PR #1904, opened 2016-01-13
remove grib_skip_computed

marqh:
Addressing #1889 

It is my view that although we stated we were using only coded keys, we were explicitly not doing this, as evidenced by the comment:

```
-        .. warning::
-            This currently does **not** return only the coded keys from a
-            message. This is because the gribapi functionality needed to
-            achieve this is broken, with a fix available from gribapi v1.13.0.
```

It is not trivial to pipe computed keys into an individual product definition function, given how we have structured our code.  At present we have coded our solution based on this bug, which we have explicitly warned about.  I don't think this is a practical situation for us.

If we want to have a strong limitation on using coded keys, but allow them in special exception cases, I think we need to consider how to do that effectively.  I propose this is a separate activity, following this PR

PR #1913, opened 2016-01-18
careful use of computed keys

marqh:
A new function is introduced on _Section, providing controlled access to computed keys

this makes it clear where we are using them, (currently GDT40 and GDT41)

Alternative to #1904

### DIFF
--- a/lib/iris/fileformats/grib/_message.py
+++ b/lib/iris/fileformats/grib/_message.py
@@ -1,4 +1,4 @@
-# (C) British Crown Copyright 2014 - 2015, Met Office
+# (C) British Crown Copyright 2014 - 2016, Met Office
 #
 # This file is part of Iris.
 #
@@ -376,7 +376,7 @@ class _Section(object):
                        'satelliteNumber', 'instrumentType',
                        'scaleFactorOfCentralWaveNumber',
                        'scaledValueOfCentralWaveNumber',
-                       'longitudes', 'latitudes', 'distinctLatitudes')
+                       'longitudes', 'latitudes')
         if key in vector_keys:
             res = gribapi.grib_get_array(self._message_id, key)
         elif key == 'bitmap':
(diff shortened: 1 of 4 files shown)


marqh:
@ajdawson does this seem like a clean approach to handling gdt40/41 with GRIB API versions including 1.14?

lib/iris/fileformats/grib/_message.py
```diff
@@ -376,7 +376,7 @@ def _get_key_value(self, key):
                        'satelliteNumber', 'instrumentType',
                        'scaleFactorOfCentralWaveNumber',
                        'scaledValueOfCentralWaveNumber',
-                       'longitudes', 'latitudes', 'distinctLatitudes')
+                       'longitudes', 'latitudes')
```
ajdawson: Are `longitudes` and `latitudes` computed or coded keys?

marqh: > Are longitudes and latitudes computed or coded keys?

This distinction appears to be template dependent.  Due to bugs (and perhaps other factors)   it is also grib api version dependent.

So, using grib api 1.14:
 - in GDT4 & GDT5, latitudes are reported as coded;
   - this is true, there is an array of numbers
 - in GDt40 & GDT41 latitudes are report as computed;
   - this seems true, these are worked out

I originally included an exception report is 'get_computed_key' retrieved a coded key, but this was grib api version intolerant so I removed it.

At version 1.14, this approach appears to mandate that the new get_computed_key function is used whenever a gribapi computation is used.
This seems like a safe, but not so gribapi version critical interpretation for our use

ajdawson: OK, so having these listed in the _Section class is safe, because they can only be accessed as coded keys if gribapi knows this to be true for the particular message being handled.

rhattersley:
FWIW, I suggested this addition of a method to the _Section class to @marqh offline so I'm implicitly :+1: for this approach.

ajdawson:
I'm happy with this I think. Can you clarify the behavior under earlier gribapi version @marqh? Is it just that these will erroneously list computed keys when asked for coded keys only? If so the only problem is that code may be developed using older libraries and the mechanism for catching it not triggered, which can be mitigated by using later gribapi versions for testing. Have I got that right?

marqh:
> Is it just that these will erroneously list computed keys when asked for coded keys only? 

exactly; this behaviour is improved/fixed at 1.13

> If so the only problem is that code may be developed using older libraries and the mechanism for catching it not triggered, which can be mitigated by using later gribapi versions for testing. Have I got that right?

yes

ajdawson:
OK, then I'm going to merge this on the assumption that the CI configuration is updated to work with a version of gribapi >= 1.13.

PR #1908, opened 2016-01-14
Check aux coords match before concatenating

lbdreyer:
Currently, for concatenation, only AuxCoord metadata is being checked. This allows two cubes with different AuxCoord points to be concatenated. For an example case, see the integration test I added.

This PR ensures the AuxCoords match before the concatenation can happen.

PR #1915, opened 2016-01-21
Allow concatenation of cubes with anonymous dimensions

lbdreyer:
This is dependent on PR #1908

### DIFF
--- a/lib/iris/_concatenate.py
+++ b/lib/iris/_concatenate.py
@@ -27,6 +27,7 @@ from collections import defaultdict, namedtuple
 
 import biggus
 import numpy as np
+import warnings
 
 import iris.coords
 import iris.cube
@@ -37,8 +38,6 @@ from iris.util import guess_coord_axis, array_equal, unify_time_units
 # TODO:
 #
 #   * Cope with auxiliary coordinate factories.
-#
-#   * Deal with anonymous dimensions.
 #
 #   * Allow concatentation over a user specified dimension.
 #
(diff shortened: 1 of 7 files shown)


lib/iris/_concatenate.py
```diff
@@ -431,9 +431,8 @@ def match(self, other, error_on_mismatch):
 
         # Check if either cube is anonymous.
         if self.anonymous or other.anonymous:
-            msg = ('Dimensions differ: one or both cubes have anonymous '
-                   'dimensions')
-            msgs.append(msg)
+            msg = ('One or both cubes have anonymous dimensions')
```
cpelley: Its perfectly valid to have anonymous dimensions and since concatenation now handles them so why issue the warning?
Is iris making an assumption under these circumstances which might lead it to doing the 'wrong' thing, hence pointing it out to the user in the form of a warning?

In our application that we provide to our users, we currently have to filter certain unnecessary warnings issues by certain libraries so as to be left with those which the user does need be aware of.  Is there a purpose to this warning that I'm missing?

Cheers

cpelley:
ping :)

lib/iris/_concatenate.py
lbdreyer: I think I used a warning as I was concerned about someone trying to concatenate a cube that wasn't fully defined. However, I realise this is a risk with any concatenation, regardless of whether there is an anonymous dimension.

I will put up another pull request that removes this warning.

PR #1917, opened 2016-01-26
ENH: LBPROC IMDI

cpelley:
Currently iris does not accept LBPROC values of IMDI.  The UM F03 spec specifies:
"It should be 0 if no processing has been done, otherwise the relevant numbers from the list below are added together"

```Python
...
  File "/home/h05/cpelley/git/iris/lib/iris/fileformats/pp.py", line 694, in __init__
    raise ValueError('Negative numbers not supported with '
ValueError: Negative numbers not supported with splittable integers object
```

In reality, real datasets can be found with its value set to IMDI.  I see no problem in interpreting LPBROC of IMDI as a 0.  This PR does exactly this.

### DIFF
--- a/lib/iris/fileformats/ff.py
+++ b/lib/iris/fileformats/ff.py
@@ -1,4 +1,4 @@
-# (C) British Crown Copyright 2010 - 2015, Met Office
+# (C) British Crown Copyright 2010 - 2016, Met Office
 #
 # This file is part of Iris.
 #
@@ -32,7 +32,8 @@ from iris.fileformats._ff_cross_references import STASH_TRANS
 from . import pp
 
 
-IMDI = -32768
+IMDI = pp.IMDI
+
 
 FF_HEADER_DEPTH = 256      # In words (64-bit).
 DEFAULT_FF_WORD_DEPTH = 8  # In bytes.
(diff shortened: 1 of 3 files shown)


rhattersley:
> I see no problem in interpreting LPBROC of IMDI as a 0.

If IMDI should be interpreted as 0 then that needs to be made clear in F3. IMDI is effectively "I don't know what's been done", which isn't the same as "nothing has been done".

cpelley:
OK thanks @rhattersley I'll run it passed them and post an update.

rhattersley:
> I'll run it passed them and post an update

That'd be very helpful - thanks @cpelley. :+1:

cpelley:
It seems there is no appetite to cater for this case.  I'll close this PR.

Does beg the question though, people should have a means to apply a callback of sorts on the field so that they can load it into iris (they know what it means), why should they be forced to change their source data?

rhattersley:
> It seems there is no appetite to cater for this case

Hi @cpelley - just to be clear, *who* has no appetite? The guardians of F3?

cpelley:
Apologies for my vagueness, I meant both colleagues and 'guardians of F03'.
Thanks for your interest on this in any case, appreciated.

PR #1918, opened 2016-01-27
unify time dtype awareness

marqh:
make unify time units utilitiy function dtype aware and careful to preserve dtype for coordinate instances

### DIFF
--- a/lib/iris/tests/unit/util/test_unify_time_units.py
+++ b/lib/iris/tests/unit/util/test_unify_time_units.py
@@ -1,4 +1,4 @@
-# (C) British Crown Copyright 2014 - 2015, Met Office
+# (C) British Crown Copyright 2014 - 2016, Met Office
 #
 # This file is part of Iris.
 #
@@ -118,6 +118,14 @@ class Test(tests.IrisTest):
         expected = 'hours since 1970-01-01 00:00:00'
         unify_time_units(cubelist)
         self._common(expected, cubelist)
+
+    def test_time_dtype(self):
+        cubelist = self.simple_1d_time_cubes()
+        for cube in cubelist:
+            cube.coord('time').points = cube.coord('time').points.astype(int)
+        unify_time_units(cubelist)
+        for cube in cubelist:
+            self.assertEqual(cube.coord('time').points.dtype, int)
 
 
 if __name__ == '__main__':
--- a/lib/iris/tests/unit/util/test_unify_time_units.py
+++ b/lib/iris/tests/unit/util/test_unify_time_units.py
@@ -127,6 +127,52 @@ class Test(tests.IrisTest):
         for cube in cubelist:
             self.assertEqual(cube.coord('time').points.dtype, int)
 
+    def test_int32_overflow_points(self):
+        coord_points = [9000]
+        data_points = [273]
+        reftimes = ['seconds since 1970-01-01 00:00:00',
+                    'days since 2015-01-27 10:21:17']
+        list_of_cubes = []
+        for reftime in reftimes:
+            cube = iris.cube.Cube(np.array(data_points, dtype=np.float32),
+                                  standard_name='air_temperature',
+                                  units='K')
+            unit = cf_units.Unit(reftime, calendar='gregorian')
+            coord = iris.coords.DimCoord(points=np.array(coord_points,
+                                                         dtype=np.int32),
+                                         standard_name='time',
+                                         units=unit)
+            cube.add_dim_coord(coord, 0)
+            list_of_cubes.append(cube)
+        expected = ('Unifying units would have altered the points array of the'
+                    ' time coord')
+        with self.assertRaisesRegexp(ValueError, expected):
+            unify_time_units(list_of_cubes)
+
+    def test_int32_overflow_bounds(self):
+        coord_points = [8000]
+        data_points = [273]
+        bounds = [7000, 9000]
+        reftimes = ['seconds since 1970-01-01 00:00:00',
+                    'days since 2015-01-27 10:21:17']
+        list_of_cubes = []
+        for reftime in reftimes:
+            cube = iris.cube.Cube(np.array(data_points, dtype=np.float32),
+                                  standard_name='air_temperature',
+                                  units='K')
+            unit = cf_units.Unit(reftime, calendar='gregorian')
+            coord = iris.coords.DimCoord(points=np.array(coord_points,
+                                                         dtype=np.int32),
+                                         bounds=np.array(bounds,
+                                                         dtype=np.int32),
+                                         standard_name='time',
+                                         units=unit)
+            cube.add_dim_coord(coord, 0)
+            list_of_cubes.append(cube)
+        expected = ('Unifying units would have altered the bounds array of the'
+                    ' time coord')
+        with self.assertRaisesRegexp(ValueError, expected):
+            unify_time_units(list_of_cubes)
 
 if __name__ == '__main__':
     tests.main()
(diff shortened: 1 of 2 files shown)


ajdawson:
Does this change introduce a problem where the resulting times might not be correct? I'm imagining a unit something like 'seconds since 1800-01-01 00:00:0.0'. This unit will require a 64-bit float in order to represent today's date. If all cubes are to have their time dimensions converted to this unit, and some of them represent today and have 32-bit float dtype, then won't the resulting coordinates be incorrect?

```python
from datetime import datetime

import cf_units
import numpy as np

base_unit = cf_units.Unit('seconds since 1800-01-01 00:00:0.0', calendar='gregorian')
today = datetime(2015, 1, 27, 10, 21, 17)
today_double = base_unit.date2num(today)
print(base_unit.num2date(today_double))
print(base_unit.num2date(np.float32(today_double)))
```

resulting in

```
2015-01-27 10:21:17
2015-01-27 10:25:04
```

marqh:
Hi @ajdawson 

in the case you describe, there is a possibility of problems, but in this implementation, I think this is protected

The numerical conversion of the data values is done by convert_unit.  This will return a float.  If the dtype of the original coord was also a float, then there is no issue

If the dtype of the original coordinate was an int, then the float will be cast to an int.

All unify time units is doing is altering the epoch so that it will be consistent.  If the input is an int and the epoch is changed, for the same core unit, then the implications on precision are minute, to the point of being ignored.

If you'd like a test in the the tests to illustrate this fact, then let me know, and I can try to construct one

cheers
mark

ajdawson:
I actually think the current behaviour is buggy, your new version introduces new buggy behaviour. Take the following example:

```python

from datetime import datetime

from cf_units import Unit
from iris.cube import Cube, CubeList
from iris.util import unify_time_units
import numpy as np

today = datetime(2015, 1, 27, 10, 21, 17)

# construct two cubes, with time dimensions that have 1 entry, both
# representing the same date but with different units
u0 = Unit('seconds since 1800-01-01 00:00:0.0', calendar='gregorian')
u1 = Unit('days since 2015-01-27 10:21:17', calendar='gregorian')
t0 = DimCoord([u0.date2num(today)], standard_name='time', units=u0)
# Note: type-cast here
t1 = DimCoord([np.float32(u1.date2num(today))], standard_name='time', units=u1)
c0 = Cube([2.716], standard_name='air_temperature', units='K')
c0.add_dim_coord(t0, 0)
c1 = Cube([1.821], standard_name='air_temperature', units='K')
c1.add_dim_coord(t1, 0)

# contruct a cube list and print information about each cube's time
# dimension
cl = CubeList([c0, c1])
print('before unify_time_units')
for i in (0, 1):
    print('cube {}'.format(i))
    print('    {!s}'.format(cl[i].coord('time')))
    print('    {!r}'.format(cl[i].coord('time')))
    print('    {!s}'.format(cl[i].coord('time').points.dtype))

# run unify_time_units on the cube list to set a common epoch, then print
# information about each cube's time dimension again
unify_time_units(cl)
print('')
print('after unify_time_units')
for i in (0, 1):
    print('cube {}'.format(i))
    print('    {!s}'.format(cl[i].coord('time')))
    print('    {!r}'.format(cl[i].coord('time')))
    print('    {!s}'.format(cl[i].coord('time').points.dtype))
```

Against current master and your branch this produces the following result, which shows the time represented by the second cube's time coordinate has been changed:

```
Before unify_time_units():
cl[0]
    DimCoord([2015-01-27 10:21:17], standard_name='time', calendar='gregorian')
    DimCoord(array([  6.78701648e+09]), standard_name='time', units=Unit('seconds since 1800-01-01 00:00:0.0', calendar='gregorian'))
    float64
cl[1]
    DimCoord([2015-01-27 10:21:17], standard_name='time', calendar='gregorian')
    DimCoord(array([ 0.], dtype=float32), standard_name='time', units=Unit('days since 2015-01-27 10:21:17', calendar='gregorian'))
    float32

After unify_time_units():
cl[0]
    DimCoord([2015-01-27 10:21:17], standard_name='time', calendar='gregorian')
    DimCoord(array([  6.78701648e+09]), standard_name='time', units=Unit('seconds since 1800-01-01 00:00:0.0', calendar='gregorian'))
    float64
cl[1]
    DimCoord([2015-01-27 10:25:04], standard_name='time', calendar='gregorian')
    DimCoord(array([  6.78701670e+09], dtype=float32), standard_name='time', units=Unit('seconds since 1800-01-01 00:00:0.0', calendar='gregorian'))
    float32
```

So this is bad, but not to do with your proposed change. Next consider the same example. but with the type of the second time coordinate set to `np.int32`. From your branch I get this:

```
cl[0]
    DimCoord([2015-01-27 10:21:17], standard_name='time', calendar='gregorian')
    DimCoord(array([  6.78701648e+09]), standard_name='time', units=Unit('seconds since 1800-01-01 00:00:0.0', calendar='gregorian'))
    float64
cl[1]
    DimCoord([2015-01-27 10:21:17], standard_name='time', calendar='gregorian')
    DimCoord(array([0], dtype=int32), standard_name='time', units=Unit('days since 2015-01-27 10:21:17', calendar='gregorian'))
    int32

After unify_time_units():
cl[0]
    DimCoord([2015-01-27 10:21:17], standard_name='time', calendar='gregorian')
    DimCoord(array([  6.78701648e+09]), standard_name='time', units=Unit('seconds since 1800-01-01 00:00:0.0', calendar='gregorian'))
    float64
cl[1]
    DimCoord([1731-12-13 20:45:52], standard_name='time', calendar='gregorian')
    DimCoord(array([-2147483648], dtype=int32), standard_name='time', units=Unit('seconds since 1800-01-01 00:00:0.0', calendar='gregorian'))
    int32
```

which is very bad, but from master I don't get any change to the time (but the new coordinate has dtype `np.float64`).

The existing buggy behaviour needs to be fixed, and I'm not in favour of your change until the existing behaviour can be fixed. Even then I'm not convinced you can guarantee the safety of casting each coordinate back to its original dtype.

rhattersley:
(@ajdawson - thanks for taking the time to investigate so thoroughly!)

marqh:
Hi @ajdawson 

cheers for the input

I can't explicitly reproduce your test case on my machine, it passes, but I can reproduce a similar one (maybe this is to do with architecture, numpy etc.)

Anyhow, i think this is to do with overrun, particularly with things like int32

I've updated the behaviour to check for this and added tests


> The existing buggy behaviour needs to be fixed, and I'm not in favour of your change until the existing behaviour can be fixed. 

I'm not sure I agree with this.  I think this PR is additive and helpful and that the other behaviour is independent: should also be fixed, but perhaps not in this PR or as a prerequisite to it. (imho, as always)

> Even then I'm not convinced you can guarantee the safety of casting each coordinate back to its original dtype.

I think I have done this (by throwing an exception if there's a problem)

ajdawson:
>I'm not sure I agree with this. I think this PR is additive and helpful and that the other behaviour is independent

I don't believe it is independent, the second test case I provided fails with your PR (raises an exception telling me the result would be wrong)* but without your PR this case is working correctly, and therefore could be in user code that also functions correctly. You PR could break this existing working code.

*[Ubuntu 14.04, Python 2.7.10 (Continuum analytics), NumPy v1.8.2]

ajdawson:
A good question to ask is what is desired from unifying time units? Would we prefer an error and be prevented from doing so if the type cannot be preserved, or would we prefer the new time coordinate to have a new dtype (and maybe a warning)? We could even envision a situation where the typecast is done only if needed.

marqh:
> I don't believe it is independent, the second test case I provided fails with your PR

in that case I will continue and try to fix the bug i did not know about alongside the one I am chasing

> A good question to ask is what is desired from unifying time units? 

indeed.  The primary use case from my point of view is to enable concatenate or merge to work effectively.  To do so, relevant coordinates must meet conditions, including having consistent dtype and, for temporal coordinates, the same epoch and calendar

In these cases, succeed correctly or throw an exception are the only useful results

marqh:
i will not be putting any further work into this pull request

PR #1923, opened 2016-01-29
Add windspeed example code

djkirkham:
Adding an example to the Iris Gallery.

The example plots windspeed direction using quiver to plot arrows. It also demonstrates normalising the length of the arrows.

Note: change relies on pull request https://github.com/SciTools/iris-sample-data/pull/12


marqh:
@djkirkham I think that some locational information would help

a high resolution coastline showing the lake would help

I would also recommend a nice title, explaining what is shown and where it is

docs/iris/example_code/Meteorology/wind_speed.py
```diff
@@ -0,0 +1,69 @@
...
+    windspeed.rename( 'windspeed' )
+
+    x = ulon.points
+    y = uwind.coord('latitude').points
+    u = uwind.data
+    v = vwind.data
+
+    # Plot the wind speed as a contour plot
+    plt.figure()
+    plt.axes(projection=cartopy.crs.PlateCarree())
```
dkillick: Typically we would write this as `projection=ccrs.PlateCarree()`, having imported cartopy.crs specifically at the top of the module: `import cartopy.crs as ccrs`. It's just a stylistic thing, but it's good to maintain a consistent style :smile:

docs/iris/example_code/Meteorology/wind_speed.py
```diff
@@ -0,0 +1,69 @@
...
+    plt.quiver(x, y, u, v, pivot = 'middle')
+
+    iplt.show()
+
+    # Normalise the data for uniform arrow size
+    u_norm = u / np.sqrt( u ** 2.0 + v ** 2.0 )
+    v_norm = v / np.sqrt( u ** 2.0 + v ** 2.0 )
+
+    plt.figure()
+    plt.axes(projection=cartopy.crs.PlateCarree())
```
dkillick: As above

docs/iris/example_code/Meteorology/wind_speed.py
```diff
@@ -0,0 +1,69 @@
...
+
+    uwind = iris.load_cube(infile, 'x_wind')
+    vwind = iris.load_cube(infile, 'y_wind')
+
+    ulon = uwind.coord('longitude')
+    vlon = vwind.coord('longitude')
+    
+    # The longitude points go from 180 to 540, so subtract 360 from them
+    ulon.points = ulon.points - 360.0  
+    vlon.points = vlon.points - 360.0
```
dkillick: As you are effectively operating in-place on `ulon.points` and `vlon.points` here, you could use the in-place operator `-=`. For example: `ulon.points -= 360.0`

ajdawson:
This example has to extract data from a cube and coordinates in order to plot, and has thus exposed our lack of support for vector plotting in `iris.plot`... we should get round to doing something about that fairly soon.

docs/iris/example_code/Meteorology/wind_speed.py
```diff
@@ -0,0 +1,69 @@
...
+    windspeed = ( uwind ** 2 + vwind ** 2 ) ** 0.5 
+    windspeed.rename( 'windspeed' )
+
+    x = ulon.points
+    y = uwind.coord('latitude').points
+    u = uwind.data
+    v = vwind.data
+
+    # Plot the wind speed as a contour plot
+    plt.figure()
```
dkillick: You don't actually need this line. Remember from the matplotlib course - if an axes doesn't have a current figure to be written onto then it will go away and create a new one :smiley:

docs/iris/example_code/Meteorology/wind_speed.py
```diff
@@ -0,0 +1,69 @@
...
+The second plot uses normalised data to produce arrows with a uniform size
+
+"""
+
+import matplotlib.pyplot as plt
+import numpy as np
+
+import iris
+import iris.coord_categorisation
+import iris.plot as iplt
```
dkillick: You only use `iplt` to show your plots. You could remove this import entirely and use `qplt.show` instead to show your plots.

dkillick:
> a high resolution coastline showing the lake would help

@marqh I thought this too. The downside will be if adding code to do this makes it harder to follow the code that produces the example plot. If it doesn't I'm :+1: for this idea.

docs/iris/example_code/Meteorology/wind_speed.py
```diff
@@ -0,0 +1,77 @@
+"""
+Plotting wind direction using quiver
+===========================================================
+
+This example demonstrates using quiver to plot wind speed
```
marqh: contours and wind direction arrows from wind vector component input data.  The vector components are co-located in space in this case.

docs/iris/example_code/Meteorology/wind_speed.py
```diff
@@ -0,0 +1,77 @@
...
+
+The second plot uses normalised data to produce arrows with a uniform size
```
marqh: For the second plot, the data used for the arrows is normalised to produce arrows with a uniform size on the plot.

docs/iris/example_code/Meteorology/wind_speed.py
```diff
@@ -0,0 +1,77 @@
...
+    # Normalise the data for uniform arrow size
+    u_norm = u / np.sqrt(u ** 2.0 + v ** 2.0)
+    v_norm = v / np.sqrt(u ** 2.0 + v ** 2.0)
+
+    ax = plt.axes(projection=ccrs.PlateCarree())
+    ax.add_feature(lakes)
+
+    qplt.contourf(windspeed, 20)
+
+    plt.quiver(x, y, u_norm, v_norm, pivot='middle')
```
ajdawson: I think this should probably have a `transform` keyword argument since the axes are a `GeoAxes` instance.

marqh:
hi @djkirkham 

this is looking good now, tests are passing

I have made two suggestions for the introductory docstring.  if you're happy, please include these, then rebase to squish your commits

mark

marqh:
> This example has to extract data from a cube and coordinates in order to plot, and has thus exposed our lack of support for vector plotting in iris.plot... we should get round to doing something about that fairly soon.

@ajdawson indeed...
something not altogether unlike:
https://github.com/marqh/vectorfield
perhaps, maybe, one day

marqh:
@djkirkham excellent, many thanks for the contribution

ajdawson:
>something not altogether unlike: https://github.com/marqh/vectorfield perhaps, maybe, one day

I wasn't thinking of anything like that actually, just a plot wrapper for quiver that extracts the transform and coordinates just like we do for other plot types. A much smaller amount of effort would be required for this.

PR #1929, opened 2016-02-03
new translations

marqh:
GRIB2 precipitation_amount
and
implied height for stash m01s03i227

replaces #1924 

derived from
https://github.com/metarelate/metOcean/issues/61
https://github.com/metarelate/metOcean/issues/70

### DIFF
--- a/lib/iris/fileformats/grib/_grib_cf_map.py
+++ b/lib/iris/fileformats/grib/_grib_cf_map.py
@@ -1,4 +1,4 @@
-# (C) British Crown Copyright 2013 - 2015, Met Office
+# (C) British Crown Copyright 2013 - 2016, Met Office
 #
 # This file is part of Iris.
 #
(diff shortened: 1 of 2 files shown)


rhattersley:
Perhaps it would be helpful to embed the corresponding metarelate/metOcean commit hash in lib/iris/fileformats/grib/_grib_cf_map.py?

marqh:
> Perhaps it would be helpful to embed the corresponding metarelate/metOcean commit hash in lib/iris/fileformats/grib/_grib_cf_map.py?

I am concerned that this level of traceability would be complicated to implement, given the current functionality of the gen_translations tool whilst the searchability of metarelate enables fine grained analysis of translations and their history, e.g.
http://www.metarelate.net/metOcean/list_mappings/False/?ref=[{%22predicate%22%3A+%22%22%2C+%22rdfobject%22%3A+%22precipitation_amount%22}]
(search on precipitation_amount)

I'm don't think there is enough value in having the github commits embedded and referenced within these map files in Iris to be worth the effort to implement such referencing.

rhattersley:
> I'm don't think there is enough value in having the github commits embedded and referenced

To be clear, I'm **not** talking about including a list of commits that correspond with a change in Iris. I'm simply suggesting that the mappings in the file are created from a particular state of the metOcean database, and that state can be succinctly expressed with a _single_ commit hash.

For example, the header of `lib/iris/fileformats/grib/_grib_cf_map.py` would look like:

```python
# DO NOT EDIT: AUTO-GENERATED
"""
Provides GRIB/CF phenomenon translations.

Generated from metarelate/metOcean @ cf419fba84a70fba5f394f1481cfcdbba28877ff

"""

```

marqh:
ah, i see; I was overcomplicating my interpretation

I think this is plausible, we'd need a small tweak to the metarelate API to get the commit sha.  I will try to get to that and get it running properly soon.

Would you be content to take this PR as is and aim for a follow up providing the commit sha?

cheers
mark

rhattersley:
> Would you be content to take this PR as is and aim for a follow up providing the commit sha?

Absolutely. :+1:

marqh:
https://github.com/metarelate/metarelate/issues/36

marqh:
thank you

PR #1934, opened 2016-02-05
WIP: Candidate PPField->Cube interface.

rhattersley:
See #1917 - ping @cpelley.

Provides a way to work with PP fields that raise errors during the conversion to Cubes. For example, fields with a negative LBPROC value give:
```python
>>> cube = iris.load_cube(path)
...
ValueError: Negative numbers not supported with splittable integers object
```
Working at the lower, PP field level we can adjust the fields before conversion:
```python
fields = iris.fileformats.pp.load(path)
def mutate(field):
    field.lbproc = 0
    return field
tweaked_fields = (mutate(field) for field in fields)
cubes = iris.cube.CubeList(iris.fileformats.pp.as_cubes(tweaked_fields))
cube = cubes.merge_cube()
```

### DIFF
--- a/lib/iris/fileformats/pp.py
+++ b/lib/iris/fileformats/pp.py
@@ -2028,6 +2028,23 @@ def load_cubes(filenames, callback=None, constraints=None):
                                        constraints=constraints)
 
 
+def as_cubes(pp_fields):
+    """
+    Convert an iterable of PP fields into an iterable of Cubes.
+
+    Args:
+
+    * pp_fields:
+        An iterable of :class:`iris.fileformats.pp.PPField`.
+
+    Returns:
+        An iterable of :class:`iris.cube.Cube`s.
+
+    """
+    return iris.fileformats.rules.as_cubes(pp_fields,
+                                           iris.fileformats.pp_rules.convert)
+
+
 def _load_cubes_variable_loader(filenames, callback, loading_function,
                                 loading_function_kwargs=None,
                                 constraints=None):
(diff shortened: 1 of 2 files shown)


cpelley:
Hi @rhattersley.  This is great!

We have a rather less than ideal solution that we had been working with in our project which patches `iris.fileformats.pp.load`:
https://code.metoffice.gov.uk/trac/ancil/browser/ants/trunk/lib/ants/fileformats/ff.py?rev=1319#L806

We have corresponding unit tests that would be directly applicable to this PR:
https://code.metoffice.gov.uk/trac/ancil/browser/ants/trunk/lib/ants/tests/fileformats/ff/test_pp2cubes.py?rev=1319

@jkettleb you will definitely like this!
@ehogan, I think this would also be applicable to you in your work as I understand it requires the filtering of fields before turning them into cubes.

Thanks @rhattersley, I'm more than happy to review when your ready or contribute to its development (your preference).

jkettleb:
This looks like a good idea.

I think it t would be neat if  `load_cubes` could call `as_cubes`.  (Less code, encapsulation of the `as_cubes` logic in one place). Do you need to call the legacy rules and callback before the dereferencing logic?  Is this order dependence the thing that makes it harder to call `as_cubes` from `load_cube`   (Obviously ignore this if you disagree its a good idea - and sorry, its quite nosy of me.)

I think `as_cubes` would work with fields files too - or did I miss something?

marqh:
please note I am continuing work on this in #1941, in case that wasn't clear

rhattersley:
Closing as the action has moved to #1941.

PR #1936, opened 2016-02-08
Test image for hovmoller and sample data path ref updated

corinnebosley:
Recent pull request has been merged which changes a data file in the sample data path.  This data file has been used in another gallery image, so this branch points to the new data and provides an updated test image for the gallery testing.


marqh:
I have looked into this data and I agree that the original data was intended to be masked to sea points only.

This plot better represents the data intent than the previous version, based on unmasked data

thank you @corinnebosley

PR #1937, opened 2016-02-09
Add add_hour function to coord_categorisation

djkirkham:
Cubes can now have an 'hour' coordinate added by calling `iris.coord_categorisation.add_hour(<cube>, <time-coord>)`.

### DIFF
--- a/lib/iris/tests/unit/coord_categorisation/test_add_hour.py
+++ b/lib/iris/tests/unit/coord_categorisation/test_add_hour.py
@@ -1,0 +1,92 @@
+# (C) British Crown Copyright 2016, Met Office
+#
+# This file is part of Iris.
+#
+# Iris is free software: you can redistribute it and/or modify it under
+# the terms of the GNU Lesser General Public License as published by the
+# Free Software Foundation, either version 3 of the License, or
+# (at your option) any later version.
+#
+# Iris is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU Lesser General Public License for more details.
+#
+# You should have received a copy of the GNU Lesser General Public License
+# along with Iris.  If not, see <http://www.gnu.org/licenses/>.
+"""
+Test coordinate categorisation function add_hour.
+"""
+
+from __future__ import (absolute_import, division, print_function)
+from six.moves import (filter, input, map, range, zip)  # noqa
+
+# import iris tests first so that some things can be initialised before
+# importing anything else
+import iris.tests as tests
+
+import cf_units
+import numpy as np
+
+import iris
+import iris.coord_categorisation as ccat
+
+
+class Test_add_hour(tests.IrisTest):
+    def setUp(self):
+        # make a series of 'hour numbers' for the time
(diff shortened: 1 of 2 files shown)


rhattersley:
Thanks @djkirkham. The new function looks :+1:

It'd be nicer to do your testing as unit tests in `lib/iris/tests/unit/coord_categorisation/test_add_hour.py` instead of the legacy `lib/iris/tests/test_coord_categorisation.py`. Are you up for tackling that?

rhattersley:
NB. There is a strange error with ncdump that needs resolving.

djkirkham:
> It'd be nicer to do your testing as unit tests in lib/iris/tests/unit/coord_categorisation/test_add_hour.py

Are you sure you want me to add a new file? There is already a test class for add_day_of_year in lib/iris/tests/unit/coord_categorisation/test_add_categorised_coord.py. Shall I add my test code in there?

The ncdump issue seems to have gone away. I don't think it was anything to do with my code, just the vagaries of Travis. It's now encountering TimeoutExceptions...

rhattersley:
> Are you sure you want me to add a new file? There is already a test class for add_day_of_year in lib/iris/tests/unit/coord_categorisation/test_add_categorised_coord.py. Shall I add my test code in there?

Thanks for checking :+1:,  but a new file is fine - it's the [standard way](http://scitools.org.uk/iris/docs/latest/developers_guide/tests.html#functions) we structure the unit tests. It makes it easy to find the tests for a given piece of code, and vice versa.

djkirkham:
Cool, I'll make the changes you suggested.

djkirkham:
Rebased due to reversion of file causing header test errors (copyright year)

lib/iris/tests/unit/coord_categorisation/test_add_hour.py
```diff
@@ -0,0 +1,93 @@
...
+        self.assertEqual(len(result_coords), 1, msg)
+
+    def test_basic(self):
+        coord_name = 'my_hour'
+        cube = self.cube
+        time_coord = self.time_coord
+        expected_coord = iris.coords.AuxCoord(
+            self.hour_numbers % 24, long_name=coord_name)
+
+        with warnings.catch_warnings(record=True):
```
rhattersley: I don't think you need this? (The same applies to the other two occurrences and the initial `import warnings`.)

lib/iris/tests/unit/coord_categorisation/test_add_hour.py
```diff
@@ -0,0 +1,93 @@
...
+        msg = 'Missing/incorrectly named result for add_hour'
+
+        # Specify source coordinate by name
+        cube = self.cube.copy()
+        with warnings.catch_warnings(record=True):
+            ccat.add_hour(cube, 'time', name=coord_name)
+        result_coords = cube.coords(coord_name)
+        self.assertEqual(len(result_coords), 1, msg)
+
+        # Specify source coordinate by coordinate reference
```
rhattersley: This would be better as a separate test.

lib/iris/tests/unit/coord_categorisation/test_add_hour.py
djkirkham: You're right. I used another test as a template and left that in without really thinking about it.

marqh:
excellent, thank you @djkirkham

PR #1939, opened 2016-02-11
metarelate date and git sha

marqh:
date and commit sha stamps for metarelate sourced content

as requested in #1929

### DIFF
--- a/lib/iris/fileformats/grib/_grib_cf_map.py
+++ b/lib/iris/fileformats/grib/_grib_cf_map.py
@@ -16,6 +16,12 @@
 # along with Iris.  If not, see <http://www.gnu.org/licenses/>.
 #
 # DO NOT EDIT: AUTO-GENERATED
+# Created on 12 February 2016 17:02 from 
+# http://www.metarelate.net/metOcean
+# at commit cf419fba84a70fba5f394f1481cfcdbba28877ff
+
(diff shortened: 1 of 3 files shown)


lib/iris/fileformats/grib/_grib_cf_map.py
```diff
@@ -16,6 +16,12 @@
 # along with Iris.  If not, see <http://www.gnu.org/licenses/>.
 #
 # DO NOT EDIT: AUTO-GENERATED
+# Created on 16 February 11 12:02 from
```
rhattersley: Thanks for adding this @marqh :+1: :grin: ... just one small extra favour, presumably `16 February 11` => 11th of February 2016? Could you change the date format to be less ambiguous?

tools/gen_translations.py
```diff
@@ -203,10 +210,11 @@ def build_grib_cf_map(fuseki, filename):
 
 
 def main():
+    now = datetime.utcnow().strftime('%y %B %d %H:%m')
```
rhattersley: Aha ... this is the culprit!

marqh:
> Could you change the date format to be less ambiguous?

eerk, sorry, that was somewhat unclear

i've altered this and added a check, just to protect against a race condition

rhattersley:
Super - thanks @marqh . :+1:

PR #1943, opened 2016-02-17
Fix to automatic deriving of units when saving in netcdf format

djkirkham:
Make coords with standard name 'latitude' and 'longitude' which have no coordsystem to be saved with units 'degree_north' and 'degree_east' respectively in netcdf files. Remove other unsafe unit changes.

Change some unit test expected output to recognise these changes.

### DIFF
--- a/lib/iris/fileformats/netcdf.py
+++ b/lib/iris/fileformats/netcdf.py
@@ -1118,20 +1118,16 @@ class Saver(object):
         """
 
         units = str(coord.units)
-
-        # TODO: Use #61 to get the units.
-        if isinstance(coord.coord_system, iris.coord_systems.GeogCS):
-            if "latitude" in coord.standard_name:
+        # Set the 'units' of 'latitude' and 'longitude' coordinates specified
+        # in 'degrees' to 'degrees_north' and 'degrees_east' respectively,
+        # as defined in the CF conventions for netCDF files: sections 4.1 and
+        # 4.2.
+        if ((isinstance(coord.coord_system, iris.coord_systems.GeogCS) or
+                coord.coord_system is None) and coord.units == 'degrees'):
+            if coord.standard_name == "latitude":
                 units = 'degrees_north'
-            elif "longitude" in coord.standard_name:
+            elif coord.standard_name == "longitude":
                 units = 'degrees_east'
-
-        elif isinstance(coord.coord_system, iris.coord_systems.RotatedGeogCS):
-            units = 'degrees'
-
-        elif isinstance(coord.coord_system,
-                        iris.coord_systems.TransverseMercator):
-            units = 'm'
 
         return coord.standard_name, coord.long_name, units
 
--- a/lib/iris/tests/results/netcdf/netcdf_save_ndim_auxiliary.cdl
+++ b/lib/iris/tests/results/netcdf/netcdf_save_ndim_auxiliary.cdl
@@ -36,11 +36,11 @@ variables:
 		rlon:standard_name = "grid_longitude" ;
 		rlon:long_name = "rotated longitude" ;
 	float lat(rlat, rlon) ;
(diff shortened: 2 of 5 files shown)


marqh:
please add a what's new contribution for this change

lib/iris/fileformats/netcdf.py
```diff
@@ -1119,20 +1119,13 @@ def _cf_coord_identity(self, coord):
 
         units = str(coord.units)
 
-        # TODO: Use #61 to get the units.
-        if isinstance(coord.coord_system, iris.coord_systems.GeogCS):
-            if "latitude" in coord.standard_name:
+        if ((isinstance(coord.coord_system, iris.coord_systems.GeogCS) or
```
marqh: perhaps an in-line comment here would help a little.  I would suggest

Set the 'units' of 'latitude' and 'longitude' coordinates specified in 'degrees' to 'degrees_north' and 'degrees_east' respectively, as defined in the CF conventions for netCDF files: sections 4.1 and 4.2.

lib/iris/tests/results/netcdf/netcdf_save_ndim_auxiliary.cdl
```diff
@@ -36,11 +36,11 @@ variables:
 		rlon:standard_name = "grid_longitude" ;
 		rlon:long_name = "rotated longitude" ;
 	float lat(rlat, rlon) ;
-		lat:units = "degrees" ;
+		lat:units = "degrees_north" ;
```
marqh: this is the test result I want to see change, good

lib/iris/tests/results/netcdf/netcdf_save_wcoord.cdl
```diff
@@ -8,7 +8,7 @@ variables:
 		temp:units = "K" ;
 	int64 lon(lon) ;
 		lon:axis = "X" ;
-		lon:units = "degree_east" ;
+		lon:units = "degrees_east" ;
```
marqh: that's an oddity, but the result is better

lib/iris/tests/results/netcdf/netcdf_save_gridmapmulti.cdl
```diff
@@ -15,12 +15,12 @@ variables:
 		latitude_longitude:earth_radius = 6371229. ;
 	int64 longitude(longitude) ;
 		longitude:axis = "X" ;
-		longitude:units = "degrees_east" ;
+		longitude:units = "1" ;
```
marqh: i'm surprised by the test result changes in this file.  I would have expected all of these to stay the same

@djkirkham why do you think this has occurred?

djkirkham: It appears the units of these coords are '1', so they fail the if check. I'll look into why they have units of '1'.

djkirkham: In the test these coordinates are created from numpy arrays without specifying a unit, so it defaults to '1'.

marqh: which test is it?
may I suggest that perhaps a more useful test would be where they are created with a units of 'degrees'?

djkirkham: It's this test: https://github.com/SciTools/iris/blob/30f3e2dc62308f1abdffa055e177ad746ef6d075/lib/iris/tests/test_netcdf.py#L599

> may I suggest that perhaps a more useful test would be where they are created with a units of 'degrees'?

I agree

marqh:
looking good @djkirkham I think this is ready to be adopted

please may you squish the commits?

marqh:
excellent, thanks @djkirkham

PR #1956, opened 2016-03-08
Patch page 404 mailing list url replaced

corinnebosley:
@pelson: 
I have changed the mailing list url (that was harder to find than I expected, but I feel good about myself now).

Hope this is ok.

### DIFF
--- a/docs/iris/src/developers_guide/gitwash/this_project.inc
+++ b/docs/iris/src/developers_guide/gitwash/this_project.inc
@@ -2,4 +2,4 @@
 .. _`iris`: http://scitools.org.uk/iris
 .. _`iris github`: http://github.com/SciTools/iris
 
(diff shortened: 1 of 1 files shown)


ajdawson:
I thought this was auto-generated from [docs/iris/src/developers_guide/gitwash_build.sh](https://github.com/SciTools/iris/blob/master/docs/iris/src/developers_guide/gitwash_build.sh)...

corinnebosley:
Is it? Should I have changed something else instead?

ajdawson:
My understanding of gitwash is that the project name, url and mailing list are specified when you build the dev guide using gitwash. If you change the mailing list in the wrapper script I mentioned then run it, I think it should regenerate the file you have changed with the correct mailing list address.

ajdawson:
Closing after merge of #1957 which fixes the same issue.

PR #1957, opened 2016-03-08
Patch page 404 url replaced in the correct location.

corinnebosley:
@ajdawson @pelson 
How about this then?  (Feeling a bit less good about myself now.)

### DIFF
--- a/docs/iris/src/developers_guide/gitwash/development_workflow.rst
+++ b/docs/iris/src/developers_guide/gitwash/development_workflow.rst
@@ -334,7 +334,7 @@ Rewriting commit history
 
    Do this only for your own feature branches.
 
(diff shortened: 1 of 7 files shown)


ajdawson:
I think this is the correct place to make the change, but you'll need to run the script now you have changed it, and then commit the resulting changes, which should be just the updated mailing list url (it is possible there may be other minor changes due to upstream gitwash changes, although it doesn't usually change much or often).

corinnebosley:
How about this?

ajdawson:
Looks good :+1:.

This should fix things for the next release. To close #1954 it may be necessary to fix this in the most up-to-date version of the live online documentation.

PR #1959, opened 2016-03-09
Pp rules onepath

pp-mo:
This effort re-combines the 'vector' and 'scalar' rules approaches, as discussed in [this comment] (https://github.com/SciTools/iris/pull/1953#issuecomment-193329427)

This re-reroutes the 'scalar' funcitons "_convert_scalar_time/vertical_coords" via the the equivalent 'vector' routines "_convert_time/vertical_coords" .

The resulting new combined routines should be compared with both the old versions.
The intention is that the operation of the 'scalar' cases should be completely unaffected.
The existing 'vector' routines *have* been modified, to incorporate the additions to the rules logic in the scalar routines since the two groups were split.

There are no new tests here, it is just a refactoring.
It might be approriate to test the new 'year' routine though.


pp-mo:
> The intention is that the operation of the 'scalar' cases should be completely unaffected

This might be true anyway, but to make sure the "_convert_time_coords" routine has a new control flag ```vector_operation```, that allows the scalar call behaviour to work differently from a vector style call.  This makes the routine work on plain values instead of arrays.

In practice, none of the basic rules logic differs (since I updated the vector one).
The only difference is that the 'non-vector' call will not attempt to use the '_reshape_vector_args' routine to broadcast the input metadata, or the _reduce_points_and_bounds' routine to collapse duplicate values.  

This seems like it may be only a technical difference, so it might still be possible to remove the 'vector_operation' control altogether.
In fact, I'm no longer sure why we made the 'vector' routines a separate path in the first place...
I think I recall that there were doubts about the exact equivalence in some cases.
I also think that only the **_time_**  coords were potentially controversial?
  - the time, forecast_period and forecast_reference_time all have the potential to vary, which leads to different decisions about which may be DimCoord and which AuxCoord, which possibly the vectorised code may force in a way that was not wanted.

Vertical vector coords were added here : #1395
..and time coords here : #1394

I'm still not entirely sure about this.
@rhattersley @bjlittle any clues you recognise here ?

pp-mo:
##### NOTE:
(mostly to self!)
it would make more sense to define 'year' as a function 'date2year', in parallel to 'date2hours', and vectorise it when needed in the same way.

rhattersley:
Not had time to look into the equivalence issues yet, but do you have a handle on any performance impact?

pp-mo:
Thanks for looking @rhattersley 

Please don't get too involved with minute details of this, as I'm still investigating some improvements + simplifications....

I've been pursuing any problems generated if you bypass the special 'scalar' code in the time coords rules, and simply use the _same_ (vector) code for all cases.
In fact this does seem to work without producing any differences in the general tests, and it makes the code simpler, so I'm very hopeful about that.

Unfortunately I'm not quite ready to post that yet :
  * I had to fix some knockon problems in certain tests, which took a lot of effort sorting out
  * This will make it possible to remove the 'do_vector' arg altogether again, which will put the code closer to where it started again, but I haven't integrated that yet.

I'm also thinking that, once that principle is accepted, I can also combine the repetitive use of calls to ```_reduce_points_and_bounds```, ```_dim_or_aux``` and ```coords_and_dims.append```, which should make all the rules code much more readable.

*Meanwhile* I must also run some performance benchmarking, as you suggest

rhattersley:
It's a lot to take on! I'm impressed @pp-mo ! Thank you! :+1:

pp-mo:
Content of latest commits : 
  * _Firstly_, re-combined the two scalar/vector invocation paths for ```_convert(_scalar)_time/vertical_coords```
    * but we retain this distinction internally within the 'time' conversion, since testing revealed that treating scalars as arrays had a nasty impact on execution speed.
    * ```_convert_time_coords``` now detects internally when it has scalar/vector inputs, and disables/enables the extra operations needed
  * _Secondly_, re-combined the exisiting scalar/vector tests -- see the **latest** commit for this.
    * There was a lot of duplication, which this removes.
    * A quick comparison between the 'old' scalar tests and the 'new enhanced' vector ones (before + after the latest commit) should reveal that all the previous testcases are now combined into one routine in both (i.e. 'time' and 'vertical) cases.

pp-mo:
#### Performance testing summary
Adding the following to the standard metrics ...
( these files are chosen from dataZoo, to be fairly representative of 'typical' large loads, similar to "decoupler" project inputs )
```
def large_ff_load_setup(context):
    context.filename = '/data/local/dataZoo/FF/alyku.pp0'

def large_pp_load_setup(context):
    context.filename = '/data/local/dataZoo/PP/mogrepsR/2011032906.qtey06.oper12.036.pp'

def huge_ff_load_setup(context):
    context.filename = '/data/local/dataZoo/FF/ff_variants/ff_64.ff'

 . . .

    TimeMetric(raw_load_iteration, large_ff_load_setup,
               repeat=3, name='ff_large_raw_load'),
    TimeMetric(raw_load_iteration, large_pp_load_setup,
               repeat=3, name='pp_large_raw_load'),
    TimeMetric(ff_fast_load_iteration, huge_ff_load_setup,
               repeat=3, name='ff_huge_unstructured_load'),
```

#### "Before":
**timeit-ff_large_raw_load**
        = 1.43213605881 ;  1.44219398499 ;  1.44876003265
~1.44 (seconds)

**timeit-pp_large_raw_load**
        = 0.334093093872 ;  0.335901021957 ;  0.334613084793
~0.335

**timeit-ff_huge_unstructured_load**
        = 0.83291387558 ;  0.838693857193 ;  0.832060813904
~0.835

#### "After":
**timeit-ff_large_raw_load**
    = 1.47655582428 ;  1.48132610321 ;  1.49532103539
~1.48
cf. 1.44
percentage increase : + 3%

**timeit-pp_large_raw_load**
    = 0.347663164139 ;  0.343266010284 ;  0.346315860748
~0.345
cf. 0.335
percentage increase : + 3%

**timeit-ff_huge_unstructured_load**
    = 0.841286897659 ;  0.843652963638 ;  0.831774950027
~0.838
cf. 0.835
percentage increase : ~ 0%

pp-mo:
### Decoupler impact testing

I ran the decoupler 'acceptance tests' via PBS, via my local Iris checkout, both immediately before and after the changes introduced here.
 * all tests ran ok
 * the newer code was slightly faster on both 'user' and 'sys' timings (by an insignificant amount).

#### Actual timings:

##### BEFORE (branch "pp_rules_onepath_BASIS"):
Ran 140 tests in 692.084s

real	11m36.485s
user	14m4.257s
sys	9m20.255s

##### AFTER  (branch "pp_rules_onepath"):
Ran 143 tests in 717.130s

real	12m4.482s
user	13m15.118s
sys	8m35.168s

#### interpretation:
"user" is apparently about 6% faster, and "sys" about 8% faster
This is unlikely to represent anything significant, but it is at least not worse !

marqh:
thanks for the metrics @pp-mo 

I think these represent a safe, low concern change, 

I think that we can adopt these changes for code clarity and maintainability, without adversely affecting loading significantly

@rhattersley what are your views?

lib/iris/tests/unit/fileformats/pp_rules/test__reduced_points_and_bounds.py
```diff
@@ -1,4 +1,4 @@
-# (C) British Crown Copyright 2014 - 2015, Met Office
+# (C) British Crown Copyright 2014 - 2016, Met Office
```
marqh: i think this is unnecessary now

marqh:
great stuff @pp-mo, this should improve maintenance

please check if 
test__reduced_points_and_bounds.py
can be left unchanged, then squish

i will have another look, but I am minded to merge this

lib/iris/tests/unit/fileformats/pp_rules/test__reduced_points_and_bounds.py
pp-mo: Thanks for the spot.
I think it was changed and then changed back, so the header update is the only thing left.
In which case (I'll check), that will need removing when I squish.

rhattersley:
> do you have a handle on any performance impact?

The trade-off seems OK to me. :+1:

marqh:
thank you @pp-mo

PR #1962, opened 2016-03-16
Grib dep2

marqh:
deprecations and publications for the GRIB loading, to prepare for strict_grib_load to become the default and to support as_cubes (#1941)

### DIFF
--- a/docs/iris/src/whatsnew/contributions_1.10/deprecate_2016-Mar-17_GribWrapper.txt
+++ b/docs/iris/src/whatsnew/contributions_1.10/deprecate_2016-Mar-17_GribWrapper.txt
@@ -1,0 +1,5 @@
+* deprecated:
+ iris.fileformats.grib.hindcast_workaround,
+ :class: `iris.fileformats.grib.GribWrapper`.
+ The class :class:`iris.fileformats.grib.message.GribMessage` provides
+ alternative means of working with GRIB message instances.
--- a/lib/iris/fileformats/grib/__init__.py
+++ b/lib/iris/fileformats/grib/__init__.py
@@ -1,4 +1,4 @@
-# (C) British Crown Copyright 2010 - 2015, Met Office
+# (C) British Crown Copyright 2010 - 2016, Met Office
 #
 # This file is part of Iris.
 #
@@ -44,7 +44,7 @@ from iris.exceptions import TranslationError
 from iris.fileformats.grib import grib_phenom_translation as gptx
 from iris.fileformats.grib import _save_rules
 import iris.fileformats.grib._load_convert
-from iris.fileformats.grib._message import _GribMessage
+from iris.fileformats.grib.message import GribMessage
 import iris.fileformats.grib.load_rules
 
 
@@ -55,6 +55,7 @@ __all__ = ['as_messages', 'as_pairs', 'grib_generator', 'load_cubes',
 
 #: Set this flag to True to enable support of negative forecast periods
 #: when loading and saving GRIB files.
+#: .. deprecated:: 1.10
 hindcast_workaround = False
 
 
@@ -175,8 +176,14 @@ class GribWrapper(object):
     """
     Contains a pygrib object plus some extra keys of our own.
 
+    .. deprecated:: 1.10
+
+    The class :class:`iris.fileformats.grib.message.GribMessage`
+    provides alternative means of working with GRIB message instances.
+
     """
     def __init__(self, grib_message, grib_fh=None, auto_regularise=True):
+        warnings.warn('Deprecated at version 1.10')
         """Store the grib message and compute our extra keys."""
         self.grib_message = grib_message
         deferred = grib_fh is not None
@@ -834,6 +841,12 @@ def grib_generator(filename, auto_regularise=True):
     Returns a generator of :class:`~iris.fileformats.grib.GribWrapper`
     fields from the given filename.
 
+    .. deprecated:: 1.10
+
+    The function:
+    :meth:`iris.fileformats.grib.message.GribMessage.messages_from_filename`
+    provides alternative means of obtainig GRIB messages from a file.
+
     Args:
 
     * filename (string):
@@ -849,6 +862,7 @@ def grib_generator(filename, auto_regularise=True):
         reduced grid to an equivalent regular grid.
 
     """
+    warnings.warn('Deprecated at version 1.10')
     with open(filename, 'rb') as grib_fh:
         while True:
             grib_message = gribapi.grib_new_from_file(grib_fh)
@@ -892,7 +906,7 @@ def load_cubes(filenames, callback=None, auto_regularise=True):
     """
     if iris.FUTURE.strict_grib_load:
         grib_loader = iris.fileformats.rules.Loader(
-            _GribMessage.messages_from_filename,
+            GribMessage.messages_from_filename,
             {},
             iris.fileformats.grib._load_convert.convert)
     else:
@@ -962,7 +976,7 @@ def as_pairs(cube):
 def as_messages(cube):
     """
     Convert one or more cubes to GRIB messages.
-    Returns an iterable of grib_api GRIB messages.
+    Returns an iterable of grib_api GRIB message IDs.
 
     Args:
         * cube      - A :class:`iris.cube.Cube`, :class:`iris.cube.CubeList` or
--- a/lib/iris/fileformats/grib/__init__.py
+++ b/lib/iris/fileformats/grib/__init__.py
@@ -55,6 +55,7 @@ __all__ = ['as_messages', 'as_pairs', 'grib_generator', 'load_cubes',
 
 #: Set this flag to True to enable support of negative forecast periods
 #: when loading and saving GRIB files.
+#:
 #: .. deprecated:: 1.10
 hindcast_workaround = False
 
--- a/lib/iris/fileformats/grib/message.py
+++ b/lib/iris/fileformats/grib/message.py
@@ -33,17 +33,31 @@ import numpy as np
 from iris.exceptions import TranslationError
 
 
-class _GribMessage(object):
+class _OpenFileRef(object):
     """
-    Lightweight GRIB message wrapper, containing **only** the coded keys and
-    data attribute of the input GRIB message.
+    A reference to an open file that ensures that the file is closed
+    when the object is garbage collected.
+    """
+    def __init__(self, open_file):
+        self.open_file = open_file
+
+    def __del__(self):
+        if not self.open_file.closed:
+            self.open_file.close()
+
+
+class GribMessage(object):
+    """
+    An in-memory representation of a GribMessage, providing
+    access to the :meth:`GribMessage.data` payload and the metadata
+    elements by section via the :meth:`GribMessage.sections` property.
 
     """
 
     @staticmethod
     def messages_from_filename(filename):
         """
-        Return a generator of :class:`_GribMessage` instances; one for
+        Return a generator of :class:`GribMessage` instances; one for
         each message in the supplied GRIB file.
 
         Args:
@@ -52,31 +66,53 @@ class _GribMessage(object):
             Name of the file to generate fields from.
 
         """
-        with open(filename, 'rb') as grib_fh:
-            while True:
-                offset = grib_fh.tell()
-                grib_id = gribapi.grib_new_from_file(grib_fh)
-                if grib_id is None:
-                    break
-                raw_message = _RawGribMessage(grib_id)
-                recreate_raw = _MessageLocation(filename, offset)
-                yield _GribMessage(raw_message, recreate_raw)
+        grib_fh = open(filename, 'rb')
+        # create an _OpenFileRef to manage the closure of the file handle
+        file_ref = _OpenFileRef(grib_fh)
 
-    def __init__(self, raw_message, recreate_raw):
+        while True:
+            offset = grib_fh.tell()
+            grib_id = gribapi.grib_new_from_file(grib_fh)
+            if grib_id is None:
+                break
+            raw_message = _RawGribMessage(grib_id)
+            recreate_raw = _MessageLocation(filename, offset)
+            yield GribMessage(raw_message, recreate_raw, file_ref=file_ref)
+
+    def __init__(self, raw_message, recreate_raw, file_ref=None):
         """
-
-        Args:
-
-        * raw_message:
-            The _RawGribMessage instance which should be wrapped to
-            provide the `data` attribute.
+        It is recommended to obtain GribMessage instance from the static method
+        :meth:`GribMessage.messages_from_filename`, rather than creating
+        them directly.
 
         """
+        # A RawGribMessage giving gribapi access to the original grib message.
         self._raw_message = raw_message
+        # A _MessageLocation which biggus uses to read the message data array,
+        # by which time this message may be dead and the original grib file
+        # closed.
         self._recreate_raw = recreate_raw
+        # An _OpenFileRef to keep the grib file open while this GribMessage is
+        # alive, so that we can always use self._raw_message to fetch keys.
+        self._file_ref = file_ref
 
     @property
     def sections(self):
+        """
+        Return the key-value pairs of the message keys, grouped by containing
+        section.
+
+        Key-value pairs are collected into a dictionary of
+        :class:`Section` objects. One such object is made for
+        each section in the message, such that the section number is the
+        object's key in the containing dictionary. Each object contains
+        key-value pairs for all of the message keys in the given section.
+
+        For example::
+
+            print(grib_message[4]['parameterNumber'])
+
+        """
         return self._raw_message.sections
 
     @property
@@ -123,6 +159,15 @@ class _GribMessage(object):
             fmt = 'Grid definition template {} is not supported'
             raise TranslationError(fmt.format(template))
         return data
+
+    def __getstate__(self):
+        """
+        Alter state of object prior to pickle, ensure open file is closed.
+
+        """
+        if not self._file_ref.open_file.closed:
+            self._file_ref.open_file.close()
+        return self
 
 
 class _MessageLocation(namedtuple('_MessageLocation', 'filename offset')):
@@ -273,15 +318,10 @@ class _RawGribMessage(object):
         section.
 
         Key-value pairs are collected into a dictionary of
-        :class:`_Section` objects. One such object is made for
+        :class:`Section` objects. One such object is made for
         each section in the message, such that the section number is the
         object's key in the containing dictionary. Each object contains
         key-value pairs for all of the message keys in the given section.
-
-        .. warning::
-            This currently does **not** return only the coded keys from a
-            message. This is because the gribapi functionality needed to
-            achieve this is broken, with a fix available from gribapi v1.13.0.
 
         """
         if self._sections is None:
@@ -301,7 +341,7 @@ class _RawGribMessage(object):
         """
         Group keys by section.
 
-        Returns a dictionary mapping section number to :class:`_Section`
+        Returns a dictionary mapping section number to :class:`Section`
         instance.
 
         .. seealso::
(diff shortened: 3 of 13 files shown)


lib/iris/fileformats/grib/message.py
```diff
@@ -323,16 +323,16 @@ def _get_message_sections(self):
...
                 section_keys = []
                 section = new_section
             section_keys.append(key_name)
-        sections[section] = _Section(self._message_id, section, section_keys)
+        sections[section] = Section(self._message_id, section, section_keys)
         return sections
 
 
-class _Section(object):
+class Section(object):
```
pp-mo: Needs a docstring

ajdawson:
Can you comment in some more detail why you want to make a large chunk of the deliberately private grib API public? This seems like an unusual step considering the difficulties we have had already with modifying public APIs (complex deprecation cycles and waiting for major releases).

marqh:
> Can you comment in some more detail why you want to make a large chunk of the deliberately private grib API public? This seems like an unusual step considering the difficulties we have had already with modifying public APIs (complex deprecation cycles and waiting for major releases).

of course

this is not without some consideration and a measure of trepidation

#1941 provides an important functionality extension, required by a number of users: it enables the use of 'file format fields' such as GRIB messages as feeds into the convert process.  Currently only file names are well supported

For this to be useful, we have to interact with messages.  We can do that with the legacy GRIB interface, but this PR is deprecating that, to enable deletion at a future release.

To deprecate these and support #1941 we need to be able to work with GRIB messages, and all the newer code is private.

So, I have promoted 2 Classes to the public API to enable us to do this

ajdawson:
Wehn you say it needs to be public, does it need to be visible to the end-user in order for the pairs loading to work, or just used by those modules?

marqh:
> When you say it needs to be public, does it need to be visible to the end-user in order for the pairs loading to work, or just used by those modules?

this needs to be visible to the end user in order for the pairs loading to be useful for users.

pp-mo:
> @marqh this needs to be visible to the end user in order for the pairs loading to be useful for users

I'd support that:
By  [#1941 "as_cubes"](#1941), we now have new routines (```messages_from_filename``` and ```load_pairs_from_fields```) which handle _GribMessage objects.
The whole point of that is so that you can do stuff with them (if only to filter on their contents).
So their definition must be public.

pp-mo:
As a general point, I think there should be more pointers here to go with the statements "X is deprecated" and add "go look at Y instead".
I'll add some short notes in a few places in the code...

docs/iris/src/whatsnew/contributions_1.10/deprecate_2016-Mar-17_GribWrapper.txt
```diff
@@ -0,0 +1,3 @@
+* deprecated:
+ iris.fileformats.grib.hindcast_workaround,
+ :class: `iris.fileformats.grib.GribWrapper`.
```
pp-mo: cf : [comment: add links into deprecation notices](https://github.com/SciTools/iris/pull/1962#issuecomment-197895178)
Reference ```GribMessage`` as an alternative.

lib/iris/fileformats/grib/__init__.py
```diff
@@ -175,8 +176,12 @@ class GribWrapper(object):
     """
     Contains a pygrib object plus some extra keys of our own.
 
+    .. deprecated:: 1.10
+
```
pp-mo: cf : [comment: add links into deprecation notices](https://github.com/SciTools/iris/pull/1962#issuecomment-197895178)
Reference ```GribMessage`` as an alternative.

lib/iris/fileformats/grib/__init__.py
```diff
@@ -175,8 +176,12 @@ class GribWrapper(object):
...
+
     """
     def __init__(self, grib_message, grib_fh=None, auto_regularise=True):
+        warnings.warn('Deprecated at version 1.10')
```
pp-mo: cf : [comment: add links into deprecation notices](https://github.com/SciTools/iris/pull/1962#issuecomment-197895178)
Maybe here too?

lib/iris/fileformats/grib/__init__.py
```diff
@@ -834,6 +839,8 @@ def grib_generator(filename, auto_regularise=True):
     Returns a generator of :class:`~iris.fileformats.grib.GribWrapper`
     fields from the given filename.
 
+    .. deprecated:: 1.10
+
```
pp-mo: cf : [comment: add links into deprecation notices](https://github.com/SciTools/iris/pull/1962#issuecomment-197895178)
Reference ```fileformats.message.GribMessage.messages_from_filename`` as an alternative.

lib/iris/fileformats/grib/__init__.py
```diff
@@ -962,7 +970,7 @@ def as_pairs(cube):
 def as_messages(cube):
     """
     Convert one or more cubes to GRIB messages.
-    Returns an iterable of grib_api GRIB messages.
+    Returns an iterable of grib_api GRIB message IDs.
```
pp-mo: I think 'handles' would be lot clearer than 'IDs', as Python objects have 'ids'.  But then I see that 'id' is used in the grib_api documentation.

This applies to all the last 3 low-level routines : as_pairs / as_messages / save_messages.
At least this one is explicit in referring to **_"grib_api_** GRIB message IDs"
I can't create a comment for it, but I think the line 950 docstring should be improved to a similar style.

pp-mo: Actually, there is some serious potential for confusion here, when we make 'message.GribMessage' public, about the fact that 'as_pairs' and 'as_messages' do **_not_** handle those, but grib_api handles instead.
I.E. despite losing 'GribWrapper', we still have a low-level and a higher-level message concept.

We could consider renaming 'GribMessage' to something that made it clearer that this is **_not_** the things that the grib_api deals in ?  I can now see that was the point of the name 'GribWrapper' !

lib/iris/fileformats/grib/message.py
```diff
@@ -60,7 +60,7 @@ def messages_from_filename(filename):
                     break
                 raw_message = _RawGribMessage(grib_id)
                 recreate_raw = _MessageLocation(filename, offset)
-                yield _GribMessage(raw_message, recreate_raw)
+                yield GribMessage(raw_message, recreate_raw)
 
     def __init__(self, raw_message, recreate_raw):
         """
```
pp-mo: The docs don't explain the meaning of 'recreate_raw' at all.
I think that needs fixing now we're making it public.

pp-mo: Also needed (but outside the scope of the current diffs):
The 'GribMessage.sections' property should have a docstring explaining that it is a **_mapping_**
 i.e.  " int ::--> :class:`Section` "
A quick usage example should be worth a lot of prose there, I think.

lib/iris/fileformats/grib/message.py
```diff
@@ -33,7 +33,7 @@
 from iris.exceptions import TranslationError
 
 
-class _GribMessage(object):
+class GribMessage(object):
     """
     Lightweight GRIB message wrapper, containing **only** the coded keys and
     data attribute of the input GRIB message.
```
pp-mo: This dosctring is basically a clone of the old 'GribWrapper' one, and wants revisiting.
I don't know if we still have an ambition to limit it to coded keys, but at present of course it does not so maybe best not even to say that.
Suggest something like.. """An object wrapping a grib_api message ID and exposing grib keys via the "sections" property."""

pp-mo:
After a longer think + further trawl through the code, it seems to me that the existing "GribMessage" class is not really ready yet to be exposed as a public general-purpose solution.  It seems at present to be designed for only specific, limited purposes : in particular, as required by the **_loading_** operation -- any possibility of using it for saving as well will require a good deal more work.
A number of points emerged ...
 * you can't modify keys in a Section, which limits the potential for customising the loading process (let alone saving)
 * it makes no good sense for a user to create a GribMessage via the normal \_\_init__ constructor --especially as, I think, we'd still much rather keep _RawGribMessage and  _MessageLocation private :  So, you can really only get valid ones from ```GribMessage.messages_from_filename```
   * in particular, you can't sensibly create a 'new from scratch' GribMessage, independent from a file, e.g. for use in saving.
 * the proposal in #1941 adds a ```load_pairs_from_fields``` function, to return GribMessage objects.  However, the existing **_save_** functionality, and the rules operation,  uses grib_api handles, and there is currently no easy route to rationalising the 'save' side :  we can't use GribMessage in its current form.
 * the caching of keys is currently weird + clumsy, given the code in ```RawGribMessage``` :
   * the code in ```Section``` looks like it intends to caches keys independently, but ...
   * GribMessage.sections --> RawGribMessage.sections --> RawGribMessage._get_message_sections()
   * ... so in fact it will read the whole lot at once, but only at the first attempt to access anything
 * although data access can be deferred till after the source file is closed, the key access cannot : It would fail if the original grib_api handle was discarded before reading any keys.

The issue about closing files is awkward because GribMessage.messages_from_filename() will close the file when the iteration is terminated.  So fine if you deal with each message in a generator, but not if you make a list of all to process them 'later'.

**_If_** we decide we still want to publish something at least barely usable in this area, 
then I suggest the best immediate resolution we can make is to force the loading of the sections, and document the resulting message objects as **_read-only_**.
 * it's possibly a bit of a disappointment not to be able to modify the messages within the load chain, as then you can't leverage the action of the rules on a modified message.  Nor can you fix a message that triggers a rules error.
 * however, you can still inspect messages and modify cubes at the 'load_pairs_...' level.
 * ...and even more simply, you can still just filter messages before generating cubes, which can help with slow loading or bad messages

pp-mo:
> the existing "GribMessage" class is not really ready yet

Regardless of the above...
**_If_** we decide we still want to publish something at least barely usable in this area, 
then I suggest the best immediate resolution we can make is to force the loading of the sections, and document the resulting message objects as **_read-only_**.
 * it's possibly a bit of a disappointment not to be able to modify the messages within the load chain, as then you can't leverage the action of the rules on a modified message.  Nor can you fix a message that triggers a rules error.
 * however, you can still inspect messages and modify cubes at the 'load_pairs_...' level.
 * ...and even more simply, you can still just filter messages before generating cubes, which can help with slow loading or bad messages

lib/iris/fileformats/grib/__init__.py
marqh: in my mind, a key deliverable for this change is to ensure that there is no need to use the grib_api directly

So, I would like to put together a proposal which gets all the functionality we need from GribMessage

There is a planned update to as_pairs and as_messages in #1941, which we can adapt to meet this aim

http://scitools.org.uk/iris/docs/v1.9.2/userguide/saving_iris_cubes.html#customising-the-save-process
would look much better if it did not include the gribapi calls, imo

pp-mo:
Correction:
> the caching of keys is currently weird + clumsy ...  in fact it will read the whole lot at once

Creating section does not fetch all key _values_, it only gets a list of key names.

pp-mo:
I think latest changes are an improvement, but I'm partial !

Probably now wants some tests to exercise ```GribMessage.__setitem__``` and (possibly) ```_cache_header```.

Also, I'm still wondering if naming "GribMessage" as something else might avoid any further confusion with gribapi message handles (similarly named in various existing documentation).
But I can't think of anything much better than "GribField", which is still pretty awful.

lib/iris/fileformats/grib/message.py
```diff
@@ -259,6 +274,16 @@ def __init__(self, message_id):
         self._message_id = message_id
         self._sections = None
 
+    def _cache_header(self):
+        """
+        Create sections and cache the header keys and values
+        as the file will be closed at some point.
+
+        """
+        for section in self.sections:
```
pp-mo: Needs to be ```for section in self.sections.values():```.
N.B. existing works in Python3 !

marqh:
I have pre-cached the header, such that it is usable outside of the iterator, to see the effect

some timings for integration.test_grib2:

with this change:
    2.530 (sigma 0.173)

master:
    0.835 (sigma 0.016)

@pp-mo  as you suspected, the _cache_sections is pretty expensive, a factor 3 increase in pain

hmmm....

pp-mo:
> a factor 3 increase in pain

Sounds big, but it's only significant in terms of some "total" operation time.
I'd say it's reasonable to measure iris.raw_load, at least, i.e. to include rules processing.
Are these numbers seconds, and for what operation over how many messages ?

lib/iris/fileformats/grib/message.py
pp-mo: As it is now, the thing called 'section' is an index.
But it's only used the context "self.sections[section]".
So I still think it's clearer to iterate over the actual section objects...
```
    for section in self.sections.values():
        for key in section.keys():
            ....
```

lib/iris/fileformats/grib/__init__.py
marqh: hmm, ok; this isn't going to be delivered by in this PR, we will ahve to wait until we are confident in message.GribMessage as a file writer, with appropriate grib save refactoring, before we can deliver this

lib/iris/fileformats/grib/message.py
```diff
@@ -53,27 +62,28 @@ def messages_from_filename(filename):
...
+        file_ref = _OpenFileRef(grib_fh)
+
+        while True:
+            offset = grib_fh.tell()
+            grib_id = gribapi.grib_new_from_file(grib_fh)
+            if grib_id is None:
+                break
+            raw_message = _RawGribMessage(grib_id)
+            recreate_raw = _MessageLocation(filename, offset)
+            yield GribMessage(raw_message, recreate_raw)
```
pp-mo: Important bit missing : should say "GribMessage(raw_message, recreate_raw**, file_ref=file_ref**)"

lib/iris/fileformats/grib/message.py
```diff
@@ -400,7 +391,16 @@ def __getitem__(self, key):
...
                     key, self._number))
             else:
-                value = self._get_key_value(key)
+                # _get_key_value can fail, if the gribapi has released
+                # the message.  In this case, the user can recreate the
+                # message an cache all the unread headers
+                try:
+                    value = self._get_key_value(key)
+                except gribapi.GribInternalError:
+                    msg = ('The underlying grib message is no longer'
```
pp-mo: Probably now obsolete ?

marqh:
ok, this is now looking like a coherent change.

The message module, the GribMessage and the Section classes are now public.  
The message maintains a link to it's file until references to it are removed, so it can be used outside of the from_filenames yield.
The cache on each section is the primary access point.  __setitem__ updates this cache and leaves the file on disk alone

I think this represents a good access pattern for now, able to support #1941 (I hope)

@pp-mo has been instrumental in this implementation.
@rhattersley @ajdawson we would greatly value your opinions on this implementation and the API additions at this stage

lib/iris/fileformats/grib/message.py
```diff
@@ -52,31 +66,46 @@ def messages_from_filename(filename):
...
-
-        * raw_message:
-            The _RawGribMessage instance which should be wrapped to
-            provide the `data` attribute.
+        It is recommended to obtain GribMessage instance from the static method
+        :meth:`GribMessage.messages_from_filename`, rather than creating
+        them directly.
 
         """
         self._raw_message = raw_message
```
pp-mo: Some (private) comments here clarifying the purpose of the constructor args would be good.
Maybe something like... :
```
    # A RawGribMessage giving gribapi access to the original grib message.
    self._raw_message = raw_message
    # A _MessageLocation which biggus uses to read the message data array,
    # by which time this message may be dead and the original grib file closed.
    self._recreate_raw = recreate_raw
    # An _OpenFileRef to keep the grib file open while this GribMessage is alive,
    # so that we can always use self._raw_message to fetch keys.
    self._file_ref = file_ref
```

I admit that seems a bit wordy though...

lib/iris/fileformats/grib/__init__.py
```diff
@@ -55,6 +55,7 @@
 
 #: Set this flag to True to enable support of negative forecast periods
 #: when loading and saving GRIB files.
+#: .. deprecated:: 1.10
```
pp-mo: An extra "#:" spacing line is wanted before the 'deprecated' line, to make this display nicely.

lib/iris/fileformats/grib/message.py
```diff
@@ -33,17 +33,31 @@
...
+
+    def __del__(self):
+        if not self.open_file.closed:
+            self.open_file.close()
+
+
+class GribMessage(object):
+    """
+    An in-memory representation of a GribMessage, providing
+    access to the :meth:`GribMessage.data` payload and the metadata
```
pp-mo: These display as "GribMessage.data()" and "GribMessage.sections()".
You can avoid the noise of repeating the module path, by using the form :meth:`~GribMessage.data`  (i.e. add a tilde)

lib/iris/fileformats/grib/message.py
```diff
@@ -33,17 +33,31 @@
...
+    elements by section via the :meth:`GribMessage.sections` property.
 
     """
 
     @staticmethod
     def messages_from_filename(filename):
         """
-        Return a generator of :class:`_GribMessage` instances; one for
+        Return a generator of :class:`GribMessage` instances; one for
```
pp-mo: "~" useful here too.

lib/iris/fileformats/grib/message.py
```diff
@@ -52,31 +66,53 @@ def messages_from_filename(filename):
...
+    def __init__(self, raw_message, recreate_raw, file_ref=None):
         """
-
-        Args:
-
-        * raw_message:
-            The _RawGribMessage instance which should be wrapped to
-            provide the `data` attribute.
+        It is recommended to obtain GribMessage instance from the static method
+        :meth:`GribMessage.messages_from_filename`, rather than creating
```
pp-mo: "~" useful here too.

lib/iris/fileformats/grib/message.py
```diff
@@ -52,31 +66,53 @@ def messages_from_filename(filename):
...
+        # alive, so that we can always use self._raw_message to fetch keys.
+        self._file_ref = file_ref
 
     @property
     def sections(self):
+        """
+        Return the key-value pairs of the message keys, grouped by containing
+        section.
+
+        Key-value pairs are collected into a dictionary of
```
pp-mo: Maybe doesn't read that clearly. 
"collected into" is describing from the implementation POV rather than usage.
Maybe something like this?..
```
Sections in a message are indexed by GRIB section-number, and values in a section are indexed by key strings.

.. For example::

    print(grib_message.sections[4]['parameterNumber'])
    grib_message.sections[1]['minute'] = 0
```

lib/iris/fileformats/grib/message.py
```diff
@@ -52,31 +66,53 @@ def messages_from_filename(filename):
...
+        :class:`Section` objects. One such object is made for
+        each section in the message, such that the section number is the
+        object's key in the containing dictionary. Each object contains
+        key-value pairs for all of the message keys in the given section.
+
+        For example::
+
+            print(grib_message[4]['parameterNumber'])
```
pp-mo: Should read ```grib_message```**```.sections```**```[4]```
See also above note..

lib/iris/fileformats/grib/message.py
```diff
@@ -124,6 +160,15 @@ def data(self):
             raise TranslationError(fmt.format(template))
         return data
 
+    def __getstate__(self):
```
pp-mo: I'm not convinced we need this, or for that matter any of "tests.integration.test_pickle.TestGribMessage".
As far as I can see, messages don't form any part of a cube :
only the ```fileformats.grib.message._DataProxy``` and ```._MessageLocation```do that
 * If this is ***not*** needed it should really go (and probably that pickle test too).
 * or, if it ***is*** needed then this doesn't necessarily solve the problem, as an unpickled message could not access any keys not already read in.

@rhattersley can you shed light on why we though we needed to **pickle** grib messages,
and could this maybe now be obsolete ??

pp-mo:
@marqh sorry for pickiness.
Really close now !

I would like to pursue the mysterious pickling requirement, though.

lib/iris/fileformats/grib/message.py
marqh: > If this is not needed it should really go (and probably that pickle test too).
> 
>     ( Else we are just passing on the burden of not knowing why this is needed ! )

May I suggest, given the lack of feedback, that this is an issue that I can take on and follow up in a subsequent PR as required, rather than blocking this issue.

> or, if it is needed then this doesn't necessarily solve the problem, as an unpickled message could not access any keys not already read in.

I disagee a little, as the unpickled current implementation has the same feature, if the keys are not cached, then they are inaccessible following an unpickle

So, may I suggest that I create and follow up on the 'why is this here, is it needed' but we don't hold off on this PR to wait for that.  This is a blocking PR already, blocking a blocking PR awaiting more input makes me feel :(

marqh:
s/pickiness/vital feedback

suggestions adopted for docs, much clearer thank you

lib/iris/fileformats/grib/message.py
pp-mo: Okay, on reflection I think you're right -- previous behaviour has not changed anyway, and this tweak just sidesteps the resulting exception.

pp-mo:
Thanks @marqh for sticking with it, sorry if I dragged my feet a bit

PR #1963, opened 2016-03-31
Regrid fix circular extend

pp-mo:
Addresses a reported problem whereby regridded data came out with large values instead of the expected masked points.
There was already a longstanding workaround for this, related to a known problem with numpy 'append', but it had a bug in its own logic.
I replaced the earlier fix with one using either np.concatenate or np.ma.concatenate, as required.  This seemed necessary to ensure the required result type, i.e. masked-for-masked-input and also unmasked-for-unmasked-input.

### DIFF
--- a/lib/iris/analysis/_interpolation.py
+++ b/lib/iris/analysis/_interpolation.py
@@ -1,4 +1,4 @@
-# (C) British Crown Copyright 2014 - 2015, Met Office
+# (C) British Crown Copyright 2014 - 2016, Met Office
 #
 # This file is part of Iris.
 #
@@ -101,27 +101,10 @@ def extend_circular_data(data, coord_dim):
     coord_slice_in_cube = [slice(None)] * data.ndim
     coord_slice_in_cube[coord_dim] = slice(0, 1)
 
-    # TODO: Restore this code after resolution of the following issue:
-    # https://github.com/numpy/numpy/issues/478
-    # data = np.append(cube.data,
-    #                  cube.data[tuple(coord_slice_in_cube)],
-    #                  axis=sample_dim)
-    # This is the alternative, temporary workaround.
-    # It doesn't use append on an nD mask.
-    if not (isinstance(data, ma.MaskedArray) and
-            not isinstance(data.mask, np.ndarray)) or \
-            len(data.mask.shape) == 0:
-        data = np.append(data,
-                         data[tuple(coord_slice_in_cube)],
-                         axis=coord_dim)
-    else:
-        new_data = np.append(data.data,
-                             data.data[tuple(coord_slice_in_cube)],
-                             axis=coord_dim)
-        new_mask = np.append(data.mask,
-                             data.mask[tuple(coord_slice_in_cube)],
-                             axis=coord_dim)
-        data = ma.array(new_data, mask=new_mask)
+    mod = ma if isinstance(data, ma.MaskedArray) else np
+    data = mod.concatenate((data,
+                            data[tuple(coord_slice_in_cube)]),
+                           axis=coord_dim)
     return data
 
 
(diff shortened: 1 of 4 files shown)


lib/iris/analysis/_interpolation.py
```diff
@@ -101,27 +101,10 @@ def extend_circular_data(data, coord_dim):
...
 
-    # TODO: Restore this code after resolution of the following issue:
-    # https://github.com/numpy/numpy/issues/478
-    # data = np.append(cube.data,
-    #                  cube.data[tuple(coord_slice_in_cube)],
-    #                  axis=sample_dim)
-    # This is the alternative, temporary workaround.
-    # It doesn't use append on an nD mask.
-    if not (isinstance(data, ma.MaskedArray) and
-            not isinstance(data.mask, np.ndarray)) or \
```
pp-mo: This was a bug I think : in "not isinstance(data.mask, np.ndarray)",  the "not" is not wanted !

lib/iris/analysis/_interpolation.py
```diff
@@ -101,27 +101,10 @@ def extend_circular_data(data, coord_dim):
...
-    else:
-        new_data = np.append(data.data,
-                             data.data[tuple(coord_slice_in_cube)],
-                             axis=coord_dim)
-        new_mask = np.append(data.mask,
-                             data.mask[tuple(coord_slice_in_cube)],
-                             axis=coord_dim)
-        data = ma.array(new_data, mask=new_mask)
+    mod = ma if isinstance(data, ma.MaskedArray) else np
+    data = mod.concatenate((data,
```
pp-mo: Replaced with concatenate as, though the original quoted bug is now fixed, 'append' still doesn't do what we want.

pp-mo:
**Note**
No whatsnew 
Will add that ...

lib/iris/tests/test_interpolation.py
```diff
@@ -510,25 +510,6 @@ def test_points_datatype_casting(self):
         # this test tries to extract a float from an array of type integer. the result should be of type float.
         r = iris.analysis.interpolate.linear(self.simple2d_cube_extended, [('shared_x_coord', 7.5)])
         self.assertCML(r, ('analysis', 'interpolation', 'linear', 'simple_casting_datatype.cml'))
-
```
pp-mo: Note: these do still work, but I think they are now obsolete

marqh: I understand the sentiment, I would just like to understand why these are now obsolete.  please could you expand a little on why these are not necessary and which remaining tests provide the test coverage?
thank you

marqh:
the change looks good to me; the implementation is simpler and easier to understand, as well as fixing the reported bug

the new tests look good to me as well

lib/iris/tests/test_interpolation.py
pp-mo: These tests exercise the exact codepaths that previously failed, as fixed in #155.
These cases now have new tests:
 * 'test_mask' exercises the case where -- before the original fix, and in earlier versions of numpy -- it would return an array whose mask was a different shape than the data. 
   * now replaced by ```test_circular_src_masked``` [here](https://github.com/SciTools/iris/pull/1963/files#diff-b79409258b6aa9758c83b4f54c8d09d0R1245)
 * 'test_scalar_mask' checks the case mask=False, which was bugged in earliest version of the fix
    * now replaced by ```test_circular_src__masked_missingmask``` [here](https://github.com/SciTools/iris/pull/1963/files#diff-b79409258b6aa9758c83b4f54c8d09d0R1233)

marqh:
great, I am happy now

@pp-mo please rebase/squish prior to a merge

thank you

PR #1964, opened 2016-04-06
Save rules for maximum and minimum values corrected

jonseddon:
This fixes a problem where LBPROC is not set when saving a minimum to a PP file. It also sets the IA component of LBTIM when saving a maximum or a minimum.

I have really struggled with the unit tests for these fixes. I have one for LBPROC, although this creates 84 warnings when run. I can't write a unit test for the LBTIM fix. When using the mock approach that other tests in the unit test module use, LBTIM is always set to an integer and so the IA, IB and IC components can't be set. I tried creating an additional mock for LBTIM that replaced it with an object with these components, but this was overwritten with an integer.

I also looked at the approach used in tests/integration/test_pp.py but got an error: "TypeError: 'NoneType' object has no attribute '__getitem__'" when I called "iris.fileformats.pp._save_rules.verify(cube, field)".

I have tested the fixes on real PP data and they work well with that.

### DIFF
--- a/lib/iris/etc/pp_save_rules.txt
+++ b/lib/iris/etc/pp_save_rules.txt
@@ -1,4 +1,4 @@
-# (C) British Crown Copyright 2010 - 2014, Met Office
+# (C) British Crown Copyright 2010 - 2016, Met Office
 #
 # This file is part of Iris.
 #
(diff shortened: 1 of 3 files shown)


lib/iris/etc/pp_save_rules.txt
```diff
@@ -150,6 +150,31 @@ IF
...
+    # interval
+    scalar_coord(cm, 'time') is not None
+    scalar_coord(cm, 'time').has_bounds()
+    scalar_coord(cm, 'clim_season') is None
+    scalar_coord(cm, 'forecast_period') is not None or scalar_coord(cm, 'forecast_reference_time') is not None
+    scalar_cell_method(cm, 'minimum', 'time') is not None
+    scalar_cell_method(cm, 'minimum', 'time').intervals != ()
+    scalar_cell_method(cm, 'minimum', 'time').intervals[0].endswith('hour')
+THEN
+    pp.lbtim.ia = int(scalar_cell_method(cm, 'minimum', 'time').intervals[0][:-5])
```
marqh: please could you unpack the logic of this line a little for me, with a comment in the code?

jonseddon: Sorry, I'm just getting to grips with Github, do you mean a comment on line 150, or comment on the new lines 153 to 164?

marqh:
Hello @jonseddon 

this looks good now, thank you

please may you 'squish' your commits into a single commit representing this whole change?
(this helps keep our commit history well managed)
for example
```
git rebase -i upstream/master
```

and 'fixup' the 'year in copyright' and 'logic unpacked' commits into the first one; then you have to 'force' the push (as you've changed the history)

```
git push -f origin pp_save_minimum
```

jonseddon:
Hi @marqh 

The commits should be rebased now.

Thanks,

Jon

QuLogic:
I'm not sure you remembered to change the rebase settings, or haven't yet pushed the changes, because the individual commit are still here.

jonseddon:
Sorry, hopefully I've done that correctly now.

rhattersley:
> Sorry, hopefully I've done that correctly now.

Yup. Looks good. Thanks  @jonseddon

PR #1966, opened 2016-04-06
scalar_cell_method now checks coord_name

jonseddon:
While doing some other work I noticed that scalar_cell_method() in lib/iris/fileformats/rules.py didn't check its coord_name argument, which I've now fixed.


marqh:
this looks great, thank you Jon

PR #1968, opened 2016-04-07
Fix the 'Nearest' scheme to return the same datatype as the input.

pp-mo:
Stumbled on this while refactoring regridding code.

I want Nearest() to behave sensibly so I can use it to replace the usage of the obsolete methods we are deprecating.  Refer : https://github.com/pp-mo/iris/pull/20


marqh:
this looks grand to me

I cannot see why nearest neighbour should coerce types

thanks @pp-mo

PR #1971, opened 2016-04-12
Added integration test which demonstrates a bug.

pp-mo:
One of these tests currently fails:
It should be possible to load CF data with a grid mapping variable that specifies no more than the type of mapping.
This currently works with a rotated-lat-lon, which is perhaps surprising as it has to use a default pole position.
However it fails for a "plain" (unrotated) lat-lon.

**This should be fixed**
(PR is a placeholder for a future fix)

### DIFF
--- a/lib/iris/tests/integration/test_nc_load_no_ellipsoid.py
+++ b/lib/iris/tests/integration/test_nc_load_no_ellipsoid.py
@@ -1,4 +1,4 @@
-# (C) British Crown Copyright 2014 - 2015, Met Office
+# (C) British Crown Copyright 2014 - 2016, Met Office
 #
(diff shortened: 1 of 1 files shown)


pp-mo:
This problem is derived from a Google Group report : https://groups.google.com/forum/?hl=en-GB&fromgroups=#!topic/scitools-iris/FDA_sGzFPbQ

The resulting error message is shown there.
That matches the error message produced by the failing test here.

corinnebosley:
@pp-mo This is very old now.  Would you mind rebasing to see whether this still needs fixing please?

corinnebosley:
It says it doesn't need rebasing but I'm pretty sure that's a fib.

PR #1972, opened 2016-04-12
Curl docs change

corinnebosley:
@pp-mo @rhattersley @marqh @pelson 
The documentation for this function was very vague and unclear, so I have added paragraphs which should make it more clear and understandable.  It now states the type of units that need to be on the cube data and what the different results will be when the k-component cube is and is not used.  It also states which coordinate systems will lead to the spherical calculus being used.

Please tell me if this documentation is clear and makes sense.  I would be very keen to hear feedback and suggestions.

### DIFF
--- a/lib/iris/analysis/calculus.py
+++ b/lib/iris/analysis/calculus.py
@@ -443,7 +443,12 @@ def _trig_method(coord, trig_function):
 
 def curl(i_cube, j_cube, k_cube=None, ignore=None):
     r'''
-    Calculate the 3d curl of the given vector of cubes.
+    Calculate the 2-dimensional or 3-dimensional spherical or cartesian
+    curl of the given vector of cubes.
+
+    As well as the standard x and y coordinates, this function requires each
+    cube to possess a vertical or z-like coordinate (representing some form
+    of height or pressure).  This can be a scalar or dimension coordinate.
 
     Args:
 
@@ -458,15 +463,31 @@ def curl(i_cube, j_cube, k_cube=None, ignore=None):
         The k cube of the vector to operate on
     * ignore
         This argument is not used.
+
         .. deprecated:: 0.8
             The coordinates to ignore are determined automatically.
 
     Return (i_cmpt_curl_cube, j_cmpt_curl_cube, k_cmpt_curl_cube)
 
+    If the k-cube is not passed in then the 2-dimensional curl will
+    be calculated, yielding a single cube representing the vertical component
+    as the result.  If the k-cube is passed in, the 3-dimensional curl will
+    be calculated, returning 3 component cubes.
+
+    All cubes passed in must have the same data units, and those units
+    must be spatially-derived (i.e. m/s or km/h).
+
     The calculation of curl is dependent on the type of
-    :func:`iris.coord_systems.CoordSystem` in the cube:
+    :func:`iris.coord_systems.CoordSystem` in the cube.
+    If the :func:`iris.coord_systems.CoordSystem` is either
+    GeogCS or RotatedGeogCS, the spherical curl will be calculated; otherwise
+    the cartesian curl will be calculated:
 
         Cartesian curl
+
+            When cartesian calculus is used, i_cube is the u vector component
+            (e.g. eastward), j_cube is the v component (e.g. northward) and
+            k_cube is the w component(vertical).
 
             The Cartesian curl is defined as:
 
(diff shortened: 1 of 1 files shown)


lib/iris/analysis/calculus.py
```diff
@@ -458,16 +463,32 @@ def curl(i_cube, j_cube, k_cube=None, ignore=None):
...
 
     Return (i_cmpt_curl_cube, j_cmpt_curl_cube, k_cmpt_curl_cube)
 
+    If the k-cube is not passed in then the 2-dimensional curl will
+    be calculated, yielding a single cube representing the vertical component
+    as the result.  If the k-cube is passed in, the 3-dimensional curl will
+    be calculated, returning 3 component cubes.
+
+    All cubes passed in must have the same data units, and those units
+    must be spatially-derived (i.e. m/s or km/h).
```
pp-mo: 'i.e.' should be 'e.g.' ?

lib/iris/analysis/calculus.py
```diff
@@ -458,16 +463,32 @@ def curl(i_cube, j_cube, k_cube=None, ignore=None):
...
     * ignore
         This argument is not used.
+
         .. deprecated:: 0.8
             The coordinates to ignore are determined automatically.
 
     Return (i_cmpt_curl_cube, j_cmpt_curl_cube, k_cmpt_curl_cube)
 
+    If the k-cube is not passed in then the 2-dimensional curl will
+    be calculated, yielding a single cube representing the vertical component
```
pp-mo: > "yielding a single cube"
I think it returns [None, None, k_cube], so this could be clearer.

lib/iris/analysis/calculus.py
```diff
@@ -458,16 +463,32 @@ def curl(i_cube, j_cube, k_cube=None, ignore=None):
...
     The calculation of curl is dependent on the type of
-    :func:`iris.coord_systems.CoordSystem` in the cube:
+    :func:`iris.coord_systems.CoordSystem` in the cube.
+    If the :func:`iris.coord_systems.CoordSystem` is either
+    GeogCS or RotatedGeogCS, the spherical curl will be calculated; otherwise
+    the cartesian curl will be calculated:
 
         Cartesian curl
 
+            When cartesian calculus is used, i_cube is the u vector component
```
pp-mo: This explanation is now identical to the 'spherical' case (!)
I think in this case we probably should **not** say eastward/northward, as that is no longer (always) correct.
So I think in the cartesian mode we should just call them "x/y/z- components", .

pp-mo:
Not your fault, nor even touched here, but I think the name "ellipsoidal" for the variable in line 570 is deeply confusing and wrong !
Given that "spherical" is already used to mean "really spherical, not ellipsoidal", you could use "cartesian", with the opposite sense ?

lib/iris/analysis/calculus.py
```diff
@@ -458,16 +463,32 @@ def curl(i_cube, j_cube, k_cube=None, ignore=None):
...
+    be calculated, yielding a single cube representing the vertical component
+    as the result.  If the k-cube is passed in, the 3-dimensional curl will
+    be calculated, returning 3 component cubes.
+
+    All cubes passed in must have the same data units, and those units
+    must be spatially-derived (i.e. m/s or km/h).
+
     The calculation of curl is dependent on the type of
-    :func:`iris.coord_systems.CoordSystem` in the cube:
+    :func:`iris.coord_systems.CoordSystem` in the cube.
```
pp-mo: handy hint : you can reduce the link text to the final name (leaving out module path) with a tilde (\~).
E.G. :func:\`~iris.coord_systems.CoordSystem\`

lib/iris/analysis/calculus.py
```diff
@@ -481,8 +502,8 @@ def curl(i_cube, j_cube, k_cube=None, ignore=None):
 
         Spherical curl
 
-            When spherical calculus is used, i_cube is the phi vector
-            component (e.g. eastward), j_cube is the theta component
+            When spherical calculus is used, i_cube is the :math:\phi vector
```
pp-mo: These :math: bits need backticks around.
E.G. :math:\`\phi`\

pp-mo:
Thanks @corinnebosley 
OWOK I think.
This is so worth doing !!!

lib/iris/analysis/calculus.py
corinnebosley: @pp-mo 

With velocities, lots of people use u, v and w.  I am concerned that it will be a little more confusing if we assign the cubes to x, y and z because each individual cube also needs an x, y and z.

lib/iris/analysis/calculus.py
corinnebosley: I did wonder about that...

corinnebosley:
Thank you @pp-mo , I have applied most of those changes.  I kept the cartesian description in terms of u, v and w but took out eastward and northward (and vertical).  I appreciate the input.

pp-mo:
Thanks @corinnebosley 
I like it !

corinnebosley:
Thank you @pp-mo 
That's one more thing I can be not getting on with.

PR #1973, opened 2016-04-13
Making Cubes From Scratch - Scitools gallery project

davidnmurray:
Here are my proposed changes in respect of the SciTools gallery exercise. My example consists of cube_from_scratch.py (the actual code; it makes an Iris cube from nothing) and the accompanying test (test_cube_from_scratch.py).

I also have a .png image that can accompany the exercise, showing an example output from cube_from_scratch. Do I need to upload this with the code, or is the idea that the image is generated during or after the merge process?


dkillick:
Welcome @davidnmurray!

Perhaps you could now alter your PR so that you push two different files? At the moment you have pushed the same file to two different locations.

Can you also please add a comment that describes the rationale for this work? We already have perfectly good example code (if admittedly not on here) that demonstrates creating a cube from scratch. I am also not convinced that this worthwhile thing to add to the gallery / example code, although any further thinking on this will have to wait until you've pushed the actual example code...

davidnmurray:
Hi Pete! Many thanks for your very prompt feedback. I've corrected the file upload - I must have accidentally clicked the same thing twice, or something like that. Apologies :)

As for rationales for purpose, my thoughts are a) as you point out, we don't have anything exactly like this on the gallery already (in fact I'm pretty sure there's nothing close?) and b) I think it would be good for new users to have something that can introduce them to making Iris cubes. This approach is quick, reasonably painless and doesn't have anything too abstruse going on inside it, so I feel it's a nice place to start. I hope that's some helpful clarification?

marqh:
Hi @davidnmurray 

 docs/iris/example_code/General/cube_from_scratch.py
needs to conform to 
https://www.python.org/dev/peps/pep-0008/

there's a neat tool available to help with this, at a terminal try:
```
pep8 
docs/iris/example_code/General/cube_from_scratch.py
```

this should give you a report on the lines which need altering to meet the pep8 coding standard

thank you
marqh

docs/iris/example_code/General/cube_from_scratch.py
```diff
@@ -0,0 +1,83 @@
...
+
+def main():
+    # Here we create our lists for latitude and longitude.
+    latitudes = [0, 10, 20, 30, 40]
+    longitudes = [5, 15, 25, 35, 45]
+
+    # Now we generate an array of random values. These values will be the basis for our cube's data.
+    data = (np.random.rand(3, 5, 5))
+
+    print data
```
marqh: In order to run in Python3 as well as Python2 this would have to be
```
print(data)
```

However, I would prefer not to have a print line in an example, so I suggest simply removing the line.

marqh:
> we don't have anything exactly like this on the gallery already (in fact I'm pretty sure there's nothing close?) and b) I think it would be good for new users to have something that can introduce them to making Iris cubes. This approach is quick, reasonably painless and doesn't have anything too abstruse going on inside it, so I feel it's a nice place to start.

I think this is a neat addition to the gallery.  Much of our documentation is built around our standard loaders and these do not always fit people's needs.

Building a new cube seems like a great addition to the example gallery to me.

davidnmurray:
Hi @marqh ! Many thanks for the useful feedback. I intend to put out a proper update soon...

davidnmurray:
Hi @marqh - I've uploaded a new cube_from_scratch version, corrected for Pep 8 compliance. I have two issues left over from that which I'm not quite sure how to handle. These two nagging bits are lines 44 and 45, which are the middle ones in the below chunk of code:

43    time_coord = iris.coords.DimCoord([349618., 349619., 349620.],
44                                      units=cf_units.Unit('hours since 1970-01-01 00:00:00',
45                                                          calendar='gregorian'),
46                                      standard_name='time')

They're both a few characters over the line length-limit, but I'm not 100% sure how to chop them down any further.

QuLogic:
After an opening parenthesis, place each parameter on a new line and indented by 4 spaces.

davidnmurray:
@QuLogic many thanks, I'll edit accordingly.

marqh:
Hello @davidnmurray 

the tests are still showing a set of PEP8 failures for
iris/docs/iris/example_code/General/cube_from_scratch.py

https://travis-ci.org/SciTools/iris/jobs/123869270

these need fixing before this can be adopted.

you can do this locally with:
```
pep8 docs/iris/example_code/General/cube_from_scratch.py
```

this should return no response for a python module which meets PEP8

mark

davidnmurray:
Hi @marqh ! Thanks again for your feedback.
I've run pep8 on my local script at the command line, and I have no errors. So, I'm going to upload that very shortly. Hopefully this should resolve the issue.

davidnmurray:
I see that the checks seem to have failed again, which is something of a surprise. The command line utility returned no problems. Do we know of anything that might account for the divergence?

ajdawson:
Your last commit doesn't contain any changes, perhaps you made a mistake updating this branch?

davidnmurray:
Thanks @ajdawson (and apologies for the delay in getting back to you). I'm not quite sure what happened with the last branch update, but it clearly wasn't what I intended :)

I'm going to upload a new edition of cube_from_scratch. This should be free of any pep8 problems - or at least, neither PyCharm nor the pep8 tool (as @marqh suggested) report any, and I can't see any from looking at the code. So, hopefully, the code should pass Travis this time.

ajdawson:
From your 4th commit you have been adding/editing a file called `cube_from_scratch.py` in the root of the iris directory. You need to make changes to the file `docs/iris/example_code/General/cube_from_scratch.py`.

I reckon you need to move the file in the root directory (which looks PEP8 compliant) into the place of `docs/iris/example_code/General/cube_from_scratch.py` where it belongs.

ajdawson:
>I am also not convinced that this worthwhile thing to add to the gallery / example code

I think I agree with this, might this be better in the user guide rather than the gallery, since the gallery image won't be identifiable as a cube created from scratch?

> We already have perfectly good example code (if admittedly not on here) that demonstrates creating a cube from scratch.

Where is the example you are referring to @dkillick?

davidnmurray:
Thanks again @ajdawson - I'd missed the fact that the files were committing to the root instead of the correct directory. I've corrected that now.

dkillick:
**tl;dr** this is useful code that isn't gallery material; it should be released on `iris_example_code` in the far more appropriate IPython notebook format.

> the gallery image won't be identifiable as a cube created from scratch

@ajdawson I almost wrote something like this in my original comment that you quoted from above. My logic went something like this: it feels like the gallery image that we'll get from this is almost being shoe-horned in to justify the existence of this piece of example code. I didn't write this because the example code exists separately to, if tightly coupled to, the gallery examples.

Nevertheless, the more changes are made to this PR, the more I become convinced that this is **not** gallery example material. I don't say this as a slight on @davidnmurray but as a statement that he has been asked to do something that is not appropriate, for want of a better word.

Also, as the data is being generated randomly, what we'll actually end up with is a random smattering of brightly-hued squares as a gallery example. And that isn't going to be at all helpful or look good on the gallery thumbnails page. It certainly isn't going to be clear what it's doing there without accessing the gallery example itself. I guess we could _mitigate_ this by setting the data to something generated trigonometrically for example, but that will only mitigate the randomly-hued squares; it won't make it any clearer what the example is demonstrating.

> Where is the example you are referring to?

Ah - unfortunately it (currently) isn't publically available; @mo-g wrote an IPython notebook that was only published here. However, it demonstrates **exactly** the same code as what @davidnmurray has written here, just in a far more appropriate format for telling the story that this example is trying to tell.

You only have to look at the latest iteration of the example code to see that this should be being provided as a notebook - each individual step is prefaced by a hefty multi-line comment. This is ideal territory for producing a notebook - then we can have a markdown cell that describes the steps taken in each successive code cell, which happens to be what @mo-g has already done...

Therefore I believe that rather than adding this to the gallery or the user guide we should release the existing notebook to the `iris_example_code` repo, which was introduced for precisely this type of example - something that's really useful to know but not really gallery example material.

ajdawson:
>Also, as the data is being generated randomly, what we'll actually end up with is a random smattering of brightly-hued squares as a gallery example.

Good point. It also makes it more annoying to test (but that could be solved).

I have more comments on details of this PR's code, but I'll save them until the direction (notebook or gallery or other) is determined.

marqh:
Hello @davidnmurray 

interesting thoughts from @dkillick @ajdawson 

I think that the user guide already covers many topics and that adding a new topic for this would not be the best option, the user guide is already a long and heavy read.  The `iris-example-code` is not the best place for this either, in my view: it has too low a profile within the user community.

I think that there is a sensible approach, which takes this essence of this PR and makes it more gallery like.  I would like to see code like this in the examples gallery in some form.
 
Having spoken to @dkillick I am interested in updating this example to produce an interesting picture.  If the coordinates are extended (https://github.com/davidnmurray/iris/blob/GalleryUpdate/docs/iris/example_code/General/cube_from_scratch.py#L22) to be larger arrays with a finer resolution and a coord_system added to each (https://github.com/davidnmurray/iris/blob/GalleryUpdate/docs/iris/example_code/General/cube_from_scratch.py#L33, https://github.com/davidnmurray/iris/blob/GalleryUpdate/docs/iris/example_code/General/cube_from_scratch.py#L38)

e.g.:
```
import numpy as np
import iris
lon_coord = iris.coords.DimCoord(np.linspace(-10, 5, 100),
                                      standard_name='longitude',
                                      units='degrees',
                                      coord_system=iris.coord_systems.GeogCS(6371229.0))

```
i suggest 40 to 50 for latitudes, as this will capture the UK
(the location is clearly not an important factor though)

this would open up the opportunity to calculate a data field based on the horizontal coordinate values and times and produce an interesting 2D pattern.  

All that would be needed is a arbitrary function which takes lat, lon and time values and calculates a 3D array from the function.
This could then be nicely plotted over a map

So, replacing https://github.com/davidnmurray/iris/blob/GalleryUpdate/docs/iris/example_code/General/cube_from_scratch.py#L27 with such a function, once all the coordinates are created

This would make an image more in keeping with the Iris gallery and avoid the use of random values, which make it harder to test

Any function would do, something which made a nice pattern would be better; suggestions are always welcome

mark

ajdawson:
Not a bad suggestion @marqh. Would we be willing to sacrifice something pretty for something meteorologically relevant and useful? If so may I suggest deriving the coriolis parameter `f = 2 * omega * sin(phi)` (where omega is Earth's rotation rate and phi is latitude) from a global latitude coordinate and plotting it on a global projection of your choice?

marqh:
> Not a bad suggestion @marqh. Would we be willing to sacrifice something pretty for something meteorologically relevant and useful? If so may I suggest deriving the coriolis parameter f = 2 * omega * sin(phi) (where omega is Earth's rotation rate and phi is latitude) from a global latitude coordinate and plotting it on a global projection of your choice?

that is a great suggestion, this would absolutely address the situation

a couple of clarifications please:
* what units are omega in?
* what units would the 'coriolis parameter' expect to be in?

many thanks @ajdawson

dkillick:
> > deriving the coriolis parameter

> this is a great suggestion

Absolutely seconded - this would be an ideal pattern for demonstrating creating a cube from scratch as the data has not come from an existing dataset but still has physical meaning within the sphere that we're working in.

ajdawson:
The units of omega are `s-1`, typical value is around 7.29x10-5 I think (but please check, might want to match UM value which would mimic a realistic requirement). The unit of the Coriolis itself is also `s-1` since the other terms are dimensionless.

marqh:
> I also have a .png image that can accompany the exercise, showing an example output from cube_from_scratch. Do I need to upload this with the code, or is the idea that the image is generated during or after the merge process?

you will need to upload the image of the expected plot as part of this PR.  The image will go here:
https://github.com/SciTools/iris/tree/master/lib/iris/tests/results/visual_tests

the filename is important, it must be:
<name of test module>.<name of test class>.<name of test function>.0.png
so that it is picked up for the correct test

davidnmurray:
Hi everyone; many thanks again for your very helpful feedback. A new iteration of the gallery example is currently being developed, and I hope to be able to share it soon.

@marqh many thanks for that clarification. I assume then that if a script was called (for instance) thing.py then the correct name for the associated image would be thing0.png?

davidnmurray:
Hi everyone! Following on from your suggestions, @ajdawson, @dkillick and @marqh, I've uploaded a new bit of code. This plots the Coriolis parameter across the UK, as suggested. (I've also removed the old cube_from_scratch, as it's not needed anymore.)

It currently has one outstanding issue, which is a colorbar with cluttered labels. This is under investigation right now, but I wanted to out the code out there so you could all see it. Of course, if anyone knows of a good way to control the number of labels on a color bar, I'm open to suggestions :)

docs/iris/example_code/General/coriolis_plot.py
```diff
@@ -0,0 +1,99 @@
+"""
+Coriolis parameters over the UK
```
ajdawson: Not a plural, just "parameter". However, "Coriolis frequency" might be more correct. Happy to have discussion on this.

docs/iris/example_code/General/coriolis_plot.py
```diff
@@ -0,0 +1,99 @@
...
+===============================
+
+This example shows the coriolis parameters for the area encompassing the UK.
```
ajdawson: Coriolis is a proper noun so should be capitalised. A more correct term might be the "Coriolis frequency".

docs/iris/example_code/General/coriolis_plot.py
```diff
@@ -0,0 +1,99 @@
...
+
+It creates and plots a map showing the coriolis parameters, overlaid on top of
```
ajdawson: Same deal with "coriolis" and "parameters".

docs/iris/example_code/General/coriolis_plot.py
```diff
@@ -0,0 +1,99 @@
...
+the UK coastline. The code also creates a custom cube to store the data in and
```
ajdawson: The first line should explain the code derives the Coriolis frequency and stores it in a cube, then go on to mention plotting.

docs/iris/example_code/General/coriolis_plot.py
```diff
@@ -0,0 +1,99 @@
...
+
+def calculate_coriolis_parameter(latitude):
+    """
+    This function computes the coriolis parameters for the points in an array.
+
+    It takes a 2D latitude array as its input.
+
+    It returns a 2D array for coriolis parameters.
+    """
+    omega = 7.29E-5
```
ajdawson: Lower case "e" would be nicer style in my opinion.

docs/iris/example_code/General/coriolis_plot.py
```diff
@@ -0,0 +1,99 @@
...
+    # The above process is repeated for the longitude coordinates.
+    long_coord = iris.coords.DimCoord(longitudes,
+                                      standard_name='longitude',
+                                      units='degrees',
+                                      coord_system=cs)
+
+    # The next step is to add vertical structure to the cube.
+    # This cube has only one level, but a cube with multiple levels
+    # can be made by feeding in a list rather than a single value.
+    height_coord = iris.coords.DimCoord(100.,
```
ajdawson: A height coordinate does not belong on this cube anymore. It should be latitude and longitude only.

docs/iris/example_code/General/coriolis_plot.py
```diff
@@ -0,0 +1,99 @@
...
+    new_cube = iris.cube.Cube(data,
+                              standard_name='coriolis_parameter',
+                              units='s-1',
+                              dim_coords_and_dims=[(long_coord, 0),
+                                                   (lat_coord, 1)],
+                              aux_coords_and_dims=[(height_coord, None)])
+
+    # This section checks to ensure that the coordinates are bounded.
+    # If they have no bounds, the code applies some via guess_bounds()
+    for coord in new_cube.coords():
```
ajdawson: Why not add bounds yourself? Since this is a cube from scratch, you have full control and don't need iris to guess the bounds, you can choose them.

docs/iris/example_code/General/coriolis_plot.py
```diff
@@ -0,0 +1,99 @@
...
+        if not isinstance(coord.bounds, np.ndarray):
+            coord.guess_bounds()
+
+    # Now let's see what our cube looks like:
+    ax = plt.axes(projection=ccrs.Mercator())
```
ajdawson: I think a global plot would be more appropriate.

docs/iris/example_code/General/coriolis_plot.py
```diff
@@ -0,0 +1,99 @@
...
+    ax.coastlines(resolution='10m')
+    qplt.pcolormesh(new_cube)
```
ajdawson: Can we have a diverging colormap too since the data (in a global sense) are centred on zero.

docs/iris/example_code/General/coriolis_plot.py
```diff
@@ -0,0 +1,99 @@
...
+    It returns a 2D array for coriolis parameters.
+    """
+    omega = 7.29E-5
+    return 2. * omega * np.sin(latitude)
+
+
+def main():
+    # Start with arrays for latitudes and longitudes, with a given resolution.
+    resolution = 50
+    latitudes = np.linspace(49.0, 60.0, resolution)
```
ajdawson: I think these coordinates should be expanded to cover the whole globe, this would make much more sense in the new context.

docs/iris/example_code/General/coriolis_plot.py
```diff
@@ -0,0 +1,99 @@
...
+    longitudes = np.linspace(-11.0, 3.0, resolution)
+    lat_calc = np.deg2rad(latitudes)
+    lon_calc = np.deg2rad(longitudes)
+    lat1, long1 = np.meshgrid(lat_calc, lon_calc)
+
+    # Now to call the coriolis function
+    data = calculate_coriolis_parameter(lat1)
```
ajdawson: It is a one liner, perhaps just keep it inline instead of using a function call.

docs/iris/example_code/General/coriolis_plot.py
```diff
@@ -0,0 +1,99 @@
...
+
+
+def main():
+    # Start with arrays for latitudes and longitudes, with a given resolution.
+    resolution = 50
+    latitudes = np.linspace(49.0, 60.0, resolution)
+    longitudes = np.linspace(-11.0, 3.0, resolution)
+    lat_calc = np.deg2rad(latitudes)
+    lon_calc = np.deg2rad(longitudes)
+    lat1, long1 = np.meshgrid(lat_calc, lon_calc)
```
ajdawson: The lat and lon are in the opposite order for normal conventions, I'd suggest switching them round then making the cube dimension order latitude longitude.

Also, the names don't make sense, you might consider `lat2d` instead of `lat1` since that nicely describes the content of the variable. Also I wouldn't bother assigning the longitudes to the variable `long1` since they aren't ever used. 

I'd also be tempted not to do the conversion to radians first and include that in the computation of the coriolis (which I think could be inline):

```python
_, lat2d = np.meshgrid(longitudes, latitudes)

omega = 7.29e-5
data = 2. * omega * np.sin(np.deg2rad(lat2d))
```

docs/iris/example_code/General/coriolis_plot.py
```diff
@@ -0,0 +1,99 @@
...
+It creates and plots a map showing the coriolis parameters, overlaid on top of
+the UK coastline. The code also creates a custom cube to store the data in and
+to plot it from.
+
+"""
+
+import cartopy.crs as ccrs
+
+import iris
+
```
ajdawson: Don't need spaces between the imports.

docs/iris/example_code/General/coriolis_plot.py
```diff
@@ -0,0 +1,99 @@
...
+                                        standard_name=None,
+                                        units='m',
+                                        long_name='level_height',
+                                        attributes={'positive': 'up'})
+
+    # Now we input our data array into the cube.
+    new_cube = iris.cube.Cube(data,
+                              standard_name='coriolis_parameter',
+                              units='s-1',
+                              dim_coords_and_dims=[(long_coord, 0),
```
ajdawson: As I've suggested elsewhere, I'd prefer the order [latitude, longitude]. If you make the suggested change to the `meshgrid` call this will be necessary.

ajdawson:
I've made a lot of inline comments, in summary:

* I think this example would be more useful if it computed the Coriolis parameter over the whole globe, and plotted it on a global projection (or orthographic or similar).

* The computation could be simplified somewhat.

* Terminology needs to be correct and consistent.

docs/iris/example_code/General/coriolis_plot.py
QuLogic: The blank line between Cartopy and Iris is (PEP8-)correct though. However, matplotlib and numpy should go in the same block as Cartopy.

ajdawson: I disagree. In the context of an example program iris should be considered as a third party library just like cartopy or matplotlib etc.

If you were writing a program yourself that used iris you would consider it a third party library so there is no reason we shouldn't do that here in our own examples (which are not part of the iris package).

QuLogic: Hrm, yes, good point. They can all squash together then, no need to change the order.

docs/iris/example_code/General/coriolis_plot.py
dkillick: @ajdawson the benefit of separating out the computation of the coriolis parameter was that it gave us a nice, small and single purpose function whose sole purpose was performing the computation. This meant we could separate the data computation out from the rest of the code, which may aid the code's legibility.

I agree that performing the `deg2rad` conversion inline is a good idea though.

docs/iris/example_code/General/coriolis_plot.py
dkillick: See above for our rationale for having this in a separate function.

docs/iris/example_code/General/coriolis_plot.py
dkillick: While it doesn't now make physical sense for this cube I think demonstrating how to add a scalar coordinate to a cube is still a beneficial thing to show in this example - perhaps we could change the scalar coordinate to something more realistic.

docs/iris/example_code/General/coriolis_plot.py
davidnmurray: The original logic was to show people an approach they could use to automatically check for bounds. Of course, that was several versions ago, so the guess_bounds() section may well be obsolete now. I'm fine with removing it if it no longer serves much purpose.

docs/iris/example_code/General/coriolis_plot.py
davidnmurray: Okay; I've edited the opening line of the doc string to read "This code computes the Coriolis parameter for an area and stores it in a cube."

docs/iris/example_code/General/coriolis_plot.py
ajdawson: Actually I'm not sure if bounds are appropriate or not here. Either way, I'm not sure the example should use `guess_bounds`.

docs/iris/example_code/General/coriolis_plot.py
davidnmurray: Fair enough; I'll make an appropriate amendment.

docs/iris/example_code/General/coriolis_plot.py
ajdawson: >gave us a nice, small and single purpose function

Containing one line of operation (so small, I agree!) and only called once. Is this good practice even? I'm just not convinced that a whole function is justified for a one-liner you only use once.

dkillick: > Is this good practice even?

This is a debate of separation of concerns and code clarity (how the code currently is) against best practice (your proposal) - and as the gallery should be demonstrating best-practice code I think that you are entirely right here and we should bring the calculation of the coriolis parameter into the main function. Thanks for pointing this out @ajdawson!

docs/iris/example_code/General/coriolis_plot.py
davidnmurray: My feeling is that the bounds section isn't adding any clarity, so perhaps it's for the best to remove it.

docs/iris/example_code/General/coriolis_plot.py
davidnmurray: @dkillick @ajdawson Okay, I think we have a consensus that the Coriolis parameter needs to come into the main function. I'll make the appropriate edit.

davidnmurray: With regards to the issue of long1, on balance I'm inclined to keep it. My thought is, a new user looking at this code might wonder what's going on with meshgrid, and where has the longitude 'gone'? I think there's a risk that without an explicit long1 output, a user could think that lat1 is a merged lat/long array, which is not really the case.

davidnmurray: (I should add I'm in agreement on everything else, though.)

ajdawson: I was just suggesting using the underscore to replace `long1`, but your point is reasonable. Perhaps we can have

    lon2d, lat2d = meshgrid(latitudes, longitudes)

which would hopefully make the usage of meshgrid very clear.

davidnmurray: @ajdawson agreed. I think lon2d, lat2d would work very well.

davidnmurray:
I've uploaded a new iteration incorporating everyone's feedback; many thanks again! Regarding the guess_bounds() issue, I've left that 'as is' for now. This is because the code essentially assumes an arbitrary resolution, and with explicit bounds on latitude and longitude, that would add more things that users would need to change.

docs/iris/example_code/General/coriolis_plot.py
```diff
@@ -0,0 +1,88 @@
...
+    # Start with arrays for latitudes and longitudes, with a given resolution.
+    resolution = 200
+    longitudes = np.linspace(-180.0, 180.0, resolution)
+    latitudes = np.linspace(-90.0, 90.0, resolution)
+    lon2d, lat2d = np.meshgrid(longitudes, latitudes)
+
+    # Omega is the Earth's rotation rate, expressed in radians per second
+    omega = 7.29e-5
+
+    # Now to create our data
```
ajdawson: Can we have a comment that states what the data are instead? Something along the lines of

```
Compute the Corilois parameter from the latitude values
```

docs/iris/example_code/General/coriolis_plot.py
```diff
@@ -0,0 +1,88 @@
...
+import iris
+from iris.coord_systems import GeogCS
+import iris.plot as iplt
+import matplotlib.pyplot as plt
+import numpy as np
+
+
+def main():
+    # Start with arrays for latitudes and longitudes, with a given resolution.
+    resolution = 200
```
ajdawson: Not really resolution, but coordinate size. Can we choose a better name for this?

docs/iris/example_code/General/coriolis_plot.py
```diff
@@ -0,0 +1,88 @@
...
+
+    # The Iris coords module turns the latitude list into a coordinate array.
+    # Coords then applies an appropriate standard name and unit to it.
+    lat_coord = iris.coords.DimCoord(latitudes,
+                                     standard_name='latitude',
+                                     units='degrees',
+                                     coord_system=cs)
+
+    # The above process is repeated for the longitude coordinates.
+    long_coord = iris.coords.DimCoord(longitudes,
```
ajdawson: I have a (perhaps unreasonably) strong personal preference for `lon_coord` over `long_coord`!

The following is a hand-wavy explanation for my feeling on this. I think this is for a few reasons, firstly it makes the variable name the same length as the corresponding variable for latitude, perhaps making it easier to see these two are of a kind... Secondly, `long_coord` makes me think about the length of the coordinate, not about it containing longitudes! These may be the ramblings of a mad man, but I think I genuinely would prefer `lon_coord` here!

docs/iris/example_code/General/coriolis_plot.py
```diff
@@ -0,0 +1,88 @@
...
+                                      standard_name='longitude',
+                                      units='degrees',
+                                      coord_system=cs)
+
+    # The next step is to add vertical structure to the cube.
+    # This cube has only one level, but a cube with multiple model levels
+    # can be made by feeding in a list rather than a single value.
+    height_coord = iris.coords.DimCoord(100.,
```
ajdawson: I still don't think this should be here. Why is the Coriolis parameter defined on a level 100 m above the surface? (i.e. it isn't, it is treated as height invariant).

I appreciate there is some appetite to include a demonstration of a scalar coordinate, but this feels like it is shoe-horned in for no reason other than we can.

docs/iris/example_code/General/coriolis_plot.py
davidnmurray: Okay; I'm thinking of substituting in something like "We will create our example data using the 2D latitude array", so that it's clear to the user that they need the result from meshgrid and not the raw one-d latitudes.

docs/iris/example_code/General/coriolis_plot.py
davidnmurray: Good point. I'll need to think a bit about specifically what - the obvious one is "coordinate_size", but I want to avoid any risk of confusion with "cs", which obviously is something else entirely.

docs/iris/example_code/General/coriolis_plot.py
davidnmurray: Okay, I'll make the change to lon_coord.

docs/iris/example_code/General/coriolis_plot.py
```diff
@@ -0,0 +1,88 @@
...
+    for coord in new_cube.coords():
+        if not isinstance(coord.bounds, np.ndarray):
+            coord.guess_bounds()
+
+    # Now let's plot our cube, along with coastlines, a title and an
+    # appropriately-labelled colour bar:
+    ax = plt.axes(projection=ccrs.Orthographic())
+    ax.coastlines(resolution='10m')
+    mesh = iplt.pcolormesh(new_cube, cmap='seismic')
+    bar = plt.colorbar(mesh, orientation='horizontal')
```
ajdawson: You could do all the colorbar work in one line, which might be more intuitive for users. I also suggest the addition of a label formatter to make the resulting plot a little tidier:

```python
tick_levels = [-0.00012, -0.00006, 0.0, 0.00006, 0.00012]
bar = plt.colorbar(mesh, orientation='horizontal', label='s-1',
                   ticks=tick_levels, format='%.1e')

File: docs/iris/example_code/General/coriolis_plot.py (hunk 18 above)
Comment (ajdawson) [inline docs/iris/example_code/General/coriolis_plot.py]: You'll be making a grumpy, set-in-his-ways old man very happy!

File: docs/iris/example_code/General/coriolis_plot.py (hunk 20 above)
Comment (davidnmurray) [inline docs/iris/example_code/General/coriolis_plot.py]: That makes sense. And I suppose labels with exponential notation is something science users would find useful, so I'll see to adopting that.

File: docs/iris/example_code/General/coriolis_plot.py (hunk 17 above)
Comment (ajdawson) [inline docs/iris/example_code/General/coriolis_plot.py]: It isn't a big deal really, don't invest much time here.

File: docs/iris/example_code/General/coriolis_plot.py [hunk 21]
```diff
@@ -0,0 +1,88 @@
+"""
+Coriolis parameters across the world
+====================================
+
+This code computes the Coriolis parameter for an area and stores it in a cube.
+
+It creates and plots a map showing the Coriolis parameters, overlaid on top of
```
ajdawson: parameters -> parameter

docs/iris/example_code/General/coriolis_plot.py
```diff
@@ -0,0 +1,88 @@
+"""
+Coriolis parameters across the world
+====================================
+
+This code computes the Coriolis parameter for an area and stores it in a cube.
```
ajdawson: Can we change this slightly:
```
This code computes the Coriolis parameter and stores it in a cube with associated metadata.
```

docs/iris/example_code/General/coriolis_plot.py
davidnmurray: Thank you, good catch! The inevitable typo finds its way in...

docs/iris/example_code/General/coriolis_plot.py
davidnmurray: I'm not absolutely-wedded to having a model level in there, and I agree that it probably made more sense in the older version than in this one. However, I think it would be useful to see what the consensus of opinion is on removing it/keeping it/substituting something else - @dkillick perhaps might have a view on this?

ajdawson: Well, think about it like this: we are demonstrating how to create metadata for a quantity we have derived ourselves. There is no logical reason why you would introduce this level coordinate, since the Coriolis parameter is independent of height. Adding the height coordinate teaches users to add spurious metadata to their cubes, which is not desirable. If you can't justify it in the context of the example, it shouldn't be there.

If you can think of another scalar coordinate that makes sense in the context of this example, please go ahead and add it, if not then I suggest you don't bother.

docs/iris/example_code/General/coriolis_plot.py
davidnmurray: Okay, I'll make that edit shortly.

docs/iris/example_code/General/coriolis_plot.py
davidnmurray: Fair enough. In the interests of keeping up some forward momentum, I'll push through a version without the model level section. (I suppose if it's subsequently decided it's needed, it can always be brought back later.)

docs/iris/example_code/General/coriolis_plot.py
davidnmurray: I've amended it to 'coordinate_points', as it basically it is a number of points of coordinates, I suppose.

docs/iris/example_code/General/coriolis_plot.py
davidnmurray: Always glad to be of service :)

davidnmurray:
Here is another iteration of coriolis_plot, incorporating the new feedback. I've also changed a detail of the colorbar code, as I realised that the statement "bar =" in "bar=plt.colorbar..." was unnecessary.

docs/iris/example_code/General/coriolis_plot.py
```diff
@@ -0,0 +1,77 @@
+"""
+Coriolis parameters across the world
```
ajdawson: Again, the use of the plural parameters is wrong. It should be parameter. This is clearly confusing, and I therefore assume it could be confusing to users reading this example, so please replace with:

```Deriving the Coriolis frequency over the globe```

This terminology has the possible advantage of being consistent with [Wikipedia](https://en.wikipedia.org/wiki/Coriolis_frequency).

File: docs/iris/example_code/General/coriolis_plot.py [hunk 24]
```diff
@@ -0,0 +1,77 @@
...
+====================================
+
+This code computes Coriolis parameters and stores them in a cube with
```
ajdawson: "Coriolis parameters" -> "Coriolis frequency"

docs/iris/example_code/General/coriolis_plot.py
```diff
@@ -0,0 +1,77 @@
...
+associated metadata.
+
+It creates and plots a map showing the parameters, overlaid on top of an
```
ajdawson: This reads oddly and uses incorrect terminology. Replace with:

```It plots the Coriolis frequency on an orthographic projection.```

I'd suggest that this line is part of the previous paragraph too.

File: docs/iris/example_code/General/coriolis_plot.py [hunk 26]
```diff
@@ -0,0 +1,77 @@
...
+
+    # Now let's plot our cube, along with coastlines, a title and an
+    # appropriately-labelled colour bar:
+    ax = plt.axes(projection=ccrs.Orthographic())
+    ax.coastlines(resolution='10m')
+    mesh = iplt.pcolormesh(new_cube, cmap='seismic')
+    tick_levels = [-0.00012, -0.00006, 0.0, 0.00006, 0.00012]
+    plt.colorbar(mesh, orientation='horizontal', label='s-1',
+                 ticks=tick_levels, format='%.1e')
+    plt.title('Coriolis parameter')
```
ajdawson: "Corilois parameter" -> "Coriolis frequency"

docs/iris/example_code/General/coriolis_plot.py
```diff
@@ -0,0 +1,77 @@
...
+    # Start with arrays for latitudes and longitudes, with a given resolution.
+    coordinate_points = 200
+    longitudes = np.linspace(-180.0, 180.0, coordinate_points)
+    latitudes = np.linspace(-90.0, 90.0, coordinate_points)
+    lon2d, lat2d = np.meshgrid(longitudes, latitudes)
+
+    # Omega is the Earth's rotation rate, expressed in radians per second
+    omega = 7.29e-5
+
+    # We will create our example data using the 2D latitude array
```
ajdawson: Might want to include a mention of what `data` actually is. Perhaps:

```
The data for our cube is the Coriolis frequency `f = 2 * omega * sin(phi)`, which is
computed for each grid point over the globe from the 2-dimensional latitude array.
```

davidnmurray:
@ajdawson many thanks again for your thorough comments. I've made all the appropriate text edits, and shall push through an updated script very shortly.

davidnmurray:
I see the Travis build has failed; this might be because it slipped my mind to upload a new version of the test image. (The plot title, of course, is different.) I'll try that and see if that helps.

docs/iris/example_code/General/coriolis_plot.py
```diff
@@ -0,0 +1,78 @@
...
+    # Now we input our data array into the cube.
+    new_cube = iris.cube.Cube(data,
+                              standard_name='coriolis_parameter',
+                              units='s-1',
+                              dim_coords_and_dims=[(lat_coord, 0),
+                                                   (lon_coord, 1)])
+
+    # This section checks to ensure that the coordinates are bounded.
+    # If they have no bounds, the code applies some via guess_bounds()
+    for coord in new_cube.coords():
```
ajdawson: This is the only problematic part left. I'm not totally opposed to calling `guess_bounds()` as long as it is explained properly. You should also avoid the loop and the isinstance check. Again, you are creating from scratch so you don't need to inspect the coordinates, you made them so you know if they have bounds or not! Perhaps this could be moved to before the cube is created and be restructured as in:

```python
# Add bounds to our latitude and longitude coordinates. We want simple
# contiguous bounds for our regularly spaced coordinate points so we just
# use the guess_bounds() method of the coordinate which does exactly this.
# If our coordinates were more complex we could derive and set the bounds
# manually.
lat_coord.guess_bounds()
lon_coord.guess_bounds()
```

This leaves me with only 1 minor concern, you latitude coordinate includes the poles and therefore the generated bounds may lie outside the valid [-90,90] range. I think @rhattersley fixed this in #1978 though... so maybe this is fine now.

ajdawson:
Test failures are unrelated. Only one issue left to resolve.

docs/iris/example_code/General/coriolis_plot.py
davidnmurray: @ajdawson okay, thanks for that. I think this approach is a good way to proceed. I've just given this code a local run, and I didn't see any errors or problems, so it looks like the poles aren't introducing any problems.

That said, what I'll also do is restrict the latitude range to (say) -89.5 to 89.5, to avoid the poles entirely. I'd been worried it might create some empty space on the plot, but having run the code with those terms, I'm not seeing any problems.

ajdawson: OK, well #1978 has been merged anyway so there should be no issue with the bounds regardless of your choice of points. Feel free to proceed as you see fit with regards to choosing the points.

davidnmurray: In that case, what I might do is just put the latitudes as -90.0 to 90.0, so it looks tidier and I can have one less line of comments than I do right now. It seems to work, so that's the main thing :)

davidnmurray:
Here is another update to the coriolis_plot code, incorporating this morning's discussions.

davidnmurray:
It looks like the Travis failures were due to a possible pep8 issue; one of the comment lines was nudging up against the 80 character limit. I've reworded line 56 to attempt to remedy this problem.

docs/iris/example_code/General/coriolis_plot.py
```diff
@@ -0,0 +1,80 @@
+"""
+Deriving the Coriolis frequency over the globe
+==============================================
+
+This code computes Coriolis frequency and stores them in a cube with
```
ajdawson: This code computes Coriolis frequency and stores them in a cube with
This code computes **the** Coriolis frequency and stores ~~them~~ **it** in a cube with

docs/iris/example_code/General/coriolis_plot.py
```diff
@@ -0,0 +1,80 @@
...
+associated metadata.
+
+It plots the Coriolis frequency on an orthographic projection.
```
ajdawson: It **then** plots ...

Make this line part of the first paragraph.

ajdawson:
Just a few more grammatical changes that got missed.

davidnmurray:
Here's a quick set of changes to reflect the grammatical changes. I note there was still one failing Travis check on the last commit, although it wasn't very clear to me what the root problem actually was.

ajdawson:
Failing tests look like timeouts on unrelated units, this is a known issue and should not be a concern here.

ajdawson:
OK I'm happy with this now. Can you please squash it into a single commit @davidnmurray.

davidnmurray:
@ajdawson excellent, glad to hear it :) I'll get this sorted out in a little while.

davidnmurray:
The commits have now been squashed (and apologies for the delay!).

ajdawson:
Thanks, can I just confirm that you are employed by the Met Office @davidnmurray and therefore do not need to sign a CLA?

davidnmurray:
@ajdawson  Yes, that's correct. I started here in late December, so it's all still quite recent!

ajdawson:
Thanks, I have to check for new contributors that aren't listed [here](http://scitools.org.uk/about.html).

ajdawson:
5 failed tests are not related to this change and are known to cause timeout errors, therefore I am satisfied these can be ignored.

ajdawson:
Good work @davidnmurray :tada: Turned out to be a nice example.

dkillick:
Hurrah! Well done on persevering with this @davidnmurray :+1:

davidnmurray:
Many thanks @ajdawson and @dkillick - I'm very glad we found a way through :)

PR #1974, opened 2016-04-14
Additions to cartography.project documentation

corinnebosley:
@pp-mo @marqh 
I have added a note about the target projection not being assigned to the grid if it's not an iris coordinate system, and a usage example for each type of target CS (one iris and one cartopy).

Let me know if you think I could add more.

### DIFF
--- a/lib/iris/analysis/cartography.py
+++ b/lib/iris/analysis/cartography.py
@@ -1,4 +1,4 @@
-# (C) British Crown Copyright 2010 - 2015, Met Office
+# (C) British Crown Copyright 2010 - 2016, Met Office
 #
 # This file is part of Iris.
 #
--- a/lib/iris/analysis/cartography.py
+++ b/lib/iris/analysis/cartography.py
@@ -480,6 +480,7 @@ def project(cube, target_proj, nx=None, ny=None):
     Return a new cube that is the result of projecting a cube with 1 or 2
     dimensional latitude-longitude coordinates from its coordinate system into
     a specified projection e.g. Robinson or Polar Stereographic.
+
     This function is intended to be used in cases where the cube's coordinates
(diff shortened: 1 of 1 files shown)


lib/iris/analysis/cartography.py
```diff
@@ -500,8 +501,20 @@ def project(cube, target_proj, nx=None, ny=None):
             covering the globe.
 
     Returns:
-        An instance of :class:`iris.cube.Cube` and a list describing the
-        extent of the projection.
+        An tuple containing an instance of :class:`iris.cube.Cube` and a list
```
djkirkham: **A** tuple

lib/iris/analysis/cartography.py
```diff
@@ -500,8 +501,20 @@ def project(cube, target_proj, nx=None, ny=None):
...
+        describing the extent of the projection.
```
rhattersley: It'd be helpful to describe the structure of the extent, i.e. `[x_min, x_max, y_min, y_max]`.

lib/iris/analysis/cartography.py
```diff
@@ -500,8 +501,20 @@ def project(cube, target_proj, nx=None, ny=None):
...
+
+    Example::
+
+        from iris.analysis.cartography import project
+
+        iris_cs = iris.coord_systems.Stereographic()
+        iris_cs_cube = project(cube, iris_cs)
+
+        cartopy_cs = ccrs.PlateCarree()
+        cartopy_cs_cube = project(cube, cartopy_cs)
```
rhattersley: In both these examples I think it'd be clearer to unpack the return value, e.g.:
```python
reprojected_cube, extent = project(cube, cartopy_cs)
```

PR #1975, opened 2016-04-20
Added zonal (i.e. longitudinal) mean support to LBPROC save rule

hdyson:
This is a fix to correctly set the LBPROC upon saving as a pp file for zonal mean data.  I have not included a test - if you can point me in the direction of where the tests should be for a pp_save_rules fix, I'd be grateful.  

For a bit of context, the closest test I could find was [here](https://github.com/SciTools/iris/blob/master/lib/iris/tests/test_cube_to_pp.py#L265)  However, that is acting on zonal mean being set in a particular cube attribute rather than the cube cell methods.  Simplifying that test down to just time and longitudinal mean yields this:

> cube = stock.lat_lon_cube()
cube.attributes["ukmo__process_flags"] = ['Time mean field', 'Zonal mean field']
temp_filename = iris.util.create_temp_filename(".pp")
iris.save(cube, temp_filename)
result = iris.fileformats.pp.load(temp_filename).next()
result.lbproc

Which correctly reports 192.  However, this snippet using cell methods instead reports 128 (i.e. just time mean):

> cube = stock.lat_lon_cube()
cube.cell_methods = (
    iris.coords.CellMethod(method=u'mean', coords=u'time'),
    iris.coords.CellMethod(method=u'mean', coords=u'longitude')
)
temp_filename = iris.util.create_temp_filename(".pp")
iris.save(cube, temp_filename)
result = iris.fileformats.pp.load(temp_filename).next()
result.lbproc

This PR corrects that.

### DIFF
--- a/lib/iris/tests/integration/test_pp.py
+++ b/lib/iris/tests/integration/test_pp.py
@@ -25,8 +25,9 @@ import iris.tests as tests
 
 import numpy as np
 
+from cf_units import Unit
 from iris.aux_factory import HybridHeightFactory, HybridPressureFactory
-from iris.coords import AuxCoord, CellMethod
+from iris.coords import AuxCoord, CellMethod, DimCoord
 from iris.cube import Cube
 import iris.fileformats.pp
 import iris.fileformats.pp_rules
@@ -585,6 +586,54 @@ class TestAsCubes(tests.IrisTest):
                 cubes.append(cube)
         for cube in cubes:
             self.assertEqual(cube.attributes['pseudo level'], 3)
+
+
+class TestSaveLBPROC(tests.IrisTest):
+
+    def create_cube(self, longitude_coord='longitude'):
+        cube = Cube(np.zeros((2, 3, 4)))
+        tunit = Unit('days since epoch', calendar='gregorian')
+        tcoord = DimCoord(np.arange(2), standard_name='time', units=tunit)
+        xcoord = DimCoord(np.arange(3), standard_name=longitude_coord,
+                          units='degrees')
+        ycoord = DimCoord(points=np.arange(4))
+        cube.add_dim_coord(tcoord, 0)
+        cube.add_dim_coord(xcoord, 1)
+        cube.add_dim_coord(ycoord, 2)
+        return cube
+
+    def convert_cube_to_field(self, cube):
+        field = iris.fileformats.pp.PPField3()
+        field.lbvc = 0
+        iris.fileformats.pp._ensure_save_rules_loaded()
+        iris.fileformats.pp._save_rules.verify(cube, field)
+        return field
+
+    def test_time_mean_only(self):
+        cube = self.create_cube()
+        cube.add_cell_method(CellMethod(method='mean', coords='time'))
+        field = self.convert_cube_to_field(cube)
+        self.assertEqual(int(field.lbproc), 128)
+
+    def test_longitudinal_mean_only(self):
+        cube = self.create_cube()
+        cube.add_cell_method(CellMethod(method=u'mean', coords=u'longitude'))
+        field = self.convert_cube_to_field(cube)
+        self.assertEqual(int(field.lbproc), 64)
+
+    def test_grid_longitudinal_mean_only(self):
+        cube = self.create_cube(longitude_coord='grid_longitude')
+        cube.add_cell_method(CellMethod(method=u'mean',
+                                        coords=u'grid_longitude'))
+        field = self.convert_cube_to_field(cube)
+        self.assertEqual(int(field.lbproc), 64)
+
+    def test_time_mean_and_zonal_mean(self):
+        cube = self.create_cube()
+        cube.add_cell_method(CellMethod(method=u'mean', coords=u'time'))
+        cube.add_cell_method(CellMethod(method=u'mean', coords=u'longitude'))
+        field = self.convert_cube_to_field(cube)
+        self.assertEqual(int(field.lbproc), 192)
 
 
 if __name__ == "__main__":
--- a/lib/iris/etc/pp_save_rules.txt
+++ b/lib/iris/etc/pp_save_rules.txt
@@ -544,6 +544,19 @@ IF
 THEN
     pp.lbproc += sum([iris.fileformats.pp.lbproc_map[name] for name in cm.attributes["ukmo__process_flags"]])
 
+#zonal-mean
+IF
+    # Look for a CellMethod which is a "mean" over "longitude".
+    scalar_cell_method(cm, 'mean', 'longitude') is not None
+THEN
+    pp.lbproc += 64
+
+IF
+    # Look for a CellMethod which is a "mean" over "grid longitude".
+    scalar_cell_method(cm, 'mean', 'grid_longitude') is not None
+THEN
+    pp.lbproc += 64
+
 #time-mean
 IF
     # Look for a CellMethod which is a "mean" over "time".
@@ -564,7 +577,6 @@ IF
     scalar_cell_method(cm, 'maximum', 'time') is not None
 THEN
     pp.lbproc += 8192
-
 
 ##########################
 ### vertical - lbuser5 ###


cpelley:
I would be inclined to create a new test class in [iris/tests/integration/test_pp.py](https://github.com/SciTools/iris/blob/e5f883dc21ac5159d324ce931bc060064da3fc94/lib/iris/tests/integration/test_pp.py) called `TestSaveLBPROC`.

I would also be inclined to write a new function in [`iris/fileformats/rules.py`](https://github.com/SciTools/iris/blob/e5f883dc21ac5159d324ce931bc060064da3fc94/lib/iris/fileformats/rules.py) called `is_zonal_mean` perhaps.

[`pp_save_rules.txt`](https://github.com/SciTools/iris/blob/e5f883dc21ac5159d324ce931bc060064da3fc94/lib/iris/etc/pp_save_rules.txt) change then perhaps become something like:

```
IF
    is_zonal_mean(cm)
THEN
    pp.lbproc += 128
```

This function can then handle latitude, longitude, grid_longitude and grid_latitude etc.  Not handling these cases might be seen as deceptive behaviour to the end user.

**Can an iris developer confirm what actions will be required to get this through a review?**

Cheers

marqh:
the handling of the pp_save_rules is planned to change in teh near future, so I would be inclined to keep the code in the save rules, and not add a helper function into the rules module

@pp-mo was working in this space, and may care to comment

i agree with @cpelley that a new test class in iris/tests/integration/test_pp.py would be sensible; we would certainly appreciate a test for the new behaviour

hdyson:
I've added tests that cater to the specific cases we're interested in (time mean, zonal mean, time & zonal mean).  There's two failing tests on an area of code that I don't think these changes touch - TimedOutExceptions on AreaWeightedRegrid tests i.e. similar issues to #1986.  Any ideas?

ajdawson:
>There's two failing tests on an area of code that I don't think these changes touch - TimedOutExceptions on AreaWeightedRegrid tests i.e. similar issues to #1986. Any ideas?

Those can be safely ignored in this case. These tests fail intermittently due to timeouts, I believe @cpelley is on the case already so this shouldn't be a problem for too much longer.

pp-mo:
> Those can be safely ignored in this case. 

To clarify, if you haven't seen this before: 
 * click on the red 'X' next to your commit in github, and then to sub-tests as required.
 * re-run the ones which are timing out, by hitting the 'Restart Build' control (clockwise round-and-again arrow)
 * wait to see if it happens again
 * repeat until it obeys !

ajdawson:
Don't you need to be a project dev to restart builds?

pp-mo:
>  I would be inclined to keep the code in the save rules, and not add a helper function into the rules module

I'm agreed with that in this case.
I doubt that a specialist "is_zonal_mean" function makes the logic any much clearer, as the original is a one-line test anyway, and we're only likely to use it twice?  As proposed, at least we see exactly what it is testing.
**On the other hand, I do agree that both longitude and grid_longitude should be handled.**

As for latitude/grid_latitude, I've not heard of that -- would you call it a "meridional mean" ? -- I don't think there is an LBPROC bit encoding for such a thing.

pp-mo:
> Don't you need to be a project dev to restart builds?

Sorry, I didn't know that.
I have re-spun it myself...

hdyson:
> On the other hand, I do agree that both longitude and grid_longitude should be handled.

I've added grid_longitude to the pp_save_rules - I have not added a separate test for this, it seems to me as though it's a bit redundant with the existing tests.  If you'd prefer an additional test, I'll certainly add one.

> Sorry, I didn't know that.
> I have re-spun it myself...

If you could re-do your magic on the timing out tests, I'd be grateful.

hdyson:
I hadn't noticed that there were conflicts again on this one - is the procedure that I keep fixing this each time there's a conflict, or do I wait and resolve the conflicts after it's reviewed?

Given how close the conflicting change is to this one, it's a shame the two fixes weren't combined.

cpelley:
> I hadn't noticed that there were conflicts again on this one - is the procedure that I keep fixing this each time there's a conflict, or do I wait and resolve the conflicts after it's reviewed?

Ordinarily rebasing first is preferable otherwise the reviewer is not reviewing the changes that are going onto trunk, however, before going to the effort of rebasing again.  I'm happy to review this myself but were still going to need someone with the green button to commit once I give it the thumbs up.

Any takers?

hdyson:
If someone could wave a magic wand at the timing out test, I'd be grateful.

bjlittle:
@hdyson Done :wink:

lib/iris/tests/integration/test_pp.py
```diff
@@ -587,5 +588,56 @@ def test_as_pairs(self):
...
+class TestSaveLBPROC(tests.IrisTest):
+
+    def create_cube(self):
+        cube = Cube(np.zeros((2, 3, 4)))
+        cube.add_dim_coord(
+            DimCoord(standard_name='time',
+                     units=Unit('days since 0001-01-01 00:00:00',
+                                calendar='gregorian'),
+                     points=np.arange(2),),
+            0)
```
cpelley: Formatting on this makes it difficult to read, instead:

```Python
    def create_cube(self):
        cube = Cube(np.zeros((2, 3, 4)))
    	tunit = Unit('days since 0001-01-01 00:00:00', calendar='gregorian')
	    tcoord = DimCoord(np.arange(2), standard_name='time', units=tunit)
        cube.add_dim_coord(tcoord, 0)
        ...
```

lib/iris/tests/integration/test_pp.py
```diff
@@ -587,5 +588,56 @@ def test_as_pairs(self):
...
+        cube.add_dim_coord(DimCoord(standard_name='longitude',
+                                    units='degrees',
+                                    points=np.arange(3),), 1)
```
cpelley: Not as bad, but while your at it, why not similarly change as above.

lib/iris/tests/integration/test_pp.py
```diff
@@ -587,5 +588,56 @@ def test_as_pairs(self):
...
+        cube.add_dim_coord(DimCoord(standard_name='latitude',
+                                    units='degrees',
+                                    points=np.arange(4),), 2)
```
cpelley: And here.  +make the the first argument points, I wouldn't bother specifying the name.

i.e. `DimCoord(np.arange(4), ...)`

lib/iris/tests/integration/test_pp.py
hdyson: For brevity (and to keep it to one line within PEP8 constraints), I've switched to var_name rather than standard_name.

ajdawson: `var_name` is very different to `standard_name`. `var_name` is for the name you'd like a variable/coordinate to be called when saved to netCDF, and I think is a special case for netCDF only.

hdyson: @ajdawson I recognise that var_name and standard_name are different: but for the purposes of these tests, they're functionally the same.  If there's no name at all on the coordinate, the relevant LBPROC isn't set.  However, var_name is sufficient to trigger setting LBPROC, which is all we're interested in here (the test failure below is another timed out regrid).

Now, if you're concerned that future changes around this area of the code (as I understand it, the pp_save_rules are being migrated to python?) will break the usage of var_name here, that's fine: let me know, and I'll revert to standard_name.

ajdawson: I just don't think there is a need to change it, I think @cpelley just meant to change from using points as a keyword to using it as a positional argument, otherwise this is fine.

hdyson: @ajdawson From my perspective, the choice is arbitrary, so the decision to try var_name was purely aesthetic.  Given it was added as part of this fix - rather than modifying an existing test - I'm not sure I see the downside, but that's a bit of a moot point: I'll change it back momentarily.

ajdawson: I just feel on principal that your choice of implementation (i.e. choosing `var_name` or `standard_name` which are functionally different in the wider context) should not be dictated by the coding standards we choose to enforce. You chose `standard_name` initially, presumably because that is the implementation you really wanted, and you shouldn't have to change it to conform to how the code should look. Coding standards are important, but there are better ways of improving to conform.

hdyson: No problem, and thanks for clarifying.  The reason for choosing standard_name was almost certainly the result of a copy and paste from an existing test, for what it's worth, so I wouldn't read too much into it.  That would explain why the redundant latitude name was still present though.

hdyson:
There's another timing out test, so if someone could retry that test, that'd be great.

lib/iris/tests/integration/test_pp.py
```diff
@@ -587,5 +588,52 @@ def test_as_pairs(self):
...
+        field.lbvc = 0
+        iris.fileformats.pp._ensure_save_rules_loaded()
+        iris.fileformats.pp._save_rules.verify(cube, field)
+        return field
+
+    def test_time_mean_only(self):
+        cube = self.create_cube()
+        cube.cell_methods = (
+            iris.coords.CellMethod(method='mean', coords='time'),
+        )
```
bjlittle: @hdyson 
A minor point, but it's probably safer (from a maintenance point of view) to use the `add_cell_method` method ...
```python
cm = CellMethod(method='mean', coords='time')
cube.add_cell_method(cm)
```

Note that, `CellMethod` is imported directly `from iris.coords` ... so there's no need to specify the full namespace i.e. `iris.coords.CellMethod`

If you agree, then rinse-and-repeat for all below ... ?

lib/iris/tests/integration/test_pp.py
```diff
@@ -587,5 +588,52 @@ def test_as_pairs(self):
...
+
+    def test_time_mean_and_zonal_mean(self):
+        cube = self.create_cube()
+        cube.cell_methods = (
+            iris.coords.CellMethod(method=u'mean', coords=u'time'),
+            iris.coords.CellMethod(method=u'mean', coords=u'longitude')
+        )
+        field = self.convert_cube_to_field(cube)
+        self.assertEqual(int(field.lbproc), 192)
+
```
bjlittle: @hdyson Just for completeness, how about a test case to cover `grid_longitude' ... that would be dandy!

bjlittle:
@hdyson I'm :+1: for this PR, once you serviced some minor comments.

lib/iris/tests/integration/test_pp.py
hdyson: Thanks @bjlittle - will do.

lib/iris/tests/integration/test_pp.py
```diff
@@ -587,5 +588,52 @@ def test_as_pairs(self):
             self.assertEqual(cube.attributes['pseudo level'], 3)
 
 
+class TestSaveLBPROC(tests.IrisTest):
+
+    def create_cube(self):
+        cube = Cube(np.zeros((2, 3, 4)))
+        tunit = Unit('days since 0001-01-01 00:00:00', calendar='gregorian')
```
bjlittle: @hdyson 
Just noticed this line (after a double-take) ... did you know that you can also do this `Unit('days since epoch')` ?

Which is kinda handy :wink:

hdyson: @bjlittle Nope, that's a new one on me - but one I'll certainly recycle.

lib/iris/tests/integration/test_pp.py
hdyson: @bjlittle I've only added a single test for grid_longitude mean without time mean- I expressed misgivings above about the potential redundancy of grid longitude tests.  That said, I can trivially replicate the longitude mean + time mean test to grid_longitude mean + time mean, if you prefer?

bjlittle:
@hdyson 
Travis is failing due to PEP8 failures ...
`test_pp.py:627:42: E127 continuation line over-indented for visual indent`

bjlittle:
@hdyson Push up your PEP8 fix and we'll get this merged :wink:

hdyson:
@bjlittle Apologies - I'd missed the '-a' when making the commit yesterday, so didn't even commit the fix, let alone push it.

Is something up with travis?  It seems to be having problems running the tests.

bjlittle:
@hdyson No worries. I'll try restarting Travis, although it seems to be a systemic problem (at their end?)

bjlittle:
@hdyson Hmmm seems like some dependencies may have moved under our feet. Not your doing. 

I'll need to investigate further, hang tight.

bjlittle:
@hdyson I try to recreate the problem locally so that I can identify the dependency issue and generate a fix.

bjlittle:
@hdyson Could you please rebase against upstream/master (see #2053) to include a fix that resolves the travis failures, thanks :smile:

hdyson:
@bjlittle Thanks, that solved it.

bjlittle:
@hdyson Finally, a clean run through travis, phew!

I'm :+1: for this PR so could you please squash your commits into one and I'll happily merge!

hdyson:
@bjlittle You may want to double check the squashed commit - it's not something I've done before.  Is it worth perhaps adding the procedure to the iris development workflow instructions?

Thanks for all your help on this, everyone.

bjlittle:
@hdyson Your squashed commit looks good to me.

Typically, if a PR contains many commits the reviewer can request that the author squashes their commits into one, as it makes the commit history on master that much less noisy. It's a normal practice, and a good skill to have, so congrats to you!

As of today, we've just changed the settings on the iris repo to take advantage of a new GitHub feature that allows the reviewer to automatically squash and merge multiple commits into one new merge commit. This saves the whole reviewer and author ding dong, requesting the author to squash. So, looking forward this shouldn't be much of an issue.

ajdawson:
@hdyson - we have some documentaion on this in the developer's guide: http://scitools.org.uk/iris/docs/latest/developers_guide/gitwash/development_workflow.html#rewriting-commit-history. Even though we can squash on merge through the Github GUI, it is still worth understanding this aspect of git as it can be very helpful.

bjlittle:
@hdyson Also, [progit](https://git-scm.com/book/en/v2) is a pretty handy free git resource.

bjlittle:
@hdyson ... and we're over the line! Thanks!

hdyson:
Thanks again everyone - much appreciated.  

@ajdawson Sorry, my mistake: I didn't make the mental connection between rewriting commit history and squashing commits, so I had the right page, but not the right understanding.  Consequently, I didn't read that section in sufficient depth to spot where it talks about combining commits.

PR #1976, opened 2016-04-20
Update deprecation rules

rhattersley:
Since the switch to semver the old two-release-or-six-months rule doesn't really make sense. This change simplifies it to just a one-release minimum.

(When someone chooses to _install_ a major release is an independent question.)

### DIFF
--- a/docs/iris/src/developers_guide/deprecations.rst
+++ b/docs/iris/src/developers_guide/deprecations.rst
@@ -2,11 +2,9 @@ Deprecations
 ************
 
 If you need to make a backwards-incompatible change to a public API
-[#public-api]_ that has been included in a release, then you should
-only make that change after a deprecation period. This deprecation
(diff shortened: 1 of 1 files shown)


pp-mo:
I'm :+1: for this.
Any dissenters ??

ajdawson:
:+1: let's do it

pp-mo:
Ooops, just spotted, the six-month/two-revisions part is mentioned again lower down ~Line 75
( can't link to a .rst ?)

pp-mo:
I think we should be mentioning SemVer explicity.
That can be handled as a separate issue though.

rhattersley:
> Ooops, just spotted, the six-month/two-revisions part is mentioned again lower down ~Line 75
> ( can't link to a .rst ?)

Well spotted :eyes: - thank you. I've updated that now.

PR #1978, opened 2016-04-20
Clip guessed latitude bounds to [-90, 90].

rhattersley:
Using `Coord.guess_bounds()` on a latitude coordinate which extends too close to the poles gives rise to "silly" bounds that extend beyond the range [-90, 90]. (e.g. The sample data `GloSea4/ensemble_001.pp`.)

This PR checks for a latitude coordinate, and if all the _points_ are inside the range [-90, 90] then it clips the guessed bounds to [-90, 90].

NB. Because this is a backwards-incompatible change of behaviour I've controlled it via an `iris.FUTURE.clip_latitudes` flag.

### DIFF
--- a/lib/iris/coords.py
+++ b/lib/iris/coords.py
@@ -1011,8 +1011,9 @@ class Coord(six.with_metaclass(ABCMeta, CFVariableMixin)):
 
         Kwargs:
 
-        * bound_position - The desired position of the bounds relative to the
-                           position of the points.
+        * bound_position:
+            The desired position of the bounds relative to the position
+            of the points.
 
         Returns:
             A numpy array of shape (len(self.points), 2).
@@ -1023,10 +1024,12 @@ class Coord(six.with_metaclass(ABCMeta, CFVariableMixin)):
 
         .. note::
 
-            If `iris.FUTURE.clip_latitudes` is True, then this
-            method will clip the coordinate bounds to the range
-            [-90, 90] for latitude and grid_latitude coordinates
-            measured in degrees.
+            If `iris.FUTURE.clip_latitudes` is True, then this method
+            will clip the coordinate bounds to the range [-90, 90] when:
+
+            - it is a `latitude` or `grid_latitude` coordinate,
+            - the units are degrees,
+            - all the points are in the range [-90, 90].
 
         """
         # XXX Consider moving into DimCoord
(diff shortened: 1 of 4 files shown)


lib/iris/tests/unit/coords/test_Coord.py
```diff
@@ -158,6 +159,58 @@ def test_circular_decreasing_alt_range(self):
...
+        lat.guess_bounds()
+        self.assertArrayEqual(lat.bounds, [[-150, -50], [-50, 60], [60, 180]])
+
+
+class Test_guess_bounds__enabled_latitude_clipping(tests.IrisTest):
+    def setUp(self):
+        iris.FUTURE.clip_latitudes = True
+
+    def tearDown(self):
+        iris.FUTURE.clip_latitudes = True
```
bjlittle: @rhattersley Did you mean for this to be `True` in the `tearDown` ?

rhattersley: Doh!

lib/iris/coords.py
```diff
@@ -1056,6 +1056,11 @@ def _guess_bounds(self, bound_position=0.5):
 
         bounds = np.array([min_bounds, max_bounds]).transpose()
 
+        if iris.FUTURE.clip_latitudes and 'latitude' in self.name():
+            points = self.points
+            if (points >= -90).all() and (points <= 90).all():
+                np.clip(bounds, -90, 90, out=bounds)
```
bjlittle: @rhattersley We might need a `units` check here ...

rhattersley: :+1:

rhattersley: And something in a docstring might be nice ... :wink:

lib/iris/coords.py
```diff
@@ -1021,6 +1021,13 @@ def _guess_bounds(self, bound_position=0.5):
 
             This method only works for coordinates with ``coord.ndim == 1``.
 
+        .. note::
+
+            If `iris.FUTURE.clip_latitudes` is True, then this
+            method will clip the coordinate bounds to the range
+            [-90, 90] for latitude and grid_latitude coordinates
+            measured in degrees.
```
bjlittle: @rhattersley Is it also worth explicitly saying that "... it clips iff *all* points are within [-90, 90]"

rhattersley: Double doh! :finnadie:

bjlittle:
@rhattersley Looks good to me :smile:  ... just waiting on Travis, then we'll merge.

Thanks! :+1:

cpelley:
Thanks @rhattersley, very useful.  We have had our own `guess_bounds` function in our project for a long time now.  Will be good to deprecate it. :+1:

PR #1980, opened 2016-04-20
Fix the plot rendering within sphinx.

### LINKED ISSUE #1979: Broken example code
See https://travis-ci.org/SciTools/iris/jobs/124401763#L772

rhattersley:
1. Sphinx splits the code when it sees `plt.show` which creates invalid syntax for code with this kind of structure. We use `qplt.show` to avoid that, as for the other examples.
2. Because sphinx doesn't get to do its splitting thing, we have to explicitly create a new figure after each show().

Fixes #1979.


rhattersley:
I restarted the travis build that had timed out and now we have a lovely green tick.

PR #1983, opened 2016-04-25
ENH: Lazy in-place Cube transpose

cpelley:
1. Simply utilises the capabilities of biggus to provide a transposed version of the data rather than making a copy (necessary with the sizes of data we are working with).
2. ~~Using 1, also provides in-place dimension inversion (intentionally constrained implementation).~~ - now moving to a new PR after the merge of this one.

### DIFF
--- a/docs/iris/src/whatsnew/contributions_1.10/newfeature_2016-May-20_lazy_cube_transpose.txt
+++ b/docs/iris/src/whatsnew/contributions_1.10/newfeature_2016-May-20_lazy_cube_transpose.txt
@@ -1,0 +1,1 @@
+* The transpose method of a Cube now results in a transposed view of the original data rather than a transposed copy.
--- a/docs/iris/src/whatsnew/contributions_1.10/newfeature_2016-May-20_lazy_cube_transpose.txt
+++ b/docs/iris/src/whatsnew/contributions_1.10/newfeature_2016-May-20_lazy_cube_transpose.txt
@@ -1,1 +1,1 @@
-* The transpose method of a Cube now results in a transposed view of the original data rather than a transposed copy.
+* The transpose method of a Cube now results in a lazy transposed view of the original rather than realising the data then transposing it.
--- a/lib/iris/cube.py
+++ b/lib/iris/cube.py
@@ -2806,10 +2806,7 @@ bound=(1994-12-01 00:00:00, 1998-12-01 00:00:00)
         elif len(new_order) != self.data.ndim:
             raise ValueError('Incorrect number of dimensions.')
 
-        # The data needs to be copied, otherwise this view of the transposed
-        # data will not be contiguous. Ensure not to assign via the cube.data
-        # setter property since we are reshaping the cube payload in-place.
-        self._my_data = np.transpose(self.data, new_order).copy()
+        self._my_data = self.lazy_data().transpose(new_order)
 
         dim_mapping = {src: dest for dest, src in enumerate(new_order)}
 
--- a/lib/iris/cube.py
+++ b/lib/iris/cube.py
@@ -2802,11 +2802,14 @@ bound=(1994-12-01 00:00:00, 1998-12-01 00:00:00)
 
         """
         if new_order is None:
-            new_order = np.arange(self.data.ndim)[::-1]
-        elif len(new_order) != self.data.ndim:
+            new_order = np.arange(self.ndim)[::-1]
+        elif len(new_order) != self.ndim:
             raise ValueError('Incorrect number of dimensions.')
 
-        self._my_data = self.lazy_data().transpose(new_order)
+        if self.has_lazy_data():
+            self._my_data = self.lazy_data().transpose(new_order)
+        else:
+            self._my_data = self.data.transpose(new_order)
 
         dim_mapping = {src: dest for dest, src in enumerate(new_order)}
 
--- a/lib/iris/tests/unit/cube/test_Cube.py
+++ b/lib/iris/tests/unit/cube/test_Cube.py
@@ -1410,5 +1410,14 @@ class TestCellMeasures(tests.IrisTest):
             cm_dims = self.cube.cell_measure_dims(a_cell_measure)
 
 
+class Test_transpose(tests.IrisTest):
+    def test_lazy_data(self):
+        data = np.arange(12).reshape(3, 4)
+        cube = Cube(biggus.NumpyArrayAdapter(data))
+        cube.transpose()
+        self.assertTrue(cube.has_lazy_data())
+        self.assertArrayEqual(data.T, cube.data)
+
+
 if __name__ == '__main__':
     tests.main()
--- a/lib/iris/tests/unit/cube/test_Cube.py
+++ b/lib/iris/tests/unit/cube/test_Cube.py
@@ -1418,6 +1418,14 @@ class Test_transpose(tests.IrisTest):
         self.assertTrue(cube.has_lazy_data())
         self.assertArrayEqual(data.T, cube.data)
 
+    def test_not_lazy_data(self):
+        data = np.arange(12).reshape(3, 4)
+        cube = Cube(data)
+        cube.transpose()
+        self.assertFalse(cube.has_lazy_data())
+        self.assertIs(data.base, cube.data.base)
+        self.assertArrayEqual(data.T, cube.data)
+
 
 if __name__ == '__main__':
     tests.main()


cpelley:
`iris.tests.experimental.regrid.test_regrid_area_weighted_rectilinear_src_and_grid.TestAreaWeightedRegrid` are failling on travis but I'm not seeing any issue locally :(

Any ideas anyone?

Chers

ajdawson:
> failling on travis but I'm not seeing any issue locally :( Any ideas anyone?

We should increase the timeout for nose multiprocessing...

ajdawson:
So, can you describe how the inversion would proceed if I invert a 1d dimension coordinate that has a 1d Auxcoord describing it also?

rhattersley:
> We should increase the timeout for nose multiprocessing...

:+1: ... or speed up/remove the offending re-projection tests.

cpelley:
> So, can you describe how the inversion would proceed if I invert a 1d dimension coordinate that has a 1d Auxcoord describing it also?

Certainly, it will fall over as I have called `self.coord` not `self.coords` (exception [here](https://github.com/cpelley/iris/blob/LAZY_CUBE_TRANSPOSE/lib/iris/cube.py#L1436)).
I can add a test if you really want me to but I see no benefit to adding duplicate testing for `Cube.coord`.

Cheers

Thanks both for your interest on this.

cpelley:
This implementation is intentionally constrained.  In future it will support multi-dimensional coordinates and multiple coordinates mapped to the same dimension (as/when driven by requirement).

cpelley:
>>  We should increase the timeout for nose multiprocessing...

> :+1: ... or speed up/remove the offending re-projection tests.

Thanks both, I hadn't noticed that it was a `TimedOutException` raised.  I'll look at how how easy they might be to migrate convert to true unittests.

ajdawson:
>This implementation is intentionally constrained. In future it will support multi-dimensional coordinates and multiple coordinates mapped to the same dimension (as/when driven by requirement).

I wanted to know what happens in the case I specified. Is an exception raised? Where is it raised from?

lib/iris/cube.py
```diff
@@ -2785,6 +2785,34 @@ def slices(self, ref_to_slice, ordered=True):
...
+        """
+        dimensions = tuple(set([int(dim) for dim in dimensions]))
+
+        for dim in dimensions:
+            coord = self.coord(contains_dimension=dim)
+            coord_dims = self.coord_dims(coord)
+            if len(coord_dims) > 1:
+                msg = ('Currently multidimensional coordinate inversion is '
+                       'not supported')
+                raise RuntimeError(msg)
```
ajdawson: Is `RuntimeError` the best class of exception to be used here?

ajdawson: Perhaps `CoordinateMultiDimError` would be better? (https://github.com/SciTools/iris/blob/master/lib/iris/exceptions.py#L48)

cpelley:
> I wanted to know what happens in the case I specified. Is an exception raised? Where is it raised from?

@ajdawson I thought I already did?

@cpelley: "Certainly, it will fall over as I have called [self.coord](https://github.com/SciTools/iris/pull/1983/files#diff-d1f30240d0d146f09f58577e718ba6a6R2801) not self.coords (exception [here](https://github.com/cpelley/iris/blob/LAZY_CUBE_TRANSPOSE/lib/iris/cube.py#L1436))..."

Perhaps an example might help?

```Python
>>> print cube
thingness / (1)                     (bar: 3; foo: 4)
     Dimension coordinates:
          bar                           x       -
          foo                           -       x
     Auxiliary coordinates:
          bing                          x       -
>>> cube.invert_dims((0,))
  File "/home/h05/cpelley/git/iris/lib/iris/cube.py", line 2801, in invert_dims
    coord = self.coord(contains_dimension=dim)
  File "/home/h05/cpelley/git/iris/lib/iris/cube.py", line 1439, in coord
    raise iris.exceptions.CoordinateNotFoundError(msg)
iris.exceptions.CoordinateNotFoundError: 'Expected to find exactly 1 coordinate, but found 2. They were: bar, bing.'
```

lib/iris/cube.py
cpelley: Thanks @ajdawson, sure.

ajdawson:
See above modified comment.

cpelley:
> You may want to consider catching this exception and re-raising with a more appropriate error message?

I thought the exception was clear enough along with the traceback that says where it was raised (see above example).

ajdawson:
> I thought the exception was clear enough

It doesn't tell you what actually went wrong though. The real problem is that in-place inversion doesn't support multiple coordinates spanning a single dimension. The end-user shouldn't have to know that you called the `coord` method to determine this (or know anything about the functionality and exceptions raised by the `coord` method). You should raise an exception that describes the immediate problem, in this case something like:

```
Inversion of dimensions that are spanned by multiple coordinates is not currently supported.
Dimension {} is spanned by ...
```

lib/iris/cube.py
```diff
@@ -2785,6 +2785,34 @@ def slices(self, ref_to_slice, ordered=True):
...
+            Determines which dimensions will be inverted.
+
+        """
+        dimensions = tuple(set([int(dim) for dim in dimensions]))
+
+        for dim in dimensions:
+            coord = self.coord(contains_dimension=dim)
+            coord_dims = self.coord_dims(coord)
+            if len(coord_dims) > 1:
+                msg = ('Currently multidimensional coordinate inversion is '
```
ajdawson: You shouldn't start the sentence with "Currently", try something along the lines of:

```
Inversion of multidimensional coordinates is not supported.
```

lib/iris/cube.py
```diff
@@ -2785,6 +2785,34 @@ def slices(self, ref_to_slice, ordered=True):
 
         return _SliceIterator(self, dims_index, dim_to_slice, ordered)
 
+    def invert_dims(self, dimensions):
+        """
+        Invert the specified dimensions of a cube in-place.
+
+        Args:
+
+        * dimensions (iterable of ints):
```
ajdawson: Would it be useful to be able to specify the coordinate by name or coordinate instance here in stead of number? If we imagine a future time when inversion of dimensions with AuxCoords spanning them is implemented, this may be very useful. Since this affects the public API it should be ironed out now (the actual implementation for AuxCoords/multi-dimensional coords can indeed be added at a later time as the need arises).

I don't think this would change the implementation much at all. We already have a way of accepting "string or instance" as input which you could reuse. Then you just proceed as before, check if the dimension spanned by the coord has other coordinates or if it spans multiple dimensions and reject if so.

rhattersley: > Since this affects the public API it should be ironed out now...

@ajdawson: Thanks for getting this back on my radar with the magic words "public API". :wink: 

@cpelley: :+1: for lazy tranpose, but `invert_dims` seems too niche to be on the Cube. What's stopping you from indexing your cube with a `::-1` slice? (NB. I'd prefer to deal with @ajdawson's excellent suggestion of being able to use coordinate names by using pandas/xarray style name-based indexing and have been trying to put together [a proposal](https://github.com/SciTools/iris/wiki/IEP-1) that would be relevant.)

cpelley:
> What's stopping you from indexing your cube with a[::-1] slice?

If the data is not lazy, that would create a copy (i.e. consume twice as much as necessary).

> ...but invert_dims seems too niche to be on the Cube

OK.  I followed the following structural layout for util within our own project (with the location describing the context of what objects the utility works on):
`util.coords`, `util.ndarray`, `util.cube` etc.

I bring this up only as I would guess that my adding yet another function to `util` will bring up the debate of the complexity and length of `utils`.

Would you be happy for me to put follow my proposed structure change (making util a sub-package and putting `invert_dims` within it?
(i.e. `iris/util/cube/invert_dims`)

cpelley:
I have re-written some of the `TestAreaWeightedRegrid` tests over in the unit testing area, however, those causing a `TimedOutException` probably need removal and completely new true unit tests written.  I don't think this is trivial enough for this PR as it would require evaluation of new numbers for the area weighted regrid results by hand.

I'll await a response to my last comment before moving `invert_dims` out of the Cube class.

Cheers both.

rhattersley:
>> What's stopping you from indexing your cube with a[::-1] slice?
> 
> If the data is not lazy, that would create a copy (i.e. consume twice as much as necessary).

What if that was [no longer the case](https://github.com/SciTools/iris/pull/939)?

cpelley:
> @rhattersley: What if that was no longer the case?

Are you suggesting that indexing doesn't copy or that you are open to removing the need to copying data by indexing the cube?

cpelley:
I think I'm tempted to limit this PR to the initial commit (ENH: Lazy cube data transpose - [00dae8](https://github.com/SciTools/iris/pull/1983/commits/00dae8b1d843681d47ee759214de5d0517227773)) unless there are any objections?

ajdawson:
Not from me, we can revisit the inversion issue in a separate PR if required.

cpelley:
I have reverted the cube_inverse idea so we are left with the simple transpose change (I'll put a separate PR up after this PR is merged dealing with the invert_dims).

Cheers

ajdawson:
I don't see why you are modifying tests about regridding, this change has nothing to do with the regrid code as far as I can see.

cpelley:
> I don't see why you are modifying tests about regridding, this change has nothing to do with the regrid code as far as I can see.

See [comment](https://github.com/SciTools/iris/pull/1983#issuecomment-214775481)

ajdawson:
I don't know if you need to do anything about it in this PR, it has nothing to do with the changes you have introduced. Changes to these tests should probably be in a separate PR where we can review carefully and have some record of the aim and scope of the changes.

cpelley:
> Changes to these tests should probably be in a separate PR...

I only did so because possible solutions were discussed on this ticket.
Updated commit removing these changes added and created a new PR (#1986 to continue the issue of test failure of the regrid on travis).
I'll squash when your happy.

Cheers

lib/iris/tests/unit/cube/test_Cube.py
```diff
@@ -1410,5 +1410,15 @@ def test_fail_cell_measure_dims(self):
...
 
 
+class Test_transpose(tests.IrisTest):
+    def setUp(self):
+        self.cube = stock.lat_lon_cube()
+
+    def test_inplace_data_transpose(self):
+        target_id = id(self.cube.lazy_data())
+        self.cube.transpose((0, 1))
+        self.assertEqual(id(self.cube.lazy_data().array), target_id)
```
rhattersley: By using `.array` you're relying on a particular implementation of the transpose algorithm. I'd suggest checking the cube is still lazy with `has_lazy_data` and then resolving the numbers and checking they actually match.

Also, I'd do an actual transpose rather than deliberately doing a "null" transpose.

Putting those together:
```python
data = np.arange(12).reshape(3, 4)
cube = Cube(biggus.NumpyArrayAdapter(data))
cube.transpose()
self.assertTrue(cube.has_lazy_data())
self.assertArrayEqual(data.T, cube.data)
```

lib/iris/tests/unit/cube/test_Cube.py
```diff
@@ -1410,5 +1410,15 @@ def test_fail_cell_measure_dims(self):
             cm_dims = self.cube.cell_measure_dims(a_cell_measure)
 
 
+class Test_transpose(tests.IrisTest):
+    def setUp(self):
+        self.cube = stock.lat_lon_cube()
+
+    def test_inplace_data_transpose(self):
```
rhattersley: The repetition of "transpose" doesn't add anything, and it doesn't mention that we're interested in the laziness. Perhaps just `test_lazy_data`?

rhattersley:
> Are you suggesting that indexing doesn't copy or that you are open to removing the need to copying data by indexing the cube?

I'm suggesting that we need to change cube indexing so that it doesn't copy the data.

lib/iris/tests/unit/cube/test_Cube.py
cpelley: Good point, thanks @rhattersley

lib/iris/tests/unit/cube/test_Cube.py
cpelley: Thanks @rhattersley.

cpelley:
Should be ready now, let me know and I'll squash :)

Cheers

rhattersley:
Squash-away my friend! Squash-away!

cpelley:
squashed :)

Cheers

rhattersley:
> ooh just remembered, don't we need to update what's new for this?

My turn to remember the "what's new" entry! :wink:

rhattersley:
> ooh just remembered, don't we need to update what's new for this?
> 
> My turn to remember the "what's new" entry! :wink:

@cpelley In other words... the only thing this needs is a what's new entry and it's ready to merge.

cpelley:
Thanks @rhattersley yes sorry I'm on it.

cpelley:
whatsnew done.

Cheers

docs/iris/src/whatsnew/contributions_1.10/newfeature_2016-May-20_lazy_cube_transpose.txt
```diff
@@ -0,0 +1 @@
+* The transpose method of a Cube now results in a transposed view of the original data rather than a transposed copy.
```
rhattersley: I'm not sure this really captures the lazy-loading aspect...?

cpelley:
I have updated the wording of the whatsnew.

Cheers

cpelley:
Can we get this into the v1.10 milestone?

dkillick:
@cpelley realistically no - v1.10 is big enough already and we just need to get it out the door now rather than adding new functionality to the milestone that might hold it up.

cpelley:
> @cpelley realistically no - v1.10 is big enough already and we just need to get it out the door now rather than adding new functionality to the milestone that might hold it up.

Can you please reconsider?  This ticket has gone through a review with the only remaining thing flagged by @rhattersley being the whatsnew entry.  This was completed over two weeks ago.

djkirkham:
I know I'm late to the party, but this change will mean that when transpose() is called on a cube without lazy data, it will then appear to have lazy data, i.e. cube.has_lazy_data() will return true.

cpelley:
Thanks for your interest @djkirkham.  This behaviour was expected: [iris.tests.unit.cube.test_Cube.Test_transpose.test_lazy_data](https://github.com/SciTools/iris/pull/1983/files#diff-1a002eee61524000a58a8e337597c7d7)

To modify the behaviour of `has_lazy_data` (i.e. change the implied meaning of this method) I think is outside the scope of this ticket.
Take for example the following simple example that based on your comment you might also have a problem with?

```Python
>>> arr = biggus.NumpyArrayAdapter(np.arange(4).reshape(2, 2))
>>> cube = iris.cube.Cube(arr)
>>> cube.has_lazy_data()
True
```

I should say that I don't have a problem with the above (but then I have taken the meaning of iris `lazy_data` to be the behaviour of `has_lazy_data`, which I think is perfectly reasonable especially given that there is no docstring.  Initialising a cube with a biggus array is a very use case in case.

Is there a preference for doing some type checking in transpose?
This would be contrary to current iris behaviour in similar situations like shown in my example above.

Cheers

djkirkham:
I'm admittedly a novice when it comes to Iris, but my understanding of the meaning of `has_lazy_data()` is "has the data been loaded from its initial data source?". Thus, I don't have a problem with your example, since the data hasn't been "loaded" from the data source, which happens to be a numpy array.

For me, the usefulness of `has_lazy_data()` comes from the ability to use it to check whether cube operations load in the data. With the current change that ability is lost with respect to transpose.

> Is there a preference for doing some type checking in transpose?

My preference would be to use `self.lazy_data()` if `self.has_lazy_data()` returns `True`, and use `self.data` otherwise, although maybe that defeats the object of this PR.

cpelley:
Thanks @djkirkham I'll have a closer look and see what I can do.

Cheers

cpelley:
> ...Thus, I don't have a problem with your example, since the data hasn't been "loaded" from the data source, which happens to be a numpy array.

I don't see how the term "load" comes into it in the context of my example.  Right now, `has_lazy_data` simply means "is biggus array".  I suggest we add documentation to `has_lazy_data` to clarify that (new ticket).

I think I agree in principle with your expected return value though.  I have pushed a new commit which now does this.

Thanks @djkirkham

djkirkham:
Thanks @cpelley, the changes look good.

>  Right now, has_lazy_data simply means "is biggus array".

I would say that's just the way it's implemented, rather than the intended meaning of the method. Correct me if I'm wrong, but currently the only way `cube._my_data` ever gets assigned a biggus array is if it's passed in by the user, and if the data in that array is ever loaded, `cube._my_data` gets assigned a NumPy array.

I agree that it would be a good idea to have the intended meaning documented somewhere.

cpelley:
Ping, the test failures are timeouts.

Cheers

cpelley:
ping

cpelley:
Please let me know if there's anything I can do to help to get this in.

Cheers

marqh:
> Please let me know if there's anything I can do to help to get this in.
> This would provide very useful capability for CMIP6 due to the sheer size of data involved.

this looks like a done deal to me, thanks @cpelley

cpelley:
Thanks @marqh, this will make a significant difference to us.  Appreciated.

PR #1984, opened 2016-04-25
Remove obsolete import of iris.fileformats.dot

rhattersley:
This explicit import should not be needed - the import is [intended to be deferred](https://github.com/rhattersley/iris/blob/20d3a7841b9d80aa7dfcc0fbb9c8910d22cf8ba4/lib/iris/io/__init__.py#L229-L242).


pp-mo:
Can't argue with that !

PR #1985, opened 2016-04-25
Watch out for broken `dot` installations.

### LINKED ISSUE #1899: Importing iris shouldn't result in a subprocess.check_call
https://github.com/SciTools/iris/pull/1266 introduced an automatic check_call to identify the dot executable. Because of our import machinery, this ends up getting done when ```import iris``` is run. Clearly that is not desirable - ideally it should only run when we actually *want* dot to work.

rhattersley:
Closes #1899.

### DIFF
--- a/lib/iris/fileformats/dot.py
+++ b/lib/iris/fileformats/dot.py
@@ -1,4 +1,4 @@
-# (C) British Crown Copyright 2010 - 2015, Met Office
+# (C) British Crown Copyright 2010 - 2016, Met Office
 #
 # This file is part of Iris.
 #
@@ -33,20 +33,35 @@ import iris.util
 _GRAPH_INDENT = ' ' * 4
 _SUBGRAPH_INDENT = ' ' * 8
 
-_DOT_EXECUTABLE_PATH = iris.config.get_option('System', 'dot_path',
-                                              default='dot')
-if not os.path.exists(_DOT_EXECUTABLE_PATH):
-    if not os.path.isabs(_DOT_EXECUTABLE_PATH):
-        try:
-            # Check PATH
-            subprocess.check_output([_DOT_EXECUTABLE_PATH, '-V'],
-                                    stderr=subprocess.STDOUT)
-        except OSError:
-            _DOT_EXECUTABLE_PATH = None
+
+_DOT_CHECKED = False
+_DOT_EXECUTABLE_PATH = None
+
+
+def _dot_path():
+    global _DOT_CHECKED, _DOT_EXECUTABLE_PATH
+
+    if _DOT_CHECKED:
+        path = _DOT_EXECUTABLE_PATH
     else:
-        _DOT_EXECUTABLE_PATH = None
+        path = iris.config.get_option('System', 'dot_path', default='dot')
+        if not os.path.exists(path):
+            if not os.path.isabs(path):
+                try:
+                    # Check PATH
+                    subprocess.check_output([path, '-V'],
+                                            stderr=subprocess.STDOUT)
+                except (OSError, subprocess.CalledProcessError):
+                    path = None
+            else:
+                path = None
+        _DOT_EXECUTABLE_PATH = path
+        _DOT_CHECKED = True
+    return path
+
+
 #: Whether the 'dot' program is present (required for "dotpng" output).
-DOT_AVAILABLE = _DOT_EXECUTABLE_PATH is not None
+DOT_AVAILABLE = _dot_path() is not None
 
 
 def save(cube, target):
@@ -104,17 +119,17 @@ def save_png(source, target, launch=False):
         raise ValueError("Can only write dot png for a Cube or DOT file")
 
     # Create png data
-    if not _DOT_EXECUTABLE_PATH:
+    if not _dot_path():
         raise ValueError('Executable "dot" not found: '
                          'Review dot_path setting in site.cfg.')
     # To filename or open file handle?
     if isinstance(target, six.string_types):
-        subprocess.call([_DOT_EXECUTABLE_PATH, '-T', 'png', '-o', target,
+        subprocess.call([_dot_path(), '-T', 'png', '-o', target,
                          dot_file_path])
     elif hasattr(target, "write"):
         if hasattr(target, "mode") and "b" not in target.mode:
             raise ValueError("Target not binary")
-        subprocess.call([_DOT_EXECUTABLE_PATH, '-T', 'png', dot_file_path],
+        subprocess.call([_dot_path(), '-T', 'png', dot_file_path],
                         stdout=target)
     else:
         raise ValueError("Can only write dot png for a filename or writable")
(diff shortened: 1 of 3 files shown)


lib/iris/fileformats/dot.py
```diff
@@ -33,20 +33,35 @@
...
+            else:
+                path = None
+        _DOT_EXECUTABLE_PATH = path
+        _DOT_CHECKED = True
+    return path
+
+
 #: Whether the 'dot' program is present (required for "dotpng" output).
-DOT_AVAILABLE = _DOT_EXECUTABLE_PATH is not None
+DOT_AVAILABLE = _dot_path() is not None
```
pp-mo: Surely, with this in place then it is **still making the call at import time** ?

I don't think DOT_AVAILABLE is used anywhere, but then I suppose it _is_ now part of the public API.

If that's important, then its awkward, as you can't wrap a module variable as a "property".
I suppose you could define an object with a custom "bool" conversion ??
It can be done by defining a special method: see http://stackoverflow.com/questions/1087135/boolean-value-of-objects-in-python

Pretty obscure + nasty though...

lib/iris/fileformats/dot.py
```diff
@@ -33,20 +33,35 @@
...
     else:
-        _DOT_EXECUTABLE_PATH = None
+        path = iris.config.get_option('System', 'dot_path', default='dot')
+        if not os.path.exists(path):
+            if not os.path.isabs(path):
+                try:
+                    # Check PATH
+                    subprocess.check_output([path, '-V'],
+                                            stderr=subprocess.STDOUT)
+                except (OSError, subprocess.CalledProcessError):
```
rhattersley: This is the key change: adding `CalledProcessError`. It allows us to catch circumstances where we have the path to an installation of dot that explodes when you run it.

lib/iris/fileformats/dot.py
rhattersley: We don't need to defer the check at import time and that's not what this PR is about. The deferred aspect was already taken care of in #1900.

The only reason I stuck the logic in a function is so I can more easily test it.

lib/iris/tests/unit/fileformats/dot/test__dot_path.py
```diff
@@ -0,0 +1,51 @@
...
+import os.path
+import subprocess
+
+import mock
+
+from iris.fileformats.dot import _dot_path
+
+
+class Test(tests.IrisTest):
+    def test_broken_install(self):
```
pp-mo: Might we also want to test this when it *does* have a valid path ?
Mind you, it must work because the dot functionality itself is tested elsewhere (```tests.test_cdm``` and ```tests.test_file_save```)

pp-mo:
> Might we also want to test this when it does have a valid path ?

OWOK, I think

lib/iris/fileformats/dot.py
pp-mo: > We don't need to defer the check.. already taken care of

Sorry, got that now !

rhattersley:
Tests extended and a tick from travis. Over to you @pp-mo ...

lib/iris/tests/unit/fileformats/dot/test__dot_path.py
```diff
@@ -0,0 +1,82 @@
...
+
+from iris.fileformats.dot import _dot_path
+
+
+class Test(tests.IrisTest):
+    def setUp(self):
+        # Because _dot_path is triggered by the initial import we
+        # reset the caching status to allow us to see what happens
+        # under different circumstances.
+        patcher = mock.patch('iris.fileformats.dot._DOT_CHECKED', new=False)
```
pp-mo: Did you do know you can just use self.patch(...) to do all of this for you ??

lib/iris/tests/unit/fileformats/dot/test__dot_path.py
```diff
@@ -0,0 +1,82 @@
...
+
+    def test_invalid_absolute_path(self):
+        # Override the configuration value for System.dot_path
+        dummy_path = '/not_a_real_path' * 10
+        assert not os.path.exists(dummy_path)
+        with mock.patch('iris.config.get_option', return_value=dummy_path):
+            result = _dot_path()
+        self.assertIsNone(result)
+
+    def test_valid_relative_path(self):
```
pp-mo: Not sure if it's too picky but the "valid" caused me some raised eyebrows

rhattersley: I reckon you are being too "picky". As far as the routine under test is concerned it _is_ valid relative path - it even managed to "execute it".

lib/iris/tests/unit/fileformats/dot/test__dot_path.py
rhattersley: Thanks for the reminder.

lib/iris/tests/unit/fileformats/dot/test__dot_path.py
pp-mo: but it had to *fail* a "path.exists" check to get there...

QuLogic:
The previous issue was tagged for 1.9; is this going to be backported?

rhattersley:
I guess that depends if there's sufficient appetite to make another 1.9 release with v1.10 being relatively close.

lib/iris/fileformats/dot.py
envhyf: I met the same problem when I import `iris`. I edited the `dot.py` taking @rhattersley's work  as reference. But what beneath the ` except (OSError, subprocess.CalledProcessError):`?   


    ` 36 _DOT_EXECUTABLE_PATH = iris.config.get_option('System', 'dot_path',
     37                                               default='dot')
     38 if not os.path.exists(_DOT_EXECUTABLE_PATH):
     39     if not os.path.isabs(_DOT_EXECUTABLE_PATH):
     40         try:
     41             # Check PATH
     42             subprocess.check_output([_DOT_EXECUTABLE_PATH, '-V'],
     43                                     stderr=subprocess.STDOUT)
     44         except OSError:
     45             _DOT_EXECUTABLE_PATH = None
     46     else:
     47         path = iris.config.get_option('System', 'dot_path', default='dot')
     48         if not os.path.exists(path):
     49             if not os.path.isabs(path):
     50                 try:
     51                     # Check PATH
     52                     subprocess.check_output([path, '-V'],
     53                                             stderr=subprocess.STDOUT)
     54                 except (OSError, subprocess.CalledProcessError):
     55 # Edit in 20160606
     56 #        _DOT_EXECUTABLE_PATH = None
     57 #: Whether the 'dot' program is present (required for "dotpng" output).
     58 #DOT_AVAILABLE = _DOT_EXECUTABLE_PATH is not None
     59 DOT_AVAILABLE = _dot_path() is not None
     60 
     61 def save(cube, target):
     62     """Save a dot representation of the cube.
     63 
     64     Args:`

PR #1989, opened 2016-04-27
Update code examples to avoid deprecations.

pp-mo:
I've added FUTURE options to avoid any deprecations in the examples code.
I also fixed all examples tests so they will fail if they hit a deprecation.
Rationale : the example code is supposed to be "best practice", but we have failed to keep it up-to-date.  In future, we'll see that they need updating when they fail.

N.B. They could probably still do with a critical review.
This is just one semi-automatic improvement.

### DIFF
--- a/docs/iris/example_code/General/SOI_filtering.py
+++ b/docs/iris/example_code/General/SOI_filtering.py
@@ -52,6 +52,8 @@ def low_pass_weights(window, cutoff):
 
 
 def main():
+    # Enable future options to avoid deprecations.
+    iris.FUTURE.netcdf_promote = True
 
     # Load the monthly-valued Southern Oscillation Index (SOI) time-series.
     fname = iris.sample_data_path('SOI_Darwin.nc')
--- a/docs/iris/example_code/General/anomaly_log_colouring.py
+++ b/docs/iris/example_code/General/anomaly_log_colouring.py
@@ -36,6 +36,9 @@ import matplotlib.colors as mcols
 
 
 def main():
+    # Enable future options to avoid deprecations.
+    iris.FUTURE.netcdf_promote = True
+
     # Load a sample air temperatures sequence.
     file_path = iris.sample_data_path('E1_north_america.nc')
     temperatures = iris.load_cube(file_path)
--- a/docs/iris/example_code/General/cross_section.py
+++ b/docs/iris/example_code/General/cross_section.py
@@ -15,8 +15,11 @@ import iris.quickplot as qplt
 
 
 def main():
+    # Enable future options to avoid deprecations.
+    iris.FUTURE.netcdf_promote = True
+
     fname = iris.sample_data_path('hybrid_height.nc')
-    theta = iris.load_cube(fname)
+    theta = iris.load_cube(fname, 'air_potential_temperature')
 
     # Extract a single height vs longitude cross-section. N.B. This could
     # easily be changed to extract a specific slice, or even to loop over *all*
--- a/docs/iris/example_code/General/custom_aggregation.py
+++ b/docs/iris/example_code/General/custom_aggregation.py
@@ -65,6 +65,9 @@ def count_spells(data, threshold, axis, spell_length):
 
 
 def main():
+    # Enable future options to avoid deprecations.
+    iris.FUTURE.netcdf_promote = True
+
     # Load the whole time-sequence as a single cube.
     file_path = iris.sample_data_path('E1_north_america.nc')
     cube = iris.load_cube(file_path)
--- a/docs/iris/example_code/General/custom_file_loading.py
+++ b/docs/iris/example_code/General/custom_file_loading.py
@@ -59,6 +59,8 @@ import datetime
 import matplotlib.pyplot as plt
 import numpy as np
 
+from cf_units import Unit, CALENDAR_GREGORIAN
+
 import iris
 import iris.coords as icoords
 import iris.coord_systems as icoord_systems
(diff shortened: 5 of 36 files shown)


ajdawson:
># Enable future options to avoid deprecations.

I'd argue that in most cases you are not avoiding deprecations. Since you only change the future option and none of the existing code in most cases, you are avoiding *deprecation warnings*. This is an important distinction, especially in the examples.

pp-mo:
>  you are avoiding deprecation warnings. This is an important distinction, especially in the examples.

I don't quite get this.
Do you have a specific example of the difference, e.g. where I am missing some remaining deprecated usage ?

My understanding :
According to our developer guide, deprecations warnings and usage "ought" to be the same thing, as when you deprecate something you ensure that runtime warnings also occur wherever appropriate.

The test changes make all deprecation warnings into errors, so we must avoid them.  
That means the tests must avoid using any deprecated features (in principle, at least).

I'm nearly ready to post a user-oriented guide to change management covering versions, releases, deprecations and the FUTURE options.
So I'm really interested in getting this stuff right, and maybe that will expose where I am wrong !

pp-mo:
Oh dear, those errors are awkward + I think I'm going to have to make a specific exception which is scarcely in the spirit of this PR !

This has prompted me to also error all DeprecationWarnings.
That makes sense, and has exposed another (soluble) problem in numpy usage (see on).
But I can't reproduce _this_ actual error on my desktop install :angry:

pp-mo:
> I can't reproduce this actual error 

Oh, I can if I use an env with a later dateutil.
Added a fix for this..

ajdawson:
What I mean is perhaps best illustrated by picking one of the example files you have changed, lets choose `docs/iris/example_code/General/SOI_filtering.py`. The only change you have made to this file is to set a future option, you made no changes to the rest of the example. Since this example runs properly with or without the future option set, it therefore isn't really using any deprecated functionality, so you aren't fixing it because it uses a deprecated feature, you are just avoiding seeing a warning about a deprecation that doesn't affect the code at all.

My quibble is entriely with terminology in the comment above the setting of the future option. It makes it look like a failure due to use of a deprecated feature is being avoided, but in fact this is not correct, it is just a warning that is avoided.

Really the issue is with an over-zealous deprecation warning in the first place. The data load in the `SOI_filtering.py` example causes the deprecation warning:

```
UserWarning: NetCDF default loading behaviour currently does not expose variables which
define reference surfaces for dimensionless vertical coordinates as independent Cubes.
This behaviour is deprecated in favour of automatic promotion to Cubes. To switch to the
new behaviour, set iris.FUTURE.netcdf_promote to True.
```

However, the loaded file file contains no such coordinate (it contains a variable called "SOI_Darwin" and a coordinate variable named "time" and that is all), and as such the warning is absolutely redundant, and in fact misleading...

rhattersley:
> over-zealous

i.e. buggy :unamused:

ajdawson:
>>over-zealous

>i.e. buggy ￼

Yeah basically. It doesn't seem right to warn users they are using deprecated behaviour when they aren't. I think the same problem probably exists with the grib regularize warning.

In terms of the example code, I'd argue that there doesn't need to be a change to most of these example programs, because they don't actually use the deprecated behaviour. The problem is with the warnings themselves.

pp-mo:
Hmmm.
I think, considering the relevance of the deprecation to a given piece of user code, it **may** still make sense to issue it in this case, because the behaviour depends on the contents of the file, which might well be different next time he runs it.
So there is a rationale for issuing the warning, not dependent on whether it "actually" hits the problem this time.

In the case of the example, though, I suppose you are right, as the file is a fixed item.
Though at least one example *does* need to be changed to work with the new behaviour.

A major driver of my attitude here is that we must make more effort to encourage users to take these warnings seriously and take time to eradicate them, to keep code up-to-date and avoid things breaking.
Version 2.0 is coming...

ajdawson:
I can see your side @pp-mo, time to formulate a compromise...

How about this: we change the terminology to express what you are really aiming for. I don't like the use of "deprecations" as a noun, it is potentially ambiguous. I take it to mean "deprecated features", you might mean something else, so let's be clear.

Instead of the comments:

    # Enable future options to avoid deprecations.
    iris.FUTURE.netcdf_promote = True

we use something like

    # Enable future option for netcdf loading to ensure we get the same results in future
    # versions of iris
    iris.FUTURE.netcdf_promote

I'm not at all set on the wording, but the key point is we say why we want to encourage this instead of  just what is does. How does that sound?

pp-mo:
> How does that sound?

Ok, I think that makes sense: I'll revisit  the various wordings

pp-mo:
I seem to be heading for rough waters here.
The kludge of filtering the warnings on text is just about ok for python2.7, 
but at 3.4 it seems to be opening a whole new can of worms.
Plus, the actual deprecation messages I'm trying to exclude are all very version-specific anyway.

I'm going to try making all _Iris_ deprecations distinguishable to avoid this problem.
I'm thinking of a distinctive warnings class : We really should have done that long ago anyway I think.

I have a prototype branch, and I'll see if I can get that to a PR tomorrow.
In the meantime, I'll close this until I've worked out whether that is the right solution...

pp-mo:
> rough waters ...  kludge of filtering the warnings on text ... prototype branch

I'm thinking #1993 is the right way forward, so I've rebased this onto that for now.
Also adjusted all the comments as requested ... if you have time.

I'll re-open this when we have the other sorted out, and this re-rebased if required.

ajdawson:
>so I've rebased this onto that for now ... I'll re-open this when we have the other sorted out, and this re-rebased if required.

Just FYI, and I've been burned by this before, once you rebase a branch that is the subject of a closed PR you cannot reopen it. It means you'll have to open a new PR and we'll lose this discussion. Worth remembering in future.

pp-mo:
> once you rebase a branch that is the subject of a closed PR you cannot reopen it

Rats, how right you are, but just too late !
Replaced by #1996

PR #1997, opened 2016-04-29
Cache linear interpolator

corinnebosley:
@rhattersley @pp-mo @pelson 
This is the first change that I implemented under PR #1946 which achieves a 25% increase in efficiency of nearest neighbour interpolation.

Richard suggested that I make this change in the original PR, but I would like to split up the issues brought up in that conversation and so I have made a fresh PR for this.

I would like to be involved in more changes to these routines (and I would like to do this soon, please don't let it all be forgotten about...) but I think this is a solid start.  Let's not run before we can walk, and all that.

### DIFF
--- a/lib/iris/analysis/trajectory.py
+++ b/lib/iris/analysis/trajectory.py
@@ -237,12 +237,17 @@ def interpolate(cube, sample_points, method=None):
     # Use a cache with _nearest_neighbour_indices_ndcoords()
     cache = {}
 
+    # Cache the linear interpolator
+    scheme = iris.analysis.Linear()
+    coords, points = zip(*sample_points)
+    interpolator = scheme.interpolator(cube, coords)
+
     for i in range(trajectory_size):
         point = [(coord, values[i]) for coord, values in sample_points]
 
         if method in ["linear", None]:
-            column = linear_regrid(cube, point)
-            new_cube.data[..., i] = column.data
+            column = interpolator([val[i] for _, val in sample_points])
+            new_cube.data[..., i] = column
         elif method == "nearest":
             column_index = _nearest_neighbour_indices_ndcoords(cube, point, cache=cache)
             column = cube[column_index]


lib/iris/analysis/trajectory.py
```diff
@@ -237,12 +237,17 @@ def interpolate(cube, sample_points, method=None):
     # Use a cache with _nearest_neighbour_indices_ndcoords()
     cache = {}
 
+    # Cache the linear interpolator
+    scheme = iris.analysis.Linear()
+    coords, points = zip(*sample_points)
+    interpolator = scheme.interpolator(cube, coords)
```
rhattersley: These lines need to be protected by `if method in ('linear', None)` - partly because it's wasted effort otherwise, but mostly because they will fail when trying to do "nearest-neighbour interpolation over multi-dimensional coordinates.

NB. The docstring for `iris.analysis.trajectory.interpolate` is a little ambiguous about the default value of `method`, but it turns out that the default is not just simply "linear", but switches to "nearest" when using multi-dimensional coordinates. :-1: I definitely don't propose to change that behaviour in _this_ PR, but it would be nice to clarify the docstring (again, possibly in a separate PR - small is beautiful).

lib/iris/analysis/trajectory.py
```diff
@@ -237,12 +237,17 @@ def interpolate(cube, sample_points, method=None):
...
+
     for i in range(trajectory_size):
         point = [(coord, values[i]) for coord, values in sample_points]
 
         if method in ["linear", None]:
-            column = linear_regrid(cube, point)
```
rhattersley: The corresponding import statement needs to be removed.

lib/iris/analysis/trajectory.py
```diff
@@ -237,12 +237,17 @@ def interpolate(cube, sample_points, method=None):
...
-            new_cube.data[..., i] = column.data
+            column = interpolator([val[i] for _, val in sample_points])
+            new_cube.data[..., i] = column
```
rhattersley: This still needs the `.data` suffix, i.e.:
```python
new_cube.data[..., i] = column.data
```

PR #1998, opened 2016-05-03
Structured load

bjlittle:
Extended the current experimental structured FF load capability to also load structured PP files.

It is now possible to pass a mixture of both structured FF and structured PP files to the structured loader.

Requires supporting iris-test-data:
- [x] https://github.com/SciTools/iris-test-data/pull/41


bjlittle:
@marqh Head's up :wink:

lib/iris/experimental/fieldsfile.py
```diff
@@ -20,12 +20,15 @@
 """
 
 from __future__ import (absolute_import, division, print_function)
+import os
```
QuLogic: This import doesn't belong here; it should be in a (new) section between here and the iris ones.

bjlittle: Yup, good spot, thanks @QuLogic

bjlittle:
@marqh Over to you ...

lib/iris/experimental/fieldsfile.py
```diff
@@ -40,15 +44,32 @@
 _HINT_COORDS = ['time', 'forecast_reference_time', 'model_level_number']
 _HINTS = {name: i for i, name in zip(range(len(_HINT_COORDS)), _HINT_COORDS)}
 
+_FF_SPEC_NAME = 'UM Fieldsfile'
+_PP_SPEC_NAME = 'UM Post Processing file'
+
```
bjlittle: I find this pretty clumsy ...

marqh:
looking good @bjlittle 

we will definitely want a "what's new" entry for this.  I would suggest a slightly enhanced entry, further stressing the notes and warnings described here:
http://scitools.org.uk/iris/docs/latest/iris/iris/experimental/fieldsfile.html

I also wonder whether the sentence 

"The results are equivalent to those from ..."
https://github.com/bjlittle/iris/blob/structured-load/lib/iris/experimental/fieldsfile.py#L97

could be usefully nuanced, to stress that we aim for equivalence, but small deviations from the expected structure can have significant effects and that exceptions will not be thrown in these cases.

I think this will be a helpful part of the message which we will need to communicate

marqh:
the two CML test results look identical to me, and I think they should be given how we made the files and teh code.

can we represent that by having the two tests reference the same 'expected' file?

bjlittle:
@marqh The CML are *almost* identical, the differing line is [simple_callback.cml#L6](https://github.com/bjlittle/iris/blob/structured-load/lib/iris/tests/results/experimental/fieldsfile/TestStructuredLoadPP/simple_callback.cml#L6) ... where the cube `attributes` has an entry resulting from the callback.

These new PP tests are based on the existing FF tests.

lib/iris/experimental/fieldsfile.py
marqh: i'm struggling to see a neater way at present

marqh:
i think this is pretty much good to go no @bjlittle 

a what's new and a squish and I am minded to merge

bjlittle:
@marqh Just penning a what's new entry ...

marqh:
this is go for launch

please  squish and commence countdown

marqh:
excellent, thank you @bjlittle

PR #2001, opened 2016-05-04
Small changes to ESMF api; add ESMPy as dependency, for Travis tests.

pp-mo:
Some ***very*** simple fixes to make our old "regrid_conservative_via_esmpy" routine work with ESMPy v7.0.0

For now, this is getting its ESMPy from scitools
See [conda-recipes-scitools#149](https://github.com/SciTools/conda-recipes-scitools/pull/149)


pp-mo:
NOTE: this routine could quite possibly benefit from a complete review now, as I doubt it shows best practice (and probably never did).
We should tackle that separately though, as this does the minimum business for now.
See #2002

PR #2008, opened 2016-05-10
Edit of aggregated_by docs

davidnmurray:
This is a proposed fix in respect of https://github.com/SciTools/iris/issues/1758

The behaviour of aggregated_by has been tested, and it doesn't affect the results whether the identical values are consecutive or otherwise. So, I propose to remove the "Todo..." line.

### DIFF
--- a/lib/iris/cube.py
+++ b/lib/iris/cube.py
@@ -3368,8 +3368,7 @@ bound=(1994-12-01 00:00:00, 1998-12-01 00:00:00)
 
         A "group coordinate" is a coordinate where repeating values represent a
         single group, such as a month coordinate on a daily time slice.
-        TODO: It is not clear if repeating values must be consecutive to form a
-        group.
+        Repeated values will form a group even if they are not consecutive.
 
         The group coordinates must all be over the same cube dimension. Each
         common value group identified over all the group-by coordinates is


rhattersley:
Thanks for testing this! :+1:

lib/iris/cube.py
```diff
@@ -3368,8 +3368,7 @@ def aggregated_by(self, coords, aggregator, **kwargs):
 
         A "group coordinate" is a coordinate where repeating values represent a
         single group, such as a month coordinate on a daily time slice.
-        TODO: It is not clear if repeating values must be consecutive to form a
-        group.
+        Repeated values can form a group even if they are not consecutive.
```
rhattersley: The word "can" suggests it doesn't always happen, perhaps "Repeated values will form a single group even ..."?

davidnmurray: Hi @rhattersley ! Many thanks for that, that's an excellent point. I'll make the appropriate change and put through an update very shortly...

davidnmurray:
... and after an unexpectedly-long trip down the version control rabbit-hole, here is the updated documentation.

rhattersley:
Thanks @davidnmurray. :+1: I hate to have to ask, but would you mind squashing your commits?

rhattersley:
> would you mind squashing your commits?

@SciTools/iris-devs - for simple cases such as this, perhaps we should consider enabling the [squash and merge](https://github.com/blog/2141-squash-your-commits) option?

davidnmurray:
Squashing is done.

rhattersley:
Super - thanks @davidnmurray :+1:

PR #2009, opened 2016-05-11
Remove 'experimental' from description in iris/fileformats/um/_fast_load_structured_fields.py

djkirkham:
This code is no longer experimental so this comment should be updated.

### DIFF
--- a/lib/iris/fileformats/um/_fast_load_structured_fields.py
+++ b/lib/iris/fileformats/um/_fast_load_structured_fields.py
@@ -14,7 +14,7 @@
 #
 # You should have received a copy of the GNU Lesser General Public License
 # along with Iris.  If not, see <http://www.gnu.org/licenses/>.
-"""Experimental code for fast loading of structured UM data."""
+"""Code for fast loading of structured UM data."""
 
 from __future__ import (absolute_import, division, print_function)
 from six.moves import (filter, input, map, range, zip)  # noqa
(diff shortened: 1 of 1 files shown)


rhattersley:
:+1: 

There's another occurrence of `experimental` in a docstring near the end of the file - would you be OK to update that one too?

djkirkham:
Urgh... now the line is too long...

djkirkham:
@rhattersley Can I do 
`:class:`
`` `~iris.fileformats.um._fast_load_structured_fields.FieldCollation` ``
to get it to fit within 80 characters?

ajdawson:
Can we set a default role for iris documentation of `:py:obj:` or even `:any:` so we don't have to specify `:class:`, `:meth:`, `:function:` etc.? Sphinx has supported this for a really long time and it makes documentation easier to write, and easier to read.

ajdawson:
See:
* http://www.sphinx-doc.org/en/stable/config.html#confval-default_role
* http://www.sphinx-doc.org/en/stable/markup/inline.html#inline-markup

dkillick:
> Can we set a default role for iris documentation

@ajdawson that's a great idea in principle but perhaps this PR in particular isn't the best place to try and introduce it? It could make this small PR to fix one specific documentation oversight into a much larger piece of work...

ajdawson:
I'm not suggesting it gets changed here. Just that it gets changed to make this kind of small fix unnecessary (or occur less often) in future.

dkillick:
> Just that it gets changed

:+1:

lib/iris/fileformats/um/_fast_load_structured_fields.py
```diff
@@ -240,7 +240,7 @@ def group_structured_fields(field_iterator):
 
     Returns:
         An generator of
```
rhattersley: An -> A

rhattersley:
> @rhattersley Can I do
> :class:
> `~iris.fileformats.um._fast_load_structured_fields.FieldCollation`
> to get it to fit within 80 characters?

It's in a private module so it won't end up in our Sphinx documentation. So how about ditching the Sphinx markup and just saying `A generator of FieldCollation objects...`

rhattersley:
Thanks @djkirkham :+1:

PR #2010, opened 2016-05-11
Simplify the deprecation of 'iris.fileformats.ff'.

pp-mo:
As @rhattersley pointed out yesterday, the previous effort is over-complicated in it's attempt to provide runtime deprecation warnings for everything in the interface.
We need only ensure that importing "iris.fileformats.ff" itself produces a warning :  The others are not needed, which makes it all much simpler.

It has proved awkward in several of these deprecation efforts to ensure the Sphinx documentation is still equivalent.
I think this still achieves that, but you may want to check it out.

### DIFF
--- a/lib/iris/fileformats/_ff.py
+++ b/lib/iris/fileformats/_ff.py
@@ -809,6 +809,11 @@ def load_cubes(filenames, callback, constraints=None):
         file (order is not preserved when there is a field with
         orography references).
 
+    .. deprecated:: 1.10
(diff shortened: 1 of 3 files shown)


rhattersley:
> It has proved awkward in several of these deprecation efforts to ensure the Sphinx documentation is still equivalent.
> I think this still achieves that, but you may want to check it out.

I've had a look and all seems well. :+1:

PR #2011, opened 2016-05-11
Reinstate grib as_messages.

pp-mo:
This function was removed by #1941, when it should just have been deprecated.
The test ```iris.tests.unit.fileformats.grib``` test "test_as_messages" is not reinstated, as "test_as_pairs" covers it.
As a complete aside, I also spotted the ECMWF link in the module header was broken + fixed that.

### DIFF
--- a/lib/iris/fileformats/grib/__init__.py
+++ b/lib/iris/fileformats/grib/__init__.py
@@ -17,7 +17,7 @@
 """
 Conversion of cubes to/from GRIB.
 
-See also: `ECMWF GRIB API <http://www.ecmwf.int/publications/manuals/grib_api/index.html>`_.
+See also: `ECMWF GRIB API <https://software.ecmwf.int/wiki/display/GRIB/Home>`_.
 
 """
 
@@ -50,7 +50,7 @@ import iris.fileformats.grib.load_rules
 
 __all__ = ['load_cubes', 'save_grib2', 'load_pairs_from_fields',
            'save_pairs_from_cube', 'save_messages', 'GribWrapper',
-           'as_pairs', 'grib_generator', 'reset_load_rules',
+           'as_messages', 'as_pairs', 'grib_generator', 'reset_load_rules',
            'hindcast_workaround']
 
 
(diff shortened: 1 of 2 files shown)


pp-mo:
Given this, I think we again have a consistent form of grib support.
This can serve as a sensible basis for subbing out grib support to "iris_grib" [as proposed here](https://github.com/SciTools/iris/issues/1948#issuecomment-218139311)

Remaining PR not included :
#1755 "grib1 future" is **not** in included in this.
It seeks to provides Grib1 support through the new "strict" loader , but we aren't finished agreeing it.
It should be pretty easy to carry those changes over into the new iris-grib package when required.
The point is, hopefully, that by pushing those problems into iris-grib, we *don't need to* provide alternatives to grib1 loading before cutting Iris 1.10.
Instead we propose: 
 * iris 1.10 deprecates iris.fileformats.grib in favour of iris-grib
 * iris 2.0 removes iris.fileformats.grib
 * grib-api 1.0 retains both 'old' + 'new' grib loaders
 * non-strict support can be removed at grib-api 2.0

pp-mo:
> Remaining PR not included :

There are also a couple of older ones still in the pot, but I think no urgency on these ? ...
 * #1703 @ajdawson "fault tolerant Grib loading."
 * #1687 "Add GRIB2 load support for GDT-50 (spherical harmonics)."

rhattersley:
> This can serve as a sensible basis for subbing out grib support to "iris_grib" as proposed here

NB. Irrespective of whether we go the iris-grib route we want this change.

rhattersley:
> The test iris.tests.unit.fileformats.grib test "test_as_messages" is not reinstated

I don't understand why you've not reinstated this - by default I'd rather it was back again.

rhattersley:
With the exception of the `test_as_messages` question I'm :+1:

pp-mo:
> I don't understand why you've not reinstated this

Ok, you're probably right.

rhattersley:
Super, thanks. :+1: Any chance of a squash?

pp-mo:
sqwished !

rhattersley:
Excellent - that gets grib handling back into the state where it doesn't need a major version change. :+1:

PR #2018, opened 2016-05-13
Fix warn_deprecation to display where it was called from.

pp-mo:
Otherwise all deprecation warnings appear to come from a line in ```iris._deprecation.py```

Turns something like:
```...lib/iris/_deprecation.py:36: IrisDeprecation: the`auto_regularise` kwarg is deprecated...```
into:
```...lib/iris/fileformats/grib/__init__.py:945: IrisDeprecation: the`auto_regularise` kwarg is deprecated ...```

### DIFF
--- a/lib/iris/_deprecation.py
+++ b/lib/iris/_deprecation.py
@@ -32,8 +32,34 @@ class IrisDeprecation(UserWarning):
     pass
 
 
-def warn_deprecated(msg, **kwargs):
-    warnings.warn(msg, IrisDeprecation, **kwargs)
+def warn_deprecated(msg, stacklevel=2):
+    """
+    Issue an Iris deprecation warning.
+
+    Calls :func:`warnings.warn', to emit the message 'msg' as a
+    :class:`warnings.warning`, of the subclass :class:`IrisDeprecationWarning`.
+
+    The 'stacklevel' keyword is passed through to warnings.warn.  However by
+    default this is set to 2, which ensures that the identified code line is in
+    the caller, rather than in this routine.
+    See :mod:`warnings` module documentation.
+
+    For example::
+
+        >>> from iris._deprecation import warn_deprecated
+        >>> def arrgh():
+        ...    warn_deprecated('"arrgh" is deprecated since version 3.5.')
+        ...    return 1
+        ...
+        >>> arrgh()
+        __main__:2: IrisDeprecation: "arrgh" is deprecated since version 3.5.
+        1
+        >>> arrgh()
+        1
+        >>>
+
+    """
+    warnings.warn(msg, IrisDeprecation, stacklevel=stacklevel)
 
 
 # A Mixin for a wrapper class that copies the docstring of the wrapped class

### CODE REVIEW

File: lib/iris/_deprecation.py [hunk 1]
```diff
@@ -33,6 +33,8 @@ class IrisDeprecation(UserWarning):
 
 
 def warn_deprecated(msg, **kwargs):
+    stacklevel = kwargs.get('stacklevel', 0) + 2
```
rhattersley: A short docstring explaining that `stacklevel` is now relative to the calling frame would be the icing on the cake. Perhaps with a simple example?
```python
def warn_deprecated(msg, **kwargs):
    """
    $ cat << EOF > example.py

    def foo():
        warn_deprecated('Danger!')
        warn_deprecated('Awooga!', stacklevel=1)

    def bar():
        warn_deprecated('Evacuate!')
        foo()

    if __name__ == '__main__':
        bar()
    EOF
    $ python example.py
    example.py:8: IrisDeprecation: Evacuate!
      warn_deprecated('Evacuate!')
    example.py:4: IrisDeprecation: Danger!
      warn_deprecated('Danger!')
    example.py:9: IrisDeprecation: Awooga!
      foo()
    """
```

rhattersley:
Other than the docstring ... definitely :+1: :grin:

pp-mo:
I only supported an existing keyword as it was so easy.
But looking again, I have thought now to simplify it too.
Rather than give a full explanation I deferred to the existing documentation.

lib/iris/_deprecation.py
```diff
@@ -32,8 +32,34 @@ class IrisDeprecation(UserWarning):
...
+    See :mod:`warnings` module documentation.
+
+    For example::
+
+        >>> from iris._deprecation import warn_deprecated
+        >>> def arrgh():
+        ...    warn_deprecated('"arrgh" is deprecated since version 3.5.')
+        ...    return 1
+        ...
+        >>> print arrgh()
```
QuLogic: `print()`, not `print`.

pp-mo: I've used command-line response instead.
I find the P3-style print function **much** less readable, especially in a command-line context, and as pep8 doesn't allow me to add spaces inside it.

ajdawson: I thought we were trying to maintain Python 3 compatibility everywhere, including doctests/documentation. `print(arrgh())` vs `print arrgh()`, doesn't look that bad to me, the function name probably is more to blame than the parentheses!

pp-mo: > I thought we were trying to maintain Python 3 compatibility everywhere

We are.
It's just that I happen to hate the look of this one thing !
It may well be "the function name is more to blame", but I still find it a backward step overall in terms of readability.

QuLogic: I'm not really sure that's a convincing argument, but in this case, why do you need the `print` at all? At the interactive prompt, it'll print the same thing out either way.

pp-mo: > in this case, why do you need the print at all?

That's what the latest commit does.

QuLogic: No, it's good; I just did not notice it for some reason.

marqh:
i'm minded to merge

any objections please?

@pp-mo please squish

pp-mo:
Skwished
Thanks @marqh

Issue #2021, opened 2016-05-17
BUG: Unexpected concatenate failure


cpelley:
```Python
>>> import iris
>>> import iris.tests.stock
>>> cube = iris.tests.stock.lat_lon_cube()[:, 1:]
>>> cube.coord(axis='x').points = [360., 350., 10.]
>>> iris.cube.CubeList([cube[:, 2:], cube[:, :2]]).concatenate_cube()
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File ".../git/iris/lib/iris/cube.py", line 487, in concatenate_cube
    res = iris._concatenate.concatenate(self, error_on_mismatch=True)
  File ".../git/iris/lib/iris/_concatenate.py", line 291, in concatenate
    concatenated_cubes.append(proto_cube.concatenate())
  File ".../git/iris/lib/iris/_concatenate.py", line 653, in concatenate
    dim_coords_and_dims = self._build_dim_coordinates()
  File ".../git/iris/lib/iris/_concatenate.py", line 851, in _build_dim_coordinates
    dim_coord = iris.coords.DimCoord(points, bounds=bounds, **kwargs)
  File ".../git/iris/lib/iris/coords.py", line 1390, in __init__
    coord_system=coord_system)
  File ".../git/iris/lib/iris/coords.py", line 445, in __init__
    self.points = points
  File ".../cpelley/git/iris/lib/iris/coords.py", line 1464, in points
    raise ValueError('The points array must be strictly monotonic.')
ValueError: The points array must be strictly monotonic.
>>> iris.cube.CubeList([cube[:, :2], cube[:, 2:]]).concatenate_cube()
<iris 'Cube' of unknown / (unknown) (latitude: 3; longitude: 3)>
```

As you can see, concatenation succeeds when the cubelist has a different order.

ajdawson:
Is this so unexpected? With the first order the longitudes are [[10], [360, 350]] which do not join (in the order specified) to form a monotonic coordinate. In the second case the longitudes are [[360, 350], [10]] which form a monotonically decreasing sequence.

cpelley:
> Is this so unexpected? With the first order the longitudes are [[10], [360, 350]] which do not join (in the order specified) to form a monotonic coordinate.

Yes.  That is not how concatenate is documented to behave and this behaviour is not consistent in any case.  See:

```Python
>>> cube = iris.tests.stock.lat_lon_cube()       
[...]

PR #2022, opened 2016-05-18
Fix concatenation with singleton dimensions

ajdawson:
This PR fixes a bug reported in #2021. The problem arises when we concatenate a list of cubes where the first one has a singleton dimension and one of the others doesn't. The dimension ordering determined by the `_ProtoCube` would be based only on the first cube, and be set to `_CONSTANT`. This later implies sorting coordinate points in ascending order, which is only correct if the non-singleton dimension is ascending.

This PR addresses the issue by ensuring a `_ProtoCube` with a `_CONSTANT` dimension order checks the dimension order of subsequently registered cubes and chooses the first non-constant one it finds for the dimension order. This ultimately ensures that the sorting is done in the correct order and the cubes will concatenate as expected.

### DIFF
--- a/lib/iris/_concatenate.py
+++ b/lib/iris/_concatenate.py
@@ -742,6 +742,17 @@ class _ProtoCube(object):
             # Declare the nominated axis of concatenation.
             self._axis = candidate_axis
 
+        if match:
+            # If the protocube dimension order is constant (indicating it was
+            # created from a cube with a length 1 dimension coordinate) but
+            # a subsequently registered cube has a non-constant dimension
+            # order we should use that instead of _CONSTANT to make sure all
+            # the ordering checks and sorts work as expected.
+            existing_order = self._coord_signature.dim_order[self.axis]
+            this_order = coord_signature.dim_order[self.axis]
+            if existing_order == _CONSTANT and this_order != _CONSTANT:
+                self._coord_signature.dim_order[self.axis] = this_order
+
         return match
 
     def _add_skeleton(self, coord_signature, data):
--- a/lib/iris/tests/unit/concatenate/test_concatenate.py
+++ b/lib/iris/tests/unit/concatenate/test_concatenate.py
@@ -194,6 +194,30 @@ class TestOrder(tests.IrisTest):
         result = concatenate([top, bottom])
         self.assertEqual(len(result), 1)
 
+    def test_asc_points_with_singleton_ordered(self):
+        top = self._make_cube([5])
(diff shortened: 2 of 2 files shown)


ajdawson:
Ping @cpelley

cpelley:
Thanks @ajdawson taking such an interest in this.
I don't see a test for the case added for where they are both cubes with length 1 dimension coordinates.  I'm guessing that this works and that we always end up with increasing coordinates on the resulting concatenated cube, however it would be good if there was a test (if there isn't one).

Cheers

ajdawson:
This always worked previously and yes, the ordering is determined by reversing if dimension ordering is `_DECREASING`, so if all were `_CONSTANT` you would get an increasing ordering (which usually makes the most sense).

cpelley:
> This always worked previously and yes..

Yes, should we not add a test though to make sure this remains the case?

Apart from that, PR looks fine to go in!

Thanks so much for this @ajdawson

ajdawson:
I'm working on some better test coverage. Thanks for pointing it out.

ajdawson:
I have updated with some additional tests. These cover the case of all singleton dimensions, and additionally cases where bounds are present (although the bounds don't make any difference to the order, these do capture known good behaviour so are probably worth keeping).

cpelley:
+1 very thorough testing thanks @ajdawson 
Looks good to go.

lib/iris/_concatenate.py
```diff
@@ -742,6 +742,17 @@ def register(self, cube, axis=None, error_on_mismatch=False,
             # Declare the nominated axis of concatenation.
             self._axis = candidate_axis
 
+        if match:
+            # If the protocube dimension order is constant (indicating it was
```
marqh: sorry to be picky, but please may you trim these lines to pep8 (80 characters) they seem a bit long

marqh:
@ajdawson tests looks great, the implementation also looks great

please trim line length and squish, then i'm minded to merge

lib/iris/_concatenate.py
cpelley: Line length looks fine to me, longest here is 78 characters?

ajdawson: That's what I thought too... This file is checked by the coding standards tests too I think.

cpelley:
ping @marqh

ajdawson:
@marqh can you please reconsider this as is? We can't see the issue with line length you describe...

marqh:
pong

finally ..... :$

PR #2023, opened 2016-05-20
Update CF tables: std names v33, area types v4

rhattersley:


marqh:
looks good, thanks @rhattersley 
I am glad we are updating these now

rhattersley:
Thanks @marqh

PR #2027, opened 2016-05-23
Relax DimCoord bound monotonicity

rhattersley:
In the light of a [recent comment](http://mailman.cgd.ucar.edu/pipermail/cf-metadata/2016/058820.html) perhaps we should relax the monotonicity constraint on DimCoord.bounds?


ajdawson:
This seems sensible. Is it worth checking if the corresponding point actually lies within (or on one of) the bounds? Seems like this would be a requirement.

rhattersley:
The check was simple enough ... but look at all those exciting knock-on issues! :grin: I'll dig deeper...

ajdawson:
First off, I'm reassured that you find these issues exciting! It looks like most of them are from a few stock generators (stock cube generators? Mmmm tasty).

This brings about the question: do we want to do this now since it does not appear to be backwards compatible (i.e. existing code may break because of it)?

It also raises a slightly deeper question: do the points need to lie inside the bounds, or do the bounds need to bound the point, or both? What I'm getting at is that it would be equally valid to have this check when setting the points too... Are bounds second class to points or should both be considered equals?

With the behaviour as it stands we can't change the bounds to a set that don't contain the points (seems reasonable) but we can change the points so they lie outside the bounds... If we checked when setting points too it would mean one could not change the points to be outside the bounds without first resetting the bounds to `None`. This may prove annoying to someone who wants to change the points then the bounds.

I'm not sure what the use cases are here. It is tempting to go for consistency and check that points lie within bounds when setting either points or bounds (and perhaps if setting points, remove the bounds if they aren't compatible?). On the other hand, is it just a load of effort for no real gain, and we just end up annoying people who know what they are doing?

lib/iris/tests/unit/fileformats/pp_rules/test__convert_vertical_coords.py
```diff
@@ -188,7 +188,7 @@ def test_unbounded__vector_no_depth(self):
                           brlev=brlev, expect_mixed=True, dim=0)
 
     def test_bounded(self):
-        self._check_depth(_lbcode(1), lblev=23.0, brlev=22.5, brsvd1=23.5,
+        self._check_depth(_lbcode(1), blev=23.0, brlev=22.5, brsvd1=23.5,
```
rhattersley: :bug: ahoy!

rhattersley:
> it would be equally valid to have this check when setting the points too

If we're going to have checks at all, then :+1: 

> This may prove annoying to someone who wants to change the points then the bounds.

If they want to change both then perhaps `coord.copy(points=..., bounds=...)` would be more appropriate?

> On the other hand, is it just a load of effort for no real gain, and we just end up annoying people who know what they are doing?

The test errors are just highlighting places where the test code was creating silly coordinates with the points outside the bounds. Mostly because the test didn't care one way or the other. But in one case it found a [genuine bug](https://github.com/SciTools/iris/pull/2027/files#r64550847) ... so perhaps +1 for checks?

The only exceptions to the silliness issues are a couple of tests that are explicitly checking for the old non-overlapping bounds behaviour.

rhattersley:
Grrr :angry: ... "The Curse of the :copyright:" strikes again! Fixes inbound...

rhattersley:
>> it would be equally valid to have this check when setting the points too
> 
> If we're going to have checks at all, then :+1:

This brings back the idea of having immutable bounds values. :unamused:

rhattersley:
> This brings back the idea of having immutable bounds values. :unamused:

I'm ~~in danger of~~ opening up a whole can of worms, but here's my train of thought:

1. It'd be good to have some sort of guiding philosophy/strategy for this sort of thing. 
2. The flexible-load/strict-save idea (if that's what we want) needs to be part of that strategy.

One outline strategy is:
* Allow non-compliant values in our data model. (Helpful when "compliant" is ill-defined!)
* Error/warn (depending on how strict we want to be) about non-compliant values on save.
* Provide compliance checking functions/methods.

=> Don't bother checking point-bound inclusion until saving. Don't even bother checking for monotonicity until saving (so we don't even need to make the points or bounds read-only). Perhaps don't even bother to discriminate between DimCoord & AuxCoord?!

Not clear if we would still have caught the lblev vs. blev issue.

Our current strategy is more like:
* Only allow compliant values in our data model.
* Provide sufficient flexibility in our load process for a user to intervene where necessary to ensure compliance - otherwise some datasets won't load!

=> Ensure it is impossible to set non-compliant values. => Read-only points & bounds. Check compliance when setting points and/or bounds.

Ping @SciTools/iris-devs

corinnebosley:
On the basis that this is over 2 years old, stalled for over a year and introduces more complications than benefits (by the looks of it), I'm shutting this one down.  Sorry @rhattersley.

rhattersley:
No apology necessary 😁 👍

PR #2029, opened 2016-05-24
Added missing `tests.skip_data` decorators

dkillick:
A couple of integration test classes seem to have missed their skip_data decorators, so I've added them.

I also slightly refactored one of the tests to add a `setUp` method and pull common setup code into it.


rhattersley:
Thanks @dkillick :+1:

PR #2030, opened 2016-05-24
Fix as_compatible_shape for transposed auxcoords

rcomer:
A minor fix for `as_compatible_shape`.

Sometimes Cube dimensions are transposed by the operation.  If an AuxCoord spans these dimensions, we need to either transpose the AuxCoord's points or rearrange its associated dims.  I went for the latter, making it consistent with the `cube.transpose()` treatment.


marqh:
this seems like a sensible addition
many thanks @rcomer

PR #2033, opened 2016-05-25
Add support for loading stereographic grid mappings from netCDF files

batsc:
Add support for loading stereographic grid mappings from netCDF files where the scale factor at projection origin is 1

### DIFF
--- a/lib/iris/tests/unit/fileformats/netcdf/test_Saver.py
+++ b/lib/iris/tests/unit/fileformats/netcdf/test_Saver.py
@@ -29,7 +29,7 @@ import numpy as np
 
 import iris
 from iris.coord_systems import (GeogCS, TransverseMercator, RotatedGeogCS,
-                                Mercator)
+                                Mercator, Stereographic)
 from iris.coords import DimCoord
 from iris.cube import Cube
 from iris.fileformats.netcdf import Saver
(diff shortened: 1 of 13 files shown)


dkillick:
Other than the failing test (again caused by the changes to the test data repo not being merged) this looks pretty good! Seems a lot of lessons learnt in #2026 have been applied here :+1:

lib/iris/tests/unit/fileformats/netcdf/test_Saver.py
```diff
@@ -51,6 +51,7 @@ def _transverse_mercator_cube(self, ellipsoid=None):
         cube.add_dim_coord(coord, 1)
         return cube
 
+
```
dkillick: Whoops! I think this line is tripping up the code compatibility check. Remove it and all the tests should pass :smile:

lib/iris/tests/unit/fileformats/pyke_rules/compiled_krb/fc_rules_cf_fc/test_build_stereographic_coordinate_system.py
```diff
@@ -0,0 +1,88 @@
...
+"""
+
+from __future__ import (absolute_import, division, print_function)
+from six.moves import (filter, input, map, range, zip)  # noqa
+
+# import iris tests first so that some things can be initialised before
+# importing anything else
+import iris.tests as tests
+
+import numpy as np
```
dkillick: This import is unused (and it's the same in the similar Mercator test). I don't want the hassle of removing it in this PR, so we'll sort it in a follow-up PR.

dkillick:
Argh, more process timout exceptions :hourglass_flowing_sand: :angry: 

But these timeouts have nothing to do with the code changes introduced here, so I'm going to get merge-happy anyway!

dkillick:
I like the fact that the approach for #2026 could be replicated here. Good work @djkirkham and @batsc for getting this in (and welcome @batsc in your first PR :+1:)

PR #2035, opened 2016-05-26
Add warning about HDF5 thread-safety.

rhattersley:
A user has reported a crash (which I've also replicated) when combining lazy evaluation and a non-thread-safe build of HDF5.

To re-create:
```python
import iris
path = '/path/to/a/tens_of_MB_cube.nc'
cubes = [iris.load_cube(path) for i in range(3)]
total = sum(cubes)
result = total.data  # This line triggers the lazy evaluation and the subsequent seg fault
```


marqh:
thanks @rhattersley

Issue #1167, opened 2014-06-09
Unwise use of setattr prevents saving a cube to file


stephenpascoe:
Iris NetCDF saving uses setattr to copy attributes to the NetCDF4-python object.  This causes problems if your source cube has an attribute which clash with  NetCDF4-python method names.  E.g. the ECHAM5 Aerocom simulations have contain the "dimensions" variable attribute which prevents saving a cube derived from this data.

```
# Executed from /group_workspaces/jasmin/cis/AeroCom/A2/MPIHAM_V2_KZ.A2.CTRL/download on JASMIN

import iris
cube = iris.load_cube('aerocom.MPIHAM_V2_KZ.AEROCOM_A2.CTRL.monthly.loadbc.2007.nc')
iris.save(cube, '/tmp/tmp.nc')
```

Causes the following stack trace in Iris 1.6.1:
```
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "/usr/lib/python2.7/site-packages/iris/io/__init__.py", line 341, in save
    saver(source, target, **kwargs)
  File "/usr/lib/python2.7/site-packages/iris/fileformats/netcdf.py", line 1568, in save
    least_significant_digit)
  File "/usr/lib/python2.7/site-packages/iris/fileformats/netcdf.py", line 671, in write
    self.update_global_attributes(global_attributes)
  File "/usr/lib/python2.7/site-packages/iris/fileformats/netcdf.py", line 695, in update_global_attributes
    setattr(self._dataset, attr_name, attributes[attr_name])
  File "netCDF4.pyx", line 1832, in netCDF4.Dataset.__setattr__ (netCDF4.c:23365)
AttributeError: 'dimensions' is one of the reserved attributes ('_grpid', '_grp', '_varid', 'groups', 'dimensions', 'variables', 'dtype', 'file_format', '_nunlimdim', 'path', 'parent', 'ndim', 'maskandscale', 'cmptypes', 'vltypes', '_isprimitive', '_isvlen', '_iscompound', '_cmptype', '_vltype'), cannot rebind. Use setncattr instead.
```
I am working with Iris 1.6.1 but I have confirmed the code still uses setattr.

The setncattr() method of NetCDF4-python has been provided to avoid this name clash.

ajdawson:
Closed by #1326.

PR #2036, opened 2016-05-31
Fix for setting protected attributes on NetCDF variable objects

duncanwp:
I hit an issue trying to save cubes with variable attributes including 'name', which is related (but not the same as #1167) when using netcdf4>=1.2. This is a test which reproduces the problem and an associated fix.

### DIFF
--- a/lib/iris/tests/unit/fileformats/netcdf/test_save.py
+++ b/lib/iris/tests/unit/fileformats/netcdf/test_save.py
@@ -1,4 +1,4 @@
-# (C) British Crown Copyright 2014 - 2015, Met Office
+# (C) British Crown Copyright 2014 - 2016, Met Office
 #
 # This file is part of Iris.
 #
@@ -58,6 +58,24 @@ class Test_attributes(tests.IrisTest):
             res = ds.getncattr('bar')
             ds.close()
         self.assertArrayEqual(res, np.arange(2))
+
+    def test_no_special_attribute_clash(self):
+        # Ensure that saving multiple cubes with netCDF4 protected attributes
+        # works as expected.
+        # Note that here we are testing variable attribute clashes only - by
+        # saving multiple cubes the attributes are saved as variable
+        # attributes rather than global attributes.
+        c1 = Cube([0], var_name='test', attributes={'name': 'bar'})
+        c2 = Cube([0], var_name='test_1', attributes={'name': 'bar_1'})
+
+        with self.temp_filename('foo.nc') as nc_out:
+            save([c1, c2], nc_out)
+            ds = nc.Dataset(nc_out)
+            res = ds.variables['test'].getncattr('name')
+            res_1 = ds.variables['test_1'].getncattr('name')
+            ds.close()
+        self.assertEqual(res, 'bar')
+        self.assertEqual(res_1, 'bar_1')
 
 
 class Test_unlimited_dims(tests.IrisTest):
--- a/lib/iris/fileformats/netcdf.py
+++ b/lib/iris/fileformats/netcdf.py
@@ -1298,7 +1298,7 @@ class Saver(object):
 
             # Don't clobber existing attributes.
             if not hasattr(cf_var, name):
-                setattr(cf_var, name, value)
+                cf_var.setncattr(name, value)
 
         return cf_name
 
@@ -1416,7 +1416,7 @@ class Saver(object):
 
             # Don't clobber existing attributes.
             if not hasattr(cf_var, name):
-                setattr(cf_var, name, value)
+                cf_var.setncattr(name, value)
 
         return cf_name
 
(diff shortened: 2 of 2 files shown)


lib/iris/tests/unit/fileformats/netcdf/test_save.py
```diff
@@ -59,6 +59,17 @@ def test_attributes_arrays(self):
             ds.close()
         self.assertArrayEqual(res, np.arange(2))
 
+    def test_no_special_attribute_clash(self):
+        c1 = Cube([0], var_name='test', attributes={'name': 'bar'})
+        c2 = Cube([0], var_name='test_1', attributes={'name': 'bar_1'})
```
ajdawson: What is `c2` for? you save it to the file but you don't read it back.

ajdawson:
Seems reasonable enough. If you want to make a head start on the [CLA](http://scitools.org.uk/documents/cla_form.pdf) it may avoid delays later.

There is a test fail on coding standards, you just need to change the year in the copyright statement of the test file from 2015 to 2016. The other fails look to be timeouts from the always irritating regrid tests.

lib/iris/tests/unit/fileformats/netcdf/test_save.py
duncanwp: I need it to reproduce the problem - I've not looked at the NetCDF code in detail, but I think it needs to be trying to save a CubeList to hit the setattr lines. Presumably the attributes get set as global if it's saving one Cube, but variable attributes if it's saving many?

Happy to add a comment to that effect if you think it needs it.

ajdawson: That sounds about right actually. Perhaps a comment to that effect would be of value. I guess you could also check the second cube's attribute too for completeness, I'll leave that up to you.

duncanwp:
I've submitted the CLA just waiting for a response now :-)

lib/iris/tests/unit/fileformats/netcdf/test_save.py
```diff
@@ -59,6 +59,22 @@ def test_attributes_arrays(self):
             ds.close()
         self.assertArrayEqual(res, np.arange(2))
 
+    def test_no_special_attribute_clash(self):
+        # Ensure that saving multiple cubes with netCDF4 protected attributes works as expected.
+        # Note that here we are testing variable attribute clashes only - by saving multiple 
+        # cubes the attributes are saved as variable attributes rather than glocal attributes.
```
ajdawson: Typo: glocal -> global. The line length looks too long also, it should be <80 characters. Also some trailing whitespace that needs to go ([the Travis CI test result](https://travis-ci.org/SciTools/iris/jobs/134382366) for coding standards should list lines and columns for these).

lib/iris/tests/unit/fileformats/netcdf/test_save.py
```diff
@@ -1,4 +1,4 @@
-# (C) British Crown Copyright 2014 - 2015, Met Office
+# (C) British Crown Copyright 2016, Met Office
```
ajdawson: This should be `2014 - 2016`.

duncanwp:
Thanks, that's what I get for editing it in GitHub!

ajdawson:
I've never dared try that option, it looks super convenient for small changes.

ajdawson:
Thanks @duncanwp, can you squash this into a single commit please? After that I'm happy to merge once the CLA is confirmed by the MO.

duncanwp:
Great, I'm pretty sure you can squash-on-merge in Github though? I would do it locally but I'm not on my work machine and don't have it checked out locally (hence editing in Github!).

ajdawson:
Squash on merge is disabled for this repository.

ajdawson:
Just noticed this work is done on your master branch. Not a problem here, but in future it may be worth working on a [feature branch](http://scitools.org.uk/iris/docs/latest/developers_guide/gitwash/development_workflow.html#make-a-new-feature-branch), particularly if you have a few things on the go at any one time.

duncanwp:
Yeah, I should get into the habbit of that - I didn't think it was really necassary for this though. 

Why is squashing on merge disabled out of interest?

ajdawson:
Not sure, @rhattersley will be able to provide the answer to your question.

marqh:
CLA approval 
https://github.com/SciTools/scitools.org.uk/pull/131

marqh:
> Why is squashing on merge disabled out of interest?

i think this has not been turned on on this repository so far.  I don't have a good reason why not and I would support implementing this feature
@bjlittle @pelson 
would you agree?
do you think this would reduce the PR noise?

ajdawson:
Would anyone more familiar with the netcdf backend like to take a look at this before merge? @pp-mo perhaps?

marqh:
i'm having a look into this, i'll try to report back tomorrow on my conclusions

pp-mo:
Thanks for the mention @ajdawson !
It makes sense to me.

marqh:
i agree, this looks like a sensible change

many thanks @duncanwp

PR #2038, opened 2016-06-02
Add name iii version2 fileformat

SusanJL:
We now have a new version of NAME output called NAME III version 2. The changes to name_loaders.py allows these files to be read in without breaking the ability to read the older NAME formats. The reader also allows for the reading of files with different sets of coordinates - something which was not possible with the older NAME formats. I have added the code to my branch and tested my branch on new and old format NAME files. It could probably be made more efficient and I probably need to add a test file but need help with both of these. I've attached and example file.

[lidar_grid1_C1.txt](https://github.com/SciTools/iris/files/295921/lidar_grid1_C1.txt)


lbdreyer:
Whenever you create a PR or push up new commits to a branch that is an open PR, this automatically triggers travis to run all the tests with a built copy of iris that includes your changes. The :x: indicates that there are some test failures. If you click on the :x:  or the 'Details' link it will take to a list of the build jobs. If you click on any that have failed (i.e. any that are red) and scroll right to bottom of the page, you will see what tests have failed.

* **'CML do not match' test failures -** There are expected cml results that we test against (located [here](https://github.com/SciTools/iris/tree/master/lib/iris/tests/results/name)). You will have to update these expected results cml files too.

* **Pep8 conformance failure -** It looks like you have a few lines that are longer than 79 characters that you will need to shorten in order to conform with pep8. You can check a file conforms with pep8 by running 
`$ pep8 file.py`
This command returns all the pep8 conformance failures. If it doesn't return anything, then that file conforms!

SusanJL:
Thank you.
 I'm afraid I have some very basic questions.

1. I guess the CML results are different because I have removed some unnecessary attributes from the cubes. How do I go about updating the results cml files.
2. I've found the line which is too long. Do I just need to correct that line and push to the add_nameIII_version2_fileformat branch on github or do I need to resubmit the pull request.

ajdawson:
>I guess the CML results are different because I have removed some unnecessary attributes from the cubes. How do I go about updating the results cml files.

You can just edit the CML file by hand to account for any differences this code change introduces. While you are thinking about testing, it would be nice to have a test for the new functionality.

>I've found the line which is too long. Do I just need to correct that line and push to the add_nameIII_version2_fileformat branch on github or do I need to resubmit the pull request.

Just push a new commit to this branch and the PR will update, no need to make a new one.

SusanJL:
Thank you. I've corrected the various errors and fails flagged by the Travis testing.

As this code as the ability to read a new NAME output format I don't think I can add any tests of the new functionality until I have data in a location where it can be seen by Travis. I've submitted a pull request to iris-test-data (https://github.com/SciTools/iris-test-data/pull/44) to initiate this process.

marqh:
> As this code as the ability to read a new NAME output format I don't think I can add any tests of the new functionality until I have data in a location where it can be seen by Travis. I've submitted a pull request to iris-test-data (SciTools/iris-test-data#44) to initiate this process.

this is now merged

to use it on travis, you will need to update:
https://github.com/SciTools/iris/blob/master/.travis.yml#L25
to use the latest commit sha1 from :
https://github.com/SciTools/iris-test-data/commit/e8e62ab79a2f8789f01f8dfa9829915984a39b2a
`e8e62ab79a2f8789f01f8dfa9829915984a39b2a`

SusanJL:
Is there a guide to running the tests offline?

lbdreyer:
@SusanJL I am not aware of any documentation for running the tests but I can hopefully point you in the direction of what to do if you want to run multiple tests (e.g. all the coding standard tests, or even all the tests).

To run the tests,  you can make use of the test runner, i.e.
```
$ python -m iris.tests.runner
```

We give travis instructions on what to test in the [travis.yml file ](https://github.com/SciTools/iris/blob/master/.travis.yml). There are 5 different types of testing (default, default with minimal package dependency requirements, coding, example and docs). We test each of these against python 2.7 and python 3.4, hence the total of 10 different tests for each [build](https://travis-ci.org/SciTools/iris/builds/136428950).

The instructions for each type of test can be found in the travis.yml file [here](https://github.com/SciTools/iris/blob/master/.travis.yml#L106).
So, for example, if you want to run the “default” tests on your desktop, you would run
```
$ python -m iris.tests.runner --default-tests --system-tests --print-failed-images
```

As for the environment you run the tests in, I think you should be okay. I think all the dependencies should be available to you on the desktop. At the start of the travis.yml file you will see that it creates a conda environment then installs all the necessary dependencies (e.g. matplotlib) in there, so if you are having problems you could always try following those instructions.

Another thing you could do (if you wanted to check all the tests pass before you put your changes up as a PR) would be to add travis testing to *your* iris repo. This will start a travis build everytime you push a branch to your iris repo. To do this you will need a travis account (I believe you can just log in with github). Then, if you click on your travis profile (click on your username in the top right hand corner) you should see a list of your repositories. If you flick the switch for your Iris repo, it will then start a travis build next time you push to your iris repo.

marqh:
@SusanJL @lbdreyer 
all the test are passing and the code is looking good; are there any further barriers to adopting this code, from either of your perspectives?

thank you
mark

SusanJL:
@marqh 
I'm trying to add additional tests (to test the load of the new data) as recommended by @ajdawson but am currently struggling to work out how to test files from iris-test-data offline. The test code is skipping tests which require external data.

SusanJL:
Additional tests now added. One test checks that the modified format of NAME data files can be correctly loaded and the other checks that heights provided in scientific notation (e.g. 10E+00) can be correctly read as floats.

marqh:
this all looks good to me

thank you @SusanJL

PR #2039, opened 2016-06-03
Exposecm

marqh:
a simple rebase of 
#1981

### DIFF
--- a/docs/iris/src/whatsnew/contributions_1.10/newfeature_2016-Jun-03_cell_method_parsing.txt
+++ b/docs/iris/src/whatsnew/contributions_1.10/newfeature_2016-Jun-03_cell_method_parsing.txt
@@ -1,0 +1,1 @@
+* The parsing functionality for Cell Methods from netCDF files is available as part of the :mod:`iris.fileformats.netcdf` module.
--- a/lib/iris/fileformats/_pyke_rules/fc_rules_cf.krb
+++ b/lib/iris/fileformats/_pyke_rules/fc_rules_cf.krb
@@ -884,7 +884,6 @@ fc_formula_terms
 fc_extras
     import six
 
-    import re
     import warnings
 
     import biggus
@@ -897,6 +896,7 @@ fc_extras
     import iris.coords
     import iris.coord_systems
     import iris.fileformats.cf as cf
+    import iris.fileformats.netcdf
     import iris.fileformats.pp as pp
     import iris.exceptions
     import iris.std_names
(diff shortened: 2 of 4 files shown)


dkillick:
Note the fact that this is just a rebased version of #1981 means that it retains that PR's unanswered question relating to `cf_var_name`. Options:
 * we merge this as it is, which has the advantage of being nice and quick but the disadvantage of potentially introducing lower-quality code into Iris. That said, apparently cell methods parsing has worked since it was introduced, so it can't be _that_ bad...
 * we wait for the extra effort to get to the bottom of the issue, decide whether it is an issue, make the relevant changes, test them, remove/edit other tests, and _then_ try to get this merged. Oh, and we potentially need to wait for the xarray guys (see [comment to this effect](https://github.com/SciTools/iris/pull/1981#issuecomment-218742408)). Benefits... higher quality code; downsides... missing the 1.10 milestone.

Considering these, I'm in favour of the former - we can add missing stuff later. @pelson @bjlittle thoughts? You both have prior on these changes.

dkillick:
I guess this should also have a whatsnew entry.

marqh:
> I guess this should also have a whatsnew entry.

added

+1 for
> merge this as it is

dkillick:
Good stuff @marqh, I'm merging. Follow-up work can be done as a follow-up once it's better understood what, if anything, needs to be done about `cf_var_name`.

dkillick:
Thanks to @rhattersley (for originally doing the work) and @marqh (for updating) :tada:

pp-mo:
> we merge this as it is,...
> apparently cell methods parsing has worked since it was introduced, so it can't be that bad..
> we wait ... decide whether it is an issue

~~Re-reading, I don't think there is any outstanding *functional* problem, and I wasn't proposing we should enhance the routine :  I'd like it to be a bit clearer about what it can and can't handle, and cover that in the documentation.~~

~~I don't think the xarray related issue makes any difference at this point.~~

~~Ideally, we would remove the functionally superfluous 'cf_name' arg.
As a minimum, just enhance the documentation + move on.~~

you merged it ?!!

dkillick:
>> I'm in favour of the former - we can add missing stuff later

> you merged it?!!

Yes. Feel free to add anything you feel is missing; I didn't feel anything was missing that was dramatic enough to add yet further drag to the 1.10 release .

pp-mo:
> Feel free to add anything you feel is missing

Fair enough "
I was just preparing a suggested tweak.  I may as well complete it, and will post separately..

Issue #2034, opened 2016-05-25
Biggus version compatibility


alaniwi:
The latest Iris appears to require Biggus 0.13.0 or it will give wrong answers to some array manipulations. @matthew-mizielinski has some test code to show this.

Two things (hope it's okay to combine them in one issue):

1) the latest version on PyPI is 0.12.0 but from my experimentation with Matthew's test code this does not fix the problem, so please can you put Biggus 0.13.0 onto PyPI

2) please can you add a check that will raise an exception when trying to import iris while an incompatible Biggus version is installed

Thanks.

marqh:
hello @alaniwi 

> 1) the latest version on PyPI is 0.12.0 but from my experimentation with Matthew's test code this does not fix the problem, so please can you put Biggus 0.13.0 onto PyPI

I'll get onto this and get an update to you

> 2) please can you add a check that will raise an exception when trying to import iris while an incompatible Biggus version is installed

I'm not sure that this is a step which we will take.  I will discuss this, but raising an exception on import is a  pretty hard failure mode with no way of affecting, so it's quite a drastic approach.

I'll get back to you on this
mark

matthew-mizielinski:
If failing on import is too aggressive, could a warning or a unit test be
used to avoid a recurrence of such an issue?

M

ajdawson:
Would you perhaps describe the wrong answers it gets with biggus < 0.13.0? We are clear to note in the install documentation for Iris 1.9.2 that Biggus v0.12.0 or higher is required: http://scitools.org.uk/iris/docs/latest/installing.html#build-and-runtime-requirements. You seem to indicate this is not sufficient, but I think we were unaware of any issues with v0.12.0, so describing your problem would be a good starting point.

matthew-mizielinski:
The way this was originally identified was through a simple difference of lazy loaded cubes; when plotted the areas masked out in the source data were suddenly full of zeros when plotted. The fix for the user was fairly straightforward; force the data to be loaded before performing the difference, i.e. circumvent the use of biggus.

For plotting, this is annoying, but for calculating global means this modifies scientific results (hence why I was worried about this).

The code below demonstrates this error when working with iris v1.9.2 and biggus v0.11.0;
```
import iris, biggus, numpy as np

print iris, iris.__version__
print biggus, biggus.__version__
print np, np.__version__

#create two cubes with masked data, with different values under the mask

[...]

PR #2041, opened 2016-06-03
Biggus version requirement update

ajdawson:
Updates the minimum version of biggus to v0.13 to help users avoid SciTools/biggus#166. I've also added a unit test that checks masked addition. This will fail with biggus < v0.13, and I've taken the unusual step of explaining this to the end user. This is felt to be necessary due to the possible mode of failure, where a masked array would lose its mask silently, endangering the correctness of scientific results. See #2034 for background and the basis for this new test.

Ping @matthew-mizielinski

### DIFF
--- a/conda-requirements.txt
+++ b/conda-requirements.txt
@@ -2,7 +2,7 @@
 # conda create -n <name> --file conda-requirements.txt
 
 # Mandatory dependencies
-biggus>=0.12.0
+biggus>=0.13.0
 cartopy
 matplotlib=1.3.1
 netcdf4
--- a/lib/iris/tests/unit/cube/test_Cube__operators.py
+++ b/lib/iris/tests/unit/cube/test_Cube__operators.py
@@ -22,6 +22,7 @@ from six.moves import (filter, input, map, range, zip)  # noqa
 import iris
 import iris.tests as tests
 import numpy as np
+import numpy.ma as ma
 import biggus
 
 
@@ -159,6 +160,29 @@ class Test_Scalar_Cube_Lazy_Maths(tests.IrisTest):
         data = cube.data
         self.assertTrue(isinstance(data, np.ndarray))
         self.assertEqual(data.shape, ())
+
+
+class Test_Masked_Lazy_Maths(tests.IrisTest):
+
+    def build_lazy_cube(self):
+        data = ma.array([[1., 1.], [1., 100000.]], mask=[[0, 0], [0, 1]])
+        data = biggus.NumpyArrayAdapter(data)
+        cube = iris.cube.Cube(data, standard_name='air_temperature', units='K')
+        lat = iris.coords.DimCoord([-10, 10], 'latitude')
+        lon = iris.coords.DimCoord([10, 20], 'longitude')
+        cube.add_dim_coord(lat, 0)
+        cube.add_dim_coord(lon, 1)
+        return cube
+
+    def test_subtract(self):
+        cube_a = self.build_lazy_cube()
+        cube_b = self.build_lazy_cube()
+        cube_c = cube_a - cube_b
+        self.assertIsInstance(
+            cube_c.data,
+            ma.MaskedArray,
+            msg='known fail with biggus < 0.13.0, consider upgrading')
+
 
 if __name__ == "__main__":
     tests.main()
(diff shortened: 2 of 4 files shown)


lib/iris/tests/unit/cube/test_Cube__operators.py
```diff
@@ -160,5 +161,26 @@ def test_div_cubes(self):
...
+        lon = iris.coords.DimCoord([10, 20], 'longitude')
+        cube.add_dim_coord(lat, 0)
+        cube.add_dim_coord(lon, 1)
+        return cube
+
+    def test_add(self):
+        cube_a = self.build_lazy_cube()
+        cube_b = self.build_lazy_cube()
+        cube_c = cube_a - cube_b
+        self.assertTrue(isinstance(cube_c.data, ma.MaskedArray),
```
dkillick: `self.assertIsInstance(...)`?

lib/iris/tests/unit/cube/test_Cube__operators.py
```diff
@@ -160,5 +161,26 @@ def test_div_cubes(self):
         self.assertTrue(isinstance(data, np.ndarray))
         self.assertEqual(data.shape, ())
 
+
+class Test_Masked_Lazy_Maths(tests.IrisTest):
+
+    def build_lazy_cube(self):
```
dkillick: We have so many examples of this sort of thing in Iris tests! I'm not suggesting you do it here, but a lazy cube provider in `iris.tests.stock` might be a good thing...

lib/iris/tests/unit/cube/test_Cube__operators.py
ajdawson: Perhaps, just copy-paste from previous test in the same file.

lib/iris/tests/unit/cube/test_Cube__operators.py
ajdawson: Indeed, I agree. All 3 in this file are different though so I'm not going to implement anything in this PR.

lib/iris/tests/unit/cube/test_Cube__operators.py
dkillick: > copy-paste from previous test in the same file

I'm much a fan of using the appropriate test method for the test being carried out - really this should have been caught by the reviewer of the change that added the other instances of this. I'd prefer this test method changed nevertheless.

ajdawson:
Changed `assertTrue` to `assertIsInstance`.

lib/iris/tests/unit/cube/test_Cube__operators.py
```diff
@@ -160,5 +161,28 @@ def test_div_cubes(self):
...
+        lat = iris.coords.DimCoord([-10, 10], 'latitude')
+        lon = iris.coords.DimCoord([10, 20], 'longitude')
+        cube.add_dim_coord(lat, 0)
+        cube.add_dim_coord(lon, 1)
+        return cube
+
+    def test_add(self):
+        cube_a = self.build_lazy_cube()
+        cube_b = self.build_lazy_cube()
+        cube_c = cube_a - cube_b
```
dkillick: &#161;Pedantry alert! @ajdawson I'm really sorry but this test method is called `test_add` and you're subtracting your two cubes... :sweat:

ajdawson:
Have updated @dkillick

conda-requirements.txt
```diff
@@ -2,7 +2,7 @@
 # conda create -n <name> --file conda-requirements.txt
 
 # Mandatory dependencies
-biggus>=0.12.0
+biggus>=0.13.0
```
dkillick: @ajdawson you need to make an equivalent change in `minimal_conda_requirements.txt` too. Given that the minimum requirements Travis run is not failing regardless, I am concerned that your test here is not doing what you want...

ajdawson: It does do what I want, I've checked it locally, it just isn't going to get tested on Travis CI because it picks the latest version which is v0.13.0 (see https://travis-ci.org/SciTools/iris/jobs/135538831#L330).

Regardless this line does need to be changed in minimal requirements file, I  will do this.

dkillick:
I'm happy with this, and the failing tests are, once again, timeouts.

dkillick:
Thanks @ajdawson - a useful update.

PR #2043, opened 2016-06-03
Remove redundant arg from parse_cell_methods.

pp-mo:
Some slightly unfinished business from #2039

The docs enhancement is a bit hurried : suggestions to improve ?


marqh:
> The docs enhancement is a bit hurried : suggestions to improve ?

i think this is good enough for now

PR #2044, opened 2016-06-06
Fix to allow zero spaces after colon in cell_method for NetCDF

hdyson:
At present, iris requires a cell method to have at least one space after the colon for a cell_method in a NetCDF file.  For example, this is valid:

> 'time: mean'

While this is not recognised:

> 'time:mean'

This PR fixes that.

As an aside, for the invalid cell method example above, iris currently silently removes the cell method.  Would a warning (or possibly exception) be more appropriate - something like 'cell method foo:bar not recognised'?

### DIFF
--- a/lib/iris/fileformats/netcdf.py
+++ b/lib/iris/fileformats/netcdf.py
@@ -148,7 +148,7 @@ _CM_INTERVAL = 'interval'
 _CM_METHOD = 'method'
 _CM_NAME = 'name'
 _CM_PARSE = re.compile(r'''
-                           (?P<name>([\w_]+\s*?:\s+)+)
+                           (?P<name>([\w_]+\s*?:\s*?)+)
                            (?P<method>[\w_\s]+(?![\w_]*\s*?:))\s*
                            (?:
                                \(\s*
--- a/lib/iris/tests/unit/fileformats/netcdf/test_parse_cell_methods.py
+++ b/lib/iris/tests/unit/fileformats/netcdf/test_parse_cell_methods.py
@@ -36,6 +36,7 @@ class Test(tests.IrisTest):
         cell_method_strings = [
             'time: mean',
             'time : mean',
+            'time :mean',
             ]
         expected = (CellMethod(method='mean', coords='time'),)
         for cell_method_str in cell_method_strings:


cpelley:
Though not strictly following formatting described by the [CF conventions](http://cfconventions.org/cf-conventions/v1.6.0/cf-conventions.html#cell-methods) I can't think of a reason not to support this. +1

hdyson:
I can't replicate the test failure locally (I don't have python3.4) so some help in tracking down what the issue is would be appreciated.  The traceback isn't really giving enough information to understand the issue.

marqh:
> I can't replicate the test failure locally (I don't have python3.4) so some help in tracking down what the issue is would be appreciated. The traceback isn't really giving enough information to understand the issue.

the test failure is an illusion, a false negative

http://cfconventions.org/cf-conventions/v1.6.0/cf-conventions.html#cell-methods
states
```
This is a string attribute comprising a list of blank-separated words of the form "name: method". 
```
The 'space' is part of this definition in the doc, so I'm a little wary of just accepting it.

Do you have a file with such a filename?
Have you tried using CF's cfchecker on it to see if it reports an issue?

I would prefer not to raise an exception during the load process, but i'm a bit worried that by missing this parse rule the metadata is lost.

If you can get me a pointer to a file with this characteristic, I'm happy to have a closer  look.

mark

hdyson:
Hi @marqh 

Thanks for clearing up the test.  

I do have files with 'time:mean' cell methods, yes.  The current (24th May) and earlier test data from here exhibit this: http://www.globalchange.umd.edu/CEDS/  Note that I've been in touch with them, and future data releases should use 'time: mean'.

cfchecker isn't happy with the files either, but I'm not sure that this error is related:  

```
$ cfchecker CEDS_CO_anthro_2000_0.5_v_05_24_2016.nc

CHECKING NetCDF FILE: CEDS_CO_anthro_2000_0.5_v_05_24_2016.nc
=====================
Using CF Checker Version 2.0.3
Using Standard Name Table Version 18 (2011-07-22T10:58:54Z)
Using Area Type Table Version 1 (5 December 2008)

ncopen: filename "CEDS_CO_anthro_2000_0.5_v_05_24_2016.nc": Can't open HDF5 attribute

COULD NOT OPEN FILE, PLEASE CHECK THAT NETCDF IS FORMATTED CORRECTLY.

ERRORS detected: 1
```

marqh:
i think that may be an error with your instance of cfchecker, it's not got as far as checking the file metadata

hdyson:
Hi @marqh, I'm using the centrally installed Met Office version - is there a more recent version I should be using somewhere?

marqh:
hello @hdyson 

I'm afraid that I am not in favour of this change.  I think that the CellMethod attribute syntax for CF should be honoured.

I also think that there is a manageable work around, using file based tools such as NCO's ncatted which enable in place alteration of attributes in netCDF files, so this should be fixable where the siutation arises

I understand this is not ideal for you, so I will leave this PR open, to allow you to restate your case and to enable core developers to disagree with my point of view
@pp-mo @dkillick @bjlittle @pelson @ajdawson 

mark

hdyson:
Hi @marqh,

Thanks for picking this up.  I no longer have a case to restate - upstream fixed their data, so I'm happy.  The question is really whether anyone will trip over this again.   It would perhaps be better to close this off, and add a separate issue for the absence of a warning if there's unrecognised cell methods?

cpelley:
> I'm afraid that I am not in favour of this change. I think that the CellMethod attribute syntax for CF should be honoured.

CF says the following 'of the form "name: method"'.  Is there a disadvantage to not interpreting this in the most strict manner?  Perhaps I'm missing something?
I know what ' mean' means, are we realistically in danger of interpreting cell methods incorrectly due to this change in future?  I think this risk is rather small compared to the considerable benefit of interpreting such files in iris.  Of course I could be missing something, can you elaborate @marqh?

hdyson:
I don't think there's any merit in leaving this open - we have no use case for it now, and if it recurs there are workarounds.

PR #2051, opened 2016-06-10
Make iris.util.new_axis handle lazy data

ajdawson:
Make `iris.util.new_axis` work without loading a deferred data payload. I've used `iris.cube.Cube.lazy_data()` for this, I didn't use a check to see if the cube actually has lazy data, it seems not to matter, but advice is welcome.

### DIFF
--- a/lib/iris/tests/integration/test_new_axis.py
+++ b/lib/iris/tests/integration/test_new_axis.py
@@ -1,0 +1,43 @@
+# (C) British Crown Copyright 2016, Met Office
+#
+# This file is part of Iris.
+#
+# Iris is free software: you can redistribute it and/or modify it under
+# the terms of the GNU Lesser General Public License as published by the
+# Free Software Foundation, either version 3 of the License, or
+# (at your option) any later version.
+#
+# Iris is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU Lesser General Public License for more details.
+#
+# You should have received a copy of the GNU Lesser General Public License
+# along with Iris.  If not, see <http://www.gnu.org/licenses/>.
+"""Integration tests for :func:`iris.util.new_axis`."""
+
+from __future__ import (absolute_import, division, print_function)
+from six.moves import (filter, input, map, range, zip)  # noqa
+
+# Import iris.tests first so that some things can be initialised before
+# importing anything else.
+import iris.tests as tests
+
+import iris
+from iris.util import new_axis
+
+
+class Test(tests.IrisTest):
+
+    @tests.skip_data
+    def test_lazy_data(self):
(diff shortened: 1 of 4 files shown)


ajdawson:
I'd really like to squeeze this into the next release if possible @bjlittle? (I know you said it is closed for further PRs, but the branch has yet to be cut...).

bjlittle:
@ajdawson Our motivation was to push 1.10 out asap, but we're hitting issues with iris-grib baggage.

I suspect that this PR could slip in :wink: , but if not we want a swift point release after 1.10. I'll have a huddle with the other devs to discuss. Because we're just not slick enough with the release mechanism "at the moment" it's inevitable that more PRs will come knocking on the door ...

In the meantime, something has broken the channel dependencies. I re-ran your travis tests to check that this was the case ... and sure enough it is. 

I can't recreate the problem locally, in a fresh conda env and manual stepping through the .travis.yml.

Any ideas? Have you seen this before?

ajdawson:
This perhaps? https://github.com/conda/conda/issues/2675

bjlittle:
Shameful if true ... awesome spot, I'll give it a try, thanks!

bjlittle:
@ajdawson You never guess what ... that was it :scream: 

See #2053

bjlittle:
@ajdawson I think that it would be worth while adding an integration test that actually has a cube with deferred data e.g. `iris.load_cube(iris.sample_data_path('air_temp.pp'))` and uses `iris.util.new_axis` to promote a scalar coordinate e.g. `pressure`, to just make sure that all is well with the world in terms of lazy payload.

What do you think?

ajdawson:
I think the integration test is a nice idea. I will do that.

bjlittle:
Awesome.

ajdawson:
I've added an integration test using a file from the test data repository (not from the sample data repo as you suggested @bjlittle :wink:). I'll rebase onto #2053 once it is merged (I will assume @dkillick is taking care of this).

lib/iris/tests/integration/test_new_axis.py
```diff
@@ -0,0 +1,42 @@
...
+
+class Test(tests.IrisTest):
+
+    @tests.skip_data
+    def test_lazy_data(self):
+        filename = tests.get_data_path(('PP', 'globClim1', 'theta.pp'))
+        cube = iris.load_cube(filename)
+        new_cube = new_axis(cube, 'time')
+        self.assertTrue(cube.has_lazy_data())
+        self.assertTrue(new_cube.has_lazy_data())
```
bjlittle: @ajdawson Did you also want to check the `shape` ?

ajdawson: Sure why not.

ajdawson:
OK I've modified the unit and integration tests. They are now essentially the same as each other, but one uses real data.

bjlittle:
@ajdawson Looks great to me. Thanks! 

Happy for you to squash (now if you want) and rebase (once #2053 is in) :+1:

bjlittle:
@ajdawson #2053 is in ... over to you!

PR #2057, opened 2016-06-16
Add SAMPLE_DATA_DIR deprecation warning.

bjlittle:
This PR addresses the [comment](https://github.com/SciTools/iris/pull/2047#discussion_r66404539) raised by @pp-mo on PR #2047, and deprecates the use of `iris.config.SAMPLE_DATA_DIR` in favour of using the newly packaged and importable `iris_sample_data`.

**Caveat emptor** - Merging this PR means that we are committed to bundling `iris_sample_data` with iris and an update to the iris conda recipe is required, otherwise a deprecation warning will always be raised on using the `iris.sample_data_path` function ... (see discussion thread below)

To do
- [x] Release `iris_sample_data` at `v2.0.0`
- [x] Purge redundant tests `test_sample_data_path`
- [x] Check if `skip_sample_data` required
- [x] Add test coverage for deprecation warning being raised
- [x] Push conda recipe for `iris_sample_data` to the scitools `main` label, see [conda-recipes-scitools PR-169](https://github.com/SciTools/conda-recipes-scitools/pull/169)
- [x] Update the `conda-requirements.txt`
- [x] Passing travis-ci for `example_tests`
- [x] Update .travis.yml to purge `site.cfg` iris-sample-data download

### DIFF
--- a/lib/iris/tests/__init__.py
+++ b/lib/iris/tests/__init__.py
@@ -91,6 +91,13 @@ else:
     GRIB_AVAILABLE = True
     from iris.fileformats.grib.message import GribMessage
 
+try:
+    import iris_sample_data
+except ImportError:
+    SAMPLE_DATA_AVAILABLE = False
+else:
+    SAMPLE_DATA_AVAILABLE = True
+
 
 #: Basepath for test results.
 _RESULT_PATH = os.path.join(os.path.dirname(__file__), 'results')
@@ -917,6 +924,11 @@ def skip_plot(fn):
 
 skip_grib = unittest.skipIf(not GRIB_AVAILABLE, 'Test(s) require "gribapi", '
                                                 'which is not available.')
+
+
+skip_sample_data = unittest.skipIf(not SAMPLE_DATA_AVAILABLE,
+                                   ('Test(s) require "iris_sample_data", '
+                                    'which is not available.'))
 
 
 def no_warnings(func):
--- a/lib/iris/tests/unit/test_sample_data_path.py
+++ b/lib/iris/tests/unit/test_sample_data_path.py
@@ -41,6 +41,7 @@ def _temp_file(sample_dir):
     return sample_path
 
 
+@tests.skip_sample_data
 class TestIrisSampleData_path(tests.IrisTest):
     def setUp(self):
         self.sample_dir = tempfile.mkdtemp()
@@ -49,22 +50,16 @@ class TestIrisSampleData_path(tests.IrisTest):
         shutil.rmtree(self.sample_dir)
 
     def test_path(self):
-        try:
-            with mock.patch('iris_sample_data.path', self.sample_dir):
-                import iris_sample_data
-                self.assertEqual(iris_sample_data.path, self.sample_dir)
-        except ImportError:
-            pass
+        with mock.patch('iris_sample_data.path', self.sample_dir):
+            import iris_sample_data
+            self.assertEqual(iris_sample_data.path, self.sample_dir)
 
     def test_call(self):
-        try:
-            sample_file = _temp_file(self.sample_dir)
-            with mock.patch('iris_sample_data.path', self.sample_dir):
-                import iris_sample_data
-                result = sample_data_path(os.path.basename(sample_file))
-                self.assertEqual(result, sample_file)
-        except ImportError:
-            pass
+        sample_file = _temp_file(self.sample_dir)
+        with mock.patch('iris_sample_data.path', self.sample_dir):
+            import iris_sample_data
+            result = sample_data_path(os.path.basename(sample_file))
+            self.assertEqual(result, sample_file)
 
 
 class TestConfig(tests.IrisTest):
--- a/lib/iris/tests/unit/test_sample_data_path.py
+++ b/lib/iris/tests/unit/test_sample_data_path.py
@@ -41,18 +41,6 @@ def _temp_file(sample_dir):
     return sample_path
 
 
-class TestIrisSampleData__get_path(tests.IrisTest):
-    def test(self):
-        try:
-            path = mock.sentinel.path
-            with mock.patch('iris_sample_data._get_path', return_value=path):
-                import iris_sample_data
-                path = iris_sample_data._get_path()
-                self.assertEqual(path, self.path)
-        except ImportError:
-            pass
-
-
 class TestIrisSampleData_path(tests.IrisTest):
     def setUp(self):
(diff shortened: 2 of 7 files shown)


ajdawson:
>Caveat emptor - Merging this PR means that we are committed to bundling iris_sample_data with iris and an update to the iris conda recipe is required, otherwise a deprecation warning will always be raised on using the iris.sample_data_path function.

This at first does not sound like a good plan because the sample data should always be optional. However, it only crops up if you actually try and use the sample data, so it seems reasonable to see the warning if you want sample data but haven't installed it. In future releases this will just become an exception.

I don't really see why this change would necessitate bundling iris_sample_data with iris... If some one wants to load sample data but uses the old method, the message will be useful, if they don't use sample data then no problem.

bjlittle:
@ajdawson I guess I should re-phrase my statement as a question ... knowing what we know, and given this PR, and say that this PR is merged (once I sort out the example test failures!), should we then add `iris_sample_data` as a dependency in to the iris conda recipe?

For me, the benefit of doing this is that users who choose to install and use iris via the conda scitools channel don't have to jump through the "setting up the iris-sample-data hoop". They get it for free, no harm done.

Like you said, if we don't, it's not a disaster. They either set up the iris-sample-data themselves via manually installing `iris_sample_data` or hook-up the data via the `site.cfg` and simply get a deprecation warning.

I guess I'm saying (at the moment) that I'm kinda in favour joining the dots, and putting `iris_sample_data` into the conda recipe as it's a simple win, but I'm not totally married to the idea.

ajdawson:
I guess my point was more that we (or indeed anyone else packaging iris) don't *have* to make the sample data a dependency. For example, Debian packaging typically incorporates data repositories like this as separate optional packages. We can do what we like with our own conda channel, but I wanted to make it clear that this change seems not to impose a requirement to depend on the sample data package.

bjlittle:
Yup, gotcha. I agree and thanks for making that clear. It's a really good point.

bjlittle:
Ahhh okay, so the issue is being forced somewhat ...

The `docs.iris.example_tests` are failing simply because we throw an exception if any `IrisDeprecation` warnings are raised. Which is fair enough, as we don't want gallery examples spewing out deprecation's.

And of course, since the `iris_sample_data` isn't installed this PRs deprecation warning is being raised, hence the example test failures.

Hmmmm ...

ajdawson:
Well, I guess this means you know it works! It would make sense to couple this change with moving to the new form of obtaining sample data, if it is deprecated we shouldn't be relying on it in our tests.

bjlittle:
@ajdawson Okay, so now I think that it's appropriate to tag the first `iris-sample-data` at `v1.0` ... that will make life a lot simpler all round.

But first, there is a little bit of tidying up to do there first ...

ajdawson:
This makes sense.

bjlittle:
By Jove, it works!

lib/iris/tests/__init__.py
```diff
@@ -91,6 +91,13 @@
     GRIB_AVAILABLE = True
     from iris.fileformats.grib.message import GribMessage
 
+try:
+    import iris_sample_data
+except ImportError:
+    SAMPLE_DATA_AVAILABLE = False
+else:
+    SAMPLE_DATA_AVAILABLE = True
```
dkillick: So slightly weirdly you just removed some very similar code from `iris.__init__` :stuck_out_tongue_winking_eye:

lib/iris/tests/unit/test_sample_data_path.py
```diff
@@ -120,6 +103,16 @@ def test_glob_absolute(self):
         with self.assertRaisesRegexp(ValueError, 'Absolute path'):
             sample_data_path(os.path.abspath('foo.*'))
 
+    def test_warn_deprecated(self):
+        sample_path = _temp_file(self.sample_dir)
+        with mock.patch('warnings.warn') as warn:
+            sample_data_path(os.path.basename(sample_path))
+            self.assertEqual(warn.call_count, 1)
+            (warn_msg, warn_exception), _ = warn.call_args
```
dkillick: Do your two named variables need to be in the tuple..?

lib/iris/tests/__init__.py
bjlittle: Yup, but for continuity I'm using the same pattern within `iris.tests.__init__` that's already established there.

Happy to change this if you want.

lib/iris/tests/unit/test_sample_data_path.py
```diff
@@ -41,18 +42,7 @@ def _temp_file(sample_dir):
...
-            path = mock.sentinel.path
-            with mock.patch('iris_sample_data._get_path', return_value=path):
-                import iris_sample_data
-                path = iris_sample_data._get_path()
-                self.assertEqual(path, self.path)
-        except ImportError:
-            pass
-
-
+@tests.skip_sample_data
```
dkillick: Are there any other tests that use the Iris sample data?

lib/iris/tests/unit/test_sample_data_path.py
bjlittle: Yes, I'm unpacking the arguments passed to the `warn_deprecated` function.

lib/iris/tests/unit/test_sample_data_path.py
bjlittle: Good question! I'll check ...

bjlittle: No other tests use `sample_data_path`, just `iris.tests.unit.test_sample_data_path`

dkillick:
Woo deprecated sample data! Haha, depre-data.

Thanks @bjlittle!

bjlittle:
Awesome, thanks @dkillick!

pp-mo:
> This PR addresses the comment raised by @pp-mo on PR #2047 

Sorry for not responding.
This is all good, but there was one extra thing I was hoping we'd do, which is to get the config module to emit a deprecation warning if there is 'sample_data_path' setting in your config file :  To encourage the removal of any obsolete "magic" from the config file.
What does anyone think about the merits of that ?

pp-mo:
> get the config module to emit a deprecation warning if there is 'sample_data_path' setting in your config file

After more thought, I'm not sure this is appropriate.
There is a problem over when you would remove a "deprecation" :  You want to do this sometime, to reduce clutter.  On removal, an "obsolete" setting becomes just an "unrecognised" one again.  So I don't think this really makes any sense unless an unrecognised setting is actually an _error_.

PR #2059, opened 2016-06-22
Fix PP callback bug

### LINKED ISSUE #2052: IgnoreCubeException not working in v1.10
The IgnoreCubeException does not fire in version 1.10-DEV from a load callback when loading '.pp' files.

For example:

```python
import iris 
iris.FUTURE.netcdf_promote=True
iris.FUTURE.netcdf_no_unlimited=True

def exc_callback(cube, field, filename):
    if cube.name() != 'air_potential_temperature':
        raise iris.exceptions.IgnoreCubeException


def main():
    print iris.__version__

    # Loading a 'pp' file
    # IgnoreCubeException does not ignored the 'air_potential_temperature' cube on load
    fn = 'uk_hires.pp'
    cubes = iris.load(iris.sample_data_path(fn), callback=exc_callback)
    print len(cubes)

    # Loading a 'netcdf' file
    # IgnoreCubeException correctly ignores the 'air_potential_temperature' cube on load
    fn = 'uk_hires.nc'
    iris.save(cubes, fn)
    cubes = iris.load(fn, callback=exc_callback)
    print len(cubes)


if __name__ == '__main__':
    main()
```

And the result of running this...
```
> python ignore_cube_exception_test.py
1.9.2
1
1
> python ignore_cube_exception_test.py
1.10.0-DEV
2
1
```

dkillick:
Fixes a bug (see #2052) where the `IgnoreCubeException` was not being respected in loading the loading of PP files with a user callback.

No tests yet, primarily because there are no tests for this behaviour at all (which is probably why this slipped through in the first place).

Closes #2052

### DIFF
--- a/lib/iris/fileformats/rules.py
+++ b/lib/iris/fileformats/rules.py
@@ -950,7 +950,7 @@ def _resolve_factory_references(cube, factories, concrete_reference_targets,
 
 
 def _load_pairs_from_fields_and_filenames(fields_and_filenames, converter,
-                                          user_callback=None):
+                                          user_callback_wrapper=None):
     # The underlying mechanism for the public 'load_pairs_from_fields' and
     # 'load_cubes'.
     # Slightly more complicated than 'load_pairs_from_fields', only because it
@@ -962,12 +962,12 @@ def _load_pairs_from_fields_and_filenames(fields_and_filenames, converter,
         cube, factories, references = _make_cube(field, converter)
 
         # Post modify the new cube with a user-callback.
-        # This an ordinary Iris load callback, so it takes the filename.
-        if user_callback is not None:
-            cube = user_callback(cube, field, filename)
-            # Callback mechanism may return None, which must not be yielded.
-            if cube is None:
-                continue
+        # This is an ordinary Iris load callback, so it takes the filename.
+        cube = iris.io.run_callback(user_callback_wrapper,
+                                    cube, field, filename)
+        # Callback mechanism may return None, which must not be yielded.
+        if cube is None:
+            continue
 
         # Cross referencing.
         for reference in references:
@@ -1035,13 +1035,15 @@ def load_cubes(filenames, user_callback, loader, filter_function=None):
             loader.legacy_custom_rules.verify(cube, field)
 
         # Then also run user-provided original callback function.
-        cube = iris.io.run_callback(user_callback, cube, field, filename)
-        return cube
+        result = cube
+        if user_callback is not None:
+            result = user_callback(cube, field, filename)
+        return result
 
     all_fields_and_filenames = _generate_all_fields_and_filenames()
     for cube, field in _load_pairs_from_fields_and_filenames(
             all_fields_and_filenames,
             converter=loader.converter,
-            user_callback=loadcubes_user_callback_wrapper):
+            user_callback_wrapper=loadcubes_user_callback_wrapper):
         yield cube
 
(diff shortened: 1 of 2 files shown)


dkillick:
I've now added a test! Its an integration test, because I couldn't figure out how to break the functionality of `fileformats/rules.load_cubes` down into unit-test-sized chunks. Also the behaviour caught here is integrated into the whole PP load framework, so I figure it fits into the integration tests.

dkillick:
Hopefully this will get the tests passing...

dkillick:
ping @bjlittle - passing tests!

bjlittle:
@dkillick On it now ...

lib/iris/fileformats/rules.py
```diff
@@ -1014,15 +1019,6 @@ def load_cubes(filenames, user_callback, loader, filter_function=None):
...
 
-    def _generate_all_fields_and_filenames():
-        for filename in filenames:
-            for field in loader.field_generator(
-                filename, **loader.field_generator_kwargs):
-                # evaluate field against format specific desired attributes
-                # load if no format specific desired attributes are violated
-                if filter_function is None or filter_function(field):
-                    yield (field, filename)
-
```
bjlittle: @dkillick This private functions doesn't appear to have changed, apart from its position in the file.

Sorry for being a pain, but could you paste it back, as it introduces unnecessary noise, thanks.

dkillick: Oops! Swap did occur :unamused: Will shift it back!

lib/iris/fileformats/rules.py
```diff
@@ -1034,6 +1030,15 @@ def loadcubes_user_callback_wrapper(cube, field, filename):
         cube = iris.io.run_callback(user_callback, cube, field, filename)
```
bjlittle: @dkillick For me, this is the nub of the problem, and it's not your doing.

I believe `lines +1034-1035` should now simply be the following:
```python
result = cube
if user_callback is not None:
    result = user_callback(cube, field, filename)
return result
```

I believe the intent of the `loadcubes_user_callback_wrapper` was simply to 

1. raise the deprecation warning
1. fire any registered custom user rules, then
1. execute the `user_callback`. 

The `user_callback` should **not** be executed here by `iris.io.run_callback`. That's the problem.

Rather, it is the `loadcubes_user_callback_wrapper` that should be executed by `iris.io.run_callback` downstream within `_load_pairs_from_fields_and_filenames`, as was prior to your changes.

So make the above change, then back-out your patch to [_load_pairs_from_fields_and_filenames](https://github.com/SciTools/iris/pull/2059/files#diff-71a6ac4e2a835fb3910a8a8a8045474eR952) and replace it with
```python
cube = iris.io.run_callback(user_callback, cube, field, filename)
if cube is None:
    continue
```

lib/iris/fileformats/rules.py
```diff
@@ -963,9 +964,13 @@ def _load_pairs_from_fields_and_filenames(fields_and_filenames, converter,
 
         # Post modify the new cube with a user-callback.
         # This an ordinary Iris load callback, so it takes the filename.
-        cube = iris.io.run_callback(user_callback, cube, field, filename)
+        if user_callback is not None:
+            cube = user_callback(cube, field, filename)
+            # Callback mechanism may return None, which must not be yielded.
+            if cube is None:
+                continue
```
bjlittle: Replace `lines +967-971` with
```python
cube = iris.io.run_callback(user_callback, cube, field, filename)
if cube is None:
    continue
```

bjlittle:
@dkillick Great, thanks!

PR #2062, opened 2016-06-27
BUG: Tolerant crs comparison

### LINKED ISSUE #716: BUG: coord_system datatypes equality beyond original precision
Comparing the coord system of one coord with another returns false it it differs by only floating point precision (in the original data type from which it was derived).
These are stored on our coord_system object as python floats (which is a C double) even if it is derived from a 32bit float within a pp.

Example:

``` Python
cd1 = coord1.coord_system.__dict__
print cd1
{'grid_north_pole_latitude': 35.5, 'ellipsoid': GeogCS(6371229.0), 'north_pole_grid_longitude': 0.0, 'grid_north_pole_longitude': 322.9}
cd2 = coord2.coord_system.__dict__
print cd2
{'grid_north_pole_latitude': 35.5, 'ellipsoid': GeogCS(6371229.0), 'north_pole_grid_longitude': 0.0, 'grid_north_pole_longitude': 322.8999938964844}

cd1['grid_north_pole_longitude'] == cd2['grid_north_pole_longitude']
False

np.float32(cd1['grid_north_pole_longitude']) == np.float32(cd2['grid_north_pole_longitude'])
True
```

It was a 32bit float when it was in a pp.

cpelley:
Provide tolerant coordinate system comparison in iris.

Currently, coordinate system comparison in iris has the following issues associated that this PR addresses:
1. Some attributes for specifying a crs can be specified or derived (this makes floating point comparison flawd) e.g. inverse flattening for defining the ellipse rather than defining both semi major AND minor axes.
2. crs attributes can be specified as 32-bit floats (consider pp as a source for the instantiation of the crs (see #716 for an illustrative case).
3. crs equality was using the class name in its equality measure.

**Solution:**

I make a 'tolerant' dict comparison between the coordinate systems (also I remove the requirement for both to be defined from the same class).  Tolerance value derived by the unittests written (see  [lib/iris/tests/unit/coord_systems/test_GeogCS.py](https://github.com/SciTools/iris/pull/2062/files#diff-c049627b2a11991ec3132128f483e441)).

**Context:**

In our current project, we are monkey-patching the `iris.coord_system.Coor.__eq__` method in order that all coordinate systems can be handled effectively.  This PR is about removing the need for our own project hack.

Closes #716

**DO NOT MERGE**

### DIFF
--- a/lib/iris/coord_systems.py
+++ b/lib/iris/coord_systems.py
@@ -31,6 +31,30 @@ import cartopy
 import cartopy.crs as ccrs
 
 
+def _dict_nearly_equal(dict1, dict2):
+    # Allow coordinate system comparison (floating point numbers with
+    # tolerance).
+    for key, value in dict1.iteritems():
+        if key not in dict2:
+            result = False
+            break
+        if hasattr(dict2[key], '__dict__') and hasattr(value, '__dict__'):
+            result = _dict_nearly_equal(dict2[key].__dict__,
+                                        value.__dict__)
+        elif hasattr(dict2[key], '__dict__') != hasattr(value, '__dict__'):
+            result = False
+        else:
+            # We intentionally do not use a global tolerance for coordinate
+            # system comparison.
+            try:
+                result = np.allclose(dict2[key], value, 1e-5)
+            except (TypeError, ValueError):
+                result = dict2[key] == value
+        if not result:
+            break
+    return result
+
+
 class CoordSystem(six.with_metaclass(ABCMeta, object)):
     """
     Abstract base class for coordinate systems.
@@ -40,8 +64,10 @@ class CoordSystem(six.with_metaclass(ABCMeta, object)):
     grid_mapping_name = None
 
     def __eq__(self, other):
-        return (self.__class__ == other.__class__ and
-                self.__dict__ == other.__dict__)
+        res = False
+        if other is not None:
+            res = _dict_nearly_equal(self.__dict__, other.__dict__)
+        return res
 
     def __ne__(self, other):
         # Must supply __ne__, Python does not defer to __eq__ for
--- a/lib/iris/tests/unit/coord_systems/test_GeogCS.py
+++ b/lib/iris/tests/unit/coord_systems/test_GeogCS.py
@@ -1,0 +1,79 @@
+# (C) British Crown Copyright 2016, Met Office
+#
+# This file is part of Iris.
+#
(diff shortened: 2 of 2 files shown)


lib/iris/coord_systems.py
```diff
@@ -31,6 +31,34 @@
...
+                                            value.__dict__)
+            elif hasattr(dict2[key], '__dict__') != hasattr(value, '__dict__'):
+                result = False
+            else:
+                # We intentionally do not use a global tolerance for coordinate
+                # system comparison.
+                if True in [dict2[key] is None, value is None]:
+                    result = dict2[key] == value
+                elif np.isreal(dict2[key]) and np.isreal(value):
+                    result = abs(dict2[key] - value) < 1
```
cpelley: This is an intentionally naive/very accepting here.  Feel free to share ideas you might have (np.allclose etc.  what might fit more generally package wide??).

cpelley:
I have intentionally not changed those existing tests which are effected by this change.  It would be good to get agreement in principle that this change seems reasonable first.

Cheers

cpelley:
Is there no interest in this change?
I hope we won't have to monkey-patch the `iris.coord_system.Coord.__eq__` forever :disappointed:

cpelley:
Closing as I don't see any interest on this.

jkettleb:
Is it clear why this isn't a good idea and so not worthy including in iris? I can't see any comments giving us any indication why is not been picked up.

cpelley:
Hi @jkettleb 

Thanks for the prod, I should have provided some context:

I began a wider discussion around float handling in iris some time ago:
https://groups.google.com/forum/#!topic/scitools-iris-dev/aD4cessUZ8A

I should probably stress that I my view remains that iris float handling is a serious problem (if not the most problematic we face).  After all, we and our users experience many issues as a result, some mild (failures), some serious (silent continuation with incorrect results).  This PR was intended to provide a local fix for a short term fix for a small subset of the problem in order to allow us to remove a monkey-patch we have for iris in our package.
The discussion on the google forums intended to expose the problem package wide and offer a hypothetical solution (framework) to solving this package wide problem and spark discussion.  Unfortunately, I don't see that the problem of float handling in iris is understood or indeed it's implications.  This is apparent from the forum discussion.  I don't see how we can gain traction on this PR while we can't agree that there is a serious problem to fix... :(

I won't delete this branch in case in case this sparks some interest.

PR #2067, opened 2016-06-29
Support plotting non-gregorian calendars

lbdreyer:
This depends on a release of [nc_time_axis](https://github.com/SciTools/nc-time-axis) and for [the recipe](https://github.com/SciTools/conda-recipes-scitools/pull/173) to be added to the scitools channel


lbdreyer:
The graphics tests it is failing on for the two updated gallery images **don't** fail when I reproduce the conda environment on my machine!

lbdreyer:
Even though I had recreated the conda environment that travis uses, I still had some pth files hanging around. Once I got rid of those, it changed the images from the failing tests.

It may be of interest that during these investigations I found that the rms for the Orca Projection test, on my machine was ~8.9 and the rms for the coriolis plot test was ~0.3.

lib/iris/plot.py
```diff
@@ -403,8 +405,14 @@ def _draw_2d_from_points(draw_method_name, arg_func, cube, *args, **kwargs):
 
 def _fixup_dates(coord, values):
     if coord.units.calendar is not None and values.ndim == 1:
-        r = [datetime.datetime(*(coord.units.num2date(val).timetuple()[0:6]))
-             for val in values]
+        num2date = coord.units.num2date
+        if coord.units.calendar == 'gregorian':
```
marqh: is this code path necessary? can netcdftime handle the gregorian calendar as well?

lib/iris/plot.py
```diff
@@ -403,8 +405,14 @@ def _draw_2d_from_points(draw_method_name, arg_func, cube, *args, **kwargs):
...
+            r = [datetime.datetime(*(num2date(val).timetuple()[0:6]))
```
marqh: the double use of 
```
*(num2date(val).timetuple()[0:6])
```
is a shame, could this be factored out?

also, a comment, regarding the syntax, would help, what is the functionality provided by the `*`, why `[0:6]`

lib/iris/plot.py
lbdreyer: Right now, nc_time_axis doesn't provide all the functionality that matplotlib can do with datetime.datetimes (e.g. extended handling of tick locators). Until this is the case, it would be best to make use of datetimes for gregorian calendars.

lib/iris/plot.py
lbdreyer: I have factored out `num2date(val).timetuple()[0:6]` but I have to leave the unpacking until the datetime call as each of the 6 values in the tuple are arguments of the datetime.datetime or netcdftime.datetime
i.e.
```
>>> d = (coord.units.num2date(2).timetuple()[0:6])
>>> print d
(2000, 2, 3, 0, 0, 0)
>>> print datetime.datetime(*d)
2000-02-03 00:00:00

```

lbdreyer:
The tests had time out errors so I've tried to restart the build.

marqh:
thank you @lbdreyer

PR #1911, opened 2016-01-15
Remove coord checks when adding/subtracting cubes

rcomer:
As discussed at #1887, this PR removes checks which are applied to coordinates before adding and subtracting cubes.  It makes these operations consistent with multiplication and division in terms of whether two cubes are compatible.

Having bypassed `_add_subtract_common`, scalar coords which exist on `cube` but not on `other` ("ignorable" coords) are no longer removed as part of the operation.  I think this is consistent with allowing the vector coords to differ (if I've understood correctly, `_binary_op_common` ignores all of the coords on `other`).  There are no unit tests to check whether the "ignorable" coords are removed, but I am getting a failure in the doctests for `userguide/cube_maths.rst` because a print out of a cube shows more metadata than previously.

I have also removed the depreciated (since 0.8.0) keyword `ignore`.  It's not 100% clear what it did, but I guess it triggered the above mentioned removal of "ignorable" coords.

PR #2068, opened 2016-07-01
Addition to pearsonr documentation

MahatmaCane:
Possible clarification of broadcastability of cubes in iris.analysis.stats.pearsonr function, by request of internal user via AVD Support e-mail account

### DIFF
--- a/lib/iris/analysis/stats.py
+++ b/lib/iris/analysis/stats.py
@@ -48,9 +48,14 @@ def pearsonr(cube_a, cube_b, corr_coords=None, weights=None, mdtol=1.,
     Args:
 
     * cube_a, cube_b (cubes):
-        Cubes between which the correlation will be calculated.  The cubes
+        Cubes between which the correlation will be calculated. The cubes
         should either be the same shape and have the same dimension coordinates
-        or one cube should be broadcastable to the other.
+        or one cube should be broadcastable to the other. Currently, this means
+        the trailing dimensions of the larger cube (those to the right of the
+        dimension of interest in cube.shape) should be the same shape as the
+        corresponding dimensions of the smaller cube and the leading dimensions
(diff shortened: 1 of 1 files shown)


MahatmaCane:
Ping @bjlittle - would you mind taking a look at this?

ajdawson:
I'm not sure this actually makes it clearer. I think the broadcasting rules are the same as for numpy arrays, so perhaps we could just link to their existing description? (http://docs.scipy.org/doc/numpy/user/basics.broadcasting.html)

corinnebosley:
I think a link is a good idea.  The user also suggested adding an example, which I always find helpful in docstrings.

corinnebosley:
Also, if you look at the details for the failing tests, it shows that you need to change the date at the top of the altered module to 2016 instead of 2015, and shorten line 197 by 3 letters (or split it up so that neither lines exceed 79 letters).  After you've changed these things, you should pass all the integration tests when you push.

rcomer:
The broadcasting rules here come from those in cube multiplication, which are more flexible than numpy broadcasting.  If the shapes don't match in a numpy-broadcasting sense, there is an attempt to rearrange the second cube.  See this try-except loop in [`_binary_op_common`](https://github.com/SciTools/iris/blob/master/lib/iris/analysis/maths.py#L592) for the relevant logic.

There is some discussion of broadcasting under [Cube Maths](http://scitools.org.uk/iris/docs/latest/userguide/cube_maths.html) in the Iris User Guide, but perhaps we'd benefit from an extra section there that spells out what does and doesn't work.  Then we could link to there from the pearsonr doc and anywhere else that uses it (and only have to update in one place if/when cube broadcasting rules change).

The flaw in my suggestion is that there are some inconsistencies with broadcasting for different types of operations, which are proving difficult to disentangle - see #1911.

PR #2070, opened 2016-07-06
naming type for CellMethod changed

corinnebosley:
@pp-mo Previously the name used for Cell Methods was the var_name, but this was causing some human confusion and there was also a discrepancy between the name used in different ways:

```
precipitation_flux / (kg m-2 s-1)   (time: 30; latitude: 144; longitude: 192)
     Dimension coordinates:
          time                           x             -               -
          latitude                       -             x               -
          longitude                      -             -               x
     Auxiliary coordinates:
          forecast_period                x             -               -
     Scalar coordinates:
          forecast_reference_time: 1988-09-01 00:00:00
     Attributes:
          Conventions: CF-1.5
          um_version: 10.3
     Cell methods:
          mean: time_0 (1 hour)
```
```
In [24]: CellMethod('mean', cube.coord('time'), '1 hour')
Out[24]: CellMethod(method='mean', coord_names=(u'time',), intervals=('1 hour',), comments=())
```

There should be consistency in the naming type of the Cell Method, so I have added a post-processing method which changes the var_name to the standard_name.  

I am not convinced in the slightest that this is efficient, and I am also unsure that this will work correctly for cubes with multiple Cell Methods.  Advice would be much appreciated.

### DIFF
--- a/lib/iris/fileformats/netcdf.py
+++ b/lib/iris/fileformats/netcdf.py
@@ -531,6 +531,24 @@ def _load_cube(engine, cf, cf_var, filename):
     for attr_name, attr_value in tmpvar:
         _set_attributes(cube.attributes, attr_name, attr_value)
 
+    names_dict = {}
+    for coord in cube.coords():
+        names_dict[coord.var_name] = coord.name()
+
+    for i in range(len(cube.cell_methods)):
+        for j in range(len(cube.cell_methods[i].coord_names)):
+            method = cube.cell_methods[i].method[j]
+            intervals = cube.cell_methods[i].intervals[j]
+            comments = cube.cell_methods[i].comments
+            pattern = re.compile(r'\b(' + '|'.join(names_dict.keys()) + r')\b')
+            new_name = pattern.sub(lambda x: names_dict[x.group()],
+                                   cube.cell_methods[i].coord_names[j])
+            cube.cell_methods = (iris.coords.CellMethod(method=method,
+                                                        coords=new_name,
+                                                        intervals=intervals,
+                                                        comments=comments),)
+
+
     # Show pyke session statistics.
     _pyke_stats(engine, cf_var.cf_name)
 
--- a/lib/iris/fileformats/netcdf.py
+++ b/lib/iris/fileformats/netcdf.py
@@ -531,22 +531,15 @@ def _load_cube(engine, cf, cf_var, filename):
     for attr_name, attr_value in tmpvar:
         _set_attributes(cube.attributes, attr_name, attr_value)
 
-    names_dict = {}
-    for coord in cube.coords():
-        names_dict[coord.var_name] = coord.name()
+    names_dict = {coord.var_name: coord.name() for coord in cube.coords()}
 
-    for i in range(len(cube.cell_methods)):
-        for j in range(len(cube.cell_methods[i].coord_names)):
-            method = cube.cell_methods[i].method[j]
-            intervals = cube.cell_methods[i].intervals[j]
-            comments = cube.cell_methods[i].comments
-            pattern = re.compile(r'\b(' + '|'.join(names_dict.keys()) + r')\b')
-            new_name = pattern.sub(lambda x: names_dict[x.group()],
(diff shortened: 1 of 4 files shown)


corinnebosley:
@pp-mo This test is failing because it expects to see the var_name as the cell_method coord_name in test_trajectory.py.

I am not sure how to change the reference result so that it matches, and I would also like to double check that this is a reasonable change before I go changing reference results and Iris behaviour.

marqh:
I have some waryness about this change, I'm not saying it's wrong, but I want to be cautious.

please read http://cfconventions.org/cf-conventions/v1.6.0/cf-conventions.html#cell-methods and consider 
```
Each "name: method" pair indicates that for an axis identified by name, the cell values representing the field have been determined or derived by the specified method. 

In the specification of this attribute, name can be a dimension of the variable, a scalar coordinate variable, a valid standard name, or the word "area"
```

Is this the approach we want to take?
How widespread a change could this be for users?

tbc...

lib/iris/fileformats/netcdf.py
```diff
@@ -531,6 +531,18 @@ def _load_cube(engine, cf, cf_var, filename):
     for attr_name, attr_value in tmpvar:
         _set_attributes(cube.attributes, attr_name, attr_value)
 
+    names = {coord.var_name: coord.name() for coord in cube.coords()}
+
+    cube.cell_methods = [iris.coords.CellMethod(method=method.method,
+                                                intervals=method.intervals,
+                                                comments=method.comments,
+                                                coords=[names[coord_name]
```
pp-mo: You could maybe restart the indent here, it seems to have got a bit extreme!

pp-mo:
You still have a couple of tests failing here :
 * ```iris.tests.test_netcdf.TestNetCDFLoad.test_cell_methods``` 
 * ```iris.tests.test_trajectory.TestTrajectory.test_tri_polar```

Both need an updated CML reference file.
The actual changes look ok to me.

lib/iris/fileformats/netcdf.py
```diff
@@ -534,14 +534,13 @@ def _load_cube(engine, cf, cf_var, filename):
...
-                                                 coord_name for coord_name in
-                                                 method.coord_names])
+    cube.cell_methods = [iris.coords.CellMethod(
+                             method=method.method,
+                             intervals=method.intervals,
+                             comments=method.comments,
+                             coords=[names[coord_name]
+                                     if coord_name in names else coord_name
+                                     for coord_name in method.coord_names])
                          for method in cube.cell_methods]
```
corinnebosley: @marqh @pp-mo 
I am at a loss.  I can't see why this is failing pep8.

mo-g: Mixed indentation methods. You've tried to do a step indent on top of a visual indent, and pep8 doesn't like it. You need to either do a visual indent to level with Method( - I assume this would take you to past 79chars, or - switch the whole statement to 4char step indenting.

corinnebosley:
@marqh @pp-mo @mo-g 
I got 19 TimedOutExceptions.  How do I fix that?

marqh:
The change being proposed here is that Iris will use standard_name in preference to var_name to identify the coordinate a CellMethod is defined with respect to.

This is a different preference order from the one used in CF-netCDF, although both are valid

When Iris writes a CF-netCDF file it will use the dimension name to identify the cell method

This is a subtle representation change within Iris, which is more aligned with Iris' coordinate identification scheme.  It is not expected to manifest itself in any netCDF files written out by Iris

marqh:
> I got 19 TimedOutExceptions. How do I fix that?

i restarted them for you

corinnebosley:
@pp-mo @marqh 
Just waiting for a merge on this now, if everyone is happy...

marqh:
i'm now in favour of this change; I think it makes CellMethod instances more Iris like, without altering the CF-netCDF encoding preference for dimension names, despite these not being standardised

I think the impact on users will be minimal, as there are no `get_cell_method` functions which use name as part of their call

@corinnebosley please may you add a what's new snippet to 
describing this change in representation?

PR #2074, opened 2016-07-07
move ff warning

marqh:
The warning for a lack of FF subgrid is moved to a place where it will only warn if it is to be used.

This avoids raising warnings about code which is not being used


marqh:
@pp-mo 

please may you consider whether this can just be merged, or whether there is a need for more work?

thank you

Issue #2073, opened 2016-07-07
Merge error for a mixture of loaded and lazy data


rcomer:
I've stumbled across a problem whereby cubes would not merge, and eventually realized it was because some of the data has been loaded but some hadn't.  Here's an example that reproduces the error:

```python
filename = iris.sample_data_path('uk_hires.pp')
cubes =  iris.load_raw(filename)[1:3]
a = cubes[0].data
print cubes.merge_cube()
```
the merge fails with this error:
```
    raise iris.exceptions.MergeError(msgs)
iris.exceptions.MergeError: failed to merge into a single cube.
  cube data fill_value differs: 1e+20 != -1e+30
```
In this example the cubes will obviously merge fine if they are either both loaded or both lazy.  I doubt it makes sense to try to fix this so that they merge in the case where only some data is loaded (presumably you'd have to load everything?)  However, it might be helpful to have a specific error message for the loaded/lazy case, to make a work-around more obvious to the user.

ajdawson:
After #2075 fill value is simply ignored on merge, fixing this issue.

PR #2075, opened 2016-07-08
Ignore fill_value on merge

ajdawson:
I'm going to be bold and propose that we allow cubes to merge when their fill values are different, with the idea that it shouldn't matter what the fill value is, all that matters is that masks are preserved. Biggus is perfectly capable of creating an `ArrayStack` from arrays with different fill values, resulting in an array that is correctly masked. Therefore I don't think we have a problem.

This PR simply removes `fill_value` from `_CubeSignature` so it is not considered when merging. This change resolves the issue noted in #2073 and should resolve other issues noted in the discussion list [here](https://groups.google.com/forum/?fromgroups#!topic/scitools-iris/NeDiBpWPc7Y) and [here](https://groups.google.com/forum/?fromgroups#!topic/scitools-iris/sRhwpSOigGg). I'd also like to note that concatenation does not seem to care about fill value, so there is an element of consistency in doing this.

I'd appreciate feedback on the idea, in particular and caveats of doing this that I may have overlooked.


bjlittle:
@ajdawson Spooky, I literally tripped over this yesterday!

Interesting approach @ajdawson ... also pointing out the inconsistency between merge and concatenate plays into this PRs favour.

marqh:
:+1:

marqh:
all in favour so far, i'm going to merge this

thank you @ajdawson

PR #2079, opened 2016-07-10
Extend point in cell

pp-mo:
Adapt PointInCell regridder to make it work for cross-coordinate-system regridding, including non-latlon systems.

NOTE: this removes certain input requirements, but also adds some, so is not totally backwards-compatible :
**requirements removed:**
  * source and target must have the same coordinate system
  * coord systems must be latlon
  * source data X and Y coords must be 2d (can now have any dimensionality)

**requirements added:**
  * source and target must both have a **defined** coord-system
    * previously 'None' was acceptable, but that is problematic now it isn't limited to latlon systems
  * source data X and Y coordinates must now have **the same** cube dimensions
    * previously they had to be 2d, and the same size, but could have different dimension ordering

**I think these changes are acceptable, as it is experimental**
It basically *will* work for all existing cases, with just simple adaptations in certain cases

### DIFF
--- a/lib/iris/experimental/regrid.py
+++ b/lib/iris/experimental/regrid.py
@@ -1,4 +1,4 @@
-# (C) British Crown Copyright 2013 - 2015, Met Office
+# (C) British Crown Copyright 2013 - 2016, Met Office
 #
 # This file is part of Iris.
 #
--- a/lib/iris/experimental/regrid.py
+++ b/lib/iris/experimental/regrid.py
@@ -1072,6 +1072,11 @@ def _regrid_weighted_curvilinear_to_rectilinear__prepare(
     # contribution from one or more source points.
     rows = np.nonzero(sum_weights)
 
+    # NOTE: when source points are masked, this 'sum_weights' is possibly
+    # incorrect and needs re-calculating.  Likewise 'rows' may cover target
+    # cells which happen to get no data.  This is dealt with by adjusting as
+    # required in the '__perform' function, below.
+
     regrid_info = (sparse_matrix, sum_weights, rows, grid_cube)
     return regrid_info
 
@@ -1087,15 +1092,35 @@ def _regrid_weighted_curvilinear_to_rectilinear__perform(
     sparse_matrix, sum_weights, rows, grid_cube = regrid_info
 
     # Calculate the numerator of the weighted mean (M, 1).
-    if not ma.isMaskedArray(src_cube.data):
+    is_masked = ma.isMaskedArray(src_cube.data)
+    if not is_masked:
         data = src_cube.data
     else:
         # Use raw data array
         data = src_cube.data.data
-        # Zero any masked source points so they add nothing in output sums.
-        if np.ma.is_masked(src_cube.data):
+        # Check if there are any masked source points to take account of.
+        is_masked = np.ma.is_masked(src_cube.data)
+        if is_masked:
+            # Zero any masked source points so they add nothing in output sums.
             mask = src_cube.data.mask
             data[mask] = 0.0
+            # Calculate a new 'sum_weights' to allow for missing source points.
+            # N.B. it is more efficient to use the original once-calculated
+            # sparse matrix, but in this case we can't.
+            # Hopefully, this post-multiplying by the validities is less costly
+            # than repeating the whole sparse calculation.
+            valid_src_cells = ~mask.flat[:]
+            src_cell_validity_factors = sparse_diags(
+                np.array(valid_src_cells, dtype=int),
+                0)
+            valid_weights = sparse_matrix * src_cell_validity_factors
+            sum_weights = valid_weights.sum(axis=1).getA()
+            # Work out where output cells are missing all contributions.
+            # This allows for where 'rows' contains output cells that have no
+            # data because of missing input points.
+            zero_sums = sum_weights == 0.0
+            # Make sure we can still divide by sum_weights[rows].
+            sum_weights[zero_sums] = 1.0
 
     # Calculate sum in each target cell, over contributions from each source
     # cell.
@@ -1104,19 +1129,15 @@ def _regrid_weighted_curvilinear_to_rectilinear__perform(
     # Create a template for the weighted mean result.
     weighted_mean = ma.masked_all(numerator.shape, dtype=numerator.dtype)
 
-    # Account for any missing input data points.
-    if np.ma.is_masked(src_cube.data):
-        # Calculate a new 'sum_weights' to account for missing source points.
-        valid_src_cells = ~mask.flat[:]
-        src_cell_validity_factors = sparse_diags(
-            np.array(valid_src_cells, dtype=int),
-            0)
-        valid_weights = sparse_matrix * src_cell_validity_factors
-        sum_weights = valid_weights.sum(axis=1)
-        sum_weights = np.maximum(1, sum_weights).getA()
-
     # Calculate final results in all relevant places.
     weighted_mean[rows] = numerator[rows] / sum_weights[rows]
+    if is_masked:
+        # Ensure masked points where relevant source cells were all missing.
+        if np.any(zero_sums):
+            # Make masked if it wasn't.
+            weighted_mean = np.ma.asarray(weighted_mean)
+            # Mask where contributing sums were zero.
+            weighted_mean[zero_sums] = np.ma.masked
 
     # Construct the final regridded weighted mean cube.
     tx = grid_cube.coord(axis='x', dim_coords=True)
--- a/lib/iris/experimental/regrid.py
+++ b/lib/iris/experimental/regrid.py
@@ -836,10 +836,10 @@ def regrid_weighted_curvilinear_to_rectilinear(src_cube, weights, grid_cube):
     * src_cube:
         A :class:`iris.cube.Cube` instance that defines the source
         variable grid to be regridded.
-    * weights:
+    * weights (array or None):
         A :class:`numpy.ndarray` instance that defines the weights
         for the source variable grid cells. Must have the same shape
-        as the :data:`src_cube.data`.
+        as the X and Y coordinates.  If weights is None, all-ones will be used.
     * grid_cube:
         A :class:`iris.cube.Cube` instance that defines the target
         rectilinear grid.
@@ -848,10 +848,21 @@ def regrid_weighted_curvilinear_to_rectilinear(src_cube, weights, grid_cube):
         A :class:`iris.cube.Cube` instance.
 
     """
-    if src_cube.shape != weights.shape:
-        msg = 'The source cube and weights require the same data shape.'
-        raise ValueError(msg)
+    regrid_info = \
+        _regrid_weighted_curvilinear_to_rectilinear__calculate_regrid_info(
+            src_cube, weights, grid_cube)
+    result = _regrid_weighted_curvilinear_to_rectilinear__perform_regrid(
+        src_cube, regrid_info)
+    return result
 
+
+def _regrid_weighted_curvilinear_to_rectilinear__calculate_regrid_info(
+        src_cube, weights, grid_cube):
+    """
+    Check inputs and calculate the sparse info needed for the
+    'regrid_weighted_curvilinear_to_rectilinear' calculation.
+
+    """
     if src_cube.aux_factories:
         msg = 'All source cube derived coordinates will be ignored.'
         warnings.warn(msg)
@@ -898,6 +909,13 @@ def regrid_weighted_curvilinear_to_rectilinear(src_cube, weights, grid_cube):
         msg = 'The source cube x ({!r}) and y ({!r}) coordinates must ' \
             'map onto the same cube dimensions.'
         raise ValueError(msg.format(sx.name(), sy.name()))
+
+    if weights is None:
+        weights = np.ones(sx.shape)
+    if weights.shape != sx.shape:
+        msg = ('Provided weights must have the same shape as the X and Y '
+               'coordinates.')
+        raise ValueError(msg)
 
     if tx.coord_system != ty.coord_system:
         msg = 'The target grid cube x ({!r}) and y ({!r}) coordinates must ' \
@@ -1023,13 +1041,14 @@ def regrid_weighted_curvilinear_to_rectilinear(src_cube, weights, grid_cube):
     # target cube cell.
 
     # Determine the valid indices and their offsets in M x N space.
-    if ma.isMaskedArray(src_cube.data):
-        # Calculate the valid M offsets, accounting for the source cube mask.
-        mask = ~src_cube.data.mask.flatten()
-        cols = np.where((y_indices >= 0) & (y_indices < ty_depth) &
-                        (x_indices >= 0) & (x_indices < tx_depth) &
-                        mask)[0]
-    else:
+#    if ma.isMaskedArray(src_cube.data):
+#        # Calculate the valid M offsets, accounting for the source cube mask.
+#        mask = ~src_cube.data.mask.flatten()
+#        cols = np.where((y_indices >= 0) & (y_indices < ty_depth) &
+#                        (x_indices >= 0) & (x_indices < tx_depth) &
+#                        mask)[0]
+#    else:
+    if 1:
         # Calculate the valid M offsets.
         cols = np.where((y_indices >= 0) & (y_indices < ty_depth) &
                         (x_indices >= 0) & (x_indices < tx_depth))[0]
@@ -1052,6 +1071,28 @@ def regrid_weighted_curvilinear_to_rectilinear(src_cube, weights, grid_cube):
     sparse_matrix = csc_matrix((data, (rows, cols)),
                                shape=(grid_cube.data.size, src_cube.data.size))
 
+    regrid_info = (sparse_matrix, grid_cube)
+    return regrid_info
+
+
+def _regrid_weighted_curvilinear_to_rectilinear__perform_regrid(
+        src_cube, regrid_info):
+    """
+    Perform actual regrid calculation for
+    'regrid_weighted_curvilinear_to_rectilinear'.
+
+    """
+    sparse_matrix, grid_cube = regrid_info
+
+    if not ma.isMaskedArray(src_cube.data):
+        data = src_cube.data
+    else:
+        data = src_cube.data.data
+        if np.ma.is_masked(src_cube.data):
+            # Also must adjust sparse matrix for missing data ...
+            # sparse matrix is (M,N), grid is (M,), data is (N,).
+            sparse_matrix[:, src_cube.data.mask] = 0.0
+
     # Performing a sparse sum to collapse the matrix to (M, 1).
     sum_weights = sparse_matrix.sum(axis=1).getA()
 
@@ -1067,6 +1108,10 @@ def regrid_weighted_curvilinear_to_rectilinear(src_cube, weights, grid_cube):
     weighted_mean[rows] = numerator[rows] / sum_weights[rows]
 
     # Construct the final regridded weighted mean cube.
+    tx = grid_cube.coord(axis='x', dim_coords=True)
+    ty = grid_cube.coord(axis='y', dim_coords=True)
+    tx_dim, = grid_cube.coord_dims(tx)
+    ty_dim, = grid_cube.coord_dims(ty)
     dim_coords_and_dims = list(zip((ty.copy(), tx.copy()), (ty_dim, tx_dim)))
     cube = iris.cube.Cube(weighted_mean.reshape(grid_cube.shape),
                           dim_coords_and_dims=dim_coords_and_dims)
@@ -1170,9 +1215,25 @@ class _CurvilinearRegridder(object):
                              'source grid as this regridder.')
 
         # Call the regridder function.
-        res = regrid_weighted_curvilinear_to_rectilinear(src, self.weights,
-                                                         self._target_cube)
-        return res
+#        res = regrid_weighted_curvilinear_to_rectilinear(
+#            src, self.weights, self._target_cube)
+#        return res
+        # This includes repeating over any non-XY dimensions, because the
+        # underlying routine does not support this.
+        # Just for now, we will use cube.slices and merge to achieve this,
+        # though that is not a terribly efficient method ...
+        regrid_info = None
+        result_slices = iris.cube.CubeList([])
+        for slice_cube in src.slices(sx):
+            if regrid_info is None:
+                # Calculate the basic regrid info just once.
+                regrid_info = \
+                    _regrid_weighted_curvilinear_to_rectilinear__calculate_regrid_info(
+                        slice_cube, self.weights, self._target_cube)
+            result_slices.append(_regrid_weighted_curvilinear_to_rectilinear__perform_regrid(
+                slice_cube, regrid_info))
+        result = result_slices.merge_cube()
+        return result
 
 
 class PointInCell(object):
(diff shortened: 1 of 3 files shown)


pp-mo:
Taking a quick overview of this, I think it is best seen as a single change after all, rather than trying to understand the development process.
In fact only 3 files changed, and I think we can squash all commits, when discussions + further changes are done.
Meanwhile, quite a few of the comments in the basic ```experimental/regrid.py``` source look like they need reviewing and tidying.  I will get onto that...

lib/iris/experimental/regrid.py
```diff
@@ -827,43 +874,59 @@ def regrid_weighted_curvilinear_to_rectilinear(src_cube, weights, grid_cube):
...
         raise ValueError(msg.format(sx.name(), sy.name()))
 
-    if sx.ndim != 2:
-        msg = 'The source cube x ({!r}) and y ({!r}) coordinates must ' \
-            'be 2D auxiliary coordinates.'
-        raise ValueError(msg.format(sx.name(), sy.name()))
+#    if sx.ndim != 2:
+#        msg = 'The source cube x ({!r}) and y ({!r}) coordinates must ' \
+#            'be 2D auxiliary coordinates.'
+#        raise ValueError(msg.format(sx.name(), sy.name()))
```
dkillick: As above.

lib/iris/experimental/regrid.py
```diff
@@ -827,43 +874,59 @@ def regrid_weighted_curvilinear_to_rectilinear(src_cube, weights, grid_cube):
...
-        msg = 'The source cube and target grid cube must have the same ' \
-            'coordinate system.'
+    if src_cube.coord_dims(sx) != src_cube.coord_dims(sy):
+        msg = 'The source cube x ({!r}) and y ({!r}) coordinates must ' \
+            'map onto the same cube dimensions.'
+        raise ValueError(msg.format(sx.name(), sy.name()))
+
+    if weights is None:
+        weights = np.ones(sx.shape)
+    if weights.shape != sx.shape:
```
dkillick: This requires that `sx` and `sy` have the same shape, which is fine in the multiple '1D' horizontal coords case. But what about when we have '2D' horizontal coords - can we now only regrid square cubes?

lib/iris/experimental/regrid.py
```diff
@@ -827,43 +874,59 @@ def regrid_weighted_curvilinear_to_rectilinear(src_cube, weights, grid_cube):
...
+        msg = ('Provided weights must have the same shape as the X and Y '
+               'coordinates.')
+        raise ValueError(msg)
+
+    if tx.coord_system != ty.coord_system:
+        msg = 'The target grid cube x ({!r}) and y ({!r}) coordinates must ' \
+            'have the same coordinate system.'
+        raise ValueError(msg.format(tx.name(), ty.name()))
+
+    if sx.coord_system is None or tx.coord_system is None:
```
dkillick: I wonder if it would be worth putting these two coord system checks up just after the checks on source coord system? It would make the flow of these checks more logical (and I almost added a comment asking where the source coord system checks were :stuck_out_tongue_winking_eye: )

lib/iris/experimental/regrid.py
```diff
@@ -879,29 +942,51 @@ def regrid_weighted_curvilinear_to_rectilinear(src_cube, weights, grid_cube):
...
-    # cube x coordinate range.
-    min_sx, min_tx = np.min(sx.points), np.min(tx.points)
+    # Transform source X and Y points into the target coord-system, if needed.
+    if sx.coord_system != tx.coord_system:
+        src_crs = sx.coord_system.as_cartopy_projection()
+        tgt_crs = tx.coord_system.as_cartopy_projection()
+        sx_points, sy_points = _transform_xy_arrays(
+            src_crs, sx_points, sy_points, tgt_crs)
+    #
+    # PP-TODO: how does this work with scaled units ??
```
dkillick: I think given the tight schedule for this change that if we don't need to solve it here we don't solve it here. You could always leave this comment in place but remove the commented code that's below.

lib/iris/experimental/regrid.py
```diff
@@ -958,16 +1043,9 @@ def _regrid_indices(cells, depth, points):
...
-        # Calculate the valid M offsets, accounting for the source cube mask.
-        mask = ~src_cube.data.mask.flatten()
-        cols = np.where((y_indices >= 0) & (y_indices < ty_depth) &
-                        (x_indices >= 0) & (x_indices < tx_depth) &
-                        mask)[0]
-    else:
-        # Calculate the valid M offsets.
-        cols = np.where((y_indices >= 0) & (y_indices < ty_depth) &
-                        (x_indices >= 0) & (x_indices < tx_depth))[0]
+    # Calculate the valid M offsets.
```
dkillick: How did you manage to throw away the mask processing?

lib/iris/experimental/regrid.py
```diff
@@ -994,14 +1072,78 @@ def _regrid_indices(cells, depth, points):
...
     rows = np.nonzero(sum_weights)
 
-    # Calculate the numerator of the weighted mean (M, 1).
-    numerator = sparse_matrix * src_cube.data.reshape(-1, 1)
+    # NOTE: when source points are masked, this 'sum_weights' is possibly
+    # incorrect and needs re-calculating.  Likewise 'rows' may cover target
+    # cells which happen to get no data.  This is dealt with by adjusting as
+    # required in the '__perform' function, below.
+
+    regrid_info = (sparse_matrix, sum_weights, rows, grid_cube)
```
dkillick: Um. Where is this sparse matrix defined? I can't find any reference to it before this line other than in the line you've removed...

lib/iris/experimental/regrid.py
```diff
@@ -994,14 +1072,78 @@ def _regrid_indices(cells, depth, points):
...
-    # Calculate the weighted mean payload.
+    Perform the prepared regrid calculation on a single 2d cube.
+
+    """
+    sparse_matrix, sum_weights, rows, grid_cube = regrid_info
+
+    # Calculate the numerator of the weighted mean (M, 1).
+    is_masked = ma.isMaskedArray(src_cube.data)
+    if not is_masked:
+        data = src_cube.data
```
dkillick: Whoa, well this isn't lazy :wink:

lib/iris/experimental/regrid.py
```diff
@@ -994,14 +1072,78 @@ def _regrid_indices(cells, depth, points):
...
+    else:
+        # Use raw data array
+        data = src_cube.data.data
+        # Check if there are any masked source points to take account of.
+        is_masked = np.ma.is_masked(src_cube.data)
```
dkillick: This sort of sounds like the same check as in L1095 - I guess the difference is that here you're looking for masked points rather than a masked array? You _could_ simplify this by expecting that if you have a masked array you will always have masked points, but I'm not too fussed if you don't do this.

lib/iris/experimental/regrid.py
pp-mo: No, we can for instance have x.shape == y.shape == (200, 300).

lib/iris/experimental/regrid.py
```diff
@@ -994,14 +1072,78 @@ def _regrid_indices(cells, depth, points):
...
+            # N.B. it is more efficient to use the original once-calculated
+            # sparse matrix, but in this case we can't.
+            # Hopefully, this post-multiplying by the validities is less costly
+            # than repeating the whole sparse calculation.
+            valid_src_cells = ~mask.flat[:]
+            src_cell_validity_factors = sparse_diags(
+                np.array(valid_src_cells, dtype=int),
+                0)
+            valid_weights = sparse_matrix * src_cell_validity_factors
+            sum_weights = valid_weights.sum(axis=1).getA()
```
dkillick: Whoa. `getA()` could almost do with a comment of its very own...

lib/iris/experimental/regrid.py
pp-mo: Because it combined the which-src-in-what-tgt-cell calculation, which depends only on the grid, with the source-data-missing-points, which can vary between calls.
So the sparse matrix computed like this is specific to the source field, and can't be re-used on another 2d field.
The code in the "second part" (__perform) routine now handles the effects of the source .

lib/iris/experimental/regrid.py
pp-mo: line 1065

lib/iris/experimental/regrid.py
```diff
@@ -1105,9 +1252,25 @@ def __call__(self, src):
                              'source grid as this regridder.')
 
         # Call the regridder function.
-        res = regrid_weighted_curvilinear_to_rectilinear(src, self.weights,
-                                                         self._target_cube)
-        return res
+        # This includes repeating over any non-XY dimensions, because the
+        # underlying routine does not support this.
+        # FOR NOW: we will use cube.slices and merge to achieve this,
+        # though that is not a terribly efficient method ...
```
dkillick: But it will do for now :smile: We could always look into duplicating the behaviour of the "big" regridders for this functionality at a later date.

lib/iris/experimental/regrid.py
```diff
@@ -1105,9 +1252,25 @@ def __call__(self, src):
...
+        # TODO: create a template result cube and paste data slices into it,
+        # which would be more efficient.
+        result_slices = iris.cube.CubeList([])
+        for slice_cube in src.slices(sx):
```
dkillick: Does slicing on `sx` implicitly expected the source cube is '1D'?

lib/iris/experimental/regrid.py
```diff
@@ -994,14 +1072,78 @@ def _regrid_indices(cells, depth, points):
...
 
-    # Calculate the numerator of the weighted mean (M, 1).
-    numerator = sparse_matrix * src_cube.data.reshape(-1, 1)
+    # NOTE: when source points are masked, this 'sum_weights' is possibly
+    # incorrect and needs re-calculating.  Likewise 'rows' may cover target
+    # cells which happen to get no data.  This is dealt with by adjusting as
+    # required in the '__perform' function, below.
+
+    regrid_info = (sparse_matrix, sum_weights, rows, grid_cube)
+    return regrid_info
```
dkillick: Ooh, if it will fit into 80 characters you could just `return (sparse_matrix, sum_weights, rows, grid_cube)` rather than defining the throw-away variable `regrid_info`...

lib/iris/tests/unit/experimental/regrid/test__CurvilinearRegridder.py
```diff
@@ -35,76 +40,152 @@
 
 class Test___init__(tests.IrisTest):
     def setUp(self):
-        self.ok = lat_lon_cube()
+        self.src_grid = lat_lon_cube()
```
dkillick: :+1: for `ok` --> `src_grid`

lib/iris/experimental/regrid.py
pp-mo: TBH I just followed the existing code for this.
I think it is clear if you properly understand the sparse matrix provisions, but that's tricky because it is very poorly explained in scipy (!!)
I **think** that sparse getA() is like biggus ndarray() -- i.e. it is only then that you get a *real* array.

The key problem with understanding this code as I received it is, IMHO, that it is not too clear which things are sparse and which ordinary matrices.  
So, I thought of trying to improve on that by changing variables, but I thought that then the effect of my changes on top of the existing code would just be even harder to follow...

lib/iris/experimental/regrid.py
pp-mo: Very much a problem for another day I think.

I consulted @marqh on this, and he thought it quite possible that we just haven't considered this in other places in the Iris code where we transform points with Cartopy.
Cartopy itself has no concept of units :  I think it basically assumes either degrees or metres for everything, on the principle that 'metres' is a defined concept on a map surface.  The problem is, it publishes no obvious record of **which** projections are in degrees.
It might be a good idea for every Cartopy projection to somehow specify the coordinate units for each projection, then we could close the loop on this uncertainty.

lib/iris/tests/unit/experimental/regrid/test__CurvilinearRegridder.py
```diff
@@ -35,76 +40,152 @@
...
+        y_coord_2d = AuxCoord(yy,
+                              standard_name=y_coord.standard_name,
+                              units=y_coord.units,
+                              coord_system=y_coord.coord_system)
+        self.src_grid.add_aux_coord(x_coord_2d, (0, 1))
+        self.src_grid.add_aux_coord(y_coord_2d, (0, 1))
+        self.weights = np.ones(self.src_grid.shape, self.src_grid.dtype)
+        # Define an actual, dummy cube for the internal partial result, so we
+        # can do a cubelist merge on it, which is too complicated to mock out.
+        self.mock_slice_result = Cube([1])
```
dkillick: A little thing - with the `mock` library flying about in these tests you could rename this cube to `fake_slice_result` to avoid overloading the work 'mock'...

lib/iris/tests/unit/experimental/regrid/test__CurvilinearRegridder.py
```diff
@@ -35,76 +40,152 @@
...
-        clr.assert_called_once_with(src_grid, self.weights, target_grid)
-        self.assertIs(result, mock.sentinel.regridded)
+    def test_no_weights(self):
+        # Check we can use the regridder without weights.
+        src_grid = self.src_grid
+        target_grid = self.tgt_grid
+        regridder = Regridder(src_grid, target_grid)
+        # Note: for now, patch the internal partial result to a "real" cube,
+        # so we can do a cubelist merge, which is too complicated to mock out.
+        mock_slice_result = Cube([1])
```
dkillick: Can you not get this from `self`? And rename it while you're at it? :wink:

lib/iris/tests/unit/experimental/regrid/test__CurvilinearRegridder.py
```diff
@@ -35,76 +40,152 @@
...
+        target_grid = self.tgt_grid
         regridder = Regridder(src_grid, target_grid, self.weights)
-        src = self.ok.copy()
-        src.rename('SRC')
-        with mock.patch(self.func,
-                        return_value=mock.sentinel.regridded) as clr:
-            result = regridder(src)
+        # Note: for now, patch the internal partial result to a "real" cube,
+        # so we can do a cubelist merge, which is too complicated to mock out.
+        mock_slice_result = Cube([1])
```
dkillick: Again with the rename, (and maybe also getting it from `self`?)

lib/iris/experimental/regrid.py
pp-mo: We could use that simplification, but I was a bit concerned about the efficiency of re-doing the sparse calculation, so I thought it best to always do the minimum necessary.

lib/iris/tests/unit/experimental/regrid/test__CurvilinearRegridder.py
```diff
@@ -35,76 +40,152 @@
...
 
-        clr.assert_called_once_with(src, self.weights, target_grid)
-        self.assertIs(result, mock.sentinel.regridded)
+    def test_caching(self):
+        # Check that it calculates regrid info just once, and re-uses it in
+        # subsequent calls.
+        src_grid = self.src_grid
+        target_grid = self.tgt_grid
+        regridder = Regridder(src_grid, target_grid, self.weights)
+        mock_slice_result = Cube([1])
```
dkillick: And again.

lib/iris/tests/unit/experimental/regrid/test__CurvilinearRegridder.py
```diff
@@ -35,76 +40,152 @@
...
-        src = self.ok.copy()
-        src.rename('SRC')
-        with mock.patch(self.func,
-                        return_value=mock.sentinel.regridded) as clr:
-            result = regridder(src)
+        # Note: for now, patch the internal partial result to a "real" cube,
+        # so we can do a cubelist merge, which is too complicated to mock out.
+        mock_slice_result = Cube([1])
+        # Provide a "different" cube for the actual regrid.
+        different_src_cube = self.src_grid.copy()
```
dkillick: Does just copying the source cube make it different enough from the original source cube to test this behaviour?

lib/iris/tests/unit/experimental/regrid/test__CurvilinearRegridder.py
```diff
@@ -35,76 +40,152 @@
...
-        clr.assert_called_once_with(src, self.weights, target_grid)
-        self.assertIs(result, mock.sentinel.regridded)
+    def test_caching(self):
+        # Check that it calculates regrid info just once, and re-uses it in
+        # subsequent calls.
+        src_grid = self.src_grid
+        target_grid = self.tgt_grid
+        regridder = Regridder(src_grid, target_grid, self.weights)
+        mock_slice_result = Cube([1])
+        different_src_cube = self.src_grid.copy()
```
dkillick: And again again.

lib/iris/tests/unit/experimental/regrid/test__CurvilinearRegridder.py
```diff
@@ -35,76 +40,152 @@
...
-        self.regridder = Regridder(self.ok, self.ok, weights)
+        self.src_grid = global_pp()
+        y = self.src_grid.coord('latitude')
+        x = self.src_grid.coord('longitude')
+        self.src_grid.remove_coord('latitude')
+        self.src_grid.remove_coord('longitude')
+        self.src_grid.add_aux_coord(y, 0)
+        self.src_grid.add_aux_coord(x, 1)
+        weights = np.ones(self.src_grid.shape, self.src_grid.dtype)
+        self.regridder = Regridder(self.src_grid, self.src_grid, weights)
```
dkillick: Why does this regridder require weights?

lib/iris/tests/unit/experimental/regrid/test__CurvilinearRegridder.py
```diff
@@ -113,8 +194,111 @@ def test_bad_src_type(self):
...
+        result = regridder(src_cube)
+        self.assertEqual(result.coord('z'), src_cube.coord('z'))
+        self.assertEqual(result.coord('extra_scalar_coord'),
+                         src_cube.coord('extra_scalar_coord'))
+        self.assertEqual(result.coord('longitude'),
+                         grid_cube.coord('longitude'))
+        self.assertEqual(result.coord('latitude'),
+                         grid_cube.coord('latitude'))
+        self.assertMaskedArrayAlmostEqual(result.data, expected_result)
+        t_dbg = 0
```
dkillick: ?

lib/iris/tests/unit/experimental/regrid/test_regrid_weighted_curvilinear_to_rectilinear.py
```diff
@@ -34,81 +34,109 @@
...
 import iris.coords
+from iris.coords import AuxCoord, DimCoord
+from iris.coord_systems import GeogCS, LambertConformal
+from iris.fileformats.pp import EARTH_RADIUS
 import iris.cube
 from iris.experimental.regrid \
     import regrid_weighted_curvilinear_to_rectilinear as regrid
 
 
+_PLAIN_LATLON_CS = GeogCS(EARTH_RADIUS)
```
dkillick: This probably needn't be private...

lib/iris/tests/unit/experimental/regrid/test_regrid_weighted_curvilinear_to_rectilinear.py
```diff
@@ -149,64 +177,103 @@ def test_aligned_src_x(self):
...
+        self.src.add_aux_coord(self.src_x_positive, (0, 1))
+        # ... given that, we expect exactly the same 'normal' result.
+        data = np.array([0,
+                         self._weighted_mean([3]),
+                         self._weighted_mean([7, 8]),
+                         self._weighted_mean([9, 10, 11])]).reshape(2, 2)
         expected = self._expected_cube(data)
         self.assertEqual(result, expected)
-        mask = np.array([[True, True], [True, False]])
+        mask = np.array([[True, False], [False, False]])
```
dkillick: Why has this answer changed?

lib/iris/tests/unit/experimental/regrid/test_regrid_weighted_curvilinear_to_rectilinear.py
```diff
@@ -149,64 +177,103 @@ def test_aligned_src_x(self):
...
         result = regrid(self.src, self.weights, self.grid)
-        data = np.array([0,
-                         self._weighted_mean([3]),
-                         self._weighted_mean([7, 8]),
-                         self._weighted_mean([9, 10, 11])]).reshape(2, 2)
+        data = np.array([0, 0, 0, self._weighted_mean([9, 10])]).reshape(2, 2)
         expected = self._expected_cube(data)
         self.assertEqual(result, expected)
-        mask = np.array([[True, False], [False, False]])
+        mask = np.array([[True, True], [True, False]])
```
dkillick: Again.

lib/iris/experimental/regrid.py
pp-mo: No way it can be, we're about to use it all.

lib/iris/experimental/regrid.py
pp-mo: Dead right.  
In fact, I think it could be a useful 'general Iris tidy' to rationalise how this is done in a bunch of different places.  Using a common code structure, helper routines or whatever.

lib/iris/tests/unit/experimental/regrid/test_regrid_weighted_curvilinear_to_rectilinear.py
pp-mo: It hasn't : diff has compared it with code from a different test :-(

lib/iris/experimental/regrid.py
pp-mo: I'd rather not as I'm seeing "regrid_info" as a black box : Odd context returned by one routine, to be passed unchanged to the other.
Just to prove it, I have actually changed what was in there several times during development !

lib/iris/experimental/regrid.py
pp-mo: No, X and Y have the "whole grid shape" (explained above), so each slice does the same.

lib/iris/tests/unit/experimental/regrid/test__CurvilinearRegridder.py
pp-mo: No, but the subsequent rename should do it.
We just want to check that the 'second part' routine operates on the given slice rather that the original 'sample' source cube used to create the regridder.

lib/iris/tests/unit/experimental/regrid/test__CurvilinearRegridder.py
pp-mo: It doesn't any more : just legacy code from when it did.
I will remove.

lib/iris/tests/unit/experimental/regrid/test_regrid_weighted_curvilinear_to_rectilinear.py
dkillick: For reference, this is a diff weirdness too :expressionless:

dkillick:
@pp-mo great! I reckon a squash would be good, so if I can't squash on here I will be knocking :facepunch: on your door :door: for a squash...

dkillick:
Yaay nD regridding :grinning: Good work @pp-mo

PR #2084, opened 2016-07-19
Iris grib switch

pp-mo:
Closes #2019

PR #2085, opened 2016-07-22
Refactor grib tests

pp-mo:
Preparing tests for the switch to iris_grib :
A sort of preliminary to #2084

They will still have to work without it (as they do here), but we will shortly switch over + stop testing that.

What remains is to actually enable iris_grib (i.e. in Travis testing also).
There will still be additional changes to make that work, but this should make it less painful.

NOTE: the diffs are awkward to follow because I moved some code in the tests, so 
  a) you can't so easily see what changed between old tests/test_grib_load and new tests/integration/test_grib_load
  b) some results in test/results/visual_tests are renamed also (checked: none of these have changed)


marqh:
all this seems sensible to me, i have no material suggestions, only pernickety comments which don't add noticeable value

on on @pp-mo

PR #2087, opened 2016-07-25
Use iris grib

### LINKED ISSUE #2019: Deprecate iris.FUTURE.strict_grib_load
This is redundant thanks to iris-grib.

pp-mo:
Includes several minor tweaks to get around non-working tests,
plus various modified results (new loader behaves differently) which I'm hoping will all be "ok".

Closes #2019


pp-mo:
We still have a problem here.

I have this working with all tests on my machine with iris_grib 0.9 and ecmwf_grib 1.16.
In Travis, however, the main full test (python2.7) is still picking up ecmwf_grib version 1.14, not 1.16 for some reason.

The 'conda list' command in the Travis build output shows that it installed ecmwf_grib version 1.14.2, build "np18py27_0".
However, the channel also lists "1.16.0-np18py27_0", which you would think should take precedence.
A simple search with "conda search ecmwf_grib" shows the later version.

A previous effort had some channel internal error though, so _just possibly_ there is something odd going on with builds generally.

@marqh can you shed any light ?

pp-mo:
We got this to work by re-building iris_grib with modified dependencies.
That solution is pretty nasty + we do need to fix it later ..  See https://github.com/SciTools/conda-recipes-scitools/issues/185

marqh:
this looks great @pp-mo, excellent work

PR #2089, opened 2016-07-26
Fix test for biggus 0.14.

pp-mo:
Tiny fix needed as private element got moved.

### DIFF
--- a/lib/iris/tests/unit/cube/test_Cube__operators.py
+++ b/lib/iris/tests/unit/cube/test_Cube__operators.py
@@ -24,6 +24,7 @@ import iris.tests as tests
 import numpy as np
 import numpy.ma as ma
 import biggus
+from biggus._init import _Elementwise as Elementwise
 
 
 class Test_Lazy_Maths(tests.IrisTest):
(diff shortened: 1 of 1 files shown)


lib/iris/tests/unit/cube/test_Cube__operators.py
```diff
@@ -24,6 +24,7 @@
 import numpy as np
 import numpy.ma as ma
 import biggus
+from biggus._init import _Elementwise as Elementwise
```
marqh: i do not like importing a private class definition and changing its name to remove the leading underscore

marqh:
thanks @pp-mo

QuLogic:
Does this not mean the minimum dependency needs to be updated?

pp-mo:
> Does this not mean the minimum dependency needs to be updated?

No, this is not an API change :  The _Elementwise class was, and remains, a private element.

QuLogic:
But there is no `biggus._init` in Biggus 0.13, though I guess it _is_ only a test...

ajdawson:
You could use a try/except to avoid this, e.g.:

```python
try:
    from biggus._init import _Elementwise
except ImportError:
    _Elementwise = biggus._Elementwise
```

PR #2094, opened 2016-08-01
Remove restriction on whitespace in netCDF variable names

djkirkham:
The netCDF conventions state:

> Variable, dimension and attribute names should begin with a letter and be composed of letters, digits, and underscores. 

However, the netCDF4 Python library does not enforce this convention. Is there any reason Iris should too?

I have a user with a netCDF file with a variable containing whitespace that can't be used in Iris due to this restriction. I imagine there are others who have encountered the issue as well.

So, what are the arguments against relaxing the restriction?

### DIFF
--- a/lib/iris/_cube_coord_common.py
+++ b/lib/iris/_cube_coord_common.py
@@ -142,13 +142,9 @@ class CFVariableMixin(object):
 
     @var_name.setter
     def var_name(self, name):
-        if name is not None:
-            if not name:
-                raise ValueError('An empty string is not a valid CF variable '
-                                 'name.')
-            elif set(name).intersection(string.whitespace):
-                raise ValueError('{!r} is not a valid CF variable name because'
-                                 ' it contains whitespace.'.format(name))
+        if name is not None and not name:
+            raise ValueError('An empty string is not a valid CF variable '
+                             'name.')
         self._var_name = name
 
     @property


dkillick:
I don't think we should relax this. If the NetCDF conventions state that variable names may not include spaces, then Iris should not allow variable names that contain spaces. In this case (awkwardly) it's the NetCDF python library that is wrong, not Iris...

ajdawson:
Is there room for compromise where we allow names with white space but replace with underscores, following the principal of being lenient on load but strict on save?

dkillick:
@ajdawson that might be a nice (and well-principled) solution to this :+1: We could (optionally!) raise a warning that such a replacement is happening, too.

djkirkham:
Can I suggest we allow loading with no modification, but raise an error on save? I can see things getting a bit fiddly when you try to deal with the corner cases where replacing whitespace with underscores produces an existing variable name.

This relates to the discussion on [this pull request](https://github.com/SciTools/iris/pull/2027) about whether to allow non-compliant values in the data at all.

ajdawson:
Can we currently assign a name with spaces once data are loaded? If so then I'm in favour of this not causing any error on load, since one could have generated the same dataset manually. If not then we might want to think about sanitizing on input.

djkirkham:
> Can we currently assign a name with spaces once data are loaded?

No, it gives the same error as when loading

djkirkham:
I've closed this because I think trying to sanitise the inputs will be two fiddly and not worth the time

PR #2095, opened 2016-08-01
Issues with NetCDF files

juanmcloaiza:
- NetCDF files were not correctly read, throwing the following error:
 

`"TypeError: ufunc did not contain a loop with signature matching types dtype(<U6) dtype(<U6) dtype(<U6)"
`

Apparently the function in question was trying to multiply two strings. I corrected this issue converting the strings to ints or floats.
 
 - Every unit was invalid while reading NetCDF files. 

I corrected this issue by adding a function "sanitize units". The function takes a look at the string containing the units, converts it to lower case and then decides which units they are among a variety of possibilities (finite and rather small). Work still needs to be done here for every new missing unit found.


ajdawson:
Hi @juanmcloaiza. I'm going to close this PR since it has so many commits not relevant to the issue at hand (added unintentionally I'm sure, since you made the PR from your master branch which you then worked hard on by the looks of it!).

If you want to make a new PR with only the commits that you are proposing be merged into iris, we'd be happy to take a look. Contributions are always welcome.

juanmcloaiza:
Hi @ajdawson, Thanks for writing. Actually the problems I addressed were just present when using python 3.4.5, now I downgraded to python 2.7.6 and it seems to work fine out of the box. However I find it strange that no one else has reported similar issues, which makes me wonder whether no one is using python 3 or I am actually missing something...

ajdawson:
It would help if we could see the ncdump header for a netcdf file that causes these problems. We do run tests on Python 3 and there are certainly many users (myself included). Perhaps there is some issue with the specific form of the file you are loading that triggers different behaviour, and if so this would be worth investigation.

PR #2091, opened 2016-07-27
Name format files in National grid coordinates

SusanJL:
This change correctly maps Name III format field files and Version 2 format files containing National grid coordinates onto cubes with National Grid coordinates.

PR #2097, opened 2016-08-02
NameIII format files and Version 2 format files containing National g…

SusanJL:
…rid coordinates are now correctly mapped into cubes with National Grid coordinates


marqh:
that's great, thank you @SusanJL 

i'm merging this and closing #2091

PR #2101, opened 2016-08-04
Adding a number of missing "What's new" entries

agbutteriss:
Since the 1.10 release is coming up, I was asked by @marqh (offline) to go through the pull requests associated with this milestone and write "What's new" entries for those which were missing them.
Right now all of the commits are separate, since it will help in reviewing what actually belongs there (I might have been over-zealous in recording documentation changes).
The relevant pull requests are included in the message for the commit that added the relevant "What's new" entry, and it should be one commit per PR.

I've gone for using the PR start date as the date in the file name, since that seems consistent with previous entries.

Since I wasn't involved in writing any of the updates I'm summarising, I would not be surprised if I went wrong with some of them, so if anyone notices any mistakes/omissions please do point it out.


agbutteriss:
Wait, hang on, wrong base.

PR #2108, opened 2016-08-09
Change default value for check_aux_coords argument of concatenate_cube()

djkirkham:
To maintain compatibility with previous versions of Iris, this PR changes the default value of the `check_aux_coords` argument of `CubeList.concatenate_cube()` from `True` to `False`

PR #2109, opened 2016-08-09
Change default value for check_aux_coords argument of concatenate_cube()

djkirkham:
To maintain compatibility with previous versions of Iris, this PR changes the default value of the `check_aux_coords` argument of `CubeList.concatenate_cube()` from `True` to `False`

### DIFF
--- a/lib/iris/cube.py
+++ b/lib/iris/cube.py
@@ -469,7 +469,7 @@ class CubeList(list):
 
         return merged_cubes
 
-    def concatenate_cube(self, check_aux_coords=True):
+    def concatenate_cube(self, check_aux_coords=False):
         """
         Return the concatenated contents of the :class:`CubeList` as a single
         :class:`Cube`.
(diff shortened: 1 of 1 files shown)


dkillick:
Having [said that this change should be targetted at v1.10.x](https://github.com/SciTools/iris/pull/2108#issuecomment-238508137), it's not so clear that it should be! We're going to do some further investigation and make a decision once we're more informed.

djkirkham:
OK. I've emailed the user who requested it for more information on his use-case, so I'll let you know when I hear back.

marqh:
This was introduced in #1908 

It is my view that this check is always valid, and that the kwarg is there for potential performance issue work arounds.

I am concerned that without this check the resulting metadata may be wrong

It would be very helpful to see a dataset where setting this flag off enables concatenation with consistent resulting metadata.
I'd like to understand such a case before supporting such a change.

djkirkham:
The user has fixed his code and is now (relatively) happy for default value to be True. So unless there are complaints from other users there's no need to make the change.

Issue #2104, opened 2016-08-05
Making compiling the "What's new" easier


agbutteriss:
From working on the "What's New" page, I've noticed that the style of individual "What's new" entries isn't always consistent, and there isn't that clear an indication of what style, if any, they should follow.

While the [relevant section](http://scitools.org.uk/iris/docs/latest/developers_guide/documenting/whats_new_contributions.html) of the developer's guide does mention that a "what's new" entry is ideally "written as a single concise bullet point", it doesn't clarify that the entry should start with the bullet point (going off how they compile), or that the normal character-per-line limits don't apply. The page also doesn't mention what mark-up syntax is used, so it'd be perfectly reasonable for a new developer to just write a single line in plain-text (indeed, for all they know that's what they _should_ do). This means that some of the entries need to be manually changed for the "What's New" page to look right, which obviously takes a non-0 amount of time.

To help with ensuring that "What's New" entries are in the form we want, it seems like it might be worth adding some sort of representative example entry to the page, ideally one that takes advantage of Sphinx's features (e.g. linking to a specific class within the documentation). I'm not sure what style that should be in though (e.g. is the "fixed a bug where this function did a broken thing" style preferable to "this function no longer does this broken thing" for "Bug Fix" entries or vice versa? This isn't necessarily important in and of itself, but if one is definitely preferred it'd be better for the example to use it). I imagine it might also be possible to add some sort of test for compliance with some of the rules for the entries themselves(e.g. start with '\* '). 
Is there some sort of pressing reason that this hasn't been done already that I've missed?

It might also be worth making it clearer in the documentation that (from what I can gather) "What's new" entries are a thing that _should_ be written, not just _could_ (which is the sense I got from just reading the guide). Right now the [Pull Request checklist](http://scitools.org.uk/iris/docs/latest/developers_guide/pulls.html) doesn't even mention them, for instance. Adding something like "If your change is non-trivial, have you added a "What's New" entry?" could be useful, since ultimately the entry needs to get written at some point anyway.

**Edit:** Now I've reread it, there is a passing mention in the "Adding contribution files" section that the entry must be written in markdown, which means the statement that it doesn't make an indication is inaccurate. However, it making an _incorrect_ one (the what's new page is written in reStructuredText, and for some relevant things the syntax is fairly different) isn't exactly better, and it does need to be more clearly highlighted (after being corrected), probably in the Writing a Contribution section.

tkknight:
#3769 primed to improve the what's new process and documentation.

PR #2111, opened 2016-08-09
Minor changes to the pull request checklist

agbutteriss:
I've added a line about including a what's new entry (see #2104), and removed the stuff about iris-code-generators at @pp-mo's suggestion, since it's obsolete.


agbutteriss:
Now I'm looking at it, it seems like it could be worth including "have you added any dependencies to conda-requirements.txt/the install page" as well.

pp-mo:
> @agbutteriss Now I'm looking at it, it seems like it could be worth including "have you added any dependencies to conda-requirements.txt/the install page" as well. 

Definitely worth it :+1:

docs/iris/src/developers_guide/pulls.rst
```diff
@@ -62,20 +63,20 @@ The Iris Check List
 
  * ``make doctest``, ``make extest``  in ``./docs/iris``.
```
agbutteriss: It seems like ``make extest`` isn't actually a thing, but before I remove it does anyone know what, if anything, replaces it and if it's worth including here?

ajdawson: `extest` is a target in the documentation Makefile that runs tests on the example code (code that produces gallery examples).

agbutteriss: Ah, I was in iris/docs/src which has its own Makefile. Thanks.

Issue #2105, opened 2016-08-08
Clarify deprecation of iris.FUTURE.strict_grib_load


pp-mo:
See : https://github.com/SciTools/iris/issues/1948#issuecomment-237171470

pp-mo:
See #2112 
Addresses the issue, but is it too much fuss ?

marqh:
closed by #2112

PR #2112, opened 2016-08-11
Add per-option explanatory string to warning for deprecated FUTURE options.

pp-mo:
In response to #2105 

I'm not really committed to this, let's see who likes it / not.

### DIFF
--- a/lib/iris/__init__.py
+++ b/lib/iris/__init__.py
@@ -215,13 +215,17 @@ class Future(threading.local):
                           self.strict_grib_load, self.netcdf_no_unlimited,
                           self.clip_latitudes)
 
-    deprecated_options = ['strict_grib_load']
+    deprecated_options = {
+        'strict_grib_load': ('This is because "iris.fileformats.grib" is now '
+                             'deprecated :  Please install the "iris_grib" '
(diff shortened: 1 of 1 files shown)


rhattersley:
I'm +0 ... using a mapping seems a little premature since there's only one entry, but it's not a big deal. It's certainly better than not having a verbose message. :smile:

pp-mo:
> a little premature since there's only one entry

There will be a whole bunch of them in v2.0, of course, so I thought I'd just tackle it now.

rhattersley:
> ... I thought I'd just tackle it now

Sounds like an epitaph. :stuck_out_tongue_winking_eye:

marqh:
this looks fine to me; i'd be happy for this to be a pattern which we can use as we require it

marqh:
however, this PR is targeting master, whilst within the 1.10 milestone

if this is a1.10 change, please may you target the v1.10.x branch @pp-mo

QuLogic:
> @pp-mo changed the base branch to `SciTools:v1.10.x` from `SciTools:master`

Whoa, hold on, did GitHub finally fix this?

PR #2115, opened 2016-08-12
Defer 'nc_time_axis' import, making dependency optional.

### LINKED ISSUE #2113: Make nc_time_axis an optional dependency
For a library with such a specific purpose, it is a surprise to find that `nc_time_axis` is now a first-class Iris dependency. At the very least `nc_time_axis` should be only a deferred load (i.e. only when necessary) within `iris.plot`, but even this will still mean that we cannot build the documentation without nc_time_axis. This clearly is not a desirable situation to be in.

An ideal solution to this would be that `nc_time_axis` behaves like `iris_grib`; that is that it is a plugin module for Iris that relies on Iris, rather than being a library that Iris relies on.

pp-mo:
Hopefully closes #2113

### DIFF
--- a/lib/iris/plot.py
+++ b/lib/iris/plot.py
@@ -37,7 +37,6 @@ import matplotlib.transforms as mpl_transforms
 import matplotlib.pyplot as plt
 import matplotlib.ticker as mpl_ticker
 from mpl_toolkits.axes_grid.anchored_artists import AnchoredText
-import nc_time_axis
 import netcdftime
 import numpy as np
 import numpy.ma as ma
@@ -45,6 +44,7 @@ import numpy.ma as ma
 import iris.cube
 import iris.analysis.cartography as cartography
 import iris.coords
+from iris.exceptions import IrisError
 # Importing iris.palette to register the brewer palettes.
 import iris.palette
 
@@ -412,6 +412,16 @@ def _fixup_dates(coord, values):
         if coord.units.calendar == 'gregorian':
             r = [datetime.datetime(*date) for date in dates]
         else:
+            try:
+                import nc_time_axis
+            except ImportError:
+                msg = ('Cannot plot against time in a non-gregorian '
+                       'calendar, because "nc_time_axis" is not available :  '
+                       'Install the package from '
+                       'https://github.com/SciTools/nc-time-axis to enable '
+                       'this usage.')
+                raise IrisError(msg)
+
             r = [nc_time_axis.CalendarDateTime(
                  netcdftime.datetime(*date), coord.units.calendar)
                  for date in dates]
--- a/conda-requirements.txt
+++ b/conda-requirements.txt
@@ -5,7 +5,6 @@
 biggus>=0.13.0
 cartopy
 matplotlib=1.3.1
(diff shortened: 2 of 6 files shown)


lib/iris/plot.py
```diff
@@ -37,14 +37,14 @@
 import matplotlib.pyplot as plt
 import matplotlib.ticker as mpl_ticker
 from mpl_toolkits.axes_grid.anchored_artists import AnchoredText
-import nc_time_axis
 import netcdftime
```
dkillick: You could do the same with this import too: it also was added by #2067.

dkillick:
Thanks @pp-mo! Do we also want to remove nc_time_axis from the conda requirements file? Or is that one dependency removal too far?

lib/iris/plot.py
```diff
@@ -412,6 +412,16 @@ def _fixup_dates(coord, values):
         if coord.units.calendar == 'gregorian':
             r = [datetime.datetime(*date) for date in dates]
         else:
+            try:
+                import nc_time_axis
+            except ImportError:
```
dkillick: Of course, the problem with raising an error is that the failing examples will still fail, just with this error message instead...

pp-mo:
> failing examples will still fail

I think I've sorted this out now...
As nc_time_axis is an _optional_ dependency, the tests should run without it.
They now do (and before was not absent from the 'minimal' test set).

Meanwhile, the examples can still require it, as they are only run with the full dependencies.

lib/iris/plot.py
pp-mo: But I think it's not worth it, as netcdf is a required dependency, ***not*** optional
(and it's just part of that)

Plus, it has nothing to do with the issue in hand.

lib/iris/plot.py
```diff
@@ -412,6 +412,16 @@ def _fixup_dates(coord, values):
...
+                msg = ('Cannot plot against time in a non-gregorian '
+                       'calendar, because "nc_time_axis" is not available :  '
+                       'Install the package from '
+                       'https://github.com/SciTools/nc-time-axis to enable '
+                       'this usage.')
+                raise IrisError(msg)
```
dkillick: What was the purpose of raising this as the slightly esoteric `IrisError`?

marqh:
I think that this looks ready to merge.

The use of the IrisError makes sense to me, as Iris is unable to function in this case, due to a dependency problem.  Unless there's an established pattern for this kind of error, I'm content with this one

@pp-mo anything outstanding from your p.o.v.?

marqh:
@pp-mo 
this is a 1.10 issue, please target the v1.10.x branch, rather than master?

marqh:
Hi @pp-mo 
please may you remove the reference to nc_time_axis from 
`minimal-conda-requirements`
as part of this PR

thank you

pp-mo:
> @marqh please may you remove the reference to nc_time_axis from minimal-conda-requirements
Some confusion here? : this was already done

marqh:
> Some confusion here? : this was already done

so I see, my mistake
thank you

PR #2117, opened 2016-08-12
DOC: usage example for nearest_neighbour_indices

dkillick:
Iris 1.10 deprecated `iris.analysis.interpolate.nearest_neighbour_indices` in favour of `iris.coords.Coord.nearest_neighbour_index`, which was a nice step forward in deprecating the interpolate module. The docs for the interpolate function contained a usage example, however, that was not ported to the docs for the new function. This PR addresses that oversight.


marqh:
looks good to me

Issue #2116, opened 2016-08-12
iris.analysis.interpolate failure


djkirkham:
A user is reporting that some library code based on Iris is failing in  v1.10 now that iris.analysis.interpolate has been deprecated. The code in question has the following form:

```
import iris.analysis
...
iris.analysis.interpolate.linear(...)
```

I know this is incorrect usage (thanks @dkillick), but given that it works in v1.9.2, is there a case for making sure it works in v1.10?

dkillick:
There is, it's...

```python
from iris.analysis.interpolate import linear

...
linear(...)
```

:wink: 

Sometimes moving forwards means breaking stuff. Typically this should be done in a major version, but I think we can get away with it in this case, given that this was never intentional behaviour on Iris' part, and should not have been treated by end users as such.

djkirkham:
This puts the user in question in a bit of a bind. He can't control which version of Iris is used, so his library will suddenly break for all his users until he can get a fix out.

I would normally agree with you that we shouldn't worry about this sort of thing, and if the code in question were, say, using a private variable I wouldn't be in favour of keeping it working. But this case is kind of a Python gotcha. I didn't even know it was incorrect usage myself.

ajdawson:
>This puts the user in question in a bit of a bind. He can't control which version of Iris is used, so his library will suddenly break for all his users until he can get a fix out.

This is an unfortunate scenario for the user in question, but this does happen. The fix for the library in question seems very simple, and this is exactly what a bugfix release is for.

Since we've deprecated this module we shouldn't be importing it into the `iris.analysis` namespace for backwards compatibility, after all this will break end user code in the next release when the interpolate module disappears.

ajdawson:
There are also bigger problems on the horizon for this user, of course... The library in question will need to be updated for compatibility with iris 2.0 when there is no `iris.analysis.interpolate` module at all. Now might be the time to guide them towards using the `interpolate` and/or `regrid` methods on the cube to ensure they don't have similar breakage when their end users upgrade to iris 2.0.

dkillick:
> Now might be the time to guide them towards using the `interpolate` and/or `regrid` methods on the cube

:+1:

djkirkham:
I can see I'm fighting a losing battle here, but my suggestion is only to keep the user's code working in the same way `import iris.analysis.interpolate` will still work, and with the same deprecation warning. Both will break in the next release, but at least the user would have some warning.

ajdawson:
[...]

PR #2119, opened 2016-08-17
Reinstate implicit import of iris.analysis.interpolate with iris.analysis

djkirkham:
Fixes issue #2116

The fix is pretty messy. Perhaps @pp-mo or someone else can suggest a better way?


pp-mo:
> Perhaps @pp-mo or someone else can suggest a better way?

I'm proposing : #2121
It's really **not** any less messy, but it retains the behaviour that importing 'iris.analysis.interpolate' emits a deprecation warning.

marqh:
i think this is now obsolete.
@djkirkham please close this if so

PR #2121, opened 2016-08-22
Import of 'iris.analysis' provides 'iris.analysis.interpolate' again.

pp-mo:
My attempt to improve on @djkirkham  #2119.
TBD: which is better ?

### DIFF
--- a/lib/iris/analysis/__init__.py
+++ b/lib/iris/analysis/__init__.py
@@ -1,4 +1,4 @@
-# (C) British Crown Copyright 2010 - 2015, Met Office
+# (C) British Crown Copyright 2010 - 2016, Met Office
 #
 # This file is part of Iris.
 #
@@ -64,7 +64,6 @@ from iris.analysis._interpolation import (EXTRAPOLATION_MODES,
 from iris.analysis._regrid import RectilinearRegridder
 import iris.coords
 from iris.exceptions import LazyAggregatorError
-
 
 __all__ = ('COUNT', 'GMEAN', 'HMEAN', 'MAX', 'MEAN', 'MEDIAN', 'MIN',
            'PEAK', 'PERCENTILE', 'PROPORTION', 'RMS', 'STD_DEV', 'SUM',
@@ -2327,3 +2326,7 @@ class Nearest(object):
         """
         return RectilinearRegridder(src_grid, target_grid, 'nearest',
                                     self.extrapolation_mode)
+
+# Import "iris.analysis.interpolate" to replicate older automatic imports.
+# NOTE: do this at end, as otherwise its import of 'Linear' will fail.
+from . import _interpolate_backdoor as interpolate
--- a/lib/iris/analysis/_interpolate_backdoor.py
+++ b/lib/iris/analysis/_interpolate_backdoor.py
@@ -1,0 +1,139 @@
+# (C) British Crown Copyright 2010 - 2016, Met Office
+#
+# This file is part of Iris.
+#
+# Iris is free software: you can redistribute it and/or modify it under
+# the terms of the GNU Lesser General Public License as published by the
+# Free Software Foundation, either version 3 of the License, or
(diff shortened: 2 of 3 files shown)


djkirkham:
This is better. I didn't realise I'd removed the deprecation warning when import iris.analysis.interpolate directly.

dkillick:
I have to query the betterness of this... having to introduce an entire private module just to support functionality that's being deprecated seems to be a severe retrograde step. Surely there is a better, more simple way to do this??

djkirkham:
Well, it's better than mine because it works properly and mine didn't.

Really it's just renaming an existing module, not introducing a new one. And it'll be gone in the next release anyway.

pp-mo:
> Really it's just renaming an existing module, not introducing a new one.

\<RANT>
After a lot of code refactoring recently I have become heartily sick of git diff missing the "humanly obvious" when you move or rename things :  In these cases, a reviewer sometimes has no good way of seeing what you did, even when it was relatively simple.
I can understand that if you reorder or shuffle things within or between files it can't be expected to cope, but I'm fed up with it failing to spot simple file renames.
If you *just* rename a file it can cope, also if changes are really minimal.  But even that only works when the difference is restricted to that single commit -- if you include other changes, it is likely to get hopelessly confused again.  It seems that the difference algorithm can not make use of individual commits to get more clues, so even when the changes are comprehensible across each individual commit, they will often fail to be once you squash.
The worst case is if I rename A as OLD_A, and also provide a new A which really looks totally different.  Usually it will fail to make sense of that, and insist that (a) OLD_A is a new file with no history, i.e. unrelated to the earlier code, and (b) the new A is derived from the old one with a huge number of changes.
\</RANT>

I have just discovered the command-line "git diff" flags -M and -C flags, which do seem to help with this.  For example in this case...
```
itpp@eld238: /net/home/h05/itpp/git/iris/iris_main > git diff --stat=165,100 HEAD~1..HEAD 
 lib/iris/analysis/__init__.py              |    7 ++++-
 lib/iris/analysis/_interpolate_backdoor.py |  139 +++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++
 lib/iris/analysis/interpolate.py           |  108 +++++----------------------------------------------------------------------------------
 3 files changed, 151 insertions(+), 103 deletions(-)
```
This diff says that '_interpolate_backdoor' is a new file, unrelated to any previous one.

BUT...
You can add "-M -C" to do extra "copy + move" check (but oddly, you can't say "-MC"), 
which gives :
```
itpp@eld238: /net/home/h05/itpp/git/iris/iris_main > git diff -M -C --stat=165,100 HEAD~1..HEAD
 lib/iris/analysis/__init__.py                                  |    7 ++++-
 lib/iris/analysis/{interpolate.py => _interpolate_backdoor.py} |   21 +++++++++--------
 lib/iris/analysis/interpolate.py                               |  108 ++++++--------------------------------------------------------------------------------------
 3 files changed, 23 insertions(+), 113 deletions(-)
itpp@eld238: /net/home/h05/itpp/git/iris/iris_main > 
```
Now it can see that "_interpolate_backdoor.py" is not a brand new file, but a slightly modified version of the old "interpolate.py".
It does however still insist on regarding the new 'interpolate' as being derived from the old one, though to be fair the boilerplate header could be responsible for that.

Now, it would be really great if you could apply this in github diffs, but I don't think that is possible.

**Does anyone know of other better ways ??**

marqh:
this looks like a sensible implementation to me.  Attempting to keep the imports working as before is a helpful part of the deprecation process and it does not look like a particularly large change, apart from in githubDiff's eyes, which can be somewhat clouded at times.

I'm minded to merge unless there are further comments

marqh:
@pp-mo
this is a 1.10 issue, please target the v1.10.x branch, rather than master?

PR #2128, opened 2016-09-05
Fix issues where +=, -= were being used on arrays with incompatible dtypes

djkirkham:
In Numpy1.10, you cannot use in-place operators on arrays where the RHS cannot be upcast to the LHS, e.g., `int += float`

### DIFF
--- a/lib/iris/experimental/regrid.py
+++ b/lib/iris/experimental/regrid.py
@@ -953,10 +953,10 @@ def _regrid_weighted_curvilinear_to_rectilinear__prepare(
         min_sx, min_tx = np.min(sx.points), np.min(tx.points)
         if min_sx < 0 and min_tx >= 0:
             indices = np.where(sx_points < 0)
-            sx_points[indices] += modulus
+            sx_points[indices] = sx_points[indices] + modulus
         elif min_sx >= 0 and min_tx < 0:
             indices = np.where(sx_points > (modulus / 2))
-            sx_points[indices] -= modulus
+            sx_points[indices] = sx_points[indices] - modulus
 
     # Create target grid cube x and y cell boundaries.
     tx_depth, ty_depth = tx.points.size, ty.points.size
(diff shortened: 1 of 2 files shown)


lib/iris/tests/unit/analysis/test_WPERCENTILE.py
```diff
@@ -1,4 +1,4 @@
-# (C) British Crown Copyright 2015, Met Office
+# (C) British Crown Copyright 2015-2016, Met Office
```
dkillick: This needs to be formatted as 2015 - 2016 (note the spaces surrounding the dash -) as the current formatting is causing some of the license header tests to barf.

djkirkham: ok

dkillick:
> In Numpy1.10, you cannot use in-place operators on arrays where the RHS cannot be upcast to the LHS

@djkirkham do you happen to have a reference for this assertion?

djkirkham:
@dkillick There's this section of the user guide. https://docs.scipy.org/doc/numpy-dev/numpy-user.pdf#page=13&zoom=auto,-266,577

lib/iris/experimental/regrid.py
```diff
@@ -953,10 +953,10 @@ def _src_align_and_flatten(coord):
         min_sx, min_tx = np.min(sx.points), np.min(tx.points)
         if min_sx < 0 and min_tx >= 0:
             indices = np.where(sx_points < 0)
-            sx_points[indices] += modulus
+            sx_points[indices] = sx_points[indices] + modulus
```
dkillick: In the cases where this change fires we are now going to change (i.e. upcast) the dtype of `sx_points`, yes? Is this desirable behaviour, even if it is unavoidable?

Alternatively, was the previous behaviour even correct? Or was it upcasting anyway and just not raising an exception when it did so?

djkirkham: Yes, it will upcast. Previously, int the case where the dtype of the array was int and `modulus` was a float, it would simply round `modulus` down to the nearest integer, which is clearly wrong.

djkirkham:
Sorry this still isn't working...

dkillick:
@djkirkham I agree that this fixes the errors, thanks!

QuLogic:
Can this be backported to v1.10.x?

PR #2130, opened 2016-09-06
tests for util.new_axis added pending bugfix

corinnebosley:
iris.util.new_axis() does not work for a MaskedConstant.  This is now more of an issue since the change to make iris.util.new_axis() handle lazy data, because it now turns any single-point masked array into a MaskedConstant and then doesn't add a new axis to the mask, only to the data.  

For now I have only written tests for this PR; I have not yet added the fix because I don't know what it is.


corinnebosley:
Have found tests for this function in alternate location, so have relocated them and added them to PR#2143: https://github.com/SciTools/iris/pull/2143

PR #2134, opened 2016-09-08
Fix datetime rounding errors.

djkirkham:
Since the `netcdftime.datetime` object now has microsecond precision, some tests are failing due to rounding errors.

This change fixes the errors by introducing a function `num2date_to_nearest_second`, which attempts to replicate the old behaviour of `netcdftime.num2date` (although it accepts a `cf_units.Unit` rather than _unit string_ and _calendar_ arguments.)

I say 'attempts' because when `num2date` is given a value in seconds representing a time with a half second, it will round it up or down arbitrarily, and it is apparently impossible to replicate this behaviour with the new version. The new function will always round up.

### DIFF
--- a/lib/iris/util.py
+++ b/lib/iris/util.py
@@ -1612,7 +1612,7 @@ def _num2date_to_nearest_second(time_value, units):
     except AttributeError:
         microsecond = 0
     if has_half_second or microsecond > 0:
-        if has_half_second or round(microsecond, -6) == 1000000:
+        if has_half_second or microsecond >= 500000:
             seconds = cf_units.Unit('second')
             second_frac = seconds.convert(0.75, units.utime().units)
             time_value += second_frac
--- a/lib/iris/util.py
+++ b/lib/iris/util.py
@@ -1589,3 +1589,35 @@ def demote_dim_coord_to_aux_coord(cube, name_or_coord):
     cube.remove_coord(dim_coord)
 
     cube.add_aux_coord(dim_coord, coord_dim)
+
+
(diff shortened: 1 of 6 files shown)


lib/iris/util.py
```diff
@@ -1589,3 +1589,35 @@ def demote_dim_coord_to_aux_coord(cube, name_or_coord):
     cube.remove_coord(dim_coord)
 
     cube.add_aux_coord(dim_coord, coord_dim)
+
+
+def _num2date_to_nearest_second(time_value, units):
```
marqh: I think that this function may prove to be useful for users experiencing issues with netcdf4 times

I think that it is worth considering making it public and adapting the first few comment lines as the docstring.

I would likely leave the `note that this behaviour...` as a developer comment and not include it in teh docstring in this case

marqh:
this looks like a sensible and well crafted change.  It deals with a set of known issues within netcdf4-python and should make Iris less fragile with respect to versions of netcdf4

the open question for me is the public/private nature of the `iris.util` function

djkirkham:
Perhaps it could be add to the cf_units api, possibly as an optional argument to `Unit.num2date`

marqh:
> Perhaps it could be added to the cf_units api, possibly as an optional argument to Unit.num2date

I like this apporach, I think that num <> date is already in scope for cf_units

@bjlittle @pelson would you support this approach?
Is it useful for @djkirkham to propose such a change to cf_units for us to see and him to test?

djkirkham:
The change has been made to cf_units so this PR is no longer needed.

PR #2135, opened 2016-09-08
Graphic Tests

marqh:
I have:
* decreased the check_graphic tolerance to 0.1
* removed 3 tests which are visually reasonable but which now fail with tolerance > 1
* updated two example gallery images, which are failing at tol=0.1 
* fixed the `iris.util._is_circular` function so that it correctly handles -ve diffs using np.abs 
 * added unit tests to check _is_circular behaviour
 * updated the CML for the two GRIB load tests, which should be circular, according to the plot images stored (RMS Value: 9.683)
  * these plots now pass with an RMS < 0.1


lib/iris/tests/test_plot.py
```diff
@@ -323,6 +323,12 @@ def setUp(self):
             'Temperature')
         self.draw_method = qplt.scatter
 
+    def test_cube_coord(self):
+        pass
+
+    def test_coord_coord(self):
+        pass
```
dkillick: This is a bit odd... It looks like we could either do with some content for these tests, or for them to be removed.

marqh: this is an issue of test specialization.  this test class is a specialization of a parent class, which implements a range of tests.  
Some of these fail, some of these pass.

options:
 1. remove all the tests, including passing ones
 1. specialize these tests in this class and skip them (implemented)

would an extra developer comment stating that this is a skip pattern for the sub-class be sufficient in this case?

marqh: > would an extra developer comment stating that this is a skip pattern for the sub-class be sufficient in this case?

(see last commit)

dkillick:
@marqh it would be good to get an idea of...

* why the coriolis frequency image has been changed despite the fact that old and new appear identical, and
* why the orca projection 2 image has changed so subtly.

marqh:
@dkillick I am not sure why such subtle changes in these 2 images occur.  I have tried to mimick travis' environment as closely as I can to create these images, perhaps previous contributors had not been so careful, used different versions of cartopy, proj4, etc.

The images are pretty close to the previous ones though

bjlittle:
@marqh @dkillick Could the subtle graphic shift be related to the change in `_is_circular` ?

marqh:
> @marqh @dkillick Could the subtle graphic shift be related to the change in _is_circular ?

I believe not, one cube is built from scratch, one is loaded from a CF netCDF file, so neither should make use of that utility at all

QuLogic:
The Coriolis example is a small change in the outline of Seychelles or the Maldives, maybe due to new data. The Orca projection example is a change in scaling, which I think may have originated in Matplotlib somewhere.

marqh:
obsoleted by #2164 and #2162

PR #2139, opened 2016-09-12
Moveplottests

marqh:
remove graphic tests in favour of numerical tests for testing other parts of the code base

assertCML (where already in use) is left as sufficient test coverage

otherwise, data operations are tested using statistics of the data array


marqh:
interestingly, given my aim to reduce version dependency, there are 4 numerical check failures, compared to a local platform (where all these pass) which tries to mimick the travis environment

lib/iris/tests/experimental/regrid/test_regrid_area_weighted_rectilinear_src_and_grid.py
```diff
@@ -495,39 +466,35 @@ def test_scalar_source_cube(self):
...
-    @tests.skip_plot
     def test_global_data_reduce_res(self):
         src = iris.tests.stock.global_pp()
         src.coord('latitude').guess_bounds()
         src.coord('longitude').guess_bounds()
         dest = _resampled_grid(src, 0.4, 0.3)
         res = regrid_area_weighted(src, dest)
-        qplt.pcolormesh(res)
-        self.check_graphic()
+        self.assertArrayAlmostEqual(res.data.mean(), 280.137698)
```
djkirkham: Here and elsewhere you're using `assertArrayAlmostEqual` for scalars.

djkirkham: Here and elsewhere you're using `assertArrayAlmostEqual` for scalars.

djkirkham: Out of interest, why have you used 'AlmostEqual', rather than 'Equal'?

marqh: > Out of interest, why have you used 'AlmostEqual', rather than 'Equal'?

because every time someone compares two floats for equality in a test, a fairy dies ;)

dkillick:
:+1: for removing extraneous graphics tests!

Some thoughts on the strategy proposed here...

 * There are still a number of cases of testing for CML equality. This will of course be open to the same issues in many ways as the graphics tests: a small change in the CML leads straight to a test failure.
 * I like the approach of checking for near equality of mean & stdev of the data arrays, but I wonder whether these provide enough information to test equality of the two data arrays?
 * What this of course means is that we need a different approach for equality testing. As has been discussed, for graphics tests we could check that multiple items of the two matplotlib graphics (e.g. the extents of the x and y axes, presence of colorbar, ...) are equal or near-equal.

lib/iris/tests/integration/test_regridding.py
```diff
@@ -32,20 +32,15 @@
...
-@tests.skip_plot
-class TestOSGBToLatLon(tests.GraphicsTest):
+class TestOSGBToLatLon(tests.IrisTest):
     def setUp(self):
         path = tests.get_data_path(
             ('NIMROD', 'uk2km', 'WO0000000003452',
              '201007020900_u1096_ng_ey00_visibility0180_screen_2km'))
         self.src = iris.load_cube(path)[0]
-        self.src.data = self.src.data.astype(np.float32)
+        self.src.data = self.src.data.astype(np.float64)
```
bjlittle: @marqh Is this up-casting to avoid `numpy 1.8` issues?

lib/iris/tests/integration/test_regridding.py
```diff
@@ -58,23 +53,27 @@ def setUp(self):
...
+        self.assertArrayAlmostEqual(result.data.std(), 5612.657155)
 
 
 @tests.skip_data
-@tests.skip_plot
-class TestGlobalSubsample(tests.GraphicsTest):
+class TestGlobalSubsample(tests.IrisTest):
     def setUp(self):
         self.src = global_pp()
+        _ = self.src.data
```
bjlittle: `self.src.data` on it's own will do the job, no need to assign

marqh: it will, but it can make for messy output during debugging
I tend to prefer to assign to _ as a way to throw this away (so to speak)

lib/iris/tests/experimental/regrid/test_regrid_area_weighted_rectilinear_src_and_grid.py
djkirkham: So it's a precision issue? In that case, maybe a hash of data would be better to test. NumPy arrays can't be hashed, but the underyling buffer (`ndarray.data`) can be, if the `writeable` flag of the `ndarray` is  set to `False`.

I'm wondering if testing the mean/standard deviation means we wouldn't catch bugs which only changed the data slightly. It's probably nothing to worry about, but just a thought.

lib/iris/tests/experimental/regrid/test_regrid_area_weighted_rectilinear_src_and_grid.py
```diff
@@ -367,25 +361,11 @@ def test_regrid_to_higher_res(self):
...
-        qplt.pcolormesh(res)
-        self.check_graphic()
-        plt.contourf(res.coord('grid_longitude').points,
-                     res.coord('grid_latitude').points,
-                     res.coord('altitude').points)
-        self.check_graphic()
-        plt.contourf(res.coord('grid_longitude').points,
-                     res.coord('grid_latitude').points,
-                     res.coord('surface_altitude').points)
-        self.check_graphic()
```
bjlittle: @marqh Nice. If we have `assertCML` coverage then there's no need for `check_graphic` coverage as well.

marqh:
@bjlittle I understand, this seems like a useful pattern.

I have implemented `assertArrayShapeStats` for `IrisTest`

dkillick:
@bjlittle @marqh like it! :+1:

bjlittle:
@marqh LGTM :+1:

QuLogic:
> near equality of mean & stdev of the data arrays, but I wonder whether these provide enough information to test equality of the two data arrays?

Anscombe's quartet says no.

PR #2140, opened 2016-09-12
Correct "precipitation_amount" to grib mapping

lbdreyer:
According to the documentation:

||Number ||Parameter||
|8 |Total precipitation*** |
|49 |Total water precipitation*** |


lbdreyer:
Should I rather be changing this in metarelate, @marqh?

marqh:
> Should I rather be changing this in metarelate, @marqh?

yes, this should be changed in metarelate.  It should not really be changed here at all, this whole area of the code base is essentially frozen, awaiting deprecation.

this should be a change to iris-grib, via metarelate, in my view

it's not immediately clear to me that this change is valid, but that discussion belongs in metarelate

lbdreyer:
> It should not really be changed here at all, this whole area of the code base is essentially frozen, awaiting deprecation

Ah you're right! I completely forgot about iris_grib.

PR #2142, opened 2016-09-14
Remove AssertString tests from test_pandas.py, replace with in-code tests

djkirkham:
Replaces uses of assertString in Pandas tests, one of which was failing due to a slight formatting change, with tests on the data.


marqh:
i think these test are maintaining test coverage whilst being more explicit about what is being tested in the test code

They also remove a fragility, exposed by a change to the pandas version and the way they choose to implement 
`str(iris.pandas.as_series(cube))`
and change it in newer version of their library.
I don't think this is should be a test failure

I am very much minded to merge this as is

PR #2144, opened 2016-09-15
Fix unit test failing due to change in cf_units

djkirkham:
A unit test was passing a mock object to `cf_units.date2num`, and failing with the changes in https://github.com/SciTools/cf_units/pull/68. This change ensures that actual `datetime` objects are passed to that function.

The unit test in question doesn't care about the values returned by the function, so the datetimes passed are essentially arbitrary.


djkirkham:
@pelson @dkillick Since there has now been a release of `cf_units`, I suspect that the build will fail unless this PR is merged.

djkirkham:
@bjlittle

dkillick:
Looks good :+1:

PR #2161, opened 2016-10-05
Check graphic hash

marqh:


.travis.yml
```diff
@@ -105,10 +105,10 @@ install:
 
 script:
   - if [[ $TEST_TARGET == 'default' ]]; then
-      python -m iris.tests.runner --default-tests --system-tests --print-failed-images;
+      python -m iris.tests.runner --default-tests --system-tests;
```
pp-mo: I think it could make sense to remove the support for the  "print-failed-images" key from setup.py.

lib/iris/tests/__init__.py
```diff
@@ -695,18 +703,50 @@ def check_graphic(self, tol=_DEFAULT_IMAGE_TOLERANCE):
...
+                    os.makedirs(dname)
+                wmsg = 'Created image for test {}'
+                warnings.warn(wmsg.format(unique_id))
                 shutil.copy2(result_fname, expected_fname)
-
-            err = mcompare.compare_images(expected_fname, result_fname, tol=tol)
+            else:
+                # Cherry-pick the registered expected hashes from the
+                # test case uri/s.
+                expected = []
```
pp-mo: I think a list comprehension would be clearer than init + append here.
Seems an obvious edit of what is here already.

lib/iris/tests/__init__.py
```diff
@@ -695,18 +703,50 @@ def check_graphic(self, tol=_DEFAULT_IMAGE_TOLERANCE):
...
+                for uri in uris:
+                    ehash = os.path.splitext(os.path.basename(uri))[0]
```
pp-mo: I'd still prefer that the url + hash were kept logically separate, and not insist that the filename "is" the hash.
The decision to do that may be enforced elsewhere, but better not to assume / require it here ?

.travis.yml
marqh: for now i'm putting this back in; if it is removed then we can discuss removing the functionality too

marqh:
replaced by #2162

QuLogic:
It wasn't really necessary to open a new PR; just force push on this one.

PR #2162, opened 2016-10-05
Rebase hash

marqh:
new PR to replace #2161 with a cleaner history (perhaps)


lib/iris/tests/__init__.py
```diff
@@ -620,16 +623,21 @@ def _unique_id(self):
...
-        bits = self.id().split('.')[-3:]
+        bits = self.id().split('.')
         if bits[0] == '__main__':
-            file_name = os.path.basename(sys.modules['__main__'].__file__)
+            floc = sys.modules['__main__'].__file__
+            path, file_name = os.path.split(os.path.abspath(floc))
             bits[0] = os.path.splitext(file_name)[0]
+            folder, location = os.path.split(path)
+            bits = [location] + bits
+            while location != 'iris':
```
bjlittle: @marqh :bomb: This won't work for documentation [example_tests](https://github.com/SciTools/iris/pull/2162/files#diff-53068f7ead231d14ec56529c4837ca8eR2), `iris` isn't the only root namespace module in the testing context.

You'll need to account for `example_tests` also being a terminal module.

lib/iris/tests/__init__.py
```diff
@@ -695,18 +704,38 @@ def check_graphic(self, tol=_DEFAULT_IMAGE_TOLERANCE):
...
-                shutil.copy2(result_fname, expected_fname)
+            # XXX: Deal with a new test result i.e. it's not in the repo
 
-            err = mcompare.compare_images(expected_fname, result_fname, tol=tol)
+            # hash the created image using sha1
+            with open(result_fname, 'rb') as res_file:
+                sha1 = hashlib.sha1(res_file.read())
+            if unique_id not in repo:
+                msg = 'Image comparison failed: Created image {} for test {}.'
+                raise ValueError(msg.format(result_fname, unique_id))
```
bjlittle: @marqh You've changed the logic here from the original, which I disagree with.

As it now stands any developer who adds a **new** graphics test, has to manually craft the changes to the `imagerepo.json` and manage the hash renaming of the `png` ... this doesn't scale very well.

I'll propose an alternative approach on your branch as a PR that (I believe) improves this work-flow. We can discuss that further there ... so hang tight for that.

lib/iris/tests/test_image_json.py
```diff
@@ -0,0 +1,93 @@
...
+# importing anything else
+import iris.tests as tests
+
+import codecs
+from collections import deque
+import json
+import logging
+import os
+from six.moves.queue import Queue
+import requests
```
bjlittle: @marqh Move this `import` above `six.moves.queue`, thanks.

lib/iris/tests/test_image_json.py
```diff
@@ -0,0 +1,93 @@
...
+
+class TestImageFile(tests.IrisTest):
+    def test_resolve(self):
+        repo_fname = os.path.join(os.path.dirname(__file__), 'results',
+                                  'imagerepo.json')
+        with open(repo_fname, 'rb') as fi:
+            repo = json.load(codecs.getreader('utf-8')(fi))
+        uris = []
+        for k, v in six.iteritems(repo):
+            uris = uris + v
```
bjlittle: @marqh Drop the usage of `k` ... you don't need it, then use `six.itervalues`

Another neat pattern from `itertools` to flatten a list of lists is ...
```python
from itertools import chain
uris = list(chain.from_iterable(six.itervalues(repo)))
```

lib/iris/tests/__init__.py
bjlittle: @marqh Okay, I'm working on that `rebase_hash` branch PR ... in-bound soon, thanks

lib/iris/tests/__init__.py
marqh: indeed

lib/iris/tests/__init__.py
marqh: I understand your concerns and I am keen to see a work flow for updating lots of images

however, in your previous implementation, the effect was that tests with no checksum to check would pass, for developers and for travis.
This is a more serious problem.  I simplified the logic to allow me to fix this problem, which I deemed critical.
If you are working with respect to this branch, please make sure that an implementation does not allow tests to pass where there is no entry in the imagerepo or where the image checksum comparison fails.

any tooling to help updating must (imho) deal with these as test failures, then help the developer fix them, not enable them to pass

I hope that point of view is clear and coherent

marqh:
#2163 concern about testing stability 

i suggest we move forward, not treat this as a showstopper

lib/iris/tests/__init__.py
bjlittle: @marqh I completely understand your concerns, and I think that there is a suitable compromise to be had here to meet all our needs.

I don't think it would be wise of me to mix this into this PR - let's do that in a separate PR :+1:

marqh:
> #2163 concern about testing stability
> 
> i suggest we move forward, not treat this as a showstopper

@bjlittle it's gone green

nothing to see here ;) :/

bjlittle:
LGTM

bjlittle:
@marqh Not yet ...

bjlittle:
@marqh Now that your happy LGTM

lib/iris/tests/test_image_json.py
```diff
@@ -0,0 +1,92 @@
...
+        Thread.__init__(self)
+        self.daemon = True
+
+    def run(self):
+        while not self.queue.empty():
+            resource = self.queue.get()
+            try:
+                result = requests.head(resource)
+                if (result.status_code == 200 and
+                   resource.startswith('https://scitools.github.io')):
```
QuLogic: This is not aligned correctly.

QuLogic: Also, why not check the `startswith` beforehand and save a request?

lib/iris/tests/test_image_json.py
```diff
@@ -0,0 +1,92 @@
...
+import logging
+import os
+import requests
+from collections import deque
+from itertools import chain
+from six.moves.queue import Queue
+from threading import Thread
+
+# maximum number of threads for multi-threading code
+MAXTHREADS = 128
```
QuLogic: Do you really need to run 128 parallel HEAD requests? Browsers limit themselves to something like 2-3 connections per host; this seems like overkill.

When I run it with 128, I get a bunch of `ConnectionError(MaxRetryError...` in the exception deque and it takes 22.277s. If I cut threads down to _2_, there are no errors and it runs in 16.276s. Running with 1 thread takes 33.248s, 4 threads is 8.292s and 8 threads is 5.938s. The delta is already diminishing there, so I doubt there is any benefit to bumping it all the way to 128.

PR #2165, opened 2016-10-06
removed visual_test images from source tree,

marqh:
see https://github.com/scitools/test-images-scitools/ for details


bjlittle:
@marqh I'm :+1: for removing the **contents** of the `iris/tests/results/visual_tests`, but I'm :-1: on removing the directory itself.

As part of the proposed developer work-flow in #2166, I've used the `iris/tests/results/visual_tests` directory as a staging area for **new** `<hash>.png` files that require to be pushed into [test-images-scitools](https://github.com/SciTools/test-images-scitools). The `idiff.py` tool now consistently updates the `imagerepo.json` store with regards to those new image files.

The advantage of using the known directory `iris/tests/results/visual_tests`, which is under git source code control, is that it clearly highlights to the iris developer that those git **untracked files** exist and require to be dealt with appropriately ... rather than hiding them in "some other" (hidden and) uncontrolled directory.

pp-mo:
> @bjlittle on removing the directory itself...

If you agree with @marqh [comment](https://github.com/SciTools/iris/pull/2166#issuecomment-251938835), which [I do](https://github.com/SciTools/iris/pull/2166#issuecomment-251988702), then completely losing this *does* make sense.

PR #2166, opened 2016-10-06
Refactor idiff for imagerepo.json capability.

bjlittle:
Extend the capability of the iris developer `idiff.py` tool to assist with graphical test differences and updating the `imagerepo.json` store.

Note that, `idiff.py` will consistently update the `imagerepo.json` store if a graphic test image change is accepted **and** place the new `<hash>.png` file into the `iris/tests/resuts/visual_tests` directory in preparation for the developer to add to the  [test-images-scitools](https://github.com/SciTools/test-images-scitools) repo as a new PR

- [x] Resolve linkage with #2165
- [x] Rebase after #2165 
- [x] Refactor

### DIFF
--- a/lib/iris/tests/idiff.py
+++ b/lib/iris/tests/idiff.py
@@ -28,19 +28,20 @@ from six.moves import (filter, input, map, range, zip)  # noqa
 import argparse
 import codecs
 import contextlib
+from glob import glob
 import hashlib
 import json
 import os.path
-import requests
-import shutil
 import sys
 import warnings
 
+from PIL import Image
 import filelock
 import matplotlib.pyplot as plt
 import matplotlib.image as mimg
 import matplotlib.testing.compare as mcompare
 import matplotlib.widgets as mwidget
+import requests
 
 # Force iris.tests to use the ```tkagg``` backend by using the '-d'
 # command-line argument as idiff is an interactive tool that requires a
@@ -50,9 +51,9 @@ import iris.tests
 import iris.util as iutil
 
 
-_HASH_DIR = os.path.join('results', 'visual_tests')
-_POSTFIX_LOCK = os.path.join('results', 'imagerepo.lock')
+_POSTFIX_DIFF = '-failed-diff.png'
 _POSTFIX_JSON = os.path.join('results', 'imagerepo.json')
+_POSTFIX_LOCK = os.path.join('results', 'imagerepo.lock')
 _PREFIX_URI = 'https://scitools.github.io/test-images-scitools/image_files'
 _TIMEOUT = 30
 _TOL = 0
@@ -69,7 +70,8 @@ def temp_png(suffix=''):
         os.remove(fname)
 
 
-def diff_viewer(repo, key, dname, expected_fname, result_fname, diff_fname):
+def diff_viewer(repo, key, repo_fname,
+                expected_fname, result_fname, diff_fname):
     plt.figure(figsize=(14, 12))
     plt.suptitle(os.path.basename(expected_fname))
     ax = plt.subplot(221)
@@ -82,8 +84,10 @@ def diff_viewer(repo, key, dname, expected_fname, result_fname, diff_fname):
     # Determine the new image hash.png name.
     with open(result_fname, 'rb') as fi:
         sha1 = hashlib.sha1(fi.read())
-    hash_fname = os.path.join(dname, _HASH_DIR, sha1.hexdigest()+'.png')
-    uri = os.path.join(_PREFIX_URI, os.path.basename(hash_fname))
+    result_dir = os.path.dirname(result_fname)
+    fname = sha1.hexdigest() + '.png'
+    uri = os.path.join(_PREFIX_URI, fname)
+    hash_fname = os.path.join(result_dir, fname)
 
     def accept(event):
         if uri not in repo[key]:
@@ -91,9 +95,10 @@ def diff_viewer(repo, key, dname, expected_fname, result_fname, diff_fname):
             # associated with the repo key is the oldest, and the last
             # uri is the youngest
             repo[key].append(uri)
-            with open(os.path.join(dname, _POSTFIX_JSON), 'wb') as fo:
+            # Update the image repo.
+            with open(repo_fname, 'wb') as fo:
                 json.dump(repo, fo, indent=4, sort_keys=True)
-            shutil.copy2(result_fname, hash_fname)
+            os.rename(result_fname, hash_fname)
             msg = 'ACCEPTED:  {} -> {}'
             print(msg.format(os.path.basename(result_fname),
                              os.path.basename(hash_fname)))
@@ -101,7 +106,7 @@ def diff_viewer(repo, key, dname, expected_fname, result_fname, diff_fname):
             msg = 'DUPLICATE: {} -> {} (ignored)'
             print(msg.format(os.path.basename(result_fname),
                              os.path.basename(hash_fname)))
-        os.remove(result_fname)
+            os.remove(result_fname)
         os.remove(diff_fname)
         plt.close()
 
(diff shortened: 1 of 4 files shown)


lib/iris/tests/idiff.py
```diff
@@ -25,17 +25,52 @@
...
+# command-line argument as idiff is an interactive tool that requires a
+# gui interface.
+sys.argv.append('-d')
+import iris.tests
+import iris.util as iutil
 
-def diff_viewer(expected_fname, result_fname, diff_fname):
-    plt.figure(figsize=(16, 16))
+
+_HASH_DIR = os.path.join('results', 'visual_tests')
```
marqh: this folder is under source control, In don't think any code should be writing images to a folder under iris source control.  i think the potential for new image files to get added by mistake is too high

could hashed named image files be put elsewhere please?
they'll need to be moved by a developer again anyway

bjlittle:
> i think the potential for new image files to get added by mistake is too high

This is the case with *any* directory. See this [comment](https://github.com/SciTools/iris/pull/2165#issuecomment-251932108) for further detail. So what other **uncontrolled, random directory** would you suggest that the live in?

> they'll need to be moved by a developer again anyway

True. But git sees them as **untracked files**, which is the massive hint to the iris developer that they need to deal with them appropriately.

I think that it's a good work-flow to remove the **contents** of `iris/test/results/visual_tests` to ensure it's empty (perhaps add a readme.txt with a developer note) but keep the actual directory itself.

marqh:
> So what other uncontrolled, random directory would you suggest that the live in?

i think that the functionality already implemented to decide where to write files better fits the bill

https://github.com/SciTools/iris/blob/master/lib/iris/tests/__init__.py#L680

makes some decisions about where to write image_results files.  I think the 'hashed' copies of files which may need to be managed should also live in this folder, which should never be under source control
this folder is also writeable to, even if the iris installation is not, which can occur where testing is occurring on servers and the like

I feel this is a better pattern to use

pp-mo:
> @marqh i think that the functionality already implemented to decide where to write files better fits the bill

I was previously an advocate for putting all outputs into the managed dir, 
but @marqh has now convinced me that will  cause problems in an unwriteable installation.
For which https://github.com/SciTools/iris/blob/master/lib/iris/tests/__init__.py#L683 is a simple fix.
So I'm now cool with that need.

I'm also now agreeing with 
> @marqh the 'hashed' copies of files which may need to be managed should also live in this [[unmanaged]] folder

... and also #2165

pp-mo:
I understand from @marqh that the pair of you agreed that 'adding a graphical test' need not be simple
However, I do note that at present '''check_graphic''' makes a bit of hash of this in that the message produced for "new unrecognised test" is the **same** as for "test result not an expected one".
We just need to remember to sort this out.

pp-mo:
The way I think it should now go ...
 * results are created in '''tests/result_image_comparison'''  (the unmanaged dir, or it's cwd-relative equivalent)
 * results that are already recognised (hashes in repo) are then deleted, while new ones remain.
 * an unrecognised test produces an error
 * an unrecognised result hash produces a different error

lib/iris/tests/idiff.py
```diff
@@ -44,49 +79,114 @@ def diff_viewer(expected_fname, result_fname, diff_fname):
...
-        os.remove(expected_fname)
-        shutil.copy2(result_fname, expected_fname)
+        if uri not in repo[key]:
+            # Ensure to maintain strict time order where the first uri
+            # associated with the repo key is the oldest, and the last
+            # uri is the youngest
+            repo[key].append(uri)
+            with open(os.path.join(dname, _POSTFIX_JSON), 'wb') as fo:
+                json.dump(repo, fo, indent=4, sort_keys=True)
+            shutil.copy2(result_fname, hash_fname)
```
pp-mo: Should possibly rename rather than copy (see general behaviour discussion)

lib/iris/tests/idiff.py
```diff
@@ -44,49 +79,114 @@ def diff_viewer(expected_fname, result_fname, diff_fname):
...
+            print(msg.format(os.path.basename(result_fname),
+                             os.path.basename(hash_fname)))
+        os.remove(result_fname)
         os.remove(diff_fname)
         plt.close()
 
     def reject(event):
-        print('REJECTED: %s' % (os.path.basename(expected_fname), ))
+        if uri not in repo[key]:
+            print('REJECTED:  {}'.format(os.path.basename(result_fname)))
```
pp-mo: I think this branch should *also* remove the result
(i.e. REJECT always removes the result)
see general behaviour discussion

lib/iris/tests/idiff.py
```diff
@@ -44,49 +79,114 @@ def diff_viewer(expected_fname, result_fname, diff_fname):
...
 
-    ax_accept = plt.axes([0.7, 0.05, 0.1, 0.075])
-    ax_reject = plt.axes([0.81, 0.05, 0.1, 0.075])
-    bnext = mwidget.Button(ax_accept, 'Accept change')
-    bnext.on_clicked(accept)
-    bprev = mwidget.Button(ax_reject, 'Reject')
-    bprev.on_clicked(reject)
+    def skip(event):
+        print('SKIPPED:   {}'.format(os.path.basename(result_fname)))
+        os.remove(diff_fname)
```
pp-mo: I think this should leave the result in-place, so we can deal with it later by re-running
(see general behaviour discussion)

lib/iris/tests/idiff.py
```diff
@@ -44,49 +79,114 @@ def diff_viewer(expected_fname, result_fname, diff_fname):
...
+                wmsg = 'Ignoring unregistered test result {!r}.'
+                warnings.warn(wmsg.format(key))
+                continue
+            with temp_png(key) as expected_fname:
+                processed = True
+                resource = requests.get(uri)
+                with open(expected_fname, 'wb') as fo:
+                    fo.write(resource.content)
+                mcompare.compare_images(expected_fname, result_fname, tol=_TOL)
+                diff_fname = result_fname[:-4] + '-failed-diff.png'
```
QuLogic: Why not use `splitext` this time?

lib/iris/tests/idiff.py
```diff
@@ -44,49 +79,114 @@ def diff_viewer(expected_fname, result_fname, diff_fname):
...
+                        default=default,
+                        help=help)
+    help = 'Force "iris.tests" to use the tkagg backend (default: %(default)s)'
+    parser.add_argument('-d',
+                        action='store_true',
+                        default=True,
+                        help=help)
+    args = parser.parse_args()
+    result_dir = args.resultdir
+    if not os.path.isdir(result_dir):
```
QuLogic: You can't just create the directory?

lib/iris/tests/idiff.py
```diff
@@ -25,17 +25,52 @@
 from __future__ import (absolute_import, division, print_function)
 from six.moves import (filter, input, map, range, zip)  # noqa
 
+import argparse
+import codecs
+import contextlib
+import hashlib
+import json
 import os.path
+import requests
```
QuLogic: `requests` is not a builtin library.

lib/iris/tests/idiff.py
```diff
@@ -44,49 +79,114 @@ def diff_viewer(expected_fname, result_fname, diff_fname):
     ax = plt.subplot(223, sharex=ax, sharey=ax)
     ax.imshow(mimg.imread(diff_fname))
 
+    # Determine the new image hash.png name.
+    with open(result_fname, 'rb') as fi:
+        sha1 = hashlib.sha1(fi.read())
+    hash_fname = os.path.join(dname, _HASH_DIR, sha1.hexdigest()+'.png')
+    uri = os.path.join(_PREFIX_URI, os.path.basename(hash_fname))
```
QuLogic: One more redundant join and split.

bjlittle:
This PR needs a rebase, refactor and some TLC ...

bjlittle:
@marqh ping

marqh:
thanks @bjlittle

PR #2167, opened 2016-10-06
is_circular: grib load tests and what's new

marqh:
required update to 
#2164 (already merged :/)

### DIFF
--- a/docs/iris/src/whatsnew/contributions_1.11/bugfix_2016-Oct-06_is_circular.txt
+++ b/docs/iris/src/whatsnew/contributions_1.11/bugfix_2016-Oct-06_is_circular.txt
@@ -1,2 +1,2 @@
-* Fixed a bug where :func:iris.util._is_circular would erroneously return false
+* Fixed a bug where :func:iris.util._is_circular would erroneously return False
   when coordinate values are decreasing.
--- a/docs/iris/src/whatsnew/contributions_1.11/bugfix_2016-Oct-06_is_circular.txt
+++ b/docs/iris/src/whatsnew/contributions_1.11/bugfix_2016-Oct-06_is_circular.txt
@@ -1,2 +1,2 @@
-- Fixed a bug where the :func:`iris.util._is_circular` would return false where 
+* Fixed a bug where the :func:`iris.util._is_circular` would return false where 
   the coordinate values are decreasing.
--- a/docs/iris/src/whatsnew/contributions_1.11/bugfix_2016-Oct-06_is_circular.txt
+++ b/docs/iris/src/whatsnew/contributions_1.11/bugfix_2016-Oct-06_is_circular.txt
@@ -1,2 +1,2 @@
-* Fixed a bug where the :func:`iris.util._is_circular` would return false where 
-  the coordinate values are decreasing.
+* Fixed a bug where :func:iris.util._is_circular would erroneously return false
+  when coordinate values are decreasing.
--- a/docs/iris/src/whatsnew/contributions_1.11/bugfix_2016-Oct-06_is_circular.txt
+++ b/docs/iris/src/whatsnew/contributions_1.11/bugfix_2016-Oct-06_is_circular.txt
@@ -1,0 +1,2 @@
+- Fixed a bug where the :func:`iris.util._is_circular` would return false where 
+  the coordinate values are decreasing.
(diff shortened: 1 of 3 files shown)


docs/iris/src/whatsnew/contributions_1.11/bugfix_2016-Oct-06_is_circular.txt
```diff
@@ -0,0 +1,2 @@
+- Fixed a bug where the :func:`iris.util._is_circular` would return false where
```
ajdawson: Previously we started each contribution with a `*`, should we keep consistency with that? I know `-` is the correct character, but `*` is also allowed, and mixing them in a single list would be a bad idea so we have to pick one.

marqh: i'd rather stick with '-' 

as this is the first entry for this collection, perhaps it can set a precedent?

I believe that the 1.10 contributions mixed the - and * use

ajdawson: Fine, I'd prefer if we tried to stick to the same throughout, but depending on restructured-text backend it is worst case a warning if you mix them, so I guess it doesn't matter as much as I thought.

bjlittle: @marqh LGTM ... but @ajdawson is right, after a quick scan through the last few `whatsnew` all points are marked with `*` ...

... does Sphinx render the html differently with `-` instead of `*` ... I can't tell quickly from documentation, and I can't be bother building the docs to see.  

http://thomas-cokelaer.info/tutorials/sphinx/rest_syntax.html#list-and-bullets doesn't mention the use of `-` 

@marqh could I nudge you to in the direction of using `*` ? Then I'll merge ... I'll even update the dev notes :wink:

ajdawson: It should render fine with either, but if you mix the styles you get a warning from sphinx but the output is likely correct (although in some circumstances the intent may be misinterpreted. I'd advocate using strictly 1 style only.

marqh:
> could I nudge you in the direction of using * ? Then I'll merge ... I'll even update the dev notes in a separate PR

done

ping @bjlittle

docs/iris/src/whatsnew/contributions_1.11/bugfix_2016-Oct-06_is_circular.txt
bjlittle: I vote for `*`

docs/iris/src/whatsnew/contributions_1.11/bugfix_2016-Oct-06_is_circular.txt
```diff
@@ -0,0 +1,2 @@
+* Fixed a bug where the :func:`iris.util._is_circular` would return false where
```
ajdawson: Now that we've dealt with character number 1, we can move onto the rest :wink:  Can it be changed like this:

Fixed a bug where ~~the~~ :func:`iris.util._is_circular` would *erroneously* return false ~~where~~ when ~~the~~ coordinate values are decreasing.

marqh:
updated with new wording, please squish and merge if you are content

marqh:
note: tests passed with the first commit, all further commits are just changes to the what's new wording, so you may choose not to wait for travis, if you so desire

docs/iris/src/whatsnew/contributions_1.11/bugfix_2016-Oct-06_is_circular.txt
```diff
@@ -0,0 +1,2 @@
+* Fixed a bug where :func:iris.util._is_circular would erroneously return False
```
ajdawson: The markup here is no longer correct, it is missing the back-ticks ( I suspect because copy-paste from the Github comment, which turned the back-ticks into an `object`). You actually don't need to capitalise False in this usage unless you are referring explicitly to the Python object `False`...

marqh:
> The markup here is no longer correct, it is missing the back-ticks

sorry, i should have spotted that; hopefully fixed now

bjlittle:
I'm not going to wait for travis, this is safe to merge, as the first commit passed fine on travis, so that's good enough for me.

Thanks @marqh and thanks @ajdawson

PR #2173, opened 2016-10-10
Hash corrections

QuLogic:
This is a followup to my comments on #2162. The number of threads does not need to be so high, and if the hostname is invalid, we should not even bother to make the request.

### DIFF
--- a/lib/iris/tests/test_image_json.py
+++ b/lib/iris/tests/test_image_json.py
@@ -58,8 +58,7 @@ class _ResolveWorkerThread(Thread):
             resource = self.queue.get()
             try:
                 result = requests.head(resource)
-                if (result.status_code == 200 and
-                   resource.startswith('https://scitools.github.io')):
+                if result.status_code == 200:
                     self.deque.append(resource)
                 else:
                     msg = '{} is not resolving correctly.'.format(resource)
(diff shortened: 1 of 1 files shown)


lib/iris/tests/test_image_json.py
```diff
@@ -80,7 +79,12 @@ def test_resolve(self):
         exceptions = deque()
         uri_queue = Queue()
         for uri in uris:
-            uri_queue.put(uri)
+            if uri.startswith('https://scitools.github.io'):
+                uri_queue.put(uri)
+            else:
+                msg = '{} is not a valid resource.'.format(uri)
+                exceptions.append(ValueError(msg))
+                continue
```
marqh: what is the `continue` here for please, I am not convinced it is doing anything active

lib/iris/tests/test_image_json.py
```diff
@@ -35,7 +35,7 @@
 from threading import Thread
 
 # maximum number of threads for multi-threading code
-MAXTHREADS = 128
+MAXTHREADS = 8
```
marqh: this seems sensible, given your analysis of run times

lib/iris/tests/test_image_json.py
```diff
@@ -58,8 +58,7 @@ def run(self):
             resource = self.queue.get()
             try:
                 result = requests.head(resource)
-                if (result.status_code == 200 and
-                   resource.startswith('https://scitools.github.io')):
```
marqh: i agree with this logic change

lib/iris/tests/test_image_json.py
QuLogic: Copy/paste leftover; will remove.

marqh:
neat improvements; thank you @QuLogic

PR #2174, opened 2016-10-10
PEP8 updates

QuLogic:
The `pep8` package was pinned in #1558. To be able to run on the latest packages, this pinning needs to be removed. So these style issues need to be corrected as well.

On a related note, when should we depend on `pycodestyle` instead of `pep8`?

### DIFF
--- a/lib/iris/_constraints.py
+++ b/lib/iris/_constraints.py
@@ -264,11 +264,15 @@ class _CoordConstraint(object):
             desired_values = list(self._coord_thing)
             # A dramatic speedup can be had if we don't have bounds.
             if coord.has_bounds():
-                call_func = lambda cell: cell in desired_values
+                def call_func(cell):
+                    return cell in desired_values
             else:
-                call_func = lambda cell: cell.point in desired_values
+                def call_func(cell):
+                    return cell.point in desired_values
         else:
-            call_func = lambda c: c == self._coord_thing
+            def call_func(c):
+                return c == self._coord_thing
+
             try_quick = (isinstance(coord, iris.coords.DimCoord) and
                          not isinstance(self._coord_thing, iris.coords.Cell))
 
--- a/INSTALL
+++ b/INSTALL
@@ -132,7 +132,7 @@ mock 1.0.1 (http://pypi.python.org/pypi/mock/)
 nose 1.1.2 or later (https://nose.readthedocs.io/en/latest/)
     Python package for software testing. Iris is not compatible with nose2.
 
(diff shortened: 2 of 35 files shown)


marqh:
Hi @QuLogic 

i think this is a positive step, i am in favour

I had not realised that pep8 explicitly pushed against lambda functions in the way this PR suggests

have you considered whether we should adopt this pattern within the user guide? e.g.
http://scitools.org.uk/iris/docs/latest/userguide/loading_iris_cubes.html#constrained-loading
(scroll down a bit, to:
`As with many of the examples later in this documentation, the simple function above can be conveniently written as a lambda function on a single line:`
)

or is the key pep8 suggestion that multi line lamda functions are naff, rather than all lambda functions?

ajdawson:
>is the key pep8 suggestion that multi line lamda functions are naff, rather than all lambda functions?

PEP8 suggests you prefer a `def` statement instead of binding a lambda to an identifier, it doesn't matter what the form of the lambda is, it just shouldn't be bound to a name.

>have you considered whether we should adopt this pattern within the user guide?

The user guide is suggesting ways of doing things to end users, and as such I don't think we need to be strict about following a style document there (it is the end user's decision what style to follow, if any).

lib/iris/_constraints.py
```diff
@@ -264,11 +264,15 @@ def extract(self, cube):
             desired_values = list(self._coord_thing)
             # A dramatic speedup can be had if we don't have bounds.
             if coord.has_bounds():
-                call_func = lambda cell: cell in desired_values
+                def call_func(cell):
```
ajdawson: This could be a one-liner like the original:

```python
def call_func(cell): return cell in desired_values
```

Any strong preferences?

marqh: i find reading over multiple lines clearer than having a def and func on one line, I prefer @QuLogic 's current, I think

marqh:
all the individual code changes seem positive or neutral to me

the overall aim of unpinning is clearly better

I see no reason not to merge

last call for caution.....

marqh:
Hi @QuLogic 

this is great stuff, I am fully on board.

there has been a lot of activity recently, centred on #2124, which has just been merged (woo!)

please may you rebase onto master so that the tests rerun, in case there is any pep8 fallout which we have missed

thank you
mark

marqh:
please may you also add a change to
https://github.com/SciTools/iris/blob/master/INSTALL#L135
removing the *

thank you

marqh:
excellent, many thanks @QuLogic

PR #2185, opened 2016-10-11
Graphics tests docs

pp-mo:
First cut at documenting new graphics test arrangements + workflow.

**NOT yet complete**

### DIFF
--- a/docs/iris/src/developers_guide/graphics_tests.rst
+++ b/docs/iris/src/developers_guide/graphics_tests.rst
@@ -54,8 +54,8 @@ This consists of :
  * storing the hashes of 'known accepted results' for each test in a
    database in the repo (which is actually stored in 
    ``lib/iris/tests/results/imagerepo.json``).
- * storing associated reference images for each hash value in a separate web
-   project, currently in https://github.com/SciTools/test-images-scitools ,
+ * storing associated reference images for each hash value in a separate public
+   repository, currently in https://github.com/SciTools/test-images-scitools ,
    allowing human-eye judgement of 'valid equivalent' results.
  * a new version of the 'iris/tests/idiff.py' assists in comparing proposed
    new 'correct' result images with the existing accepted ones.
--- a/docs/iris/src/developers_guide/graphics_tests.rst
+++ b/docs/iris/src/developers_guide/graphics_tests.rst
@@ -3,4 +3,111 @@
 Graphics tests
 **************
 
-TODO: a full description is pending, will be provided for release 1.12.
+The only practical way of testing plotting functionality is to check actual
+output plots.
+For this, a basic 'graphics test' assertion operation is provided in the method
+:method:`iris.tests.IrisTest.check_graphic` :  This tests plotted output for a
+match against a stored reference.
+A "graphics test" is any test which employs this.
+
+At present (Iris version 1.10), such tests include the testing for modules
+`iris.tests.test_plot` and `iris.tests.test_quickplot`, and also some other
+'legacy' style tests (as described in :ref:`developer_tests`).
+It is conceivable that new 'graphics tests' of this sort can still be added.
+However, as graphics tests are inherently "integration" style rather than true
+unit tests, results can differ with the installed versions of dependent
+libraries (see below), so this is not recommended except where no alternative
+is practical.
+
+Testing actual plot results introduces some significant difficulties :
+ * Graphics tests are inherently 'integration' style tests, so results will
+   often vary with the versions of key dependencies :  Obviously, results will
+   depend on the matplotlib version, but they can also depend on numpy and
+   other installed packages.
+ * Although it seems possible in principle to accommodate 'small' result changes
+   by distinguishing plots which are 'nearly the same' from those which are
+   'significantly different', in practice no *automatic* scheme for this can be
+   perfect :  That is, any calculated tolerance in output matching will allow
+   some changes which a human would judge as a significant error.
+ * Storing a variety of alternative 'acceptable' results as reference images
+   can easily lead to uncontrolled increases in the size of the repository,
+   given multiple independent sources of variation.
+
+
+Graphics Testing Strategy
+=========================
+
+Prior to Iris 1.10, all graphics tests compared against a stored reference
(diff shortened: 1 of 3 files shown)


ajdawson:
I haven't been following this closely but it seems like an interesting tactic for handling image tests across dependency versions and platforms. The one problem is that the hashes of 2 very similar images will be very different, meaning we need exact matches to one of our stored hashes. This seems OK as long as we can generate enough hashes, but I wonder how well this will scale in the future.

It might be valuable to consider using a different hashing algorithm, one where similar images generate hashes that are "close". Luckily such things exist already, called perceptual hashes. A quick search suggests Python implementations of [pHash](http://www.phash.org/docs/design.html) are available (e.g. https://github.com/JohannesBuchner/imagehash). This would allow storing hashes for major changes (e.g. when matplotlib changes something like default colormap) but minor variants around these could be matched by checking the distance between two hashes rather than having to add new reference hashes for each case. Perhaps worth considering.

bjlittle:
Thanks @ajdawson!

Perceptual image hashing has been on our radar and already discussed. See [here](http://nbviewer.jupyter.org/gist/bjlittle/7a8adda9d4dd351bcdf1a9a82f5ca63e) for an initial investigation. It raises the interesting question of how "similar" is a result to an expected result, and is that definition of "similarity" acceptable to suit our testing needs. We thought that we'd first exercise the use of a zero tolerance cryptographic hash approach and assess the impact and practicalities of demanding exact image equivalence rather than "similar" images based on a hamming distance hash threshold.

It's certainly a very interesting problem space. With the new work flow pattern and framework that we now have in place for image testing, I think we're in a much healthier place i.e. it's an enabler that's got us pretty darn close to unpinning mpl within Travis from 1.3 to latest mpl 1.5+, which in turn releases us from the numpy 1.8 shackles, see #2124 . So I believe perceptual image hashing is still a viable option that we can chose to use and it's a question that I think we should consider answering sooner rather than later ...

marqh:
> it's a question that I think we should consider answering sooner rather than later ...

indeed

docs/iris/src/developers_guide/graphics_tests.rst
```diff
@@ -0,0 +1,98 @@
...
+There should be sufficient work-flow detail here to allow an iris developer to:
+    * understand the new check graphic test process
+    * understand the steps to take and tools to use to add a new graphic test
+    * understand the steps to take and tools to use to diagnose and fix an graphic test failure
+
+
+Basic workflow
+==============
+#. a graphics test result has changed, following changes in Iris or
+   dependencies, so a test is failing
```
corinnebosley: I like this point.  It's important to include a trigger or reason for having to do this.

docs/iris/src/developers_guide/graphics_tests.rst
```diff
@@ -0,0 +1,98 @@
...
+#. the hash value of the new result is added into the relevant set of 'valid
+   result hashes' in the image result database file,
+   ``tests/results/imagerepo.json``.
+#. the tests can now be re-run, and the 'new' result should now be accepted.
+#. a pull request is created to add this file into the test-images-scitools
+   repository
+#. a pull request is created in the Iris repository, including the change to
+   the image results database (``tests/results/imagerepo.json``) :
+   This pull request must contain a reference to the matching one in
+   test-images-scitools.
```
corinnebosley: Good stubs.  I agree that enumerating the steps will make it easier to add the necessary detail.

docs/iris/src/developers_guide/pulls.rst
```diff
@@ -16,66 +16,78 @@ is merged.  Before submitting a pull request please consider this list.
...
+  * iris-test-data_ is a github project containing all the data to support the
+    tests.
+  * iris-sample-data_ is a github project containing all the data to support
+    the gallery and examples.
+  * test-images-scitools_ is a github project containing reference plot images
+    to support iris graphics tests : see :ref:`test graphics images`.
+ * If new files are required by tests or code examples, they must be added to
+   the appropriate supporting project via a suitable pull-request.
+   This new 'supporting pull request' should be referenced in the main Iris
+   pull request, and must be accepted and merged before the Iris one can be.
```
corinnebosley: I really like your additions to this document.  Very clear and instructive.

docs/iris/src/developers_guide/tests.rst
```diff
@@ -125,3 +127,19 @@ developers locate relevant tests. It is recommended they are named
...
 `metadata/test_pp_preservation.py`, and not named according to the
 module(s) under test.
+
+
+Graphics tests
+=================
+Certain Iris tests rely on testing plotted results.
+This is required for testing the modules :mod:`iris.plot` and
+:mod:`iris.quickplot`, but is also used for some other legacy and integration
+tests.
```
corinnebosley: There are lots of tests in this sentence.  Are there too many?

bjlittle:
@pp-mo and @corinnebosley If https://github.com/SciTools/iris/pull/2192 is merged, then it may impact this PR.

marqh:
#2192 is now merged and this branch has inherited a conflict

@pp-mo please may you attempt to capture these inputs and hook together @corinnebosley 's comments and update the PR

thank you

corinnebosley:
@pp-mo I did make a branch from this branch and made some changes to the docs in your absence, but I had trouble working out how to make a PR and where to push it to and stuff, so I stopped.

pp-mo:
> [[corinnebosley]] pp-mo I did make a branch from this branch and made some changes to the docs in your absence, but I had trouble working out how to make a PR and where to push it to and stuff, so I stopped.

Those suggestions are here : https://github.com/pp-mo/iris/pull/39

pp-mo:
:-( Messed this up be re-using this branch
but :-) all those comments + suggestions now adopted here.

pp-mo:
Rebased + resolved.
I've now made some response to all the previous comments + suggestions.
Please re-review !

marqh:
this is looking pretty close to good to go to me

and remaining concerns or objections in short order please...

marqh:
thanks @pp-mo

PR #2186, opened 2016-10-12
idiv vs itruediv test fixed

corinnebosley:
This test was dividing an integer array by another integer array and then failing when it found that it has floats in its resulting array.  This only occurs in the Python3 function 'itruediv' (the previously used Python2 function 'idiv' rounded to answer to an integer, so didn't encounter the casting error).

The change here catches the event of itruediv being used, and casts the arrays to float arrays rather than integers so that it can quite happily produce a floating point array as a result.

I have tested this with Python3 and numpy1.10 with success.

### DIFF
--- a/lib/iris/tests/test_basic_maths.py
+++ b/lib/iris/tests/test_basic_maths.py
@@ -1,4 +1,4 @@
-# (C) British Crown Copyright 2010 - 2015, Met Office
+# (C) British Crown Copyright 2010 - 2016, Met Office
 #
 # This file is part of Iris.
 #
--- a/lib/iris/tests/test_basic_maths.py
+++ b/lib/iris/tests/test_basic_maths.py
@@ -638,6 +638,8 @@ class TestLog(tests.IrisTest):
 class TestMaskedArrays(tests.IrisTest):
     ops = (operator.add, operator.sub, operator.mul)
     iops = (operator.iadd, operator.isub, operator.imul)
+    # This is to ensure compatibility for both Python2 (which uses div and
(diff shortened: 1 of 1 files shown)


corinnebosley:
@bjlittle Will you have a little look at this fix for me please?

lib/iris/tests/test_basic_maths.py
```diff
@@ -662,7 +666,15 @@ def test_operator(self):
...
         for test_op in self.iops:
             test_op(self.cube1, self.cube2)
-            test_op(self.data1, self.data2)
+            if test_op == operator.itruediv:
+                # Python3 itruediv requires floats to return floats, numpy1.10:
+                # "TypeError: ufunc 'true_divide' output (typecode 'd') could
+                # not be coerced to provided output parameter (typecode 'l')
+                # according to the casting rule ''same_kind''"
+                test_op(ma.MaskedArray(self.data1, dtype='float'),
+                        ma.MaskedArray(self.data2, dtype='float'))
```
bjlittle: @corinnebosley Looks great. My only suggestion would be to reduce this change to ...

```python
test_op(data1.astype('float'), data2.astype('float'))
```
Otherwise I'm :+1: on this PR :smile:

bjlittle:
@corinnebosley Awesome, thanks!

LGTM!

corinnebosley:
@bjlittle What?

bjlittle:
@corinnebosley Apologies "Let's Get This Merged" :stuck_out_tongue_winking_eye:

dkillick:
I always thought it meant "Looks Good To Me"... :wink:

PR #2189, opened 2016-10-12
Added last compare and codecs writer.

bjlittle:


lib/iris/tests/idiff.py
```diff
@@ -55,6 +56,7 @@
 _POSTFIX_JSON = os.path.join('results', 'imagerepo.json')
 _POSTFIX_LOCK = os.path.join('results', 'imagerepo.lock')
 _PREFIX_URI = 'https://scitools.github.io/test-images-scitools/image_files'
```
marqh: this is only used once, could we have it used in line, not obfuscated in a global please?

lib/iris/tests/idiff.py
```diff
@@ -160,28 +166,45 @@ def step_over_diffs(result_dir):
...
             try:
-                uri = repo[key][0]
+                uri = repo[key][index]
             except KeyError:
                 wmsg = 'Ignoring unregistered test result {!r}.'
                 warnings.warn(wmsg.format(key))
                 continue
             with temp_png(key) as expected_fname:
                 processed = True
                 resource = requests.get(uri)
```
marqh: `requests.head` is a bit faster than `requests.get` as it doesn't download the image file

bjlittle: @marqh But we want the image file ...

lib/iris/tests/idiff.py
bjlittle: Sure

lib/iris/tests/idiff.py
```diff
@@ -55,6 +56,7 @@
 _POSTFIX_JSON = os.path.join('results', 'imagerepo.json')
 _POSTFIX_LOCK = os.path.join('results', 'imagerepo.lock')
 _PREFIX_URI = 'https://scitools.github.io/test-images-scitools/image_files'
+_RESOLVED_STATUS = 200  # HTTP response status code - OK
```
bjlittle: Sure

marqh:
@bjlittle just pep8 to fix please

lib/iris/tests/idiff.py
```diff
@@ -86,7 +86,8 @@ def diff_viewer(repo, key, repo_fname,
         sha1 = hashlib.sha1(fi.read())
     result_dir = os.path.dirname(result_fname)
     fname = sha1.hexdigest() + '.png'
-    uri = os.path.join(_PREFIX_URI, fname)
+    base_uri = 'https://scitools.github.io/test-images-scitools/image_files'
+    uri = os.path.join(base_uri, fname)
```
marqh: I agree. even M$ windows respects the HTTP protocol's use of `/` as a separator in all circumstances.  This should always return a `/` separator, so the use of `os.path.join` is incorrect

bjlittle: Good spot, thanks @QuLogic

bjlittle:
Hmmm I keep needing to re-run some of the pk3 image tests, that are failing on travis and they shouldn't ... I see this as very bad behaviour indeed. Particularly as I've not touched them, which means to me the image testing is unstable and everyone is going to be stung by this ...

marqh:
> Hmmm I keep needing to re-run some of the pk3 image tests (twice now), that are failing on travis and they shouldn't ... I see this as very bad behaviour indeed. Particularly as I've not touched them, which means to me the image testing is somewhat unstable and everyone is going to be stung by this ...

@bjlittle I agree, there is a follow up activity here which we need to make time for.  do we have an issue on github for this? could we have one and link to it from here please

i think we should consider this outside of the 1.11 milestone, we can cut 1.11 with this, but I think it is technical debt which we have created anew, in trying to reduce our overhead.  perhaps the v2 milestone is the appropriate place for this?

bjlittle:
@marqh I've captured this in issue https://github.com/SciTools/iris/issues/2195

bjlittle:
@marqh Agreed. LGTM :smile:

PR #2192, opened 2016-10-13
Check graphic create missing

bjlittle:
The missing part of the new image test work-flow, to allow developers to create missing/new image test expected result files and associated entries in the `imagerepo.json` file.

- [x] Rebase after #2196

### DIFF
--- a/lib/iris/tests/__init__.py
+++ b/lib/iris/tests/__init__.py
@@ -672,6 +672,7 @@ class IrisTest(unittest.TestCase):
         output directory, and the imagerepo.json file being updated.
 
         """
+        dev_mode = os.environ.get('IRIS_TEST_CREATE_MISSING')
         unique_id = self._unique_id()
         repo_fname = os.path.join(os.path.dirname(__file__),
                                   'results', 'imagerepo.json')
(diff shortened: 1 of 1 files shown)


lib/iris/tests/__init__.py
```diff
@@ -733,13 +749,27 @@ def check_graphic(self, tol=None):
...
+        Checks that the image hash for the current matplotlib figure matches
+        the expected image hash for the current test.
+
+        """
+        fname = os.path.join(os.path.dirname(__file__),
+                             'results', 'imagerepo.lock')
+        lock = filelock.FileLock(fname)
+        # The imagerepo.json file is a critical resource, so ensure thread
+        # safe read/write behaviour via platform independent file locking.
+        with lock.acquire(timeout=600):
```
bjlittle: @marqh A suitably high timeout ...

Having the graphics tests fighting over the `imagerepo.lock` doesn't seem to impact the overall running time of the tests in travis ... but it's hard to tell due to the variable loads on travis.

I don't think this is a problem, but we could choose to only acquire a lock if the `IRIS_TEST_CREATE_MISSING` environment variable is set - as this means there will be potential writing to the `imagerepo.json` and this shouldn't be applicable to test runs within travis but only on the desktop of the developer.

bjlittle:
This PR requires some extra work ... in-bound

bjlittle:
@marqh Ping ...

bjlittle:
@marqh LGTM

bjlittle:
@marqh Rebased and good to go ...

marqh:
thanks @bjlittle

PR #2193, opened 2016-10-13
Netcdf packing support v2

jswanljung:
NetCDF has support for variable packing; by specifying attributes scale_factor and add_offset, data can be packed into smaller data types. 16 bit short integers provide more than enough quantization for suitably scaled temperature data, for instance. See:
http://unidata.ucar.edu/software/netcdf/docs/BestPractices.html#bp_Packed-Data-Values 
Iris already supports loading of packed netCDF because it is automatically handled by the netCDF4 module. However, there is currently no way to save data with packing.

This PR implements packing by adding an optional `packing` argument to the netcdf.save function and netcdf.Saver.write method. As described in the docstring:
 
- packing (type or string or dict or list): A numpy integer datatype
    (signed or unsigned) or a string that describes a numpy integer dtype
    (i.e. 'i2', 'short', 'u4') or a dict of packing parameters as described
    below or a list of such types, strings, or dicts. This provides support
    for netCDF data packing as described in
    If this argument is a type (or type string), appropriate values of
    scale_factor and add_offset will be automatically calculated based on
    `cube.data` and possible masking. For masked data, `fill_value` is
    taken from netCDF4.default_fillvals. For more control, pass a dict with
    one or more of the following keys: `dtype` (required), `scale_factor`,
    `add_offset`, and `fill_value`. To save multiple cubes with different
    packing parameters, pass a list of types, strings, dicts, or `None`,
    one for each cube.  Note that automatic calculation of packing
    parameters will trigger loading of lazy data; set them manually using a
    dict to avoid this. The default is `None`, in which case the datatype
    is determined from the cube and no packing will occur.

So for example:
```
# pack into np.int16 with automatic calc of scale_factor and add_offset, default fillval if applicable
 iris.save(cube, 'outfile.nc', packing='int16') 

# pack into np.int16 with manually set scale_factor and add_offset
iris.save(cube, 'outfile.nc', packing=dict(dtype=np.short, scale_factor=0.01,  add_offset=250))

# pack one cube with auto scale, the other manually, and the third not at all (cubelist has 3 cubes)
iris.save(cubelist, 'outfile.nc', packing=['int16', dict(dtype='int16', scale_factor=0.01), None])
```

This PR supercedes a previous attempt (https://github.com/SciTools/iris/pull/2152) which I withdrew because it set packing attributes in the cube. In this version packing is completely decoupled from
the cube (see the previous PR for discussion). It also uses the new pattern for  setting cf attributes (change requested by @marqh).

In the discussion in the previous PR, @ajdawson objected (and I agreed) that the method implemented here for providing different packing specifications for multiple cubes is suboptimal. Ideally we would identify cubes in the list by key, but we were unable to come up with keys that could identify cubes uniquely in a list. There is no requirement that cube names be unique, for instance. Suggestions are welcome.


lib/iris/fileformats/netcdf.py
```diff
@@ -2104,9 +2194,19 @@ def save(cube, filename, netcdf_format='NETCDF4', local_keys=None,
     with Saver(filename, netcdf_format) as sman:
         # Iterate through the cubelist.
         for cube in cubes:
+            if isinstance(packing, list):
```
bjlittle: @jswanljung The user might just use a `tuple` rather than a list, and that'll cause some carnage from this point on ... we could be graceful and deal with that.

lib/iris/fileformats/netcdf.py
```diff
@@ -2104,9 +2194,19 @@ def save(cube, filename, netcdf_format='NETCDF4', local_keys=None,
...
+                if len(packing) > 0:
+                    packspec = packing.pop(0)
+                else:
+                    msg = ('If packing is a list, it must have the '
+                           'same number of elements as the argument to'
+                           'cube.')
+                    raise ValueError(msg)
+            else:
+                packspec = packing
```
bjlittle: @jswanljung For this whole section, if we have a `list` (or a `tuple`) we're popping as we go along ... but it can be the case that the user doesn't have the same number of cubes as `packspec`s, and you've dealt with that. Great!

However, we only raise an exception in this case as we stumble across it. From a user perspective, I think I rather see that we **fail early** rather than later.

What I'm suggesting is that before entering the `Saver` context manager, we detect this exception case beforehand, as we already have the `cubes` and the `packing`, so we can easily check that they align in terms of length (associated pairs). If not, then raise the exception. This mean that when we enter the `Saver` we know that we've got a stronger saving contract i.e. we should now save all the `cubes` in the `CubeList`.

lib/iris/fileformats/netcdf.py
jswanljung: Good point.

bjlittle: That would be great :smile:

lib/iris/fileformats/netcdf.py
```diff
@@ -1831,6 +1848,54 @@ def _create_cf_data_variable(self, cube, dimension_names, local_keys=None,
             The newly created CF-netCDF data variable.
 
         """
+        if 'packing' in kwargs:
+            packing = kwargs.pop('packing')
+            if packing:
+                scale_factor = None
+                add_offset = None
+                fill_value = None
+                masked = isinstance(cube.data, np.ma.core.MaskedArray)
```
bjlittle: @jswanljung You could use `np.ma.isMaskedArray(cube.data)` instead ...

lib/iris/fileformats/netcdf.py
```diff
@@ -1831,6 +1848,54 @@ def _create_cf_data_variable(self, cube, dimension_names, local_keys=None,
...
+                    else:
+                        scale_factor = (cmax - cmin)/(2**n-1)
+                    if dtype.kind == 'u':
+                        add_offset = cmin
+                    elif dtype.kind == 'i':
+                        if masked:
+                            add_offset = (cmax + cmin)/2
+                        else:
+                            add_offset = cmin + 2**(n-1)*scale_factor
+                if masked and not fill_value:
```
bjlittle: @jswanljung Careful ... are you sure you don't want `if masked and fill_value is not None` here ... just in case someone wanted a `fill_value=0` (why? ... just because they can I suppose)

lib/iris/fileformats/netcdf.py
ajdawson: Can you use `ma.` instead of `np.ma.` please. There is already an `import numpy.ma as ma` in the imports section.

lib/iris/fileformats/netcdf.py
ajdawson: You can test for an iterable with

```python

if isinstance(some_thing, Iterable):
    ...
```

This will return `True` for string also, so you'll need to check if it is iterable, but not a string.

bjlittle: This will also return `True` for a `dict` ...

ajdawson: Alright, never mind then!

lib/iris/fileformats/netcdf.py
```diff
@@ -1839,10 +1904,12 @@ def _create_cf_data_variable(self, cube, dimension_names, local_keys=None,
...
                                                  'NETCDF3_64BIT')):
-            # Determine whether there is a cube MDI value.
-            fill_value = None
-            if isinstance(cube.data, ma.core.MaskedArray):
-                fill_value = cube.data.fill_value
+
+            if not packing:
+                # Determine whether there is a cube MDI value.
+                fill_value = None
+                if isinstance(cube.data, ma.core.MaskedArray):
```
bjlittle: @jswanljung You've already cached this in the `masked` local variable, calculated earlier.

lib/iris/fileformats/netcdf.py
```diff
@@ -1839,10 +1904,12 @@ def _create_cf_data_variable(self, cube, dimension_names, local_keys=None,
         if (not cube.has_lazy_data()
                 or self._dataset.file_format in ('NETCDF3_CLASSIC',
                                                  'NETCDF3_64BIT')):
-            # Determine whether there is a cube MDI value.
-            fill_value = None
-            if isinstance(cube.data, ma.core.MaskedArray):
-                fill_value = cube.data.fill_value
+
+            if not packing:
```
bjlittle: `if packing is not None`

lib/iris/tests/unit/fileformats/netcdf/test_Saver.py
```diff
@@ -612,5 +612,162 @@ def test_laea_cs(self):
...
+                add_offset = cmin + 2**(n-1)*scale_factor
+        return (scale_factor, add_offset)
+
+    def test_single_packed_signed(self):
+        """Test saving a single CF-netCDF file with packing."""
+        # Read PP input file.
+        file_in = tests.get_data_path(
+            ('PP', 'cf_processing',
+             '000003000000.03.236.000128.1990.12.01.00.00.b.pp'))
+        cube = iris.load_cube(file_in)
```
bjlittle: @jswanljung `L642:645` common to all `test_single_packed_...` type tests, so you could put that in a `def setUp(self)`.

lib/iris/fileformats/netcdf.py
jswanljung: And for strings!

lib/iris/fileformats/netcdf.py
jswanljung: Yup.

lib/iris/fileformats/netcdf.py
jswanljung: Agreed.

lib/iris/fileformats/netcdf.py
```diff
@@ -1831,6 +1848,54 @@ def _create_cf_data_variable(self, cube, dimension_names, local_keys=None,
...
+        if 'packing' in kwargs:
+            packing = kwargs.pop('packing')
+            if packing:
+                scale_factor = None
+                add_offset = None
+                fill_value = None
+                masked = isinstance(cube.data, np.ma.core.MaskedArray)
+                if isinstance(packing, dict):
+                    if 'dtype' not in packing:
+                        msg = ("The dtype attribute is required for packing. ")
```
jswanljung: Whoops. I think it was longer before...

lib/iris/fileformats/netcdf.py
jswanljung: True. Wasn't my code except the condition.

jswanljung: Actually the masked local was only set if packing was in kwargs, but I might as well move it outside the conditional.

lib/iris/tests/unit/fileformats/netcdf/test_Saver.py
```diff
@@ -612,5 +612,162 @@ def test_laea_cs(self):
...
+            for cube, packedcube, dtype in zip(cubes, packedcubes, dtypes):
+                if dtype:
+                    sf, ao = self._get_scale_factor_add_offset(cube, dtype)
+                    decimal = int(-np.log10(sf))
+                    # Check that packed cube is accurate to expected precision
+                    self.assertArrayAlmostEqual(cube.data, packedcube.data,
+                                                decimal=decimal)
+                else:
+                    self.assertArrayEqual(cube.data, packedcube.data)
+
```
bjlittle: @jswanljung Awesome tests, thanks!

However, I'd consider them more as `integration` tests rather than `unit` tests ... @ajdawson and @marqh would you agree?

If so, then move them to `iris/tests/integration/test_netcdf.py` instead ...

lib/iris/tests/unit/fileformats/netcdf/test_Saver.py
```diff
@@ -612,5 +612,162 @@ def test_laea_cs(self):
         self._test(coord_system, expected)
 
 
+@tests.skip_data
+class Test_packed_data(tests.IrisTest):
+    def _get_scale_factor_add_offset(self, cube, datatype):
```
bjlittle: @jswanljung If you agree to split this test class (as mentioned below) then this `method` is really a convenience `function` that would live at the module level (note no usage of `self` within the method) ... then each of the test cases can still call it as a module level `function`.

lib/iris/fileformats/netcdf.py
bjlittle: Cool, thanks!

lib/iris/tests/unit/fileformats/netcdf/test_Saver.py
jswanljung: I started with the tests in tests/test_netcdf.py, then moved them to integration and then finally (after reading this: http://scitools.org.uk/iris/docs/latest/developers_guide/tests.html) I moved them to unit tests :-). But I'm happy to put them wherever you feel they belong.

bjlittle:
@jswanljung Okay I think I'm done. 

Thanks for this PR, it's going to make a valuable contribution to the code base :smile: 

I'm going to be distracted elsewhere for a wee while, but I'll get back on this PR asap (hope you don't mind). I'd love to see it merged :+1:

bjlittle:
@jswanljung You'll need to rebase from upstream/master to resolve conflicts, otherwise travis-ci won't kick in ...

lib/iris/fileformats/netcdf.py
```diff
@@ -54,6 +54,8 @@
 import iris.fileformats._pyke_rules
 import iris.io
 import iris.util
+from itertools import repeat
```
bjlittle: @jswanljung We've adopted an import ordering within iris, which is basically builtins, third-party, then iris specific modules. 

This said, `from itertools` is a `built-in` import and belongs up after the `import collections` line.

marqh:
> However, I'd consider them more as integration tests rather than unit tests ... @ajdawson and @marqh would you agree?

I agree, i think these should live under `tests.integration`

lib/iris/fileformats/netcdf.py
jswanljung: Okay.

lib/iris/fileformats/netcdf.py
```diff
@@ -54,6 +54,8 @@
...
+from six.moves import zip_longest
```
jswanljung: iris.tests.test_coding_standards.TestFutureImports

jswanljung:
I've got all tests passing now except an image test that is failing under Python 3. I think it's this problem: https://github.com/SciTools/iris/issues/2195. I'll try rerunning the tests.

jswanljung:
Actually I don't seem to be able to rerun these tests without committing something new.

bjlittle:
@jswanljung I've restarted your failed travis test for you. Unfortunately, that's our fault and is related to https://github.com/SciTools/iris/issues/2195. Apologies.

lib/iris/tests/integration/test_netcdf.py
```diff
@@ -23,13 +23,15 @@
 # importing anything else.
 import iris.tests as tests
 
+from itertools import repeat
```
bjlittle: @jswanljung Sorry, but within each import block, the imports should be in alphabetical order, so `itertools` comes after `contextlib` and before `os.path` :smile:

bjlittle: Yeah, I'm working on a fix for #2195 :scream: 

But I'll kept an eye on your travis jobs and keep restarting them manually. We'll get there ...

lib/iris/tests/unit/fileformats/netcdf/test_Saver.py
```diff
@@ -612,5 +612,162 @@ def test_laea_cs(self):
...
+                                      'single_packed_manual.cdl'))
+            self.assertArrayAlmostEqual(cube.data, packedcube.data,
+                                        decimal=decimal)
+
+    def test_multi_packed_single_dtype(self):
+        """Test saving multiple packed cubes with the same pack_dtype."""
+        # Read PP input file.
+        file_in = tests.get_data_path(('PP', 'cf_processing',
+                                       'abcza_pa19591997_daily_29.b.pp'))
+        cubes = iris.load(file_in)
```
jswanljung: There was a lot more redundancy than just the setup. I removed the redundancy, but without splitting the class.

lib/iris/tests/integration/test_netcdf.py
jswanljung: I have no doubt.

jswanljung:
Maybe it's too obvious to need pointing out in the docstring, but iterable arguments to cubes and packing must both be ordered. No sets, for instance!

marqh:
This PR is looking great from my point of view.  many thanks @jswanljung for the contributions

I fully support the approach and the implementation looks just fine to me

merge when ready, I say :+1:

bjlittle:
@jswanljung I really wouldn't worry about explicitly talking about sets in your docstrings.

Thanks for the nudge @marqh. I'll take another pass through this PR. Let's get this merge today!

jswanljung:
Great! Just added a what's new entry dated today (no other changes). Hopefully the tests will pass again.

bjlittle:
@jswanljung Thanks, I'll watch the build for you, just in case it misbehaves :wink:

ajdawson:
Referring to the test CDL files, why is the type of a packed variable `float`? I thought the point of packing it was to be able to represent with a smaller type like a short integer, perhaps I misunderstood how this is represented?

bjlittle:
@jswanljung Unfortunately, the format of the whatsnew contributions filename is quite strict.

There's a doctest failure, so you need to change your filename from ``newfeature_2016_Oct_19_netcdf_packing_support.txt`` to ``newfeature_2016-Oct-19_netcdf_packing_support.txt``

bjlittle:
@ajdawson Good question ... I would have thought that the point was to pack down into an ``int8``, ``int16``.

There's certainly loader code in iris that works out what the intended unpacked datatype should be for packed data in order to support deferred loading.

jswanljung:
@ajdawson, that is a very good question, and very disturbing...

jswanljung:
The embarrassing truth is that this version is doing everything but setting the datatype when creating the netcdf variable, so it is not in fact packing as the first version did. I switched the CDL files for the new PR because attributes were coming in in different order, but I didn't notice the datatype. This should be a very small code change, but I will have to regenerate the CDL files.

jswanljung:
Thank you @ajdawson for spotting that minor issue. :-(

jswanljung:
@bjlittle Okay, I fixed it, but an image test failed. I saw that I had accidentally added a character to a docstring, so I added a new commit. Hopefully it will pass this time.

jswanljung:
Testing seems to be broken right now. Problems with setting up the build, maybe temporary. But if someone would restart the tests, I'd be grateful.

jswanljung:
Setup of the test environment is still failing in some cases. Do we know what's wrong?

bjlittle:
@jswanljung Is your PR rebase against the latest upstream/master?

bjlittle:
I think that the problem is actually on travis ... I've see all kinds of unusual failures today ...

jswanljung:
Okay, tests passed, but I guess I still need to do a rebase against upstream/master again?

jswanljung:
Rebased against latest upstream/master.

jswanljung:
Passing after rebase except for a random image test.

jswanljung:
Passing again! So back to where we were this morning except now the code actually does something useful.

QuLogic:
@jswanljung It appears that one of your computers does not have git configured correctly. All but three of the commits are attributed to `Johan Swanljung <johan@Agda.local>` which is presumably not what you want.

jswanljung:
@QuLogic Thanks, I'll have a look at that.

bjlittle:
@jswanljung Thanks for an awesome contribution and your sustained effort on this PR. 

It's very much appreciated! Grattis :beers:

jswanljung:
And thank you and the rest of the SciTools people! This was a real learning experience and it's gratifying to see it merged. Also something I'll use all the time myself. 🍻

Issue #2195, opened 2016-10-14
Sporadic image test sha failures


bjlittle:
Within travis-ci, we have noticed several isolated cases where there are image test sha failures with the new image testing framework.

On re-running the failed travis job, all the tests pass. How disconcerting :disappointed: 

This issue has occurred on several separate PRs where the developer has not touched any image tests nor introduced any changes that would alter any image tests.

We require to investigate and fix this issue to ensure that the image testing framework is both stable and reliable, otherwise all PRs will be open to this issue.

bjlittle:
See this [comment](https://github.com/SciTools/iris/pull/2185#issuecomment-253009764) and also this [comment](https://github.com/SciTools/iris/pull/2185#issuecomment-253060111).

QuLogic:
Travis also no longer posts images or diffs, so it's difficult to say whether the problem is small or large. Even if generating the diff is difficult, maybe it should still post the image anyway for further review by a developer.

bjlittle:
@marqh I've got an iris proof of concept working using perceptual image hashing, see https://github.com/bjlittle/iris/pull/14 and https://github.com/bjlittle/test-images-dummy :smiley:

PR #2196, opened 2016-10-14
image hash work around

marqh:
temporary work around for unknown and occasional non-repeatable check_graphic test failure

if it fails, try again, mostly it works second time round

masks #2195 does not fix it

### DIFF
--- a/lib/iris/tests/__init__.py
+++ b/lib/iris/tests/__init__.py
@@ -669,8 +669,6 @@ class IrisTest(unittest.TestCase):
 
         unique_id = self._unique_id()
 
-        figure = plt.gcf()
-    
         repo_fname = os.path.join(os.path.dirname(__file__), 'results', 'imagerepo.json')
         with open(repo_fname, 'rb') as fi:
             repo = json.load(codecs.getreader('utf-8')(fi))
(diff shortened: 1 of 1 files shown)


marqh:
@bjlittle it's not pretty, but it may be better than nothing for now

lib/iris/tests/__init__.py
```diff
@@ -702,13 +700,17 @@ def check_graphic(self, tol=None):
                     if err.errno != 17:
                         raise
 
-            figure.savefig(result_fname)
+            def save_figure_hash():
+                figure = plt.gcf()
+                figure.savefig(result_fname)
```
bjlittle: @marqh `plt.gcf().savefig(result_fname)` works for me ... but I don't see the point in making such a cosmetic change here in this PR.

I'm going to have to rebase #2192 with this anyways, so I can tweak away in my PR :wink:

bjlittle:
@marqh Agreed that this is a short-term fix that's not going to close #2195 :+1:

lib/iris/tests/__init__.py
marqh: I don't think this is a cosmetic change

previously the `figure = plt.gcf()` and `figure.savefig(result_fname)` were in the code, a little way apart.

i moved them together, without changing either line

minor point tho

bjlittle:
@marqh Unfortunately, I don't see this `fix` making much difference ... we still have this issue, or at least I'm still encountering it with this PR included :cry:

PR #2197, opened 2016-10-14
removed gdal

corinnebosley:
This PR is to see which libraries I can remove from minimal-conda-requirements.txt without breaking the iris tests.

I will make one change per PR, which will be described in the commit message.


QuLogic:
> I will make one change per PR

Since it doesn't look like you're doing this in parallel, you could push one commit at a time and just wait for the CI.

marqh:
obsoleted by #2199

PR #2198, opened 2016-10-14
generate iris-grib translations

marqh:
updates to gen_translations to target iris-grib or iris, depending on argument

### DIFF
--- a/tools/gen_translations.py
+++ b/tools/gen_translations.py
@@ -23,6 +23,7 @@ metOcean mapping translations.
 from __future__ import (absolute_import, division, print_function)
 from six.moves import (filter, input, map, range, zip)  # noqa
 
+import argparse
 from datetime import datetime
 import os.path
 import requests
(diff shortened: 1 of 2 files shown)


tools/gen_translations.py
```diff
@@ -161,7 +162,7 @@ def build_um_cf_map(fuseki, filename, now, git_sha):
         fh.writelines(cffc.lines(fuseki))
 
 
-def build_grib_cf_map(fuseki, filename, now, git_sha):
+def build_grib_cf_map(fuseki, now, git_sha, base_dir):
```
lbdreyer: A very minor point but I would prefer if `build_grib_cf_map` and `build_um_cf_map` had a similar argument order. Currently it's:

> def build_grib_cf_map(fuseki, now, git_sha, ***base_dir***):
> def build_um_cf_map(fuseki, ***filename***, now, git_sha):


The docstring of `build_grib_cf_map` also needs updating since `filename` is no longer an argument.

tools/gen_translations.py
```diff
@@ -90,7 +91,7 @@
 
 DIR_BASE = '../lib/iris/fileformats'
 FILE_UM_CF = os.path.join(DIR_BASE, 'um_cf_map.py')
-FILE_GRIB_CF = os.path.join(DIR_BASE, 'grib', '_grib_cf_map.py')
+
```
lbdreyer: Unnecessary extra line?

tools/gen_translations.py
```diff
@@ -210,11 +214,20 @@ def build_grib_cf_map(fuseki, filename, now, git_sha):
...
+    args = parser.parse_args()
     now = datetime.utcnow().strftime('%d %B %Y %H:%m')
     git_sha = requests.get('http://www.metarelate.net/metOcean/latest_sha').text
     with FusekiServer() as fuseki:
-        build_um_cf_map(fuseki, FILE_UM_CF, now, git_sha)
-        build_grib_cf_map(fuseki, FILE_GRIB_CF, now, git_sha)
+        if args.iris_grib:
+            build_grib_cf_map(fuseki, now, git_sha, args.iris_grib)
+        else:
+            build_um_cf_map(fuseki, FILE_UM_CF, now, git_sha)
```
lbdreyer: Could this not be:
```
build_um_cf_map(fuseki, FILE_UM_CF, now, git_sha)
if args.iris_grib:
    build_grib_cf_map(fuseki, now, git_sha, args.iris_grib)
```
It would save us from having to run this twice (once for the um-cf map and once for the grib-cf map).

marqh:
these changes have caused me to run the translations for um cf and they are just a little behind, so I added the updates to the 2nd commit

lbdreyer:
Tests pass!!

PR #2199, opened 2016-10-14
Creq tidy

marqh:
tiyd up INSTALL and conda-requirements

reduce the scope of minimal-conda-requirements, which isn't especially minimal at the moment

i wonder whether our graphics test skip patterns are such that we could remove matplotlib as a dependency, but I think that cartopy might bring it in anyway via conda. I have not investigated this further

### DIFF
--- a/INSTALL
+++ b/INSTALL
@@ -70,7 +70,7 @@ postfix) in addition to the core packages.
 python 2.7 or 3.4 (http://www.python.org/)
     Iris requires Python 2.7 or Python 3.4.
 
-numpy 1.6 or later (http://numpy.scipy.org/)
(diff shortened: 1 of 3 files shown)


marqh:
treading on #2197 's toes, sorry @corinnebosley

marqh:
useful tidy up perhaps, but I'm a bit disappointed that i have barely touched teh minimal test run times

corinnebosley:
@marqh At least you've dumped gdal.  That can't be a bad thing.

bjlittle:
I agree. Minimal isn't quite minimal.

Thanks @marqh :+1:

PR #2206, opened 2016-10-24
Perceptual image hashing

### LINKED ISSUE #2195: Sporadic image test sha failures

bjlittle:
This PR provides a more stable and scalable approach to the new graphic testing framework, replacing cryptographic sha based image hashes with perceptual image hashes.

It relies on the support of the new perceptual image hash repository [test-iris-imagehash](https://github.com/SciTools/test-iris-imagehash).

Perceptual image hashes allows checking for the degree of *similarity* between images through measuring the hamming distance difference (a count of bit differences) between image hashes. 

The major benefit of this perceptual image hash approach is that it is simple, scalable (already almost halved the number of required reference images) and makes us less sensitive to scientific software stack dependency changes.

This PR also addresses the issues raised in #2195. Normally, I would have pushed those changes in a separate PR, but there is clearly a major overlap here with regards to graphical testing. In short, the problem simply relates to iris plotting not being thread-safe, and our `travis-ci` testing uses `nose`, which highlighted this problem in a typically dreaded, threaded *sporadic* way.

Closes #2195

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -60,6 +60,10 @@ install:
       fi
     fi
 
+  # Experimental image hashing (TBD: push recipe to conda-forge!)
+  - conda install pip
+  - pip install imagehash
+
   - PREFIX=$HOME/miniconda/envs/$ENV_NAME
 
   # Output debug info
--- a/.travis.yml
+++ b/.travis.yml
@@ -60,7 +60,7 @@ install:
       fi
     fi
 
-  # Experimental image hashing (TBD: push recipe to conda-forge!)
+  # Perceptual image hashing (TBD: push recipe to conda-forge!)
   - conda install pip
   - pip install imagehash
 
--- a/docs/iris/example_tests/extest_util.py
+++ b/docs/iris/example_tests/extest_util.py
@@ -35,7 +35,6 @@ import iris
 from iris._deprecation import IrisDeprecation
 import iris.plot as iplt
 import iris.quickplot as qplt
-from iris.tests import _DEFAULT_IMAGE_TOLERANCE
 
 
 EXAMPLE_DIRECTORY = os.path.join(os.path.dirname(os.path.dirname(__file__)),
@@ -60,7 +59,7 @@ def add_examples_to_path():
 
 
 @contextlib.contextmanager
-def show_replaced_by_check_graphic(test_case, tol=_DEFAULT_IMAGE_TOLERANCE):
+def show_replaced_by_check_graphic(test_case):
     """
     Creates a context manager which can be used to replace the functionality
     of matplotlib.pyplot.show with a function which calls the check_graphic
@@ -69,7 +68,7 @@ def show_replaced_by_check_graphic(test_case, tol=_DEFAULT_IMAGE_TOLERANCE):
     """
     def replacement_show():
         # form a closure on test_case and tolerance
-        test_case.check_graphic(tol=tol)
+        test_case.check_graphic()
 
     orig_show = plt.show
     plt.show = iplt.show = qplt.show = replacement_show
--- a/docs/iris/example_tests/test_atlantic_profiles.py
+++ b/docs/iris/example_tests/test_atlantic_profiles.py
@@ -33,7 +33,7 @@ class TestAtlanticProfiles(tests.GraphicsTest):
         with fail_any_deprecation_warnings():
             with add_examples_to_path():
                 import atlantic_profiles
-            with show_replaced_by_check_graphic(self, tol=14.0):
+            with show_replaced_by_check_graphic(self):
                 atlantic_profiles.main()
 
 
--- a/lib/iris/plot.py
+++ b/lib/iris/plot.py
@@ -27,6 +27,8 @@ from six.moves import (filter, input, map, range, zip)  # noqa
 
 import collections
 import datetime
+from functools import wraps
+import threading
 
 import cartopy.crs as ccrs
 import cartopy.mpl.geoaxes
@@ -52,8 +54,27 @@ import iris.palette
 # Cynthia Brewer citation text.
 BREWER_CITE = 'Colours based on ColorBrewer.org'
 
-
 PlotDefn = collections.namedtuple('PlotDefn', ('coords', 'transpose'))
+
+# Threading reentrant lock to ensure thread-safe plotting.
+_lock = threading.RLock()
+
+
+def _locker(func):
+    """
+    Decorator that ensures a thread-safe atomic operation is
+    performed by the decorated function.
+
+    Uses a shared threading reentrant lock to provide thread-safe
+    plotting by public API functions.
+
+    """
+    @wraps(func)
+    def decorated_func(*args, **kwargs):
+        with _lock:
+            result = func(*args, **kwargs)
+        return result
+    return decorated_func
 
 
 def _get_plot_defn_custom_coords_picked(cube, coords, mode, ndims=2):
@@ -665,6 +686,7 @@ def _map_common(draw_method_name, arg_func, mode, cube, plot_defn,
     return plotfn(*new_args, **kwargs)
 
 
+@_locker
 def contour(cube, *args, **kwargs):
     """
     Draws contour lines based on the given Cube.
@@ -689,6 +711,7 @@ def contour(cube, *args, **kwargs):
     return result
 
 
+@_locker
 def contourf(cube, *args, **kwargs):
     """
     Draws filled contours based on the given Cube.
@@ -815,6 +838,7 @@ def _fill_orography(cube, coords, mode, vert_plot, horiz_plot, style_args):
     return result
 
 
+@_locker
 def orography_at_bounds(cube, facecolor='#888888', coords=None, axes=None):
     """Plots orography defined at cell boundaries from the given Cube."""
 
@@ -845,6 +869,7 @@ def orography_at_bounds(cube, facecolor='#888888', coords=None, axes=None):
                            horiz_plot, style_args)
 
 
+@_locker
 def orography_at_points(cube, facecolor='#888888', coords=None, axes=None):
     """Plots orography defined at sample points from the given Cube."""
 
@@ -866,6 +891,7 @@ def orography_at_points(cube, facecolor='#888888', coords=None, axes=None):
                            horiz_plot, style_args)
 
 
+@_locker
 def outline(cube, coords=None, color='k', linewidth=None, axes=None):
     """
     Draws cell outlines based on the given Cube.
@@ -903,6 +929,7 @@ def outline(cube, coords=None, color='k', linewidth=None, axes=None):
     return result
 
 
+@_locker
 def pcolor(cube, *args, **kwargs):
     """
     Draws a pseudocolor plot based on the given Cube.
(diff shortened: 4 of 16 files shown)


docs/iris/example_tests/extest_util.py
```diff
@@ -69,7 +68,7 @@ def show_replaced_by_check_graphic(test_case, tol=_DEFAULT_IMAGE_TOLERANCE):
     """
     def replacement_show():
         # form a closure on test_case and tolerance
-        test_case.check_graphic(tol=tol)
+        test_case.check_graphic()
```
bjlittle: In the hope of refraining from a culture of tweaking graphical testing tolerances, I've closed the door to passing a `tol` through to `check_graphic`.

docs/iris/example_tests/test_atlantic_profiles.py
```diff
@@ -33,7 +33,7 @@ def test_atlantic_profiles(self):
         with fail_any_deprecation_warnings():
             with add_examples_to_path():
                 import atlantic_profiles
-            with show_replaced_by_check_graphic(self, tol=14.0):
```
bjlittle: This covers up an issue in the use of `twiny` ...

lib/iris/plot.py
```diff
@@ -52,9 +54,28 @@
...
+    plotting by public API functions.
+
+    """
+    @wraps(func)
+    def decorated_func(*args, **kwargs):
+        with _lock:
+            result = func(*args, **kwargs)
+        return result
+    return decorated_func
+
```
bjlittle: A simple decorator to facilitate the use of a re-entrant lock to provide thread-safe plotting. A re-entrant lock is required as `iris.plot.contourf` calls `iris.plot.contour` ...

lib/iris/plot.py
```diff
@@ -665,6 +686,7 @@ def _map_common(draw_method_name, arg_func, mode, cube, plot_defn,
     return plotfn(*new_args, **kwargs)
 
 
+@_locker
 def contour(cube, *args, **kwargs):
```
bjlittle: Use the lock for all public api plotting functions ...

lib/iris/tests/__init__.py
```diff
@@ -143,7 +147,8 @@
     plt.switch_backend('tkagg')
     _DISPLAY_FIGURES = True
 
-_DEFAULT_IMAGE_TOLERANCE = 10.0
+# Threading non re-entrant blocking lock to ensure thread-safe plotting.
+_lock = threading.Lock()
```
bjlittle: A **non re-entrant** lock is required here to ensure that there is no cross-pollination between tests using the wrong plot figures or axes.

lib/iris/tests/__init__.py
```diff
@@ -695,93 +702,99 @@ def _assert_graphic(self):
...
                     if err.errno != 17:
                         raise
 
-            def _save_figure_hash():
-                plt.gcf().savefig(result_fname)
-                # Determine the test result image hash using sha1.
-                with open(result_fname, 'rb') as fi:
-                    sha1 = hashlib.sha1(fi.read())
-                return sha1
-
```
bjlittle: This hot-fix didn't work and needs to be purged.

At this point the plot was **already** wrong i.e. another thread had already clobbered either the plot title or axes title/s, so re-saving the clobbered plot image was never going to yield the correct sha.

lib/iris/tests/__init__.py
```diff
@@ -695,93 +702,99 @@ def _assert_graphic(self):
...
+                l = []
+                count = hash_size * (hash_size // 4)
+                if len(hexstr) != count:
+                    emsg = 'Expected hex string size of {}.'
+                    raise ValueError(emsg.format(count))
+                for i in range(count // 2):
+                    h = hexstr[i*2:i*2+2]
+                    v = int("0x" + h, 16)
+                    l.append([v & 2**i > 0 for i in range(8)])
+                return imagehash.ImageHash(np.array(l))
```
bjlittle: This fix has already been merged to `imagehash` but the fix has not yet been pushed to `pypi` ... we can remove this function when the fix is available via `pypi` .... or even conda-forge!

lib/iris/tests/__init__.py
```diff
@@ -695,93 +702,99 @@ def _assert_graphic(self):
...
+
+            # Calculate the test result perceptual image hash.
+            buffer = io.BytesIO()
+            figure = plt.gcf()
+            figure.savefig(buffer, format='png')
+            buffer.seek(0)
```
bjlittle: Memory buffers are our friend ... only save to disk on failure, rather than save to disk and delete on success.

lib/iris/tests/__init__.py
```diff
@@ -695,93 +702,99 @@ def _assert_graphic(self):
...
-                if sha1.hexdigest() not in expected:
-                    # This can be an accidental failure, unusual, but it occurs
-                    # https://github.com/SciTools/iris/issues/2195
-                    # retry once, in case it passes second time round.
-                    sha1 = _save_figure_hash()
+                # Calculate the hamming distance vector for the result hash.
+                distances = [e - phash for e in expected]
 
-                if sha1.hexdigest() not in expected:
+                if np.all([hd > tol for hd in distances]):
```
bjlittle: The result image fails to meet expectation iff it is not similar to all our registered expected images.

lib/iris/tests/experimental/test_animate.py
```diff
@@ -40,6 +40,7 @@
 @tests.skip_plot
 class IntegrationTest(tests.GraphicsTest):
     def setUp(self):
+        super(IntegrationTest, self).setUp()
```
bjlittle: We need to make sure that the `GraphicsTest.setUp` is called, as it `aquires` the non re-entrant lock that protects plot figures for a test.

lib/iris/tests/idiff.py
```diff
@@ -38,10 +37,12 @@
 
 from PIL import Image
 import filelock
+import imagehash
```
bjlittle: ... long live `imagehash`! :wink:

lib/iris/tests/idiff.py
```diff
@@ -137,42 +133,99 @@ def skip(event):
...
+    l = []
+    count = hash_size * (hash_size // 4)
+    if len(hexstr) != count:
+        emsg = 'Expected hex string size of {}.'
+        raise ValueError(emsg.format(count))
+    for i in range(count // 2):
+        h = hexstr[i*2:i*2+2]
+        v = int("0x" + h, 16)
+        l.append([v & 2**i > 0 for i in range(8)])
+    return imagehash.ImageHash(np.array(l))
```
bjlittle: Again this will disappear ...

lib/iris/tests/runner/_runner.py
```diff
@@ -24,48 +24,31 @@
...
             return fh.read().encode("base64").replace("\n", "")
 
     html = ['<!DOCTYPE html>', '<html>', '<body>']
+    rdir = os.path.join(os.path.dirname(__file__), os.path.pardir,
+                        'result_image_comparison')
+    if not os.access(rdir, os.W_OK):
+        rdir = os.path.join(os.getcwd(), 'iris_image_test_output')
 
-    for expected, actual, diff in _failed_images_iter():
+    for expected, actual, diff in step_over_diffs(rdir, 'similar', False):
```
bjlittle: This fixes the failed test runner `--print-failed-images` capability, so that we can see graphical failures on travis again :smile:

bjlittle:
Whoop! All the tests passed first time! :beers:

marqh:
> Whoop! All the tests passed first time! :beers:

:beers: indeed

**marqh** clinks glasses with @bjlittle

lib/iris/tests/experimental/test_animate.py
dkillick: @bjlittle do we need to follow this pattern for _all_ instances of test classes inheriting from `tests.GraphicsTest`? If so, you'll also need to update the following:

* In `lib/iris/tests`:
 * `test_analysis.TestRotatedPole`
 * `test_coordsystem.Test_LambertConformal`
 * `test_mapping.TestBasic`
 * In `testPlot`: `Test1dPlotMultiArgs`, `TestMissingCoord`, `TestMissingCS`, `TestAttributePositive`, TestPlotOtherCoordSystems
* All affected test classes in `lib/iris/tests/integration/test_grib_load.py`
* Test class in `lib/iris/tests/integration/test_netcdftime.py`

.travis.yml
```diff
@@ -60,6 +60,10 @@ install:
       fi
     fi
 
+  # Perceptual image hashing (TBD: push recipe to conda-forge!)
+  - conda install pip
```
QuLogic: As part of the conda environment (and installing to it), though?

ajdawson: pip is installed by default when you create a new conda environment (e.g. https://travis-ci.org/SciTools/iris/jobs/170017955#L253) so it doesn't need to be done separately. The result of this line is a no-op (https://travis-ci.org/SciTools/iris/jobs/170017955#L464).

lib/iris/plot.py
ajdawson: I'm inclined to agree... it feels like this should be done when testing. We are mostly wrapping matplotlib calls, which I guess aren't thread safe either, so it seems consistent to lock our wrappers.

.travis.yml
bjlittle: Yup

lib/iris/tests/experimental/test_animate.py
bjlittle: @dkillick Thanks. Yes, we do indeed need to follow a pattern here ...

A `GraphicsTest` subclass **must** call the `GraphicsTest.setUp` method (in order to `acquire` the non-reentrant lock) **if** the subclass in question overrides the inherited `setUp` method. If the subclass doesn't specialise the `setUp` method then, thanks to inheritance, the right thing does happen, in that the `unittest` calls the (inherited) `GraphicsTest.setUp` for the subclass (this also applies to `GraphicsTest.tearDown` btw to `release` the lock).

In **all** of the test cases that you identified above, there is **no need** to explicitly call `GraphicsTest.setUp`, simply because the subclass **does not** specialise the `setUp` method.

Note that, you incorrectly mentioned `test_mapping.TestBasic`:
```python
class TestBasic(tests.GraphicsTest):
    def setUp(self):
        super(TestBasic, self).setUp()
        self.cube = iris.tests.stock.realistic_4d()
```
which does apply the pattern, and so does `test_plot.Test1dPlotMultiArgs`:
```python
class Test1dPlotMultiArgs(tests.GraphicsTest):
    # tests for iris.plot using multi-argument calling convention

    def setUp(self):
        super(Test1dPlotMultiArgs, self).setUp()
        self.cube1d = _load_4d_testcube()[0, :, 0, 0]
        self.draw_method = iplt.plot
```
So, I'd argue that this isn't an issue, unless I'm missing something ...

dkillick: > I'd argue that this is a non-issue

Agreed. I was sure there must be a pattern to it, but I couldn't work out what the pattern was! Good ol' inheritance :wink:

PR #2208, opened 2016-10-24
Bump the nose multiprocess timeout

bjlittle:
For the first time I've seen some `nose.plugins.multiprocess.TimedOutException` being raised on `travis-ci` ... so I've bumped up the `--process-timeout`

We might have been on the edge with the current timeout of `250` seconds and introducing the thread-safe plotting may be pushing it just over the edge (plus also given the variable load on `travis-ci`)


marqh:
obsoleted by #2207

PR #2211, opened 2016-10-24
override test data path

marqh:
enable test data path to be overridden at run time, via an environment variable

### DIFF
--- a/lib/iris/config.py
+++ b/lib/iris/config.py
@@ -132,8 +132,12 @@ TEST_DATA_DIR = get_dir_option(_RESOURCE_SECTION, 'test_data_dir',
 # Override the data repository if the appropriate environment variable
 # has been set.  This is used in setup.py in the TestRunner command to
 # enable us to simulate the absence of external data.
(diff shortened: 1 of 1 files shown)


lib/iris/config.py
```diff
@@ -132,8 +132,12 @@ def get_dir_option(section, option, default=None):
 # Override the data repository if the appropriate environment variable
 # has been set.  This is used in setup.py in the TestRunner command to
 # enable us to simulate the absence of external data.
-if os.environ.get("override_test_data_repository"):
-    TEST_DATA_DIR = None
+override = os.environ.get("override_test_data_repository")
+if override:
+    if override == '1':
```
dkillick: What if the passed value is neither `'1'` nor a proper path to the test data?

marqh: then the test data directory will not be found, and the tests will fall back to the 'no data' approach of skipping

dkillick:
Other than the single comment above I think this is a useful side-step for missing test data in an environment that Iris is deployed to :+1:

PR #2212, opened 2016-10-25
Stock

marqh:
* move stock npy arrays to iris-test-data https://github.com/SciTools/iris-test-data/pull/45
* update skip_data pattern usage
* alter minimal testing environment to not use test_data

### DIFF
--- a/lib/iris/tests/__init__.py
+++ b/lib/iris/tests/__init__.py
@@ -183,7 +183,10 @@ def get_data_path(relative_path):
     """
     if not isinstance(relative_path, six.string_types):
         relative_path = os.path.join(*relative_path)
-    data_path = os.path.join(iris.config.TEST_DATA_DIR, relative_path)
+    test_data_dir = iris.config.TEST_DATA_DIR
+    if test_data_dir is None:
+        test_data_dir = ''
+    data_path = os.path.join(test_data_dir, relative_path)
 
     if _EXPORT_DATAPATHS_FILE is not None:
         _EXPORT_DATAPATHS_FILE.write(data_path + '\n')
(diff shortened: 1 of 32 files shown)


lib/iris/tests/__init__.py
```diff
@@ -183,7 +183,10 @@ def get_data_path(relative_path):
     """
     if not isinstance(relative_path, six.string_types):
         relative_path = os.path.join(*relative_path)
-    data_path = os.path.join(iris.config.TEST_DATA_DIR, relative_path)
+    test_data_dir = iris.config.TEST_DATA_DIR
+    if test_data_dir is None:
```
pp-mo: I don't see the need for this..
From how ```iris.config.TEST_DATA_DIR``` is initialised in iris.config, I think this **can't** be empty ?

marqh: it can occur, here:
https://github.com/SciTools/iris/blob/master/lib/iris/config.py#L138
```
override = os.environ.get("override_test_data_repository")
if override:
    if override == '1':
        TEST_DATA_DIR = None
    else:
        TEST_DATA_DIR = override
```

lib/iris/tests/test_mapping.py
```diff
@@ -47,8 +47,10 @@
 
 
 @tests.skip_plot
+@tests.skip_data
 class TestBasic(tests.GraphicsTest):
-    cube = iris.tests.stock.realistic_4d()
+    def setUp(self):
+        self.cube = iris.tests.stock.realistic_4d()
```
pp-mo: Needs to call the class setUp here :
```        super(TestBasic, self).setUp()```

**Comment by pp-mo** [comment]
As a cross-check, I ran all tests with control exports:
 *  ```export override_test_data_repository=1``` : breaks the config.TEST_DATA_DIR value
 * ```export IRIS_TEST_NO_DATA=1```  : force everything wrapped in "skip_data" to skip

The only failure is this one :
```
iris.tests.experimental.regrid.test_regrid_conservative_via_esmpy.TestConservativeRegrid.test_multidimensional
``` 

So, that needs wrapping with "skip_data".

PR #2215, opened 2016-10-26
imagehash via conda

marqh:
update travis to use imagehash from conda/scitools

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -60,10 +60,6 @@ install:
       fi
     fi
 
-  # Perceptual image hashing (TBD: push recipe to conda-forge!)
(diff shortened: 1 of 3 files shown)


marqh:
@lbdreyer could you check this over please

.travis.yml
```diff
@@ -60,10 +60,6 @@ install:
       fi
     fi
 
-  # Perceptual image hashing (TBD: push recipe to conda-forge!)
```
lbdreyer: It may be worth noting that the reason imagehash was put in conda-recipe-scitools rather than conda-forge is that conda-forge doesn't currently work our travis testing.

Issue #2210, opened 2016-10-24
Outstanding review comments from #2206


dkillick:
I believe that #2206 was merged a little hastily, and I have made extra comments on the PR post-merge. To try and prevent these extra comments from getting lost, here's an issue that references them.

Ping @bjlittle

QuLogic:
I believe these were addressed by #2215 and #2218. Please re-open if I'm incorrect.

dkillick:
Good spot @QuLogic - many thanks.

PR #2217, opened 2016-10-26
Remove plot locking.

bjlittle:
This PR addresses the post-merge issues raised by @dkillick on https://github.com/SciTools/iris/pull/2206.

Locking at the iris plot level will not guarantee to the user thread-safe plotting. There is still the possibility of an iris plot using an mpl figure from a different thread in certain circumstances. In the case of https://github.com/SciTools/iris/pull/2206, the implementation introduced a locking mechanism at the iris plotting api level, which is in fact at **too low** a level.

If a user insists on using iris plotting in a threaded context, then we should advocate that they take responsibility for locking themselves (which, to be fair, is pretty simple to do) in combination with creating the target mpl figure/axes.

Note that, for clarity, we do require to maintain non re-entrant locking at the individual graphics test case level to avoid the issues raised in https://github.com/SciTools/iris/issues/2195.

Closes https://github.com/SciTools/iris/issues/2210.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -61,7 +61,6 @@ install:
     fi
 
   # Perceptual image hashing (TBD: push recipe to conda-forge!)
-  - conda install pip
   - pip install imagehash
 
   - PREFIX=$HOME/miniconda/envs/$ENV_NAME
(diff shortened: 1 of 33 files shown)


QuLogic:
Wrong merge target.

PR #2218, opened 2016-10-26
Remove plot api locking.

bjlittle:
This PR addresses the post-merge issues raised by @dkillick on https://github.com/SciTools/iris/pull/2206.

Locking at the iris plot level will not guarantee to the user thread-safe plotting. In certain circumstances there is still the possibility of an iris plot using an mpl figure from a different thread. In the case of https://github.com/SciTools/iris/pull/2206, the implementation introduced a locking mechanism at the iris plotting api level, which is in fact at **too low** a level.

If a user insists on using iris plotting in a threaded context, then we should advocate that they take responsibility for locking themselves (which, to be fair, is pretty simple to do) in combination with creating the target mpl figure/axes (within the safety of the locked critical region).

Note that, for clarity, we do require to maintain non re-entrant locking at the individual graphics test case level in order to avoid the issues raised in https://github.com/SciTools/iris/issues/2195.

Closes https://github.com/SciTools/iris/issues/2210.

### DIFF
--- a/lib/iris/plot.py
+++ b/lib/iris/plot.py
@@ -27,8 +27,6 @@ from six.moves import (filter, input, map, range, zip)  # noqa
 
 import collections
 import datetime
-from functools import wraps
-import threading
 
 import cartopy.crs as ccrs
 import cartopy.mpl.geoaxes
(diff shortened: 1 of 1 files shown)


bjlittle:
@dkillick Ping ...

dkillick:
@bjlittle spotted! Now all I need is some time to look through this...

bjlittle:
@dkillick Cool, thanks. Mind to review my responses to your comments on #2206 to make sure you're in agreement and happy with the closure of #2210

dkillick:
One thing this PR does not address is [removing the pip install from travis.yml](https://github.com/SciTools/iris/blob/master/.travis.yml#L63-L65). Depending on the scope of this PR, I'm either happy with this or not happy with this:

* if the scope of this is **only remove spurious plotting locks**, them that's fine (but can we get this done in a followup PR - see [this comment](https://github.com/SciTools/iris/pull/2206/files#r84649377)), or
* if the scope of this is **backing out all that's wrong with #2206** then I think it should be done here.

Comments on a postcard :postbox:! Or, you know, on this PR.

dkillick:
Otherwise I'm happy that the changes here represent a complete removal of the additions to `plot.py` that were introduced in #2206 :+1:

bjlittle:
@dkillick I've not address removing `conda install pip` from the `.travis.yml` file as this has already been done by #2211 on `v1.11.x` branch.

This PR is only to address removing the locking from the plotting api.

I don't quite understand your comment regarding `backing out all that's wrong with #2206`, care to elaborate ... what else is wrong?

dkillick:
> I don't quite understand your comment regarding "backing out all that's wrong with #2206"

Sorry! What I meant was "do we want this PR to deal with all problems introduced in #2206", which to my mind includes the spurious `pip install` addition in `.travis.yml`. I think my question is answered by this comment...

> This PR is only to address removing the locking from the plotting api

and as such I think we might be there with this fix! Thanks @bjlittle

PR #2220, opened 2016-10-27
Prevent dot_path tests from corrupting the stored path.

pp-mo:
Without this, the tests in ```iris.tests.test_file_save``` that use "dot" can fail, if they happen to run ***after*** this one.
Especially true as the last test in "test__dot_path" leaves it set to None.

Background : I think the order of tests running is not fully guaranteed, and may not always be repeatable.
? Possibly it is  different for the different ways, i.e. "python setup.py test" // "nosetests" // "python -m unittest" ?


lbdreyer:
This will need closing now that #2222 has been merged.

PR #2221, opened 2016-10-27
limit filelock use

marqh:
remove the behaviour, introduced in #2192, where installations without write permissions to the iris source directories are unable to run tests, due to 
```
IOError: [Errno 13] Permission denied: 'imagerepo.lock'
```

@bjlittle does this maintain your desired behaviour for developers whilst removing the installation and testing  block?

### DIFF
--- a/lib/iris/tests/__init__.py
+++ b/lib/iris/tests/__init__.py
@@ -734,9 +734,14 @@ class IrisTest(unittest.TestCase):
                 figure.savefig(hash_fname)
                 msg = 'Creating imagerepo entry: {} -> {}'
                 print(msg.format(unique_id, uri))
-                with open(repo_fname, 'wb') as fo:
-                    json.dump(repo, codecs.getwriter('utf-8')(fo), indent=4,
-                              sort_keys=True)
+                lock = filelock.FileLock(repo_fname)
+                # The imagerepo.json file is a critical resource, so ensure
+                # thread safe read/write behaviour via platform independent
+                # file locking.
+                with lock.acquire(timeout=600):
+                    with open(repo_fname, 'wb') as fo:
+                        json.dump(repo, codecs.getwriter('utf-8')(fo),
+                                  indent=4, sort_keys=True)
 
             # TBD: Push this fix to imagehash (done!)
             # See https://github.com/JohannesBuchner/imagehash/pull/31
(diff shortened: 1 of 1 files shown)


lib/iris/tests/__init__.py
```diff
@@ -734,9 +734,14 @@ def _create_missing():
                 figure.savefig(hash_fname)
                 msg = 'Creating imagerepo entry: {} -> {}'
                 print(msg.format(unique_id, uri))
-                with open(repo_fname, 'wb') as fo:
-                    json.dump(repo, codecs.getwriter('utf-8')(fo), indent=4,
-                              sort_keys=True)
+                lock = filelock.FileLock(repo_fname)
```
dkillick: This looks like an improvement over the initial proposal.

lbdreyer:
Thanks @marqh :)

bjlittle:
@marqh @lbdreyer @dkillick You've introduced a bug here ...

The lock file [imagerepo.lock](https://github.com/SciTools/iris/pull/2221/files#diff-350ecfa750b99028f16446216ccbcfdbL806) is a separate file from the file [imagerepo.json](https://github.com/SciTools/iris/pull/2221/files#diff-350ecfa750b99028f16446216ccbcfdbL694) that the lock protects access to in the lock critical region.

The code change that @lbdreyer merged uses the `imagerepo.json` file as the actual lock file in error. That's bad.

[Line 737](https://github.com/SciTools/iris/pull/2221/files#diff-350ecfa750b99028f16446216ccbcfdbR737) should be replaced with the following:
```python
lock = filelock.FileLock(os.path.join(_RESULT_PATH, 'imagerepo.lock'))
```

And [line 811](https://github.com/SciTools/iris/pull/2221/files#diff-350ecfa750b99028f16446216ccbcfdbL806) now requires to be deleted ... infact the `check_graphic` function no longer needs to be a wrapper function to `_assert_graphic`. It should be the case now that `_assert_graphic` **is** `check_graphic`.

Someone needs to fix this.

lbdreyer:
Interesting that the test still pass :confused: 

Thanks for catching this @bjlittle!

I have now put up PR #2225 that should hopefully fix it.

bjlittle:
@lbdreyer There are no tests to test the testing framework.

dkillick:
> There are no tests to test the testing framework

This has always struck me as an abrupt irony.

PR #2222, opened 2016-10-28
Prevent dot_path tests from corrupting the stored path.

marqh:
replaces #2220 
same code, but targeted at v1.11.x

### DIFF
--- a/lib/iris/tests/unit/fileformats/dot/test__dot_path.py
+++ b/lib/iris/tests/unit/fileformats/dot/test__dot_path.py
@@ -28,7 +28,7 @@ import subprocess
 
 import mock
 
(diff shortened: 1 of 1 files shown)


ajdawson:
FYI you can change the target of a PR by clicking edit next to the PR title and selecting a new target branch from the drop-down menu. This way you don't lose any comments etc.

marqh:
> FYI you can change the target of a PR by clicking edit next to the PR title and selecting a new target branch from the drop-down menu. This way you don't lose any comments etc.

Hi @ajdawson 

i know about this and I tried it, but there is a commit history confusion in this case, which I don't quite follow, which made this approach not work, hence this separate PR

PR #2223, opened 2016-10-28
Dot test import

pp-mo:
Without this, the test fails when run in isolation,
with a couple of errors like "AttributeError: 'module' object has no attribute 'dot'".

It doesn't **usually** fail when running all tests, presumably because some other test imports it first.
However, think this error has occurred sometimes.

I believe the order of running tests is not dependable.
In any case, this is an obvious bug, as we _shouldn't_ have such dependencies.


QuLogic:
Great, it happened to me and I was very confused it didn't break when run by itself.

PR #2227, opened 2016-10-31
Unpin miniconda to latest

bjlittle:
Unpin versions of `miniconda` in the `.travis-ci` to be the latest.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -29,9 +29,9 @@ install:
   # -----------------
   - export CONDA_BASE=http://repo.continuum.io/miniconda/Miniconda
   - if [[ "$TRAVIS_PYTHON_VERSION" == 2* ]]; then
-      wget ${CONDA_BASE}-3.7.0-Linux-x86_64.sh -O miniconda.sh;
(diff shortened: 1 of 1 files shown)


bjlittle:
Thanks @ajdawson.

.travis.yml
```diff
@@ -29,9 +29,9 @@ install:
   # -----------------
   - export CONDA_BASE=http://repo.continuum.io/miniconda/Miniconda
```
ocefpaf: This should be `https` now. (Recent change.)

bjlittle: @ocefpaf Is `http` being deprecated? i.e. we should move to `https`?

ocefpaf: Yes. In fact I am surprised that the `wget` worked here. I spent my whole weekend fixing that in mane places in conda-forge :smile:

bjlittle: You sure know how to have fun :wink: 

Thanks for the heads-up @ocefpaf

ajdawson: It auto-redirects to the https version, which wget seems OK with... Best move to https anyway though.

PR #2230, opened 2016-11-01
Stashcubelist

marqh:
usability tweak to `CubeList.__str__` to print a STASH code, if it exists and otherwise 
`unknown / (unknown)` 
would otherwise be returned

### DIFF
--- a/lib/iris/cube.py
+++ b/lib/iris/cube.py
@@ -1835,6 +1835,10 @@ bound=(1994-12-01 00:00:00, 1998-12-01 00:00:00)
 
         nameunit = '{name} / ({units})'.format(name=self.name(),
                                                units=self.units)
+        # If all unknown and a STASH attribute exists, use it.
+        if (nameunit == 'unknown / (unknown)' and
+           'STASH' in self.attributes):
+            nameunit = '{}'.format(self.attributes.get('STASH'))
         cube_header = '{nameunit!s:{length}} ({dimension})'.format(
             length=name_padding,
             nameunit=nameunit,
--- a/lib/iris/cube.py
+++ b/lib/iris/cube.py
@@ -1836,9 +1836,8 @@ bound=(1994-12-01 00:00:00, 1998-12-01 00:00:00)
         nameunit = '{name} / ({units})'.format(name=self.name(),
                                                units=self.units)
         # If all unknown and a STASH attribute exists, use it.
-        if (nameunit == 'unknown / (unknown)' and
-           'STASH' in self.attributes):
-            nameunit = '{}'.format(self.attributes.get('STASH'))
+        if nameunit == 'unknown / (unknown)' and 'STASH' in self.attributes:
+            nameunit = '{}'.format(self.attributes['STASH'])
         cube_header = '{nameunit!s:{length}} ({dimension})'.format(
             length=name_padding,
             nameunit=nameunit,
--- a/lib/iris/tests/unit/cube/test_CubeList.py
+++ b/lib/iris/tests/unit/cube/test_CubeList.py
@@ -1,4 +1,4 @@
-# (C) British Crown Copyright 2014 - 2015, Met Office
+# (C) British Crown Copyright 2014 - 2016, Met Office
 #
 # This file is part of Iris.
 #
@@ -22,6 +22,7 @@ from six.moves import (filter, input, map, range, zip)  # noqa
 # Import iris.tests first so that some things can be initialised before
 # importing anything else.
 import iris.tests as tests
+import iris.tests.stock
 
 from cf_units import Unit
 import numpy as np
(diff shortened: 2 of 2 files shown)


lib/iris/cube.py
```diff
@@ -1835,6 +1835,10 @@ def summary(self, shorten=False, name_padding=35):
 
         nameunit = '{name} / ({units})'.format(name=self.name(),
                                                units=self.units)
+        # If all unknown and a STASH attribute exists, use it.
+        if (nameunit == 'unknown / (unknown)' and
+           'STASH' in self.attributes):
```
bjlittle: This line can be joined to the above and it'll still only be 78 characters long.

lib/iris/cube.py
```diff
@@ -1835,6 +1835,10 @@ def summary(self, shorten=False, name_padding=35):
...
+            nameunit = '{}'.format(self.attributes.get('STASH'))
```
bjlittle: `self.attributes['STASH']` ... if you wanna, as you've already checked that the `STASH` key is available.

marqh:
minor changes, as per review comments

lib/iris/cube.py
```diff
@@ -1835,6 +1835,9 @@ def summary(self, shorten=False, name_padding=35):
 
         nameunit = '{name} / ({units})'.format(name=self.name(),
                                                units=self.units)
+        # If all unknown and a STASH attribute exists, use it.
+        if (nameunit == 'unknown / (unknown)' and 'STASH' in self.attributes):
```
ajdawson: No need for parentheses when on one line.

ajdawson:
Does this affect how `Cube`s are printed too? I'm wary of removing a consistency like this, users expect `name / (unit)`, and although this is not necessarily directly useful, it does flag that something is missing. This may not be the case if this is replaced by a stash code.

marqh:
> Does this affect how Cubes are printed too? I'm wary of removing a consistency like this, users expect name / (unit), and although this is not necessarily directly useful, it does flag that something is missing. This may not be the case if this is replaced by a stash code.

this doesn't change cube print `cube.__str__`, but it does change `cube.__repr__`

I think this is a reasonable change, as most code will use print and this will still say `unknown / (unknown)` and then print the attributes, including STASH if available

the aim is to improve the printing of cube lists, where there is unknown name and unknown unit

ajdawson:
OK great, thanks for clarifying @marqh.

ajdawson:
One last thing then: did you consider `stash / (unknown)` and reject in favour of just the STASH? I can see both sides of the argument, the STASH is not the name, so it may not be appropriate to follow the `name / (unit)` convention, but on the other hand, we still can't give the units, so should we include that part?

marqh:
> One last thing then: did you consider stash / (unknown) and reject in favour of just the STASH? I can see both sides of the argument, the STASH is not the name, so it may not be appropriate to follow the name / (unit) convention, but on the other hand, we still can't give the units, so should we include that part?

I did consider this; in principal we shouldn't have a name without a unit, so if we're falling back to STASH, we already don't have a unit

i do wonder whether `stashmsi / (unknown)` might be more specific, but it is longer, and i think it will always say unknown, as, if there is a unit, we'll print that rather than a STASH code

ajdawson:
I guess the current strategy is fine, and really shines when someone knows what a STASH is and realises it probably tells them everything they need to know (or equips them with what they need to query with their data provider).

bjlittle:
@marqh Did you consider something like `stash=m01s00i004` rather than just the bare stash code (i.e. `m01s00i004`) being used instead of `unknown / (unknown)`

ajdawson:
It would certainly make it clear that the presented information was different somehow from the normal presentation of name and unit.

marqh:
> Did you consider something like stash=m01s00i004 rather than just the bare stash code (i.e. m01s00i004) being used instead of unknown / (unknown)

I am content to do this, if you think it helps, though I don't like the `=` sign in the string

I'm keen to converge on one solution, so far we have

1. m01s00i004
1. m01s00i004 / (unknown)
1. stash: m01s00i004
1. stash=m01s00i004

please vote for your favorite or nominate another representation.  I'll code up the preferred solution

marqh:
> please vote for your favorite or nominate another representation. I'll code up the preferred solution

@ajdawson @bjlittle ...

ty

ajdawson:
I don't like 3, there are too many colons already in cubelist printouts. Number 2 is my next least favourite (I know, I brought this up...). I'd go with number 4, but if there are objections to this then my next choice is plain old number 1 (i.e. no change).

bjlittle:
Apologies, I've been distracted ...

Okay, so I'm kinda in agreement with @ajdawson ... but I don't want this to be a thing, so old number 1 it is.

By golly I think we have a quorum!

Issue #2232, opened 2016-11-01
conda install error - please update install instructions


susoson:
The instructions at http://scitools.org.uk/iris/docs/latest/installing.html to install iris need improvement.

I first tried installing via conda install -c scitools iris in my ioos environment  - yikes!

So I created a new environment just for iris and ran: conda install -c scitools iris

Which as you know doesn't work - https://github.com/SciTools/iris/issues/1970

Using the conda forge install seems to have gotten me almost there. 

conda install -c conda-forge iris geos shapely

No errors associated with cartopy/dependencies/versions.

But when trying to run the first example: http://scitools.org.uk/iris/docs/latest/examples/General/SOI_filtering.html

I get the following error, which leads me to believe that there is another install step required for working with examples but I don't see reference to this anywhere in the documentation.

ValueError: Sample data file(s) at '/Users/jenniferpatterson/anaconda/envs/irisenv2/lib/python2.7/site-packages/iris/sample_data/SOI_Darwin.nc' not found.
NB. This function is only for locating files in the iris sample data collection. It is not needed or appropriate for general file access.

Please update the iris documentation with better instructions for installing and working with examples, I look forward to using iris but had hoped for smoother install/setup.

dkillick:
Hi @susoson - thanks for getting in contact with us and letting us know about the difficulties you had using the install documentation for Iris. With regard to this documentation, we deliberately kept it pretty terse as the install process when using Conda is, in most cases, as short as the documentation suggests. We also deliberately have not included any documentation on using Conda as it is unwise to duplicate documentation, especially when it is not your own (as is true here), because such duplicated documentation will just go stale and be the less useful. Instead we have included links to relevant and useful sections of the Conda documentation, which take you through the steps needed to get using Conda to the point where you could use it to install Iris as the Iris documentation demos.

The problem you're having running the examples comes from the fact that you need the Iris sample data. All the examples are based on data contained within the Iris sample data, which is available as a conda package on the SciTools channel:
```
conda install -c scitools iris_sample_data
```

This will work with Iris packages from SciTools and conda-forge. With this package installed you should be able to run the example code. It may be that the install documentation would benefit from a mention of the Iris sample data package - I will propose it as a PR and see what happens.

Hope this helps.

susoson:
Thanks @dkillick, the conda install of iris_sample_data now has me up and running! 

[...]

PR #2234, opened 2016-11-02
DOC: mention sample data in INSTALL

dkillick:
If you wish to run any of the code examples you need access to the Iris sample data, which can now be easily installed from Conda. This, however, may not be very clear to new users, so I'm proposing that we add mention of the Iris sample data to the `INSTALL` document.

Addresses part of #2232.

### DIFF
--- a/INSTALL
+++ b/INSTALL
@@ -18,7 +18,7 @@ the following command::
 
 If you wish to run any of the code examples
 (see http://scitools.org.uk/iris/docs/latest/examples/index.html) you will also
-need the Iris sample data. This can also be installed using Conda::
+need the Iris sample data. This can also be installed using conda::
 
   conda install -c scitools iris_sample_data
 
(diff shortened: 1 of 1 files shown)


ajdawson:
While you're in here, the first paragraph of the conda section claims we have 32-bit linux packages, but we don't (and neither do conda-forge). This should probably be corrected to avoid disappointment.

dkillick:
> the conda section claims we have 32-bit linux packages, but we don't

Nice spot @ajdawson - you're quite right. I'll get changing that.

INSTALL
```diff
@@ -16,6 +16,12 @@ the following command::
 
   conda install -c scitools iris
 
+If you wish to run any of the code examples
+(see http://scitools.org.uk/iris/docs/latest/examples/index.html) you will also
+need the Iris sample data. This can also be installed using Conda::
```
ajdawson: The capitalisation of conda is inconsistent, everywhere else it is all lower case. You should use lower case here too.

dkillick:
> The capitalisation of conda is inconsistent

@ajdawson nice spot - I've corrected that now.

PR #2238, opened 2016-11-09
Consistency in lon/lat return order

dkillick:
In `iris.analysis.cartography`, there are three functions that return lons and lats. Two of them return `lon(s), lat(s)`, one returns `(lat, lon)`. This changes the order of the latter to match the order of the former. The best thing is the latter (`_get_lat_lon_coords`, which I've renamed to the now more appropriate `_get_lon_lat_coords`) is a private method, so we don't need a major release for this change.

Amazingly enough, this addresses #51.

### DIFF
--- a/lib/iris/analysis/cartography.py
+++ b/lib/iris/analysis/cartography.py
@@ -162,7 +162,7 @@ def rotate_pole(lons, lats, pole_lon, pole_lat):
     return rotated_lon, rotated_lat
 
 
-def _get_lat_lon_coords(cube):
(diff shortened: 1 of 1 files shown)


marqh:
@dkillick you have missed a couple of tests which need to cahnge their pattern to use the renamed function

dkillick:
Thanks @pelson. Think I might have just beaten [your recent record](https://github.com/SciTools/iris/pull/2899#issuecomment-341489203)! 🏆

PR #2239, opened 2016-11-11
BUG: Expossing slicing bug with numpy

cpelley:
This PR demonstrates the issue of a [numpy bug](https://github.com/numpy/numpy/issues/8264) i.e. adds two failing unittests to iris.
Unfortunately, the workaround discussed in the numpy issue does not appear to work in all circumstances.

**Not for merging**

@marqh @jkettleb @hdyson

### DIFF
--- a/lib/iris/tests/unit/cube/test_Cube.py
+++ b/lib/iris/tests/unit/cube/test_Cube.py
@@ -1409,5 +1409,31 @@ class TestCellMeasures(tests.IrisTest):
             cm_dims = self.cube.cell_measure_dims(a_cell_measure)
 
 
+class Test___getitem__(tests.IrisTest):
+    def slicing_on_views(self, container):
+        # Check whether we are subject to the issues of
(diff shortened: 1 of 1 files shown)


marqh:
hello @cpelley 

this is not a test of Iris, this is a test of numpy;  i don't think this test belongs in the Iris source code, i think it belongs with 
https://github.com/numpy/numpy/issues/8264

this is not a pattern used by iris, as far as I can see.  Do you have examples of where Iris code makes use of this pattern that suffers from this behavioural problem?

mark

cpelley:
>  ...i don't think this test belongs in the Iris source code...

The purpose of this PR was not for merging but to demonstrate the numpy bug to iris users/developers which resulted from the recent dependency version increase on numpy by iris i.e. I wanted to increase its awareness.

Feel free to close this issue if you don't feel it provides any benefit.  Alternatively a known failure unittest in iris might be useful rather than assuming that we remember to keep track of this numpy issue.

PR #2242, opened 2016-11-16
Fix test runner override and no-data

bjlittle:
This PR corrects the use of the `override_test_data_repository` environment variable, and applies the standard naming convention of using upper-case environment variables :pray: 

It also fixes the broken test runner option `--no-data` to **actually** ignore the test data.

This corrects #2211 and addresses the ignored [comment](https://github.com/SciTools/iris/pull/2211#discussion_r84714014)

@dkillick and @lbdreyer


dkillick:
> addresses the ignored comment

To be fair, I'd already merged the PR by that point, if, apparently prematurely...

bjlittle:
@dkillick Yeah, I came to that party slightly late ...

PR #2246, opened 2016-11-24
New regridding scheme that uses scipy.interpolate.griddata

lbdreyer:
This is a first attempt at creating a new Iris regridding scheme that uses scipy.interpolate.griddata.

We think that most of the code is there, we're just in the middle of writing some tests.


lib/iris/analysis/_regrid.py
```diff
@@ -543,3 +544,336 @@ def __call__(self, src):
...
+        * src_cube:
+            The :class:`~iris.cube.Cube` providing the source points.
+        * tgt_grid_cube:
+            The :class:`~iris.cube.Cube` providing the target grid.
+        * method:
+            Either 'linear' or 'nearest'.
+        * projection:
+            The projection in which the interpolation is performed, or None. If
+            None, a XXX projection is used.
+        """
```
marqh: then again, perhaps making this in the __init__ makes sense

lib/iris/analysis/_regrid.py
```diff
@@ -543,3 +544,336 @@ def __call__(self, src):
...
+        * regrid_callback:
+            The routine that will be used to calculate the interpolated
+            values of any reference surfaces.
+
+        Returns:
+            The new, regridded Cube.
+
+        """
+        #
+        # XXX: At the moment requires to be a static method as used by
```
marqh: it would be helpful to understand why this is used by the experimental code. It's not clear to me why this should be a static method

lib/iris/analysis/_regrid.py
```diff
@@ -543,3 +544,336 @@ def __call__(self, src):
                                    sample_grid_x, sample_grid_y,
                                    regrid_callback)
         return result
+
+
+class UnstructuredProjectedRegridder(object):
```
marqh: it seems to me that there is code sharing taking place between this class and the rectilinear class

is this the case

how much effort is it to refactor to deliver a base class and two specialisations?

lib/iris/experimental/regrid.py
```diff
@@ -746,14 +751,10 @@ def regrid_area_weighted_rectilinear_src_and_grid(src_cube, grid_cube,
...
                                            area_func, circular, mdtol)
 
-    # Wrap up the data as a Cube.
-    # Create 2d meshgrids as required by _create_cube func.
-    meshgrid_x, meshgrid_y = np.meshgrid(grid_x.points, grid_y.points)
     regrid_callback = RectilinearRegridder._regrid
     new_cube = RectilinearRegridder._create_cube(new_data, src_cube,
                                                  src_x_dim, src_y_dim,
                                                  src_x, src_y, grid_x, grid_y,
-                                                 meshgrid_x, meshgrid_y,
```
marqh: i'm not convinced you meant to take these two out.  there are tests failing due to this

lib/iris/analysis/_regrid.py
```diff
@@ -452,7 +455,13 @@ def regrid_reference_surface(src_surface_coord, surface_dims,
                 warnings.warn(msg)
         return result
 
-    def _check_units(self, coord):
+    @staticmethod
+    def _check_units(coord):
```
marqh: i'd rather this wasn't changed to a static method, I think.

If we need to reuse the functionality, we should either factor it out to share, or reproduce a similar in experimental in the interim

lib/iris/analysis/__init__.py
```diff
@@ -2407,35 +2405,26 @@ def regridder(self, src_cube, target_grid):
...
             A callable with the interface:
-
                 `callable(cube)`
-
             where `cube` is a cube with the same grid as `src_cube`
             that is to be regridded to the `target_grid`.
-
         """
         from iris.analysis.trajectory import \
             UnstructuredNearestNeigbourRegridder
```
marqh: the changes left in this file are non-functional, as far as I can see.  If so, please may we revert them, to reduce the change noise?

lib/iris/experimental/regrid.py
```diff
@@ -1312,3 +1313,401 @@ def regridder(self, src_grid, target_grid):
 
         """
         return _CurvilinearRegridder(src_grid, target_grid, self.weights)
+
+
+class _ProjectedUnstructuredRegridder(object):
```
marqh: i am interested why this is private.  it is in experimental, I don't think it has to be further obfuscated.  Do you think it ought to stay private?

lib/iris/analysis/_regrid.py
lbdreyer: My reasoning behind trying to share as much code as possible was that it saved time spent writing unit tests since (hopefully) this method would already be tested.

Do you mean factor it out as in have it as a separate function?

lib/iris/experimental/regrid.py
lbdreyer: Thank you!

lib/iris/experimental/regrid.py
```diff
@@ -1312,3 +1313,401 @@ def regridder(self, src_grid, target_grid):
...
+                                            method=self._method,
+                                            projection=self._projection)
+
+        new_cube = self._create_cube(data, src_cube, src_xy_dim,
+                                     src_x_coord, src_y_coord,
+                                     tgt_x_coord, tgt_y_coord,
+                                     regrid_callback)
+
+        promote_aux_coord_to_dim_coord(new_cube, 'latitude')
+        promote_aux_coord_to_dim_coord(new_cube, 'longitude')
```
djkirkham: I don't think we can assume the coordinates are `latitude` and `longitude`; they could be `grid_latitude`, `projection_x_coordinate` etc., right?

Besides, I don't think this should be done here, I think we should refactor the `copy_coords` method in `_create_cube` to promote them there.

lbdreyer:
Note that this change requires https://github.com/SciTools/iris-test-data/pull/47 to be merged in

lib/iris/experimental/regrid.py
djkirkham: Should be able to add the following to `_create_cube` after the calls to `copy_coords`:

```
promote_aux_coord_to_dim_coord(result, grid_x_coord)
promote_aux_coord_to_dim_coord(result, grid_y_coord)
```

marqh: can we make use of the ability to get an 'x' like or 'y' like coord from a cube

`cube.coord(axis='X')`

marqh:
> Note that this change requires SciTools/iris-test-data#47 to be merged in

merged

lbdreyer:
I'm not sure why the last commit didn't build.

I'm going to try pushing it up again.

marqh:
excellent, thank you @lbdreyer

PR #2245, opened 2016-11-23
Test traj

marqh:
tidy up test trajectory code

PR #2247, opened 2016-11-25
Trajectory test tweak

pp-mo:
Some odd small improvements that seemed relevant, following on from #2245

### DIFF
--- a/lib/iris/analysis/_interpolate_private.py
+++ b/lib/iris/analysis/_interpolate_private.py
@@ -465,7 +465,7 @@ def regrid(source_cube, grid_cube, mode='bilinear', **kwargs):
 
     """
     if mode == 'bilinear':
-        scheme = iris.analysis.Linear(**kwargs)
+        scheme = Linear(**kwargs)
         return source_cube.regrid(grid_cube, scheme)
 
     # Condition 1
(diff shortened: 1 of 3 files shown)


lib/iris/analysis/trajectory.py
```diff
@@ -87,6 +87,7 @@ def __init__(self, waypoints, sample_count=10):
         self.length = sum([seg.length for seg in segments])
 
         # generate our sampled points
+        #: The trajectory points, as dictionaries of {coord_name: value}.
```
pp-mo: NOTE:  Includes the 'sample_points' property in the documentation listings

lib/iris/analysis/_interpolate_private.py
```diff
@@ -465,7 +465,7 @@ def regrid(source_cube, grid_cube, mode='bilinear', **kwargs):
 
     """
     if mode == 'bilinear':
-        scheme = iris.analysis.Linear(**kwargs)
+        scheme = Linear(**kwargs)
```
pp-mo: NOTE: 
existing version is "technically" wrong, because iris.analysis is not itself imported.
Not a real practical problem, though, as iris itself was imported.

marqh:
Hi @pp-mo 

please may you check this PR and if it is valid update it so we can look to merge it

mark

pp-mo:
Rebased.
I really couldn't understand how it got that problem, it's like the merge with master was only partial but I couldn't see why.
Let's hope this works !

marqh:
thanks @pp-mo

PR #2249, opened 2016-11-28
Cube sphere regridder

corinnebosley:


lib/iris/tests/integration/test_regridding.py
```diff
@@ -30,7 +30,10 @@
 from iris.coord_systems import GeogCS
 from iris.coords import DimCoord
 from iris.cube import Cube
-from iris.tests.stock import global_pp
+from iris.tests.stock import global_pp, simple_3d
+from iris.analysis import UnstructuredNearest
+
+iris.FUTURE.netcdf_promote = True
```
marqh: i think this is causing a problem with your tests
it is a global switch, so can have wide ranging effects.  to use it within a test, use it as a context manager
```
with iris.FUTURE.netcdf_promote = True:
    do_stuff()
```
or somesuch

corinnebosley:
@pp-mo @marqh Finally got the tests to pass (was missing a skipper), and I have removed the optimisation change that didn't work and wasn't necessary in this PR.

pp-mo:
I saw you'd addressed *almost* all pep8 issues with analysis.trajectory
Hence my [corinnebosley#3](https://github.com/corinnebosley/iris/issues/3)

**Otherwise** this all looks good to me now.
@corinnebosley please check that out + see if you're okay to merge that in here, too
 \- or if not, just say so + we can just skip it.
Either way, I'll merge this.

pp-mo:
[@corinnebosley#4](https://github.com/corinnebosley/iris/issues/4) shows a silly bug.

We clearly need some more tests on this.
Plus, there are other improvements we really need to the code  currently in the Unstructured.
 (E.G. bare "assert"s in the code; need to check passed cubes better; should handle additional properties of the passed cubes...).

I intend to address these by a separate effort ***after*** merging to master..
(promise :wink: )

pp-mo:
Moving too fast here, made another mistake.
We now need [@corinnebosley#5](https://github.com/corinnebosley/iris/issues/5) here, **too**.

PR #2253, opened 2016-11-30
Correct PP deprecation notices

dkillick:
The PP load/save pipeline has been substantially changed for v1.10, with a number of old bits of functionality being deprecated. Unfortunately the deprecation messaged for two of the deprecated functions (being `add_save_rules` and `reset_save_rules`) actually pointed the user to go and use another function (`as_pairs`) that had itself been deprecated by `save_pairs_from_cubes`.

This PR corrects the deprecation warning messages to point to the non-deprecated function rather than the deprecated function.

### DIFF
--- a/lib/iris/fileformats/pp.py
+++ b/lib/iris/fileformats/pp.py
@@ -2030,16 +2030,16 @@ def add_save_rules(filename):
     .. deprecated:: 1.10
 
         If you need to customise pp field saving, please refer to the functions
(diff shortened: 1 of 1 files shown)


QuLogic:
Aim for 1.11.1, or not bother because 1.12 was supposed to be out two days ago?

dkillick:
@QuLogic let's just stick it in master and see what releases it first...

dkillick:
Thanks @bjlittle!

PR #2257, opened 2016-12-02
Relax data length constraint in PP load

dkillick:
If the length of a PP field's data record as specified by the field's header and the integer word that follows the header do not match, Iris very reasonably currently fails to load the whole PP file.

This PR reduces the barrier in such a case so that the PP loader pipeline will raise a warning in such a case and then exit, returning all successfully loaded fields up to this point. This at least means that Iris users can get to most of a PP file rather than none of it.

This PR very much follows the approach introduced in #1444 for relaxing loading of bad PP headers.

~~Note: no testing.~~

### DIFF
--- a/lib/iris/fileformats/pp.py
+++ b/lib/iris/fileformats/pp.py
@@ -1950,10 +1950,12 @@ def _field_gen(filename, read_data_bytes, little_ended=False):
                 '%cL' % dtype_endian_char,
                 pp_file_read(PP_WORD_DEPTH))[0]
             if len_of_data_plus_extra != pp_field.lblrec * PP_WORD_DEPTH:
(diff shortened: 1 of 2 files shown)


bjlittle:
@dkillick Looks good ... and agrees with the strategy within `_field_gen` for dealing with other [bad](https://github.com/SciTools/iris/pull/2257/files#diff-066f828420b27bfd2a03868477b98415R1942) pp fields ...

When travis is happy, then so am I :smile: ...

bjlittle:
@dkillick Your change has tripped a test over ... that's good, right?

Can I leave you to investigate and fix, thanks!

dkillick:
Hey @bjlittle we've got the green light from Travis :+1:

PR #2261, opened 2016-12-08
ENH: Remove deepcopies when slicing cubes and copying coords

cpelley:
Summary:
- When indexing a cube when the data is reaslised (not lazy), a view of the original array is returned where possible (subject to the rules when slicing in numpy).
- When indexing a cube when the data is not reaslised (lazy), realising the data on one will still not realise the data on the other.
- Existing behaviour is that slicing coordinates returns views of the original points and bounds (where possible).  This was likely chosen behaviour on the basis that DimCoords at least are not writeable.  This is not the same however for Auxiliary coordinates and likely raises the likely case for this being a bug (i.e. one can modify AuxCoord object points and bounds).
- DimCoord slicing will now return views of its data like AuxCoords.  DimCoords will continue to realise data unlike AuxCoords due to the validation necessary for being monotonically increasing.

Replaces #1992


cpelley:
Happy to squash when ready

lib/iris/coords.py
```diff
@@ -516,13 +516,26 @@ def copy(self, points=None, bounds=None):
         .. note:: If the points argument is specified and bounds are not, the
                   resulting coordinate will have no bounds.
 
+        .. deprecated:: 1.12
+
+            By default the new coordinate's `points` and `bounds` will
```
pp-mo: I think this explains it from slightly the wrong angle : see the `Cube.__getitem__` comment...

lib/iris/coords.py
```diff
@@ -516,13 +516,26 @@ def copy(self, points=None, bounds=None):
...
             raise ValueError('If bounds are specified, points must also be '
                              'specified')
 
-        new_coord = copy.deepcopy(self)
+        if iris.FUTURE.share_data:
+            new_coord = copy.copy(self)
+            new_coord.attributes = copy.deepcopy(self.attributes)
+            new_coord.coord_system = copy.deepcopy(self.coord_system)
+        else:
+            new_coord = copy.deepcopy(self)
```
pp-mo: This branch should issue a deprecation warning.

lib/iris/cube.py
```diff
@@ -2174,23 +2182,31 @@ def new_cell_measure_dims(cm_):
...
 
         for other_slice in slice_gen:
             data = data[other_slice]
 
-        # We don't want a view of the data, so take a copy of it if it's
-        # not already our own.
-        if isinstance(data, biggus.Array) or not data.flags['OWNDATA']:
-            data = copy.deepcopy(data)
+        if not iris.FUTURE.share_data:
+            # We don't want a view of the data, so take a copy of it if it's
```
pp-mo: This branch should issue a deprecation warning.

lib/iris/coords.py
```diff
@@ -516,13 +516,26 @@ def copy(self, points=None, bounds=None):
         .. note:: If the points argument is specified and bounds are not, the
                   resulting coordinate will have no bounds.
 
+        .. deprecated:: 1.12
```
pp-mo: On a completely seperate issue from deprecations etc...
I'm a bit bothered about the scope of the new behaviour..

Firstly, I'm not convinced that it is expected for a method called "copy" to return a view at all.
I'm thinking that it would make more sense to have ```.copy()``` return a full copy, and use `[:]` for a toplevel-copy.
Isn't that what numpy does, so that is what we should emulate ?
E.G.
```
>>> a=np.arange(10)
>>> b=a[2:4]
>>> b[:] = 0
>>> a
array([0, 1, 0, 0, 4, 5, 6, 7, 8, 9])
>>> b=a.copy()
>>> b[:] = -1
>>> a
array([0, 1, 0, 0, 4, 5, 6, 7, 8, 9])
>>> 
```

Secondly, as I've mentioned it shouldn't ```Coord.__getitem__``` be returning copies ??
(that is, ***instead***, if you accept the previous...)

cpelley:
Thanks @pp-mo ah your onto something here, `Coord.__getitem__` returns a copy (i.e. calls `self.copy`).
I'll have a thinking about the logic of __getitem__ and copy and get back to you.

cpelley:
Hi @pp-mo, I think I have found a more reasonable behaviour for this views/copy business for the coords object (see the latest commit).  Summary:

1. `Coord.copy` now always returns a deepcopy but doesn't unnecessarily copy the original points before replacing it when supplying new points and bounds (this provides a significant memory optimisation especially in the case of multidimensional coordinates present).
2. `Coord.__getitem__` returns a view when possible, much like the behaviour of a numpy array.

I haven't updated the deprecation issues you mentioned yet (but I will do).  Let me know if your happy with the underlying chosen behaviour.

Cheers

pp-mo:
> I think I have found a more reasonable behaviour for this views/copy business for the coords object 
Thanks @cpelley 

I'm a bit confused now as to what we are (and are not) doing with coords...
 * Of course, unlike data, the coords data arrays are not modifiable, so we don't have to make copies just to avoid unintentional change to a 'parent' coord.
 * You are now addressing a gross issue with copying the coord points only to throw them away (!)
 * I think there is also another potentially large benefit from being able to index lazy coords without realising them (consider making cube slices, for example), but **that is simply not in scope at this point, is that right ??**

I'm also wondering whether you need to apply these tests to a DimCoord as well ?

pp-mo:
Meanwhile... 
the change in Cube looks good to me, but I think the code can be simplified.
After some effort, I convinced myself that this should be equivalent to your [lines 2185-2220](https://github.com/cpelley/iris/blob/281077934d8b8b02293522cdebb1e800903b424b/lib/iris/cube.py#L2185) ...
```
        try:
            first_slice = next(slice_gen)
        except StopIteration:
            first_slice = Ellipsis if iris.FUTURE.share_data else None

        if first_slice is not None:
            data = self._my_data[first_slice]
        else:
            data = copy.deepcopy(self._my_data)

        for other_slice in slice_gen:
            data = data[other_slice]
```

I think it passes the tests, anyway.
Does you think that is right ?

pp-mo:
@cpelley uncovered some unexpected behaviour (as discussed offline):
```
>>> co = iris.coords.AuxCoord(np.arange(5))
>>> co.points
array([0, 1, 2, 3, 4])
>>> co.points[2] = 77
>>> co.points
array([ 0,  1, 77,  3,  4])
>>> 
>>> co2 = co[2:4]
>>> co2.points
array([77,  3])
>>> co2.points[1] = 9999
>>> co.points
array([   0,    1,   77, 9999,    4])
>>> 
```

This only applies to an AuxCoord, as the DimCoord value arrays are not writeable.
We think this is a **bug** really : it is not _intended_ behaviour that changing a derived (sliced) coord can affect the original.

For the present case, we might consider _introducing_ this type of behaviour, i.e. a sliced coord is linked to the original.  We might also consider extending that to DimCoords.
However, from a practical, performance point of view, such behaviour won't achieve anything much in most cases.   That is, only if coords (a) are quite large (meaning, usually, multidimensional) and (b) have deferred data (which at present generally means loaded from netcdf).

pp-mo:
**NOTE:**
The change proposed here is practical, but doesn't go as far as making cubes behave "just like arrays" :
The original design completely precludes that by ensuring that views on cube data and coord values are ***never*** returned.
The proposed change, however, only affects the data at this point :  Even if we add other changes to let coords share points arrays, cubes will not share coordinate objects, so various things done to cube.slices() will still not work the way you might expect...

pp-mo:
**Another NOTE:**
Another thing that this _won't_ deliver is to have copies of a cube with deferred data "share" data when the original is realised (i.e. loaded)  :  Because we _replace_ "cube._my_data", the copied cubes still have deferred arrays, and when requested the data will be fetched again for each one.

cpelley:
Thanks @pp-mo, I'll update the ticket description.  There are so many subtleties that I think it worth going in some detail in the description for the record

> ...After some effort, I convinced myself that this should be equivalent to your lines 2185-2220 ...

Thanks @pp-mo, yes looks neater and simpler, thanks.

cpelley:
You will be pleased to know that I have found some further subtleties.  Some I have fixed and others I'm working on.
- DimCoords return copies of points and bounds (I have fixed this now).
- DimCoords realises its data while AuxCoords does not realise its data.  Working on fixing this now.

Gosh!

cpelley:
@pp-mo I think I'm ready.  The travis failure is unrelated.

```

FAIL: test_resolve (iris.tests.test_image_json.TestImageFile)

  File "/home/travis/miniconda/envs/test-environment/lib/python2.7/site-packages/Iris-1.12.0.dev0-py2.7-linux-x86_64.egg/iris/tests/test_image_json.py", line 95, in test_resolve

    self.assertEqual(deque(), exceptions)

AssertionError: deque([]) != deque([ConnectionError(MaxRetryError("HTTPSConnectionPool(host='scitools.github.io', port=443): Max retries exceeded with url: /test-iris-imagehash/images/155fb4e2e9a1a9a16da9b9e006fa9176078a965e0b869b5f0b8659751b807b75.png (Caused by NewConnectionError('<requests.packages.urllib3.connection.VerifiedHTTPSConnection object at 0x7f8260507050>: Failed to establish a new connection: [Errno 110] Connection timed out',))",),)])```

Cheers

**Comment by marqh** [comment]
Hi @cpelley 

your patience is appreciated, I hope you are still with us.

We have been having trouble with the above test failure in processing the preparation for the 1.12 release.

There is now a fix on the 1.12.x branch, which hasn't been merged back into master yet.

I'd like this functionality in 1.12 and to see the tests passing; so, please rebase and retarget (in a new PR is fine if that is easier) and we'll do our best to get this in

thank you
mark

**Comment by cpelley** [comment]
Thanks @marqh, I have changed base and squashed.  I still get troubles with travis however (though this time with the build).

**Comment by cpelley** [comment]
Test failures are due to unrelated (files untouched) license header failures (problems with 1.12.x).

**Comment by cpelley** [comment]
Rebased but I'm still getting what looks like unrelated problems...

```

ERROR: test_resolve (iris.tests.test_image_json.TestImageFile)

  File "/home/travis/miniconda/envs/test-environment/lib/python2.7/site-packages/Iris-1.12.0rc0-py2.7-linux-x86_64.egg/iris/tests/test_image_json.py", line 46, in test_resolve

    raise ValueError('Github API get failed: {}'.format(iuri))

ValueError: Github API get failed: https://api.github.com/repos/scitools/test-iris-imagehash/contents/images

-------------------- >> begin captured logging << --------------------

requests.packages.urllib3.connectionpool: DEBUG: Starting new HTTPS connection (1): api.github.com

requests.packages.urllib3.connectionpool: DEBUG: https://api.github.com:443 "GET /repos/scitools/test-iris-imagehash/contents/images HTTP/1.1" 403 None

--------------------- >> end captured logging << ---------------------
```

cpelley:
ping

cpelley:
Hi @marqh, I have rebased again and have the same error (looks unrelated to me??)

cpelley:
TimedOutException on the Python3.4 TARGET=DEFAULT tests :(

The example tests (i.e. TARGET=example) are failing because the warning about this deprecation is issued.  Before I look at this, do we really want the example tests failing just because deprecation warnings are issued??

marqh:
> The example tests (i.e. TARGET=example) are failing because the warning about this deprecation is issued. Before I look at this, do we really want the example tests failing just because deprecation warnings are issued??

this is an explicit choice, and one that I support

I do not want 'example' code using deprecated features

cpelley:
Finally got the time to get back to this.
All tests passing and no timeouts!

ping @marqh 

Cheers

cpelley:
Closed in favour of #2549

PR #2263, opened 2016-12-09
Fix import bug.

pp-mo:
My editor (Eclipse/PyDev) does some static analysis and it flags up [this line](https://github.com/SciTools/iris/blob/f10eb54d527cdf094c5f38179a800a68e0c9e19f/lib/iris/experimental/regrid.py#L1539) as a bug, because 'six' is not imported.
I guess this has never actually been exercised ?

I s'pose, ideally, all codepaths would be tested ...

### DIFF
--- a/lib/iris/experimental/regrid.py
+++ b/lib/iris/experimental/regrid.py
@@ -33,6 +33,7 @@ import numpy as np
 import numpy.ma as ma
 import scipy.interpolate
 from scipy.sparse import csc_matrix, diags as sparse_diags
+import six
 
 import iris.analysis.cartography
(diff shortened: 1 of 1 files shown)


djkirkham:
Oh dear... something for me and/or @lbdreyer to work on I guess.

lbdreyer:
Thanks for pointing this out @pp-mo !

We chose not to add so many tests for this bit of code because it was only in experimental... but I was concerned with this decision, and what you found proves the risk of making that decision...

I am shortly going to add a PR that includes a test for regridding cubes with aux_factories which would use this code (and would have flagged up the missing import).

PR #2264, opened 2016-12-09
Add scoping to docs.

pp-mo:
Just to clarify the usage scope of 'AreaWeighted'.
Existing notes are a bit wrong, it looks like they were copied from Linear/Nearest, which work in a different way and are not specific to X/Y dimensions.


marqh:
i agree, thanks @pp-mo

PR #2266, opened 2016-12-14
Added a new datetime format including seconds and used this to parse …

SusanJL:
I've modified the name_loaders to allow the reading of "seconds" in the date-times provided in NAME trajectory files.


marqh:
i have retargeted this at 1.12.x to try to get it included in the upcoming release

PR #2267, opened 2016-12-15
Add linear scipy projected regridder

lbdreyer:
This PR adds the linear scheme of the regridder that uses `scipy.interpolate.griddata`.

I have also added some more tests for the nearest-neighbour scheme as source cubes with aux_factories weren't being tested, as highlighted by @pp-mo in #2263 

Note the test data doesn't have much variation in the data within the layers themselves (between the layers there's quite a difference, hence the large values for the standard deviation which is calculated for the **_whole_** cube), so I was getting similar values for mean and standard deviation, regardless of whether I used the linear or nearest (see the test `TestProjectedUnstructured.test_use_default_rtol`, which I do intend to remove, I just added it to prove what I had found).
I have solved this problem by allowing the tolerance to be set but I don't think this is necessarily a great solution. Another option would be to calculate the mean and standard deviation for a level.

### DIFF
--- a/lib/iris/experimental/regrid.py
+++ b/lib/iris/experimental/regrid.py
@@ -1,4 +1,4 @@
-# (C) British Crown Copyright 2013 - 2016, Met Office
+# (C) British Crown Copyright 2013 - 2017, Met Office
 #
 # This file is part of Iris.
 #
--- a/lib/iris/experimental/regrid.py
+++ b/lib/iris/experimental/regrid.py
@@ -21,6 +21,7 @@ Regridding functions.
 
 from __future__ import (absolute_import, division, print_function)
 from six.moves import (filter, input, map, range, zip)  # noqa
+import six
 
 from collections import namedtuple
 import copy
--- a/lib/iris/experimental/regrid.py
+++ b/lib/iris/experimental/regrid.py
@@ -1549,7 +1549,7 @@ class _ProjectedUnstructuredRegridder(object):
                                                             grid_x_coord,
                                                             grid_y_coord,
                                                             regrid_callback)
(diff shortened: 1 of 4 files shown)


lib/iris/experimental/regrid.py
```diff
@@ -1549,7 +1550,7 @@ def regrid_reference_surface(src_surface_coord, surface_dims,
                                                             grid_x_coord,
                                                             grid_y_coord,
                                                             regrid_callback)
-                    result.add_aux_coord(result_coord, dims)
+                    result.add_aux_coord(result_coord, (dims[0], dims[0]+1))
```
lbdreyer: The benefits of actually testing your code! It highlights errors like this!

marqh:
Hi @lbdreyer 

given out current status, I would appreciate you targeting this against 1.12.x, rather than master
this has just had #2275 merged, so the chance of failure due to http retry is reduced for 1.12.x PRs

thank you
mark

marqh:
i'm pretty happy with the code and testing.
please may we have a what's new entry?

lbdreyer:
@marqh I have added a whatsnew entry which applies to the ProjectedUnstructured nearest and linear schemes.

I have also removed the test I used to show that testing with a low tolerance was giving the same results regardless of whether the linear or nearest scheme was used.
I have instead replaced this with testing against the whole cube (with a high tolerance), as well as a single layer.
([see the most recent commit](https://github.com/SciTools/iris/pull/2267/commits/7175798f71f83703c5cb5618df4893be942c3043))

marqh:
this looks good to go now, thanks @lbdreyer

PR #2241, opened 2016-11-14
Structured load api

pp-mo:
This now replaces the discussion PR on the "feature branch" in my repo : https://github.com/pp-mo/iris/pull/27

**NOTE NOT READY YET :** documentation changes still need to be reviewed/finalised/agreed.

from notes there:
> As everything else was already agreed + merged to here, I've merged the outstanding [pp-mo/iris#31 "Revised documentation, whatsnew and deprecations."](https://github.com/pp-mo/iris/pull/#31)  here, as that is all that remains to discuss.
So I closed this and opened a new PR to the main Iris master instead : #2241
We can discuss what remains to be discussed there.

PR #2269, opened 2016-12-16
Link callback rewrite example into structured-load docstring.

pp-mo:
Suggested as a good idea for #2241, but it got forgotten.


pp-mo:
replacing with re-targetted version, as I tried re-targetting this but I'm not convinced Travis is then testing the right thing

PR #2270, opened 2016-12-16
Tidy wrapper imports and reinstate docs for iris.analysis.interpolate.

pp-mo:
In response to : https://github.com/SciTools/iris/pull/1982#issuecomment-267242056
It seems when we deprecated the iris.analysis.interpolate functions, we also stopped building docs for them, which is unhelpful + was really not intended.


pp-mo:
replacing with re-targetted version, as I tried re-targetting this but I'm not convinced Travis is then testing the right thing

PR #2272, opened 2016-12-19
Tidy wrapper imports and reinstate docs for iris.analysis.interpolate.

pp-mo:
In response to : #1982 (comment)
It seems when we deprecated the iris.analysis.interpolate functions, we also stopped building docs for them, which is unhelpful + was really not intended.

Retry of #2270, now targetting v1.12.x


pp-mo:
Wanted for 1.12 release
@marqh ping
Can you see the sense in this ?

PR #2273, opened 2016-12-20
Unstructured regridder : usage tidying

pp-mo:
A few functional changes / fixes.
A lot of documentation and testing enhancements.

### DIFF
--- a/lib/iris/analysis/__init__.py
+++ b/lib/iris/analysis/__init__.py
@@ -69,7 +69,7 @@ __all__ = ('COUNT', 'GMEAN', 'HMEAN', 'MAX', 'MEAN', 'MEDIAN', 'MIN',
            'PEAK', 'PERCENTILE', 'PROPORTION', 'RMS', 'STD_DEV', 'SUM',
            'VARIANCE', 'WPERCENTILE', 'coord_comparison', 'Aggregator',
            'WeightedAggregator', 'clear_phenomenon_identity', 'Linear',
-           'AreaWeighted', 'Nearest')
+           'AreaWeighted', 'Nearest', 'UnstructuredNearest')
 
 
 class _CoordGroup(object):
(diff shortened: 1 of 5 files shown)


lib/iris/analysis/__init__.py
```diff
@@ -2348,13 +2348,31 @@ def regridder(self, src_grid, target_grid):
...
+    1-dimensional X and Y coordinates.
+
+    Source and target XY coordinates must have the same coordinate system,
+    which may also be None.
+    If any of the XY coordinates are latitudes or longitudes, then they *all*
+    must be.  Otherwise, the corresponding X and Y coordinates must have the
+    same units in the source and grid cubes.
+
+    .. Note::
+        Currently only supports regridding, not interpolation.
```
marqh: This distinction isn't immediately obvious to me: interpolation is a numerical method which can support a regrid operation.
As this is the docstring for a class called `UnstructuredNearest` I suggest that this note is simply not required

lib/iris/analysis/_interpolate_private.py
```diff
@@ -206,6 +206,11 @@ def _nearest_neighbour_indices_ndcoords(cube, sample_points, cache=None):
     Because this function can be slow for multidimensional coordinates,
     a 'cache' dictionary can be provided by the calling code.
 
+    .. Note::
+
+        If the points are longitudes/latitudes, these are handled correctly as
+        points on the sphere, but the values must be in 'degrees'.
```
marqh: is this protected somewhere?  if i supply radians, will a meaningful exception be raised?

lib/iris/analysis/trajectory.py
```diff
@@ -380,13 +403,15 @@ def __init__(self, src_cube, target_grid):
 
         * src_cube:
             The :class:`~iris.cube.Cube` defining the source grid.
-            The X and Y coordinates must be mapped over the same dimensions.
+            The X and Y coordinates can have any shape, but must be mapped over
```
marqh: excellent

lib/iris/tests/integration/test_regridding.py
```diff
@@ -106,7 +106,7 @@ def setUp(self):
 
     def test_nearest(self):
         res = self.src.regrid(self.grid, UnstructuredNearest())
-        self.assertArrayShapeStats(res, (1, 6, 3, 4), 315.888737, 11.000729)
+        self.assertArrayShapeStats(res, (1, 6, 3, 4), 315.890808, 11.000724)
```
marqh: are these changes expected?
the value differences are small, but non-zero

lib/iris/analysis/__init__.py
pp-mo: What 'regrid' and 'interpolate' these words mean in ***Iris*** terms is that Cubes support both a 'regrid' and an 'interpolate' method, which can both specify a "scheme".
UnstructuredNearest is a scheme, and these can *generally* be used with either interpolate or regrid
 \-- except that this one can't.  ( Similarly, neither can PointInCell ).

The practical difference is that a "regrid" uses a grid cube as its reference, while "interpolate" takes coordinate values.
This arrangement is now well established in Iris.
*But I do agree that the need for both is questionable, and the choice between is nowhere very clearly explained !!*

marqh: ok, I'm content with this

lib/iris/analysis/_interpolate_private.py
pp-mo: > is this protected somewhere? if i supply radians, will a meaningful exception be raised?

No, it is not checked and will just give wrong answers !

I decided to document this problem rather than fix it, as it is an existing low-level routine.
As this is a private routine I thought that was ok.
Arguably, we could maybe fix the existing analysis.trajectory.interpolate, as it calls this and so exposes the same problem.
? what do you think @marqh ?

I **have** fixed it in the UnstructuredNearest classes, which now convert everything to degrees (and complain if it fails).

marqh:
> It just goes to show you have to be careful with statistical tests.

indeed

marqh:
it seems there's no outstanding issues here, I am go for launch

cheers @pp-mo

PR #2274, opened 2016-12-20
Structured load pseudolevels

pp-mo:
This should allow structured-loading to work with most files where pseudo-levels are used.

If 'ordinary' levels (LBLEV values) and 'pseudo-levels' (LBUSER5) both vary at once, this rather inelegant solution could still have problems, potentially producing 2 vertical coordinates.
But only if that occurs ***within*** a phenomenon, which we think is probably not realistic ?

Note: 
Possibly needs tidying a bit further.
? Should I remove that commented-out 'fixed' failure code in ```tests.integration.fast_load.MixinProblemCases.test_FAIL_pseudo_levels``` ?
I have left it, assuming it is a useful statement of what the key problem actually is.

### DIFF
--- a/lib/iris/tests/integration/fast_load/test_fast_load.py
+++ b/lib/iris/tests/integration/fast_load/test_fast_load.py
@@ -93,7 +93,7 @@ class Mixin_FieldTest(object):
         return file_path
 
     def fields(self, c_t=None, cft=None, ctp=None,
-               c_h=None, c_p=None, phn=0, mmm=None):
+               c_h=None, c_p=None, phn=0, mmm=None, pse=None):
         # Return a list of 2d cubes representing raw PPFields, from args
         # specifying sequences of (scalar) coordinate values.
         # TODO? : add bounds somehow ?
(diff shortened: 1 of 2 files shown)


marqh:
> If 'ordinary' levels (LBLEV values) and 'pseudo-levels' (LBUSER5) both vary at once, this rather inelegant solution could still have problems, potentially producing 2 vertical coordinates.
> But only if that occurs within a phenomenon, which we think is probably not realistic ?

I agree, this seems like a pretty edgy edge case to me

lib/iris/tests/integration/fast_load/test_fast_load.py
```diff
@@ -583,6 +521,109 @@ def test_missing_combination(self):
...
+
+    def test_FAIL_pseudo_levels(self):
+        # Show how pseudo levels are handled.
+        flds = self.fields(c_t='000111222',
+                           pse='123123123')
+        file = self.save_fieldcubes(flds)
+        results = iris.load(file)
+        expected = CubeList(flds).merge()
+
+# NOTE: this problem is now fixed : Structured load gives the same answer.
```
marqh: do we need to hold onto these notes explicitly in the source?
I'd have thought that this is what the git history is for
may we remove this?

pp-mo: The problem is that the version 
Ok, I guess it is immortalised in the commit https://github.com/SciTools/iris/pull/2274/commits/33eb8d877201d7c63bef11fca87a249ee60f42f1
(though that won't be appearing in the main log).
I will tidy this up...

marqh: following an off line discussion about this I am persuaded to leave this in for now, as this commit will be squished so it will never be in teh history.

the handling of pseudo levels may require more effort, so these reminders may prove useful

PR #2282, opened 2017-01-04
Lbrel

marqh:
try except pattern for individual fields with missing lbrel values in Fields Files

these fields are skipped with a warning

requires 
https://github.com/SciTools/iris-test-data/pull/48
and updated test sha

### DIFF
--- a/lib/iris/fileformats/_ff.py
+++ b/lib/iris/fileformats/_ff.py
@@ -717,76 +717,85 @@ class FF2PP(object):
                 # read from the current FF LOOKUP table.
                 # (The PPField sub-class will depend on the header release
                 # number.)
-                field = pp.make_pp_field(header)
+                # (Some Fields File fields are UM specific scratch spaces, with
+                # no header release number, these will throw an exception from
+                # the PP module and are skipped to enable reading of the file.
+                try:
+                    field = pp.make_pp_field(header)
 
-                # Fast stash look-up.
-                stash_s = field.lbuser[3] // 1000
-                stash_i = field.lbuser[3] % 1000
-                stash = 'm{:02}s{:02}i{:03}'.format(field.lbuser[6],
-                                                    stash_s, stash_i)
-                stash_entry = STASH_TRANS.get(stash, None)
-                if stash_entry is None:
-                    subgrid = None
-                else:
-                    subgrid = stash_entry.grid_code
-                    if subgrid not in HANDLED_GRIDS:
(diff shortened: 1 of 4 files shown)


marqh:
@pp-mo 

this would be very useful for a user if it made it into 1.12.  I have added it to the milestone, please may you consider whether it is suitable and in time?

thank you
mark

lib/iris/fileformats/_ff.py
```diff
@@ -717,76 +717,85 @@ def _extract_field(self):
...
+                            # later.
+                            result_field._data = (self._filename, data_offset,
+                                                  data_depth, data_type)
 
-                    data_offset += data_depth
+                        data_offset += data_depth
 
-                    yield result_field
+                        yield result_field
+                except ValueError as valerr:
```
pp-mo: I think the code here is a bit too coupled to the pp module implementation.
It has the following problems :
 * we are *assuming* here that any ValueError must be an LBREL match, but that could easily change in the pp module  -- as we are creating a new field, the ```PPField.__init__``` call could easily add some other possible failure mode.
 * we are throwing away the detail of the provided message, so the user does not see the intended information about the reason (the LBREL value it didn't like).

I think it is a better idea to defer the *reason* for failure and the explanation to the PP module itself and just say here something like :
```
    msg = "Input field skipped because PPField creation failed : error = {!r}"
    raise ValueError(msg.format(str(valerr)))
```

lib/iris/fileformats/_ff.py
```diff
@@ -717,76 +717,85 @@ def _extract_field(self):
...
-                    fields = [field]
-                for result_field in fields:
-                    # Add a field data element.
-                    if self._read_data:
-                        # Read the actual bytes. This can then be converted to
-                        # a numpy array at a higher level.
-                        ff_file_seek(data_offset, os.SEEK_SET)
-                        result_field._data = pp.LoadedArrayBytes(
-                            ff_file.read(data_depth), data_type)
+                # (Some Fields File fields are UM specific scratch spaces, with
```
pp-mo: Shouldn't state here what the underlying problem is, as that is the PP module's business.
(see other comment)...
We should just say that the field creation failed here.

marqh:
> However, I think skipping the field is an equally reasonable response to any other possible problems ?

I think this is particularly the case for fields files, yes

the conversion to a PP object in our code may go wrong for a field, and at present that junks the whole load.

I have updated the skip message to pass through the PP exception, I hope this now represents a reasonable approach

PR #2283, opened 2017-01-05
retry github api get if it fails

marqh:


marqh:
i do not understand the test failures, it is like the merge has gone wrong; or something.

i'm leaving this untouched whilst i investigate, i have replaced the functionality with #2285

Issue #2281, opened 2017-01-04
Fix import bug in idiff


pp-mo:
I think this got fixed in one place, but recent merge has brought the problem back.
Here  : https://github.com/SciTools/iris/blob/b05bf5314d5be2cf9a4afb35bae40b04f6e4e95b/lib/iris/tests/idiff.py#L144

(Checked: same in master and v1.12.x)

lbdreyer:
Fixed in #2284

PR #2284, opened 2017-01-05
Fix import usage in idiff.

pp-mo:
Addresses #2281


lbdreyer:
Looks good. Thanks @pp-mo

PR #2286, opened 2017-01-05
Lbrelwn

marqh:
* clean up legacy contributions folders
* add 1.12 contribution for #2282


lbdreyer:
Good idea cleaning up the old contribution folders. I had been meaning to do that!

PR #2288, opened 2017-01-06
Reinstated lost whatsnew items, now to appear in the 1.12 whatsnew.

pp-mo:
I think this fixes a problem introduced by #2286
These changes were proposed before recent releases, but only now merged, so the whatsnew contributions now belong to 1.12.

Actually, I think it would be better if we did ***not*** include the version in the name of the contributions directory, but just had a fixed name for it.  Then we wouldn't need to destroy + recreate it either.
It's not a trivial change though, so I've made a separate issue to cover it : #2289


pp-mo:
@marqh can you check whether this makes sense ?

pp-mo:
Note: associated dates in the filenames now seem odd, because they are before earlier releases.
But that is ok, as they are only used to time-order the contributions in the 'aggregate_directory.py' operation, after which the date info is lost.

Issue #2280, opened 2017-01-04
Unstructured-regrid remaining issues


pp-mo:
Stuff left over from recent work refactoring the trajectory regridder and adding ```iris.analysis.UnstructuredNearest``` and ```iris.analysis.trajectory.UnstructuredNearestNeigbourRegridder```
I'm assuming for now these are **all** wanted for 1.12. 

 * [ ] support masked data and add integration test on ORCA data (which is masked !)
 * [x] add a whatsnew for UnstructuredNearest
    * #2290 addresses
 * [ ] add a whatsnew for changes to existing trajectory behaviour (like masking)

**NOTE:** for forthcoming RC1, only _second_ item is needed : brief, preliminary whatsnew for UnstructuredNearest

PR #2290, opened 2017-01-09
More docs for UnstructuredNearest

pp-mo:
Add whatsnew, and explain difference to ProjectedUnstructuredNearest.

Addresses one of the items in #2280.
That should be good enough for us to cut an RC1.


pp-mo:
@lbdreyer does this sound right to you ?

pp-mo:
Failures are fixed by #2291 
I will rebase when that is in ...

pp-mo:
At last it passes !
( Several respins required to get around test timeouts, which is getting tiresome. )

what you think @bjlittle ?

PR #2292, opened 2017-01-09
Document pseudo-level support in structured um loading.

pp-mo:
Just explains the temporary solution introduced in #2274

### DIFF
--- a/lib/iris/fileformats/um/_fast_load.py
+++ b/lib/iris/fileformats/um/_fast_load.py
@@ -1,4 +1,4 @@
-# (C) British Crown Copyright 2016, Met Office
+# (C) British Crown Copyright 2016 - 2017, Met Office
 #
 # This file is part of Iris.
 #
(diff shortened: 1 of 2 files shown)


pp-mo:
Problems still to fix.
Fixing ...

pp-mo:
... fixed
( but several timeout respins needed :angry: )

Can you check it out @marqh ?

lib/iris/fileformats/um/_fast_load.py
```diff
@@ -373,16 +373,18 @@ def structured_um_loading():
         fields of each phenomenon, such that different phenomena may have
         different field structures, and can be interleaved in any way at all.
 
+        .. note::
+
+             At present, fields with different values of 'LBUSER5'
+             (pseudo-level) are *also* treated as different phenomena.
+             This is a temporary fix, standing in place for a more correct
```
marqh: i don't think i would use the term 'temproary fix'

I would prefer to state what the behaviour is and how it will be experienced.

perhaps
`This will result in separate cubes for differnet values of LBUSER, with a coordinate representing the value.  These can be concatenated/merged to form cubes with a pseudo level dimension.`
or somesuch

PR #2295, opened 2017-01-10
github api user agent

marqh:
hopefully this will mitigate the risk that github api calls return a 403 forbidden

the docs suggest this should always have happened, if you read them a certain way
https://developer.github.com/v3/#user-agent-required

so i'm not sure this is the cause/fix for intermittent 403s, we shall see


pp-mo:
Now respinning the 'minimal' tests a few times, to exercise this fix a bit more ...

pp-mo:
> respinning the 'minimal' tests a few times

Done that + all seems solid.
Much nicer than a retry.

:clap: Awesome righteous fix @marqh !

PR #2297, opened 2017-01-10
Set clabel inline=False to fix contout plot

lbdreyer:
The example in the userguide section for [plotting a contour of a cube](http://scitools.org.uk/iris/docs/latest/userguide/plotting_a_cube.html?highlight=clabel#cube-contour) is not correct. The are spurious horizontal lines across the image.

Not being terribly familiar with how clabel works, I suspect that this is due to clabel removing some of the contour near where it wants to put the label, so that the label is more visible.
By setting inline to be False, clabel does not remove any of the contour so it renders correctly but it does make it a little harder to see

air_temp_inline_false

This is not an iris issue, I think the issue is in matplotlib itself. Below I have plotted the data in the cube using just matplotlib and I still get an error with missing contours
clabel_behaviour

### DIFF
--- a/docs/iris/src/userguide/plotting_examples/cube_contour.py
+++ b/docs/iris/src/userguide/plotting_examples/cube_contour.py
@@ -18,6 +18,6 @@ contour = qplt.contour(temperature_cube)
 plt.gca().coastlines()
 
 # Add contour labels based on the contour we have just created.
-plt.clabel(contour)
+plt.clabel(contour, inline=False)
 
 plt.show()


marqh:
I am just rebrowsing this PR and I am quite minded to keep this on the 1.12.x milestone and try to get it into the release

it is a bug fix and the bug is quite problematic in many cases

@lbdreyer the what's new is a real file now, please may you add an entry to it regarding this

@bjlittle @pp-mo would you be comfortable putting this into the release without it being part of rc1 or would you like an rc2 if we merge this?

mark

dkillick:
Did this seriously not cause any test failures?? That's a bit worrying...

marqh:
> Did this seriously not cause any test failures?? That's a bit worrying...

we don't test plot outputs in the user guide testing, so this won't have come up

as this is only a user guide change, I don't think it requires a what's new at all, i think we should just merge this

going once...
twice...

dkillick:
> this is only a user guide change

And a bugfix to a user guide at that. I agree, let's get it in; it's a useful improvement.

> we don't test plot outputs in the user guide testing

Maybe we should? Would have shown up this bug before a user guide reader brought it to our attention...

PR #2317, opened 2017-01-18
Zonal mean bounds

jonseddon:
This PR is required because when zonal means are loaded from a PP file (i.e. lbproc contains 64) the CF Standard says that cell bounds must be set for any axis that is referred to by a cell method (see the final sentence of the third paragraph of http://cfconventions.org/cf-conventions/v1.6.0/cf-conventions.html#cell-methods). This currently doesn't happen and it breaks the downstream CMIP6 processing tools that we don't have control of.

I'm not that experienced with Iris development and so apologies about anything that I've done that's not Iris like. Please let me know of any problems and I'll try and fix them

For the tests, https://github.com/SciTools/iris-test-data/pull/49 is required.


marqh:
Hi @jonseddon 

many thanks for the proposal, this seems like a logical and sensible change to me.

Please may you move your new test case into the PP integration test module
https://github.com/SciTools/iris/blob/master/lib/iris/tests/integration/test_pp.py
add the class (with the data skipper) it at the bottom of this module?

I have just merged the new test data file.  
https://github.com/SciTools/iris-test-data/commits/master
you will need to add the commit sha1 into
https://github.com/SciTools/iris/blob/master/.travis.yml#L25

to make sure the testing picks up this new file

hopefully the tests will then pass for us

thank you
mark

jonseddon:
Hi @marqh,

I made those changes but looking at the Travis logs (https://travis-ci.org/SciTools/iris/jobs/194505888#L151), the previous .travis.yml file is being used as the old SHA1 value is shown rather than the new value (942e6344d6e6d00c85d824dadd49b5d6f57f4b4a) that I just committed. Can you suggest how I should get the updated .travis.yml to be used?


Thanks,

Jon

marqh:
> I made those changes but looking at the Travis logs (https://travis-ci.org/SciTools/iris/jobs/194505888#L151), the previous .travis.yml file is being used as the old SHA1 value is shown rather than the new value (942e6344d6e6d00c85d824dadd49b5d6f57f4b4a) that I just committed. Can you suggest how I should get the updated .travis.yml to be used?

Hi @jonseddon 

I am really struggling to understand this.  i can see the change in your code and it is not being used in Travis, i cannot fathom how.

I don't see why this should happen

I am happy, the change looks fine.  Perhaps, as we work nearby, you could pop over monday/tuesday and we could fix this between us?

marqh:
closed by #2336

Issue #2308, opened 2017-01-17
(prior) netCDF loading


pp-mo:
Load to a cube with dask array content

pp-mo:
**Relevant code sections**
```
fileformats
    _pyke_rules."fc_rules_cf.krb"
        build_auxiliary_coordinate : OrthoArrayAdapter
        build_cell_measures : OrthoArrayAdapter
    netcdf
        _load_cube() : OrthoArrayAdapter
        Saver._create_cf_data_variable : save
```

marqh:
progress so far:
https://github.com/SciTools/iris/compare/master...marqh:dasking?expand=1

(remember to rebuild or hack the pyke rules)

marqh:
i chose to ignore unit tests, as we are changing behaviour, so:
`python -m unittest discover -v -s iris.tests.integration`


```
======================================================================
ERROR: test_round_trip (iris.tests.integration.test_netcdf.TestCellMeasures)
----------------------------------------------------------------------
Traceback (most recent call last):
  File "/home/marky/coding/scitools/iris/lib/iris/tests/integration/test_netcdf.py", line 275, in test_round_trip
    iris.save(cube, filename, unlimited_dimensions=[])
  File "/home/marky/coding/scitools/iris/lib/iris/io/__init__.py", line 402, in save
    saver(source, target, **kwargs)
  File "/home/marky/coding/scitools/iris/lib/iris/fileformats/netcdf.py", line 2247, in save
    least_significant_digit, packing=packspec)
  File "/home/marky/coding/scitools/iris/lib/iris/fileformats/netcdf.py", line 967, in write
    self._add_cell_measures(cube, cf_var_cube, dimension_names)
  File "/home/marky/coding/scitools/iris/lib/iris/fileformats/netcdf.py", line 1150, in _add_cell_measures
    cm)
  File "/home/marky/coding/scitools/iris/lib/iris/fileformats/netcdf.py", line 1497, in _create_cf_cell_measure_variable
    data = self._ensure_valid_dtype(cell_measure.data, 'coordinate',
  File "/home/marky/coding/scitools/iris/lib/iris/coords.py", line 1749, in data
    return data.view()
TypeError: view() missing 1 required positional argument: 'dtype'

======================================================================
ERROR: test_lazy_mask_preserve_fill_value (iris.tests.integration.test_netcdf.TestLazySave)
----------------------------------------------------------------------
Traceback (most recent call last):
[...]

PR #2318, opened 2017-01-18
Dask pp

pp-mo:
Note "temporary" testing approach : just selected new tests, for the new code functions : 
```python -m unittest discover -v lib/iris/tests/integration/temp_dask```

Addresses #2307

### CODE REVIEW

**Comment by marqh** [comment]
Hi @pp-mo 

please may you target the dask feature branch with such changes in the short term, rather than master?  

thank you

**Comment by pp-mo** [comment]
> target the dask feature branch 

Sorry, forgot about that !

**Comment by pp-mo** [comment]
**WIP**
I'm going to try amending the travis.yml to get it to install dask + run the temporary tests...

**Comment by pp-mo** [comment]
Selected tests now working.

Hoping you like it, as I'm wanting to build on this testing strategy in further work ...

File: lib/iris/util.py [hunk 1]
```diff
@@ -1,4 +1,4 @@
-# (C) British Crown Copyright 2010 - 2016, Met Office
+# (C) British Crown Copyright 2010 - 2017, Met Office
```
dkillick: What happened here?

lib/iris/tests/integration/temp_dask/test_pp_lazy.py
```diff
@@ -0,0 +1,103 @@
...
+        self.assertTrue(cube.has_lazy_data())
+        cube.data
+        self.assertFalse(cube.has_lazy_data())
+
+    def test_lazy_data(self):
+        # Check cube.lazy_data().
+        cube = self.cube
+        raw_data = cube._my_data
+        lazy = cube.lazy_data()
+        self.assertIs(cube.lazy_data(), raw_data)
```
dkillick: Did you want to check the type of `lazy` here too? I don't think you've checked that `cube.lazy_data` returns a dask object.

If you do choose to do this then this test method will get quite long, with lots of assertions, so it would be worth looking into splitting it up a bit.

lib/iris/tests/integration/temp_dask/test_pp_lazy.py
```diff
@@ -0,0 +1,103 @@
...
+        cube.data
+        lazy = cube.lazy_data()
+        self.assertIsNot(cube.lazy_data(), raw_data)
+        self.assertArrayAllClose(lazy.compute(), raw_data.compute())
+
+    def test_lazy_data__set(self):
```
dkillick: To me this name implies you're going to run a `set()` operation on the lazy data.

lib/iris/fileformats/pp.py
```diff
@@ -1642,12 +1641,8 @@ def __eq__(self, other):
...
                 attrs = [hasattr(self, attr), hasattr(other, attr)]
                 if all(attrs):
-                    self_attr = getattr(self, attr)
-                    other_attr = getattr(other, attr)
-                    if isinstance(self_attr, biggus.NumpyArrayAdapter):
-                        self_attr = self_attr.concrete
-                    if isinstance(other_attr, biggus.NumpyArrayAdapter):
-                        other_attr = other_attr.concrete
+                    self_attr = as_concrete_data(getattr(self, attr))
+                    other_attr = as_concrete_data(getattr(other, attr))
```
pp-mo: Well, I think we really want to be explicit here that we are realising the content.

In fact ```(self_attr == other+attr)``` ***is*** itself a lazy object, i.e. it hasn't looked at the data yet.
I fully expected that you would be *required* to call 'compute' + nothing else will do it (as with biggus),
however ***in fact*** it seems that applying np.all() will realise it + do the compare.
In my view that is not nice, and may even be a bug, as it's not documented anywhere -- see [comment on #2308](https://github.com/SciTools/iris/issues/2308#issuecomment-273499060)

lib/iris/util.py
pp-mo: I made some changes, then didn't need them + reverted them all in a subsequent commit.
But then you can't revert the date without the licence check failing.
I think you *can* remove this when you squish.

lib/iris/tests/integration/temp_dask/test_pp_lazy.py
pp-mo: Ok, I'll rename these functions "test_lazy_data__new_content"...

lib/iris/tests/integration/temp_dask/test_pp_lazy.py
```diff
@@ -0,0 +1,103 @@
...
+#
+# Iris is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU Lesser General Public License for more details.
+#
+# You should have received a copy of the GNU Lesser General Public License
+# along with Iris.  If not, see <http://www.gnu.org/licenses/>.
+"""
+Test lazy data handlingin iris.fileformats.pp.
```
pp-mo: Will fix!

lib/iris/tests/integration/temp_dask/test_pp_lazy.py
pp-mo: I didn't want to be explicit about the type, having hidden details behind the iris._lazy_data interface. 
It also doesn't seem right to use ```is_lazy_data``` when we are also testing that here in another class.
I give the game away anyway by calling ".compute()" on them just below.

lib/iris/_lazy_data.py
```diff
@@ -0,0 +1,79 @@
...
+    If lazy, return the realised data, otherwise return the argument unchanged.
+
+    """
+    if is_lazy_data(data):
+        data = data.compute()
+    return data
+
+
+# A magic value, borrowed from biggus
+_MAX_CHUNK_SIZE = 8 * 1024 * 1024 * 2
```
bjlittle: @pp-mo Difficult to know what this should default to when it's unknown what type of operation is going to be performed, as the choice of chunking should really be aligned with the expected operation/use in order to be optimal (from what I understand)

lib/iris/cube.py
```diff
@@ -1630,8 +1631,8 @@ def lazy_data(self, array=None):
 
         """
```
bjlittle: @pp-mo Should we care about updating the doc-string at this point ?

lib/iris/cube.py
```diff
@@ -1694,7 +1695,8 @@ def data(self):
                 msg = msg.format(self.shape, data.dtype)
                 raise MemoryError(msg)
             # Unmask the array only if it is filled.
-            if isinstance(data, np.ndarray) and ma.count_masked(data) == 0:
+            if (isinstance(data, np.ma.masked_array) and
+                    ma.count_masked(data) == 0):
```
bjlittle: @pp-mo It's not possible to put a masked array into a `dask.array.Array` (well, certainly one that **does** have its mask set) ... so is this really still valid in the new `dask` world?

Are you imagining that a user has a non-masked masked array wrapped up in a `dask.array.Array` ?

lib/iris/_lazy_data.py
pp-mo: Well this is obviously a preliminary.
I propose to not worry about it, for now !

lib/iris/cube.py
pp-mo: At the moment I'm just totally ignoring the masked issue.
This code is effectively a dead branch with the latest changes, as ```as_concrete_data``` will never return a masked result.
But this will need fixing later, so I left it.

lib/iris/cube.py
pp-mo: Ok, I might as well...

lib/iris/fileformats/pp.py
bjlittle: @dkillick Yup, I can confirm that `dask` maintains the laziness when comparing lazy objects; it just generates another lazy dask graph, which you need to realize to get the answer. So the following gives a lazy `result`:
```python
result = getattr(self, attr) == getattr(other, attr)
```
And the answer is realized with:
```python
>>> result.compute()
array([ True,  True,  True,  True,  True,  True,  True,  True,  True,  True], dtype=bool)
```
Or indeed, as @pp-mo suggests, the lazy dask result is made concreate by `np.all`:
```python
>>> np.all(result)
True
```

lib/iris/_lazy_data.py
bjlittle: Agreed

lib/iris/cube.py
bjlittle: As discussed at length :smile:

lib/iris/tests/integration/temp_dask/test_lazy_utils.py
```diff
@@ -0,0 +1,79 @@
...
+
+
+import numpy as np
+import dask.array as da
+
+
+from iris._lazy_data import is_lazy_data, as_lazy_data, as_concrete_data
+
+
+class MixinLazyTests(object):
```
bjlittle: @pp-mo `MixinLazyTestData` really ... kinda, sorta, rather than tests.

lib/iris/tests/integration/temp_dask/test_lazy_utils.py
```diff
@@ -0,0 +1,79 @@
...
+class Test_is_lazy_data(MixinLazyTests, tests.IrisTest):
+    def test_lazy(self):
+        self.assertTrue(is_lazy_data(self.lazy_array))
+
+    def test_real(self):
+        self.assertFalse(is_lazy_data(self.real_array))
+
+
+class Test_as_lazy_data(MixinLazyTests, tests.IrisTest):
+    def test_lazy(self):
```
bjlittle: @pp-mo This is really a `test_lazy_pass_thru` ...

lib/iris/tests/integration/temp_dask/test_lazy_utils.py
pp-mo: ok

lib/iris/tests/integration/temp_dask/test_lazy_utils.py
pp-mo: renamed now (coming soon)...

pp-mo:
Latest hopefully addresses the outstanding review comments ?
Please re-review @dkillick @bjlittle

marqh:
ok, late to the party here, but I am fully on board with this implementation, given it's target is the 'dask' feature branch

i am highly minded to merge this and work from here, e.g. in #2319 

any objections to a merge, i'll aim to merge GMT am tomorrow otherwise

PR #2323, opened 2017-01-20
Replace masked cube data with nanarrays.

pp-mo:
**EXPERIMENTAL ONLY : DO NOT MERGE !**

This is a tryout for how we might implement a changed approach to missing data representation.
Something similar could allow us to use dask as our underlying lazy operations support.

Rationale:
Dask doesn't support masked arrays at all.
However, a lot of Iris (especially analysis) relies on them.
In particular, we explicitly handle masking in our merging operations, which are at present all performed using biggus.ArrayStack .

So, the idea here is :
 * translate any masked cube data uniformly to an array with NaNs in instead
 * translate back again when fetching (or realising), so that existing Iris code still gets data it can understand.

### DIFF
--- a/lib/iris/cube.py
+++ b/lib/iris/cube.py
@@ -1,4 +1,4 @@
-# (C) British Crown Copyright 2010 - 2016, Met Office
+# (C) British Crown Copyright 2010 - 2017, Met Office
 #
 # This file is part of Iris.
 #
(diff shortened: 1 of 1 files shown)


pp-mo:
**More background and practical notes**
There are quite a few test errors, of course, but most of those are understandable.
*Some* of those can be fixed, but certain operations will simply yield somewhat different results.
This would be a "breaking change", so we can probably we would ***expect and accept*** that
 -- i.e. some "new behaviour" will make perfect sense in some cases.

Known problems:
 * no good solution for masked *integer* results
   * one proposal : user can transform to floating-point types, with known fill-values
 * we will need to carry fill-values to output data in the same way as it read in (e.g. NetCDF and PP)
   * we don't have a universal strategy for that, and there probably isn't a perfect one anyway
 * we are lumping together masks and NaNs, and can no longer distinguish the two
   * which certain existing operations actually do (see test failures).
 * probable inefficiencies in some cases

pp-mo:
Ping @bjlittle 
I'm hoping this covers all the main bits from our discussion today.
What do you think?

PR #2324, opened 2017-01-23
More stable area weights computation

ajdawson:
This PR changes the computation of area weights to a more numerically stable form. The previous form converted latitude to colatitude and used difference of cosines in the cell area computation. This formulation uses latitude and difference of sines. The conversion from latitude to colatitude at lower precision causes errors when computing the cell areas (see [this google group thread](https://groups.google.com/forum/?fromgroups=#!topic/scitools-iris/yTw387ZXRAk)). This should be considered a bug fix in my opinion.

I've provided 2 new tests, the 32-bit test fails with the old formulation but passes with the new one. All the existing tests still pass.

### DIFF
--- a/docs/iris/src/whatsnew/1.12.rst
+++ b/docs/iris/src/whatsnew/1.12.rst
@@ -104,6 +104,8 @@ Iris 1.12 Features
 
 * The transpose method of a Cube now results in a lazy transposed view of the original rather than realising the data then transposing it.
 
+* The :func:`iris.analysis.cartography.area_weights` function is now more accurate for single precision input bounds.
+
 Deprecations
 ============
 * The module :mod:`iris.experimental.fieldsfile` has been deprecated, in favour
--- a/lib/iris/analysis/cartography.py
+++ b/lib/iris/analysis/cartography.py
@@ -1,4 +1,4 @@
-# (C) British Crown Copyright 2010 - 2016, Met Office
+# (C) British Crown Copyright 2010 - 2017, Met Office
 #
 # This file is part of Iris.
 #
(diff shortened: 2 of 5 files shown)


ajdawson:
I forgot license header check locally. I will fix that and update.

ajdawson:
> All the existing tests still pass.

Not true after all :disappointed:. Two things:

1. I missed the old test.
2. It looks like the `iris.experimental.regrid` module is using the private `_quadrant_area` function from outside the `iris.analysis.cartography` module... Can be fixed easy enough, but this external usage surprised me.

ajdawson:
OK looks better now, got some seemingly unrelated test failures from testing images, not sure what is happening there.

@marqh - Can we squeeze this into v1.12 please since it is a bug fix.

marqh:
Hi @ajdawson 

it is a bit close to the line for 1.12.0, but let's talk it through

* what makes this calculation more stable?
* the numerical differences in https://github.com/SciTools/iris/pull/2324/files#diff-add4e098694aa6d2db30b2ef6f6f2680L645 are noticeable, how can we provide confidence to users that they are improved?

please may you add some text directly to the 'nearly complete' what's new for 1.12 describing this change, so we can consider and make a judgement

many thanks

ajdawson:
It is more stable because we don't perform an addition operation on the latitude bounds (to convert to co-latitude) before computing the difference of cosines. Instead we just compute a difference of sines on the bounds as represented on the cube. This make absolutely no difference in double precision, but eliminates a noticeable error in single precision.

The way you can be confident the new method is more stable is to compute area weights of a cube with a symmetric latitude dimension in single precision. The old method will produce asymmetric weights (i.e. a grid cell spanning 89N-90N will have a different area to a grid cell spanning 89S-90S, which is clearly incorrect) and the new method will produce symmetric weights. This is what the new unit tests are checking. The different values for the existing test you mention are because that test uses 32-bit input bounds, and the previous results were slightly in error (again, you could verify by using a larger extent to see asymmetry in the weights for a symmetric latitude dimension).

The calculations themselves are mathematically equivalent, but rounding errors are less of a problem when you remove the addition operation.

ajdawson:
Any happier @marqh?

marqh:
> Any happier @marqh?

yes, thank you

but no, as now the tests are failing, this is intolerable

there is an update on master which may help, though it is not a fix.  i can't get a handle on the github api atm, perhaps we need to ditch it

in the meantime, please rebase onto master for me, in case that helps
and, sorry...

marqh:
> in the meantime, please rebase onto master for me, in case that helps

apologies again, don't do this

this change is on master and not 1.12.x

please leave this as is, the tests passed fine the last time they run and you have only added docs

marqh:
Hi @ajdawson 

so, this has been a frustration to me for some time.

I spent a while investigating the github API and understanding it's rate limits and options for putting tokens into travis.
the rate limit is harsh (60 hits per hour per IP) and prone to problems from IP sharing on Travis, which we have no control over.

my compromise option is to ask first, and only GET if there are remaining hits left for that IP
#2328 

I think this is a reasonable approach.  Authenticating on PRs is a potential security leak hole and does not seem like good practice.

if you are content, please may you look to merge #2328 then rebase onto it?  Hopefully your tests will then pass.

We can swiftly merge this back onto `master` so all developers benefit from this from there

much obliged

ajdawson:
@marq - I am on vacation, I won't have time to review #2328. Since the tests here already passed for all code commits, can we not let this issue with Github API hold up this particular PR?

marqh:
@bjlittle is on the case with #2328 and has fixed this. \o/

given the time pressures, I agree, let's take this as is and let the CI testing deal with the retest for us

marqh:
all tests passing, excellent

thank you @ajdawson

marqh:
@lbdreyer @bjlittle I think we are all sorted on the 1.12 milestone
https://github.com/SciTools/iris/pulls?q=is%3Aopen+is%3Apr+milestone%3Av1.12.x

i think the release is good to go now :)

PR #2326, opened 2017-01-25
github api management

marqh:
attempt to bypass github 403 for anonymous get, possibly caused by travis ip addresses hammering the API


marqh:
i'm not convinced this is a fix, the failures are sporadic at the best of times; i've skipped them on minimal, which seems to make sense

we'll get more error reporting this way, but not much else

further steps may involve token auth, which is implementable, but needs care, as it is a potential security hole

PR #2328, opened 2017-01-27
github api care

marqh:
improvement on #2326 (reverted by #2327)

only use anonymous github api get if there are rate limit calls remaining

risk:  an image not added to the image repo is referenced in a new test and missed; likelihood: low, this isn't going to happen often and will likely be caught quickly

benefit: the majority of PRs, which have no impact on this test coverage, will try to run this test, but only run it if it will not fail due to github's harsh rate limiting controls for an IP

the `tests.skip_data` ensures this doesn't run in minimal test mode, so the number of calls from a single test set run is 2, not 4

### DIFF
--- a/lib/iris/tests/test_image_json.py
+++ b/lib/iris/tests/test_image_json.py
@@ -30,6 +30,7 @@ import json
 import logging
 import os
 import requests
+import unittest
 import time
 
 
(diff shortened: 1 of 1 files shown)


lib/iris/tests/test_image_json.py
```diff
@@ -34,32 +34,47 @@
...
+        headers = {'User-Agent': 'scitools-bot'}
+        rate_limit_uri = 'https://api.github.com/rate_limit'
+        rl = requests.get(rate_limit_uri, headers=headers)
+        some_left = False
+        if rl.status_code == 200:
+            rates = rl.json()
+            remaining = rates.get('rate', {})
+            remaining = remaining.get('remaining')
+            if isinstance(remaining, int):
+                if remaining > 3:
```
bjlittle: @marqh What's the magic `3` ...

lib/iris/tests/test_image_json.py
```diff
@@ -34,32 +34,47 @@
...
-                'https://scitools.github.io/test-iris-imagehash/images:\n{}')
-        amsg = amsg.format('\n'.join(list(uris.difference(known_image_uris))))
+            amsg = ('Images are referenced in imagerepo.json but not published'
+                    ' in https://scitools.github.io/test-iris-imagehash/'
+                    'images:\n{}')
+            diffs = list(uris.difference(known_image_uris))
+            amsg = amsg.format('\n'.join(diffs))
 
-        self.assertTrue(uris.issubset(known_image_uris), msg=amsg)
+            self.assertTrue(uris.issubset(known_image_uris), msg=amsg)
```
bjlittle: @marqh Perhaps if we are rate limited and can't run the test, do we want to issue a warning to say that's the case ? At least that would be captured in the Travis CI log ...

lib/iris/tests/test_image_json.py
marqh: i thought about >0 but then there's a race condition between the python2 tests and the python3 tests

so, 3 is an engineering tolerance

lib/iris/tests/test_image_json.py
marqh: > At least that would be captured in the Travis CI log ...

I did think about this, but the travis log is very busy, i wasn't sure it would be easy to find

i can add it in easily if you think it adds value

marqh: it might be neater to implement a conditional skip pattern

```
@unittest.skipIf(mylib.__version__ < (1, 3),
                     "not supported in this library version")
```

marqh: at least then it shows as skipped, rather than a warning lost in the noise

lib/iris/tests/test_image_json.py
```diff
@@ -34,32 +34,47 @@
...
-        headers = {'User-Agent': 'SciTools'}
-        r = requests.get(iuri, headers=headers)
-        if r.status_code != 200:
-            raise ValueError('Github API get failed: {}'.format(iuri))
-        rj = r.json()
-        prefix = 'https://scitools.github.io/test-iris-imagehash/images/'
+        headers = {'User-Agent': 'scitools-bot'}
+        rate_limit_uri = 'https://api.github.com/rate_limit'
+        rl = requests.get(rate_limit_uri, headers=headers)
+        some_left = False
```
bjlittle: @marqh Perhaps rename `some_left` to `api_rate_limited`, and negate the logic ?

lib/iris/tests/test_image_json.py
bjlittle: Cool, could I push you a add a comment :wink:

marqh:
@bjlittle skip pattern implemented, thus hopefully improved logging

comments added too

bjlittle:
@marqh Thanks :+1:

PR #2360, opened 2017-02-10
Fix 1-dim label data

ocefpaf:
I am hitting a 

```python
TypeError: sequence item 0: expected str instance, bytes found
```

with my dataset and I believe that, based on the logic in https://github.com/SciTools/iris/blob/master/lib/iris/fileformats/cf.py#L705, this should be `b' '`  here too,


ocefpaf:
Thanks @bjlittle!

PR #2363, opened 2017-02-12
remove pp_packing.c

ocefpaf:
I believe that with https://github.com/SciTools/iris/blob/master/minimal-conda-requirements.txt#L27 `iris` no longer needs `pp_packing.c`, right?


marqh:
Hello @ocefpaf 

we have removed the build of this from build channels and deprecated it, but I would rather delay the removal of this until a major version change.  I hope there will be such a major version change, potentially this will be the next release.

I would like to hold off on this change until the 'major version' is confirmed and we are handling the list of deprecation changes as a set.

ocefpaf:
No problem @marqh. The only reason I sent this is b/c it was a low hanging fruit and I love PRs that removes code :smile: 

Feel free to close this or wait for the right moment. (As far as I know both `conda-forge` and SciTools recipes are already using the `mo_pack` and removed the dependency on `numpy`'s ABI from `iris`.)

pelson:
Looks like this got done in another PR in #2689.
Thanks for this @ocefpaf - sorry we missed it and you didn't get an Iris commit from it 😞

ocefpaf:
No problem. I'm anxiously waiting for the new version with dask ;-)

Issue #2362, opened 2017-02-12
Fix masked/unmasked test behaviour.


pp-mo:
Emerging test problem
PR : #2361

Contributes to #2364

PR #2365, opened 2017-02-12
Dask abstract lazy cubedata

pp-mo:
Get tests passing after moving to 'generic lazy data handling for cube data operations.

Changes added on top of #2356.
Also depends on #2362

**What this includes...**
  * allow the iris._lazy_data module to tolerate + understand (realise) Biggus as well as Dask arrays (temporarily)
  * add routines for converting masked-arrays to+from NaN-arrays
    * at present, 'array_as_lazy' not used by any standard code, so this should *only* affect existing data where it has NaNs
  * addressed test failures + errors with use of ```@skip_biggus```


marqh:
following long discussions in the team we are adopting this implementation in the short term.  We will trial it with the pp and netCDF loaders and the merge & concatenate capabilities

further design and implementation decisions will be made following those activities

PR #2369, opened 2017-02-14
dask based merge

marqh:
* [x] dask based merge
 * [x] pp load
   * [x] check ff load
 * [x] netCDF load
 * [x] concatenate

note to reviewers, please do not squash these commits.  If you would like a cleaned up history, please request a rebase

### DIFF
--- a/lib/iris/_merge.py
+++ b/lib/iris/_merge.py
@@ -1078,16 +1078,16 @@ def _multidim_daskstack(stack):
 
     """
     if stack.ndim == 0:
-        # Handle array scalar inputs, as biggus does.
-        dask_components = stack[()]
+        # A 0-d array cannot be merged.
+        result = stack.item()
     elif stack.ndim == 1:
         # 'Another' base case : simple 1-d goes direct in dask.
-        dask_components = list(stack)
+        result = da.stack(list(stack))
     else:
         # Recurse because dask.stack does not do multi-dimensional.
-        dask_components = [_multidim_daskstack(subarray)
-                           for subarray in stack]
-    return da.stack(dask_components)
+        result = da.stack([_multidim_daskstack(subarray)
+                           for subarray in stack])
+    return result
 
 
 class ProtoCube(object):
--- a/lib/iris/_merge.py
+++ b/lib/iris/_merge.py
@@ -1,4 +1,4 @@
-# (C) British Crown Copyright 2010 - 2016, Met Office
+# (C) British Crown Copyright 2010 - 2017, Met Office
 #
 # This file is part of Iris.
 #
@@ -29,13 +29,14 @@ import six
 from collections import namedtuple, OrderedDict
 from copy import deepcopy
 
-import biggus
+import dask.array as da
 import numpy as np
 import numpy.ma as ma
 
 import iris.cube
 import iris.coords
 import iris.exceptions
+from iris._lazy_data import is_lazy_data, as_concrete_data, as_lazy_data
 import iris.util
 
 
@@ -1068,6 +1069,27 @@ def derive_space(groups, relation_matrix, positions, function_matrix=None):
     return space
 
 
+def _multidim_daskstack(stack):
+    """
+    Recursively build a multidensional stacked dask array.
+
+    The argument is an ndarray of dask arrays.
+    This is needed because dask.array.stack only accepts a 1-dimensional list.
+
+    """
+    if stack.ndim == 0:
+        # Handle array scalar inputs, as biggus does.
+        dask_components = stack[()]
+    elif stack.ndim == 1:
+        # 'Another' base case : simple 1-d goes direct in dask.
+        dask_components = list(stack)
+    else:
+        # Recurse because dask.stack does not do multi-dimensional.
+        dask_components = [_multidim_daskstack(subarray)
+                           for subarray in stack]
+    return da.stack(dask_components)
+
+
 class ProtoCube(object):
     """
     Framework for merging source-cubes into one or more higher
@@ -1192,10 +1214,10 @@ class ProtoCube(object):
         # Generate group-depth merged cubes from the source-cubes.
         for level in range(group_depth):
             # Stack up all the data from all of the relevant source
-            # cubes in a single biggus ArrayStack.
+            # cubes in a single dask "stacked" array.
             # If it turns out that all the source cubes already had
-            # their data loaded then at the end we can convert the
-            # ArrayStack back to a numpy array.
+            # their data loaded then at the end we convert the stack back
+            # into a plain numpy array.
             stack = np.empty(self._stack_shape, 'object')
             all_have_data = True
             for nd_index in nd_indexes:
@@ -1204,17 +1226,19 @@ class ProtoCube(object):
                 group = group_by_nd_index[nd_index]
                 offset = min(level, len(group) - 1)
                 data = self._skeletons[group[offset]].data
-                # Ensure the data is represented as a biggus.Array and
-                # slot that Array into the stack.
-                if isinstance(data, biggus.Array):
+                # Ensure the data is represented as a dask array and
+                # slot that array into the stack.
+                if is_lazy_data(data):
                     all_have_data = False
                 else:
-                    data = biggus.NumpyArrayAdapter(data)
+                    data = as_lazy_data(data)
                 stack[nd_index] = data
 
-            merged_data = biggus.ArrayStack(stack)
+            merged_data = _multidim_daskstack(stack)
             if all_have_data:
-                merged_data = merged_data.masked_array()
+                # All inputs were concrete, so turn the result back into a
+                # normal array.
+                merged_data = as_concrete_data(merged_data)
                 # Unmask the array only if it is filled.
                 if (ma.isMaskedArray(merged_data) and
                         ma.count_masked(merged_data) == 0):
--- a/lib/iris/fileformats/_pyke_rules/fc_rules_cf.krb
+++ b/lib/iris/fileformats/_pyke_rules/fc_rules_cf.krb
@@ -1002,7 +1002,6 @@ fc_extras
 
     import warnings
 
-    import biggus
     import cf_units
     import netCDF4
     import numpy as np
@@ -1018,6 +1017,7 @@ fc_extras
     import iris.exceptions
     import iris.std_names
     import iris.util
+    import iris._lazy_data
     
     
     #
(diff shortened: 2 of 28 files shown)


marqh:
@djkirkham @corinnebosley 
please may you add netcdf and pp dask implementations as PRs to this PR

marqh:
@pp-mo 
your perspective on the test status would be really useful at this stage

pp-mo:
Skimming though tests (just the ...
  * the full tests are too confusing, so I'm looking at the TEST_MINIMAL ones for now.
  * ```iris.tests.integration.fast_load...``` : most failures are due to the use of concatenate, but also the FieldCollation object ".data" fetcher is using biggus.  Nothing concerning here I think
  * ```iris.tests.unit.fileformats.pp.test__interpret_field..Test__interpret_fields__land_packed_fields...``` (*4 fails) : all fail at pp.py line 1886.  Something nasty is done by the mock-testing setup here -- the field._data object has a 0 where it should have a numpy dtype.  The mocked PPDataProxy object then does not support fetching its '.dtype'.  A biggus.NumpyArrayAdapter wrapped around this **also** fails if you try to extract its dtype.  Presumably, creating the biggus adapter does not *ask* the dtype, but in dask it does.  I expect you can fix this by providing a sensible dtype in the test.
    * STOP PRESS: if in the tests setUp, you set ```self.pp_field._data[3]``` to ```np.dtype('f4')```, and ```self.land_mask_field._data[3]``` to ```np.dtype('i4')```, these tests do pass.
  *

lib/iris/fileformats/pp.py
```diff
@@ -1286,8 +1286,8 @@ def data(self):
 
         """
         # Cache the real data on first use
-        if isinstance(self._data, biggus.Array):
-            data = self._data.masked_array()
+        if iris._lazy_data.is_lazy_data(self._data):
+            data = iris._lazy_data.as_concrete_data(self._data)
             if ma.count_masked(data) == 0:
```
pp-mo: I don't think this line works any more, as the result as ```as_concrete_data``` doesn't always yield a masked array,

marqh: i think that as_concrete_data always returns a masked array now

lib/iris/tests/integration/test_aggregated_cube.py
```diff
@@ -37,15 +35,16 @@ def test_agg_by_aux_coord(self):
...
+        # This test then ensures that aggregating using `points` instead is
+        # successful.
 
-        # First confirm we've got a `biggus.Array`.
+        # First confirm we've got a lazy array.
         # NB. This checks the merge process in `load_cube()` hasn't
         # triggered the load of the coordinate's data.
         forecast_period_coord = cube.coord('forecast_period')
-        self.assertIsInstance(forecast_period_coord._points, biggus.Array)
+        self.assertTrue(forecast_period_coord._points.all(),
```
pp-mo: Something gone wrong here ?
I think this means "unless all points are zero, raise [the function] ```iris._lazy_data.is_lazy_data``` as an error message :  It does not _call_ the function.
I'm also not sure what .all() does : does this produce a computed equality check, or is a dask deferred object ?

Equivalent to old code I ***think*** should be just
```
self.assertTrue(iris._lazy_data.is_lazy_data(forecast_period_coord._points))
```

**Stop press:** I checked out dask ```.all()``, and it does return a _deferred_ object.
This probably means that we should be using "array.all()" more in place of "np.all(array)", as the array function generalises to work properly on either kind of object.

lib/iris/tests/test_netcdf.py
```diff
@@ -111,7 +113,8 @@ def test_load_rotated_xy_land(self):
         cube = iris.load_cube(tests.get_data_path(
             ('NetCDF', 'rotated', 'xy', 'rotPole_landAreaFraction.nc')))
         # Make sure the AuxCoords have lazy data.
-        self.assertIsInstance(cube.coord('latitude')._points, biggus.Array)
+        self.assertTrue(cube.coord('latitude')._points.all(),
```
pp-mo: Same problem as https://github.com/SciTools/iris/pull/2369/files#r101506208

lib/iris/tests/test_pp_module.py
```diff
@@ -44,15 +44,15 @@ def setUp(self):
     def test_copy_field_deferred(self):
         field = next(pp.load(self.filename))
         clone = field.copy()
-        self.assertIsInstance(clone._data, biggus.Array)
+        self.assertIsInstance(clone._data, dask.array.Array)
```
pp-mo: Would rather this used ```is_lazy_data()```.

lib/iris/tests/test_pp_module.py
```diff
@@ -44,15 +44,15 @@ def setUp(self):
...
         self.assertEqual(field, clone)
         clone.lbyr = 666
         self.assertNotEqual(field, clone)
 
     def test_deepcopy_field_deferred(self):
         field = next(pp.load(self.filename))
         clone = deepcopy(field)
-        self.assertIsInstance(clone._data, biggus.Array)
+        self.assertIsInstance(clone._data, dask.array.Array)
```
pp-mo: Would rather this used is_lazy_data().

lib/iris/tests/unit/fileformats/grib/message/test_GribMessage.py
```diff
@@ -182,7 +182,7 @@ def _test(self, scanning_mode):
              6: SECTION_6_NO_BITMAP,
              7: {'codedValues': np.arange(12)}})
         data = message.data
-        self.assertIsInstance(data, biggus.Array)
+        self.assertTrue(data, iris._lazy_data.is_lazy_data)
```
pp-mo: Incorrect usage again, as https://github.com/SciTools/iris/pull/2369/files#r101506208

lib/iris/tests/unit/fileformats/pp/test__create_field_data.py
```diff
@@ -62,18 +63,19 @@ def test_deferred_bytes(self):
...
         # the biggus array (it's a private attribute), so instead we
         # patch the proxy creation and check it's being created and
         # invoked correctly.
         with mock.patch('iris.fileformats.pp.PPDataProxy') as PPDataProxy:
             PPDataProxy.return_value = proxy
             pp._create_field_data(field, data_shape, land_mask)
-        # Does the biggus array look OK from the outside?
-        self.assertIsInstance(field._data, biggus.Array)
+        # Does the dask array look OK from the outside?
+        self.assertIsInstance(field._data, da.core.Array)
```
pp-mo: Use is_lazy_data ?

lib/iris/tests/unit/fileformats/um/fast_load_structured_fields/test_FieldCollation.py
```diff
@@ -82,7 +83,7 @@ def test_t1_varies_faster(self):
              _make_field(lbyr=2013, lbyrd=2001, data=3),
              _make_field(lbyr=2014, lbyrd=2001, data=4),
              _make_field(lbyr=2015, lbyrd=2001, data=5)])
-        data = collation.data.ndarray()
+        data = collation.data.compute()
```
pp-mo: Maybe use "as_concrete_data" ?

lib/iris/tests/unit/fileformats/um/fast_load_structured_fields/test_FieldCollation.py
```diff
@@ -95,7 +96,7 @@ def test_t2_varies_faster(self):
              _make_field(lbyr=2014, lbyrd=2000, data=3),
              _make_field(lbyr=2014, lbyrd=2001, data=4),
              _make_field(lbyr=2014, lbyrd=2002, data=5)])
-        data = collation.data.ndarray()
+        data = collation.data.compute()
```
pp-mo: Maybe use "as_concrete_data" ?

lib/iris/tests/test_pickling.py
```diff
@@ -48,8 +48,8 @@ def pickle_then_unpickle(self, obj):
             yield protocol, reconstructed_obj
 
     def assertCubeData(self, cube1, cube2):
-        np.testing.assert_array_equal(cube1.lazy_data().ndarray(),
-                                      cube2.lazy_data().ndarray())
+        np.testing.assert_array_equal(cube1.lazy_data().compute(),
```
pp-mo: Maybe use "as_concrete_data" ?

marqh:
> ome notes just regarding the tests.
> (Core code still to do).
> I think there are a few genuine mistakes here.

agreed, i have attempted to address all of these

lib/iris/fileformats/_pyke_rules/fc_rules_cf.krb
```diff
@@ -1623,34 +1623,34 @@ fc_extras
         # Get units
         attr_units = get_attr_units(cf_coord_var, attributes)
 
-        def cf_var_as_biggus(cf_var):
+        def cf_var_as_array(cf_var):
```
pp-mo: Is it worth reconsidering the name here ?
I'd say that "array" sounds too much like numpy + should be replaced by "lazy" -- where before we had "biggus".

lib/iris/fileformats/_pyke_rules/fc_rules_cf.krb
```diff
@@ -1695,16 +1695,16 @@ fc_extras
         # Get units
         attr_units = get_attr_units(cf_cm_attr, attributes)
 
-        def cf_var_as_biggus(cf_var):
+        def cf_var_as_array(cf_var):
```
pp-mo: Rename as [above](https://github.com/SciTools/iris/pull/2369/files#r101523690) ?

lib/iris/fileformats/pp.py
pp-mo: Agreed

lib/iris/fileformats/pp.py
```diff
@@ -1644,10 +1644,6 @@ def __eq__(self, other):
                 if all(attrs):
                     self_attr = getattr(self, attr)
                     other_attr = getattr(other, attr)
-                    if isinstance(self_attr, biggus.NumpyArrayAdapter):
```
pp-mo: Why do we think we no longer need a realisation step here 
 -- as in "if is_lazy_data(x): x=as_concrete_data(x)" for both attrs ??

lib/iris/fileformats/um/_fast_load_structured_fields.py
```diff
@@ -89,9 +90,13 @@ def data(self):
             self._calculate_structure()
         if self._data_cache is None:
             data_arrays = [f._data for f in self.fields]
-            self._data_cache = \
-                ArrayStack.multidim_array_stack(data_arrays,
-                                                self.vector_dims_shape)
+            vector_dims_list = list(self.vector_dims_shape)
```
pp-mo: I find this a wee bit confusing.
I wonder if we could usefully employ the '_multidim_daskstack' routine from iris._merge instead ?
It is a recursive rather than iterative algorithm, which is less efficient but maybe clearer, and we would have the advantage that it is done once.  ( Actually I think it lacks unit testing, but that is another issue)
I ***think*** the equivalent would be something like ...
```
data_arrays = np.array([f._data for f in self.fields], dtype=object)
data_arrays = data_arrays.reshape(self.vector_dims_shape)
self._data_cache = _multdim_daskstack(data_arrays)
```

lib/iris/_merge.py
```diff
@@ -1068,6 +1069,27 @@ def derive_space(groups, relation_matrix, positions, function_matrix=None):
     return space
 
 
+def _multidim_daskstack(stack):
```
pp-mo: I think this needs more testing.
Probably a purpose-built unit test.
Could be deferred to another ticket for now, though, if it is working for all existing tests.

lib/iris/_merge.py
```diff
@@ -1068,6 +1069,27 @@ def derive_space(groups, relation_matrix, positions, function_matrix=None):
...
+    """
+    Recursively build a multidensional stacked dask array.
+
+    The argument is an ndarray of dask arrays.
+    This is needed because dask.array.stack only accepts a 1-dimensional list.
+
+    """
+    if stack.ndim == 0:
+        # A 0-d array cannot be merged.
+        result = stack.item()
```
pp-mo: I'm still not totally sure this bit is right.
Surely for recursive calling we always need to return a dask object, so we can stack several of those ??
See call for more research [above](https://github.com/SciTools/iris/pull/2369/files#r101527126).

lib/iris/fileformats/pp.py
pp-mo: Update: works owing to odd behaviour of np.all()
Note added to #2385

marqh:
joint decision @marqh @pp-mo @corinnebosley @AlexHilson to take this onto the dask feature branch and work from there

PR #2376, opened 2017-02-15
Don't use lazy wrappers for cube shape and dtype.

pp-mo:
This fixes the severe test slownesses we have been seeing.

I think there was ***no*** good reason to implement cube.dtype and cube.shape via cube.lazy_data().
The cube._my_data will always have viable dtype and shape properties, whether real or lazy.


pp-mo:
**Closed + retargetted at marqh/dmerge** : https://github.com/marqh/iris/pull/37

PR #2386, opened 2017-02-17
Lazy array wrapper

djkirkham:
Here is my proposal for re-factoring the code. Now `Cube._my_data` is a wrapper object which wraps the lazy and 'real' data. The idea is that this wrapper would implement an array interface to allow it to be used in cube operations. I haven't made any effort to make the tests pass, it's just to get feeback at this point.

The advantages of this approach, as I see it are:
- We could easily swap out the underlying lazy data provider (e.g dask or Biggus) by implementing the same interface with a new provider
- We don't need to worry about what the underlying provider is in the main code and whether it implements the required methods and functions
- We never expose the underlying provider to the users, only the wrapper object (could be seen as a disadvantage).

Disadvantages:
- It will require a lot of boilerplate code to implement the array interface
- The way it's written, when constructing a cube I've added a 'lazy' argument which is needed to specify if the data is lazy, which makes the interface awkward to use. This could probably be changed though.
- This maintains the existing "touch the data" behaviour, but following our discussion today this might need to be changed.


pp-mo:
Thanks @djkirkham .  Here's a few thoughts..
I find I don't have any very firm opinion on whether this is needed or not.

**For user code,**
I think as long as the user is clear that they are specifically manipulating lazy data, then they should simply see it for what it is.  As far as I can see, the only benefit of exposing lazy data to the user is so they can do things with it that Iris doesn't itself support.  In that case, the use of an implementation-independent 'wrapper' object is simply self-defeating : The user doing lazy calculations wants to deal with the actual implementing module, pretty much by definition,

**In Iris code,**
I agree we should in some places make clearer distinctions between code processing 'real' arrays and 'lazy' arrays.  This approach certainly does help with that, but it is in our hands anyway without truly _requiring_ a cleaner approach
Equally, I'm not terribly convinced about further abstraction helping that much with any future change in the underlying technology :  This approach localises the changes, but the significant problems are those where the capabilities don't quite match (like our present problems with masked data):  At that point, the contracts of the 'wrapper' api break down and change spreads into the wider codebase.

djkirkham:
@pp-mo Thanks for the feedback

> As far as I can see, the only benefit of exposing lazy data to the user is so they can do things with it that Iris doesn't itself support.

What about using it to construct another cube without loading the data? Maybe it's not a common use-case though.

>  The user doing lazy calculations wants to deal with the actual implementing module, pretty much by definition,

I see your point. My concern is that any code written with the underlying implementation in mind will be invalidated if/when the underlying implementation changes. And I don't think it would be too hard to provide the all the operations which biggus currently supprts, sot the user wouldn't necessarily lose functionality. Anyway, I think it would be quite easy to change this so that `cube.lazy_data()` returns the underlying object, and any methods and functions taking lazy data accept the underlying object.

> In Iris code,

I agree with your points here, but don't they equally apply to the current approach (on the dask branch) of using the functions in `_lazy_data.py`?

PR #2392, opened 2017-02-20
Data first

marqh:
for discussion regarding how to implement this interface


lib/iris/fileformats/netcdf.py
```diff
@@ -501,12 +506,12 @@ def _load_cube(engine, cf, cf_var, filename):
...
-    fill_value = getattr(cf_var.cf_data, '_FillValue',
-                         netCDF4.default_fillvals[cf_var.dtype.str[1:]])
+    fill_value = getattr(cf_var.cf_data, '_FillValue', None)
+
     proxy = NetCDFDataProxy(cf_var.shape, dummy_data.dtype,
                             filename, cf_var.cf_name, fill_value)
-    data = iris._lazy_data.as_lazy_data(proxy)
-    cube = iris.cube.Cube(data)
+    data = da.from_array(proxy, chunks=100)
+    cube = iris.cube.Cube(data, fill_value=fill_value)
```
marqh: should this pass the data type in as well?

lib/iris/fileformats/rules.py
```diff
@@ -901,15 +902,15 @@ def _make_cube(field, converter):
...
     except AttributeError:
         data = field.data
-
     cube = iris.cube.Cube(data,
                           attributes=metadata.attributes,
                           cell_methods=metadata.cell_methods,
                           dim_coords_and_dims=metadata.dim_coords_and_dims,
-                          aux_coords_and_dims=metadata.aux_coords_and_dims)
+                          aux_coords_and_dims=metadata.aux_coords_and_dims,
+                          fill_value=field.bmdi)
```
marqh: should this also set the data type

lib/iris/fileformats/rules.py
```diff
@@ -901,15 +902,15 @@ def _make_cube(field, converter):
     metadata = converter(field)
```
marqh: please re-implement this, don't try/accept

lib/iris/cube.py
```diff
@@ -1631,9 +1636,10 @@ def lazy_data(self, array=None):
             A lazy array, representing the Cube data array.
 
         """
+        result = None
```
marqh: don't allow this, if there is not _numpy_array and no _dask_array raise an exception

docs/iris/src/developers_guide/dask_interface.rst
```diff
@@ -0,0 +1,21 @@
...
+    * in which case `cube._my_data` shall be `None`;
+  * this may wrap the numpy array in `cube._my_data`.
+* All dask graphs wrap array-like object where missing data is represented by `nan`.
+  * masked arrays derived from these arrays shall create their mask using the nan location.
+  * where dask wrapped `int` arrays require masks, these will first be cast to `float`
+* In order to support this mask conversion, cube's have a `fill_value` as part of their metadata, which may be None.
+* Array copying is kept to an absolute minimum:
+  * array references should always be passed, not new arrays created, unless an explicit copy operation is requested.
+* To test for the presence of a dask array of any sort, we use:
+  * `hasattr(data, 'compute')`
```
bjlittle: This developer note needs a little bit of TLC to align it with the code changes ...

lib/iris/_merge.py
```diff
@@ -1228,17 +1227,23 @@ def merge(self, unique=True):
...
+                        data = data.data
+                    data = da.from_array(data, chunks=data.shape)
                 stack[nd_index] = data
 
             merged_data = _multidim_daskstack(stack)
             if all_have_data:
                 # All inputs were concrete, so turn the result back into a
                 # normal array.
-                merged_data = as_concrete_data(merged_data)
+                merged_data = merged_data.compute()
```
bjlittle: @marqh `dask` is pretty greedy, and why not. By default `dask.array` uses the `dask.threaded.get` scheduler and will also default to using as many worker as there are cores.

I'd suggest that whenever we use `compute` we factor in a bit of grace and not handover **all** of the cores to `dask`. For me, a sensible *initial* default might be something like `multiprocessing.cpu_count() - 1`. 

We'll need to think (perhaps later once the dust settles) about how we pass options down into dask to control the scheduler of choice, see [here](http://dask.pydata.org/en/stable/scheduler-overview.html#configuring-the-schedulers), which may be relevant for those users targeting clusters and opting to use `distributed`.

For now, I'd suggest:
```python
import multiprocessing
...
    merged_data = merged_data.compute(num_workers=multiprocessing.cpu_count() - 1)
```

lib/iris/cube.py
```diff
@@ -1681,10 +1703,18 @@ def data(self):
             (10, 20)
 
         """
-        data = self._my_data
-        if is_lazy_data(data):
+        if self._numpy_array is None:
             try:
-                data = as_concrete_data(data)
+                data = self._dask_array.compute()
```
bjlittle: See this [comment](https://github.com/SciTools/iris/pull/2392/files#r102430258) regarding use of `num_workers`

lib/iris/util.py
```diff
@@ -1599,3 +1599,10 @@ def demote_dim_coord_to_aux_coord(cube, name_or_coord):
     cube.remove_coord(dim_coord)
 
     cube.add_aux_coord(dim_coord, coord_dim)
+
+
+def is_dask_array(data):
+    result = False
+    if hasattr(data, 'compute'):
+        result = True
+    return result
```
bjlittle: @marqh 
```python 
return hasattr(data, 'compute')
```
works for me (for now)

bjlittle: Hmmm or did we want to be more precise and go with
```python
return isinstance(data, dask.array.core.Array)
```

bjlittle: Actually, to be honest, if this turns into a one-liner, then why do we have this function call at all?

Just inline this directly into the code?

lib/iris/_merge.py
```diff
@@ -1228,17 +1227,23 @@ def merge(self, unique=True):
                 data = self._skeletons[group[offset]].data
                 # Ensure the data is represented as a dask array and
                 # slot that array into the stack.
-                if is_lazy_data(data):
+                if iris.util.is_dask_array(data):
```
bjlittle: @marqh See this [comment](https://github.com/SciTools/iris/pull/2392/files#r102432818) discussion on usage of `iris.util.is_dask_array` ...

lib/iris/_merge.py
```diff
@@ -1228,17 +1227,23 @@ def merge(self, unique=True):
...
                     all_have_data = False
                 else:
-                    data = as_lazy_data(data)
+                    if isinstance(data, np.ma.MaskedArray):
+                        if np.ma.is_masked(data):
+                            if data.dtype.kind == 'i':
+                                data = data.astype('f8')
+                            data[data.mask] = np.nan
+                        data = data.data
```
bjlittle: @marqh Okay, I've been waiting to see this kinda pattern appear in the code base, and here it is ...

On `master`, this code block replaces
```python
data = biggus.NumpyArrayAdapter(data)
```
which takes the concrete data and wraps it up as a biggus thing, so that all data is on an equal footing i.e. a biggus array.

For convenience, we could opt to do something similar, when we want a concrete array to be a lazy thing and ensure that masking is deal with appropriately. This could be achieved by having a new `DataProxy` that, akin to the  `PPDataProxy` and `NetCDFDataProxy` would honour the task of massaging the data into `nan` format if necessary.

So, for above we would have something along the lines of
```python
data = da.from_array(DataProxy(data), chunks=data.shape)
```

... and when (if) dask supports masked arrays, this line would become (something like)
```python
data = da.from_array(data, chunks=data.shape)
```

That's pretty attractive to me. It compartmentalizes the data munging to `nan`, and the `DataProxy` pattern can easily be employed else where to consistently do the job of lazifying masked data.

lib/iris/cube.py
```diff
@@ -64,7 +65,8 @@ class CubeMetadata(collections.namedtuple('CubeMetadata',
                                            'var_name',
                                            'units',
                                            'attributes',
-                                           'cell_methods'])):
+                                           'cell_methods',
+                                           'fill_value'])):
```
bjlittle: @marqh Didn't we want to add `shape` to the `CubeMetadata` ? The `shape` of a cube is immutable, and it takes care of a corner case failure that @djkirkham found the other day, when comparing cubes with the same data values but different broadcastable shapes and no coordinates e.g.
```python
>>> a = np.arange(5).reshape(1, 5)
>>> b = np.array([0, 1, 2, 3, 4, 0, 1, 2, 3, 4]).reshape(2, 5)
>>> ca = iris.cube.Cube(a)
>>> cb = iris.cube.Cube(b)
>>> ca == cb
True
```

lib/iris/cube.py
```diff
@@ -1590,15 +1602,34 @@ def cell_methods(self, cell_methods):
...
+        """
+        The data at the core of this cube.
+        May be a numpy array or a dask array.
+
+        """
+        if self._numpy_array is not None:
+            result = self._numpy_array
+        else:
+            result = self._dask_array
+        return result
```
bjlittle: @marqh What's the use case for this method?

lib/iris/cube.py
```diff
@@ -1590,15 +1602,34 @@ def cell_methods(self, cell_methods):
...
-        return self._my_data.dtype
+        if self._dtype is None:
+            result = self.core_data.dtype
+        else:
+            result = self._dtype
+        return result
+
+    @dtype.setter
+    def dtype(self, dtype):
+        self._dtype = dtype
```
bjlittle: @marqh Are we going to ensure that the `dtype` is used to automatically cast the concrete data appropriately?

Can I use this to force the cube to cast my data from `float32` to `float64` ?

lib/iris/cube.py
bjlittle: Should `dtype` be part of the `CubeMetadata` ?

marqh: I have explicitly commented on this
https://github.com/SciTools/iris/issues/2394#issuecomment-281644724

but without details

there are test problems if we try to promote shape to be cube metadata.  I'd like to hold off on doing this for now and stick with shape being inferred

i'll need to dig out the tests which caused problems, but it unravelled on me, so i left it out for now

to be continued

lib/iris/_merge.py
marqh: @pp-mo and i had a similar conversation about this pattern.  I think there is a neater abstraction to be had.

please may we continue with this slightly awkward implementation whilst this PR gets test hardy, then implement an improved pattern as a follow up?
perhaps you could raise a ticket for this for us?

lib/iris/util.py
marqh: we've been round this loop a few times

at present there is concern that we may want to change the 'how do we check for this' pattern.  Hence the suggestion of a util function.
however, this is in the public API and expected to continue, which we may not want
to be continued, i think
perhaps the death of the private `iris._lazy_data` or it's offspring was declared early?

lib/iris/cube.py
```diff
@@ -1642,11 +1671,23 @@ def lazy_data(self, array=None):
...
+        elif self._numpy_array is not None:
+            data = self._numpy_array
+            if isinstance(data, np.ma.masked_array):
+                if np.ma.is_masked(data):
+                    if data.dtype.kind == 'i':
+                        data = data.astype(np.dtype('f8'))
+                    # Where possible, write these NANs into the cube's
+                    # _numpy_array.
+                    data.data[data.mask] = np.nan
+                data = data.data
```
bjlittle: @marqh This block of code is identical to [this](https://github.com/SciTools/iris/pull/2392/files#diff-4663903670147de0963e3cd8b176d753R1233) ... we need to refactor this out.

lib/iris/cube.py
bjlittle: @AlexHilson Is floating the [proposal](https://github.com/SciTools/iris/issues/2394#issuecomment-281656918) that there is benefit to defaulting to `self._dask_array` over `self._numpy_array`.

lib/iris/cube.py
marqh: the casting only occurs at the point where dask data is realised to numpy data.
at this point the casting is implemented and teh cube's _dtype is set to None
all dtype calls then fall through to the numpy array

marqh:
okay, so i've had one green tick

i have now returned the _lazy_data module in a minimal form (@pp-mo for you :) and pushed all data setting behaviour onto the `cube.data` call

`cube.lazy_data()` is a function which cannot set lazy data

i think this represents a good place to be

@pp-mo @bjlittle please review and heck the tests; consider whether this is ready to 'squash n merge'

further comments and request are welcome (of course) 

I am keen to get this as a baseline and look to branch from here, so follow on tickets are also fine

marqh:
...fail
`fill_value=-1.0000000150474662e+30 != fill_value=-1e+30`

:(

marqh:
@bjlittle @pp-mo 
there is a green tick \o/

lib/iris/_lazy_data.py
```diff
@@ -45,102 +37,18 @@ def is_lazy_data(data):
...
-
     """
-    if (not isinstance(array, np.ma.masked_array) and
-            array.dtype.kind == 'f'):
-        mask = np.isnan(array)
-        if np.any(mask):
-            # Turn any unmasked array with NaNs into a masked array.
-            array = np.ma.masked_array(array, mask=mask)
+    if mask is None:
+        mask = array.mask
```
bjlittle: @marqh Of course if you don't pass in a masked array, then this will raise an exception ...

... every time this function is called (pretty) much, it's always within an `if` statement to check to see if the `array` is a masked array.

Why don't we refactor this, so that any array can be passed in. If the `array` is not a `MaskedArray`  then just pass it through. If it's a `MaskedArray` then do the NaN-ifycation, as is.

That would save on the use of `if isinstance(array, np.ma.MaskedArray) .... array_masked_to_nans(...)` pattern.

If you buy into that, then I'd simply suggest calling this function `array_to_nans` ... but that's minor to my previous points.

lib/iris/_merge.py
```diff
@@ -1231,14 +1231,18 @@ def merge(self, unique=True):
                 if is_lazy_data(data):
                     all_have_data = False
                 else:
-                    data = as_lazy_data(data)
+                    if isinstance(data, np.ma.MaskedArray):
```
bjlittle: @marqh This should be `ma.MaskedArray` not `np.ma.MaskedArray` to adhere to the previous namespace usgage

lib/iris/_merge.py
```diff
@@ -1231,14 +1231,18 @@ def merge(self, unique=True):
...
+                        if np.ma.is_masked(data):
```
bjlittle: `np.ma` -> `ma`

lib/iris/cube.py
```diff
@@ -714,9 +718,18 @@ def __init__(self, data, standard_name=None, long_name=None,
...
-            data = np.asarray(data)
-        self._my_data = data
+        self.fill_value = fill_value
+
+        if is_lazy_data(data):
+            self._dask_array = data
+            self._numpy_array = None
+        else:
+            self._dask_array = None
+            if not isinstance(data, np.ma.MaskedArray):
```
bjlittle: `np.ma` -> `ma`

lib/iris/cube.py
```diff
@@ -1590,63 +1604,67 @@ def cell_methods(self, cell_methods):
...
-                # 0-dimensional array into a 1-dimensional array of
-                # length one.
-                # i.e. self.shape = () and array.shape == (1,)
-                if self.shape or array.shape != (1,):
-                    raise ValueError('Require cube data with shape %r, got '
-                                     '%r.' % (self.shape, array.shape))
-            self._my_data = array
+        if self._numpy_array is not None:
+            data = self._numpy_array
+            if isinstance(data, np.ma.masked_array):
```
bjlittle: `np.ma` -> `ma`

lib/iris/fileformats/netcdf.py
```diff
@@ -392,10 +392,13 @@ def __getitem__(self, keys):
         try:
             variable = dataset.variables[self.variable_name]
             # Get the NetCDF variable data and slice.
-            data = variable[keys]
+            var = variable[keys]
         finally:
             dataset.close()
-        return data
+        if isinstance(var, np.ma.MaskedArray):
```
bjlittle: `np.ma` -> `ma`

lib/iris/tests/test_concatenate.py
```diff
@@ -383,6 +384,27 @@ def test_concat_masked_2y2d(self):
                          [True, False]], dtype=np.bool)
         self.assertArrayEqual(result[0].data.mask, mask)
 
+    def test_concat_masked_2y2d_int16(self):
+        cubes = []
+        x = (0, 2)
+        cube = _make_cube(x, (0, 2), 1, dtype=np.int16, fill_value=-37)
+        cube.data = np.ma.asarray(cube.data)
```
bjlittle: `np.ma` -> `ma`

lib/iris/tests/unit/cube/test_Cube.py
```diff
@@ -49,7 +49,6 @@ def test_ndarray(self):
         self.assertEqual(type(cube.data), np.ndarray)
         self.assertArrayEqual(cube.data, data)
 
-    @tests.skip_biggus
     def test_masked(self):
         # np.ma.MaskedArray should be allowed through
         data = np.ma.masked_greater(np.arange(12).reshape(3, 4), 1)
```
bjlittle: `np.ma` -> `ma`

lib/iris/tests/unit/cube/test_Cube.py
```diff
@@ -117,7 +116,6 @@ def test_1d_cube_noexists(self):
 
 
 class Test_xml(tests.IrisTest):
-    @tests.skip_biggus
     def test_checksum_ignores_masked_values(self):
         # Mask out an single element.
         data = np.ma.arange(12).reshape(3, 4)
```
bjlittle: `np.ma` -> `ma`

lib/iris/tests/unit/fileformats/pp/test__data_bytes_to_shaped_array.py
```diff
@@ -71,7 +71,9 @@ def test_boundary_decompression(self):
         r = pp._data_bytes_to_shaped_array(self.data_payload_bytes,
                                            lbpack, boundary_packing,
                                            self.data_shape,
-                                           self.decompressed.dtype, -99)
+                                           self.decompressed.dtype,
+                                           -9223372036854775808)
+        r = np.ma.masked_array(r, np.isnan(r), fill_value=-9223372036854775808)
```
bjlittle: `np.ma` -> `ma`

lib/iris/tests/unit/fileformats/pp/test__data_bytes_to_shaped_array.py
```diff
@@ -153,11 +155,12 @@ def check_read_data(self, field_data, lbpack, mask):
         # Calls pp._data_bytes_to_shaped_array with the necessary mocked
         # items, an lbpack instance, the correct data shape and mask instance.
         with mock.patch('numpy.frombuffer', return_value=field_data):
-            return pp._data_bytes_to_shaped_array(mock.Mock(),
+            data = pp._data_bytes_to_shaped_array(mock.Mock(),
                                                   self.create_lbpack(lbpack),
                                                   None,
                                                   mask.shape, np.dtype('>f4'),
                                                   -999, mask=mask)
+        return np.ma.masked_array(data, np.isnan(data), fill_value=-999)
```
bjlittle: `np.ma` -> `ma`

lib/iris/tests/unit/lazy_data/test_array_masked_to_nans.py
```diff
@@ -34,7 +34,7 @@ def test_masked(self):
         masked_array = np.ma.masked_array([[1.0, 2.0], [3.0, 4.0]],
```
bjlittle: `np.ma` -> `ma`

lib/iris/tests/unit/lazy_data/test_array_masked_to_nans.py
```diff
@@ -34,7 +34,7 @@ def test_masked(self):
...
                                           mask=[[0, 1], [0, 0]])
 
-        result = array_masked_to_nans(masked_array)
+        result = array_masked_to_nans(masked_array).data
 
         self.assertIsInstance(result, np.ndarray)
         self.assertFalse(isinstance(result, np.ma.MaskedArray))
```
bjlittle: `np.ma` -> `ma`

lib/iris/tests/unit/lazy_data/test_array_masked_to_nans.py
```diff
@@ -48,7 +48,7 @@ def test_masked(self):
     def test_empty_mask(self):
         masked_array = np.ma.masked_array([1.0, 2.0], mask=[0, 0])
```
bjlittle: `np.ma` -> `ma`

lib/iris/tests/unit/lazy_data/test_array_masked_to_nans.py
```diff
@@ -48,7 +48,7 @@ def test_masked(self):
...
 
-        result = array_masked_to_nans(masked_array)
+        result = array_masked_to_nans(masked_array).data
 
         self.assertIsInstance(result, np.ndarray)
         self.assertFalse(isinstance(result, np.ma.MaskedArray))
```
bjlittle: `np.ma` -> `ma`

lib/iris/_merge.py
```diff
@@ -36,8 +36,8 @@
 import iris.cube
 import iris.coords
 import iris.exceptions
-from iris._lazy_data import is_lazy_data, as_concrete_data, as_lazy_data
 import iris.util
+from iris._lazy_data import is_lazy_data, array_masked_to_nans
```
bjlittle: @marqh Minor point but `from iris._lazy ...` sorts before alphabetical, and should be the first `iris` import.

lib/iris/cube.py
```diff
@@ -46,10 +47,11 @@
 import iris.coords
 import iris._concatenate
 import iris._constraints
-from iris._lazy_data import is_lazy_data, as_lazy_data, as_concrete_data
+
 import iris._merge
 import iris.exceptions
 import iris.util
+from iris._lazy_data import is_lazy_data, array_masked_to_nans
```
bjlittle: @marqh the import order is wrong

lib/iris/cube.py
bjlittle: @marqh :bomb: If the `dtype` changes, then the cubes `metadata` surely also needs to change!

lib/iris/cube.py
```diff
@@ -1681,43 +1699,52 @@ def data(self):
...
+        if is_lazy_data(value):
+            self._dask_array = value
+            self._numpy_array = None
+
+        else:
+            self._numpy_array = value
 
     def has_lazy_data(self):
-        return is_lazy_data(self._my_data)
+        return True if self._numpy_array is None else False
```
bjlittle: @marqh I'd prefer
```python
return self._numpy_array is not None
```

Also, isn't it the case the both `cube._numpy_array` and `cube._dask_array` could be populated? So, are we saying that if that's the case, the cube isn't lazy?

marqh: > Also, isn't it the case the both cube._numpy_array and cube._dask_array could be populated? 
yes, at present this is the case
> If so, are we saying that if that's the case, the cube isn't lazy?
yes, it is not lazy
if `cube._numpy_array` exists it is always used in preference

> Is there an explicit statement to define whether it's valid for both of these attributes to be populated? 
No, there is no explicit statement.  It is an open question.  
We should answer it.  The question here is should we answer it on this PR or raise an issue and answer it?
The utility of having both is the `cube._numpy_array` may be set to None, reverting the cube back to it's initally loaded state, with a file proxy.  I don't know how valuable this may be.

> Sutely, it's totally invalid for both not to be populated ... 

Yes, that is invalid and exceptions are thrown

> do we have checks and balances for that to maintain that integrity within the cube?

there are some, are there enough?

lib/iris/_lazy_data.py
marqh: There have been multiple requests to factor out this logic, but care is required as it is used in different ways in different contexts

> ... every time this function is called (pretty) much, it's always within an if statement to check to see if the array is a masked array.

that is not the case

this function is called 3 times so far, one of which is with a numpy array, 2 with masks.  
We will want more numpy only uses, I am confident.  In these cases the mask test has to be passed in explicitly.

plausible approaches:
 * [ ] as is, if you use this wrong, it fails; it is in a private module so should only be used by Iris code
 * [ ]  test for type within the function, such that if it is called without `mask` the type must be a masked array
 * [ ] demand that a mask is always provided, such that the call becomes `array_masked_to_nans(mymaskedarray, mymaskedarray.mask)`
 * [ ] ...??

marqh:
@bjlittle 

ok, further review actions pushed.  hopefully i've maintained test success; we shall see

please squash and merge at will once travis is content

bjlittle:
@marqh Agreed. 

If the tests pass, then let's bank this PR. It's a feature branch and we need to move forwards :+1:

bjlittle:
Ping @marqh you still have a couple of tests referencing the now nuked `_data_cache` ...

marqh:
> Ping @marqh you still have a couple of tests referencing the now nuked _data_cache ...

doh
trying again

marqh:
@bjlittle 
it has gone green
...

dkillick:
Woop! :tada: :cake: :wine_glass:

lib/iris/tests/integration/test_pp.py
```diff
@@ -47,6 +47,8 @@ def _test_coord(self, cube, point, bounds=None, **kwargs):
         if bounds is not None:
             self.assertArrayEqual(coords[0].bounds, [bounds])
 
+    # hits a segfault, very odd
```
pp-mo: These should be fixed by : #2395
Short version: since we changed our test for lazy data, a simple Mock tests as lazy instead of real.
Then ```np.ma.asarray()``` can hit an endless loop (whereas ```np.asarray``` was ok).

PR #2395, opened 2017-02-20
Mocked PPFields need a useable _data property.

pp-mo:
A mock replacing a PPField now behaves differently in the PPField.data call (in pp.py).
It _used_ to have ```isinstance(data, biggus.Array)```, to which Mock replies 'False'.
Now has ```is_lazy_data(data)```, which translates to ```hasattr(data, 'compute')```, to which Mock replies True.

It seems that np.array(mock) is ok, but ```np.ma.masked_array(mock)``` does something nasty.


marqh:
@pp-mo please may you rebase/retarget this to `dask_timed`?

it's good to go apart from that, in my view

pp-mo:
Done. passes :grin: !

PR #2402, opened 2017-02-21
first run at daskifying stats and maths

corinnebosley:


lib/iris/analysis/maths.py
```diff
@@ -121,34 +121,24 @@ def _assert_is_cube(cube):
 def _assert_compatible(cube, other):
     """
     Checks to see if cube.data and another array can be broadcast to
-    the same shape using ``numpy.broadcast_arrays``.
+    the same shape using ``numpy.broadcast_arrays`
```
corinnebosley: This no longer uses numpy.broadcast_shapes.  It uses dask.array.core.broadcast_shapes.  Make this docstring correct.

lib/iris/analysis/maths.py
```diff
@@ -585,16 +575,17 @@ def _binary_op_common(operation_function, operation_name, cube, other,
                            `cube` and `cube.data`
     """
     _assert_is_cube(cube)
-
     if isinstance(other, iris.coords.Coord):
         other = _broadcast_cube_coord_data(cube, other, operation_name, dim)
     elif isinstance(other, iris.cube.Cube):
         try:
-            BA.broadcast_arrays(cube._my_data, other._my_data)
+            da.core.broadcast_shapes(cube.shape, other.shape)
```
corinnebosley: This can just be 'broadcast_shapes' rather than 'da.core.broadcast_shapes', right?

corinnebosley:
@marqh @bjlittle @pp-mo @AlexHilson
I'm confident and excited about this.  Please take a look and let us know what you think.  Thanks

lib/iris/analysis/__init__.py
```diff
@@ -561,7 +562,7 @@ def post_process(self, collapsed_cube, data_result, coords, **kwargs):
             The collapsed cube with its aggregated data payload.
 
         """
-        if isinstance(data_result, biggus.Array):
+        if isinstance(data_result, da.Array):
```
marqh: is this equivalent to 
`iris._lazy_data.is_lazy_data(data_result)`

lib/iris/cube.py
```diff
@@ -3052,7 +3052,7 @@ def _deepcopy(self, memo, data=None):
                 new_cube_data = self.data.copy()
         else:
             # Use the provided data (without copying it).
-            if not isinstance(data, biggus.Array):
+            if not isinstance(data, da.Array):
```
marqh: is this equivalent to 
`iris._lazy_data.is_lazy_data(data_result)`

marqh:
> I'm confident and excited about this. Please take a look and let us know what you think. Thanks

first look: nice job

i'll get a more detailed analysis up soon.

I am impressed by how little change there is here

AlexHilson:
Looks great.

This is an area we we _could_ use lazy (dask) only code as raised [here](https://github.com/SciTools/iris/issues/2394#issuecomment-281656918). It looks like all of the maths operations have lazy equivalents, and we can always get a dask array out of the cube (it may just be a wrapped numpy array but that's fine).

lib/iris/cube.py
corinnebosley: Yes I believe the two are equivalent.  I have not tested the is_lazy_data approach, but I can't see a good reason not to use it.

corinnebosley:
I would be interested in a discussion about @AlexHilson's suggestion of using dask-only code in this area.  In previous iterations of these changes, I discovered that some of the tests were failing (and by 'I' here, I mean 'Dan') because the result was expected to be loaded and was actually deferred, but I don't see that as a genuine problem.  I mean, either the tests are silly or I think that the problem can be easily resolved.

marqh:
> I mean, either the tests are silly or I think that the problem can be easily resolved.

agreed



> I would be interested in a discussion about @AlexHilson's suggestion of using dask-only code in this area.

given the simplicity of this PR, I am tempted to review (and merge) this change, then investigate that option straight after

AlexHilson:
With the current implementation we would expect adding two deferred cubes to result in a deferred cubes (new behaviour) and two realized cubes to result in a realized cube (old behavior). So I don't think those tests are useless.

+1 to merge now discuss later

djkirkham:
@AlexHilson @corinnebosley 

> This is an area we we could use lazy (dask) only code

I like the idea, as long as we prevent a cube seeming to be lazy when it isn't, as mentioned in your comment on the other ticket. I think we'd need to do something like `cube.data = lazy_data.compute()` if the cube were initially non-lazy.

marqh:
we could allow the cube to remain in its current state and use whatever we are provided with, via `cube.core_data`

what I don't understand is what would happen to an arithmetic operation if one cube were lazy and one not

can i add a numpy array to a dask array?

AlexHilson:
That's a valid approach, it means that the result of calling 'has_lazy_data()' doesn't change which may reduce confusion for end users.

However, if we don't call .compute then the cube _is_ lazy - the 'real' data doesn't exist yet. We're just holding onto numpy arrays rather than file handlers (possibly meaning the cube has a bigger memory footprint?). So if a user chains multiple operations we won't calculate the results until they actually ask for the data.

From a user perspective .data behaves in the same exact same way, it's just the result of 'has_lazy_data' that would change. I think with either approach we would get multi core maths with both lazy / non lazy data.

AlexHilson:
@marqh you can add a numpy array to a dask array. But I don't know if that's true for each possible operation. 

But my suggestion is that a cube _always_ has lazy data available, so this situation never comes up - we would always be adding a dask array to a dask array. In this example it would be like (pseudo code): `dask.array_from(<NetCDFArrayPRoxy>) + dask.array_from(numpy.array())`

AlexHilson:
@marqh this funky graph is what you get from
` dask.array(...) + numpy.array(...)`

dask_plus_numpy

djkirkham:
@AlexHilson I think an important use for `has_lazy_data()` is to ask the question "Does this operation cause data to be loaded from the disk?" For the arithmetic operators the answer is "No", and I think `has_lazy_data()` should reflect that, though others may disagree.

AlexHilson:
@djkirkham +1, I use has_lazy_data() exactly for that.

But that behavior is unchanged by this proposal. If cube.has_lazy_data() == True before the operation, cube.has_lazy_data() == True after the operation. The potentially confusing bit is that has_lazy_data() == False flips to True. 

So if you care about your data staying lazy this doesn't affect you. It's only if you care about your data staying realized.

marqh:
I'm not seeing any dissenting voices

I'm muchly minded to merge this

going once, twice.....

djkirkham:
@AlexHilson Good point, but there's still a problem when two operations are chained together. If the first loads the data, the second would then hide that fact. If this happened within a complex Iris function it would be a problem.

PR #2405, opened 2017-02-23
Add test timings

marqh:
this is a replacement for the dask branch

it has all the code from the branch, but is rebased to bring in test timings.

This should now be able to be merged back onto master (one day, maybe)


marqh:
@bjlittle tests are passing, if we could get this in so we can all rebase our open work onto it, that'd be super

bjlittle:
@marqh Just confirmed that the `dask_timed` branch can now merge with `master` :+1: 

Perhaps tomorrow we can arrange to delete the `dask` feature branch when it's safe to do so ...

PR #2406, opened 2017-02-23
Dask aggs

pp-mo:
This replaces biggus with dask, getting the three existing lazy statistics working again.
Not as hard as it sounded, but I'm not convinced that there is enough testing.
We might even want in-depth testing of each function to ensure the dask versions do everything that is expected ?

### DIFF
--- a/lib/iris/analysis/__init__.py
+++ b/lib/iris/analysis/__init__.py
@@ -51,8 +51,9 @@ from six.moves import (filter, input, map, range, zip)  # noqa
 import six
 
 import collections
+from functools import wraps
 
-import biggus
+import dask.array as da
 import numpy as np
 import numpy.ma as ma
 import scipy.interpolate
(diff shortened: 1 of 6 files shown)


lib/iris/analysis/__init__.py
```diff
@@ -1412,7 +1410,35 @@ def interp_order(length):
...
+    """
+    @wraps(dask_nanfunction)
+    def inner_stat(array, axis=-1, mdtol=None, **kwargs):
+        dask_result = dask_nanfunction(array, axis=axis, **kwargs)
+        if mdtol is None:
+            result = dask_result
+        else:
+            point_counts = np.sum(np.ones(array.shape), axis=axis)
+            point_mask_counts = da.sum(da.isnan(array), axis=axis)
+            masked_point_fractions = (point_mask_counts + 0.0) / point_counts
```
pp-mo: Note: the "+ 0.0" ensures a floating divison, as there is no "dask.float".
I could probably add a comment to that effect....

pp-mo:
Note that we do get an immediate benefit as I've been able to remove the limitation to a single axis for all these statistics.
Implementation is pretty easy, but I'm not sure if the testing is adequate.
Do please take a look @marqh @djkirkham @bjlittle @corinnebosley @AlexHilson.

marqh:
target dask_timed please

pp-mo:
> target dask_timed please

Now done :grin:

lib/iris/analysis/__init__.py
marqh: i would appreciate that, yes please

PR #2407, opened 2017-02-24
Fix non-lazy pp loading

djkirkham:
The code on the current dask_timed branch doesn't function correctly in the case of loading a PP file in a non-lazy way; the `_data` attribute is always wrapped up as a dask array before being passed into the `Cube` constructor.

Actually I think we need to make bigger changes to the ones I've suggested in this PR. The problem has arisen because `PPField._data` is, in the lazy case, a `PPFieldProxy` object, and so can't be used directly in  `Cube.__init__`. Using `_data` outside of `PPField` class methods is probably a bad idea anyway.

### DIFF
--- a/lib/iris/fileformats/pp.py
+++ b/lib/iris/fileformats/pp.py
@@ -1288,6 +1288,9 @@ class PPField(six.with_metaclass(abc.ABCMeta, object)):
     def data(self, value):
         self._data = value
 
+    def has_lazy_data(self):
+        return not isinstance(self._data, LoadedArrayBytes)
+
     @property
     def calendar(self):
         """Return the calendar of the field."""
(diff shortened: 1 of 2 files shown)


marqh:
@djkirkham 
i agree that this needs to be fixed, it is a concerning instability in the code at present
i am not immediately confident this is the fix, it may be, but i'm not sure

there are also fragile tests in this space, using mock PPField things, which add to the concerns.

there is a syntax error in here, so the code isn't testing

i think i'd like this written up as an issue, on the project board, referencing this ticket.

there's some careful implementation decisions needed here

djkirkham:
See https://github.com/SciTools/iris/pull/2410 for an alternative implementation

marqh:
replaced

PR #2409, opened 2017-02-24
Replace intersection

corinnebosley:
This is not a complete fix for cube.intersection, this is just a first little attempt I have made at swapping out Biggus.  Not convinced that all the tests will pass, but it's a good start at the very least.


corinnebosley:
@marqh I have removed some intersection test skippers, are there any others that you know about?

corinnebosley:
@marqh @pp-mo

marqh:
neat, many thanks @corinnebosley

PR #2410, opened 2017-02-24
PPfield core_data

djkirkham:
In at least one place (`iris.fileformats.pp._make_cube()`) there is code to handle an object which may be either a `PPField` or a `FieldCollation`. The code in `_make_cube()` does this with try/except on the existence of a `_data` attribute, which is only present on PPField and is intended to be either lazy or not, akin to `Cube._my_data`. As mentioned in a code comment, this is causing tests using `MagicMock`s to magically pass.

The code on the current dask_timed branch doesn't function correctly in the non-lazy case; the `_data` attribute is always wrapped up as a dask array before being passed into the `Cube` constructor.

This PR adds a `core_data` property to both the `PPField` and the `FieldCollation` classes, which behaves similarly to `Cube.core_data` in the `PPField` case, and simply calls `data` in the `FieldCollation` case. I've also made the `_data` attribute of `PPField` a dask array again.

### DIFF
--- a/lib/iris/fileformats/_ff.py
+++ b/lib/iris/fileformats/_ff.py
@@ -782,13 +782,13 @@ class FF2PP(object):
                             # Read the actual bytes. This can then be converted
                             # to a numpy array at a higher level.
                             ff_file_seek(data_offset, os.SEEK_SET)
-                            result_field._data = pp.LoadedArrayBytes(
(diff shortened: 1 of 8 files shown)


marqh:
excellent, this looks like a much neater implementation.  I am very pleased to see the back of the `try:except` in the rules module

thank you

PR #2415, opened 2017-03-06
Coding standards tweaks to cube code

dkillick:
There were a few inconsistencies in the `cube` + `dask` code, and instead of noting them down for further discussion, I thought I'd make a vessel for discussion (that is, this PR).

### DIFF
--- a/lib/iris/cube.py
+++ b/lib/iris/cube.py
@@ -716,8 +716,6 @@ bound=(1994-12-01 00:00:00, 1998-12-01 00:00:00)
         if isinstance(data, six.string_types):
             raise TypeError('Invalid data type: {!r}.'.format(data))
 
-        self.fill_value = fill_value
-
         if is_lazy_data(data):
             self._dask_array = data
             self._numpy_array = None
@@ -726,8 +724,6 @@ bound=(1994-12-01 00:00:00, 1998-12-01 00:00:00)
             if not isinstance(data, ma.MaskedArray):
                 data = np.asarray(data)
             self._numpy_array = data
-
-        self._dtype = dtype
 
         #: The "standard name" for the Cube's phenomenon.
         self.standard_name = standard_name
@@ -754,6 +750,10 @@ bound=(1994-12-01 00:00:00, 1998-12-01 00:00:00)
 
         # Cell Measures
         self._cell_measures_and_dims = []
+
+        self.fill_value = fill_value
+
+        self._dtype = dtype
 
         identities = set()
         if dim_coords_and_dims:
@@ -1611,10 +1611,10 @@ bound=(1994-12-01 00:00:00, 1998-12-01 00:00:00)
         to be decided: should this be public??
 
         """
-        if self._numpy_array is not None:
-            result = self._numpy_array
-        else:
+        if self.has_lazy_data():
             result = self._dask_array
+        else:
+            result = self._numpy_array
         return result
 
     @property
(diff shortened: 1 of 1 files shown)


lib/iris/cube.py
```diff
@@ -755,6 +751,10 @@ def __init__(self, data, standard_name=None, long_name=None,
         # Cell Measures
         self._cell_measures_and_dims = []
 
+        self.fill_value = fill_value
+
+        self._dtype = dtype
```
dkillick: Assigning args and kwargs into the instance in the constructor should be done in the order the args and kwargs are passed to the constructor, as is true for all the other args.

lib/iris/cube.py
```diff
@@ -1611,10 +1611,10 @@ def core_data(self):
         to be decided: should this be public??
 
         """
-        if self._numpy_array is not None:
```
dkillick: We've defined a method to make this check, so let's use it rather than repeating the same code.

lib/iris/cube.py
```diff
@@ -1654,14 +1654,14 @@ def lazy_data(self):
             A lazy array, representing the Cube data array.
 
         """
-        if self._numpy_array is not None:
```
dkillick: We've defined a method to make this check, so let's use it rather than repeating the same code.

lib/iris/cube.py
```diff
@@ -1697,7 +1697,7 @@ def data(self):
             (10, 20)
 
         """
-        if self._numpy_array is None:
```
dkillick: We've defined a method to make this check, so let's use it rather than repeating the same code.

lib/iris/cube.py
```diff
@@ -1742,7 +1742,7 @@ def data(self, value):
             self._numpy_array = value
 
     def has_lazy_data(self):
-        return True if self._numpy_array is None else False
```
dkillick: What? `self._numpy_array is None` is already a logical test, so return the result of the logical test.

lib/iris/cube.py
```diff
@@ -2205,12 +2205,11 @@ def new_cell_measure_dims(cm_):
             first_slice = next(slice_gen)
         except StopIteration:
             first_slice = None
-        if self._numpy_array is not None:
```
dkillick: We've defined a method to make this check, so let's use it rather than repeating the same code.

lib/iris/cube.py
```diff
@@ -2205,12 +2205,11 @@ def new_cell_measure_dims(cm_):
...
-            cube_data = self._numpy_array
-        elif self._dask_array is not None:
+
+        if self.has_lazy_data():
             cube_data = self._dask_array
         else:
-            raise ValueError('This cube has no data, slicing is not supported')
```
dkillick: This can never be hit. You have to supply a `data` variable to a cube constructor, and even if the data variable is not an array and not a dask object, whatever it _is_ still gets passed through `np.asarray()` and assigned to `cube._numpy_array`.

dkillick:
Replaced by #2416.

lib/iris/cube.py
pp-mo: Do you have a reference for this ?
I had not heard of this as a "rule" --  I can see why it might be important in other languages, but why in Python ?

As Python is so dynamic, I can also see there ceratainly _could_ be good reasons for breaking this in particular cases, where one property may need to exist so that we can leverage generic code to set another state aspect.
For instance, here in cube init, we might require dtype and/or shape to be set up so that we could use ```cube.data = data```.
(Which actually would make good sense now, since data assignment has become generic to real or lazy data).

dkillick:
Closing this again - looks like it will be easier to merge #2416 after all.

PR #2416, opened 2017-03-06
Coding standards tweaks to daskified cube code

dkillick:
There were a few inconsistencies in the `cube` + `dask` code, and instead of noting them down for further discussion, I thought I'd make a vessel for discussion (that is, this PR).

A retargetting of #2415 to the base `dask_timed`.

### DIFF
--- a/lib/iris/cube.py
+++ b/lib/iris/cube.py
@@ -716,8 +716,6 @@ bound=(1994-12-01 00:00:00, 1998-12-01 00:00:00)
         if isinstance(data, six.string_types):
             raise TypeError('Invalid data type: {!r}.'.format(data))
 
-        self.fill_value = fill_value
-
         if is_lazy_data(data):
             self._dask_array = data
             self._numpy_array = None
@@ -726,8 +724,6 @@ bound=(1994-12-01 00:00:00, 1998-12-01 00:00:00)
             if not isinstance(data, ma.MaskedArray):
                 data = np.asarray(data)
             self._numpy_array = data
-
-        self._dtype = dtype
 
         #: The "standard name" for the Cube's phenomenon.
         self.standard_name = standard_name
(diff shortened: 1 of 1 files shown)


dkillick:
See the comments on #2415 for a discussion of why I've made these changes.

dkillick:
Ping @bjlittle @pp-mo @lbdreyer

pp-mo:
> See the comments on #2415 for a discussion of why I've made these changes.

See related query, here : https://github.com/SciTools/iris/pull/2415/files#r104770825

lib/iris/cube.py
```diff
@@ -2205,12 +2205,11 @@ def new_cell_measure_dims(cm_):
...
             first_slice = None
-        if self._numpy_array is not None:
-            cube_data = self._numpy_array
-        elif self._dask_array is not None:
+
+        if self.has_lazy_data():
             cube_data = self._dask_array
         else:
-            raise ValueError('This cube has no data, slicing is not supported')
+            cube_data = self._numpy_array
```
bjlittle: @dkillick At this point, we're confident that either `self._dask_array` or `self._numpy_array` is populated.

Prior to this change, there seems to be a hint of that might not be the case, hence the `ValueError` exception being raised. I'd certainly hope that there is a data payload of some kind to process - we need to make sure that internally that is always the case.

bjlittle:
@dkillick @pp-mo @lbdreyer I'm :+1: on this PR ... are there any reasons why we can't bank this now?

lib/iris/cube.py
dkillick: I think if we end up at this point having a cube with no data we're deeper in the sticky than a `ValueError` will help with! The rest of the code in the class makes it clear that we can never _construct_ a cube with no data, so if an end user has done something that means their cube no longer has any data then IMO they can dig their own way out of the sticky stuff...

dkillick:
@bjlittle let's just make sure @pp-mo is happy with my comment at his link above :smile:

dkillick:
@bjlittle I think @pp-mo and I might be happy with this now - we invite you to hit the merge button once Travis is happy!

lib/iris/cube.py
bjlittle: I'm really hoping that there never, ever is any `sticky stuff` :wink:

It's just a note to ourselves really, that we need to ensure the integrity of the cube as best we can

PR #2417, opened 2017-03-07
Trajectories by time

SusanJL:
Added code to allow the reading of NAME trajectories stored by time instead of by particle number. This results in a cube with a dimension coordinate of ppindex (c.f. trajectories stored by particle number will produce a cube with a dimension coordinate of time). Also added some of the file header information into the cube attributes.


marqh:
many thanks for the update @SusanJL 

this looks grand

PR #2421, opened 2017-03-08
Reinstate `as_lazy_data`

dkillick:
The return of the `as_lazy_data` function, which acts as Iris' interface to `da.from_array`. I've added tests for the function and changed the Iris code and tests so that it always uses the new function and does not call `da.from_array` directly.

### DIFF
--- a/lib/iris/_lazy_data.py
+++ b/lib/iris/_lazy_data.py
@@ -66,7 +66,8 @@ def as_lazy_data(data, chunks=_MAX_CHUNK_SIZE):
 
     """
     if not is_lazy_data(data):
-        data = array_masked_to_nans(data)
+        if isinstance(data, np.ma.MaskedArray):
+            data = array_masked_to_nans(data)
         data = da.from_array(data, chunks=chunks)
     return data
 
--- a/lib/iris/_lazy_data.py
+++ b/lib/iris/_lazy_data.py
@@ -25,6 +25,7 @@ from six.moves import (filter, input, map, range, zip)  # noqa
 
 import dask.array as da
 import numpy as np
+import numpy.ma as ma
 
 
 def is_lazy_data(data):
@@ -66,7 +67,7 @@ def as_lazy_data(data, chunks=_MAX_CHUNK_SIZE):
 
     """
     if not is_lazy_data(data):
-        if isinstance(data, np.ma.MaskedArray):
+        if isinstance(data, ma.MaskedArray):
             data = array_masked_to_nans(data)
             data = data.data
         data = da.from_array(data, chunks=chunks)
--- a/lib/iris/_lazy_data.py
+++ b/lib/iris/_lazy_data.py
@@ -40,11 +40,43 @@ def is_lazy_data(data):
     return result
 
 
+# A magic value, borrowed from biggus
+_MAX_CHUNK_SIZE = 8 * 1024 * 1024 * 2
+
+
+def as_lazy_data(data, chunks=_MAX_CHUNK_SIZE):
+    """
+    Convert the input array `data` to a lazy dask array.
+
+    Args:
+
+    * data:
+        An array. This will be converted to a lazy dask array.
+
+    Kwargs:
+
+    * chunks:
+        Describes how the created dask array should be split up. Defaults to a
+        value first defined in biggus (being `8 * 1024 * 1024 * 2`).
+        For more information see
+        http://dask.pydata.org/en/latest/array-creation.html#chunks.
+
+    Returns:
+        The input array converted to a lazy dask array.
+
+    """
+    if not is_lazy_data(data):
+        data = array_masked_to_nans(data)
+        data = da.from_array(data, chunks=chunks)
+    return data
+
+
 def array_masked_to_nans(array, mask=None):
     """
-    Convert a masked array to a normal array with NaNs at masked points.
+    Convert a masked array to an `ndarray` with NaNs at masked points.
     This is used for dask integration, as dask does not support masked arrays.
     Note that any fill value will be lost.
+
     """
     if mask is None:
         mask = array.mask
--- a/lib/iris/_lazy_data.py
+++ b/lib/iris/_lazy_data.py
@@ -68,6 +68,7 @@ def as_lazy_data(data, chunks=_MAX_CHUNK_SIZE):
     if not is_lazy_data(data):
         if isinstance(data, np.ma.MaskedArray):
             data = array_masked_to_nans(data)
+            data = data.data
         data = da.from_array(data, chunks=chunks)
     return data
 
--- a/lib/iris/_merge.py
+++ b/lib/iris/_merge.py
@@ -33,7 +33,7 @@ import dask.array as da
 import numpy as np
 import numpy.ma as ma
 
-from iris._lazy_data import array_masked_to_nans, as_lazy_data, is_lazy_data 
+from iris._lazy_data import array_masked_to_nans, as_lazy_data, is_lazy_data
 import iris.cube
 import iris.coords
 import iris.exceptions
@@ -1231,10 +1231,6 @@ class ProtoCube(object):
                 if is_lazy_data(data):
                     all_have_data = False
                 else:
-                    if isinstance(data, ma.MaskedArray):
-                        if ma.is_masked(data):
-                            data = array_masked_to_nans(data)
-                        data = data.data
                     data = as_lazy_data(data, chunks=data.shape)
                 stack[nd_index] = data
 
--- a/lib/iris/_merge.py
+++ b/lib/iris/_merge.py
@@ -33,7 +33,7 @@ import dask.array as da
 import numpy as np
 import numpy.ma as ma
 
-from iris._lazy_data import is_lazy_data, array_masked_to_nans
+from iris._lazy_data import array_masked_to_nans, as_lazy_data, is_lazy_data 
 import iris.cube
 import iris.coords
 import iris.exceptions
@@ -1235,7 +1235,7 @@ class ProtoCube(object):
                         if ma.is_masked(data):
                             data = array_masked_to_nans(data)
                         data = data.data
-                    data = da.from_array(data, chunks=data.shape)
+                    data = as_lazy_data(data, chunks=data.shape)
                 stack[nd_index] = data
 
             merged_data = _multidim_daskstack(stack)
--- a/lib/iris/fileformats/_pyke_rules/fc_rules_cf.krb
+++ b/lib/iris/fileformats/_pyke_rules/fc_rules_cf.krb
@@ -1003,7 +1003,6 @@ fc_extras
     import warnings
 
     import cf_units
-    import dask.array as da
     import netCDF4
     import numpy as np
     import numpy.ma as ma
@@ -1018,6 +1017,7 @@ fc_extras
     import iris.exceptions
     import iris.std_names
     import iris.util
+    from iris._lazy_data import as_lazy_data
     
     
     #
@@ -1630,7 +1630,7 @@ fc_extras
             proxy = iris.fileformats.netcdf.NetCDFDataProxy(
                 cf_var.shape, dtype, engine.filename,
                 cf_var.cf_name, fill_value)
-            return da.from_array(proxy, chunks=proxy.shape)
+            return as_lazy_data(proxy, chunks=proxy.shape)
 
         # Get any coordinate point data.
         if isinstance(cf_coord_var, cf.CFLabelVariable):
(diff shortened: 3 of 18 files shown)


dkillick:
ping @bjlittle @lbdreyer @pp-mo

lib/iris/fileformats/_pyke_rules/fc_rules_cf.krb
```diff
@@ -1018,6 +1017,7 @@ fc_extras
     import iris.exceptions
     import iris.std_names
     import iris.util
+    from iris._lazy_data import as_lazy_data
```
lbdreyer: Does it matter what order these are in? I just noticed that in the other files the private modules are imported first.

dkillick: Ah, I put this here because I'd seen examples where the private modules were imported last :wink:

lib/iris/tests/unit/lazy_data/test_is_lazy_data.py
```diff
@@ -26,13 +26,13 @@
...
-from iris._lazy_data import is_lazy_data
+from iris._lazy_data import as_lazy_data, is_lazy_data, _MAX_CHUNK_SIZE
 
 
 class Test_is_lazy_data(tests.IrisTest):
     def test_lazy(self):
-        lazy_values = np.arange(30).reshape((2, 5, 3))
-        lazy_array = da.from_array(lazy_values, 1e6)
+        values = np.arange(30).reshape((2, 5, 3))
+        lazy_array = da.from_array(values, _MAX_CHUNK_SIZE)
```
lbdreyer: Minor, but it might be worth using the kwarg i.e.
`lazy_array = da.from_array(values, chunks=_MAX_CHUNK_SIZE)`
to be consistent with the other times da.from_array is called.

lib/iris/_merge.py
```diff
@@ -33,7 +33,7 @@
 import numpy as np
 import numpy.ma as ma
 
-from iris._lazy_data import is_lazy_data, array_masked_to_nans
+from iris._lazy_data import array_masked_to_nans, as_lazy_data, is_lazy_data
```
lbdreyer: I think pep8 is complaining about this trailing whitespace

lib/iris/_merge.py
```diff
@@ -1235,7 +1235,7 @@ def merge(self, unique=True):
                         if ma.is_masked(data):
                             data = array_masked_to_nans(data)
                         data = data.data
-                    data = da.from_array(data, chunks=data.shape)
+                    data = as_lazy_data(data, chunks=data.shape)
```
lbdreyer: The lines doing the mask->nans above this can now be removed as it's being done in `as_lazy_data`

lib/iris/tests/unit/lazy_data/test_is_lazy_data.py
dkillick: Agreed.

lib/iris/_merge.py
dkillick: Good spot!

lib/iris/tests/integration/test_trajectory.py
```diff
@@ -233,7 +233,7 @@ class TestLazyData(tests.IrisTest):
     def test_hybrid_height(self):
         cube = istk.simple_4d_with_hybrid_height()
         # Put a biggus array on the cube so we can test deferred loading.
-        cube.data = da.from_array(cube.data, chunks=cube.data.shape)
+        cube.data = as_lazy_data(cube.data)
```
marqh: i'm interested why you have removed teh chunking here, and in other tests

is there a reason to use the 'magic chunking number' in these cases?

marqh:
i'm not against this change.  I would like to understand a little about it's motivation.  It appears at first glance that calls to
`da.from_array()`
are replaced by
`iris._lazy_data.as_lazy_data()`
calls with a very similar call signature

the only difference I can see is that 
`iris._lazy_data.as_lazy_data()`
always calls
`iris._lazy_data.array_masked_to_nans()`
where as previously this had to be called explicitly if required

is this the benefit which is being targeted?
Is there extra benefit which is being targeted which I am not observing?

Additionally, it seems a shame to reintroduce
```
# A magic value, borrowed from biggus
MAX_CHUNK_SIZE = 8 * 1024 * 1024 * 2
```
to me, I think a better chunking approach is required than this

none of this is a barrier to adoption, I am an interested watcher in this case, keen to comprehend

dkillick:
@marqh we have identified the following benefits with reintroducing `as_masked_array`:

 * factoring repeated code: though there's not a lot of code been factored, if we wanted to change the way that iris interfaced with dask then this means we need only change one piece of code
 * similarly we get to define a single common interface for what to do when constructing a lazy array: we always want to ensure there are no masks present when setting up a lazy array and we want to have a consistent approach to chunking; these are both most efficiently done by having this common interface. 
 * disentangling Iris and dask code: I think it would be good to have as little direct Iris interaction with dask as possible, not least for reasons of factoring common code
 * reintroducing the biggus magic number is an improvement to chunking. The arbitrary scattergun approach to chunking that existed before was naive at best and only likely to negatively impact dask performance. This magic number was actually carefully selected to maximise the performance improvement of chunking while minimising the overhead of setting up chunking, so reintroducing it is a very positive step. Certainly further improvements to how chunks are set up are necessary, but at the moment there is no driver to implement anything here (although having a single function to set up lazy arrays will greatly ease the introduction of it).

lib/iris/tests/integration/test_trajectory.py
dkillick: Primarily this has been done for consistency, but there are also performance improvements to be had for setting the chunk size to the magic chunk size.

lib/iris/tests/unit/lazy_data/test_as_lazy_data.py
```diff
@@ -0,0 +1,65 @@
...
+# (at your option) any later version.
+#
+# Iris is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU Lesser General Public License for more details.
+#
+# You should have received a copy of the GNU Lesser General Public License
+# along with Iris.  If not, see <http://www.gnu.org/licenses/>.
+"""Test :meth:`iris._lazy data.as_lazy_data` method."""
```
lbdreyer:
```
"""Test function :func:`iris._lazy data.as_lazy_data`."""
```

I also just ran into this in the tests I'm writing. It looks the other two unit tests for iris._lazy_data also have this mistake.

marqh:
so
> always calls
> `iris._lazy_data.array_masked_to_nans()`
> where as previously this had to be called explicitly if required
seems a fair assessment of benefit.  I can see that this is deemed useful.

> disentangling Iris and dask code: I think it would be good to have as little direct Iris interaction with dask as possible, not least for reasons of factoring common code

I don't see the benefit of this in principal.  I think Iris using dask is beneficial and Iris should be clear to make calls to dask where useful, I don't think that, in principal, as little direct interaction as possible is a good thing.
In this case it can be seen as a practical step, to handle the work around for missing data, but this isn't an approach that I would advocate across the board.

marqh:
> reintroducing the biggus magic number is an improvement to chunking. The arbitrary scattergun approach to chunking that existed before was naive at best and only likely to negatively impact dask performance. This magic number was actually carefully selected to maximise the performance improvement of chunking while minimising the overhead of setting up chunking, so reintroducing it is a very positive step. Certainly further improvements to how chunks are set up are necessary, but at the moment there is no driver to implement anything here (although having a single function to set up lazy arrays will greatly ease the introduction of it).

this is far from clear to me.  
How is this magic number an improvement to chunking please?  
Is this a performance benefit?  A memory overhead benefit? Have we measured these somewhere that I have not seen?
How is 16777216 carefully selected to deliver performance?

Dask does not provide a magic chunk size, it requests it is provided based on context.  I think it is worth considering whether Iris should respect Dask's API in this case and not provide a 'default value'

It looks in this PR like the only places this is used is in testing, that all of the functional code supplies it's own chunk size.  If this is not the case and I'm missing something, I'd be interested to see it.

With this in mind, I wonder whether not including the magic number in this PR and looking into it as a follow on activity might better suit that introducing it here as part of a non-functional code refactor for purposes discussed above.

lib/iris/tests/unit/lazy_data/test_as_lazy_data.py
dkillick: As in, it should have a `:func:` decorator?

dkillick:
> How is 16777216 carefully selected to deliver performance?

I spoke to @rhattersley about why he chose this number for biggus. What it comes down to is that chunking the data (and in dask, setting up the graph) commands a not insignificant memory/time overhead. The performance tests for biggus showed that setting the chunks too small, as was being done here, ruins performance as any improvement gained through chunking is destroyed through setting up the chunking. Conversely, setting the chunking too large also caused a (less significant) performance impact due to the chunks becoming larger and slower to process.

The value selected is set to be just to the right of peak chunking performance in biggus and I am asserting that this is true in dask too, not least because the [dask docs say](http://dask.pydata.org/en/latest/faq.html?highlight=chunk) (see FAQ 3) that chunks are best set to be between 10MB and 100MB in size, which this number handily is.

I'm going to do some performance (timing) testings on this though, as I'm interested to see how performance will vary by chunk size.

One more thing to note is that you can `rechunk` dask arrays. The dask docs suggest that a lot of the performance improvements in chunking come from setting the _correct chunk size for the operation_. I think that at a later date we should look into rechunking lazy data cubes based upon the operation we're about to perform on it. It is still the case though that setting a default max chunk size of 16MB in both cases will make a noticeable performance improvement than the naive and inconsistent chunk sizes that I have replaced.

dkillick:
> the only places this is used is in testing

Not true:

* almost all operations now use the default chunk size as this will be better in the general case.
* the fileformats interface specifies a chunk size that is the shape of the 2D field coming in.
* in general usage, especially when we have users supplying large datasets, having a default chunk size set to a sensible value is a better approach.

dkillick:
> looking into it as a follow on activity 

This will be much easier done if we have followed the good software design pattern reintroduced here of having a single consistent point of entry for creating dask arrays for Iris cubes rather than having duplicated code scattered all across the Iris codebase.

marqh:
> almost all operations now use the default chunk size as this will be better in the general case.

please may you point me to where this occurs?
all of the calls to the
`iris._lazy_data.as_lazy_data()`
that I can find on this branch explicitly pass in chunk size

dkillick:
Well, there's all the tests - they now use the default chunk size. Cube also uses the default chunk size. The other non-test calls to `as_lazy_data` are all made by code that's about to stack up multiple 2D fields.

lib/iris/tests/unit/lazy_data/test_as_lazy_data.py
lbdreyer: Yep, as it's a function not a method.

lib/iris/_lazy_data.py
```diff
@@ -40,11 +40,45 @@ def is_lazy_data(data):
...
+    Args:
+
+    * data:
+        An array. This will be converted to a lazy dask array.
+
+    Kwargs:
+
+    * chunks:
+        Describes how the created dask array should be split up. Defaults to a
+        value first defined in biggus (being `8 * 1024 * 1024 * 2`).
```
lbdreyer: I do wonder whether the references to biggus here and on line 43 are necessary. I want to be able to read this code without needing to also know the history of Iris lazy data handling.
Not that this is a blocker to this PR.

lib/iris/_lazy_data.py
```diff
@@ -40,11 +40,45 @@ def is_lazy_data(data):
...
+        For more information see
+        http://dask.pydata.org/en/latest/array-creation.html#chunks.
+
+    Returns:
+        The input array converted to a lazy dask array.
+
+    """
+    if not is_lazy_data(data):
+        if isinstance(data, np.ma.MaskedArray):
```
lbdreyer: Is it worth importing `np.ma` rather than just `np`

i.e. 
```
import numpy as np
import numpy.ma as ma
```

lib/iris/_lazy_data.py
dkillick: Yeah, I guess this is a temporary note for us to describe the background to choosing this number in particular...

lib/iris/_lazy_data.py
dkillick: Probably!

lbdreyer:
I'm gonna merge this in as I agree with the principle. 

(@bjlittle @pp-mo I would still encourage you to have a look over this.)

We will have to look into optimising chunking at a later stage.

dkillick:
Thanks @lbdreyer!

dkillick:
> We will have to look into optimising chunking at a later stage.

Agreed. Solid API first, performance improvements thereafter.

PR #2423, opened 2017-03-09
Reuse multidim_daskstack in merge + fast um loading.

lbdreyer:
This reuses the code that recursively builds a multidimensional stacked dask array in two places.


lbdreyer:
@dkillick @pp-mo @bjlittle

lib/iris/_lazy_data.py
```diff
@@ -52,3 +52,31 @@ def array_masked_to_nans(array, mask=None):
...
+
+def multidim_daskstack(stack):
+    """
+    Recursively build a multidimensional stacked dask array.
+
+    This is needed because dask.array.stack only accepts a 1-dimensional list.
+
+    Args:
+
+    * data:
```
dkillick: Your input argument is called "stack", not "data".

lib/iris/fileformats/um/_fast_load_structured_fields.py
```diff
@@ -88,15 +88,11 @@ def data(self):
...
-            data_arrays = [da.from_array(f._data, f._data.shape)
-                           for f in self.fields]
-            vector_dims_list = list(self.vector_dims_shape)
-            vector_dims_list.reverse()
-            self._data_cache = data_arrays
-            for size in vector_dims_list:
-                self._data_cache = [da.stack(self._data_cache[i:i+size]) for i
-                                    in range(0, len(self._data_cache), size)]
-            self._data_cache, = self._data_cache
+            data_arrays = np.array([f._data for f in self.fields],
```
dkillick: I think you need to convert each `f._data` into a dask array here. This isn't done by `multidim_daskstack`, unless calling `da.stack` converts the inputs into dask arrays and then stacks them? Either way, the conversion to arrays is not being done by the common API from #2421, which it probably should be.

lib/iris/tests/unit/lazy_data/test_multidim_daskarray.py
```diff
@@ -0,0 +1,58 @@
...
+import numpy as np
+import dask.array as da
+
+from iris._lazy_data import multidim_daskstack
+
+
+class Test_multidim_daskstack(tests.IrisTest):
+    def test_0d(self):
+        value = 4
+        data = np.array(da.from_array(np.array(value), chunks=1), dtype=object)
```
dkillick: These `from_array` calls will need to be replaced after #2421 has been merged...

lib/iris/tests/unit/lazy_data/test_multidim_daskarray.py
```diff
@@ -0,0 +1,58 @@
...
+                         in vals], dtype=object)
+        result = multidim_daskstack(data)
+        self.assertEqual(result.shape, (2, 3, 3))
+        self.assertArrayEqual(result[:, 0, 0], np.array(vals))
+
+    def test_2d(self):
+        vals = [4, 8, 11]
+        data = np.array([da.from_array(val*np.ones((2, 2)), chunks=2) for val
+                         in vals], dtype=object)
+        data = data.reshape(3, 1, 2, 2)
```
dkillick: I'd argue this data is no longer 2D! Perhaps you could rename this test to `test_nd`?

lib/iris/tests/unit/lazy_data/test_array_masked_to_nans.py
```diff
@@ -14,7 +14,7 @@
 #
 # You should have received a copy of the GNU Lesser General Public License
 # along with Iris.  If not, see <http://www.gnu.org/licenses/>.
-"""Test :meth:`iris._lazy data.array_masked_to_nans` method."""
+"""Test function :func:`iris._lazy data.array_masked_to_nans`."""
```
dkillick: :+1:

lib/iris/tests/unit/lazy_data/test_is_lazy_data.py
```diff
@@ -14,7 +14,7 @@
 #
 # You should have received a copy of the GNU Lesser General Public License
 # along with Iris.  If not, see <http://www.gnu.org/licenses/>.
-"""Test :meth:`iris._lazy data.is_lazy_data` method."""
+"""Test function :func:`iris._lazy data.is_lazy_data`."""
```
dkillick: :+1:

lib/iris/tests/unit/lazy_data/test_multidim_daskarray.py
```diff
@@ -0,0 +1,58 @@
...
+        result = multidim_daskstack(data)
+        self.assertEqual(result.shape, (3, 1, 2, 2))
+        self.assertArrayEqual(result[:, :, 0, 0], np.array(vals).reshape(3, 1))
+
+
+if __name__ == '__main__':
```
dkillick: Given my comment above about whether you have NumPy arrays or dask arrays at the end of this process, you could consider adding a test to check what happens when you pass NumPy arrays in and dask arrays in.

pp-mo: Actually, I don't think we need to worry about what happens if you pass the routine numpy arrays (i.e. a numpy object array containing numpy arrays).
In intended use the argument is as stated "an ndarray of dask arrays", so we should just test that.

lbdreyer:
@dkillick @pp-mo the changes based on you two's reviews are now up!

lib/iris/tests/unit/lazy_data/test_multidim_lazy_stack.py
```diff
@@ -0,0 +1,57 @@
...
+import dask.array as da
+
+from iris._lazy_data import as_lazy_data, multidim_lazy_stack
+
+
+class Test_multidim_lazy_stack(tests.IrisTest):
+    def _check(self, stack_shape):
+        vals = np.arange(np.prod(stack_shape)).reshape(stack_shape)
+        stack = np.empty(stack_shape, 'object')
+        stack_element_shape = (4, 5)
```
dkillick: These magic numbers could probably do with a little explanation...

lib/iris/tests/unit/lazy_data/test_multidim_lazy_stack.py
```diff
@@ -0,0 +1,57 @@
...
+
+    def test_0d_lazy_stack(self):
+        shape = ()
+        result = self._check(shape)
+
+    def test_1d_lazy_stack(self):
+        shape = (2,)
+        result = self._check(shape)
+
+    def test_2d_lazy_stack(self):
```
dkillick: Is there any chance of a >2D test?

lib/iris/fileformats/um/_fast_load_structured_fields.py
```diff
@@ -88,11 +87,11 @@ def data(self):
         if not self._structure_calculated:
             self._calculate_structure()
         if self._data_cache is None:
-            data_arrays = np.array([as_lazy_data(f._data, chunks=f._data.shape)
-                for f in self.fields], dtype=object)
-            data_arrays = data_arrays.reshape(
-                self.vector_dims_shape + data_arrays.shape[1:])
-            self._data_cache = multidim_daskstack(data_arrays)            
+            stack = np.empty(np.prod(self.vector_dims_shape), 'object')
+            for index, f in enumerate(self.fields):
```
dkillick: Haha, I so would have gone for "`i, field`" here...

lbdreyer: Good point. I used f as that was what was used in the previous code. I will change it to 
`for index, field in ...`

lib/iris/tests/unit/lazy_data/test_multidim_lazy_stack.py
lbdreyer: It seems a bit unnecessary? The 2d test is already testing the recursivity

lib/iris/tests/unit/lazy_data/test_multidim_lazy_stack.py
lbdreyer: Not magic numbers. I just chose them as they were small enough that I could produce a 4D result (for `test_2d_lazy_stack` of shape (3, 2, 4, 5) where the length of the dimensions is not repeated.

It's equivalent to a pp field data of shape (4,5)

lib/iris/tests/unit/lazy_data/test_multidim_lazy_stack.py
dkillick: Indeed, however part of testing is checking edge cases, and I _don't know_ what would happen if I passed a 7D array to it. Evidently we don't need to test all possible dimensionalities (!), but it would be good to test something that's outside of the boundary of logical intent for the functionality being tested. This is just in case an unsafe assumption has been made about how the function will be used; "of course, _no-one_ would ever use this for more than a 2D input".

dkillick: I'm not hung up on this though, so I'm not going to make a right fuss if it's left out. @bjlittle @pp-mo do you have any inputs to this?

lib/iris/tests/unit/lazy_data/test_multidim_lazy_stack.py
lbdreyer: Possibly a better way to think of the numbers as "test input", similar to when you create an array to test an array operation, that array is thought of as "test input data".

In these tests, I am creating a stack of `val*np.ones((4,5))` arrays where the shape of each element in the stack is (4,5). So there's nothing special about the numbers I just need something to create an array and then I reuse those numbers to check I'm getting the right output shape.

My worry with the comment "equivalent to a pp field data of shape (4,5)"  is that that is an example of an implementation of multidim_lazy_stack. 

I would consider renaming the variable though. Would `stack_element_dask_array_shape`  be clearer?

dkillick:
@bjlittle @pp-mo do you have any comments to add? Otherwise I'm minded to merge this when Travis is happy.

lib/iris/fileformats/um/_fast_load_structured_fields.py
```diff
@@ -89,15 +87,12 @@ def data(self):
...
-            data_arrays = [as_lazy_data(f._data, chunks=f._data.shape)
-                           for f in self.fields]
-            vector_dims_list = list(self.vector_dims_shape)
-            vector_dims_list.reverse()
-            self._data_cache = data_arrays
-            for size in vector_dims_list:
-                self._data_cache = [da.stack(self._data_cache[i:i+size]) for i
-                                    in range(0, len(self._data_cache), size)]
-            self._data_cache, = self._data_cache
+            stack = np.empty(np.prod(self.vector_dims_shape), 'object')
```
pp-mo: I think it would be neater + clearer to iterate with a multidimensional index here, thus avoiding the ```np.prod(shape)``` and the subsequent ```stack.reshape```.
Your magic friend for that is `np.ndindex(shape)`, which gives you indices to all elements of a multidimensional array, as used [here for example](https://github.com/SciTools/biggus/blob/master/biggus/_init.py#L1812)
I think that gets you something like ...
```
stack = np.empty(self.vector_dims_shape, 'object')
for field, nd_index in zip(fields, np.ndindex(self.vector_dims_shape)):
    stack[nd_index] = as_lazy_data(field._data, chunks=field._data.shape)
self._data_cache = multidim_lazy_stack(stack)
```

lib/iris/tests/unit/lazy_data/test_multidim_lazy_stack.py
pp-mo: From a mockist / white-boxy point of view, the existing tests do already cover all 3 code branches.
The implementation relies on iteration over the passed stack to deconstruct the input one dimension at a time, but 2d is already checking that.
One more would do no harm, I suppose, but I don't really expect it to go wrong !

lib/iris/tests/unit/lazy_data/test_multidim_lazy_stack.py
```diff
@@ -0,0 +1,58 @@
...
+class Test_multidim_lazy_stack(tests.IrisTest):
+    def _check(self, stack_shape):
+        vals = np.arange(np.prod(stack_shape)).reshape(stack_shape)
+        stack = np.empty(stack_shape, 'object')
+        # Define the shape of each element in the stack.
+        stack_element_shape = (4, 5)
+        for index, val in np.ndenumerate(vals):
+            stack[index] = as_lazy_data(val*np.ones(stack_element_shape))
+        result = multidim_lazy_stack(stack)
+        self.assertEqual(result.shape, stack_shape + stack_element_shape)
```
pp-mo: I wonder if we could also check the actual values, to ensure that the ordering of elements is definitely correct ?
I think all that is needed is something like...
```
expected = np.empty(list(stack_shape)+list(stack_element_shape), dtype=int)
for index, val in np.ndenumerate(vals):
    stack[index] = as_lazy_data(val*np.ones(stack_element_shape))
    expected[index] = val
...
self.assertArrayAllClose(result.compute(), expected)
```

lbdreyer: > ensure that the ordering of elements is definitely correct

Surely, that's already achieved by using dimensions of different lengths?

Not that I'm against adding extra assurance.

lbdreyer:
@dkillick @pp-mo I have made the final changes based of your reviews, and tests pass...Merge?

(Please squash and merge)

PR #2424, opened 2017-03-09
Tighten purpose of `array_masked_to_nans`

dkillick:
I think that `array_masked_to_nans` should _always_ return a `numpy.ndarray` and not a `ma.core.masked_array` as used to happen. This means we don't have to call `data = data.data` after every time we use `array_masked_to_nans` in Iris code.

I've also removed the optional `mask` argument from `array_masked_to_nans` as it was only used once and ran the risk of confusing the usage of the function. At its worst, it would have been possible to pass a masked array and an unmatched mask, at which point the function would have returned an array that was not completely filled with `nan`s.

### DIFF
--- a/lib/iris/_lazy_data.py
+++ b/lib/iris/_lazy_data.py
@@ -69,24 +69,43 @@ def as_lazy_data(data, chunks=_MAX_CHUNK_SIZE):
     if not is_lazy_data(data):
         if isinstance(data, ma.MaskedArray):
             data = array_masked_to_nans(data)
-            data = data.data
         data = da.from_array(data, chunks=chunks)
     return data
 
 
-def array_masked_to_nans(array, mask=None):
+def array_masked_to_nans(masked_array):
     """
-    Convert a masked array to an `ndarray` with NaNs at masked points.
+    Convert a masked array to a NumPy `ndarray` filled with NaN values. Input
+    NumPy arrays with no mask are returned unchanged.
     This is used for dask integration, as dask does not support masked arrays.
-    Note that any fill value will be lost.
+
+    Args:
+
+    * masked_array:
+        A NumPy array. If masked, the masked points will be filled with
+        `np.nan` values.
+
+    Returns:
+        The input array if unmasked, or a NaN-filled masked array of
+        floating-point values if masked.
+
+    .. note::
+        The fill value and mask of the input masked array will be lost.
+
+    .. note::
+        Integer masked arrays are cast to floats because NaN is a
+        floating-point value.
 
     """
-    if mask is None:
-        mask = array.mask
-    if array.dtype.kind == 'i':
-        array = array.astype(np.dtype('f8'))
-    array[mask] = np.nan
-    return array
+    if not hasattr(masked_array, 'mask'):
+        result = masked_array
+    else:
+        if masked_array.dtype.kind == 'i':
+            masked_array = masked_array.astype(np.dtype('f8'))
+        mask = masked_array.mask
+        masked_array[mask] = np.nan
+        result = masked_array.data
+    return result
 
 
 def multidim_lazy_stack(stack):
--- a/lib/iris/_lazy_data.py
+++ b/lib/iris/_lazy_data.py
@@ -81,7 +81,7 @@ def array_masked_to_nans(array):
 
     Args:
 
-    * masked_array:
+    * array:
         A NumPy `ndarray` or masked array.
 
     Returns:
@@ -100,10 +100,11 @@ def array_masked_to_nans(array):
     if not ma.isMaskedArray(array):
         result = array
     else:
-        if array.dtype.kind == 'i':
-            array = array.astype(np.dtype('f8'))
-        mask = array.mask
-        array[mask] = np.nan
+        if ma.is_masked(array):
+            if array.dtype.kind == 'i':
+                array = array.astype(np.dtype('f8'))
+            mask = array.mask
+            array[mask] = np.nan
         result = array.data
     return result
 
--- a/lib/iris/_lazy_data.py
+++ b/lib/iris/_lazy_data.py
@@ -73,7 +73,7 @@ def as_lazy_data(data, chunks=_MAX_CHUNK_SIZE):
     return data
 
 
-def array_masked_to_nans(masked_array):
+def array_masked_to_nans(array):
     """
     Convert a masked array to a NumPy `ndarray` filled with NaN values. Input
     NumPy arrays with no mask are returned unchanged.
@@ -82,29 +82,29 @@ def array_masked_to_nans(masked_array):
     Args:
 
     * masked_array:
-        A NumPy array. If masked, the masked points will be filled with
-        `np.nan` values.
+        A NumPy `ndarray` or masked array.
 
     Returns:
-        The input array if unmasked, or a NaN-filled masked array of
-        floating-point values if masked.
+        A NumPy `ndarray`. This is the input array if unmasked, or an array
+        of floating-point values with NaN values where the mask was `True` if
+        the input array is masked.
 
     .. note::
         The fill value and mask of the input masked array will be lost.
 
     .. note::
-        Integer masked arrays are cast to floats because NaN is a
+        Integer masked arrays are cast to 8-byte floats because NaN is a
         floating-point value.
 
     """
-    if not hasattr(masked_array, 'mask'):
-        result = masked_array
+    if not ma.isMaskedArray(array):
+        result = array
     else:
-        if masked_array.dtype.kind == 'i':
-            masked_array = masked_array.astype(np.dtype('f8'))
-        mask = masked_array.mask
-        masked_array[mask] = np.nan
-        result = masked_array.data
+        if array.dtype.kind == 'i':
+            array = array.astype(np.dtype('f8'))
+        mask = array.mask
+        array[mask] = np.nan
+        result = array.data
     return result
 
 
--- a/lib/iris/fileformats/pp.py
+++ b/lib/iris/fileformats/pp.py
@@ -1039,7 +1039,8 @@ def _data_bytes_to_shaped_array(data_bytes, lbpack, boundary_packing,
 
     # Mask the array?
     if mdi in data:
-        data = array_masked_to_nans(data, data == mdi)
+        data = np.ma.masked_array(data, data == mdi)
+        data = array_masked_to_nans(data)
 
     return data
 
--- a/lib/iris/fileformats/pp.py
+++ b/lib/iris/fileformats/pp.py
@@ -1037,10 +1037,11 @@ def _data_bytes_to_shaped_array(data_bytes, lbpack, boundary_packing,
         # Reform in row-column order
         data.shape = data_shape
 
-    # Mask the array?
+    # Convert mdi to NaN.
     if mdi in data:
-        data = np.ma.masked_array(data, data == mdi)
-        data = array_masked_to_nans(data)
+        if data.dtype.kind == 'i':
+            data = data.astype(np.dtype('f8'))
+        data[data == mdi] = np.nan
 
     return data
 
--- a/lib/iris/tests/results/unit/merge/ProtoCube/register__CubeSig/noise.txt
+++ b/lib/iris/tests/results/unit/merge/ProtoCube/register__CubeSig/noise.txt
@@ -4,4 +4,4 @@ failed to merge into a single cube.
   cube.attributes keys differ: 'stuffed'
   cube.cell_methods differ
   cube.shape differs: (3,) != (2,)
-  cube data dtype differs: int64 != float64
+  cube data dtype differs: int64 != int8
--- a/lib/iris/tests/unit/lazy_data/test_array_masked_to_nans.py
+++ b/lib/iris/tests/unit/lazy_data/test_array_masked_to_nans.py
@@ -31,41 +31,45 @@ from iris._lazy_data import array_masked_to_nans
 
 
 class Test(tests.IrisTest):
-    def test_masked(self):
-        masked_array = ma.masked_array([[1.0, 2.0], [3.0, 4.0]],
-                                       mask=[[0, 1], [0, 0]])
-
-        result = array_masked_to_nans(masked_array).data
-
+    def _common_checks(self, result):
         self.assertIsInstance(result, np.ndarray)
         self.assertFalse(isinstance(result, ma.MaskedArray))
         self.assertFalse(ma.is_masked(result))
 
+    def test_masked_input(self):
+        masked_array = ma.masked_array([[1.0, 2.0], [3.0, 4.0]],
+                                       mask=[[0, 1], [0, 0]])
+
+        result = array_masked_to_nans(masked_array)
+
+        self._common_checks(result)
         self.assertArrayAllClose(np.isnan(result),
                                  [[False, True], [False, False]])
         result[0, 1] = 777.7
         self.assertArrayAllClose(result, [[1.0, 777.7], [3.0, 4.0]])
 
+    def test_unmasked_input(self):
+        unmasked_array = np.array([1.0, 2.0])
+        result = array_masked_to_nans(unmasked_array)
+        # Non-masked array is returned as-is, without copying.
+        self.assertIs(result, unmasked_array)
+
     def test_empty_mask(self):
         masked_array = ma.masked_array([1.0, 2.0], mask=[0, 0])
 
-        result = array_masked_to_nans(masked_array).data
+        result = array_masked_to_nans(masked_array)
 
-        self.assertIsInstance(result, np.ndarray)
-        self.assertFalse(isinstance(result, ma.MaskedArray))
-        self.assertFalse(ma.is_masked(result))
-
-        # self.assertIs(result, masked_array.data)
-        # NOTE: Wanted to check that result in this case is delivered without
-        # copying.  However, it seems that ".data" is not just an internal
-        # reference, so copying *does* occur in this case.
+        self._common_checks(result)
         self.assertArrayAllClose(result, masked_array.data)
 
-    def test_non_masked(self):
-        unmasked_array = np.array([1.0, 2.0])
-        result = array_masked_to_nans(unmasked_array, mask=False)
-        # Non-masked array is returned as-is, without copying.
-        self.assertIs(result, unmasked_array)
+    def test_integer_array(self):
+        masked_array = ma.masked_array([[1, 2], [3, 4]],
+                                       mask=[[0, 1], [0, 0]])
+
+        result = array_masked_to_nans(masked_array)
+
+        self._common_checks(result)
+        self.assertEqual(result.dtype, np.dtype('f8'))
 
 
 if __name__ == '__main__':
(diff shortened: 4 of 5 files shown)


lib/iris/_lazy_data.py
```diff
@@ -40,15 +40,36 @@ def is_lazy_data(data):
...
-        mask = array.mask
-    if array.dtype.kind == 'i':
-        array = array.astype(np.dtype('f8'))
-    array[mask] = np.nan
-    return array
+    if not hasattr(masked_array, 'mask'):
+        result = masked_array
+    else:
+        if masked_array.dtype.kind == 'i':
+            masked_array = masked_array.astype(np.dtype('f8'))
```
dkillick: Should we do this for all integer array inputs, not just masked ones?

lib/iris/fileformats/pp.py
```diff
@@ -1040,7 +1040,8 @@ def _data_bytes_to_shaped_array(data_bytes, lbpack, boundary_packing,
 
     # Mask the array?
     if mdi in data:
-        data = array_masked_to_nans(data, data == mdi)
+        data = np.ma.masked_array(data, data == mdi)
```
dkillick: This allows the `mask` kwarg to be dropped from `array_masked_to_nans`. (This is also the only place the kwarg was used!)

marqh:
please note, this behaviour was explicitly added in #2392 
https://github.com/SciTools/iris/commit/d2116d9a30451e96c54f427fb994de3ffb318c06#diff-6f32db0fa1a9558ed4ded7a2425d9fcdL111

enabling the logic for type changing from int to float to be shared between instances where a masked array exists and needs to be converted to a nan array and instances where a numpy array exists with a known condition as to where to put the nan values

these two cases exist in our different file format loaders and I expect both behaviours to be used more

Additionally, the detail change to `iris.fileformats.pp` doesn't make sense to me, 
`data`
at https://github.com/SciTools/iris/pull/2424/files#diff-066f828420b27bfd2a03868477b98415L1043
will never be a masked array, always an array, so it will never get masked
perhaps our testing is lacking here, if this isn't resulting in a test failure

in its current form this proposed change appears to be a regressive step to me

dkillick:
@marqh #2392 is so large as to be unusable, can you point to a particular commit?

> these two cases exist

Then they should probably be handled separately. As it stood, `array_masked_to_nans` could very easily be misused by passing both a masked array and a mask, at which point the masked array's mask would be ignored and the supplied mask used to set nan point. In this case the returned value would be partly nan-filled with some masked points not filled with anything.

> will never be a masked array

I know, that's why I've converted it to a masked array first, using the same logic as was before used to supply a boolean "mask" to the previous `array_masked_to_nans`.

In summary, `array_masked_to_nans` seemed to have become confused in its purpose so I've made a positive assertion so that it does one job well rather than trying to do two overlapping things.

marqh:
> @marqh #2392 is so large as to be unusable, can you point to a particular commit?

There is one commit on this PR, so I have explicitly pointed to the change in the file:
> please note, this behaviour was explicitly added in #2392
> d2116d9#diff-6f32db0fa1a9558ed4ded7a2425d9fcdL111

> I know, that's why I've converted it to a masked array first, using the same logic as was before used to supply a boolean "mask" to the previous array_masked_to_nans.

It was that creation of a mask to immediately remove it again that was identified as an unrequired performance overhead and avoided

pp-mo:
As far as I can see, this single place where the 'mask' keyword is used is also redundant, as this code should not need to use masked arrays at all any more :  All the sources of the 'data' variable that  this processes now produce ***nan-arrays instead***.
Thus, no need for a mask keyword!

However this code as we have it is now mixing masks and NaNs, and I think getting it wrong...
For instance, it now uses NaNs for the missing parts of LBCs (on the dask branch), but then uses mask= to convert that to a masked array with land or sea points masked out.  The two types of "missing" don't get combined any more, which I think is a bug. 
Worse, the dual use of '_data_bytes_to_shaped_array' now looks wrong to me
 -- though again,  not because of any change here ...
  * as '_data_bytes_to_shaped_array' now returns a *nanarray*, this it is ok to wrap a dask wrapper around it + use it as your cube lazy data.  This means the code at [line 883 in ](https://github.com/SciTools/iris/blob/afe7183704fdc5317b9e4bd762665fc6d8979373/lib/iris/fileformats/pp.py#L883) is good
  * but the other call in [_create_field_data when field.data is a LoadedArrayBytes](https://github.com/SciTools/iris/blob/afe7183704fdc5317b9e4bd762665fc6d8979373/lib/iris/fileformats/pp.py#L1868) seems to me now to be ***wrong*** :  When this becomes cube.data, it is real data not lazy, so I think it nevers gets "converted back" from nans to a masked array -- which for Iris use it should do.

I'm investigating that...

dkillick:
> It was that creation of a mask to immediately remove it again that was identified as an unrequired performance overhead and avoided

Then I guess the PP loader does not need to call `array_masked_to_nans` at all! And this very much reinforces the fact that `array_masked_to_nans` does not require a `mask` kwarg.

marqh:
> Then I guess the PP loader does not need to call array_masked_to_nans at all

it shares the logic about type coercion where nans are required, i.e. the change from int to float. That is the logic which was factored out into the `_lazy_data` module on request that it be shared

in order to avoid the call at all, this logic would have to be copied back into the `fileformats.pp` module

lib/iris/fileformats/pp.py
marqh: on reflection, both of these calls (previous and proposed) are creating a 'mask array', so perhaps the concerns over creating more arrays which I raised are less important than I thought

It seems odd to create a masked array and immediately turn that into another numpy array but perhaps it is a reasonable approach in this case

dkillick: @marqh we're seeing that there's no obvious answer to this. When I discussed this offline with @pp-mo he raised similar concerns about what was happening here...

dkillick: So all we're trying to do here is replace instances of `mdi` values with `nan` values (and cast int --> float). Both ways that have been tried so far have ended up producing some sort of mask that has ultimately been thrown away; I wonder if we need another way of making this conversion?

As a follow-on question; do we need `array_masked_to_nans` at all - will there _ever_ be masked arrays at any of the places this function is used that will need to be nan-filled?

dkillick:
Now with extra added rebase after #2421 got merged!

lib/iris/fileformats/pp.py
marqh: i think in my previous implementation this was all done in line and the reassignment happened conditionally

in factoring this out to share the type change code, i think this performance benefit was lost; I think that your refactoring just highlighted its loss

I am tempted to suggest that sharing code is one module is not always the best thing to do and this may be a context where it is easier to just process the data in line and accept and recognise a bit of shared logic, it is not like it is a lot of code

something like
https://github.com/SciTools/iris/blob/dask/lib/iris/fileformats/pp.py#L1040
```
    if mdi in data:
        if masked_array.dtype.kind == 'i':
            masked_array = masked_array.astype(np.dtype('f8'))
        data[data == mdi] = np.nan
```
perhaps with a code comment that this logic is reproduced here to avoid creating transient arrays or somesuch

lib/iris/fileformats/pp.py
```diff
@@ -1037,9 +1037,11 @@ def _data_bytes_to_shaped_array(data_bytes, lbpack, boundary_packing,
         # Reform in row-column order
         data.shape = data_shape
 
-    # Mask the array?
+    # Convert mdi to NaN.
     if mdi in data:
-        data = array_masked_to_nans(data, data == mdi)
```
dkillick: If we pay attention to the name of `array_masked_to_nans`, it tells us that it will convert a masked array to a nan-filled array. If we don't have masked data (such as here), let's just not use it and instead do the processing we actually need to. That is, we cast int -> float if necessary and replace `mdi` values with `NaN`.

lib/iris/tests/unit/lazy_data/test_array_masked_to_nans.py
```diff
@@ -31,41 +31,45 @@
...
-    def test_masked(self):
-        masked_array = ma.masked_array([[1.0, 2.0], [3.0, 4.0]],
-                                       mask=[[0, 1], [0, 0]])
-
-        result = array_masked_to_nans(masked_array).data
-
+    def _common_checks(self, result):
         self.assertIsInstance(result, np.ndarray)
         self.assertFalse(isinstance(result, ma.MaskedArray))
         self.assertFalse(ma.is_masked(result))
```
pp-mo: Isn't this redundant, given the previous line ?
(`is_masked` meaning that : "it is a masked array, and some points are actually masked")

Based on my previous comment about "standard facilities", perhaps we should just be using `self.assertFalse(ma.isMaskedArray(result))`.

lib/iris/tests/unit/lazy_data/test_array_masked_to_nans.py
```diff
@@ -31,41 +31,45 @@
...
         result[0, 1] = 777.7
         self.assertArrayAllClose(result, [[1.0, 777.7], [3.0, 4.0]])
 
+    def test_unmasked_input(self):
+        unmasked_array = np.array([1.0, 2.0])
+        result = array_masked_to_nans(unmasked_array)
+        # Non-masked array is returned as-is, without copying.
+        self.assertIs(result, unmasked_array)
+
     def test_empty_mask(self):
```
pp-mo: Masked arrays being the horrible things they are, it might be worth checking also the case where the 'mask' keyword is not set (the actual default is  "mask=np.ma.nomask").
For a ghastly distinction, this is not the same thing as `mask=zeros(...)` or `mask=None` or `mask=False`, all of which allocate an actual (empty) mask _array_: 
```
>>> print np.ma.masked_array([1, 2], mask=None).mask
[False False]
>>> print np.ma.masked_array([1, 2]).mask
False
>>> print np.ma.masked_array([1, 2], mask=np.ma.nomask).mask
False
>>> 
```
Urrrgh! Don't get me started...

lib/iris/tests/unit/lazy_data/test_array_masked_to_nans.py
```diff
@@ -31,41 +31,45 @@
...
-        # reference, so copying *does* occur in this case.
+        self._common_checks(result)
         self.assertArrayAllClose(result, masked_array.data)
 
-    def test_non_masked(self):
-        unmasked_array = np.array([1.0, 2.0])
-        result = array_masked_to_nans(unmasked_array, mask=False)
-        # Non-masked array is returned as-is, without copying.
-        self.assertIs(result, unmasked_array)
+    def test_integer_array(self):
```
pp-mo: This tests "masked integer" input.
We should have another one that tests "un-masked integers", where the input is a non-masked integer array, as in that case the dtype is preserved.
Also I suppose yet another, to test an integer masked array with no masked points (like 'test_empty_mask' above).

lib/iris/_lazy_data.py
```diff
@@ -69,21 +69,40 @@ def as_lazy_data(data, chunks=_MAX_CHUNK_SIZE):
     if not is_lazy_data(data):
         if isinstance(data, ma.MaskedArray):
             data = array_masked_to_nans(data)
-            data = data.data
         data = da.from_array(data, chunks=chunks)
     return data
 
 
-def array_masked_to_nans(array, mask=None):
+def array_masked_to_nans(masked_array):
```
dkillick: > the argument name should be changed

Good point! Shows how much change there's been to this function recently...

> Possibly we should change the function name too

To what though? Primarily what it does is put `nan`s where before there were masked values, which is what the function says it does. If we wanted to describe the non-masked array pass through too then we might end up with:
`array_masked_to_nans_unless_the_array_isnt_masked_at_which_point_just_return_it`
... which is a bit long (and very Pratchett-ian! :tophat:)

Do you have a better name suggestion? Otherwise I propose we don't touch it.

lib/iris/_lazy_data.py
```diff
@@ -69,21 +69,40 @@ def as_lazy_data(data, chunks=_MAX_CHUNK_SIZE):
...
+    Convert a masked array to a NumPy `ndarray` filled with NaN values. Input
+    NumPy arrays with no mask are returned unchanged.
     This is used for dask integration, as dask does not support masked arrays.
-    Note that any fill value will be lost.
+
+    Args:
+
+    * masked_array:
+        A NumPy array. If masked, the masked points will be filled with
+        `np.nan` values.
```
dkillick: Sure.

lib/iris/_lazy_data.py
```diff
@@ -69,21 +69,40 @@ def as_lazy_data(data, chunks=_MAX_CHUNK_SIZE):
...
+
+    Returns:
+        The input array if unmasked, or a NaN-filled masked array of
```
dkillick: Ah yes, a clunky description of what is happening. 

"Masked points are set to `nan`" might be better? I'll also add the bit about the "normal" array too.

lib/iris/_lazy_data.py
```diff
@@ -69,21 +69,40 @@ def as_lazy_data(data, chunks=_MAX_CHUNK_SIZE):
...
+        floating-point values if masked.
+
+    .. note::
+        The fill value and mask of the input masked array will be lost.
+
+    .. note::
+        Integer masked arrays are cast to floats because NaN is a
```
dkillick: Agreed.

lib/iris/tests/unit/lazy_data/test_array_masked_to_nans.py
dkillick: Sure.

FWIW, I inherited these checks. I kept these last two because I wanted to be _absolutely sure_ that my non-masked array really hadn't retained its mask somehow!

lib/iris/tests/unit/lazy_data/test_array_masked_to_nans.py
dkillick: I wonder if this is getting a little close to testing the actual `ma` implementation! Which, perhaps, it could do with a little more of...

dkillick:
Review actions to address comments from @pp-mo.

lib/iris/tests/unit/lazy_data/test_array_masked_to_nans.py
```diff
@@ -62,6 +68,14 @@ def test_empty_mask(self):
...
 
+    def test_no_mask(self):
+        masked_array = ma.masked_array([1.0, 2.0], mask=ma.nomask)
+
+        result = array_masked_to_nans(masked_array)
+
+        self._common_checks(result)
+        self.assertArrayAllClose(result, masked_array.data)
+
     def test_integer_array(self):
```
pp-mo: I think it is worth renaming this `test_masked_ints` to make clear the relationship to `test_unmasked_ints`.
Also, as this is the one that changes the type, it is worth checking the actual values, as in `test_masked_input`.

lib/iris/tests/unit/lazy_data/test_array_masked_to_nans.py
```diff
@@ -33,7 +33,7 @@
 class Test(tests.IrisTest):
     def _common_checks(self, result):
         self.assertIsInstance(result, np.ndarray)
-        self.assertFalse(isinstance(result, ma.MaskedArray))
+        self.assertFalse(ma.isMaskedArray(result))
         self.assertFalse(ma.is_masked(result))
```
pp-mo: This line is still redundant, as we already checked it's not a masked array

dkillick: I really disagree.

pp-mo: ?? what does `self.assertFalse(ma.isMaskedArray(result))` mean to you.
```
isMaskedArray(x)
    Test whether input is an instance of MaskedArray.
    
    This function returns True if `x` is an instance of MaskedArray
    and returns False otherwise.  Any object is accepted as input.

```
If the type is not a masked array, it cannot be masked, so why do you want to check this ?

dkillick: Alright, I give in...

dkillick:
Review actions: tests renamed and excess test removed.

@bjlittle @lbdreyer do you have anything to add here? It would be nice to get this in.

pp-mo:
Suggested change : https://github.com/dkillick/iris/pull/8

Other than that, I'm good with this + will merge when other have had a chance..

dkillick:
OK, I've changed `array_masked_to_nans` so that it won't recast int masked array objects with `mask=ma.nomask`.

lib/iris/_lazy_data.py
```diff
@@ -69,21 +69,41 @@ def as_lazy_data(data, chunks=_MAX_CHUNK_SIZE):
...
     """
-    Convert a masked array to an `ndarray` with NaNs at masked points.
+    Convert a masked array to a NumPy `ndarray` filled with NaN values. Input
+    NumPy arrays with no mask are returned unchanged.
     This is used for dask integration, as dask does not support masked arrays.
-    Note that any fill value will be lost.
+
+    Args:
+
+    * masked_array:
```
bjlittle: @dkillick This should be `array` surely, and not `masked_array` ...

dkillick: Bah! So it should.

lib/iris/tests/unit/lazy_data/test_array_masked_to_nans.py
```diff
@@ -31,42 +31,72 @@
...
-        result = array_masked_to_nans(masked_array).data
-
-        self.assertIsInstance(result, np.ndarray)
-        self.assertFalse(isinstance(result, ma.MaskedArray))
-        self.assertFalse(ma.is_masked(result))
+        result = array_masked_to_nans(masked_array)
 
+        self._common_checks(result)
         self.assertArrayAllClose(np.isnan(result),
                                  [[False, True], [False, False]])
```
lbdreyer: You could have always reused the mask defined on line 40.

dkillick:
Oh good, the tests have passed! 

@pp-mo @lbdreyer @bjlittle nudge nudge nudge...

:shipit:

lib/iris/tests/results/unit/merge/ProtoCube/register__CubeSig/noise.txt
```diff
@@ -4,4 +4,4 @@ failed to merge into a single cube.
   cube.attributes keys differ: 'stuffed'
   cube.cell_methods differ
   cube.shape differs: (3,) != (2,)
-  cube data dtype differs: int64 != float64
\ No newline at end of file
+  cube data dtype differs: int64 != int8
```
dkillick: So this is fun: it looks like the changes in this PR have [fixed this test result](https://github.com/SciTools/iris/blob/master/lib/iris/tests/unit/merge/test_ProtoCube.py#L298) -- note how the dtype specified in the error text here now matches to the dtype _set in the test code_...

lbdreyer: So this was what it originally was, but as part of the previous dask work it was changed to `float64` because the cube data array was a masked array. However the masked array doesn't contain any masked values so the dask array wouldn't contain any nans and therefore could be int.

PR #2427, opened 2017-03-10
CML cube file_value and dtype refactor.

bjlittle:
This PR addresses the issue of adding `fill_value` and `dtype` to the metadata of the cube.

We now need to see this shift in metadata reflected in the CML test results.

Also, this PR aligns with the new philosophy that we don't care about the `fill_value` on any concrete cube masked data, but we do care about the `fill_value` on the cube.

Note that, the associated CML for skipped tests have not be altered. We will trip over this change later, as we unskip those CML tests.

### DIFF
--- a/lib/iris/cube.py
+++ b/lib/iris/cube.py
@@ -1624,15 +1624,29 @@ bound=(1994-12-01 00:00:00, 1998-12-01 00:00:00)
 
     @property
     def dtype(self):
-        if self._dtype is None:
-            result = self.core_data.dtype
-        else:
-            result = self._dtype
-        return result
+        dtype = self._dtype
+        if dtype is None:
+            dtype = self.core_data.dtype
+        return dtype
 
     @dtype.setter
     def dtype(self, dtype):
         self._dtype = dtype
+
+    @property
+    def fill_value(self):
+        fill_value = self._fill_value
+        if fill_value is not None:
+            try:
+                fill_value = np.asarray([self._fill_value],
+                                        dtype=self.dtype)[0]
+            except OverflowError:
+                fill_value = None
+        return fill_value
+
+    @fill_value.setter
+    def fill_value(self, fill_value):
+        self._fill_value = fill_value
 
     @property
     def ndim(self):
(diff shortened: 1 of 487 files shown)


lib/iris/cube.py
```diff
@@ -1624,17 +1624,31 @@ def shape(self):
...
     def dtype(self):
-        if self._dtype is None:
-            result = self.core_data.dtype
-        else:
-            result = self._dtype
-        return result
+        dtype = self._dtype
+        if dtype is None:
+            dtype = self.core_data.dtype
+        return dtype
```
bjlittle: This is just a simple refactor. There is no functional change here.

lib/iris/cube.py
```diff
@@ -1624,17 +1624,31 @@ def shape(self):
...
 
     @property
+    def fill_value(self):
+        fill_value = self._fill_value
+        if fill_value is not None:
+            try:
+                fill_value = np.asarray([self._fill_value],
+                                        dtype=self.dtype)[0]
+            except OverflowError:
+                fill_value = None
```
bjlittle: The `fill_value` requires to be reported within the context of the `dtype`, what ever that is.

We need to catch the case where, for example, a default float `fill_value`, cannot be cast to an `int` when the `dtype` of the cube is changed.

In this case, I've opted to not clearing the `self._fill_value`, so that if the `cube.dtype` change is reverted the "old" `fill_value` is still there. Willing to take advice/thoughts on this ...

lib/iris/cube.py
```diff
@@ -2892,6 +2906,9 @@ def _xml_element(self, doc, checksum=False, order=True, byteorder=True):
         if self.var_name:
             cube_xml_element.setAttribute('var_name', self.var_name)
         cube_xml_element.setAttribute('units', str(self.units))
+        if self.fill_value is not None:
+            cube_xml_element.setAttribute('fill_value', str(self.fill_value))
+        cube_xml_element.setAttribute('dtype', self.dtype.name)
```
bjlittle: Add the cube `fill_value` and `dtype` to the CML of the cube - these attributes are akin to the metadata of the cube.

lib/iris/cube.py
```diff
@@ -2970,8 +2987,6 @@ def normalise(data):
                 else:
                     crc = 'no-masked-elements'
                 data_xml_element.setAttribute("mask_checksum", crc)
-                data_xml_element.setAttribute('fill_value',
-                                              str(data.fill_value))
```
bjlittle: We now don't care about the `fill_value` on masked data.

lib/iris/cube.py
dkillick: Might be worth some sort of note to the user as to _why_ they're getting `None` as the cube's dtype in such a case.

I think it hits plausible use-cases though :+1:

lib/iris/cube.py
dkillick: Have you also removed where (if anywhere) the data's `fill_value` and `dtype` were added into the XML? We don't want them being set twice!

dkillick:
@bjlittle just a couple of thoughts from me!

PR #2430, opened 2017-03-13
Make image has tests register as failures rather than errors

djkirkham:
When an image hash check fails, a `ValueError` is raised, causing unit tests to register as errors rather than failures. This PR changes `ValueError` to `AssertionError`.


marqh:
agreed, much neater pattern for reporting

PR #2433, opened 2017-03-13
fill_value and dtype

bjlittle:
This PR addresses the issue of dealing with the `dtype` and `fill_value` of the cube.

We need to deal with fact that `dask` does not support masked arrays, and so we need to take care to correctly handle the intended `dtype` and `fill_value` of the data payload in the cube.

The `dtype` and `fill_value` have now been promoted to be cube metadata attributes, because we care about them and define what a cube is. In the case of a `dask` lazy, masked data payload, we need to preserve the case where the intended `dtype` is integral.

We also have a shift in how `fill_value` is handled, in that previously we handed that off to `biggus` and `numpy.ma` in the hope that it would do the right thing. Now we deal with this directly within the cube through `cube.fill_value`, which is used as the intended `fill_value` of any masked data payload.

### DIFF
--- a/lib/iris/_concatenate.py
+++ b/lib/iris/_concatenate.py
@@ -28,6 +28,7 @@ from copy import deepcopy
 
 import dask.array as da
 import numpy as np
+import numpy.ma as ma
 
 import iris.coords
 import iris.cube
@@ -329,6 +330,16 @@ class _CubeSignature(object):
 
         self.defn = cube.metadata
         self.data_type = cube.dtype
+        # Concatenation will generate resultant cubes that contain lazy data.
+        # We need to deal with the special case of preserving the dtype of a
+        # candidate cube with concrete masked integral data by ensuring that
+        # the cube metadata has the correct target dtype.
+        kwargs = self.defn._asdict()
+        if kwargs['dtype'] is None and not cube.has_lazy_data() and \
+                isinstance(cube.data, ma.masked_array) and \
+                cube.data.dtype.kind == 'i':
+            kwargs['dtype'] = self.data_type
+            self.defn = iris.cube.CubeMetadata(**kwargs)
 
         #
         # Collate the dimension coordinate metadata.
@@ -402,6 +413,14 @@ class _CubeSignature(object):
                       ', '.join(diff_names))
         return result
 
+    def promote_defn(self):
+        defn = self.defn
+        kwargs = defn._asdict()
+        if kwargs['dtype'] is None and self.data_type.kind == 'i':
+            kwargs['dtype'] = self.data_type
+            defn = iris.cube.CubeMetadata(**kwargs)
+        return defn
+
     def match(self, other, error_on_mismatch):
         """
         Return whether this _CubeSignature equals another.
@@ -433,10 +452,14 @@ class _CubeSignature(object):
 
         # Check cube definitions.
         if self.defn != other.defn:
-            # Note that the case of different phenomenon names is dealt with
-            # in :meth:`iris.cube.CubeList.concatenate_cube()`.
-            msg = 'Cube metadata differs for phenomenon: {}'
-            msgs.append(msg.format(self.defn.name()))
+            promoted = self.promote_defn()
+            if promoted != other.promote_defn():
+                # Note that the case of different phenomenon names is dealt
+                # with in :meth:`iris.cube.CubeList.concatenate_cube()`.
+                msg = 'Cube metadata differs for phenomenon: {}'
+                msgs.append(msg.format(self.defn.name()))
+            else:
+                self.defn = promoted
         # Check dim coordinates.
         if self.dim_metadata != other.dim_metadata:
             differences = self._coordinate_differences(other, 'dim_metadata')
--- a/lib/iris/_concatenate.py
+++ b/lib/iris/_concatenate.py
@@ -414,6 +414,12 @@ class _CubeSignature(object):
         return result
 
     def promote_defn(self):
+        """
+        Update the metadata definition to match that of a similar but lazy cube.
+        A cube with lazy masked integral data must have its :attr:`metadata.dtype`
+        set appropriately.
+
+        """
         defn = self.defn
         kwargs = defn._asdict()
         if kwargs['dtype'] is None and self.data_type.kind == 'i':
@@ -452,6 +458,8 @@ class _CubeSignature(object):
 
         # Check cube definitions.
         if self.defn != other.defn:
+            # Attempt to match metadata for the lazy masked integral
+            # dtype case.
             promoted = self.promote_defn()
             if promoted != other.promote_defn():
                 # Note that the case of different phenomenon names is dealt
@@ -459,6 +467,7 @@ class _CubeSignature(object):
                 msg = 'Cube metadata differs for phenomenon: {}'
                 msgs.append(msg.format(self.defn.name()))
             else:
+                # Persist the promoted metadata dtype match case.
                 self.defn = promoted
         # Check dim coordinates.
         if self.dim_metadata != other.dim_metadata:
--- a/lib/iris/_concatenate.py
+++ b/lib/iris/_concatenate.py
@@ -415,9 +415,9 @@ class _CubeSignature(object):
 
     def promote_defn(self):
         """
-        Update the metadata definition to match that of a similar but lazy cube.
-        A cube with lazy masked integral data must have its :attr:`metadata.dtype`
-        set appropriately.
+        Update the metadata definition to match that of a similar but lazy
+        cube. A cube with lazy masked integral data must have its
+        :attr:`metadata.dtype` set appropriately.
 
         """
         defn = self.defn
--- a/lib/iris/cube.py
+++ b/lib/iris/cube.py
@@ -65,8 +65,8 @@ class CubeMetadata(collections.namedtuple('CubeMetadata',
                                            'units',
                                            'attributes',
                                            'cell_methods',
-                                           'fill_value',
-                                           'dtype'])):
+                                           'dtype',
+                                           'fill_value'])):
     """
     Represents the phenomenon metadata for a single :class:`Cube`.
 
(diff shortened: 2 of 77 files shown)


lib/iris/cube.py
```diff
@@ -1632,7 +1638,33 @@ def dtype(self):
 
     @dtype.setter
     def dtype(self, dtype):
-        self._dtype = dtype
+        if dtype != self.dtype:
```
pp-mo: Should we not also allow here `cube.dtype = None`, to reset the dtype to the underlying data ?

bjlittle: I can see the case that you might want to unset the dtype for lazy data to `None`, which you can't at the moment given the current implementation ... I've no argument against that.

pp-mo:
Hi @bjlittle highly impressed you have this passing now !!
Will review later...

In meantime, here's an idea to ponder on...
I analysed usecases as I was suggesting, especially for the "minimal" requirements.
The only important outcome was this :  I suspect we don't really need dtype assignment, we can allow dtype (that is, the 'alternative real dtype') to be specified ***only*** when setting lazy data.

I think we will need to retain real-data reassignment "cube.data =", but we can probably do without *lazy-data* reassignment (I need to check out existing code).  Even if not, we can define a particular function for this.  
This would mean that 'real dtype' can *only* be supplied in association with new lazy data.
I think removing the dtype assignment makes the whole API cleaner and simpler (including simpler *to explain*).

Think on, will discuss later.

bjlittle:
```shell
Ran 4006 tests in 104.584s

FAILED (SKIP=310, failures=38)
```
- [ ] `iris.tests.unit.analysis.regrid.test_RectilinearRegridder.Test___call____circular` (x1)
- [ ] `iris.tests.test_basic_maths` (x23)
- [ ] `iris.tests.test_analysis` (x10)
- [ ] `iris.tests.test_concatenate.Test2D` (x3)
- [ ] `iris.tests.test_interpolation.TestNearestLinearInterpolRealData` (x1)

lbdreyer:
Sorry @bjlittle! I accidentally pushed (still getting the hang of pycharm) the change
https://github.com/bjlittle/iris/commit/2b77106939ce5f1a443df4db520c79ea4e6a0acd
directly to this branch.

You can always revert it if you diagree with it

bjlittle:
@pp-mo @lbdreyer @dkillick Holy smoke, the tests are passing ...

pp-mo:
> Holy smoke, the tests are passing ...

!!!  :+1: :champagne: !!!

lib/iris/cube.py
```diff
@@ -751,10 +751,13 @@ def __init__(self, data, standard_name=None, long_name=None,
         # Cell Measures
         self._cell_measures_and_dims = []
 
+        # We need to set the dtype before the fill_value,
+        # as the fill_value is checked against self.dtype.
+        self._dtype = None
+        if dtype is not None:
+            self.dtype = dtype
```
bjlittle: I don't think we need to have this check here, we just need to initialise `_dtype` and then set `self.dtype = dtype`

lib/iris/cube.py
```diff
@@ -1632,7 +1638,38 @@ def dtype(self):
...
-        self._dtype = dtype
+        if dtype != self.dtype:
+            if dtype is not None:
+                if not self.has_lazy_data():
+                    emsg = 'Cube does not have lazy data, cannot set dtype.'
+                    raise ValueError(emsg)
+                dtype = np.dtype(dtype)
+                if dtype.kind != 'i':
+                    emsg = ('Can only cast lazy data to integral dtype, '
+                            'got {!r}.')
```
lbdreyer: Why?

bjlittle: We need to be strict here. We're only allowing the `dtype` to be of integral type to support the masked case. If we didn't do this, then users would abuse this to perform casting of their data.

If we do want to support casting of data, for the sake of it, then we should support that properly, say with `cube.astype(...)` or whatever. Hence, that's why we nail this `dtype` change down to only being applied to lazy, integral data.

lib/iris/cube.py
lbdreyer: Agreed

lib/iris/cube.py
pp-mo: Just a note: if you're visiting this bit, I've found that this test should also be allowing 'u' type data, i.e. unsigned ints.  E.G. `if dtype.kind not in ('i', 'u'):` ?
If not it can wait -- I think it could do with a more general code search anyway, as I think we may have missed this possibility elsewhere in the code too.

bjlittle: Hmmm good point @pp-mo ... See this [concatenate](https://github.com/SciTools/iris/pull/2433/files#diff-5fa4649482766af96ae11aead017e334R340) change ... is there a generic numpy kind to catch all integral kinds instead of `i` and `u` ?

bjlittle: This [numpy scalars link](http://numpy.readthedocs.io/en/latest/reference/arrays.scalars.html#) was handy ...

And this pattern seem to work:
```python
>>> integral = [np.arange(10, dtype=np.int8),
... np.arange(10, dtype=np.int16),
... np.arange(10, dtype=np.int32),
... np.arange(10, dtype=np.int64),
... np.arange(10, dtype=np.uint8),
... np.arange(10, dtype=np.uint16),
... np.arange(10, dtype=np.uint32),
... np.arange(10, dtype=np.uint64)]
>>> for i in integral:
...     print(isinstance(i[0], np.integer))
... 
True
True
True
True
True
True
True
True
>>> f = np.arange(10, dtype=float)
>>> isinstance(f, np.integer)
False
```

lib/iris/_concatenate.py
```diff
@@ -402,6 +413,14 @@ def _coordinate_differences(self, other, attr):
                       ', '.join(diff_names))
         return result
 
+    def promote_defn(self):
+        defn = self.defn
+        kwargs = defn._asdict()
+        if kwargs['dtype'] is None and self.data_type.kind == 'i':
+            kwargs['dtype'] = self.data_type
+            defn = iris.cube.CubeMetadata(**kwargs)
+        return defn
```
bjlittle: I should have some unit tests for this really ...

lib/iris/cube.py
lbdreyer: >We're only allowing the dtype to be of integral type to support the masked case

It might be worth putting a comment in the code about this

lib/iris/cube.py
```diff
@@ -1632,7 +1637,38 @@ def dtype(self):
...
+
+    @property
+    def fill_value(self):
+        return self._fill_value
+
+    @fill_value.setter
+    def fill_value(self, fill_value):
+        if fill_value is not None:
+            # Convert the given value to the dtype of the cube.
+            fill_value = np.asarray([fill_value])[0]
```
lbdreyer: How about
```
fill_value, = np.asarray([fill_value])
```

lib/iris/cube.py
pp-mo: I only mentioned the kind=u/i thing for completeness..
Let's not get hung up on this, please.
It will be much simpler to fix it later as I'm sure it needs doing in unrelated bits of the code.

lib/iris/cube.py
bjlittle: I'm tending to steer away from the `fill_value, = np.asarray([fill_value])` pattern as it's quite subtle.

However, we could opt for `[fill_value] = np.asarray([fill_value])` instead ...

lib/iris/cube.py
lbdreyer: I do prefer np.integer. 
It will also need to be fixed in iris/_concatenate.py e.g. https://github.com/SciTools/iris/pull/2433/files#diff-5fa4649482766af96ae11aead017e334R340

bjlittle: @pp-mo can you make an issue for this please

lib/iris/_concatenate.py
bjlittle: Actually, I'm not going to do that here as this might all change given conversations with @pp-mo

lib/iris/cube.py
bjlittle: I'll make this change ...

lib/iris/_concatenate.py
bjlittle: Created the following issue https://github.com/SciTools/iris/issues/2439 for this

lib/iris/tests/test_concatenate.py
```diff
@@ -318,42 +335,78 @@ def test_order(self):
         result = concatenate(cubes)
         self.assertEqual(len(result), 2)
 
+    def test_masked_fill_value(self):
```
lbdreyer: I do wonder whether "test_masked_diff_fill_value" would be better

lib/iris/_concatenate.py
```diff
@@ -402,6 +413,14 @@ def _coordinate_differences(self, other, attr):
                       ', '.join(diff_names))
         return result
 
+    def promote_defn(self):
+        defn = self.defn
```
pp-mo: I think this is worthy of explanation !
Suggest: 
 * call it "lazy_version_defn"
 * docstring like ....
```
Produce the defn (i.e. cube.metadata) which a lazy version of the cube would have.
A cube with lazy data representing masked ints must have its `metadata.dtype` set appropriately.
```




**Even if** we think this may all change again later ...

lib/iris/_concatenate.py
```diff
@@ -433,10 +452,14 @@ def match(self, other, error_on_mismatch):
 
         # Check cube definitions.
         if self.defn != other.defn:
-            # Note that the case of different phenomenon names is dealt with
-            # in :meth:`iris.cube.CubeList.concatenate_cube()`.
-            msg = 'Cube metadata differs for phenomenon: {}'
-            msgs.append(msg.format(self.defn.name()))
+            promoted = self.promote_defn()
```
pp-mo: Explain here too.  Something like "Check whether lazy versions of the defns would match."

bjlittle:
@pp-mo @lbdreyer Let's get this merged asap ....

lbdreyer:
The tests are failing due to pep8 @bjlittle

lbdreyer:
Once this tests are passing  I'll merge this in

lbdreyer:
Are you happy for me to squash and merge (once the tests have passed)?

bjlittle:
Ping @lbdreyer :+1:

bjlittle:
Ping @pp-mo @dkillick

lbdreyer:
Sorry about the wait. @pp-mo wanted to do some final checks

Issue #2425, opened 2017-03-10
BUG: Iris coords bounds indexing broken with multiple ellipsis


cpelley:
In numpy v1.12.0 it is no longer simply deprecated to index with a double ellipsis, it is now enforced by an IndexError exception raised.  [Here](https://github.com/numpy/numpy/commit/c87c7a612ee7f90281797e447255c402b099c73a) is the relevant numpy change.

This exposes a bug in iris as `iris.util.column_slices_generator` can return an ellipsis, then an additional ellipsis is also appended [here](https://github.com/SciTools/iris/blob/baef12627c5d840ccda7533c0fbcbcf629ee80cd/lib/iris/coords.py#L495).

```Python
>>> import numpy as np
>>> import iris.coords.DimCoord as DimCoord
>>> np.__version__
'1.12.0'
>>> coord = DimCoord([-1,  0,  1,  2])
>>> coord.guess_bounds()
>>> coord[np.array([0, 3])]
*** IndexError: an index can only have a single ellipsis ('...')
```

Here is a partial traceback:
```Python
...
    coord = super(DimCoord, self).__getitem__(key)
  File "/path/to/iris/coords.py", line 496, in __getitem__
    bounds = bounds[keys + (Ellipsis, )]
IndexError: an index can only have a single ellipsis ('...')
```

Workaround is to explicitly specify numpy as one of our dependencies and pin its version to 1.11

cpelley:
I'll post an example here shortly illustrating the problem, however this exposes a problem with the travis testing.  I notice that the irisi travis tests still appear to be using numpy v1.10 but I get numpy v1.12 (the latest) via a conda install iris through conda forge.  Since iris supports later versions of numpy, I think this exposes a frailty here.

marqh:
this is being actively worked on by myself and @djkirkham 

i know there are changes staged for the coord API to avoid double ellipsis
suggestions for increasing test coverage in this realm are most welcome, please notify @djkirkham and I of these

cpelley:
> this is being actively worked on by myself and @djkirkham

Great news, let me know if I can help in any way.

marqh:
> Great news, let me know if I can help in any way.

writing a test case for us which catches the failure mode you are interested in would help

the only failure i remember seeing was fixed inside `iris.coord` not `iris.util`

cpelley:
> the only failure i remember seeing was fixed inside iris.coord not iris.util

[...]

PR #2440, opened 2017-03-16
[WIP] Changes to make tests pass when using conda forge channel

djkirkham:
This PR contains a set of changes to make sure the tests pass when installing the environment from the conda-forge channel, with a view to ultimately using that environment for the Travis automated tests.

Due to some issues with cartopy/matplotlib detailed in https://github.com/SciTools/cartopy/issues/848, matplotlib is pinned to 1.*. (The package spec `matplotlib<2` picks up `2.0.0rc2` which is inexplicably in the conda-forge channel, so I've pinned it with `<1.9`.)

In addition to some system and example tests which are still failing, some integration tests will fail using conda-forge until https://github.com/SciTools/iris-grib/pull/75 is merged and a release has been made.


marqh:
I think that there are a range of reasons why this may well be a sensible approach for us to take.  I am keen to hear words of caution and tactical ways to protect ourselves were we to adopt this approach

djkirkham:
imagehash changes: https://github.com/SciTools/test-iris-imagehash/pull/11

marqh:
i am in favour of these changes in principal.  I haven't found the time to finish my detail review yet

i'd like to encourage comments for key stakeholders about the approach of moving CI testing to conda-forge, which is the purpose of this work

are there voices in favour?
voices with concern?
issues to address?

thoughts please

djkirkham:
The main issue as I see it is that conda forge changes so quickly that we might struggle to keep up with the changes. Additionally, it's pretty gauling if tests start failing because something on the channel changed. This is still an issue with the scitools channel of course, but at least we're in control of that.

QuLogic:
That seems like an advantage to me; having everything "work" because scitools is using a bunch of outdated packages is not really "working".

dkillick:
I agree with @djkirkham - we should not have to constantly be caught up to the latest everything. Doing so has the danger of introducing a very high workload to Iris devs that would come at the expense of the cessation of new features...

>  having everything "work" because scitools is using a bunch of outdated packages is not really "working"

Two things here:

 1. if all the tests pass, then to the best of our knowledge everything is working. The definition of working is that things are not breaking all the time; it is **not** that we are using the latest everything.
1. some packages that are pinned a few point versions behind the latest release is not "outdated".

I agree that allowing ourselves to get far behind the latest version of packages is a very bad thing. It is similarly bad, however, to be constantly trying to chase the latest everything, and I'm concerned that conda-forge just moves a little too fast and too unpredictably to be really reliable to base Iris on -- at least, not without some well thought through pinning of some packages.

ajdawson:
>This is still an issue with the scitools channel of course, but at least we're in control of that.

You'd have quite a lot of control over the conda-forge channel too, don't forget that.

I think it is an advantage to keep up with the conda-forge channel, because that is the one most users are working with, and if iris works there then it is working for a lot of people.

djkirkham:
The key issues seem to be these:

- Using the latest conda-forge packages exposes us to test failures which may be out of our control and would slow the introduction of new features
- Pinning the libraries or using the out-of-date packages on Scitools means that we're not supporting packages which are actually being used by users, and every so often there is a massive amount of work to bring things up-to-date

In an offline discussion earlier, the idea emerged to pin key libraries for the tests, but have a process in place to run the tests on the latest conda-forge libraries and notify us of any failures. That way new code could still easily be merged, and we'd be able to keep abreast of changes in conda-forge.

Thoughts?

marqh:
@djkirkham a useful summary, thank you

My present view is that I/we can quickly 'patch' a test failure by pinning a dependency; this gives me/us time to implement an improvement and to then remove the pin

> the idea emerged to pin key libraries for the tests, but have a process in place to run the tests on the latest conda-forge libraries and notify us of any failures.

I think this may be more conservative than is required.  I would feel comfortable having the tests with no pins (or as few as possible) and to add a pin if a failure mode rears its head.  Perhaps a developer rule that doing this should include raising a bug ticket and hoping to fix it in the future

Thus, I am advocating a reactive approach.
I think this could work  for us

lib/iris/coords.py
```diff
@@ -492,7 +502,14 @@ def is_full_slice(s):
                         raise IndexError('Cannot index with zero length '
                                          'slice.')
                 if bounds is not None:
-                    bounds = bounds[keys + (Ellipsis, )]
+                    # Bounds will generally have an extra dimension compared
+                    # to points, so add an Ellipsis at the end, unless there
+                    # is already on, as numpy does not support double Ellipsis.
```
pp-mo: Not quite clear:  should that be "unless there already is one" ?

djkirkham: Yes, I have corrected it.

lib/iris/tests/test_pandas.py
```diff
@@ -100,11 +101,18 @@ def test_time_360(self):
         time_coord = DimCoord([0, 100.1, 200.2, 300.3, 400.4],
                               long_name="time", units=time_unit)
         cube.add_dim_coord(time_coord, 0)
-        expected_index = [netcdftime.datetime(2000, 1, 1, 0, 0),
-                          netcdftime.datetime(2000, 4, 11, 2, 24),
-                          netcdftime.datetime(2000, 7, 21, 4, 48),
-                          netcdftime.datetime(2000, 11, 1, 7, 12),
-                          netcdftime.datetime(2001, 2, 11, 9, 36)]
+        if netCDF4.__version__ > '1.2.4':
```
dkillick: Really?!

lib/iris/tests/test_pandas.py
```diff
@@ -236,8 +244,12 @@ def test_time_360(self):
         time_coord = DimCoord([100.1, 200.2], long_name="time",
                               units=time_unit)
         cube.add_dim_coord(time_coord, 0)
-        expected_index = [netcdftime.datetime(2000, 4, 11, 2, 24),
-                          netcdftime.datetime(2000, 7, 21, 4, 48)]
+        if netCDF4.__version__ > '1.2.4':
```
dkillick: Again :scream: 

Why do we not have a test that picks this up properly rather than having two completely unrelated tests that have two different codepaths based on the version of an unrelated package?

marqh: I feel pretty bad about this approach, but it represents another step in a long line of trying to work with netcdf4python

i think that this 

there is also a particular issue with conda-forge builds of netcdf, not all versions are available as proper builds

further work is required on this, which needs to be planned
#2322

lib/iris/tests/idiff.py
```diff
@@ -236,7 +237,17 @@ def step_over_diffs(result_dir, action, display=True):
...
+                    mcompare.compare_images(expected_fname, result_fname,
+                                            tol=0)
+                except Exception as e:
+                    if isinstance(e,  ValueError) or \
+                            isinstance(e, ImageComparisonFailure):
+                        print('Could not compare {}: {}'.format(result_fname,
+                                                                e.message))
+                        continue
+                    else:
+                        raise
```
bjlittle: @djkirkham I think that this should be `raise e` to propagate the exception, surely?

lib/iris/tests/integration/plot/test_netcdftime.py
```diff
@@ -26,12 +26,15 @@
 # importing anything else
 import iris.tests as tests
 
+import numpy as np
+import netcdftime
```
bjlittle: @djkirkham Fix the sort order of the imports please

lib/iris/tests/idiff.py
djkirkham: No, `raise` on its own will keep the original stack trace. `raise e` would start a new stack trace from here. Minor detail, but it gives an extra bit of debugging information.

bjlittle: Yup, nice. But could you add a comment just to make that subtle point crystal clear to those that follow ... thanks!

lib/iris/tests/unit/coord_categorisation/test_add_categorised_coord.py
```diff
@@ -99,7 +99,6 @@ def setUp(self):
             'proleptic_gregorian': np.array(list(range(360, 367)) +
                                             list(range(1, 4))),
             'noleap': np.array(list(range(359, 366)) + list(range(1, 4))),
-            'julian': np.array(list(range(360, 367)) + list(range(1, 4))),
```
bjlittle: @marqh ( @djkirkham ) Any idea why this has been removed?

Also, the `from cf_units import CALENDARS as calendars` is now no longer used.

There is also a disconnect between the calendars in `calendars` and the calendars that this test is covering ... might want to join that up somehow ...

djkirkham: @bjlittle This is because netcdftime julian datetime objects don't give the correct day of year in the latest version; they always give a value of 1.

In fact, some of the other calendar types have the same issue if they are constructed directly:
```
>>> import netcdftime
>>> ut = netcdftime.utime('days since 1980-12-25', calendar='365_day')
>>> date = ut.num2date(0)
>>> date2 = netcdftime.DatetimeNoLeap(1980, 12, 25)
>>> print repr(date)
netcdftime._netcdftime.DatetimeNoLeap(1980, 12, 25, 0, 0, 0, 0, 2, 359)
>>> print date.dayofyr
359
>>> print date.timetuple()
(1980, 12, 25, 0, 0, 0, 2, 359, -1)
>>> print repr(date2)           
netcdftime._netcdftime.DatetimeNoLeap(1980, 12, 25, 0, 0, 0, 0, -1, 1)
>>> print date2.dayofyr
1
>>> print date2.timetuple()
(1980, 12, 25, 0, 0, 0, -1, 1, -1)
```

I think it's probably worth just commenting out the offending line in our tests and having a comment explaining why. Either that, or pin netcdf4.

bjlittle:
Where is the travis-ci output ... :confused:

lib/iris/tests/unit/coord_categorisation/test_add_categorised_coord.py
```diff
@@ -116,6 +116,11 @@ def make_cube(self, calendar):
 
     def test_calendars(self):
         for calendar in calendars:
+            # Skip the Julian calendar due to
+            # https://github.com/Unidata/netcdftime/issues/13
+            # Remove this if block once the issue is resolved.
```
cpelley: Is this issue present as a result in changing versions of netcdftime?

If so, would pinning the version of netcdftime be the preferred option until https://github.com/Unidata/netcdftime/issues/13 is resolved?

cpelley: A parallel in our own project is where we pin the numpy version to v1.11 until an iris issue is resolved:
https://github.com/SciTools/iris/issues/2425

PR #2441, opened 2017-03-16
adopt conda-forge

djkirkham:
Run Travis tests using the conda-forge channel. Will fail until https://github.com/SciTools/iris/pull/2440 is merged.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -49,12 +49,12 @@ install:
 
   # Customise the testing environment
   # ---------------------------------
-  - conda config --add channels scitools
+  - conda config --add channels conda-forge
   - if [[ "$TEST_MINIMAL" == true ]]; then
       conda install --quiet --file minimal-conda-requirements.txt;
     else
       if [[ "$TRAVIS_PYTHON_VERSION" == 3* ]]; then
-        sed -e '/ecmwf_grib/d' -e '/esmpy/d' -e '/iris_grib/d' -e 's/#.\+$//' conda-requirements.txt | xargs conda install --quiet;
+        sed -e '/ecmwf_grib/d' -e '/esmpy/d' -e '/iris-grib/d' -e 's/#.\+$//' conda-requirements.txt | xargs conda install --quiet;
       else
         conda install --quiet --file conda-requirements.txt;
       fi
--- a/.travis.yml
+++ b/.travis.yml
@@ -8,7 +8,7 @@
 language: python
 python:
   - 2.7
-  - 3.4
+  - 3.5
 sudo: false
 
 env:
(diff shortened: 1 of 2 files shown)


djkirkham:
Some tests will continue to fail even if the above is merged, until a thread-safe hdf5 is available on conda-forge (see https://github.com/conda-forge/hdf5-feedstock/pull/48)

cpelley:
> Some tests will continue to fail even if the above is merged, until a thread-safe hdf5 is available on conda-forge

In case this went unnoticed, there is a thread-safe build on conda-forge

djkirkham:
Due to some changes in Numpy 1.12, some aggregator functions in `np.ma` return a `MaskedArray` instance with no mask rather than an `ndarray`. This causes some of our tests to fail which do an instance check for a `MaskedArray`.

I've modified the checking code to use `np.ma.is_masked()`, but this causes other tests to fail where the array being checked has always been a MaskedArray instance with no mask (thus, the reference file claims it is masked when it isn't). Commit  https://github.com/SciTools/iris/pull/2441/commits/dfa75a34dd1c235f30bd7bef201589711338f238 demonstrates this - see the [Travis test failures](https://travis-ci.org/SciTools/iris/builds/218854843). Note that because some tests contain multiple checks, not all of the checks which would fail are listed in the output.

marqh:
this PR is ready to go, all the tests are passing and conda-forge is providing all of the testing matrix

The developers proximate to me have all met and discussed the strengths and weaknesses of this approach.  We have reached a consensus that the benefits outweigh the costs

cpelley:
Is there a reason why we are testing against python3.5 and not python3.6?

I assume there are missing dependencies for 3.6 at this time on conda-forge?  Either way, It would be good detail the reasoning description of this ticket.  People are trying to use iris with 3.6, see #2481.

cpelley:
In addition, the INSTALL instructions still suggest we support 3.4:
https://github.com/SciTools/iris/blob/master/INSTALL

marqh:
> In addition, the INSTALL instructions still suggest we support 3.4:
> https://github.com/SciTools/iris/blob/master/INSTALL

i have working environments of iris with python 3.4

this is only a minimum, i don't think that the minimum version of Python has changed to 3.5, just the version we are testing with on travis



> Is there a reason why we are testing against python3.5 and not python3.6?

we chose to move from 3.4 to 3.5 to ensure dependency resolution and as a minimal change in order to adopt conda forge

we will look to move testing to 3.6, but not as part of this piece of work, that will be planned as a future activity
as you have observed, there may be further wrinkles to iron out

cpelley:
Thanks @marqh

PR #2447, opened 2017-03-20
Replace dask 'compute()' usage with a common realisation call. (#2)

lbdreyer:
Replaces #2422

### DIFF
--- a/lib/iris/_lazy_data.py
+++ b/lib/iris/_lazy_data.py
@@ -73,6 +73,30 @@ def as_lazy_data(data, chunks=_MAX_CHUNK_SIZE):
     return data
 
 
+def as_concrete_data(data, nans_replacement=None, result_dtype=None):
+    """
+    Return the actual content of a lazy array, as a numpy array.
+
+    If the data is a NumPy array, return it unchanged.
+
+    If the data is lazy, return the realised result.
+
+    Where lazy data contains NaNs these are translated by filling or conversion
+    to masked data, using the :func:`convert_nans_array` function.
+    See there for usage of the 'nans_replacement' and 'result_dtype' keys.
+
+    """
+    if is_lazy_data(data):
+        # Realise dask array.
+        data = data.compute()
+        # Convert any missing data as requested.
+        data = convert_nans_array(data,
+                                  nans_replacement=nans_replacement,
+                                  result_dtype=result_dtype)
+
+    return data
+
+
 def array_masked_to_nans(array):
     """
     Convert a masked array to a NumPy `ndarray` filled with NaN values. Input
--- a/lib/iris/_lazy_data.py
+++ b/lib/iris/_lazy_data.py
@@ -73,27 +73,32 @@ def as_lazy_data(data, chunks=_MAX_CHUNK_SIZE):
     return data
 
 
-def as_concrete_data(data, nans_replacement=None, result_dtype=None):
+def as_concrete_data(data, **kwargs):
     """
     Return the actual content of a lazy array, as a numpy array.
 
-    If the data is a NumPy ndarray or masked array, return it unchanged.
+    If the input data is a NumPy `ndarray` or masked array, return it
+    unchanged.
 
-    If the data is lazy, return the realised result.
+    If the input data is lazy, return the realised result.
 
     Where lazy data contains NaNs these are translated by filling or conversion
     to masked data, using the :func:`~iris._lazy_data.convert_nans_array`
     function.
-    See there for usage of the 'nans_replacement' and 'result_dtype' keys.
+
+    Args:
+
+    * data:
+        A NumPy `ndarray` or masked array.
+
+    Kwargs are passed through to :func:`~iris._lazy_data.convert_nans_array`.
 
     """
     if is_lazy_data(data):
         # Realise dask array.
         data = data.compute()
         # Convert any missing data as requested.
-        data = convert_nans_array(data,
-                                  nans_replacement=nans_replacement,
-                                  result_dtype=result_dtype)
+        data = convert_nans_array(data, **kwargs)
 
     return data
 
--- a/lib/iris/_lazy_data.py
+++ b/lib/iris/_lazy_data.py
@@ -47,12 +47,12 @@ _MAX_CHUNK_SIZE = 8 * 1024 * 1024 * 2
 
 def as_lazy_data(data, chunks=_MAX_CHUNK_SIZE):
     """
-    Convert the input array `data` to a lazy dask array.
+    Convert the input array `data` to a dask array.
 
     Args:
 
     * data:
-        An array. This will be converted to a lazy dask array.
+        An array. This will be converted to a dask array.
 
     Kwargs:
 
@@ -63,7 +63,7 @@ def as_lazy_data(data, chunks=_MAX_CHUNK_SIZE):
         http://dask.pydata.org/en/latest/array-creation.html#chunks.
 
     Returns:
-        The input array converted to a lazy dask array.
+        The input array converted to a dask array.
 
     """
     if not is_lazy_data(data):
@@ -76,22 +76,24 @@ def as_lazy_data(data, chunks=_MAX_CHUNK_SIZE):
 def as_concrete_data(data, **kwargs):
     """
     Return the actual content of a lazy array, as a numpy array.
-
     If the input data is a NumPy `ndarray` or masked array, return it
     unchanged.
 
     If the input data is lazy, return the realised result.
 
-    Where lazy data contains NaNs these are translated by filling or conversion
+    Where lazy data contains NaNs these are translated by filling or converting
     to masked data, using the :func:`~iris._lazy_data.convert_nans_array`
     function.
 
     Args:
 
     * data:
-        A NumPy `ndarray` or masked array.
+        A dask array, NumPy `ndarray` or masked array
 
     Kwargs are passed through to :func:`~iris._lazy_data.convert_nans_array`.
+
+    Returns:
+        A NumPy `ndarray` or masked array.
 
     """
     if is_lazy_data(data):
--- a/lib/iris/_lazy_data.py
+++ b/lib/iris/_lazy_data.py
@@ -77,12 +77,13 @@ def as_concrete_data(data, nans_replacement=None, result_dtype=None):
     """
     Return the actual content of a lazy array, as a numpy array.
 
-    If the data is a NumPy array, return it unchanged.
+    If the data is a NumPy ndarray or masked array, return it unchanged.
 
     If the data is lazy, return the realised result.
 
     Where lazy data contains NaNs these are translated by filling or conversion
-    to masked data, using the :func:`convert_nans_array` function.
+    to masked data, using the :func:`~iris._lazy_data.convert_nans_array`
+    function.
     See there for usage of the 'nans_replacement' and 'result_dtype' keys.
 
     """
--- a/lib/iris/_merge.py
+++ b/lib/iris/_merge.py
@@ -33,7 +33,8 @@ import dask.array as da
 import numpy as np
 import numpy.ma as ma
 
-from iris._lazy_data import as_lazy_data, is_lazy_data, multidim_lazy_stack
+from iris._lazy_data import (as_lazy_data, is_lazy_data, multidim_lazy_stack,
+                             as_concrete_data)
 import iris.cube
 import iris.coords
 import iris.exceptions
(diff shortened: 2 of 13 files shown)


lib/iris/tests/unit/analysis/test_VARIANCE.py
```diff
@@ -23,13 +23,13 @@
...
 import iris.tests as tests
 
-from iris._lazy_data import as_lazy_data
 import numpy as np
 import numpy.ma as ma
 
 from iris.analysis import VARIANCE
 import iris.cube
 from iris.coords import DimCoord
+from iris._lazy_data import as_lazy_data, as_concrete_data
```
lbdreyer: We do need to agree on an order for importing hidden modules, atm we're inconsistent with it. 
Unfortunately, a quick google didn't unearth any standard practice
I do like this approach of putting them in alphabetical, regardless of whether or not they are private though

lib/iris/_lazy_data.py
```diff
@@ -73,6 +73,30 @@ def as_lazy_data(data, chunks=_MAX_CHUNK_SIZE):
...
+def as_concrete_data(data, nans_replacement=None, result_dtype=None):
+    """
+    Return the actual content of a lazy array, as a numpy array.
+
+    If the data is a NumPy array, return it unchanged.
+
+    If the data is lazy, return the realised result.
+
+    Where lazy data contains NaNs these are translated by filling or conversion
+    to masked data, using the :func:`convert_nans_array` function.
```
lbdreyer: >:func:`convert_nans_array`
should this be `:func:`iris._lazy_data.convert_nans_array or is okay to be relative if it's in the same module?

lib/iris/_lazy_data.py
```diff
@@ -73,6 +73,30 @@ def as_lazy_data(data, chunks=_MAX_CHUNK_SIZE):
     return data
 
 
+def as_concrete_data(data, nans_replacement=None, result_dtype=None):
+    """
+    Return the actual content of a lazy array, as a numpy array.
+
+    If the data is a NumPy array, return it unchanged.
```
bjlittle: @lbdreyer This also applies to `~numpy.ma.core.MaskedArray`.

lib/iris/_merge.py
```diff
@@ -33,7 +33,8 @@
 import numpy as np
 import numpy.ma as ma
 
-from iris._lazy_data import as_lazy_data, is_lazy_data, multidim_lazy_stack
+from iris._lazy_data import (as_lazy_data, is_lazy_data, multidim_lazy_stack,
+                             as_concrete_data)
```
bjlittle: @lbdreyer The sort order needs changes here ... make `as_concrete_data` first ...

lib/iris/tests/unit/analysis/test_VARIANCE.py
bjlittle: If we're going to go with alphabetical (which we are) then `_` sorts before all lower case letters, so IMHO private modules should go first.

lib/iris/_lazy_data.py
bjlittle: You could always make it 
```python
:func:`~iris._lazy_data.convert_nans_array`
```

lib/iris/coords.py
```diff
@@ -39,6 +39,7 @@
 
 import iris.aux_factory
 import iris.exceptions
+from iris._lazy_data import as_concrete_data
```
bjlittle: Import order ...

lib/iris/coords.py
```diff
@@ -1611,7 +1612,11 @@ def _sanitise_array(self, src, ndmin):
     def points(self):
         """Property containing the points values as a numpy array"""
         if is_lazy_data(self._points):
-            self._points = self._points.compute()
+            self._points = as_concrete_data(self._points,
+                                            nans_replacement=np.ma.masked)
+            # NOTE: we probably don't have full support for masked aux-coords.
+            # We certainly *don't* handle a _FillValue attribute (and possibly
+            # the loader will throw one away ?)
```
bjlittle: @lbdreyer We should raise a ticket to consider how we deal with masked integral data on coordinates and cell measures. At the moment we don't keep the result dtype ... this is lost in translation, which is bad.

lib/iris/cube.py
```diff
@@ -43,7 +43,8 @@
 import iris._constraints
 from iris._deprecation import warn_deprecated
 from iris._lazy_data import (array_masked_to_nans, as_lazy_data,
-                             convert_nans_array, is_lazy_data)
+                             convert_nans_array, is_lazy_data,
+                             as_concrete_data)
```
bjlittle: Import sort order ...

lib/iris/fileformats/pp.py
```diff
@@ -43,7 +43,8 @@
 import iris.fileformats.rules
 import iris.fileformats.pp_rules
 import iris.coord_systems
-from iris._lazy_data import array_masked_to_nans, as_lazy_data, is_lazy_data
+from iris._lazy_data import (array_masked_to_nans, as_lazy_data, is_lazy_data,
+                             as_concrete_data)
```
bjlittle: Import sort order ...

lib/iris/tests/unit/analysis/test_STD_DEV.py
```diff
@@ -23,7 +23,7 @@
 # importing anything else.
 import iris.tests as tests
 
-from iris._lazy_data import as_lazy_data
+from iris._lazy_data import as_lazy_data, as_concrete_data
```
bjlittle: Import sort order

lib/iris/tests/unit/lazy_data/test_multidim_lazy_stack.py
```diff
@@ -26,7 +26,7 @@
 import numpy as np
 import dask.array as da
 
-from iris._lazy_data import as_lazy_data, multidim_lazy_stack
+from iris._lazy_data import as_lazy_data, multidim_lazy_stack, as_concrete_data
```
bjlittle: Import sort order

lbdreyer:
@bjlittle Hopefully you are happy with the changes? 
(I am happy for you to squash and merge)

lib/iris/_lazy_data.py
```diff
@@ -73,6 +73,31 @@ def as_lazy_data(data, chunks=_MAX_CHUNK_SIZE):
...
+    Return the actual content of a lazy array, as a numpy array.
+
+    If the data is a NumPy ndarray or masked array, return it unchanged.
+
+    If the data is lazy, return the realised result.
+
+    Where lazy data contains NaNs these are translated by filling or conversion
+    to masked data, using the :func:`~iris._lazy_data.convert_nans_array`
+    function.
+    See there for usage of the 'nans_replacement' and 'result_dtype' keys.
```
dkillick: Typo: "there" --> "their".

Alternatively you could change the signature to 

```python
def as_concrete_data(data, **kwargs)
```

as the `kwargs` are not used at all by this function and are just passed straight to `convert_nans_array`. At that point you could replace this description of the `kwargs` with something like

"Kwargs are passed straight to :func:`~iris._lazy_data.convert_nans_array`."

It is a more typical way of doing such things in Python but also less readable...

lib/iris/_merge.py
```diff
@@ -1217,10 +1218,11 @@ def merge(self, unique=True):
...
                 # All inputs were concrete, so turn the result back into a
                 # normal array.
-                merged_data = merged_data.compute()
-                # Unmask the array only if it is filled.
+                merged_data = as_concrete_data(merged_data,
+                                               nans_replacement=ma.masked)
+                # Unmask the array if it has no masked points.
                 if (ma.isMaskedArray(merged_data) and
-                        ma.count_masked(merged_data) == 0):
+                        not ma.is_masked(merged_data)):
```
dkillick: Nice spot :wink:

lib/iris/tests/unit/lazy_data/test_as_concrete_data.py
```diff
@@ -0,0 +1,90 @@
...
+        lazy_array = as_lazy_data(data)
+        sentinel = mock.sentinel.data
+        with mock.patch('iris._lazy_data.convert_nans_array') as conv_nans:
+            conv_nans.return_value = sentinel
+            result = as_concrete_data(lazy_array, nans_replacement=7,
+                                      result_dtype=np.int16)
+        self.assertEqual(sentinel, result)
+
+        # Check call to convert_nans_array
+        conv_nans.assert_called_once()
```
dkillick: I reckon you can do [`assert_called_once_with`](http://www.voidspace.org.uk/python/mock/mock.html#mock.Mock.assert_called_once_with), which may be useful here.

lib/iris/tests/unit/lazy_data/test_as_concrete_data.py
```diff
@@ -0,0 +1,90 @@
...
+            result = as_concrete_data(lazy_array)
+        self.assertEqual(sentinel, result)
+
+        # Check call to convert_nans_array
+        conv_nans.assert_called_once()
+        args, kwargs = conv_nans.call_args
+        arg, = args
+        self.assertFalse(is_lazy_data(arg))
+        self.assertArrayEqual(arg, data)
+        self.assertEqual(kwargs, {'result_dtype': None,
```
dkillick: `assert_called_once_with`?

lib/iris/coords.py
bjlittle: @lbdreyer Did you create an issue to cover this?

lib/iris/tests/unit/lazy_data/test_as_concrete_data.py
```diff
@@ -0,0 +1,90 @@
...
+from __future__ import (absolute_import, division, print_function)
+from six.moves import (filter, input, map, range, zip)  # noqa
+
+# Import iris.tests first so that some things can be initialised before
+# importing anything else.
+import iris.tests as tests
+
+import numpy as np
+import numpy.ma as ma
+import dask.array as da
```
bjlittle: @lbdreyer Sorry, but could you fix this import order ...

lib/iris/tests/unit/lazy_data/test_as_concrete_data.py
lbdreyer: Doesn't work when one of the args is a numpy array

lib/iris/tests/unit/lazy_data/test_as_concrete_data.py
lbdreyer: Ah man! I didn't see this one!

lib/iris/_lazy_data.py
lbdreyer: I think "See there for usage" was meant as in "See over there for usage". 

I like your idea of `**kwargs`

lib/iris/coords.py
lbdreyer: No, but I'll do that now

lbdreyer: @bjlittle I've just raised https://github.com/SciTools/iris/issues/2449
 not sure if that's quite what you were after. Feel free to edit the issue

lib/iris/tests/unit/lazy_data/test_as_concrete_data.py
dkillick: It probably would if you also mocked the array.

lib/iris/_lazy_data.py
dkillick: > I think "See there for usage" was meant as in "See over there for usage".

Ah yes, on a re-reading you're quite right!

lib/iris/tests/unit/lazy_data/test_as_concrete_data.py
dkillick: Also, in this case, neither of them are NumPy arrays...

lib/iris/tests/unit/lazy_data/test_as_concrete_data.py
bjlittle: @dkillick I don't think that `assert_called_once_with` will cope with comparing numpy arrays, so checking the call count and unpacking the call args seems okay ... but you could give @dkillick's suggestion a try.

@lbdreyer Although, you could combine this test and the above `test_lazy_data` test into one and also use sentinels for the `nans_replacement` and `result_dtype` ... that would make it sufficiently generic.

bjlittle:
:+1: Hoo-rah! Awesome, thanks @lbdreyer

lbdreyer:
Thanks @bjlittle !!

Issue #2398, opened 2017-02-21
Get all Iris tests working again (5 man days' effort)


corinnebosley:
(refactor or replace implementation-specific tests)

lbdreyer:
Relevant Issue: #2443

lbdreyer:
Remove all `skip_biggus` skippers:
ldreyer@eld347 $ grep -ilr --color "@tests.skip_biggus"
- [ ] tests/test_pickling.py
- [ ] tests/test_cdm.py
- [ ] tests/test_cube_to_pp.py
- [ ] tests/test_analysis_calculus.py
- [ ] tests/test_pp_cf.py
- [ ] tests/test_pp_module.py
- [ ] tests/integration/test_aggregated_cube.py
- [ ] tests/integration/test_regridding.py
- [ ] tests/integration/test_pp.py
- [ ] tests/integration/experimental/test_regrid_ProjectedUnstructured.py
- [ ] tests/integration/test_netcdf.py
- [ ] tests/test_abf.py
- [ ] tests/test_netcdf.py
- [ ] tests/unit/concatenate/test_concatenate.py
- [ ] tests/unit/fileformats/rules/test__make_cube.py
- [ ] tests/unit/fileformats/grib/message/test_GribMessage.py
- [ ] tests/unit/fileformats/grib/test_load_cubes.py
- [ ] tests/unit/fileformats/pyke_rules/compiled_krb/fc_rules_cf_fc/test_build_auxiliary_coordinate.py
- [ ] tests/unit/fileformats/netcdf/test_save.py
- [ ] tests/unit/fileformats/netcdf/test__load_cube.py
- [ ] tests/unit/fileformats/test_rules.py
- [ ] tests/unit/fileformats/pp/test_PPField.py
- [ ] tests/unit/experimental/regrid/test__CurvilinearRegridder.py
- [ ] tests/unit/cube/test_Cube__operators.py
- [ ] tests/unit/analysis/maths/test_add.py
- [ ] tests/unit/analysis/maths/test_multiply.py
- [ ] tests/unit/analysis/maths/test_subtract.py

pp-mo:
In context of #2507, I'm grabbing these ones (so, ticked them above) :
`lib/iris/tests/test_cube_to_pp.py:236`
`lib/iris/tests/test_cube_to_pp.py:270`
`lib/iris/tests/test_pp_cf.py:84`
 ==> "pp/bmdi problem"

[...]

Issue #2448, opened 2017-03-20
Remove `core-dtype` from cml (1 man days' effort)


lbdreyer:
core-dtype was introduced whilst we were messing around with the dtype/fill_value of a cube and it's array. Once this is sorted it should be remove from the cml
https://github.com/SciTools/iris/blob/dask/lib/iris/cube.py#L2939

(see [relevant comment](https://github.com/SciTools/iris/pull/2411/files#r106912894))

dkillick:
As of #2543, we may have a different `core_dtype` to cube `dtype`. This bit of functionality makes retaining `core_dtype` in the CML beneficial, so we think we want to retain this for the time being.

PR #2450, opened 2017-03-21
Remove some `@tests.skip_biggus` that are now unnecessary

lbdreyer:
This removes some of the skippers that were put in place for the Iris-dask integration since they are no longer needed.

The cml has change for some due to core-dtype, dtype and fill value becoming cube attributes

Relevant Issue: #2398

There are still some outstanding tests that have skippers in them but they are a bit more complicated so I will deal with them in later PRs


bjlittle:
@lbdreyer Awesome, thanks :+1:

dkillick:
> The cml has change for some due to core-dtype

Is `core-dtype` something we might want to drop, a la `core_data`?

bjlittle:
@dkillick Yeah, the `core-dtype` was always intended to be an interim change, just to help with the changes in the testing ... so I'd be keen for us to purge it.

We could also then look into sorting the xml elements into an order that we prefer

lbdreyer:
> Is core-dtype something we might want to drop, a la core_data?

See #2448 so we do intend to remove core-dtype at some point

PR #2452, opened 2017-03-22
Lazy data in PP save pairs

pp-mo:
Preserve lazy data in PP save-pairs generation, so you can work with a list of save cubes and/or fields without excessive memory requirements (as intended for saving).

The existing PP save code does not preserve lazy data when making a PPField from a 2d cube slice.
There was no point bothering with that, as the PP save rules fetch the data, basically just to get the data fill-value (if there is one).

This fixes that, so that the save rules, and save field generation, both preserve the lazy aspect.
Changes here:
 * preserve lazy data in generating PP save pairs
 * add a `cube.fill_value` to generated saved cubes (only)
    * this approach anticipates a probable future API change for cube to have a standard `cube.fill_value` property -- as currently under discussion for the dask migration work.


pp-mo:
@bjlittle can you comment whether this now looks sensible, in light of ongoing Dask development ?

cpelley:
Happy to review this when it's ready.  We have external collaborators (CMIP6) who this effects and have limited availability of hardware to work around this problem.

Cheers

cpelley:
> this approach anticipates a probable future API change for cube to have a standard cube.fill_value property -- as currently under discussion for the dask migration work.

Description a little out of date?? #2433

marqh:
> Description a little out of date?? #2433

hi @cpelley 

those changes are on a feature branch, no dask code is yet on master, to my knowledge.

so, i don't think this is out of date

lib/iris/fileformats/pp.py
```diff
@@ -2404,6 +2416,11 @@ def save_pairs_from_cube(cube, field_coords=None, target=None):
 
     # Save each named or latlon slice2D in the cube
     for slice2D in cube.slices(field_coords):
+        # Attach an extra cube "fill_value" property, allowing the save rules
+        # to deduce MDI more easily without realising the data.
+        # NOTE: it is done this way because this property may exist in future.
+        slice2D.fill_value = _data_fill_value(slice2D)
```
djkirkham: Perhaps this could be avoided by making the code in `_data_fill_value()` into a set of save rules to determine `pp.bmdi`.

pp-mo: This is a purely temporary measure, based on the idea that the forthcoming dask change will definitely be introducing the generic "cube.fill_value" property.
We will remove this hack when the dask branch is merged back.
Obviously, the method actually used here is :nauseated_face: but it is only here so you guys can check that it fixes the current memory burn problem.

lib/iris/fileformats/pp.py
```diff
@@ -2426,8 +2443,12 @@ def save_pairs_from_cube(cube, field_coords=None, target=None):
         # From UM doc F3: "Set to -99 if LBEGIN not known"
         pp_field.lbuser[1] = -99
 
-        # Set the data
-        pp_field.data = slice2D.data
+        # Set the data, keeping it lazy where possible.
+        if slice2D.has_lazy_data():
+            slice_core_data = slice2D.lazy_data()
+        else:
+            slice_core_data = slice2D.data
```
pp-mo: It's best like this because creating a simple wrapper around real data is substantially more costly in dask than it was with biggus :  The earliest dask branch retained existing cube property code to get cube.shape and cube.dtype from cube.lazy_data(), and then the tests took nearly twice as long!
Having said that, this code should be removed when we transition to dask anyway, as mentioned [above](https://github.com/SciTools/iris/pull/2452#discussion_r113162934)

cpelley:
Thanks @pp-mo, ping @hdyson

hdyson:
Thanks all, it's very much appreciated.  I'll keep you informed once we've done a test with the data that revealed this issue (this may take a few days, it was an external collaborator that found it).

hdyson:
Just confirming that this does work for the use case that our collaborators found - good stuff.  We're seeing ~5x improvement in memory usage, and this is more than sufficient for our needs.

Thanks again!

Issue #2438, opened 2017-03-15
Fix dtype.kind usage


pp-mo:
Where we have code like `array.dtype.kind == 'i'` we should be allowing `'u'` also.

pp-mo:
See : [this comment](https://github.com/SciTools/iris/pull/2433#pullrequestreview-27045807)

This probably needs fixing in a number of places, not just in dask changes

Possibly `isinstance(xxx, np.integer)` is better ?

PR #2453, opened 2017-03-22
Fix for dtype.kind

bjlittle:
Fix to support `dtype.kind` to also check for `uint` and `bool` cases (as well as `int`).

### DIFF
--- a/lib/iris/cube.py
+++ b/lib/iris/cube.py
@@ -1640,8 +1640,8 @@ bound=(1994-12-01 00:00:00, 1998-12-01 00:00:00)
                     emsg = 'Cube does not have lazy data, cannot set dtype.'
                     raise ValueError(emsg)
                 if dtype.kind not in 'biu':
-                    emsg = ('Can only cast lazy data to integral dtype, '
-                            'got {!r}.')
+                    emsg = ('Can only cast lazy data to integral or bool '
+                            'dtype, got {!r}.')
                     raise ValueError(emsg.format(dtype))
                 self._fill_value = None
                 self._dtype = dtype
--- a/lib/iris/cube.py
+++ b/lib/iris/cube.py
@@ -1639,7 +1639,7 @@ bound=(1994-12-01 00:00:00, 1998-12-01 00:00:00)
                 if not self.has_lazy_data():
                     emsg = 'Cube does not have lazy data, cannot set dtype.'
                     raise ValueError(emsg)
-                if dtype.kind != 'i':
+                if dtype.kind not in 'biu':
                     emsg = ('Can only cast lazy data to integral dtype, '
                             'got {!r}.')
                     raise ValueError(emsg.format(dtype))
@@ -1658,7 +1658,7 @@ bound=(1994-12-01 00:00:00, 1998-12-01 00:00:00)
             # Convert the given value to the dtype of the cube.
             fill_value = np.asarray([fill_value])[0]
             target_dtype = self.dtype
-            if fill_value.dtype.kind == 'f' and target_dtype.kind == 'i':
+            if fill_value.dtype.kind == 'f' and target_dtype.kind in 'biu':
                 # Perform rounding when converting floats to ints.
                 fill_value = np.rint(fill_value)
             try:
--- a/lib/iris/cube.py
+++ b/lib/iris/cube.py
@@ -1640,7 +1640,7 @@ bound=(1994-12-01 00:00:00, 1998-12-01 00:00:00)
                     emsg = 'Cube does not have lazy data, cannot set dtype.'
                     raise ValueError(emsg)
                 if dtype.kind not in 'biu':
-                    emsg = ('Can only cast lazy data to integral or bool '
+                    emsg = ('Can only cast lazy data to an integer or boolean '
                             'dtype, got {!r}.')
                     raise ValueError(emsg.format(dtype))
                 self._fill_value = None
--- a/lib/iris/tests/unit/cube/test_Cube.py
+++ b/lib/iris/tests/unit/cube/test_Cube.py
@@ -1252,7 +1252,7 @@ class Test_fill_value(tests.IrisTest):
         self.assertEqual(cube.fill_value, None)
 
     def test_fill_value_bool(self):
-        fill_value = np.array([1], dtype=np.dtype('bool'))[0]
+        fill_value = np.bool_(True)
         for dtype in self.dtypes:
             data = np.array([0], dtype=dtype)
             cube = Cube(data, fill_value=fill_value)
@@ -1261,7 +1261,7 @@ class Test_fill_value(tests.IrisTest):
             self.assertEqual(cube.fill_value.dtype, dtype)
 
     def test_fill_value_int(self):
-        fill_value = np.array([123], dtype=np.dtype('int'))[0]
+        fill_value = np.int_(123)
         for dtype in self.dtypes:
             data = np.array([0], dtype=dtype)
             cube = Cube(data, fill_value=fill_value)
@@ -1270,7 +1270,7 @@ class Test_fill_value(tests.IrisTest):
             self.assertEqual(cube.fill_value.dtype, dtype)
 
     def test_fill_value_float(self):
-        fill_value = np.array([123.4], dtype=np.dtype('float'))[0]
+        fill_value = np.float_(123.4)
         for dtype in self.dtypes:
             data = np.array([0], dtype=dtype)
             cube = Cube(data, fill_value=fill_value)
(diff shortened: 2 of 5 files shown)


bjlittle:
Closes #2438

bjlittle:
Ping @lbdreyer @dkillick

lbdreyer:
How about this one:
https://github.com/bjlittle/iris/blob/6bd77b099ecf3a4ae7db2621c7b5c300be1703bb/lib/iris/fileformats/pp.py#L1044

lib/iris/cube.py
```diff
@@ -1639,7 +1639,7 @@ def dtype(self, dtype):
                 if not self.has_lazy_data():
                     emsg = 'Cube does not have lazy data, cannot set dtype.'
                     raise ValueError(emsg)
-                if dtype.kind != 'i':
+                if dtype.kind not in 'biu':
                     emsg = ('Can only cast lazy data to integral dtype, '
```
lbdreyer: This error message is now not technically true since you also check for boolean.

lib/iris/cube.py
```diff
@@ -1658,7 +1658,7 @@ def fill_value(self, fill_value):
             # Convert the given value to the dtype of the cube.
             fill_value = np.asarray([fill_value])[0]
             target_dtype = self.dtype
-            if fill_value.dtype.kind == 'f' and target_dtype.kind == 'i':
+            if fill_value.dtype.kind == 'f' and target_dtype.kind in 'biu':
                 # Perform rounding when converting floats to ints.
                 fill_value = np.rint(fill_value)
```
lbdreyer: Again, what if the target_dtype is boolean?

lib/iris/cube.py
bjlittle: Yup, you're totally right. `numpy` in it's wisdom doesn't actually have `bool_` subclassed under `integer`, see [here](http://numpy.readthedocs.io/en/latest/reference/arrays.scalars.html#) ... I'll update the `emsg`. Good spot.

lib/iris/cube.py
bjlittle: It rounds to the nearest integer and treats `0` as `False` and non-zero as `True` ... which is fair enough.

lib/iris/tests/unit/cube/test_Cube.py
```diff
@@ -1223,21 +1223,197 @@ def test__lazy(self):
...
+
+    def test_lazy_data_masked__mask_set(self):
+        for dtype in self.dtypes:
+            data = ma.array([0, 1], dtype=dtype)
+            data[0] = ma.masked
+            cube = Cube(as_lazy_data(data))
+            if dtype == np.dtype('float'):
+                self.assertEqual(cube.dtype, dtype)
+            else:
+                # For dask, integral masked data is converted to float
```
lbdreyer: "integral and bool masked data is converted to float"

lib/iris/tests/unit/cube/test_Cube.py
```diff
@@ -1223,21 +1223,197 @@ def test__lazy(self):
...
+                # to support the replacement of masks with NaNs.
+                self.assertEqual(cube.dtype, np.dtype('float'))
```
lbdreyer: This bit of logic may make it clearer what the expected behaviour is, but it is a bit unnecessary. Surely
```
if dtype == np.dtype('float'):
    self.assertEqual(cube.dtype, dtype)
else:
    self.assertEqual(cube.dtype, np.dtype('float'))
```
is the same as
```
self.assertEqual(cube.dtype, np.dtype('float'))
```

lbdreyer:
@bjlittle I think this PR is pretty much good to go, could you just address my last two comments?

dkillick:
Sorry! Turns out I was sitting on a review that I never submitted :unamused:

lib/iris/tests/unit/cube/test_Cube.py
bjlittle: Yup :blush:

lib/iris/tests/unit/cube/test_Cube.py
```diff
@@ -1223,21 +1223,197 @@ def test__lazy(self):
...
+
+    def test_cube_data_setter_with_none(self):
+        fill_value = 123
+        cube = Cube(np.array(0), fill_value=fill_value)
+        self.assertEqual(cube.fill_value, fill_value)
+        cube.data = np.array(1)
+        self.assertEqual(cube.fill_value, None)
+
+    def test_fill_value_bool(self):
+        fill_value = np.array([1], dtype=np.dtype('bool'))[0]
```
bjlittle: It would have to be `fill_value = np.array(True)[()]` which I thought was a little bit terse and magical ... and this ain't Hogwarts, so I erred on the side of verbosity :wink:

lib/iris/tests/unit/cube/test_Cube.py
```diff
@@ -1223,21 +1223,197 @@ def test__lazy(self):
...
+            self.assertEqual(cube.fill_value, expected)
+            self.assertEqual(cube.fill_value.dtype, dtype)
+
+    def test_fill_value_float(self):
+        fill_value = np.array([123.4], dtype=np.dtype('float'))[0]
+        for dtype in self.dtypes:
+            data = np.array([0], dtype=dtype)
+            cube = Cube(data, fill_value=fill_value)
+            if dtype.kind in 'biu':
+                fill_value = np.rint(fill_value)
```
bjlittle: Nope, `np.rint` is a thing ...

lib/iris/tests/unit/cube/test_Cube.py
```diff
@@ -1223,21 +1223,197 @@ def test__lazy(self):
...
+        self.assertEqual(cube.fill_value, None)
+
+    def test_fill_value_bool(self):
+        fill_value = np.array([1], dtype=np.dtype('bool'))[0]
+        for dtype in self.dtypes:
+            data = np.array([0], dtype=dtype)
+            cube = Cube(data, fill_value=fill_value)
+            [expected] = np.array([fill_value], dtype=dtype)
+            self.assertEqual(cube.fill_value, expected)
+            self.assertEqual(cube.fill_value.dtype, dtype)
```
bjlittle: The `dtype` of the `fill_value` is relative to the `dtype` of the data ... so yeah.

lbdreyer:
I'm happy with the changes. Once @dkillick is happy this can be merged in.

lib/iris/tests/unit/cube/test_Cube.py
```diff
@@ -1337,15 +1508,15 @@ def test_lazydata_maskedints(self):
...
+        self.assertTrue(ma.is_masked(data))
         self.assertMaskedArrayEqual(data, masked_data_original)
 
     def test_lazydata_floating_dtype_change(self):
         # Check that re-assigning dtype won't allow a floating type.
         cube = self._sample_cube(dtype=np.float32, lazy=True,
                                  cube_fill_value=23.7)
         self.assertArrayAllClose(cube.fill_value, 23.7)
-        msg = "Can only cast lazy data to integral dtype"
+        msg = "Can only cast lazy data to integral or bool dtype"
```
dkillick: I think "integer" might now be clearer here. You might also want to change "bool" to "boolean".

lib/iris/cube.py
```diff
@@ -1639,9 +1639,9 @@ def dtype(self, dtype):
                 if not self.has_lazy_data():
                     emsg = 'Cube does not have lazy data, cannot set dtype.'
                     raise ValueError(emsg)
-                if dtype.kind != 'i':
-                    emsg = ('Can only cast lazy data to integral dtype, '
-                            'got {!r}.')
+                if dtype.kind not in 'biu':
+                    emsg = ('Can only cast lazy data to integral or bool '
```
dkillick: Integral --> Integer; bool --> boolean?

lib/iris/tests/unit/cube/test_Cube.py
```diff
@@ -1223,21 +1223,192 @@ def test__lazy(self):
...
+        for dtype in self.dtypes:
+            data = np.array([0, 1], dtype=dtype)
+            emsg = 'Cube does not have lazy data, cannot set dtype'
+            with self.assertRaisesRegexp(ValueError, emsg):
+                Cube(data, dtype=np.complex)
+
+    def test_lazy_data_with_dtype_non_integral(self):
+        for dtype in self.dtypes:
+            data = np.array([0, 1], dtype=dtype)
+            emsg = 'Can only cast lazy data to integral or bool dtype'
```
dkillick: You'll need to change this error message too if you agree with my other suggestions.

lbdreyer:
Thanks @bjlittle !

PR #2451, opened 2017-03-22
Remove the last bits of biggus

dkillick:
Remove the final biggus imports (and usages) from the codebase*.

In essence, this means:

 * delete a couple of spurious (and unused!) biggus import statements,
 * sort out some test modules, and
 * tidy up the lazy data API where the tests very pleasingly showed up a couple of gaps in it.

Fixes #2443.
&nbsp;

\* Well, not _quite_: I explicitly **haven't** dealt with grib here - we'll [do](https://github.com/SciTools/iris/issues/2342) that [later](https://github.com/SciTools/iris/issues/2306).

PR #2454, opened 2017-03-22
WIP: changes to Cube data api

dkillick:
Specifically, try and use the data setter for _every_ change to `cube.data`. This includes:

* instantiating a cube,
* realising lazy data, and
* user changing a cube's data.

This set of changes started life in #2451 but got backed out of there when it got big and gnarly.

### DIFF
--- a/lib/iris/_lazy_data.py
+++ b/lib/iris/_lazy_data.py
@@ -97,7 +97,10 @@ def as_concrete_data(data, **kwargs):
 
     """
     if is_lazy_data(data):
-        # Realise dask array.
+        # Realise dask array, ensuring the data result is always a NumPy array.
+        # In some cases dask may return a scalar numpy.int/numpy.float object
+        # rather than a numpy.ndarray object.
+        # Recorded in https://github.com/dask/dask/issues/2111.
         data = data.compute()
         # Convert any missing data as requested.
         data = convert_nans_array(data, **kwargs)
--- a/lib/iris/cube.py
+++ b/lib/iris/cube.py
@@ -33,7 +33,6 @@ import warnings
 from xml.dom.minidom import Document
 import zlib
 
-import biggus
 import dask.array as da
 import numpy as np
 import numpy.ma as ma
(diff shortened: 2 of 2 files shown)


lib/iris/_lazy_data.py
```diff
@@ -97,7 +97,10 @@ def as_concrete_data(data, **kwargs):
 
     """
     if is_lazy_data(data):
-        # Realise dask array.
+        # Realise dask array, ensuring the data result is always a NumPy array.
+        # In some cases dask may return a scalar numpy.int/numpy.float object
+        # rather than a numpy.ndarray object.
+        # Recorded in https://github.com/dask/dask/issues/2111.
```
dkillick: This comment should probably be moved now that we're not ensuring `data` is a NumPy array here.

lib/iris/cube.py
```diff
@@ -664,9 +663,9 @@ def __init__(self, data, standard_name=None, long_name=None,
             This object defines the shape of the cube and the phenomenon
             value in each cell.
 
-            It can be a biggus array, a numpy array, a numpy array
-            subclass (such as :class:`numpy.ma.MaskedArray`), or an
-            *array_like* as described in :func:`numpy.asarray`.
+            ``data`` can be a dask array, a NumPy array, a NumPy array
+            subclass (such as :class:`numpy.ma.MaskedArray`), or
+            array_like (as described in :func:`numpy.asarray`).
```
dkillick: Whoops - looks like this change got missed when I backed this out of #2451...

lib/iris/cube.py
```diff
@@ -1754,8 +1750,16 @@ def data(self):
 
     @data.setter
     def data(self, value):
-        if not (hasattr(value, 'shape') and hasattr(value, 'dtype')):
```
dkillick: This turns out to not be the complete check of what's required here. For example, a `numpy.int32` object has a `shape` attribute but is **not** an `ndarray`...

marqh:
hi @dkillick 

please may you confirm if this PR is still a desired change?

if so, please rebase as it now conflicts
if not, please close

thank you

dkillick:
Aha, no - this all got done by @bjlittle in the end. Thanks @marqh

PR #2456, opened 2017-03-22
Cube.__init__ uses data setter.

bjlittle:
Tidy the `Cube.__init__` so that it is a good citizen and explicitly uses the `Cube.data` setter.


lib/iris/cube.py
```diff
@@ -1638,18 +1633,20 @@ def dtype(self):
 
     @dtype.setter
     def dtype(self, dtype):
```
dkillick: This setter method all seems very unstable... hopefully we will be able to reduce the scope of this (or remove it entirely!) with upcoming changes.

bjlittle: Yup, that's the plan

bjlittle:
@dkillick Thanks!

PR #2458, opened 2017-03-23
Only use truediv in cubes

dkillick:
Iris only uses `truediv` to divide a cube when the `operator` function `div` is not available (effectively, this means only in py3k). This PR changes that behaviour so that `truediv` is **always** used. 

The main driver for this was the observation that dividing an integer cube by a scalar does not produce the same result as dividing the cube's integer data by a scalar:

```python
cube = iris.cube.Cube(np.array([1,2,3,4,5]))
cube.data
array([1, 2, 3, 4, 5])
cube.data.dtype
dtype('int64')
operator.truediv(cube, 5).data
array([0, 0, 0, 0, 1])
operator.truediv(cube.data, 5)
array([ 0.2,  0.4,  0.6,  0.8,  1. ])
```

This seems to be unwanted behaviour.


dkillick:
Hurrah, tests passing! Ping @lbdreyer @bjlittle

lib/iris/analysis/maths.py
```diff
@@ -341,10 +341,16 @@ def divide(cube, other, dim=None, in_place=False):
...
-    except AttributeError:
-        op = operator.itruediv if in_place else operator.truediv
+    if in_place:
+        op = operator.itruediv
+        # Input `other` may not have a `dtype` attribute.
+        other_dtype_kind = other.dtype.kind if hasattr(other, 'dtype') else 'f'
+        if 'i' in [cube.dtype.kind, other_dtype_kind]:
+            # NumPy ufunc `true_divide` cannot coerce output to int.
+            aemsg = 'Cannot perform inplace division of integer data.'
+            raise ArithmeticError(aemsg)
```
bjlittle: @dkillick I think that you only need to check that the `cube.dtype.kind` is always `f` for `itruediv` ... right?

lib/iris/cube.py
```diff
@@ -3161,15 +3161,35 @@ def __hash__(self):
...
+        return iris.analysis.maths.subtract(self, other,
+                                            ignore=True, in_place=True)
+
     __mul__ = iris.analysis.maths.multiply
     __rmul__ = iris.analysis.maths.multiply
+
+    def __imul__(self, other):
+        return iris.analysis.maths.multiply(self, other, in_place=True)
+
     __div__ = iris.analysis.maths.divide
```
dkillick: Hang on, this might be a problem... I've asserted here that we should _only_ use `truediv` when dividing cubes, yet as it stands if I do `my_cube / 5`, it is _this_ codepath that will be followed and not `truediv`. This can't be right.

dkillick: Don't panic! We do want this behaviour as all division overrides here call `maths.divide`, which only uses the `truediv` / `itruediv` operator. So actually this should be kept so that it funnels all incoming division operations on cubes into running `maths.divide`, and all is hunky-dory.

lib/iris/tests/test_basic_maths.py
```diff
@@ -654,72 +654,92 @@ def test_log10(self):
...
     def test_operator_array(self):
         for test_op in self.ops:
             result1 = test_op(self.cube1, self.data2)
             result2 = test_op(self.data1, self.data2)
-
-            np.testing.assert_array_equal(result1.data, result2)
+            self.assertArrayAlmostEqual(result1.data, result2)
+            result3 = test_op(self.cube3, self.data4)
+            result4 = test_op(self.data3, self.data4)
+            self.assertArrayAlmostEqual(result3.data, result4)
```
bjlittle: @dkillick This test covers the `(float, float)` and `(int, int)` operator cases ... would it be cheeky of me to push for `(float, int)` and `(int, float)` combos coverage?

That might give us some more confidence that the right thing is going on under the hood ...

lib/iris/analysis/maths.py
dkillick: Correct! I'll fix.

lib/iris/tests/test_basic_maths.py
```diff
@@ -654,72 +654,92 @@ def test_log10(self):
...
+
+    def test_cube_itruediv__int(self):
+        test_op = self.iops[-1]  # `itruediv`
```
bjlittle: @dkillick This is a bit naughty ... what if the `iops` order changes in the `setUp` ... you could always be explicit and just use `operator.itruediv` ...

lib/iris/tests/test_basic_maths.py
```diff
@@ -654,72 +654,92 @@ def test_log10(self):
...
+        with self.assertRaisesRegexp(ArithmeticError,
+                                     'Cannot perform inplace division'):
+            test_op(self.cube3, self.cube4)
+
+    def test_cube_itruediv__scalar(self):
+        test_op = self.iops[-1]  # `itruediv`
```
bjlittle: Again ...

lib/iris/tests/test_basic_maths.py
dkillick: It would certainly increase the length of the test class :wink:

lib/iris/tests/test_basic_maths.py
dkillick: > you could always be explicit and just use `operator.itruediv`

Yeah I should probably just do that...

lib/iris/analysis/maths.py
```diff
@@ -311,11 +319,35 @@ def multiply(cube, other, dim=None, in_place=False):
...
 
+def _inplace_common(cube, other, math_op):
+    """
+    Check whether an inplace math operation can take place between `cube` and
+    `other`. It cannot if `cube` has integer data and `other` has float data
+    as the operation will always produce float data that cannot be 'safely'
+    cast back to the integer data of `cube`.
+
+    """
+    if cube.dtype.kind == 'i':
```
bjlittle: @dkillick We probably want to make this at least `if cube.dtype.kind in 'iu'` to include `uint`, and at a stretch `biu` to include `bool`

lib/iris/analysis/maths.py
```diff
@@ -311,11 +319,35 @@ def multiply(cube, other, dim=None, in_place=False):
...
+        if hasattr(other, 'dtype'):
+            other_kind = other.dtype.kind
+        else:
+            other_kind = np.asanyarray(other).dtype.kind
+        if other_kind != 'i':
```
bjlittle: Same here ...

lib/iris/analysis/maths.py
```diff
@@ -342,13 +374,12 @@ def divide(cube, other, dim=None, in_place=False):
     other_unit = getattr(other, 'units', '1')
     new_unit = cube.units / other_unit
     if in_place:
-        op = operator.itruediv
-        # Input `other` may not have a `dtype` attribute.
-        other_dtype_kind = other.dtype.kind if hasattr(other, 'dtype') else 'f'
-        if 'i' in [cube.dtype.kind, other_dtype_kind]:
-            # NumPy ufunc `true_divide` cannot coerce output to int.
-            aemsg = 'Cannot perform inplace division of integer data.'
+        if cube.dtype.kind == 'i':
```
bjlittle: Same extension for the kind here as above ...

lib/iris/analysis/maths.py
```diff
@@ -342,13 +374,12 @@ def divide(cube, other, dim=None, in_place=False):
...
+            # Cannot coerce float result from inplace division back to int.
+            aemsg = ('Cannot perform inplace division of cube {!r} '
+                     'with integer data.')
```
bjlittle: Would be kinda neat to use your new `_inplace_common` here, but if that's too much of a cludge, no worries ... just a thought.

lib/iris/tests/unit/analysis/maths/test__inplace_common.py
```diff
@@ -0,0 +1,96 @@
...
+
+        self.float_cube = Cube(self.float_data)
+        self.int_cube = Cube(self.int_data)
+
+        self.op = 'addition'
+        self.emsg = 'Cannot perform inplace {}'.format(self.op)
+
+    def test_float_cubes(self):
+        result = _inplace_common(self.float_cube, self.float_cube, self.op)
+        self.assertIsNone(result)
```
bjlittle: @dkillick Slightly confused, `_inplace_common` doesn't return anything so `self.asserIsNone(result)` isn't really a useful test (same for all below) ...

bjlittle: Surely you need introspect the cube to make sure that the inplace operation has changed something ...

bjlittle:
@dkillick I've added some additional minor comments, so service those and we're done. Nice one!

lib/iris/tests/unit/analysis/maths/test__inplace_common.py
bjlittle: ... or add a banner comment regarding your testing philosophy so that all the `self.assertIsNone(result)` tests make sense to the reader, given the context.

dkillick: @bjlittle interestingly `_inplace_common` doesn't change anything anywhere, it's just a pass-through function that will either:

* do nothing (most cases) -- and so have an implicit return value of `None`, or
* fail with an error iff both cube and other have a dtype kind of `'i'`.

Further, I decided I wanted to explicitly check the known good cases, because if they become no longer good for some reason we want these tests to start failing.

But a banner comment on this slightly esoteric testing approach would probably be a help to our future selves!

lib/iris/analysis/maths.py
dkillick: I'll investigate... but it will probably need a `kwarg` to `_inplace_common`, which may just move the bad stuff around. The reason I didn't use it for divide is because divide has a quite different processing requirement and there wasn't a whole lot of code duplication...

lib/iris/tests/unit/analysis/maths/test__inplace_common.py
dkillick: I should probably rename the function as well! That might make it clearer what's going on.

lib/iris/analysis/maths.py
dkillick: I'll go as far as unsigned ints but I don't believe there's a use-case for bool checks here too. In the spirit of maximising work not done I'm not going to touch bools. If someone finds there's a problem for having not done so then we can think about it further then...

dkillick:
@bjlittle I've made most of the changes you suggested :+1:

lib/iris/tests/test_basic_maths.py
```diff
@@ -653,73 +653,189 @@ def test_log10(self):
...
     def setUp(self):
-        self.data1 = ma.MaskedArray([[9, 9, 9], [8, 8, 8,]],
-                                    mask=[[0, 1, 0], [0, 0, 1]])
-        self.data2 = ma.MaskedArray([[3, 3, 3], [2, 2, 2,]],
-                                    mask=[[0, 1, 0], [0, 1, 1]])
-
-        self.cube1 = iris.cube.Cube(self.data1)
-        self.cube2 = iris.cube.Cube(self.data2)
+        self.data_1f = np.array([[9, 9, 9], [8, 8, 8]],
+                                dtype=np.float64)
```
dkillick: We actually weren't using the masked data at all, it's just what I inherited with these tests, and up until now it has been fine. But, weirdly, masked arrays screw up inplace operations on uint data, so I've removed the masking. The two tests in this module that actually were about masked data I've moved to a new class below.

lib/iris/tests/test_basic_maths.py
```diff
@@ -653,73 +653,189 @@ def test_log10(self):
...
-    # This is to ensure compatibility for both Python2 (which uses div and
-    # idiv) and Python3 (which uses truediv and itruediv).
-    try:
-        ops = ops + (operator.div, )
-        iops = iops + (operator.idiv, )
-    except AttributeError:
-        ops = ops + (operator.truediv, )
-        iops = iops + (operator.itruediv, )
-
+class TestMathOperations(tests.IrisTest):
```
dkillick: No longer about masked arrays!

bjlittle:
@dkillick Lookin' good! When travis is happy, then so am I :+1:

bjlittle:
@dkillick travis isn't happy ...

bjlittle:
Nice :+1:

PR #2459, opened 2017-03-24
Remove skip_biggus from unit/fileformats/test_rules.py

lbdreyer:
I have updated this PR so it only deals with the removal of the biggus skipper from:
unit/fileformats/test_rules.py

### DIFF
--- a/lib/iris/tests/unit/fileformats/test_rules.py
+++ b/lib/iris/tests/unit/fileformats/test_rules.py
@@ -35,12 +35,8 @@ from iris.fileformats.rules import (ConcreteReferenceTarget,
                                     Reference, ReferenceTarget, load_cubes,
                                     scalar_cell_method)
 from iris.coords import CellMethod
+from iris.tests import mock
 import iris.tests.stock as stock
-
-
-class Mock(object):
-    def __repr__(self):
-        return '<Mock {!r}>'.format(self.__dict__)
 
 
 class TestConcreteReferenceTarget(tests.IrisTest):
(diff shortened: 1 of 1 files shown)


lib/iris/tests/test_cube_to_pp.py
```diff
@@ -266,7 +266,7 @@ def lbproc_from_pp(self, filename):
         field = next(pp_file)
         return field.lbproc
 
-    # see related comment #236
+    # see related comment line number #236
```
lbdreyer: This comment confused me so I thought a quick edit might prevent someone else also getting confused.
Of course the intention is to remove this `skip_biggus` soon

lib/iris/tests/unit/fileformats/test_rules.py
```diff
@@ -105,22 +101,25 @@ def transform(cube):
...
         # Test the creation process for a factory definition which only
         # uses simple dict arguments.
 
         # The fake PPField which will be supplied to our converter.
-        field = Mock()
-        field.data = None
+        field = mock.Mock()
+        field.core_data = mock.Mock()
+        # Add a compute attribute so that the data will be treated as lazy.
+        field.core_data.compute = None
```
dkillick: Is this a bit like a statement of `lazy == hasattr(core_data, 'compute')`? Does the test function correctly without this attribute being set?

dkillick: OK, I don't understand why the mock object needs to pretend to be lazy (the runs just fine without this line). I also don't see this line as a deal-breaker, though...

Issue #2403, opened 2017-02-22
dask scheduler num_workers


bjlittle:
See [here](https://github.com/SciTools/iris/pull/2392/files#r102430258) for context.

We need to consider how we easily control the number of threads, processes or workers that the dask scheduler uses - particularly for users targeting a shared resource such as a server or cluster.

Ping @marqh

AlexHilson:
Fwiw if a Distributed scheduler is defined then that will be used by default, so the current implementation will work with either the threaded / distributed scheduler depending on what the user has done. 

Defining a way to pass arguments into the scheduler is definitely useful. I feel like I would want the default settings to use 100% of available resources and let dask / my cpu worry about the consequences, but maybe that's naive.

AlexHilson:
I briefly discussed this with @marqh, and he mentioned some of the multi-user systems where Iris is deployed and used.

I agree that we should test how dask behaves on these systems before going 'live', but I still believe that if you're running a multi-user system it's your responsibility to ensure that your users cpu access is managed. Making our code slower by default seems wrong to me, but I appreciate that we may need to be pragmatic...

dkillick:
My feeling about this is that Iris should **run single thread/process by default**, with users needing to opt in if they wish to use the multiprocess goodness that dask offers. This needn't be the default long-term, but I think it is the most appropriate introductory approach.

I have a couple of reasons to back up this assertion:

 * many Iris users may not be used to parallel processing and may be put out if an innocent processing command given to Iris uses up all the available compute resource and physical memory on their machine, causing it to crash.
 * Python multiprocessing seems to disregard the amount of resource offered by slurm, so if we open up unrestricted multiprocess capability (both in terms of being on by default in Iris and in terms of not heeding slurm limits) then we may enter a realm where Iris is regularly taking out scalable compute resource through numerous, concurrent, resource unrestricted uses of Iris. This would look bad.

Of course once the two concerns above are resolved then we can reconsider the default multiprocess behaviour in Iris.

dkillick:
In #2457 @bjlittle suggested the pattern of having `iris.options`. I wonder if we could make use of that here, both for setting the number of workers and setting the scheduler. Consider:

```python
with iris.options.parallel(num_workers=6, scheduler='multiprocessing'):
    iris.load('my_dataset.nc')
```

Or

```python
iris.options.parallel(num_workers=6, scheduler='multiprocessing')
iris.load('my_dataset.nc')
```

[...]

PR #2462, opened 2017-03-28
Iris options (dask parallel processing)

dkillick:
Introduces a new module; `iris.options`. Within this module we can provide classes that control certain elements of Iris' behaviour within a limited scope, such as the lifetime of the session, or within a context manager.

Specifically, this PR introduces the `Parallel` class, used to control dask parallel processing options for Iris. The `Parallel` class allows for control of:

* the **dask scheduler** used to process dask graphs when `compute` is called (defaults to `threaded`), and
* the **number of workers** used to process graphs in parallel (defaults to 1, only available for the `threaded` and `multiprocessing` scheduler options).

Note: no tests yet (I'll do them next); I was keen to get eyes on the functional code I've proposed here.

The pattern proposed here also has an obvious application to the proposal in #2457. 

Addresses #2403.


dkillick:
Ping @bjlittle @lbdreyer -- I'd appreciate your thoughts on this proposal.

lib/iris/__init__.py
```diff
@@ -113,6 +113,7 @@ def callback(cube, field, filename):
 from iris._deprecation import IrisDeprecation, warn_deprecated
 import iris.fileformats
 import iris.io
+import iris.options
```
lbdreyer: `iris.options` is not being used in this module, so by importing it here, you are catering to someone being able to do:
```
import iris
iris.options.parallel(scheduler='multiprocessing', num_workers=6)
```
I don't think this is a good thing. I would prefer users have to import iris.options directly

dkillick: I disagree; I think this _is_ a good thing. We want to make `options` as easy to use as possible. It makes complete sense that the controlling options for Iris come as part of the primary import.

And it's not just me that thinks so -- dask does this, as does `iris.FUTURE`, by dint of being defined in the Iris `__init__` already. And adding `options` to the `__init__` actually was a design consideration I made, which I only dropped because of the potential size of the `options` module.

lib/iris/options.py
```diff
@@ -0,0 +1,209 @@
...
+                information on setting up distributed schedulers, see
+                https://distributed.readthedocs.io/en/latest/index.html.
+
+        Example usages:
+
+        * Specify that we want to load a cube with dask parallel processing
+        using multiprocessing with six worker processes::
+
+        >>> iris.options.parallel(scheduler='multiprocessing', num_workers=6)
+        >>> iris.load('my_dataset.nc')
```
lbdreyer: To be consistent with the other code examples, you should remove the python interpreter prompts i.e. the `>>>` and the `...`

dkillick: :+1: And indent by four spaces? I didn't look very hard into how this is done elsewhere...

dkillick: > To be consistent with the other code examples

Actually, can you provide an example of this please? All the examples I just found in `cube.py` use the chevrons...

lbdreyer: I may be wrong. I didn't realise it would be so difficult to find examples

dkillick: There's no need to get snippy. All I was after was the example you found, as they were different to the examples I found...

pp-mo:
Discussed betwen @dkillick @pp-mo in offline conversation...
Probably this can all go in the existing `iris.config` instead of making a new module.
As at : https://github.com/SciTools/iris/pull/2467#issuecomment-290750566

dkillick:
Mocky tests added with some much appreciated help from @bjlittle! Think this is ready to go now, so long as Travis is happy...

lib/iris/__init__.py
bjlittle: @dkillick It's a moot point at the moment since `iris.options` doesn't actually exist :stuck_out_tongue_winking_eye:

lib/iris/_lazy_data.py
```diff
@@ -95,6 +95,10 @@ def as_concrete_data(data, **kwargs):
     Returns:
         A NumPy `ndarray` or masked array.
 
+    .. note::
+        Specific dask options for computation are controlled by
+        :class:`iris.options.Parallel`.
```
bjlittle: @dkillick This is stale, should be `iris.config.Parallel` or should it be `iris.config.parallel` ?

lib/iris/config.py
```diff
@@ -154,3 +168,267 @@ def get_dir_option(section, option, default=None):
...
+                defined when setting up the distributed scheduler. For more
+                information on setting up distributed schedulers, see
+                https://distributed.readthedocs.io/en/latest/index.html.
+
+        Example usages:
+
+        * Specify that we want to load a cube with dask parallel processing
+        using multiprocessing with six worker processes::
+
+            iris.options.parallel(scheduler='multiprocessing', num_workers=6)
```
bjlittle: @dkillick Nope ...

lib/iris/config.py
```diff
@@ -154,3 +168,267 @@ def get_dir_option(section, option, default=None):
...
+            iris.load('my_dataset.nc')
+
+        * Specify, with a context manager, that we want to load a cube with
+        dask parallel processing using four worker threads::
+
+            with iris.options.parallel(scheduler='threaded', num_workers=4):
```
bjlittle: @dkillick `iris.config.parallel`

lib/iris/config.py
```diff
@@ -154,3 +168,267 @@ def get_dir_option(section, option, default=None):
...
+                iris.load('my_dataset.nc')
+
+        * Run dask parallel processing using a distributed scheduler that has
+        been set up at the IP address and port at ``192.168.0.219:8786``::
+
+            iris.options.parallel(scheduler='192.168.0.219:8786')
```
bjlittle: @dkillick Again ...

lib/iris/tests/unit/config/test_Parallel.py
```diff
@@ -0,0 +1,287 @@
...
+from iris.config import Parallel
+from iris.tests import mock
+
+
+class Test_operation(tests.IrisTest):
+    # Check that the options are passed through to 'real' code.
+    # NOTE: tests that call the option class directly and as a contextmgr.
+
+    def test_bad_name__contextmgr(self):
+        # Check we can't do `with iris.options.parallel.context('foo'='bar')`.
```
bjlittle: @dkillick

lib/iris/config.py
```diff
@@ -66,11 +66,25 @@
...
 import warnings
 
+import dask
+import dask.multiprocessing
+
+DISTRIBUTED_AVAILABLE = True
+try:
+    import distributed
+except ImportError:
+    DISTRIBUTED_AVAILABLE = False
```
bjlittle: @dkillick Would you consider this slightly neater pattern (IMHO):
```python
try:
    import distributed
except ImportError:
    distributed = None
```
Then later in the code, rather than checking for 
```python
if DISTRIBUTED_AVAILABLE:
```
it's simply
```python
if distributed is not None:
```
See [here](https://github.com/SciTools/iris/pull/2462/files#diff-36cf2602a7c220f674c997618f233af7R342) ...

lib/iris/config.py
```diff
@@ -66,11 +66,25 @@
...
 from __future__ import (absolute_import, division, print_function)
 from six.moves import (filter, input, map, range, zip)  # noqa
-
+import six
 from six.moves import configparser
+
+import contextlib
+from multiprocessing import cpu_count
+from multiprocessing.pool import ThreadPool
+import re
```
bjlittle: @dkillick Import order, move below following line ...

lib/iris/config.py
```diff
@@ -154,3 +168,267 @@ def get_dir_option(section, option, default=None):
...
+
+        """
+        scheduler = self.get('scheduler')
+        get = self.get('dask_scheduler')
+        pool = None
+
+        if scheduler in ['threaded', 'multiprocessing']:
+            num_workers = self.get('num_workers')
+            pool = ThreadPool(num_workers)
+        if scheduler == 'distributed':
```
bjlittle: @dkillick `elif` ?

lib/iris/config.py
```diff
@@ -154,3 +168,267 @@ def get_dir_option(section, option, default=None):
...
+                }
+
+    def set__scheduler(self, value):
+        return value
+
+    def set_scheduler(self, value):
+        default = self._defaults_dict['scheduler']['default']
+        if value is None:
+            value = default
+        elif re.match(r'^(\d{1,3}\.){3}\d{1,3}:\d{1,5}$', value):
```
bjlittle: @dkillick You could compile this `re` ...

lib/iris/config.py
```diff
@@ -154,3 +168,267 @@ def get_dir_option(section, option, default=None):
...
+                    'for more details.')
+            warnings.warn(wmsg.format(scheduler))
+            value = None
+        else:
+            if value is None:
+                value = default
+            if value >= cpu_count():
+                # Limit maximum CPUs used to 1 fewer than all available CPUs.
+                wmsg = ('Requested more CPUs ({}) than total available ({}). '
+                        'Limiting number of used CPUs to {}.')
```
corinnebosley: @dkillick Is this going to be a bit confusing?  What if somebody requests that maximum number of CPUs available, and then is only allowed to use one less than the max available, but the warning message states that the total available is the same number that they requested.  

For example, let's say my CPU count is 8, and as a parallelisation beginner I want to see how fast my script is when I run it on all 8 CPUs, so I request 8.  This isn't allowed by set_num_workers, so I get a warning message which says:
'Requested more CPUs (8) than total available (8).  Limiting number of used CPUs to 7'

I would find this message confusing, and it would make me angry.  Maybe it just needs a little clarification that you can't use all the CPUs available because it's silly.

lib/iris/tests/unit/config/test_Parallel.py
```diff
@@ -0,0 +1,287 @@
...
+
+    def test_async(self):
+        scheduler = 'async'
+        Parallel(scheduler=scheduler)
+        self.assertEqual(self.patch_Client.call_count, 0)
+        self.assertEqual(self.patch_ThreadPool.call_count, 0)
+
+        pool = self.pool
+        get = dask.async.get_sync
+        self.patch_set_options.assert_called_once_with(pool=None, get=get)
```
corinnebosley: @dkillick I'm confused, if pool=None in the assert on L.137, why do you need 'pool = self.pool' on L.135?

djkirkham:
I'm not sure I like the idea of overriding `__setattr__`. It seems like it would be be better to use `__setitem__` instead and have it set values in an internal dictionary. There's not much difference in syntax (`option['key'] = value` instead of `option.key = value`) but it avoids the complications of overriding `__setattr__`, such as not being able to instance attributes.

lib/iris/config.py
```diff
@@ -154,3 +168,267 @@ def get_dir_option(section, option, default=None):
...
+                                          'async',
+                                          'distributed']},
+                'num_workers': {'default': 1, 'options': None},
+                'dask_scheduler': {'default': None, 'options': None},
+                }
+
+    def set__scheduler(self, value):
+        return value
+
+    def set_scheduler(self, value):
```
djkirkham: These `set_<key>` methods are confusingly named. They don't actually set anything. A better name would be `convert_<key>`, IMO.

lib/iris/config.py
```diff
@@ -154,3 +168,267 @@ def get_dir_option(section, option, default=None):
...
+        if value is None:
+            value = default
+        elif re.match(r'^(\d{1,3}\.){3}\d{1,3}:\d{1,5}$', value):
+            if DISTRIBUTED_AVAILABLE:
+                value = 'distributed'
+            else:
+                # Distributed not available.
+                wmsg = 'Cannot import distributed. Defaulting to {}.'
+                warnings.warn(wmsg.format(default))
+                self.set_scheduler(default)
```
djkirkham: This doesn't actually do anything, assuming the default scheduler is always the same. Did you mean to return this?

lib/iris/config.py
```diff
@@ -154,3 +168,267 @@ def get_dir_option(section, option, default=None):
...
+    def set__scheduler(self, value):
+        return value
+
+    def set_scheduler(self, value):
+        default = self._defaults_dict['scheduler']['default']
+        if value is None:
+            value = default
+        elif re.match(r'^(\d{1,3}\.){3}\d{1,3}:\d{1,5}$', value):
+            if DISTRIBUTED_AVAILABLE:
+                value = 'distributed'
```
djkirkham: If someone tries to set the scheduler to a new address, aren't you just discarding it here?

lib/iris/config.py
```diff
@@ -154,3 +168,267 @@ def get_dir_option(section, option, default=None):
...
+        options = ['{}={}'.format(k, v)
+                   for k, v in six.iteritems(self.__dict__)
+                   if not k.startswith('_')]
+        joined = ', '.join(options)
+        return msg.format(joined)
+
+    def __setattr__(self, name, value):
+        if value is None:
+            value = self._defaults_dict[name]['default']
+        attr_setter = getattr(self, 'set_{}'.format(name))
```
djkirkham: This is pretty ugly, IMO. I like the way Matplotlib handles this. In addition to storing default values, they also store a validator/converter function for each parameter (See [matplotlib/rcsetup.py](https://github.com/matplotlib/matplotlib/blob/4ba8903b83e41e4f83cc459ba0c3202dfd58c5d3/lib/matplotlib/rcsetup.py#L964))

lib/iris/config.py
dkillick: I thought I'd fixed all of them :unamused:

lib/iris/tests/unit/config/test_Parallel.py
dkillick: Because we want to make sure it isn't used in this case! Nice spot though :+1:

lib/iris/_lazy_data.py
dkillick: @bjlittle it should be `:class:iris.config.Parallel` as `Parallel` is (a) actually a class that can be linked to by Sphinx, and (b) has a docstring. I'll change this, and all of the similar below...

lib/iris/config.py
dkillick: They define values that are set for keys in the instance's underlying `__dict__`. In this sense they behave very similarly to the `@property.setter` in a more traditional class. I think renaming them to `convert_xxx` would be _more_ confusing as I think "converting" is further from what they're doing than "setting" is.

lib/iris/config.py
dkillick: My intention is that the scheduler setter is re-run on the default value in case it throws up any problems with setting the scheduler to the default in this case. Either way, the default value will very likely be returned in the next call of `set_scheduler`.

lib/iris/config.py
dkillick: The scheduler value will be set to `distributed`, true, which is used as a check on [L407](https://github.com/dkillick/iris/blob/60aebe8e96e84068cb757680bdf6f4166e712e17/lib/iris/config.py#L407), for example. The actual IP address is maintained in `_scheduler`, which is set on [L283](https://github.com/dkillick/iris/blob/60aebe8e96e84068cb757680bdf6f4166e712e17/lib/iris/config.py#L283) and then used on [L408](https://github.com/dkillick/iris/blob/60aebe8e96e84068cb757680bdf6f4166e712e17/lib/iris/config.py#L408).

lib/iris/config.py
dkillick: @djkirkham nice idea, I'll implement that.

lib/iris/config.py
djkirkham: Right, so it's only set in `__init__`. If I do `iris.options.parallel.scheduler = <address>`, that won't change what's saved in `_scheduler`.

lib/iris/config.py
djkirkham: > they behave very similarly to the @property.setter in a more traditional class

But the _don't_ behave like a setter because they don't _set_ anything!

lib/iris/config.py
djkirkham: But what is returned from this function is `value`, and that isn't changed if this `else` section is reached. So ultimately what will be set in the attribute is the original value passed in, despite what the warning says.

djkirkham: Actually, I see what happens. When `Option.__setattr__` is called, it doesn't recognise the address as a legitimate value, and sets the attribute to the default. But I still don't understand the point of this `self.set_scheduler(default)` call

dkillick:
Closed on account of excessive complexity and replaced by #2511.

PR #2466, opened 2017-03-29
Slice order test 2

corinnebosley:
Simpler approach to this bugfix.  Same problem as https://github.com/SciTools/iris/pull/2464/files, but this time I didn't refactor.  Just used the bits that I needed instead.

### DIFF
--- a/lib/iris/cube.py
+++ b/lib/iris/cube.py
@@ -1,4 +1,4 @@
-# (C) British Crown Copyright 2010 - 2016, Met Office
+# (C) British Crown Copyright 2010 - 2017, Met Office
 #
 # This file is part of Iris.
 #
@@ -3940,7 +3940,15 @@ class _SliceIterator(collections.Iterator):
 
         if self._ordered:
             if any(self._mod_requested_dims != list(range(len(cube.shape)))):
-                cube.transpose(self._mod_requested_dims)
+                sliced_dims = []
+                for i in self._requested_dims:
+                    # Convert dim numbers back to dims to match up to indices of
+                    # sliced cube
+                    dim_ref = self._cube.dim_coords[i].name()
+                    # Now convert back to dim numbers to use in transpose
+                    new_dim, = cube.coord_dims(dim_ref)
+                    sliced_dims.append(new_dim)
+                cube.transpose(sliced_dims)
 
         return cube
 
--- a/lib/iris/cube.py
+++ b/lib/iris/cube.py
@@ -3942,8 +3942,8 @@ class _SliceIterator(collections.Iterator):
             if any(self._mod_requested_dims != list(range(len(cube.shape)))):
                 sliced_dims = []
                 for i in self._requested_dims:
-                    # Convert dim numbers back to dims to match up to indices of
-                    # sliced cube
+                    # Convert dim numbers back to dims to match up to indices
+                    # of sliced cube
                     dim_ref = self._cube.dim_coords[i].name()
                     # Now convert back to dim numbers to use in transpose
                     new_dim, = cube.coord_dims(dim_ref)
--- a/lib/iris/cube.py
+++ b/lib/iris/cube.py
@@ -3940,14 +3940,9 @@ class _SliceIterator(collections.Iterator):
 
         if self._ordered:
             if any(self._mod_requested_dims != list(range(len(cube.shape)))):
-                sliced_dims = []
-                for i in self._requested_dims:
-                    # Convert dim numbers back to dims to match up to indices
-                    # of sliced cube
-                    dim_ref = self._cube.dim_coords[i].name()
-                    # Now convert back to dim numbers to use in transpose
-                    new_dim, = cube.coord_dims(dim_ref)
-                    sliced_dims.append(new_dim)
+                n = len(self._mod_requested_dims)
+                sliced_dims = np.empty(n, dtype=int)
+                sliced_dims[self._mod_requested_dims] = np.arange(n)
                 cube.transpose(sliced_dims)
 
         return cube
--- a/lib/iris/tests/unit/cube/test_Cube.py
+++ b/lib/iris/tests/unit/cube/test_Cube.py
@@ -23,6 +23,8 @@ from six.moves import (filter, input, map, range, zip)  # noqa
 # importing anything else.
 import iris.tests as tests
 
+from itertools import permutations
+
 import biggus
 import numpy as np
 import numpy.ma as ma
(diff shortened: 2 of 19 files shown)


corinnebosley:
@djkirkham @marqh @pp-mo I'm happy with this, I need you three to be happy with it too.

lib/iris/tests/unit/cube/test_Cube.py
```diff
@@ -409,6 +409,160 @@ def test_kwargs(self):
...
+    This class tests the slices capability of iris.cube.Cube
+    '''
+    def setUp(self):
+        '''
+        setup a 4D iris cube, each dimension is length 1.
+        The dimensions are;
+            dim1: height
+            dim2: time
+            dim3: longitude
+            dim4: latitude
```
corinnebosley: @djkirkham I am tempted to not use stock cubes for this setup as I don't really see the need, and if I did I would have to change all of the tests that have been written here.  I am also not sure that it is necessary to change round the order of dimensions; although they are not proper here, if I changed them around then I would also have to change all of the tests.  And how much does it matter, really?

lib/iris/cube.py
```diff
@@ -3940,7 +3940,15 @@ def __next__(self):
 
         if self._ordered:
             if any(self._mod_requested_dims != list(range(len(cube.shape)))):
-                cube.transpose(self._mod_requested_dims)
+                sliced_dims = []
+                for i in self._requested_dims:
+                    # Convert dim numbers back to dims to match up to indices
+                    # of sliced cube
+                    dim_ref = self._cube.dim_coords[i].name()
```
djkirkham: Unfortunately this causes slicing to fail for anonymous dimensions:
```
>>> import iris.tests.stock as its
>>> cube = its.realistic_4d()
>>> cube.remove_coord('grid_longitude')
>>> print cube
air_potential_temperature / (K)     (time: 6; model_level_number: 70; grid_latitude: 100; -- : 100)
     Dimension coordinates:
          time                           x                      -                  -         -
          model_level_number             -                      x                  -         -
          grid_latitude                  -                      -                  x         -
     Auxiliary coordinates:
          level_height                   -                      x                  -         -
          sigma                          -                      x                  -         -
          surface_altitude               -                      -                  x         x
     Derived coordinates:
          altitude                       -                      x                  x         x
     Scalar coordinates:
          forecast_period: 0.0 hours
     Attributes:
          source: Iris test case
>>> list(cube.slices([3,1,2]))
  File "/home/h02/dkirkham/development/iris/lib/iris/cube.py", line 3947, in __next__
    dim_ref = self._cube.dim_coords[i].name()
IndexError: tuple index out of range
```

djkirkham: I had a play around and I think the following code gets you the dimensions in the order you need them:
```
n = len(self._mod_requested_dims)
sliced_dims = np.empty(n), dtype=int)
sliced_dims[self._mod_requested_dims] = np.arange(n)
```

corinnebosley: Dammit, I was hoping that this wouldn't happen using this approach.  I'll have to see if I can do anything about it, but I'm a little worried that I might have to discuss this in-depth with somebody to be able to work out the best strategy.

corinnebosley: Although I am a little confused about why this passes all the tests but fails for that thing that you have done above.

corinnebosley: Nice.  Updating...

corinnebosley:
@djkirkham Looks like job's a good'un.

corinnebosley:
@marqh @pp-mo Please can I have your opinion on this?  I want it merged.

lib/iris/tests/unit/cube/test_Cube.py
```diff
@@ -409,6 +409,155 @@ def test_kwargs(self):
         self.assertMaskedArrayEqual(expected_result, res_cube.data)
 
 
+class TestSlices(tests.IrisTest):
```
pp-mo: By the naming rules, I think this should be called `Test_slices`.
Or in fact, as it only really tests one aspect of behaviour, you may as well call it `Test_slices__dim_order` or somesuch.

If you really wanted, you could be more comprehensive -- I that see Test_slices_over is pretty involved(!).
Why would you want to do that, though ?? I'm certainly not asking for it.

lib/iris/tests/unit/cube/test_Cube.py
```diff
@@ -409,6 +409,155 @@ def test_kwargs(self):
...
+            cube: iris cube
+                iris cube with the specified axis holding the data 8
+        '''
+        cube = iris.cube.Cube(np.array([[[8.]]]))
+        cube.add_dim_coord(iris.coords.DimCoord([0], dim1name), [0])
+        cube.add_dim_coord(iris.coords.DimCoord([0], dim2name), [1])
+        cube.add_dim_coord(iris.coords.DimCoord([0], dim3name), [2])
+        return cube
+
+    def make_the_test(self, dim1, dim2, dim3, dim_to_remove):
```
pp-mo: As, whenever you use this, you then always say `self.assertEqual(actual, expected)`, I would put that assertion inside it, and rename it e.g. `check_dims_order`.  Each testcase then just calls it with a list of coords.

lib/iris/tests/unit/cube/test_Cube.py
```diff
@@ -409,6 +409,155 @@ def test_kwargs(self):
...
+                the cube that results if slicing the original cube
+            expected_cube: 3D cube
+                a cube set up with the axis corresponding to the dims
+        '''
+        sliced_cube = next(self.cube.slices([dim1, dim2, dim3]))
+        sliced_cube.remove_coord(dim_to_remove)
+        expected_cube = self.expected_cube_setup(dim1, dim2, dim3)
+        return sliced_cube, expected_cube
+
+    def test_height_long_lat(self):
```
pp-mo: I've had a hard time trying to work out whether there is some special logic in the cases you test here (there are 7, out of a possible 4 * 3 * 2 = 24, I think ??).

This may sound over-clever, but if you are testing a whole bunch of cases then why not just have code to generate all of them??
If any ordering of the 4 coords will make a possible testcase, then you can just use `itertools.permutations()`.

If you do that, you probably have to do all the assertions within one testcase routine, so you should ensure that if one of them fails the resulting message makes it quite clear which one it was.

corinnebosley: Well that sounds like a pain in the arse, but I'll give it a go...

djkirkham:
@pp-mo @corinnebosley 

I've created a PR onto this branch with the permutations change: https://github.com/corinnebosley/iris/pull/7

corinnebosley:
@djkirkham Thanks Dan, merged your changes and all tests are passing.

pp-mo:
Checked that the test fails with current master.
:+1: now good for me.
@djkirkham are you ok with this now ?

djkirkham:
@pp-mo Yeah, looks good to me

corinnebosley:
@djkirkham @pp-mo @lbdreyer @marqh In the interests of not self-merging, and bearing in mind that this has been reviewed, edited and approved, can somebody please just push the button and get this merged?

marqh:
yes

PR #2467, opened 2017-03-29
Iris options: NetCDF

dkillick:
Adds the `options` module to master (cf. #2462). In particular, this PR adds an Option to control certain aspects of NetCDF file handling, specifically what to do with a custom conventions attribute on saving a cube to a NetCDF file.

In this sense this PR is proposing a very similar change to the excellent suggestion made by @morwenna-01 in #2457, but does so in the manner [suggested there](https://github.com/SciTools/iris/pull/2457#issuecomment-288971875) by @bjlittle.


lib/iris/options.py
```diff
@@ -0,0 +1,107 @@
...
+        * Specify, with a context manager, that we want a cube written to
+          NetCDF to define its own CF Conventions version::
+
+            with iris.options.netcdf.context(conventions_override=True):
+                iris.save('my_cube', 'my_dataset.nc')
+
+        """
+        self.__dict__['conventions_override'] = conventions_override
+
+    def __str__(self):
```
bjlittle: Use `__repr__` instead, otherwise you get this behaviour ...
```python
>>> iris.options.netcdf
<iris.options.NetCDF object at 0x7fdc0b9f60d0>
>>> print iris.options.netcdf
NetCDF options: conventions_override=False.
>>> 
```

lib/iris/options.py
```diff
@@ -0,0 +1,107 @@
...
+        msg = 'NetCDF options: conventions_override={}.'
+        return msg.format(self.conventions_override)
```
bjlittle: Since there's potential for more option attributes to come, this pattern is handy ...
```python
msg = 'NetCDF options: conventions_override={self.conventions_override!r}.'
return msg.format(self)
```

lib/iris/options.py
```diff
@@ -0,0 +1,107 @@
...
+"""
+from __future__ import (absolute_import, division, print_function)
+from six.moves import (filter, input, map, range, zip)  # noqa
+import six
+
+import contextlib
+import warnings
+
+
+class NetCDF(object):
```
bjlittle: @dkillick Is this thread safe? We'll want that for dask usage ...

lib/iris/options.py
```diff
@@ -0,0 +1,107 @@
...
+        """
+        self.__dict__['conventions_override'] = conventions_override
+
+    def __str__(self):
+        msg = 'NetCDF options: conventions_override={}.'
+        return msg.format(self.conventions_override)
+
+    def __setattr__(self, name, value):
+        if name not in self.__dict__:
+            msg = "'Future' object has no attribute {!r}".format(name)
```
bjlittle: Cut'n paste carnage ... `Future`?

lib/iris/options.py
```diff
@@ -0,0 +1,107 @@
...
+            good_value = self._defaults_dict[name]
+            wmsg = ('Attempting to set bad value {!r} for attribute {!r}. '
+                    'Defaulting to {!r}.')
+            warnings.warn(wmsg.format(value, name, good_value))
+            value = good_value
+        self.__dict__[name] = value
+
+    @property
+    def _defaults_dict(self):
+        return {'conventions_override': False}
```
bjlittle: I see the intent, but there's a slight disconnect here, in that there is a [default value](https://github.com/SciTools/iris/pull/2467/files#diff-6bf007ada426c5f36c4fc5e94665e8d1R32) in the formal argument list of the `__init__`.

These could contradict each other ... it would  be nice to specify this in one place somehow.

lib/iris/options.py
```diff
@@ -0,0 +1,107 @@
...
+
+    def __str__(self):
+        msg = 'NetCDF options: conventions_override={}.'
+        return msg.format(self.conventions_override)
+
+    def __setattr__(self, name, value):
+        if name not in self.__dict__:
+            msg = "'Future' object has no attribute {!r}".format(name)
+            raise AttributeError(msg)
+        if value not in [True, False]:
```
bjlittle: Are we restricting all option attributes to be only `bool`?

lib/iris/options.py
dkillick: Whoops! Good spot :wink:

lib/iris/options.py
dkillick: Yes, in a "maximise work not done" sense :wink: -- in this case I only need to care about bools, so I have. I do have a plan for making this more broad, but that will appear in #2462, where I need it.

lib/iris/options.py
dkillick: True! I'll poke it and see what I can do.

lib/iris/options.py
dkillick: I didn't have much success with `__repr__` either :unamused:

lib/iris/options.py
dkillick: Will we though? It is interesting that dask's `set_options` is **not** thread-safe. I left this like this for precisely this reason -- so that we'd get some commenting on it.

In fact I'd go so far as to argue that we _don't_ want options thread-safe (!) -- my reason being that if I set an option I'd like it to be available in **all** threads, not just the calling thread. Of course, this may indicate some mixed-up thinking about how threads work, so I'm happy to stand corrected about this one.

bjlittle: I guess we'll be driven by the behaviour. Let's see how it plays!

pp-mo: I'd say if you provide it as a _context manager_, then it's not a one-off switch to set-and-forget.
In that case, thread-safe is definitely preferable IMHO.

pp-mo:
I fell this doesn't follow the spirit of what you mentioned
> the manner [suggested there](https://github.com/SciTools/iris/pull/2457#issuecomment-288971875) by @bjlittle.

I think I'd find it much tidier to have 'iris.netcdf.options' instead of 'iris.options.netcdf', because it scales nicely.
However, I think it also makes very good sense to write the class code once and have that in 'iris.options'.  It should be similar to the iris.Future usage, though I know we have outstanding concerns about the thread-safe usage.
If we want a top-level 'iris.options' mainly so we can print a summary, you could make it a dictionary which the other modules populate on install.

So e.g. `netcdf.py` could have something like ...
```
from iris.options import Options, options as iris_options

options = Options(conventions_override=False)
iris_options['netcdf'] = options

```
You could in fact have

dkillick:
@pp-mo I think it is tidier to have all of the options in one place (there's another one for [controlling parallel processing options](https://github.com/SciTools/iris/pull/2462)), and I intend them to inherit from a superclass `Option`. All of this suggests to me that we lump all the Options together into one place...

pp-mo:
> it is tidier to have all of the options in one place

Not if they are going to divide into separate self-contained areas of interest.
Which it already sounds like they are doing.

This is very much like iris.Future...
That is getting a bit awkward as more things are added : they don't really interact or belong together.
The difference with iris.Future is that we expect all the components to be ***transient***
  \-- when the implementation matures + deprecation cycle is complete, they get remove.
But that ***wouldn't*** happen with iris.options -- the complexity will simply increase steadily over time.

dkillick:
Following from an offline conversation with @pp-mo I've moved the Options to `config.py`, which we think might be the best fit given the code that already exists within Iris. It is possible that doing this has blatted some of @bjlittle's comments on these changes without addressing them... sorry!

marqh:
i think i am following the intent of this

the ability to control low level options in these ways seems useful

I am wondering how these are described for users.  I think there could be some be some explanatory docs in the refdocs for config to better handle this.  
Perhaps a user guide section, or references from the user guide in the appropriate sections, e.g. save

I also wonder whether the sample.site.cfg should include this, and perhaps remove some of the other deprecated options whilst we are there:
https://github.com/dkillick/iris/blob/8dace2350feac49fa92461665233548e33ff9c58/lib/iris/etc/site.cfg.template

this file is a useful reference, under source control

dkillick:
Rebased.

lib/iris/config.py
```diff
@@ -154,3 +156,101 @@ def get_dir_option(section, option, default=None):
...
+
+        * conventions_override (bool):
+            Define whether the CF Conventions version (e.g. `CF-1.6`) set when
+            saving a cube to a NetCDF file should be defined by
+            Iris (the default) or the cube being saved.
+
+            If `False` (the default), specifies that Iris should set the
+            CF Conventions version when saving cubes as NetCDF files.
+            If `True`, specifies that the cubes being saved to NetCDF should
+            set the cf conventions version for the saved NetCDF files.
```
bjlittle: @dkillick Apologies for being picky, but is this `cf conventions` or `CF Conventions` ...

lib/iris/config.py
```diff
@@ -154,3 +156,101 @@ def get_dir_option(section, option, default=None):
...
+
+        Example usages:
+
+        * Specify, for the lifetime of the session, that we want all cubes
+          written to NetCDF to define their own CF Conventions versions::
+
+            iris.options.netcdf.conventions_override = True
```
bjlittle: @dkillick This has gone state ... should now be `iris.config.netcdf.conventions_override`

lib/iris/config.py
```diff
@@ -154,3 +156,101 @@ def get_dir_option(section, option, default=None):
...
+            iris.save('my_cube', 'my_dataset.nc')
+            iris.save('my_second_cube', 'my_second_dataset.nc')
+
+        * Specify, with a context manager, that we want a cube written to
+          NetCDF to define its own CF Conventions version::
+
+            with iris.options.netcdf.context(conventions_override=True):
```
bjlittle: @dkillick Same reek of staleness here, thanks

lib/iris/config.py
dkillick: Nono, good spot! My bad, I'll change (it should be the latter).

lib/iris/config.py
```diff
@@ -154,3 +156,101 @@ def get_dir_option(section, option, default=None):
...
+                        'Defaulting to {!r}.')
+                warnings.warn(wmsg.format(value, name, good_value))
+                value = good_value
+        self.__dict__[name] = value
+
+    @property
+    def _defaults_dict(self):
+        return {'conventions_override': {'default': False,
+                                         'options': [True, False]},
+                }
```
bjlittle: @dkillick The `_defaults_dict` is private to the class, so is there a real need to make this a property?

Why can't the `self._defaults_dict` be created and initialised just the once in the `__init__` ?

lib/iris/config.py
```diff
@@ -154,3 +156,101 @@ def get_dir_option(section, option, default=None):
...
+        setattr(self, 'conventions_override', conventions_override)
+
+    def __str__(self):
+        msg = 'NetCDF options: conventions_override={}.'
+        return msg.format(self.conventions_override)
+
+    def __setattr__(self, name, value):
+        if name not in self.__dict__:
+            # Can't add new names.
+            msg = "'Option' object has no attribute {!r}".format(name)
```
bjlittle: @dkillick This isn't an `Option` instance ... or are you thinking in the generic sense? If so, would it be better to say something along the lines of `"{!r} option has no attribute {!r}.".format(type(self).__name__, name)` ...

lib/iris/config.py
```diff
@@ -154,3 +156,101 @@ def get_dir_option(section, option, default=None):
...
+        if value is None:
+            # Set an unset value to the name's default.
+            value = self._defaults_dict[name]['default']
+        if self._defaults_dict[name]['options'] is not None:
+            # Replace a bad value with a good one if there is a defined set of
+            # specified good values. If there isn't, we can assume that
+            # anything goes.
+            if value not in self._defaults_dict[name]['options']:
+                good_value = self._defaults_dict[name]['default']
+                wmsg = ('Attempting to set bad value {!r} for attribute {!r}. '
```
bjlittle: Minor suggestion, `bad` -> `invalid`

Do we want to raise a warning here, which may be silently ignored and use the default, or raise a hard exception ...

bjlittle:
@dkillick You have quite a few test failures. 

Without delving too deeply, I think that the failures might be related to the fact that the options are not thread safe? In that, a test is setting the `conventions_override` for the purpose of what it's trying to test, but this is a global setting that is interfering with other concurrent running tests ...

I'll leave you to dig and confirm whether this is the case or not ...

bjlittle:
@dkillick We still need to consider how this approach is going to scale ... at the moment we will be adding all of the options into `iris.config` ... I think we should consider whether we can easily separate our concerns, or accept that `iris.config` may bloat quite quickly ...

Also, as a user of iris I want an easy way to check what possible options are available to me and what specific options are set for a group i.e. what NetCDF options are set?, what regrid options are set?, what are all the options groups? etc

lib/iris/fileformats/netcdf.py
```diff
@@ -2244,7 +2244,10 @@ def is_valid_packspec(p):
                        shuffle, fletcher32, contiguous, chunksizes, endian,
                        least_significant_digit, packing=packspec)
 
-        conventions = CF_CONVENTIONS_VERSION
+        if iris.config.netcdf.conventions_override:
+            conventions = cube.attributes['Conventions']
```
bjlittle: @dkillick The `cube.attributes` may not necessarily have any `Conventions` to use ...

lib/iris/tests/unit/config/test_NetCDF.py
```diff
@@ -0,0 +1,57 @@
...
+# (at your option) any later version.
+#
+# Iris is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU Lesser General Public License for more details.
+#
+# You should have received a copy of the GNU Lesser General Public License
+# along with Iris.  If not, see <http://www.gnu.org/licenses/>.
+"""Unit tests for the `iris.options.Paralle` class."""
```
bjlittle: @dkillick cut'n paste ...

dkillick: And a typo!

dkillick:
> We still need to consider how this approach is going to scale

At the moment `iris.config` is still small and, I think, the best place to have options like these. If it ends up getting big in the future then maybe we can think about shuffling things around then. It may also be clearer how/if to when we have a lot more examples of options like these.

dkillick:
@bjlittle we have a green light!! :vertical_traffic_light:

bjlittle:
@dkillick Awesome, thanks! :+1:

PR #2473, opened 2017-03-31
Remove biggus test skipper: fix pp field data

lbdreyer:
This PR removes a single `skip_biggus` test skipper.

Work done:
* In the test I now set and check the fill_value on the cube rather on the cube.data.fill_value itself (the fill_value on the data is no longer considered by Iris).
* I get/set the field bmdi using the cube.fill_value, regardless of whether or not it is masked data (someone may have real unmasked data but still want a bmdi set)
* When realising a field's data it returns a masked array not an nd array with the nans filled with the bmdi
* I also had to fix the reference text file of the pp file as the repr of a field's data has changed (it used to be [this](https://github.com/SciTools/iris/blob/master/lib/iris/fileformats/pp.py#L1192))


lbdreyer:
Tests are failing because, in those tests, the cube has a fill value of 9999.0 so iris is now setting the bmdi to be that rather than the default bmdi.

lbdreyer:
@marqh @corinnebosley Could you take a look at this PR? 

I would like this merged in as I have some upcoming PRs based off this work (but makes more sense to be in a separate PR)

marqh:
@lbdreyer  this makes sense to me, based on the updated fill_value handling in the Cube API

PR #2474, opened 2017-04-03
Ensure tests work when installing from conda-forge channel

marqh:
trying to replace #2440 which has got it's travis testing in a muddle


marqh:
I am trying to establish the validity of this PR using this branch
https://github.com/marqh/iris/pull/55

which should pass, by just adopting conda-forge

pp-mo:
Merged #2440 instead

PR #2477, opened 2017-04-05
Use np.ma.is_masked() instead of isinstance(MaskedArray) in tests

djkirkham:
Due to some changes in Numpy 1.12, some aggregator functions in np.ma return a MaskedArray instance with no mask rather than an ndarray. This causes some of our tests to fail which do an instance check for a MaskedArray.

I've modified the checking code to use np.ma.is_masked(), but this causes other tests to fail where the array being checked has always been a MaskedArray instance with no mask (thus, the reference file claims it is masked when it isn't). Commit dfa75a3 demonstrates this - see the Travis test failures. Note that because some tests contain multiple checks, not all of the checks which would fail are listed in the output.


marqh:
thank you @djkirkham 

this is a better test pattern, I think, asking whether the data is masked, rather than checking if it's type is 'maskedarray'

the only changes here are test results which were not masked, but were of masked array type, so this is a testing pattern change only

lib/iris/tests/__init__.py
```diff
@@ -421,7 +421,7 @@ def assertDataAlmostEqual(self, data, reference_filename, **kwargs):
                 stats = json.load(reference_file)
                 self.assertEqual(stats.get('shape', []), list(data.shape))
                 self.assertEqual(stats.get('masked', False),
-                                       isinstance(data, ma.MaskedArray))
+                                 np.ma.is_masked(data))
```
bjlittle: @djkirkham We have a soft coding standard where we use `ma` instead of `np.ma`, see [here](https://github.com/SciTools/iris/pull/2477/files#diff-350ecfa750b99028f16446216ccbcfdbL68).

If it's not too much trouble, could you adopt this please, thanks.

lib/iris/tests/__init__.py
```diff
@@ -435,14 +435,13 @@ def assertDataAlmostEqual(self, data, reference_filename, **kwargs):
...
-                     'std': np.float_(data.std()),
-                     'max': np.float_(data.max()),
-                     'min': np.float_(data.min()),
-                     'shape': data.shape, 'masked': masked}
+            stats = collections.OrderedDict([
+                ('std', np.float_(data.std())),
+                ('min', np.float_(data.min())),
+                ('max', np.float_(data.max())),
+                ('shape', data.shape),
+                ('masked', np.ma.is_masked(data)),
```
bjlittle: @djkirkham Here is well ... `np.ma.is_masked(data)` -> `ma.is_masked(data)`

PR #2479, opened 2017-04-06
(OLD) Use dask for Iris grib loading/saving

lbdreyer:
This is my local checked out copy of the `gribiris` feature branch (the relevant PR to that branch was #2476) which I have rebased onto the `dask` feature branch.

Targeted at the `dask` feature branch.

I haven't unskipped any tests.


conda-requirements.txt
```diff
@@ -2,7 +2,6 @@
 # conda create -n <name> --file conda-requirements.txt
 
 # Mandatory dependencies
-biggus
```
lbdreyer: Maybe a bit jumping the gun... but I have removed all biggus imports !!

lib/iris/fileformats/grib/message.py
```diff
@@ -150,14 +154,18 @@ def data(self):
                 shape = (grid_section['numberOfDataPoints'],)
             else:
                 shape = (grid_section['Nj'], grid_section['Ni'])
-            proxy = _DataProxy(shape, np.dtype('f8'), np.nan,
+            proxy = _DataProxy(shape, np.dtype('f8'), self.bmdi,
```
lbdreyer: I needed to add a bmdi as `iris.fileformats.rules._make_cube` requires it (see https://github.com/lbdreyer/iris/blob/c8b497e7a7e94ea526df3a5352f2c2b42cebf1ae/lib/iris/fileformats/rules.py#L909)

I chose nans as that is what we had been using (and I made self.bmdi a property so we could reuse it)

lbdreyer:
In a subsequent commits I will be trying to take out the `tests.skip_biggus` skippers that are relevant to the GRIB changes.

lib/iris/tests/unit/fileformats/grib/test_GribWrapper.py
```diff
@@ -118,7 +117,7 @@ def test_regular_sequential(self):
         for i, _ in enumerate(tell_tale):
             gw = GribWrapper(grib_message, grib_fh)
             self.assertTrue(is_lazy_data(gw._data))
-            proxy = gw._data.concrete
+            proxy = as_concrete_data(gw._data)
             self.assertIsInstance(proxy, GribDataProxy)
             self.assertEqual(proxy.shape, (10, 20))
```
lbdreyer: These tests are failing. It looks like `gw._data.concrete` should return the proxy rather than realising the data.
@corinnebosley is looking into this test failure.

lib/iris/tests/unit/fileformats/grib/message/test_GribMessage.py
```diff
@@ -101,7 +99,7 @@ def test_bitmap_present(self):
                                       6: {'bitMapIndicator': 0,
                                           'bitmap': self.bitmap},
                                       7: {'codedValues': input_values}})
-        result = message.data.masked_array()
+        result = as_concrete_data(message.data)
         expected = np.ma.masked_array(output_values,
                                       np.logical_not(self.bitmap))
         expected = expected.reshape(self.shape)
```
lbdreyer: This test is failing as when we realise the data, it doesn't return a masked array. So it looks like we're no handling fill values correctly.

lbdreyer: This is now fixed and the tests are passing. But I'm not convinced that we have enough test coverage of testing how missing data is handled, considering the fact when missing data wasn't being handled correctly only one (1) test failed.

lbdreyer:
**Update:**
There were two types of test failiures:
- [x] lib/iris/tests/unit/fileformats/grib/message/test_GribMessage.test_bitmap_present
Now fixed but [see comment ](https://github.com/SciTools/iris/pull/2479#discussion_r110422438)

- [ ] lib/iris/tests/unit/fileformats/grib/test_GribWrapper.py
Ref: [these type of tests](https://github.com/lbdreyer/iris/blob/3bde38e4954941b2495f55239749f1af3e182925/lib/iris/tests/unit/fileformats/grib/test_GribWrapper.py#L113)
`GribWrapper` first makes an instance of `GribDataProxy` and then gives that proxy to dask which it assigns as GribWrapper._data. We want to test that GribWrapper._data is returning a lazy/dask array as well as that the proxy is being set up correctly. We are just struggling to work out how to do this.

lbdreyer:
I could squash all the commits after the commits that were added from #2476  to make it easier to review?

lib/iris/fileformats/grib/message.py
```diff
@@ -263,6 +263,10 @@ def __getitem__(self, keys):
                 raise TranslationError(msg)
 
         data = data.reshape(self.shape)
+
+        if np.ma.isMaskedArray(data):
```
bjlittle: @lbdreyer
```python
import numpy.ma as ma
...
if ma.isMaskedArray(data):
```

lbdreyer: Darn! I thought I'd caught these. Well spotted @bjlittle ! :eagle: :eyes:

lbdreyer:
I'm under the impression strict_grib_load has been deprecated, therefore should I be removing it from the integration tests?? 
e.g. from here: https://github.com/SciTools/iris/blob/gribiris/lib/iris/tests/integration/test_grib2.py#L55

bjlittle:
@lbdreyer Make sense to me, particularly since the `dask` branch will be a major release, and `strict_grib_load` was deprecated in `1.10`, so I'm :+1: for removing this from the future support and tests.

bjlittle:
@lbdreyer It would be super nice if we removed the following comment references to `biggus`:
* see `fileformats/grib/message.py` [line 92](https://github.com/SciTools/iris/pull/2479/files#diff-91d60d4d5c80761804bb7c3e58f43646R92)
* see `fileformats/grib/message.py` [line 120](https://github.com/SciTools/iris/pull/2479/commits/e301c5991d1adb9947b2d9b997476733963facde#diff-91d60d4d5c80761804bb7c3e58f43646R120)

This would me that we've remove all references to `biggus` (apart from any skippers) in the code base comments (apart from `iris/_lazy_data.py` as well, but that's a historical reference)

Although, `docs/src/conf.py` does still reference `biggus` ... :wink:

lib/iris/fileformats/grib/__init__.py
```diff
@@ -636,6 +636,20 @@ def _get_verification_date(self):
         # Return validity_time = (reference_time + start_offset*time_unit).
         return reference_date_time + interval_delta
 
+    @property
+    def bmdi(self):
+        # Not sure of any cases where GRIB provides a fill value.
+        # Default for fill value is None.
+        return None
```
bjlittle: @lbdreyer I'm assuming that `bmdi` is used by the rules?

We need to verify GRIBs relationship with `fill_value` ...

lib/iris/fileformats/grib/__init__.py
```diff
@@ -636,6 +636,20 @@ def _get_verification_date(self):
...
+
+    @property
+    def core_data(self):
+        try:
+            data = self._data
+        except AttributeError:
+            data = self.data
+        return data
```
bjlittle: @lbdreyer Again, is `core_data` for the rules?

lib/iris/fileformats/grib/message.py
```diff
@@ -152,12 +153,22 @@ def data(self):
...
         else:
             fmt = 'Grid definition template {} is not supported'
             raise TranslationError(fmt.format(template))
         return data
 
+    @property
+    def bmdi(self):
+        # Not sure of any cases where GRIB provides a fill value.
+        # Default for fill value is None.
+        return None
```
bjlittle: @lbdreyer Same comments as for `GribWrapper`

lib/iris/tests/__init__.py
```diff
@@ -92,19 +92,12 @@
...
-        from iris.fileformats.grib.message import GribMessage
-    except ImportError:
-        GRIB_AVAILABLE = False
-
-# skip all iris_grib tests until a new dask aware iris_grib is available
-GRIB_AVAILABLE = False
+    GRIB_AVAILABLE = False
+else:
+    GRIB_AVAILABLE = True
+    from iris.fileformats.grib.message import GribMessage
```
bjlittle: @lbdreyer Is this all just equivalent to ...
```python
try:
    import gribapi
    GRIB_AVAILABLE = True
    from iris.fileformats.grib.message import GribMessage
except ImportError:
    GRIB_AVAILABLE = False
```

lib/iris/tests/integration/test_grib2.py
```diff
@@ -36,16 +36,13 @@
...
     try:
-        # Try to load the independent 'iris_grib' package.
-        from iris_grib import load_pairs_from_fields
-        from iris_grib.message import GribMessage
-    except ImportError:
-        # Try to load old inbuilt module instead.
         from iris.fileformats.grib import load_pairs_from_fields
         from iris.fileformats.grib.message import GribMessage
+    except ImportError:
+        pass
```
bjlittle: @lbdreyer If `GRIB_AVAILABLE == True` then why wrap this all in a `try ... except` block?

lib/iris/tests/integration/test_grib_load.py
```diff
@@ -83,18 +56,11 @@ def test_load_3_layer(self):
...
             ('GRIB', 'missing_values', 'missing_values.grib2'))
         cubes = iris.load(gribfile)
         self.assertCML(cubes, ('grib_load', 'missing_values_grib2.cml'))
 
-    @skip_irisgrib_fails
-    def test_y_fastest(self):
-        cubes = iris.load(tests.get_data_path(("GRIB", "y_fastest",
-                                               "y_fast.grib2")))
-        self.assertCML(cubes, ("grib_load", "y_fastest.cml"))
-
```
bjlittle: @lbdreyer I don't have the context, but why are all of these tests now not relevant?

bjlittle:
@lbdreyer I'm wondering if we have appropriate test coverage for lazy masked integral cases ... there certainly isn't any `result_dtype` being passed through the `as_concrete_data` anywhere. Nor is the `nans_replacement` being set to a `fill_value` ... we should consider this.

bjlittle:
@lbdreyer The is quite a lot of churn on this PR, so I'm going to try to understand the changes at a file level rather than a commit level

bjlittle:
@lbdreyer For `grib/_load_convert.py` this PR compared with `iris-grib` has the following differences:
* this PR is missing `product_definition_template_15`
* this PR is missing `product_definition_template_32`
  * `product_definition_template_32` in `iris-grib` introduces the `time_coords` convenience function, which should be shared/called by `product_definition_template_32` and `product_definition_template_0`
  * `product_definition_template_32` in `iris-grib` introduces the `satellite_common` convenience function, which should be shared/called by `product_definition_template_32` and `product_definition_template_31`
* `iris-grib` has a `statistical_method_name` convenience function, which is called by the new `product_definition_template_15` and `statistical_cell_method`
* in `iris-grib` the function `hybrid_factories` performs an integer division i.e. `offset += NV // 2` see line 1354 in `iris-grib`, but float division is done in this PR
* in `iris-grib` the function `generating_process` has the kwarg `include_forecast_process` that defaults to `True` but `product_definition_template_31` and `product_definition_template_32` call `generating_process` with the kwarg explicitly set to `False`

bjlittle:
@lbdreyer For `grib/__init__.py` [line 196](https://github.com/SciTools/iris/pull/2479/files#diff-c9dbbf5c6e3c981c89624886fe1740d2R196) this PR raises an `IrisError` exception, but `iris-grib` raises a `ValueError` ... I think this should be a `ValueError`

lib/iris/fileformats/grib/__init__.py
lbdreyer: Yep. It was complaining if it didn't have a core_data or bmdi

lib/iris/tests/__init__.py
lbdreyer: It is. I will change this

lib/iris/tests/integration/test_grib2.py
lbdreyer: Good point! I will remove this

bjlittle:
@lbdreyer For `grib/_save_rules.py` this PR compared with `iris-grib` has the following differences:
* this PR does not support the `LambertConformal` coordinate-system
* the PR does not support the new `grid_definition_template_30`, which uses the `LambertConformal` cs
* there are numerous peppered small differences between the two versions of `_save_rules.py` that I can't reconcile/understand ...  so its pretty difficult to tell what's right :cry:

lbdreyer:
@bjlittle I am very concerned with all these differences you are finding. It makes me have less confidence #2476 was correct, it seems like a lot was missed!

I do wonder whether an old version of iris-grib was used when copy and pasting it back into Iris.

bjlittle:
There has been some recent activity on `iris-grib`, which appears to be the source of the differences, namely:
* https://github.com/SciTools/iris-grib/pull/77
* https://github.com/SciTools/iris-grib/pull/76
* https://github.com/SciTools/iris-grib/pull/75
* https://github.com/SciTools/iris-grib/pull/73

bjlittle:
@lbdreyer I've not even dipped into the associated test changes/differences ...

lib/iris/fileformats/grib/__init__.py
```diff
@@ -32,65 +26,49 @@
...
-           'as_messages', 'as_pairs', 'grib_generator', 'reset_load_rules',
-           'hindcast_workaround']
-
-
-#: Set this flag to True to enable support of negative forecast periods
-#: when loading and saving GRIB files.
-#:
-#: .. deprecated:: 1.10
-hindcast_workaround = False
+           'save_pairs_from_cube', 'save_messages']
```
corinnebosley: What has happened to GribWrapper from this list?  The class is still in the module (although it has been altered), does it not need to be in here still as it has been undeprecated?

corinnebosley: I don't really understand what this list is a list of....

bjlittle: See https://docs.python.org/2/tutorial/modules.html#importing-from-a-package

corinnebosley: So is there a good reason for GribWrapper and GribDataProxy not being in this list?

lib/iris/fileformats/grib/_grib1_load_rules.py
```diff
@@ -358,74 +250,16 @@ def add_bounded_time_coords(aux_coords_and_dims, grib):
...
-    if \
-            (grib.edition == 2) and \
-            (grib.centre == 'ecmf') and \
-            (grib.discipline == 0) and \
-            (grib.parameterCategory == 3) and \
-            (grib.parameterNumber == 25) and \
-            (grib.typeOfFirstFixedSurface == 105):
-        references.append(ReferenceTarget('surface_air_pressure', lambda cube: {'standard_name': 'surface_air_pressure', 'units': 'Pa', 'data': np.exp(cube.data)}))
-
     return ConversionMetadata(factories, references, standard_name, long_name,
```
corinnebosley: Why has loads of stuff been taken out of here? I can't really pass comment on this file as I don't understand why these changes have been made.

lib/iris/fileformats/grib/__init__.py
bjlittle: Yes there is ... `GribDataProxy` was never in this list. It's really internal to our support of GRIB, so I believe that it shouldn't be exposed within `__all__` at all.

Also, from what I recall, during the port of the original GRIB functionality from `iris` into `iris-grib` we took the opportunity to remove `GribWrapper` from the `__all__` as the use of `GribWrapper` is the legacy implementation and was only being maintained for GRIB1 support and not for GRIB2 support. We put a big effort into moving away from `GribWrapper` for GRIB2 support.

Does that help?

corinnebosley: Yes.  That all makes sense.  I'm much happier with that now, thanks.

lib/iris/tests/integration/test_grib_load.py
corinnebosley: I would like to second that confusion...

lbdreyer:
Looks like [Iris-grib PR #72](https://github.com/SciTools/iris-grib/pull/72) is also missing

lib/iris/tests/test_grib_load_translations.py
```diff
@@ -324,55 +310,9 @@ def test_timeunits_calendar(self):
...
-            datetime.datetime(year=2007, month=3, day=23,
-                              hour=12, minute=0, second=0)
-        )
-        self.assertEqual(
-            grib_wrapper._periodStartDateTime,
-            datetime.datetime(year=2007, month=3, day=23,
-                              hour=22, minute=0, second=0)
-        )
-        self.assertEqual(grib_wrapper._periodEndDateTime, exp_end_date)
-
```
corinnebosley: Why do we not need this now?

lib/iris/tests/unit/fileformats/grib/test_GribWrapper.py
```diff
@@ -60,97 +64,81 @@ def _mock_grib_get_native_type(grib_message, key):
...
-            proxy = gw._data.concrete
-            self.assertIsInstance(proxy, GribDataProxy)
-            self.assertEqual(proxy.shape, (10, 20))
-            self.assertEqual(proxy.dtype, np.float)
-            self.assertIs(proxy.fill_value, np.nan)
-            self.assertEqual(proxy.path, grib_fh.name)
-            self.assertEqual(proxy.offset, _message_length * i)
-            self.assertEqual(proxy.regularise, auto_regularise)
-
-    def test_regular_mixed(self):
```
corinnebosley: Are there going to be lots of merge conflicts in this file because of the stuff that Laura and I have been doing this week on her branch?

I am a bit reluctant to pass comment on this file as it might just be irrelevant...

corinnebosley: Wait a minute, this file _contains_ the changes Laura and I have made, right?  In which case everything is ok.  I hope.  The whole thing still makes me nervous.

bjlittle:
@SciTools/iris-devs 

I've commented on the Google Groups iris-dev [iris grib future](https://groups.google.com/forum/#!topic/scitools-iris-dev/lFw2_CnaEHU) thread, discussing the intent of this PR ... have your say if you care!

lbdreyer:
@bjlittle I have now added the missing PRs I could squash these at a later stage. The only problems is @marqh was particular about what the commit history should look like and now this messes that up

lib/iris/tests/integration/test_grib_load.py
lbdreyer: They don't work with iris_grib (see [this skipper that was on these tests ](https://github.com/SciTools/iris/blob/master/lib/iris/tests/integration/test_grib_load.py#L54)).

I decided just to remove them but in the new PR #2486 I have commented them out for now.

lib/iris/tests/test_grib_load_translations.py
lbdreyer: This tests uses GribWrapper, which now only works for GRIB1. As you can see this uses a grib2 file. @marqh chose to just remove this tests as it was no longer correct (which is also what I chose to do in #2486 as I just didn't think it was necessary and changing the line
```
       grib_message = gribapi.grib_new_from_samples('GRIB2')`
```
to
```
       grib_message = gribapi.grib_new_from_samples('GRIB1')
```

lib/iris/tests/unit/fileformats/grib/test_GribWrapper.py
lbdreyer: You made the PR to my branch which this PR was already using, so when I merged in your PR it automatically updated this PR.
For the new PR (#2486) I just cherrypicked your commit and added it to that branch

lbdreyer:
As the initial work of bring iris_grib into iris missed a lot, it made this PR messy. I have created a new PR doing the exact same as this PR but it covers the missing parts and is hopefully more clearer (see #2486 )

I want to close this PR but I will leave it open for now in case for some reason  you'd prefer this one to #2486

lib/iris/fileformats/grib/_grib1_load_rules.py
lbdreyer: This file was called `load_rules.py` and had rules for grib edition 1 and grib edition 2. However it is now only being used to load grib 1 and so has been renamed   `_grib1_load_rules.py`. So all the edition 2 stuff needs to be taken out and all the checks for edition 1 are now unecessary as this code will only ever be used when the edition is 1

PR #2485, opened 2017-04-13
Ignore missing formula terms when constructing aux factories

djkirkham:
The CF conventions state that certain `formula_terms` terms may be omitted and assumed to be zero (http://cfconventions.org/cf-conventions/v1.6.0/cf-conventions.html#dimensionless-v-coord). This logic is implemented in the various `AuxCoordFactory` sub-classes in Iris, but when constructing factories from data loaded from a netCDF file an error is raised if a term is missing. This PR allows factories to be constructed with missing terms.

The change has the perhaps undesirable side effect that if a formula term refers to a variable which doesn't exist, or one which has been ignored because it wasn't valid, the factory will still be constructed with that coordinate omitted. Previously this would result in an error caused by a dependency coordinate not being present in an `AuxCoordFactory` when merging (see https://github.com/SciTools/iris/blob/master/lib/iris/_merge.py#L1679). But it looks like this was unintentional; the `AuxCoordFactory` class are written in a way which allows missing coordinates.

### DIFF
--- a/lib/iris/tests/unit/fileformats/netcdf/test__load_aux_factory.py
+++ b/lib/iris/tests/unit/fileformats/netcdf/test__load_aux_factory.py
@@ -1,4 +1,4 @@
-# (C) British Crown Copyright 2014 - 2015, Met Office
+# (C) British Crown Copyright 2014 - 2017, Met Office
 #
 # This file is part of Iris.
 #
--- a/lib/iris/tests/unit/fileformats/netcdf/test__load_aux_factory.py
+++ b/lib/iris/tests/unit/fileformats/netcdf/test__load_aux_factory.py
@@ -110,13 +110,38 @@ class TestAtmosphereHybridSigmaPressureCoordinate(tests.IrisTest):
             self.assertEqual(msg, str(warn[0].message))
 
     def test_formula_terms_ap_missing_coords(self):
-        coordinates = [(mock.sentinel.b, 'b'), (mock.sentinel.ps, 'ps')]
-        self.provides = dict(coordinates=coordinates)
         self.requires['formula_terms'] = dict(ap='ap', b='b', ps='ps')
         with mock.patch('warnings.warn') as warn:
             _load_aux_factory(self.engine, self.cube)
         warn.assert_called_once_with("Unable to find coordinate for variable "
                                      "'ap'")
+        # Check cube.add_aux_coord method.
+        self.assertEqual(self.cube.add_aux_coord.call_count, 0)
+        # Check cube.add_aux_factory method.
+        self.assertEqual(self.cube.add_aux_factory.call_count, 1)
+        args, _ = self.cube.add_aux_factory.call_args
+        self.assertEqual(len(args), 1)
+        factory = args[0]
+        # Check that the factory has no delta term
+        self.assertEqual(factory.delta, None)
+        self.assertEqual(factory.sigma, mock.sentinel.b)
+        self.assertEqual(factory.surface_air_pressure, mock.sentinel.ps)
+
+    def test_formula_terms_no_delta_terms(self):
+        self.requires['formula_terms'] = dict(b='b', ps='ps')
+        _load_aux_factory(self.engine, self.cube)
+        # Check cube.add_aux_coord method.
+        self.assertEqual(self.cube.add_aux_coord.call_count, 0)
+        # Check cube.add_aux_factory method.
+        self.assertEqual(self.cube.add_aux_factory.call_count, 1)
+        args, _ = self.cube.add_aux_factory.call_args
+        self.assertEqual(len(args), 1)
+        factory = args[0]
+        # Check that the factory has no delta term
+        self.assertEqual(factory.delta, None)
+        self.assertEqual(factory.sigma, mock.sentinel.b)
+        self.assertEqual(factory.surface_air_pressure, mock.sentinel.ps)
+
 
 if __name__ == '__main__':
     tests.main()
--- a/lib/iris/tests/unit/fileformats/netcdf/test__load_aux_factory.py
+++ b/lib/iris/tests/unit/fileformats/netcdf/test__load_aux_factory.py
@@ -109,12 +109,7 @@ class TestAtmosphereHybridSigmaPressureCoordinate(tests.IrisTest):
                 'coordinate {!r} bounds.'.format(coord_p0.name())
             self.assertEqual(msg, str(warn[0].message))
 
-    def test_formula_terms_ap_missing_coords(self):
-        self.requires['formula_terms'] = dict(ap='ap', b='b', ps='ps')
-        with mock.patch('warnings.warn') as warn:
-            _load_aux_factory(self.engine, self.cube)
-        warn.assert_called_once_with("Unable to find coordinate for variable "
-                                     "'ap'")
+    def _check_no_delta(self):
         # Check cube.add_aux_coord method.
         self.assertEqual(self.cube.add_aux_coord.call_count, 0)
         # Check cube.add_aux_factory method.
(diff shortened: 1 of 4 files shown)


bjlittle:
@djkirkham Do you have a concrete use case for this? 

What exception are you seeing for the failure?

I've tripped over a similar case in the past, but it was really, really subtle in the exception that it raised ... I wonder if this is straying into the same territory. I'm sure I've got a branch (which I've not pushed) resolving the issue. Let me dig ...

djkirkham:
@bjlittle It's possible to replicate the error by modifying the sample data file 'hybrid_height.nc'.

The following command modifies the file so that that the `formula_terms` attribute of the `level_height` variable contains no `b` term:
```
ncatted -a formula_terms,level_height,o,c,'a: level_height orog: surface_altitude' hybrid_height.nc hybrid_height_nosigma.nc
```
I think this should be allowed, but loading in Iris results in the following:
```
>>> import iris
>>> iris.load('hybrid_height_nosigma.nc')
[...]/iris/fileformats/cf.py:1143: IrisDeprecation: NetCDF default loading behaviour currently does not expose variables which define reference surfaces for dimensionless vertical coordinates as independent Cubes. This behaviour is deprecated in favour of automatic promotion to Cubes. To switch to the new behaviour, set iris.FUTURE.netcdf_promote to True.
  warn_deprecated(msg)
  [stack trace omitted]
  File "[...]/iris/fileformats/netcdf.py", line 578, in coord_from_term
    name = engine.requires['formula_terms'][term]
KeyError: 'b'
```
The following modifies the fomula terms so that the term `b` refers to a non-existent variable, `foo`:
```
ncatted -a formula_terms,level_height,o,c,'a: level_height b: foo orog: surface_altitude' hybrid_height.nc hybrid_height_foo.nc
```
This results in the same error with the following warning (I previously incorrectly claimed that it resulted in an error when merging):
```
[...]/iris/fileformats/cf.py:557: UserWarning: Missing CF-netCDF formula term variable u'foo', referenced by netCDF variable u'level_height'
  warnings.warn(message % (variable_name, nc_var_name))
```
There is a more subtle case which I'm unable to replicate with the sample data: if you have a data variable, 'd',  which refers to a derived coordinate variable, 'c', (i.e. one with a `formula_terms` attribute)  and 'c' has a formula terms variable 't' which 'd' does not span (still with me?), then 't' will be ignored with a warning of the form:
```
[...]/iris/fileformats/cf.py:1062: UserWarning: Ignoring formula terms variable u't' referenced by data variable u'd' via variable u'c': Dimensions (u'x', u'y') do not span (u'u', u'v')
```
This causes the error with aux coord factory dependencies I mentioned:
```
[stack trace omitted]
  File "[...]/iris/_merge.py", line 1679, in _extract_coord_payload
    dependency_defns.append((key, coord._as_defn()))
AttributeError: 'NoneType' object has no attribute '_as_defn'
```
For the latter two cases we might still want to raise an error (this PR changes that), although we are currently raising warnings at the most appropriate point where might want to raise an error, so it seems like it was a conscious decision not to, and the errors which _are_ raised are possibly unintentional.

djkirkham:
@bjlittle The use case which led me to this error was a bit of a strange one. A user had a file with a variable which didn't specify its bounds variable in the `bounds` attribute. The bounds variable was incorrectly being identified by Iris as a data variable, and failing to load because of the third case I mentioned above. With this change the data would still load, with the bounds variable as a separate cube.

corinnebosley:
This one's a bit tricky.  I agree with the intention of the change, and also the implementation.  I have concerns that this will reduce the obviousness of erroneous netCDF files, but at the same time, all you ever get is a string of warnings anyway.  This will just reduce the number of warnings by one.  If anything, this may even make it more clear what the problem is, because a user might be more inclined to pay attention to a single warning rather than a string of them.

@marqh @lbdreyer I am inclined to approve this change, but I would appreciate some thoughts from you about this.  It's definitely not one I'd merge on impulse.

lbdreyer:
@djkirkham Have you looked at writing tests for this, perhaps [a unit test](https://github.com/djkirkham/iris/blob/9a429e991bfadfdc8c8e1f0b7525e65f5c674b68/lib/iris/tests/unit/fileformats/netcdf/test__load_aux_factory.py)?

djkirkham:
@lbdreyer I've now added unit tests.

In `test__load_aux_factory` I've added a test for the case where a formula term is not present in the `formula_terms` attribute of the variable in netCDF file. The way this manifests in the code is that the `engine.requires['formula_terms']` dictionary doesn't have a key for that term. I've tested that the factory still gets created, with `None` for the relevant term. I've also modified an existing test which was checking for a warning if a required coordinate was not present: I've checked that the factory is created with the relevant term missing.

I've also added a test to the `ProtoCube` tests to check that cubes will not merge when one of them has an aux factory with missing terms. This also serves to exercise the `if` block I added in `_merge.py`.

lbdreyer:
As discussed offline, there is the potential confusion when someone loads in a cube, where the file contains a sigma variable but it is not included in the formula terms.
i.e. 
```level_height:formula_terms = "a: level_height orog: surface_altitude" ;```

In this example, the aux_factory has no sigma, but a sigma coordinate is loaded in from the sigma variable
```
>>> print cube.aux_factories[0]
<HybridHeightFactory(delta=u'atmosphere_hybrid_height_coordinate', orography=u'surface_altitude', sigma=None)>
>>> print cube.coord('sigma')
AuxCoord(array([ 0.9994238 ,  ..., 0.90498137], dtype=float32), bounds=array([[ 1.        ,  0.99846387], ...
 [ 0.91065508,  0.89896113]], dtype=float32), standard_name=None, units=Unit('1'), long_name=u'sigma', var_name='sigma')
```

**Comment by corinnebosley** [comment]
@lbdreyer I can see the confusion that may arise with this scenario, but don't think that is a reason to reject this change.  Would you suggest than @djkirkham applies a catch for that situation with a user warning?

**Comment by djkirkham** [comment]
@corinnebosley This was discussed, but I don't think it's possible because the name of the coordinate ('sigma' in this case) is irrelevant for formula terms. So there's no way to distinguish between the cases when a coordinate might have been omitted from the formula terms in error, and when the term wasn't included intentionally.

**Comment by corinnebosley** [comment]
@djkirkham But wouldn't a warning cover both of those possibilities? It would inform a user who had erroneously omitted the term, and a user for whom the omission was intentional would just ignore the warning.

**Comment by djkirkham** [comment]
@corinnebosley I don't think we should be giving warnings when the file is formatted correctly, which it would be in the second case.

**Comment by corinnebosley** [comment]
@djkirkham Doesn't the current behaviour raise an error in this case?  Isn't a warning better than an error?

File: lib/iris/tests/unit/fileformats/netcdf/test__load_aux_factory.py [hunk 1]
```diff
@@ -110,13 +110,38 @@ def test_formula_terms_p0_bounded(self):
...
+        args, _ = self.cube.add_aux_factory.call_args
+        self.assertEqual(len(args), 1)
+        factory = args[0]
+        # Check that the factory has no delta term
+        self.assertEqual(factory.delta, None)
+        self.assertEqual(factory.sigma, mock.sentinel.b)
+        self.assertEqual(factory.surface_air_pressure, mock.sentinel.ps)
+
+    def test_formula_terms_no_delta_terms(self):
+        self.requires['formula_terms'] = dict(b='b', ps='ps')
```
lbdreyer: Is it worth also adding a test for 
```self.requires['formula_terms'] = dict(a='a', b='b', ps='ps')```
```self.requires['formula_terms'] = dict(p0='p0', b='b', ps='ps')```

It then checks [this logic](https://github.com/SciTools/iris/pull/2485/files#diff-e055d3568b9b3f9cfe397cec7e016700R608)

Comment (lbdreyer) [inline lib/iris/tests/unit/fileformats/netcdf/test__load_aux_factory.py]: As in it checks when you only have one of 'a' or 'p'. Is that a possible scenario?

Comment (djkirkham) [inline lib/iris/tests/unit/fileformats/netcdf/test__load_aux_factory.py]: Yes, I think it's possible, but nonsensical. Since the two terms are multiplied together in [the formula](http://cfconventions.org/cf-conventions/v1.6.0/cf-conventions.html#_atmosphere_hybrid_sigma_pressure_coordinate), if one of them is zero it doesn't matter what the other one is, so there'd be no point having both. I'll add tests anyway though.

File: lib/iris/tests/unit/fileformats/netcdf/test__load_aux_factory.py [hunk 2]
```diff
@@ -109,14 +109,46 @@ def test_formula_terms_p0_bounded(self):
...
+        self.assertEqual(len(args), 1)
+        factory = args[0]
+        # Check that the factory has no delta term
+        self.assertEqual(factory.delta, None)
+        self.assertEqual(factory.sigma, mock.sentinel.b)
+        self.assertEqual(factory.surface_air_pressure, mock.sentinel.ps)
+
     def test_formula_terms_ap_missing_coords(self):
-        coordinates = [(mock.sentinel.b, 'b'), (mock.sentinel.ps, 'ps')]
-        self.provides = dict(coordinates=coordinates)
```
djkirkham: These lines weren't actually doing anything. Assigning to `self.provides` after `setUp` has run has no effect.

lbdreyer:
Thanks for persevering with this @djkirkham !

Issue #1887, opened 2015-12-23
Cube broadcasting for addition/subtraction.


rcomer:
As @shoyer noted at PR #888, the cube broadcasting rules for addition and subtraction are different (more strict) than for multiplication and division. Lines 260-279 in [`iris.analysis.maths`](https://github.com/SciTools/iris/blob/2eecc2e1aacca4637627b4dafa2f63009dfec7fb/lib/iris/analysis/maths.py#L260-279) have the extra add/subtract checks.  This has caught me out a few times when trying to take differences between very similar cubes.

For example, today the subtract threw an error with `This operation cannot be performed as there are differing coordinates (latitude, longitude) remaining which cannot be ignored.`  It took some digging to realise that the only difference was that `var_name` was set for both coords on only one of the cubes.

I'm wondering if these extra checks for addition and subtraction could simply be removed.  There is a specific (old) [test](https://github.com/rcomer/iris/blob/c61bde62a7db4c6a4b9da96f0be1c1511003b67d/lib/iris/tests/test_basic_maths.py#L73), so the behaviour is obviously deliberate.  I don't understand why though.

pelson:
Thanks @rcomer. I agree - this is peculiar behaviour, and there is no obvious comment justifying it. If you're keen, it might be worth putting the straw man PR up which just uses _binary_op_common for add & subtract...

rcomer:
Thanks @pelson.  I'll do that when I get a chance.

rwbodman:
Hi,
Any work around for this subtraction problem?  I have two cubes from which I have removed most of the Scalar Coordinates, leaving just:

>>> print model_ann_mean
unknown / (K)                       (latitude: 144; longitude: 192)
     Dimension coordinates:
          latitude                           x               -
          longitude                          -               x
     Scalar coordinates:
          time: 1999-01-01 12:00:00, bound=(1989-01-01 00:00:00, 2009-01-01 00:00:00)
     Attributes:
          Conventions: CF-1.5
>>> print era_ann_mean
unknown / (K)                       (latitude: 144; longitude: 192)
     Dimension coordinates:
          latitude                           x               -
          longitude                          -               x
     Scalar coordinates:
          time: 1998-12-17 00:00:00, bound=(1988-12-16 12:00:00, 2008-12-16 12:00:00)
     Attributes:
          Conventions: CF-1.5

But:

>>> bias_ann_mean = model_ann_mean  - era_ann_mean
Traceback (most recent call last):
[...]

PR #2498, opened 2017-04-21
Badmaths tests and fixes

corinnebosley:
When two cubes have coordinates which do not match, an error should be raised if a user tries to perform arithmetic operations on them.  These cases are caught by the add and subtract operators, but not the multiply and divide operators.

This PR contains tests for all the operators on a) a pair of cubes with totally non-matching coordinates and b) a pair of cubes with one coordinate array reversed.

It also extends the behaviour of add and subtract to apply to multiply and divide.

### DIFF
--- a/lib/iris/analysis/maths.py
+++ b/lib/iris/analysis/maths.py
@@ -1,4 +1,4 @@
-# (C) British Crown Copyright 2010 - 2016, Met Office
+# (C) British Crown Copyright 2010 - 2017, Met Office
 #
 # This file is part of Iris.
 #
--- a/lib/iris/analysis/maths.py
+++ b/lib/iris/analysis/maths.py
@@ -344,8 +344,8 @@ def multiply(cube, other, dim=None, in_place=False):
         # If a coordinate is to be ignored - remove it
         ignore = filter(None, [coord_grp[0] for coord_grp
                         in coord_comp['ignorable']])
-    for coord in ignore:
-        new_cube.remove_coord(coord)
+        for coord in ignore:
+            new_cube.remove_coord(coord)
 
     return new_cube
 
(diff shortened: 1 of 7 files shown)


corinnebosley:
The user who reported this bug requested that iris catch the case of reversed coordinates and work out that the coordinates match up, but one array is reversed, and simply deals with this accordingly.

I believe that this is not what iris should be doing, so I have opted to catch the case and raise an error instead, letting the user adjust the cube themself.

corinnebosley:
@marqh Default test on last run timed out.  Apart from that, everything seems to pass OK.

marqh:
@corinnebosley this looks good, thank you

djkirkham:
It has been reported that this is causing issues with user code. Is it worth releasing 1.13.1 with this change taken out?

@corinnebosley @marqh @pelson

rcomer:
Hello,

it was me that reported issues as my code is breaking at 1.13.  There is a relevant issue (#1887) and PR (#1911) with some discussion about what tests would be sensible.  Unfortunately we never came to any firm conclusions!

Ruth

corinnebosley:
Hi @rcomer, thanks for your feedback, I have raised an issue ([see this ticket](https://github.com/SciTools/iris/issues/2599)) regarding the behaviour of math ops.  I agree that the current behaviour is not quite right, but I also think that it should be consistent between add/subtract and multiply/divide, so we need to decide as a team what the behaviour should be.  I have added this to the current project so that we discuss this issue within the dev team and make sure that it gets addressed very soon.

In the meantime, you can use ```module load scitools/default-previous``` to use Iris 1.12 instead of 1.13 to be compatible with your code.

What would be really useful is if you could put together a few small use cases for us (like the gravity one in [this comment](https://github.com/SciTools/iris/pull/1911#issuecomment-191152640)) that we can use to make a decision on behaviour and implementation.

I will make sure that you are notified about any changes in this area so that we can get your opinion and suggestions.  Thanks!

PR #2511, opened 2017-04-25
Dask options for Iris processing

dkillick:
A much simplified replacement for #2462 and thus also a fix for #2403.

Instead of having a highly complex Iris option class that places an Iris API around the "non-API" of `dask.set_options` and potentially tying ourselves into a very tight coupling to `dask.set_options`, this PR just calls `dask.set_options` from within `iris._lazy_data` at import time. The call to `dask.set_options` is conditional on the user having not already specified dask options so that importing Iris does not change dask options that are not the default.

The only dask option that may be set by Iris is setting the `get` function to the synchronous scheduler `dask.async.get_sync`. This scheduler only runs on a single thread, which matches to the requirement suggested in #2403. In testing this scheduler was shown to be faster than using the conceptually equivalent dask options `{'get': dask.threaded.get, 'pool': ThreadPool(1)}`, and simpler to construct/maintain.

### DIFF
--- a/lib/iris/_lazy_data.py
+++ b/lib/iris/_lazy_data.py
@@ -23,9 +23,32 @@ To avoid replicating implementation-dependent test and conversion code.
 from __future__ import (absolute_import, division, print_function)
 from six.moves import (filter, input, map, range, zip)  # noqa
 
+import dask
 import dask.array as da
+import dask.context
 import numpy as np
 import numpy.ma as ma
+
+
+def _iris_dask_defaults():
+    """
+    Set dask defaults for Iris. The current default dask operation mode for
+    Iris is running single-threaded using `dask.async.get_sync`. This default
+    ensures that running Iris under "normal" conditions will not use up all
+    available computational resource.
+
+    .. note::
+        We only want Iris to set dask options in the case where doing so will
+        not change user-specified options that have already been set.
+
+    """
+    if 'pool' not in dask.context._globals and \
+            'get' not in dask.context._globals:
+        dask.set_options(get=dask.async.get_sync)
+
+
+# Run this at import time to set dask options for Iris.
+_iris_dask_defaults()
 
 
 def is_lazy_data(data):
--- a/lib/iris/_lazy_data.py
+++ b/lib/iris/_lazy_data.py
@@ -37,6 +37,9 @@ def _iris_dask_defaults():
     ensures that running Iris under "normal" conditions will not use up all
     available computational resource.
 
+    Otherwise, by default, `dask` will use a multi-threaded scheduler that uses
+    all available CPUs.
+
     .. note::
         We only want Iris to set dask options in the case where doing so will
         not change user-specified options that have already been set.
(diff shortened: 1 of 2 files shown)


lib/iris/tests/unit/lazy_data/test_iris_dask_defaults.py
```diff
@@ -0,0 +1,73 @@
...
+        self.context = 'dask.context'
+        self._globals = 'iris._lazy_data._globals'
+        set_options = 'dask.set_options'
+        self.patch_set_options = self.patch(set_options)
+        get_sync = 'dask.async.get_sync'
+        self.patch_get_sync = self.patch(get_sync)
+
+    def test_no_user_options(self):
+        test_dict = {}
+        with self.patch(self.context, _globals=test_dict):
```
pp-mo: You don't need to use "with" with the IrisTest.patch() method.
The point of it is to automatically remove itself after each test, so you avoid lots of extra indentation ...

pp-mo:
Actually I think that the tests can be much simplified.
As all the tests work the same way, instead of a combination of a  setUp and boilerplate code it should be clearer to use a common utility routine which centralises the testing method.
Looking at the current functionality, I suggest [these changes](https://github.com/dkillick/iris/pull/9) would be a simpler way of achieving what the current tests are doing.

pp-mo:
Thanks @dkillick 
Over to you again..

dkillick:
Thanks for all your input @pp-mo! :tada:

lib/iris/_lazy_data.py
```diff
@@ -23,11 +23,37 @@
...
+        not change user-specified options that have already been set.
+
+    """
+    if 'pool' not in dask.context._globals and \
+            'get' not in dask.context._globals:
+        dask.set_options(get=dask.async.get_sync)
+
+
+# Run this at import time to set dask options for Iris.
+_iris_dask_defaults()
```
bjlittle: @dkillick and @pp-mo I'm assuming that the user can simply override this default behaviour by using `dask.set_options()` themselves? ...

dkillick: Yup! That's the idea. You could do something like the following...

```python
from multiprocessing.pool import ThreadPool

import dask
import iris  # This will set default options for dask processing in Iris.

dask.set_options(pool=ThreadPool(5))  # Now the user can apply case-specific options.
```

bjlittle: Cool. As long as the user can override the default.

I love the `async` default. No surprises or wiggle room for odd behaviour :+1: 

Also, and most importantly for me, if the user does override the default, then **they** are taking explicit accountability for their actions ... :boom:

bjlittle: Nice job @dkillick and @pp-mo :smile:

dkillick: By the way, I'll make sure examples like the above are included in the user documentation when we come around to writing it...

bjlittle: @dkillick If it helps, add a note in the [Dask Documentation](https://github.com/SciTools/iris/projects/6) project, so that we don't forget :+1:

PR #2513, opened 2017-04-26
DataManager used for cell measures

pp-mo:
TODO:
 * [x] check error when creating cell measures with lazy integer data
 * [x] check error when saving masked cell measures data to netcdf

### DIFF
--- a/lib/iris/coords.py
+++ b/lib/iris/coords.py
@@ -1741,9 +1741,12 @@ class CellMeasure(six.with_metaclass(ABCMeta, CFVariableMixin)):
         #: have no meaning to Iris.
         self.attributes = attributes
 
-        self.data = data
-
+        #: String naming the measure type.
         self.measure = measure
+
+        # Initialise data via the data setter code, which applies standard
+        # checks and ajustments.
+        self.data = data
 
     @property
     def measure(self):
@@ -1761,13 +1764,13 @@ class CellMeasure(six.with_metaclass(ABCMeta, CFVariableMixin)):
         if data is None:
             raise ValueError('The data payload of a CellMeasure may not be '
                              'None; it must be a numpy array or equivalent.')
-        if is_lazy_data(data) and data.dtype.kind in 'iu':
-            # Disallow lazy integer data.  Integer values would be very
-            # unexpected anyway, but this avoids the *even more obscure*
-            # possibility that there are masked values, because masked integer
-            # data would cause dask to return data with the wrong dtype.
-            # This can all be fixed, as for cube data, by adding fill_value and
-            # dtype support, but for now we prefer just not to support this.
+        if is_lazy_data(data) and data.dtype.kind in 'biu':
+            # Disallow lazy integral data, as it will cause problems with dask
+            # if it turns out to contain any masked points.
+            # Non-floating cell measures are not valid up to CF v1.7 anyway,
+            # but this avoids any possible problems with non-compliant files.
+            # Future usage could be supported by adding a fill_value and dtype
+            # as for cube data.  For now, disallowing it is just simpler.
             msg = ('Cannot create cell measure with lazy data of type {}, as '
                    'integer types are not currently supported.')
             raise ValueError(msg.format(data.dtype))
@@ -1785,17 +1788,12 @@ class CellMeasure(six.with_metaclass(ABCMeta, CFVariableMixin)):
 
     @property
     def shape(self):
-        """The fundamental shape of the Cell Measure, expressed as a tuple."""
-        # Access the underlying _data_manager attribute to avoid triggering
-        # a deferred load unnecessarily.
+        """Returns the shape of the Cell Measure, expressed as a tuple."""
         return self._data_manager.shape
 
     @property
     def ndim(self):
-        """
-        Return the number of dimensions of the cell measure.
-
-        """
+        """Returns the number of dimensions of the cell measure."""
         return self._data_manager.ndim
 
     @measure.setter
@@ -1832,8 +1830,8 @@ class CellMeasure(six.with_metaclass(ABCMeta, CFVariableMixin)):
                         raise IndexError('Cannot index with zero length '
                                          'slice.')
 
-        new_cell_measure = self.copy(data=data)
-        return new_cell_measure
+        # The result is a copy with replacement data.
+        return self.copy(data=data)
 
     def copy(self, data=None):
         """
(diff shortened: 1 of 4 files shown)


lib/iris/coords.py
```diff
@@ -1762,30 +1761,42 @@ def data(self, data):
         if data is None:
             raise ValueError('The data payload of a CellMeasure may not be '
                              'None; it must be a numpy array or equivalent.')
+        if is_lazy_data(data) and data.dtype.kind in 'iu':
```
dkillick: Let's stretch a crazy thing a little further: do you think we might need to cope with boolean cell measures? We have already discussed (offline) the concept of a cell measure containing only 1 or 0, which is effectively a boolean cell measure...

lib/iris/coords.py
```diff
@@ -1762,30 +1761,42 @@ def data(self, data):
...
+            # Disallow lazy integer data.  Integer values would be very
+            # unexpected anyway, but this avoids the *even more obscure*
+            # possibility that there are masked values, because masked integer
+            # data would cause dask to return data with the wrong dtype.
+            # This can all be fixed, as for cube data, by adding fill_value and
+            # dtype support, but for now we prefer just not to support this.
```
dkillick: And we can add this as a feature request at a later date once we have a concrete example of this being needed :+1:

lib/iris/fileformats/netcdf.py
```diff
@@ -1494,9 +1494,18 @@ def _create_cf_cell_measure_variable(self, cube, dimension_names,
...
                          cube.cell_measure_dims(cell_measure)]
 
+        # Get the data values
+        data = cell_measure.data
+
+        # Disallow saving of *masked* cell measures.
+        if ma.is_masked(data):
+            # We can't really save properly if there are masked points, because
+            # we don't (yet) record an appropriate fill_value property.
+            msg = "Currently cannot save cell measures with missing data."
```
dkillick: I'd drop the "currently" -- we _just don't support this_, and that's just how it is unless someone asks for it.

lib/iris/fileformats/netcdf.py
```diff
@@ -1494,9 +1494,18 @@ def _create_cf_cell_measure_variable(self, cube, dimension_names,
         cf_dimensions = [dimension_names[dim] for dim in
                          cube.cell_measure_dims(cell_measure)]
 
+        # Get the data values
```
dkillick: Strictly you should add a full stop at the end of this comment.

lib/iris/coords.py
```diff
@@ -1762,30 +1761,42 @@ def data(self, data):
...
             data = np.array(data, ndmin=1)
-        if hasattr(self, '_data') and self._data is not None:
+        if hasattr(self, '_data_manager') and self._data_manager is not None:
             # Check that setting these data wouldn't change self.shape
             if data.shape != self.shape:
                 raise ValueError("New data shape must match existing data "
                                  "shape.")
 
-        self._data = data
+        self._data_manager = DataManager(data)
```
dkillick: Do you think there's value in a comment here explaining why we don't create a `DataManager` until this point? Might not be clear to someone fresh to the code.

lib/iris/coords.py
```diff
@@ -1762,30 +1761,42 @@ def data(self, data):
...
 
     @property
     def shape(self):
         """The fundamental shape of the Cell Measure, expressed as a tuple."""
-        # Access the underlying _data attribute to avoid triggering
+        # Access the underlying _data_manager attribute to avoid triggering
```
dkillick: We could probably drop this comment entirely now, right?

lib/iris/coords.py
pp-mo: This is still all out of scope as-of CF v1.7, but here we are covering ourselves for whatever might exist in a file, 
So yes, I'll add that as another.

lib/iris/coords.py
pp-mo: I think I'll rewrite to make that clearer...

lib/iris/coords.py
pp-mo: I'll add comments in all of `__init__`, `copy` and `__getitem__` to make clear how it all fits together.

lib/iris/coords.py
pp-mo: Yes

pp-mo:
> e6a8045 "Review changes: clearer comments."

Hopefully addresses all you asked for, plus a couple of other improvements.

pp-mo:
Still to-do : https://github.com/pp-mo/iris/pull/43

dkillick:
@pp-mo I'm :+1: with the changes to this point. Now all we need to do is integrate the outstanding testing and I think we're there.

pp-mo:
> all we need to do is integrate the outstanding testing 

Now done.
Ready to go ?

lib/iris/tests/unit/coords/test_CellMeasure.py
```diff
@@ -100,6 +100,14 @@ def test___getitem__(self):
         sub_measure = self.measure[2]
         self.assertArrayEqual(self.values[2], sub_measure.data)
 
+    def test___getitem__data_copy(self):
+        # Check that a sliced cell measure has independent data.
+        sub_measure = self.measure[1:3]
+        old_values = sub_measure.data.copy()
+        self.measure.data[:] = 0.0
+        new_values = sub_measure.data.copy()
```
dkillick: Is this second copy call necessary?

lib/iris/coords.py
```diff
@@ -1813,11 +1813,16 @@ def __getitem__(self, key):
         # each dimension of the cell_measure.
         full_slice = iris.util._build_full_slice_given_keys(key, self.ndim)
 
+        # Get the data, all or part of which will become the new data.
+        data = self._data_manager.core_data.copy()
+        # Copy the data to avoid making the new measure a view on the old one.
+        data = data.copy()
```
dkillick: You've now copied this variable _twice_. That seems like a lot of copying...

dkillick:
@pp-mo that will do nicely, thanks!

lib/iris/coords.py
```diff
@@ -1804,18 +1813,19 @@ def __getitem__(self, key):
         # each dimension of the cell_measure.
         full_slice = iris.util._build_full_slice_given_keys(key, self.ndim)
 
+        # Get the data, all or part of which will become the new data.
+        data = self._data_manager.core_data
+        # Copy the data to avoid making the new measure a view on the old one.
+        data = data.copy()
```
pp-mo: This line added to prevent views.

Issue #2490, opened 2017-04-19
Integrate DataManager with coordinates (8 man days)


bjlittle:

dkillick:
A couple of gotchas to consider while integrating the DataManager.

1. Printing an `AuxCoord` realises a lazy points array. What's hard to determine is whether this is desirable behaviour or not: on the one hand, if I print a coord I want to see what its points are; but on the other hand a print operation should not affect the underlying state of the object being printed! For example: 
```python
>>> ldc = iris.coords.AuxCoord(as_lazy_data(np.array([1, 2, 3]), units='m')
>>> is_lazy_data(ldc._points)
True
>>> print ldc
AuxCoord(array([1, 2, 3]), standard_name=None, units=Unit('m'))
>>> is_lazy_data(ldc._points)
False
```
2. A `DimCoord` does not preserve a lazy points array **at all**. It seems that we have [this line](https://github.com/SciTools/iris/blob/dask/lib/iris/coords.py#L1504) to thank for this. For example:
```python
>>> dldc = iris.coords.DimCoord(as_lazy_data(np.array([1, 2, 3]), standard_name='height', units='m')
>>> is_lazy_data(dldc._points)
False
```

pp-mo:
My views on these (for the record, already discussed offline)
> 1. Printing an AuxCoord realises a lazy points array
> ... a print operation should not affect the underlying state of the object 

I think it's acceptable.
Our whole conception of 'lazy' data is that it is transient and the "real" content of a Cube (or Coord, now) is the "real" data.  Thus, when you need the data you realise it and "the lazy" is gone, but the essential object content is _unchanged_, because that data "is" the true state of the object.
Thus, the 'lazy' aspect has to be seen as a kind-of internal concept -- even though it is not actually hidden or private.

> 2. A DimCoord does not preserve a lazy points array at all.

Given the monotonic requirement, I think this it is simply unavoidable.
We could instead load the data, make the test, throw it away + keep it lazy.
However, I'm really not in favour of that, as it is against the spirit of "touch once and it's gone" which is a nice straightforward model of the behaviour.

dkillick:
@pp-mo some helpful views, thanks! I think we might just need to be careful about when we print coords if we need (for some reason) to maintain lazy coord data...

dkillick:
For reference, the PR for this is #2527.

dkillick:
This is done now that #2527 is merged!

PR #2517, opened 2017-04-28
Prevent sliced coords sharing data with their parent coord.

pp-mo:
While considering #2490 I noticed that a sliced Coord (an AuxCoord, at least) can be linked to the original.
This is exactly what we found when working on #2513, and fixed in [this line in that PR](https://github.com/SciTools/iris/pull/2513/files#r113925900)

I think this is a simple bug, as concluded for the CellMeasures.
Obviously though, a behaviour change in coords could be a bit more dangerous.
There were some very old tests that check the exact opposite behaviour (i.e. indexed coords share data with their originals), so I got rid of those.
Otherwise, it didn't break any existing tests ! :crossed_fingers: 

Note that this modified `Coord.__getitem__` code will simplify nicely when we get rid of `LazyArray`s.
Currently we must still support LazyArray-s for points and bounds data, as aux factories are still using them.

### DIFF
--- a/lib/iris/coords.py
+++ b/lib/iris/coords.py
@@ -464,38 +464,40 @@ class Coord(six.with_metaclass(ABCMeta, CFVariableMixin)):
         # each dimension of the coord.
         full_slice = iris.util._build_full_slice_given_keys(key, self.ndim)
 
-        # If it's a "null" indexing operation (e.g. coord[:, :]) then
-        # we can preserve deferred loading by avoiding promoting _points
-        # and _bounds to full ndarray instances.
+        # Fetch the points and bounds.
+        points = self._points
+        bounds = self._bounds
+
+        # Track whether we need to copy data after any indexing operation.
+        # NOTE: when we get rid of LazyArray-s, we can remove this bit.
+        points_copied = False
+        bounds_copied = False
+
+        # If it's a "null" indexing operation (e.g. coord[:, :]) then we can
+        # avoid all the indexing.  This is desirable if we have LazyArray-s,
+        # as they must be fetched first (you can't index them).
         def is_full_slice(s):
             return isinstance(s, slice) and s == slice(None, None)
-        if all(is_full_slice(s) for s in full_slice):
-            points = self._points
-            bounds = self._bounds
-            # Take copies, to avoid making coords that are views on other ones.
-            if not isinstance(points, iris.aux_factory._LazyArray):
-                points = points.copy()
+        null_operation = all(is_full_slice(s) for s in full_slice)
+
+        if null_operation:
+            # If we have Lazy Arrays, don't copy but just allow the new coord
+            # to share the same LazyArray object.
+            if isinstance(points, iris.aux_factory._LazyArray):
+                points_copied = True
             if bounds is not None:
-                if not isinstance(bounds, iris.aux_factory._LazyArray):
-                    bounds = bounds.copy()
+                if isinstance(bounds, iris.aux_factory._LazyArray):
+                    bounds_copied = True
         else:
-            points = self._points
             if isinstance(points, iris.aux_factory._LazyArray):
                 # This triggers the LazyArray to compute its values
                 # (if it hasn't already), which will also trigger any
                 # deferred loading of its dependencies.
                 points = points.view()
-            else:
-                # Take a copy to avoid making coords that are views on other
-                # coords.  This will not realise lazy data.
-                points = points.copy()
-            bounds = self._bounds
+                points_copied = True
             if isinstance(bounds, iris.aux_factory._LazyArray):
                 bounds = bounds.view()
-            elif bounds is not None:
-                # Take a copy to avoid making coords that are views on other
-                # coords.  This will not realise lazy data.
-                bounds = bounds.copy()
+                bounds_copied = True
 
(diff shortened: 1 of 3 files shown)


pp-mo:
Assuming we are happy with this, it should help us to unify the `__getitem__` code in Cube, Coord and CellMeasure.

All those methods use a similar indexing approach based on the iris.util functions `_build_full_slice_given_keys` and `column_slices_generator`.
I think we can unify those approaches and centralise the code in the DataManager class, which would be very neat.
At present, the need for `Coord.__getitem__` to support LazyArray gets in the way of that a bit.

lib/iris/coords.py
```diff
@@ -497,6 +513,14 @@ def is_full_slice(s):
                 if bounds is not None:
                     bounds = bounds[keys + (Ellipsis, )]
 
+        # Copy data after indexing, if needed, to avoid making coords that are
+        # views on other coords.  This will not realise lazy data.
+        if not points_copied:
+            points = points.copy()
```
dkillick: @pp-mo doesn't this mean we end up effectively copying twice when combined with [L552](https://github.com/SciTools/iris/pull/2517/files#diff-fae2e3bb7827f6df8671256d1c779951R552)?

lib/iris/coords.py
```diff
@@ -497,6 +513,14 @@ def is_full_slice(s):
...
+        if bounds is not None and not bounds_copied:
+            bounds = bounds.copy()
```
dkillick: As above.

lib/iris/tests/unit/coords/test_Coord.py
```diff
@@ -368,5 +368,21 @@ def test_writable_points(self):
             coord2.bounds[:] = 0
 
 
+class Test_AuxCoord__getitem__(tests.IrisTest):
+    def test_partial_slice_data_copy(self):
+        original_coord = AuxCoord([1., 2., 3.])
+        sub_coord = original_coord[:1]
+        original_values = sub_coord.points.copy()
```
dkillick: I've no problem with the code of the tests :+1: However, I found the variable names a little confusing as you've sort of overloaded the word 'original'.

It refers to both:

* the coord you start with ("`original_coord`"), and
* the initial values of the indexed coord ("`original_values`").

Could you perhaps rename `original_values` to something like `initial_values` to make it clearer that there's no direct connection between it and `original_coord`?

lib/iris/tests/unit/coords/test_Coord.py
```diff
@@ -368,5 +368,21 @@ def test_writable_points(self):
...
+        original_coord.points[:] = -999.9
+        self.assertArrayEqual(sub_coord.points, original_values)
+
+    def test_full_slice_data_copy(self):
+        original_coord = AuxCoord([1., 2., 3.])
+        sub_coord = original_coord[:]
+        original_values = sub_coord.points.copy()
```
dkillick: The naming thing again.

lib/iris/coords.py
```diff
@@ -464,24 +464,40 @@ def __getitem__(self, key):
...
+        null_operation = all(is_full_slice(s) for s in full_slice)
+
+        if null_operation:
+            # If we have Lazy Arrays, don't copy but just allow the new coord
+            # to share the same LazyArray object.
+            if isinstance(points, iris.aux_factory._LazyArray):
+                points_copied = True
+            if bounds is not None:
+                if isinstance(bounds, iris.aux_factory._LazyArray):
+                    bounds_copied = True
```
lbdreyer: I don't know how fussy you want to be about naming, but to me this `bounds_copied` says that the action has been done. `copy_bounds = True` makes more sense to me

lbdreyer:
I am happy with the changes (assuming you address [Pete's comment](https://github.com/SciTools/iris/pull/2517#discussion_r114291305))
Although I'm not convinced you gained much splitting this out from the work on integrating DataManager with coords.

lib/iris/coords.py
pp-mo: > "copy_bounds = True makes more sense"
Well, kind of.. Actually, what it really means is more like "still_need_to_copy_bounds = False" !
I will review this...

lib/iris/tests/unit/coords/test_Coord.py
pp-mo: Hmm good point ...
Will review..

pp-mo:
Thanks Laura

> I'm not convinced you gained much splitting this out

It may not seem much, but I think it will make the next changes much easier to follow.

lib/iris/coords.py
pp-mo: No, because this is making copies of the new derived arrays, that will be passed as the new data to Coord.copy().
Meanwhile L552 is copying is the *old* (full) data -- which in this case will just get replaced (!!)
That is indeed a potential problem, but the existing code is already doing it :  As I've managed not to touch that code in this PR, I don't want to address it now.

pp-mo:
@dkillick @lbdreyer Changes up.
Does that look like improvement, is it good enough ?

dkillick:
@pp-mo looks good - if the tests are happy then so am I :+1:

dkillick:
Bingo. In it goes!

Issue #2495, opened 2017-04-20
Integrate DataManager with auxiliary coordinate factory (derived coordinates) [5 man days]


bjlittle:

dkillick:
A majority of the work detailed above is done by https://github.com/SciTools/iris/pull/2527/commits/2075f75056bb68f680fafa83d442dc582dd7effd.

dkillick:
To update [my previous comment](https://github.com/SciTools/iris/issues/2495#issuecomment-298954745):

* we now use `coord.core_points()` to ensure we use the incoming coords' lazy points, if available.
* it looks like there was never support for lazy bounds before, however we could [add that now](https://github.com/SciTools/iris/blob/master/lib/iris/aux_factory.py#L300).
* turns out there's [more test coverage](https://github.com/SciTools/iris/tree/master/lib/iris/tests/unit/aux_factory) than I first thought, including checks for [lazy points being maintained](https://github.com/SciTools/iris/blob/dask/lib/iris/tests/unit/aux_factory/test_AuxCoordFactory.py#L65-L66).
* this was done by https://github.com/SciTools/iris/pull/2527/commits/2075f75056bb68f680fafa83d442dc582dd7effd.

pp-mo:
I've looked over all the the files that mention factories + have changed on the dask branch.
There is no tricky usage that I can see -- the changes for dask should not cause any problems.

dkillick:
Thanks for your input @pp-mo -- closing this in light of recent comments.

PR #2518, opened 2017-05-02
Remove `LazyArray` class from coords

dkillick:
In the past the coord `LazyArray` class was used to defer loading of all coords, but now it is only used to defer calculation of derived coordinates. However, there was still some `LazyArray` logic hanging around in the wider coords codebase, which was getting in the way of integrating the `DataManager` with coords. We have decided to remove the `LazyArray` class entirely to help this all along.

Note that doing this does not prevent derived coords from being calculated, but it does mean that the calculation is (currently) not deferred. When we come to [integrating derived coords with the `DataManager`](https://github.com/SciTools/iris/issues/2495) we can look to correcting this change, if by using dask deferral rather than by using the `LazyArray` class.

### DIFF
--- a/lib/iris/aux_factory.py
+++ b/lib/iris/aux_factory.py
@@ -1,4 +1,4 @@
-# (C) British Crown Copyright 2010 - 2016, Met Office
+# (C) British Crown Copyright 2010 - 2017, Met Office
 #
 # This file is part of Iris.
 #
--- a/lib/iris/aux_factory.py
+++ b/lib/iris/aux_factory.py
@@ -25,121 +25,12 @@ import six
 
 from abc import ABCMeta, abstractmethod, abstractproperty
 import warnings
-import zlib
 
-import cf_units
 import numpy as np
 
-from iris._deprecation import warn_deprecated
 from iris._cube_coord_common import CFVariableMixin
 import iris.coords
 import iris.util
-
-
-class _LazyArray(object):
-    """
-    Represents a simplified NumPy array which is only computed on demand.
-
-    It provides the :meth:`view()` and :meth:`reshape()` methods so it
-    can be used in place of a standard NumPy array under some
-    circumstances.
-
-    The first use of either of these methods causes the array to be
-    computed and cached for any subsequent access.
-
-    """
-    def __init__(self, shape, func, dtype=None):
-        """
-        Args:
-
-        * shape (tuple):
-            The shape of the array which will be created.
-        * func:
-            The function which will be called to supply the real array.
-
-        Kwargs:
-
-        * dtype (np.dtype):
-            The numpy dtype of the array which will be created.
-            Defaults to None to signify the dtype is unknown.
-
-        """
-        self.shape = tuple(shape)
-        self._func = func
-        self.dtype = dtype
-        self._array = None
-
-    def __repr__(self):
-        return '<LazyArray(shape={}, dtype={!r})>'.format(self.shape,
-                                                          self.dtype)
-
-    def _cached_array(self):
-        if self._array is None:
-            self._array = np.asarray(self._func())
-            del self._func
-        return self._array
-
-    def reshape(self, *args, **kwargs):
-        """
-        Returns a view of this array with the given shape.
-
-        See :meth:`numpy.ndarray.reshape()` for argument details.
-
-        """
-        return self._cached_array().reshape(*args, **kwargs)
-
-    def to_xml_attr(self):
-        """
-        Returns a string describing this array, suitable for use in CML.
-
-        """
-        crc = zlib.crc32(np.array(self._cached_array(), order='C'))
-        crc &= 0xffffffff
-        return 'LazyArray(shape={}, checksum=0x{:08x})'.format(self.shape, crc)
-
-    def view(self, *args, **kwargs):
-        """
-        Returns a view of this array.
-
-        See :meth:`numpy.ndarray.view()` for argument details.
-
-        """
-        return self._cached_array().view(*args, **kwargs)
-
-
-class LazyArray(_LazyArray):
-    """
-    Represents a simplified NumPy array which is only computed on demand.
-
-    It provides the :meth:`view()` and :meth:`reshape()` methods so it
-    can be used in place of a standard NumPy array under some
-    circumstances.
-
-    The first use of either of these methods causes the array to be
-    computed and cached for any subsequent access.
-
-    .. deprecated:: 1.9
-
-    """
-    def __init__(self, shape, func, dtype=None):
-        """
-        Args:
-
-        * shape (tuple):
-            The shape of the array which will be created.
-        * func:
-            The function which will be called to supply the real array.
-
-        Kwargs:
-
-        * dtype (np.dtype):
-            The numpy dtype of the array which will be created.
-            Defaults to None to signify the dtype is unknown.
-
-        """
-        warn_deprecated('LazyArray is deprecated and will be removed '
-                        'in a future release.', stacklevel=2)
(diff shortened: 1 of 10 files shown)


dkillick:
The four tests that were failing all expected that coords will be lazy. In the long term we want to maintain this behaviour, so I've introduced a temporary skip for these tests that can be removed when we reintroduce lazy coords via `dask` / `DataManager`.

lib/iris/aux_factory.py
```diff
@@ -1734,23 +1565,17 @@ def make_coord(self, coord_dims_func):
...
+                              nd_points_by_key['depth'],
+                              nd_points_by_key['depth_c'])
 
         bounds = None
         if self.s.nbounds or (self.c and self.c.nbounds):
-            # Build a "lazy" bounds array.
+            # Build the bounds array.
             nd_values_by_key = self._remap_with_bounds(dependency_dims,
                                                        derived_dims)
             bounds_shape = self._shape(nd_values_by_key)
```
lbdreyer: It doesn't look like this gets used.

lib/iris/coords.py
```diff
@@ -460,29 +460,16 @@ def __getitem__(self, key):
             indexing.
 
         """
+        points = self._points
+        bounds = self._bounds
+
```
lbdreyer: Aren't these changes also in #2517 ? Why aren't there merge conflicts?

lbdreyer: Okay there are actually merge conflicts, so this'll just need to be rebased.

lib/iris/aux_factory.py
```diff
@@ -566,9 +449,7 @@ def calc_bounds():
                     orography = orography_pts.reshape(
                         orography_pts_shape.append(1))
                 return self._derive(delta, sigma, orography)
-            b_shape = self._shape(nd_values_by_key)
-            b_dtype = self._dtype(nd_values_by_key)
-            bounds = _LazyArray(b_shape, calc_bounds, b_dtype)
+            bounds = calc_bounds()
```
pp-mo: I think it would be clearer here to ditch the `calc_bounds` function and use plain inline code, exactly as you did for `calc_points`.

lib/iris/aux_factory.py
```diff
@@ -767,9 +640,7 @@ def calc_bounds():
                     surface_air_pressure = surface_air_pressure_pts.reshape(
                         surface_air_pressure_pts_shape.append(1))
                 return self._derive(delta, sigma, surface_air_pressure)
-            b_shape = self._shape(nd_values_by_key)
-            b_dtype = self._dtype(nd_values_by_key)
-            bounds = _LazyArray(b_shape, calc_bounds, b_dtype)
+            bounds = calc_bounds()
```
pp-mo: Another vote for losing `calc_bounds`.

lib/iris/aux_factory.py
```diff
@@ -1004,12 +867,10 @@ def calc_bounds():
                                     nd_values_by_key['eta'],
                                     nd_values_by_key['depth'],
                                     nd_values_by_key['depth_c'],
-                                    nd_values_by_key['nsigma'],
                                     nd_values_by_key['zlev'],
                                     bounds_shape,
                                     nsigma_slice_bounds)
-
-            bounds = _LazyArray(bounds_shape, calc_bounds, bounds_dtype)
+            bounds = calc_bounds()
```
pp-mo: Another vote for losing `calc_bounds`.

lib/iris/aux_factory.py
dkillick: Whoops, that one must have slipped through. Good spot :+1:

lib/iris/coords.py
dkillick: Yeah, too many people all working in the same place...

lib/iris/aux_factory.py
dkillick: @pp-mo I chose not to because the bounds calculation code is much longer and it was simpler to just leave the inline function and call it.

dkillick: OK, maybe it isn't so bad pulling the code inline, but it still makes for a very long `make_coord` method. I wonder whether it would be better to pull the bounds calculation code up into a new method in each case, or whether to just leave that problem for the future. @pp-mo @lbdreyer thoughts?

dkillick:
I've rebased to sort out conflicts following the merge of #2517 and removed the extraneous line of code that @lbdreyer pointed out. 

Having discussed the bounds calculation enclosed function `calc_bounds` offline with @pp-mo we've agreed to keep things as they are. We decided this approach would make future work easier through making it clearer the extent of the bounds calculation functionality for each factory class.

lib/iris/aux_factory.py
pp-mo: P.S. (following offline discussion)
Changed my mind! We could instead leave all these one-use functions in place, to emphasize the common structure of these classes in support of one day simplifying + generalising..

lbdreyer:
Travis isn't running! Do you know why?

lib/iris/aux_factory.py
```diff
@@ -957,27 +827,20 @@ def make_coord(self, coord_dims_func):
...
+                              points_shape,
+                              nsigma_slice)
 
         bounds = None
         if self.zlev.nbounds or (self.sigma and self.sigma.nbounds):
-            # Build a "lazy" bounds array.
+            # Build the bounds array.
             nd_values_by_key = self._remap_with_bounds(dependency_dims,
                                                        derived_dims)
             bounds_shape = self._shape(nd_values_by_key)
```
pp-mo: This line is still redundant ...

lib/iris/tests/unit/fileformats/pyke_rules/compiled_krb/fc_rules_cf_fc/test_reorder_bounds_data.py
```diff
@@ -56,19 +55,6 @@ def test_slowest_varying(self):
         expected = np.rollaxis(bounds_data, 0, bounds_data.ndim)
         self.assertArrayEqual(res, expected)
 
-    def test_slowest_varying_lazy(self):
```
lbdreyer: Why did you choose to just remove this test rather than skip it like the others?

lib/iris/aux_factory.py
dkillick: It actually isn't... this is passed to `self._derive` on L871...

lib/iris/aux_factory.py
```diff
@@ -25,123 +25,14 @@
...
 import warnings
-import zlib
 
-import cf_units
 import numpy as np
 
-from iris._deprecation import warn_deprecated
 from iris._cube_coord_common import CFVariableMixin
 import iris.coords
 import iris.util
```
pp-mo: This import is now obsolete.

lib/iris/tests/unit/fileformats/pyke_rules/compiled_krb/fc_rules_cf_fc/test_reorder_bounds_data.py
dkillick: This test looked like it tested part of the implementation of `LazyArray`, rather than testing the general lazy coords concept. If you disagree or would like to retain the test then I can skip it instead.

lbdreyer: It looks like the only places `reorder_bounds_data` is used is for building dimension coordinates and auxiliary coordinates at load time of netcdf files. Dimension coordinates doesn't use lazy data at this point and auxiliary coordinates uses a dask array.

PR #2519, opened 2017-05-02
Unify getitem code between Cube, Coord and CellMeasures.

pp-mo:
This is a useful preliminary to replacing lazy/real array data with DataManager in Coords code.

It reduces the indexing code in those 3 areas to suspiciously similar usages, with the bulk of the common code in a new private utility routine `iris.utils._slice_data_with_keys` :  I think that code can usefully go inside `DataManager.__getitem__` in future.

In future, in addition to rebasing all these 3 areas onto DataManager, I would also move the coord points and bounds accessors into `Coord` as generic methods, and make DimCoord extend the generic calls with its extra checks and tweaks.  The new similarities of handling make this attractive, but it is better left undone because (a) it makes this PR simpler and (b) it will be simpler when the underlying data objects are all DataManager-s.

This PR also contains a strictly temporary hack to remove the special-casing of LazyArray data within AuxCoords, by wrapping them as dask arrays instead (`WrappedLazyArray` class and `_wrap_lazy` function in coords.py).  
This should be replaced by the result of https://github.com/SciTools/iris/pull/2518 when that is ready.

### DIFF
--- a/lib/iris/aux_factory.py
+++ b/lib/iris/aux_factory.py
@@ -195,7 +195,7 @@ class AuxCoordFactory(six.with_metaclass(ABCMeta, CFVariableMixin)):
         nd_shape = [1] * ndim + [coord.nbounds]
         for dim, size in zip(dims, coord.shape):
             nd_shape[dim] = size
-        bounds.shape = tuple(nd_shape)
+        bounds = bounds.reshape(nd_shape)
         return bounds
 
     @staticmethod
@@ -297,7 +297,7 @@ class AuxCoordFactory(six.with_metaclass(ABCMeta, CFVariableMixin)):
                     # we just add an extra 1.
                     shape.append(1)
                 nd_values = np.array(nd_values)
-                nd_values.shape = shape
+                nd_values = nd_values.reshape(shape)
             else:
                 # If no coord, treat value as zero.
                 # Use a float16 to provide `shape` attribute and avoid
--- a/lib/iris/coords.py
+++ b/lib/iris/coords.py
@@ -893,8 +893,8 @@ class Coord(six.with_metaclass(ABCMeta, CFVariableMixin)):
 
         """
         nbounds = 0
-        if self.bounds is not None:
-            nbounds = self.bounds.shape[-1]
+        if self._bounds is not None:
+            nbounds = self._bounds.shape[-1]
         return nbounds
 
     def has_bounds(self):
--- a/lib/iris/coords.py
+++ b/lib/iris/coords.py
@@ -448,7 +448,7 @@ class Coord(six.with_metaclass(ABCMeta, CFVariableMixin)):
         self.points = points
         self.bounds = bounds
 
-    def __getitem__(self, key):
+    def __getitem__(self, keys):
         """
         Returns a new Coord whose values are obtained by conventional array
         indexing.
@@ -460,34 +460,19 @@ class Coord(six.with_metaclass(ABCMeta, CFVariableMixin)):
             indexing.
 
         """
-        # Turn the key(s) into a full slice spec - i.e. one entry for
-        # each dimension of the coord.
-        full_slice = iris.util._build_full_slice_given_keys(key, self.ndim)
-
         # Fetch the points and bounds.
         points = self._points
         bounds = self._bounds
 
-        def is_full_slice(s):
-            return isinstance(s, slice) and s == slice(None, None)
+        # Index both points and bounds with the keys.
+        _, points = iris.util._slice_data_with_keys(
+            points, keys, self.ndim)
+        if bounds is not None:
+            _, bounds = iris.util._slice_data_with_keys(
+                bounds, keys, self.ndim + 1)
 
-        if not all(is_full_slice(s) for s in full_slice):
-            # Make indexing on the cube column based by using the
-            # column_slices_generator (potentially requires slicing the
-            # data multiple times).
-            _, slice_gen = iris.util.column_slices_generator(full_slice,
-                                                             self.ndim)
-            for keys in slice_gen:
-                if points is not None:
-                    points = points[keys]
-                    if points.shape and min(points.shape) == 0:
-                        raise IndexError('Cannot index with zero length '
-                                         'slice.')
-                if bounds is not None:
-                    bounds = bounds[keys + (Ellipsis, )]
-
-        # Copy data after indexing to avoid making coords that are
-        # views on other coords. This will not realise lazy data.
+        # Copy data after indexing, to avoid making coords that are
+        # views on other coords.  This will not realise lazy data.
         points = points.copy()
         if bounds is not None:
             bounds = bounds.copy()
@@ -1496,7 +1481,7 @@ class DimCoord(Coord):
     @property
     def points(self):
         """The local points values as a read-only NumPy array."""
-        points = self._points.view()
+        points = self._points
         return points
 
     @points.setter
@@ -1530,7 +1515,7 @@ class DimCoord(Coord):
         """
         bounds = None
         if self._bounds is not None:
-            bounds = self._bounds.view()
+            bounds = self._bounds
         return bounds
 
     @bounds.setter
@@ -1600,9 +1585,6 @@ class AuxCoord(Coord):
         # Ensure the array has enough dimensions.
         # NB. Returns the *same object* if result.ndim >= ndmin
         result = np.array(result, ndmin=ndmin, copy=False)
-        # We don't need to copy the data, but we do need to have our
-        # own view so we can control the shape, etc.
-        result = result.view()
         return result
 
     @property
@@ -1614,7 +1596,7 @@ class AuxCoord(Coord):
             # NOTE: we probably don't have full support for masked aux-coords.
             # We certainly *don't* handle a _FillValue attribute (and possibly
             # the loader will throw one away ?)
-        return self._points.view()
+        return self._points
 
     @points.setter
     def points(self, points):
@@ -1656,7 +1638,7 @@ class AuxCoord(Coord):
                 # We certainly *don't* handle a _FillValue attribute (and
                 # possibly the loader will throw one away ?)
                 self._bounds = bounds
-            bounds = bounds.view()
+            bounds = bounds
         else:
             bounds = None
 
@@ -1797,37 +1779,22 @@ class CellMeasure(six.with_metaclass(ABCMeta, CFVariableMixin)):
                              "not {}".format(measure))
         self._measure = measure
 
-    def __getitem__(self, key):
+    def __getitem__(self, keys):
         """
         Returns a new CellMeasure whose values are obtained by
         conventional array indexing.
 
         """
-        # Turn the key(s) into a full slice spec - i.e. one entry for
-        # each dimension of the cell_measure.
-        full_slice = iris.util._build_full_slice_given_keys(key, self.ndim)
-
         # Get the data, all or part of which will become the new data.
         data = self._data_manager.core_data()
-        # Copy the data to avoid making the new measure a view on the old one.
+
+        # Index data with the keys.
+        # Note: does not copy data unless it has to.
+        _, data = iris.util._slice_data_with_keys(data, keys, self.ndim)
+
+        # Always copy data, to avoid making the new measure a view onto the old
+        # one.
         data = data.copy()
-
-        # If it's a "null" indexing operation (e.g. cell_measure[:, :]) then
-        # we can skip the indexing part.
-        def is_full_slice(s):
-            return isinstance(s, slice) and s == slice(None, None)
-
-        if not all(is_full_slice(s) for s in full_slice):
-            # Slice on each column, using `iris.util.column_slices_generator`
-            # (potentially slicing the data multiple times).
-            _, slice_gen = iris.util.column_slices_generator(full_slice,
-                                                             self.ndim)
-            for keys in slice_gen:
-                if data is not None:
-                    data = data[keys]
-                    if data.shape and min(data.shape) == 0:
-                        raise IndexError('Cannot index with zero length '
-                                         'slice.')
 
         # The result is a copy with replacement data.
         return self.copy(data=data)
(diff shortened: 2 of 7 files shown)


pp-mo:
@bjlittle @dkillick @lbdreyer Please check this out !

dkillick:
> the bulk of the common code in a new private utility routine `iris.utils._slice_data_with_keys` : I think that code can usefully go inside `DataManager.__getitem__` in future.

Out of interest, why did you choose to add this new private utils routine rather than add a `__getitem__` to the `DataManager` class in the first instance?

dkillick:
> why did you choose to add this new private utils routine ... ?

Oh, is the answer "because the routine already existed"? And, as such, was a useful place to host extended functionality...

pp-mo:
> why did you choose to add this new private utils routine 

I just needed a place for common functionality, but I'm expecting it to go into DataManager.
I guess we'll leave the two existing iris.util routines in place as status quo is simplest, though not exactly ideal.

pp-mo:
Rebased.
@dkillick @lbdreyer Please re-check !

lib/iris/coords.py
```diff
@@ -1496,7 +1481,7 @@ def _repr_other_metadata(self):
     @property
     def points(self):
         """The local points values as a read-only NumPy array."""
-        points = self._points.view()
+        points = self._points
         return points
```
dkillick: You could combine these two lines now, surely?

lib/iris/coords.py
```diff
@@ -1600,9 +1585,6 @@ def _sanitise_array(self, src, ndmin):
         # Ensure the array has enough dimensions.
         # NB. Returns the *same object* if result.ndim >= ndmin
         result = np.array(result, ndmin=ndmin, copy=False)
-        # We don't need to copy the data, but we do need to have our
-        # own view so we can control the shape, etc.
-        result = result.view()
```
dkillick: Why do we no longer want to make a view of `result`? Is this because we've removed `LazyArray` or is it more in depth than that?

lib/iris/cube.py
```diff
@@ -2168,23 +2161,14 @@ def new_cell_measure_dims(cm_):
...
 
-        if first_slice is not None:
-            data = cube_data[first_slice]
-        else:
-            data = copy.deepcopy(cube_data)
-
-        for other_slice in slice_gen:
-            data = data[other_slice]
+        # Index with the keys, using orthogonal slicing.
+        dimension_mapping, data = iris.util._slice_data_with_keys(
```
dkillick: Where is `dimension_mapping` used?

lib/iris/util.py
```diff
@@ -696,6 +696,27 @@ def _build_full_slice_given_keys(keys, ndim):
...
+    # NOTE: by using slicing on only one index at a time, this also generally
+    # avoids copying the data, except when a key contains a list of indices.
+    full_slice = _build_full_slice_given_keys(keys, n_parent_dims)
+
+    dims_mapping, slices_iter = column_slices_generator(full_slice,
+                                                        n_parent_dims)
+    for this_slice in slices_iter:
+        data = data[this_slice]
+        if data.ndim > 0 and min(data.shape) < 1:
+            raise IndexError('Cannot index with zero length slice.')
```
pp-mo: Existing code *allows* slicing a cube to 'nothing', like cube[2:2], but *disallows* this for a Coord (with the message here).  So..
```
>>> cube = iris.cube.Cube(np.arange(5))
>>> print cube[1:1]
unknown / (unknown)                 (-- : 0)
>>> cube.add_aux_coord(iris.coords.DimCoord(np.arange(5), long_name='x'), 0)
>>> print cube[1:1]
 . . .
  File "/opt/scitools/environments/experimental/2017_04_27/lib/python2.7/site-packages/iris/coords.py", line 492, in __getitem__
    raise IndexError('Cannot index with zero length '
IndexError: Cannot index with zero length slice.
>>> 
```

There are tests that check for this error, like [here](https://github.com/SciTools/iris/blob/3eb48980cb58a7784d9003d9fbe6f94003c1ce87/lib/iris/tests/test_cdm.py#L610)

So, I've retained it, but the restriction now applies to Cubes as well -- that didn't break any tests :crossed_fingers: 
This effectively disallows it for cubes as well,

lib/iris/coords.py
pp-mo: ok

lib/iris/util.py
dkillick: It's a nice outcome of this unification that the same behaviour is now observed for coords and cubes :+1: Might be nice to add a test to confirm the cube on its own also behaves like this now.

lib/iris/aux_factory.py
```diff
@@ -195,7 +195,7 @@ def _nd_bounds(self, coord, dims, ndim):
         nd_shape = [1] * ndim + [coord.nbounds]
         for dim, size in zip(dims, coord.shape):
             nd_shape[dim] = size
-        bounds.shape = tuple(nd_shape)
```
pp-mo: I think that modifying the view of an array in-place is horrible !

lib/iris/coords.py
pp-mo: My reasoning is that this **ought to be** redundant, now that we no longer support LazyArray-s.

I would generally expect taking a view of an (ordinary) real array to be a redundant operation.
Likewise, our `__init__` codes, and data assignment methods, take the view that a Coord or Cube "owns" it's data, so that should be reasonable.
However, the one place it didn't work was because of assigning array.shape as an in-place modification in aux_factoriy code, e.g. [here](https://github.com/SciTools/iris/pull/2519/files#r114524615).
I think that is highly unexpected, really horrible + shouldn't be done -- so I fixed it.

lib/iris/cube.py
pp-mo: It is used as a free variable in the two methods defined immediately _above_, `new_coord_dims` and `new_cell_measure_dims`.
Not obvious, is it ?!?
I considered moving them to below this point for more clarity, but the minimal change principle won out...

dkillick: > Not obvious, is it ?!?

In a word, no. I did notice it was being defined in the previous form of the code, but I still wasn't sure it was being used...

Still, :+1: for following the minimum change principle!

lib/iris/util.py
```diff
@@ -696,6 +696,27 @@ def _build_full_slice_given_keys(keys, ndim):
     return full_slice
 
 
+def _slice_data_with_keys(data, keys, n_parent_dims):
+    # Combine _build_full_slice_given_keys and column_slices_generator to index
```
pp-mo: I think this is rather work-in-progress, but I'll clarify it + see if it improves...

pp-mo: This new routine should probably have its own test.
I could write something now, or we could wait until we decide whether this becomes part of DataManager ?

pp-mo:
I think we should maybe have a test for the new function `iris.util._slice_data_with_keys` (which I just changed a bit).
I could do that now, or we could wait until we decide whether it lives inside DataManager ?

lib/iris/util.py
```diff
@@ -696,22 +696,39 @@ def _build_full_slice_given_keys(keys, ndim):
...
-                                                        n_parent_dims)
+def _slice_data_with_keys(data, keys):
+    """
+
+    Index an array-like object with orthogonal keys.
+
+    This enforces an orthogonal interpretation of indexing, which means that
+    both 'real' (numpy) arrays and other array-likes index in the same way,
+    instead of numpy arrays doing 'fancy indexing'.
+
```
dkillick: You could do with a section 'Args:' in the docstring that describes what each of the incoming arguments does.

lib/iris/util.py
```diff
@@ -696,22 +696,39 @@ def _build_full_slice_given_keys(keys, ndim):
...
+    Returns (dim_map, data_region), where :
+
+    *  dim_map (dict) :
+        A dimension map, as for :func:`column_slices_generator`.
```
dkillick: Doesn't the dimension map come _from_ `column_slices_generator`, rather than being used _for_ it? It might also be worth describing what the dimension map does, or how/why it might be used?

dkillick:
These simplifications are worthwhile I think @pp-mo. But I'm afraid that I think we _do_ need some testing for the new `iris.util._slice_data_with_keys` function. Perhaps try and keep the testing quite terse, given that the new function may be only temporary...

lib/iris/util.py
pp-mo: > Doesn't the dimension map come from column_slices_generator, rather than being used for it?
> ... It might also be worth describing what the dimension map does, or how/why it might be used?

By "as for" I mean "as it is for" -- so the use is described there, for the other routine.
As this calls the other routine, and it is public, I would much prefer to simply defer to that description.

lib/iris/util.py
```diff
@@ -696,6 +696,44 @@ def _build_full_slice_given_keys(keys, ndim):
     return full_slice
 
 
+def _slice_data_with_keys(data, keys):
+    """
+
```
lbdreyer: Unecessary line?

lbdreyer:
>But I'm afraid that I think we do need some testing for the new iris.util._slice_data_with_keys function. 

Agreed
>Perhaps try and keep the testing quite terse, given that the new function may be only temporary...

but the tests you write could be modified so that they become data manager tests once it's moved there

lib/iris/coords.py
```diff
@@ -1496,8 +1481,7 @@ def _repr_other_metadata(self):
     @property
     def points(self):
         """The local points values as a read-only NumPy array."""
```
lbdreyer: Why does this say `read-only NumPy array.` ?
```
>>> coord = AuxCoord([1, 2, 3])
>>> print coord.points.flags.writeable
True
>>> print coord._points.flags.writeable
True
>>> coord.points = [4, 5, 6]
>>> print coord.points.flags.writeable
True

```

lbdreyer: Okay! I see this is DimCoord not AuxCoord:
```
>>> coord = DimCoord([1, 2, 3])
>>> print coord.points.flags.writeable
False
```

lib/iris/coords.py
djkirkham: Isn't the reason for this to stop users messing with the points array 'by the back door'?
```
>>> d = DimCoord([1,2,3,4])
>>> d.points.shape = (2,2)
>>> print d
DimCoord(array([[1, 2],
       [3, 4]]), standard_name=None, units=Unit('1'))
```

pp-mo:
> @lbdreyer but the tests you write could be modified so that they become data manager tests once it's moved there

:+1:

pp-mo:
Hi @lbdreyer @dkillick 
New code with added test.
Hope this addresses all to-date.
More delay --> yet another rebase, I'm afraid.

lib/iris/tests/unit/util/test__slice_data_with_keys.py
```diff
@@ -0,0 +1,257 @@
...
+    def __getitem__(self, keys):
+        return keys
+
+
+# An Indexer object for generating indexing keys in a nice visible way.
+Index = Indexer()
+
+
+class MixinIndexingTest(object):
+    def check(self, shape, keys, expect_call_keys=None, expect_map=None):
```
djkirkham: This check feels a bit strict. As I understand it, the purpose of `_slice_data_with_keys` is to avoid using indices containing multiple tuples, in order to avoid Numpy 'fancy' indexing. So why not just do the following checks:

1. Each index only consists of either non-tuple values, or a single tuple plus any number of `slice(None,None,None)`
2. The resulting array is as expected

That way it gives freedom to tweak the algorithm in the future if necessary, without changing the tests.

Actually what is really being tested is the behaviour of `column_slices_generator`, so these tests shouldn't really be necessary. But the testing for that function is minimal...

dkillick: > what is really being tested is the behaviour of `column_slices_generator`

Or the combination with `_build_full_slice_given_keys`. See also the [note in the docstring](https://github.com/pp-mo/iris/blob/daf80ac690daabb4cc3fe9fb291a5a5fd46f49e1/lib/iris/tests/unit/util/test__slice_data_with_keys.py#L20-L24) for this test module.

> these tests shouldn't really be necessary

We've introduced a new, untested function. Of course there should be tests for it! Especially when the real driver for these tests was the terrible lack of any other testing for anything within this functionality space. Thus adding these tests is a positive step as it gives us some assurance for the first time that this functionality is doing what we thought it was...

You could perhaps argue that these should be integration tests, or alternatively you could say we're testing the unit that is `iris.util._slice_data_with_keys`.

lib/iris/tests/unit/util/test__slice_data_with_keys.py
```diff
@@ -0,0 +1,257 @@
...
+                msg += '\n]'
+                return msg
+
+            self.assertTrue(equal, errmsg.format(showkeys(calls_got),
+                                                 showkeys(expect_call_keys)))
+        if expect_map is not None:
+            self.assertEqual(dim_map, expect_map)
+
+
+class Test_indexing(MixinIndexingTest, tests.IrisTest):
```
dkillick: @pp-mo these tests look good :+1: ... and very thorough!

My only comment, which applies to both test classes here, is that there are very few failure modes. Admittedly there are only two exceptions within all the functions under test here, but are there any other edge cases that need to be considered?

lib/iris/tests/unit/util/test__slice_data_with_keys.py
djkirkham: >  Of course there should be tests for it!
You're right, and I'd forgotten that the function also used `_build_full_slice_given_keys`. But still, the checks on the indices really apply to `column_slices_generator`

pp-mo:
Following discussion, I added some purely results-based tests.
The existing mock-ist approach is liable to collapse if we need to modify any of the implementation, but the new test should be robust to refactoring.

lib/iris/tests/unit/util/test__slice_data_with_keys.py
pp-mo: This exercises errors raised by the 3 routines covered, but no more.
I'm certainly not trying to cover all cases here, as there are far too many possible forms of indexing (at least in numpy), and who knows what differences between numpy and dask may be lurking ?
I've just picked some known testcases.

dkillick:
I was minded to say I was happy with things as they were _before_ the latest commit, so I'm definitely minded to say I'm happy now! I'll wait to see if we can persuade Travis to be happy and then I think I might merge...

PR #2520, opened 2017-05-03
Dask merge concat fill value

bjlittle:
This PR refines the approach taken with handling the `cube.fill_value` by merge and concatenate.

This PR extends the expected behaviour of `numpy` with regards to how it manages the `fill_value` when combining masked arrays and/or ndarrays to cubes (it is valid for a cube to have a `fill_value` even though it has a ndarray payload).

In summary, when combining cubes via merge or concatenate, the resultant cube will have:
* a `fill_value` that is non-`None` **iff** all cubes have the same non-`None` `fill_value`
* a `fill_value` that is `None`, if:
   * at least one cube has a `fill_value` of `None`, or
   * the candidate cube and the proto-cube have a different `fill_value`

This simply means that, 
* if the `cube.fill_value` is set to a non-`None` value, and the cube has a masked payload, then the masked array returned by `cube.data` will have a `fill_value` that matches the `cube.fill_value`.
* if the `cube.fill_value` is set to `None`, and the cube has a masked payload, then the masked array returned by `cube.data` will have the default `fill_value` generated by `numpy` for that specific `dtype`.

### DIFF
--- a/lib/iris/_concatenate.py
+++ b/lib/iris/_concatenate.py
@@ -416,7 +416,6 @@ class _CubeSignature(object):
             - scalar coords
             - attributes
             - dtype
-            - fill_value
 
         Args:
 
@@ -464,14 +463,6 @@ class _CubeSignature(object):
         if self.data_type != other.data_type:
             msgs.append(msg_template.format('Data types', '',
                                             self.data_type, other.data_type))
-
-        # Check fill value.
-        if self.fill_value != other.fill_value:
-            # Allow fill-value promotion if either fill-value is None.
-            if self.fill_value is not None and other.fill_value is not None:
-                msgs.append(msg_template.format('Fill values', '',
-                                                self.fill_value,
-                                                other.fill_value))
 
         # Check _cell_measures_and_dims
         if self.cell_measures_and_dims != other.cell_measures_and_dims:
@@ -727,10 +718,14 @@ class _ProtoCube(object):
 
         # Check for compatible coordinate signatures.
         if match:
-            if self._cube_signature.fill_value is None and \
-                    cube_signature.fill_value is not None:
-                # Perform proto-cube fill-value promotion.
-                self._cube_signature.fill_value = cube_signature.fill_value
+            fill_value = self._cube_signature.fill_value
+            # Determine whether the fill value requires to be
+            # demoted to the default value.
+            if fill_value is not None:
+                if cube_signature.fill_value is None or \
+                        cube_signature.fill_value != fill_value:
+                    # Demote the fill value to the default.
+                    self._cube_signature.fill_value = None
             coord_signature = _CoordSignature(cube_signature)
             candidate_axis = self._coord_signature.candidate_axis(
                 coord_signature)
--- a/lib/iris/_concatenate.py
+++ b/lib/iris/_concatenate.py
@@ -722,8 +722,7 @@ class _ProtoCube(object):
             # Determine whether the fill value requires to be
             # demoted to the default value.
             if fill_value is not None:
-                if cube_signature.fill_value is None or \
-                        cube_signature.fill_value != fill_value:
+                if cube_signature.fill_value != fill_value:
                     # Demote the fill value to the default.
                     self._cube_signature.fill_value = None
             coord_signature = _CoordSignature(cube_signature)
--- a/lib/iris/_merge.py
+++ b/lib/iris/_merge.py
@@ -407,11 +407,6 @@ class _CubeSignature(namedtuple('CubeSignature',
         if self.data_type != other.data_type:
             msg = 'cube data dtype differs: {} != {}'
             msgs.append(msg.format(self.data_type, other.data_type))
-        if self.fill_value != other.fill_value:
-            # Allow fill-value promotion if either fill-value is None.
-            if self.fill_value is not None and other.fill_value is not None:
-                msg = 'cube fill value differs: {} != {}'
-                msgs.append(msg.format(self.fill_value, other.fill_value))
         if (self.cell_measures_and_dims != other.cell_measures_and_dims):
             msgs.append('cube.cell_measures differ')
 
@@ -1285,15 +1280,19 @@ class ProtoCube(object):
             this :class:`ProtoCube`.
 
         """
-        signature = self._cube_signature
+        cube_signature = self._cube_signature
         other = self._build_signature(cube)
-        match = signature.match(other, error_on_mismatch)
+        match = cube_signature.match(other, error_on_mismatch)
         if match:
-            if signature.fill_value is None and other.fill_value is not None:
-                # Perform proto-cube fill-value promotion.
-                fv = other.fill_value
-                self._cube_signature = self._build_signature(self._source,
-                                                             fill_value=fv)
+            # Determine whether the fill value requires to be demoted
+            # to the default value.
+            if cube_signature.fill_value is not None:
+                if other.fill_value is None or \
+                        cube_signature.fill_value != other.fill_value:
+                    # Demote the fill value to the default.
+                    signature = self._build_signature(self._source,
+                                                      default_fill_value=True)
+                    self._cube_signature = signature
             coord_payload = self._extract_coord_payload(cube)
             match = coord_payload.match_signature(self._coord_signature,
                                                   error_on_mismatch)
@@ -1614,7 +1613,7 @@ class ProtoCube(object):
                               self._vector_aux_coords_dims):
             aux_coords_and_dims.append(_CoordAndDims(item.coord, dims))
 
-    def _build_signature(self, cube, fill_value=None):
+    def _build_signature(self, cube, default_fill_value=False):
         """
         Generate the signature that defines this cube.
 
@@ -1625,16 +1624,18 @@ class ProtoCube(object):
 
         Kwargs:
 
-        * fill_value:
-            Promoted fill-value to use instead of the source cube
-            fill-value.
+        * default_fill_value:
+            Override the cube fill value with the default fill value of None.
+            Default is False i.e. use the provided cube.fill_value when
+            constructing the cube signature.
 
         Returns:
             The cube signature.
 
         """
-        if fill_value is None:
-            fill_value = cube.fill_value
+        fill_value = cube.fill_value
+        if default_fill_value:
+            fill_value = None
         return _CubeSignature(cube.metadata, cube.shape, cube.dtype,
                               fill_value, cube._cell_measures_and_dims)
 
--- a/lib/iris/_merge.py
+++ b/lib/iris/_merge.py
@@ -1287,8 +1287,7 @@ class ProtoCube(object):
             # Determine whether the fill value requires to be demoted
             # to the default value.
             if cube_signature.fill_value is not None:
-                if other.fill_value is None or \
-                        cube_signature.fill_value != other.fill_value:
+                if cube_signature.fill_value != other.fill_value:
                     # Demote the fill value to the default.
                     signature = self._build_signature(self._source,
                                                       default_fill_value=True)
--- a/lib/iris/tests/test_concatenate.py
+++ b/lib/iris/tests/test_concatenate.py
@@ -335,47 +335,23 @@ class TestNoConcat(tests.IrisTest):
         result = concatenate(cubes)
         self.assertEqual(len(result), 2)
 
-    def test_masked_fill_value_difference(self):
-        cubes = []
-        y = (0, 2)
-        cube = _make_cube((0, 2), y, 1, mask=True)
-        cube.fill_value = 10
-        cubes.append(cube)
-        cube = _make_cube((2, 4), y, 1, mask=True)
-        cube.fill_value = 20
-        cubes.append(cube)
-        result = concatenate(cubes)
-        self.assertEqual(len(result), 2)
-
-    def test_masked_fill_value_difference__exception(self):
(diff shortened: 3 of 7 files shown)


bjlittle:
Ping @djkirkham and @lbdreyer ... :smile:

lib/iris/_merge.py
```diff
@@ -407,11 +407,6 @@ def match(self, other, error_on_mismatch):
         if self.data_type != other.data_type:
             msg = 'cube data dtype differs: {} != {}'
             msgs.append(msg.format(self.data_type, other.data_type))
-        if self.fill_value != other.fill_value:
-            # Allow fill-value promotion if either fill-value is None.
-            if self.fill_value is not None and other.fill_value is not None:
-                msg = 'cube fill value differs: {} != {}'
-                msgs.append(msg.format(self.fill_value, other.fill_value))
```
lbdreyer: Why did you choose to allow for merging and concatenating cubes with different fill values?
I realise this is not a simple question so I can call you to discuss this if that's easier?

bjlittle: @lbdreyer This is more in alignment with the behaviour of `upstream/master`, which doesn't discriminate based on `fill_value`. So no breaking or surprising behaviour for the end user.

It also aligns with the behaviour of `numpy`, which will merge/concatenate arrays with different `fill_values`, but the resultant array has a `fill_value` that is replaced with the `numpy` nominated default for that `dtype`. 

With regards to cubes, we're doing the same - if we have two cubes that only differ on `fill_value`, say one is `1234` and the other is `4321`, then they will merge/concatenate into one cube, but the resultant cube will have a `fill_value` of `None. When the user does a `cube.data` to get the data, the resulting masked array (assuming the payload realizes is masked) will have a `fill_value` that is the `numpy` nominated default.

lib/iris/tests/test_concatenate.py
```diff
@@ -748,6 +730,49 @@ def test_concat_2x2d_aux_xy_bounds(self):
...
+        y = (0, 2)
+        cubes = [_make_cube((i*2, i*2+2), y, i, mask=True,
+                            fill_value=None) for i in range(3)]
+        for combo in itertools.permutations(cubes):
+            result = iris.cube.CubeList(combo).concatenate_cube()
+            self.assertIsNone(result.fill_value)
+            np_fill_value = ma.masked_array(0, dtype=result.dtype).fill_value
+            self.assertEqual(result.data.fill_value, np_fill_value)
+
+    def test_fill_value_invariant_to_order__different_non_None(self):
```
lbdreyer: Other than testing the different order, is this much different to [test_concat_masked_2y2d__default_fill_value_from_diff](https://github.com/SciTools/iris/pull/2520/files#diff-af298a2170476a32abb0c94b312b4ff5R438)?

I don't understand what you gain from doing all the permutations?

lbdreyer: Okay I understand that this concatenates cubes with x's like
```
concat [0, 1] [2, 3] [4, 5] => [0, 1, 2, 3, 4, 5]
concat [0, 1] [4, 5] [2, 3] => [0, 1, 2, 3, 4, 5]
concat [4, 5] [0, 1] 2, 3] => [0, 1, 2, 3, 4, 5]
```

and checking the fill value is correct after that

lib/iris/tests/test_concatenate.py
```diff
@@ -748,6 +730,49 @@ def test_concat_2x2d_aux_xy_bounds(self):
         self.assertEqual(len(result), 1)
         self.assertEqual(result[0].shape, (2, 4))
 
+    def test_fill_value_invariant_to_order__same_non_None(self):
```
lbdreyer: This test naming was a bit confusing to me. I only understood what `__same_non_None` meant when I started writing a comment asking you what it is. I am going to try to think of a different name

lib/iris/tests/test_concatenate.py
```diff
@@ -335,113 +336,93 @@ def test_order_difference(self):
...
-        cubes.append(cube)
-        cube = _make_cube((2, 4), y, 1, mask=True)
-        cube.fill_value = 20
-        cubes.append(cube)
-        emsg = 'Fill values differ'
-        with self.assertRaisesRegexp(iris.exceptions.ConcatenateError, emsg):
-            cubes.concatenate_cube()
-
 
 class Test2D(tests.IrisTest):
```
lbdreyer: You never test the fill value when all inputs that get concatenated are unmasked. You could always add a quick check to the test on [L535](https://github.com/bjlittle/iris/blob/fb0a3a51b308d7966c408b6ec2e498667cff29c9/lib/iris/tests/test_concatenate.py#L535)

lib/iris/tests/test_merge.py
```diff
@@ -145,81 +145,143 @@ def _make_cube(self, data, dtype=np.dtype('int32'), fill_value=None,
...
 
+    @staticmethod
+    def _expected_fill(fill0, fill1):
+        result = None
+        if fill0 == fill1:
+            result = fill0
+        return result
+
+    def _check_fill_value(self, result, fill0, fill1):
+        fill_value = self._expected_fill(fill0, fill1)
```
lbdreyer: Rename to be `expected_fill_value` or `expected_cube_fill_value`?
May make it easier to follow

lib/iris/tests/test_merge.py
```diff
@@ -145,81 +145,143 @@ def _make_cube(self, data, dtype=np.dtype('int32'), fill_value=None,
...
         self.dtype = np.dtype('int32')
-        self.fill_value = 1234
-        self.lazy_combos = itertools.product([False, True],
-                                             [False, True])
+        fill_value = 1234
+        lazy_combos = itertools.product([False, True],
+                                        [False, True])
+        fill_combos = itertools.product([None, fill_value],
+                                        [fill_value, None])
+        self.combos = itertools.product(lazy_combos, fill_combos)
```
lbdreyer: Was it too complicated to have another `itertools.product` for ndarray and masked?

lbdreyer:
This looks very close!

The testing of concatenate isn't less complete than the testing of merge, but there are endless combinations (masked/unmasked, fill_value_same/fill_value_diff/fill_value_None, lazy/real etc) that we can test for... I want to go through some possible *important* combinations that we are missing.
but otherwise I'm happy with the changes.

lib/iris/tests/test_concatenate.py
bjlittle: Yeah, test naming ... ugh ... you are right, I've don my thinking cap :cowboy_hat_face:

lbdreyer:
I've looked into testing of concatenate a little more and ignoring the endless combinations we could test I think there are two missing combinations that we should at least consider.

1) There doesn't appear to be a test for concatenating cubes that are masked and unmasked where both have a fill value of None.

2) Also there isn't testing of fil value when both cubes are unmasked.

lib/iris/tests/test_merge.py
bjlittle: I think I'm a wee bit wary of folding that in, as it seems like we'll end up with a single test that does absolutely everything and will be a tad impenetrable ... is that okay?

lbdreyer: Sure! makes sense

bjlittle:
@lbdreyer Okay, so I've addressed your review actions in the simplest way possible. Adding a concatenate equivalent to the merge `TestDataMergeCombos` isn't really practical and would involve quite a bit of effort, so I've added the missing test coverage through new tests and adding to existing tests.

lib/iris/_concatenate.py
```diff
@@ -727,10 +718,14 @@ def register(self, cube, axis=None, error_on_mismatch=False,
...
-                # Perform proto-cube fill-value promotion.
-                self._cube_signature.fill_value = cube_signature.fill_value
+            fill_value = self._cube_signature.fill_value
+            # Determine whether the fill value requires to be
+            # demoted to the default value.
+            if fill_value is not None:
+                if cube_signature.fill_value is None or \
+                        cube_signature.fill_value != fill_value:
+                    # Demote the fill value to the default.
+                    self._cube_signature.fill_value = None
```
djkirkham: This can be simplified to
```
if cube_signature.fill_value != self._cube_signature.fill_value:
    self._cube_signature.fill_value = None
```

lib/iris/_merge.py
```diff
@@ -1285,15 +1280,19 @@ def register(self, cube, error_on_mismatch=False):
...
-                fv = other.fill_value
-                self._cube_signature = self._build_signature(self._source,
-                                                             fill_value=fv)
+            # Determine whether the fill value requires to be demoted
+            # to the default value.
+            if cube_signature.fill_value is not None:
+                if other.fill_value is None or \
+                        cube_signature.fill_value != other.fill_value:
+                    # Demote the fill value to the default.
+                    signature = self._build_signature(self._source,
```
djkirkham: Why do we need a call to `_build_signature` here, but in the concatenate case you can just set `fill_value` to None?

bjlittle: One is a class the other is a specialization of a `namedtuple`

lib/iris/_concatenate.py
bjlittle: Yup, also applies to `_merge.py` :+1:

lib/iris/tests/test_concatenate.py
```diff
@@ -335,113 +337,117 @@ def test_order_difference(self):
...
-                          fill_value=fill_value)
+                          fill_value=1234)
         cubes.append(cube)
         result = concatenate(cubes)
         self.assertEqual(len(result), 1)
-        self.assertEqual(result[0].fill_value, fill_value)
-        self.assertEqual(result[0].data.fill_value, fill_value)
+        self.assertIsNone(result[0].fill_value)
+        np_fill_value = ma.masked_array(0, dtype=result[0].dtype).fill_value
+        self.assertEqual(result[0].data.fill_value, np_fill_value)
```
djkirkham: I'm wondering if we need this check. I think perhaps it should be part of the contract that if the input fill values don't match then the fill value you get out on the array shouldn't be relied upon to have any particular value. That said, we are making an effort to make sure the correct fill value is placed on the data array when it _is_ valid, so I'm not sure.

lib/iris/tests/test_concatenate.py
```diff
@@ -748,6 +760,49 @@ def test_concat_2x2d_aux_xy_bounds(self):
...
+        y = (0, 2)
+        cubes = [_make_cube((0, 2), y, 0, mask=True, fill_value=None)]
+        cubes.append(_make_cube((2, 4), y, 1, mask=True, fill_value=1234))
+        cubes.append(_make_cube((4, 6), y, 2, mask=True, fill_value=4321))
+        for combo in itertools.permutations(cubes):
+            result = iris.cube.CubeList(combo).concatenate_cube()
+            self.assertIsNone(result.fill_value)
+            np_fill_value = ma.masked_array(0, dtype=result.dtype).fill_value
+            self.assertEqual(result.data.fill_value, np_fill_value)
+
```
djkirkham: Nice test coverage!

bjlittle: I try :wink: 

At least we're trying to join the dots on what's happening here ... so if it changes, a whole bunch of tests are now gonna complain .... which is a good thing!

lib/iris/tests/test_concatenate.py
djkirkham: I'd be happy either way on both points to be honest

lib/iris/tests/test_merge.py
```diff
@@ -146,14 +146,14 @@ def _make_cube(self, data, dtype=np.dtype('int32'), fill_value=None,
...
-    def _expected_fill(fill0, fill1):
+    def _expected_fill_value(fill0, fill1):
         result = None
         if fill0 == fill1:
             result = fill0
         return result
 
     def _check_fill_value(self, result, fill0, fill1):
-        fill_value = self._expected_fill(fill0, fill1)
+        fill_value = self._expected_fill_value(fill0, fill1)
```
lbdreyer: When I said I would prefer a rename I meant the variable, not the method, 
i.e. ```expected_fill_value = self._expected_fill(fill0, fill1)```
and then the self.assertEqual checks would be:

```self.assertEqual(data.fill_value, expected_fill_value)```

File: lib/iris/tests/test_merge.py [hunk 12]
```diff
@@ -145,81 +145,143 @@ def _make_cube(self, data, dtype=np.dtype('int32'), fill_value=None,
...
+    @staticmethod
+    def _expected_fill_value(fill0, fill1):
+        result = None
+        if fill0 == fill1:
+            result = fill0
+        return result
+
+    def _check_fill_value(self, result, fill0, fill1):
+        fill_value = self._expected_fill_value(fill0, fill1)
+        if fill_value is None:
```
lbdreyer: When I said I would prefer a rename I meant the variable, not the method, 
i.e. ```expected_fill_value = self._expected_fill(fill0, fill1)```
and then the self.assertEqual checks would be:

```self.assertEqual(data.fill_value, expected_fill_value)```

lbdreyer:
@bjlittle if you adress my [last comment](https://github.com/SciTools/iris/pull/2520/files/d44458d1125f3a2f95092e46a511c2b7a32ad05c#r114774901) I will happily merge this in

bjlittle:
@lbdreyer Thanks :+1:

PR #2521, opened 2017-05-03
Dask core data

bjlittle:
`core_data` should be a function, not a read-only class property.

### DIFF
--- a/lib/iris/_data_manager.py
+++ b/lib/iris/_data_manager.py
@@ -117,7 +117,7 @@ class DataManager(object):
             same_realised_dtype = self._realised_dtype == other._realised_dtype
             same_dtype = self.dtype == other.dtype
             if same_lazy and same_realised_dtype and same_dtype:
-                result = array_equal(self.core_data, other.core_data)
+                result = array_equal(self.core_data(), other.core_data())
 
         return result
 
@@ -149,13 +149,14 @@ class DataManager(object):
         Returns an string representation of the instance.
 
         """
-        fmt = '{cls}({self.core_data!r}{dtype})'
+        fmt = '{cls}({data!r}{dtype})'
         dtype = ''
 
         if self._realised_dtype is not None:
             dtype = ', realised_dtype={!r}'.format(self._realised_dtype)
 
-        result = fmt.format(self=self, cls=type(self).__name__, dtype=dtype)
+        result = fmt.format(data=self.core_data(), cls=type(self).__name__,
+                            dtype=dtype)
 
         return result
 
@@ -217,7 +218,7 @@ class DataManager(object):
             else:
                 # Check that the replacement data is valid relative to
                 # the currently managed data.
-                DataManager(self.core_data).replace(data)
+                DataManager(self.core_data()).replace(data)
                 # If the replacement data is valid, then use it but
                 # without copying it.
 
@@ -262,7 +263,6 @@ class DataManager(object):
                 # Check the manager contract, as the managed dtype has changed.
                 self._assert_axioms()
 
-    @property
     def core_data(self):
         """
         If real data is being managed, then return the :class:`~numpy.ndarray`
@@ -377,7 +377,7 @@ class DataManager(object):
         if self._realised_dtype is not None:
             result = self._realised_dtype
         else:
-            result = self.core_data.dtype
+            result = self.core_data().dtype
 
         return result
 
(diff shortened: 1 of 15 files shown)


bjlittle:
Ping @djkirkham and @lbdreyer ...

dkillick:
@bjlittle could you provide a bit of justification for this change?

bjlittle:
@dkillick Yes, it is just a misuse of @property, and if we don't change it now, it will never be changed.

I believe that we should restrict our use of read-only properties, particularly if they are not writable. Being a read-only property, it's not clear whether this property can be assigned to for the users perspective of the API (and for developers too). Making it a function, then it's clear that it cannot be used as an assignable token. For example, that's why `cube.lazy_data` is a function and not a read-only property`.

Your not going to be tempeted to do `cube.core_data() = data`, but you would with `cube.core_data = data`.

I'm sure that @pp-mo has some strong opinions on this.

dkillick:
Thanks @bjlittle, very insightful. I guessed it was to do with how/when/why we use `@property` attributes. but it is good to get your reasoning too.

> I'm sure that @pp-mo has some strong opinions on this.

Maybe we should merge it before he gets the chance to air them!! :wink: :innocent:

bjlittle:
@dkillick Cool ... I think that @pp-mo is strongly in favour of not having read-only properties for the sake of them, and that they should be functions ... but that's me putting words in his mouth :wink:

bjlittle:
@dkillick To be honest, my next move is to remove the `cube.core_data` method ... it's a convenience method on a public API that I believe we can live without. The `Cube` really is a bit of a god class as it is, and doesn't need further bloat.

dkillick:
Thanks @bjlittle - a nice tidy to the API.

dkillick:
> "_just one more wafer-thin mint Mr. Creosote_"

I think you'll find it's "waffer-thin" :nerd_face:

dkillick:
> my next move is to remove the cube.core_data method 

My only question about doing that is how, from a `cube`, will you then access the 'whatever-the-current-data-is-as-it-is' object -- that is, get lazy data if data is currently lazy, or get real data if the data is currently realised?

QuLogic:
What's a waffer?

bjlittle:
> I think you'll find it's "waffer-thin" :nerd_face:

See [dictionary wafer](https://www.google.co.uk/search?q=dictionary+wafer&ie=utf-8&oe=utf-8&gws_rd=cr&ei=5vIKWeysFoLkUuedroAG) ... surely :stuck_out_tongue_winking_eye:

dkillick:
@QuLogic an odd pronunciation of the word "wafer". See [this Monty Python clip](https://www.youtube.com/watch?v=HJZPzQESq_0) which started it all off...

dkillick:
I'm sure John Cleese pronounces it really weirdly...

bjlittle:
@dkillick A-ha! You confused me with some subtle, cheeky humour ... gets me every time :smile:

dkillick:
Anyway, now that we've got to the bottom of all that, did you have any thoughts on my [earlier comment](https://github.com/SciTools/iris/pull/2521#issuecomment-299134345) @bjlittle?

bjlittle:
@dkillick Yes, okay ... back to being serious (for now) ...

Outside tests, there's actually only a couple of places in the code-base that we use `cube.core_data()` ... I'm of the opinion that ...
```python
if cube.has_lazy_data():
    data = cube.lazy_data()
else:
    data = cube.data
```
is no hardship ... what do you think?

dkillick:
> only a couple of places in the code-base that we use `cube.core_data()`

I'm more interested in **user-land** at this point. I think that as a user I'd like a quick easy (cube?) method to get the data as it currently is. I suspect that if we dangle your suggestion above to users:

* they'll go off and write their own utility function to do exactly the above, or
* duplicate the above code _everywhere_, or
* not understand the subtlety and end up calling `cube.data` everywhere and wonder why everything runs terribly.

Much as I'm a fan of removing bloat, I'm not sure `cube.core_data()` is the wafer (waffer) to not offer (even though I hugely dislike the naming)...

bjlittle:
@dkillick ... I'm not suggesting that we also purge `DataManager.core_data()` nor the use of `core_data` by the fileformats - that's a convenience to us (iris, under the hood) that seems useful and appropriate.

I can't quite gauge the level of utility to end users ...

... and I so dislike the name of `cube.core_data` also.

PR #2523, opened 2017-05-04
avoid runtime overflow

marqh:
There is currently the potential for a `RuntimeWarning: overflow encountered in int_scalars` which is missed during collapsed calculations, which can trip up unwary users of limited data types, such as int32 for very large numbers (e.g. seconds since 1970)

This PR pushes the casting, which already occurs, into the individual values, to limit the chance of overflow.  

The chance is still there, it is the responsibility of data creators to use appropriate data types.  this is a point fix which will help some users

### DIFF
--- a/lib/iris/tests/unit/coords/test_Coord.py
+++ b/lib/iris/tests/unit/coords/test_Coord.py
@@ -1,4 +1,4 @@
-# (C) British Crown Copyright 2013 - 2016, Met Office
+# (C) British Crown Copyright 2013 - 2017, Met Office
 #
 # This file is part of Iris.
 #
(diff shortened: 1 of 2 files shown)


lib/iris/tests/unit/coords/test_Coord.py
```diff
@@ -324,6 +324,12 @@ def test_numeric_nd(self):
         with self.assertRaises(ValueError):
             coord.collapsed()
 
+    def test_collapsed_overflow(self):
+        coord = DimCoord(points=np.array([1493892000, 1493895600, 1493899200],
```
cpelley: I suggest using `np.iinfo(np.int32).max` to get appropriate numbers here.  Up to you, just think it makes it clearer.

marqh: these numbers came from a real scenario, hence why I used them.  the key characteristic is that each value is valid, but the combination of the two of them overflows
that is the 'bug' that I wanted to fix and this test explicitly examines that characteristic

so, on reflection, I'd rather keep them

marqh:
@cpelley 
> I see @bjlittle is assigned so I have not 'approved' as I didn't want to step on his toes.

@bjlittle may be slightly difficult to pin down over the next few days and he has not commented since this was raised, so I have removed his toes from the firing line and assigned this PR to you instead

please feel free to take charge and merge if you are content

cpelley:
Looks good to me, thanks @marqh

Issue #2516, opened 2017-04-27
Outstanding - Integrate DataManager with cube


bjlittle:
Ref - #2492 

The following outstanding issues still require to be investigated and/or addressed:
- [x] fix cube arithmetic dtype, see #2514 
- [x] fix cube concatenate and merge (see #2492)
   * currently `dask` cube does not merge an `ndarray` and `masked` array to generate a `masked` result
   * we need to iterate through all permutations or real vs lazy, and ndarray vs masked to gain confidence
   * need to adopt the concept of a cube with an empty `fill_value` of `None` matching a candidate cube with an existing `fill_value` - although two candidate cubes with difference non-`None` `fill_values` should never match (obviously)
   * also relates to #2384
- [x] investigate and implement setting `cube.fill_value` through a `cube.data = masked_array` operation, see this [comment](https://github.com/SciTools/iris/pull/2492) for context, see #2524
- [x] Set `cube.fill_value` to `None` when merging or concatenating a mixture of masked array and ndarray, see #2520
- [x] ~~Consider whether to issue a warning in `cube.data` getter when setting the `fill_value` to `None`, as the returned masked array will contain a numpy default `fill_value`~~

pp-mo:
> need to adopt the concept of a cube with an empty fill_value of None matching a candidate cube with an existing fill_value - although two candidate cubes with difference non-None fill_values should never match (obviously)

I disagree with this.
I think, if you are including data with "no known suitable fill value", the result has ... "no known suitable fill value".
For instance, 
`
np.stack(
    [ma.array([0, 1, 2], fill_value=-1),
     ma.array([-1, -2, -3])])
`
In that case, "-1" would not be a suitable fill value for the result, as it has a valid value of -1 in it.

djkirkham:
Cube merging now allows cubes to merge if one of them has a fill value and the other doesn't. But this causes problems given a list where one of the cubes has no fill value and the others have differing fill values: the first cube matched with the `None`-fill-value cube will be merged with it, and subsequent cubes won't, making merging dependent on the order of the cubes in the list.

I'm not sure what the correct behaviour should be. If we want to keep fill value promotion, then it seems like we need logic along the lines of

    For any cubes A, B and C in a list, if A can merge with each of B and C, then merging should fail unless B can merge with C.

In terms of fill values, that amounts to

    For any cubes A, B and C in a list, if A has a fill value of `None`, then merging should fail unless B and C have the same fill value or one of them has a flll value of `None`.

[...]

PR #2524, opened 2017-05-04
Populate cube.fill_value thru cube.data setter.

bjlittle:
This PR is a follow-on action from #2492 and detailed in #2516 and allows the `cube.fill_value` to be set from the `fill_value` of concrete masked data being passed to `cube.data` setter.

All of the CML changes are due to either tests directly setting `cube.data` with a masked array, or tests that use testing stock cubes that set `cube.data` directly with a masked array ... thus populating the `cube.fill_value` and thus bleeding thru to the CML. This is the behaviour that we want to see.

If data passed to the `cube.data` setter is not masked data, then the `cube.fill_value` will be clear, as before.

This PR also neatly completes the circle with `fill_value` and masked arrays wrt the cube, in that the `cube.data` getter will set the `fill_value` of any masked array returned to the user with the `cube.fill_value`, and now this PR allows the `fill_value` of a masked array to set the `cube.fill_value` through the `cube.data` setter.

### DIFF
--- a/lib/iris/tests/unit/cube/test_Cube.py
+++ b/lib/iris/tests/unit/cube/test_Cube.py
@@ -1323,12 +1323,19 @@ class Test_fill_value(tests.IrisTest):
         cube.fill_value = fill_value
         self.assertEqual(cube.fill_value, fill_value)
 
-    def test_cube_data_setter_with_none(self):
+    def test_cube_data_setter_with_ndarray(self):
         fill_value = 123
         cube = Cube(np.array(0), fill_value=fill_value)
         self.assertEqual(cube.fill_value, fill_value)
         cube.data = np.array(1)
-        self.assertEqual(cube.fill_value, None)
+        self.assertIsNone(cube.fill_value)
+
+    def test_cube_data_setter_with_masked_array(self):
+        fill_value = 123
+        cube = Cube(np.array(0))
+        self.assertIsNone(cube.fill_value)
+        cube.data = ma.masked_array(1, fill_value=fill_value)
+        self.assertEqual(cube.fill_value, fill_value)
 
     def test_fill_value_bool(self):
         fill_value = np.bool_(True)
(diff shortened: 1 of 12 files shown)


bjlittle:
Ping @dkillick and @lbdreyer ...

lbdreyer:
If you create a new masked array with a specific fill value, make it lazy, then make it a cube's data
```cube.data = lazy_array_that_was_originally_masked_with_a_fill_value```
the fill value will get lost, but I guess there's nothing we can do about that and probably not something we need to account for

**Comment by lbdreyer** [comment]
This looks good to me :+1 So unless there are any objections I'm going to merge this in!

File: lib/iris/tests/unit/cube/test_Cube.py [hunk 1]
```diff
@@ -1323,12 +1323,19 @@ def test_cube_setter(self):
...
         cube.data = np.array(1)
-        self.assertEqual(cube.fill_value, None)
+        self.assertIsNone(cube.fill_value)
+
+    def test_cube_data_setter_with_masked_array(self):
+        fill_value = 123
+        cube = Cube(np.array(0))
+        self.assertIsNone(cube.fill_value)
+        cube.data = ma.masked_array(1, fill_value=fill_value)
+        self.assertEqual(cube.fill_value, fill_value)
```
lbdreyer: This doesn't check the data type of the fill value is correct but presumably the testing of that will be in the cube.fill_value setter

bjlittle: My assumption is that unit testing elsewhere covers that, and it does.

bjlittle:
@lbdreyer Awesome, thanks!

lbdreyer:
I'm having second thoughts on this PR...
It has introduced a *dis*parity between cube.fill_value behaviour depending on whether you instantiate a cube with a masked array or add it to a cube already there, i.e.:
```
>>> data = np.array([1, 2, 3])
>>> masked_data = ma.MaskedArray([1, 2, 3], mask=[True, False, False])
>>> 
>>> cube = Cube(masked_data)
>>> print cube.fill_value
None
>>> 
>>> cube = Cube(data)
>>> print cube.fill_value
None
>>> cube.data = masked_data
>>> print cube.fill_value
999999
```


I also noticed that it introduce some strange behaviour with one of our test files which has a dtype of unint8

```
>>> cube = iris.load_cube('iris-test-data/test_data/abfAVHRRBUVI01.1985apra.abf')
>>> print cube.fill_value
None
>>> cube.data = cube.data.copy(order='C')
>>> print cube.fill_value
63
```

This is because, regardless of a masked array's dtype, the fill_value will always be 999999 (:warning:)
The cube's data in the above example is a masked array with a dtype of uint8 so the cube.fill_value setter is changing the default numpy fill value to be 63:
```print np.asarray([999999], dtype=target_dtype)```

bjlittle:
@lbdreyer Nice detective work! Let's discuss a way forward off-line and commit to a follow-on PR :+1:

PR #2525, opened 2017-05-04
docstring bug

marqh:
subtle bug in docstring, use of generator means that changes are lost, so cast to a list is required

### DIFF
--- a/lib/iris/fileformats/grib/__init__.py
+++ b/lib/iris/fileformats/grib/__init__.py
@@ -767,6 +767,7 @@ def load_pairs_from_fields(grib_messages):
 
         >>> from iris.fileformats.grib import load_pairs_from_fields
         >>> cleaned_messages = GribMessage.messages_from_filename(filename)
+        >>> cleaned_messages = list(cleaned_messages)
         >>> for message in cleaned_messages:
         ...     if message.sections[1]['productionStatusOfProcessedData'] == 0:
         ...         message.sections[1]['productionStatusOfProcessedData'] = 4
--- a/lib/iris/fileformats/grib/__init__.py
+++ b/lib/iris/fileformats/grib/__init__.py
@@ -765,7 +765,6 @@ def load_pairs_from_fields(grib_messages):
     the load pipeline.  Fields with out of specification header elements can
     be cleaned up this way and cubes created:
 
-        >>> from iris.fileformats.grib import load_pairs_from_fields
         >>> cleaned_messages = GribMessage.messages_from_filename(filename)
         >>> cleaned_messages = list(cleaned_messages)
         >>> for message in cleaned_messages:


lib/iris/fileformats/grib/__init__.py
```diff
@@ -767,6 +767,7 @@ def load_pairs_from_fields(grib_messages):
 
         >>> from iris.fileformats.grib import load_pairs_from_fields
```
lbdreyer: Could you also remove this import? I think it just confuses things.

This example uses `filename` and `GribMessage` which are defined/imported in the above example as well as  `load_pairs_from_fields` which is imported in both places.

Either reimport/define everything so that this second example looks like:
```
        >>> from iris.fileformats.grib import load_pairs_from_fields
        >>> from iris.fileformats.grib.message import GribMessage
        >>> filename = iris.sample_data_path('polar_stereo.grib2')
        >>> cleaned_messages = GribMessage.messages_from_filename(filename)
        >>> cleaned_messages = list(cleaned_messages)
        >>> for message in cleaned_messages:
        ...     if message.sections[1]['productionStatusOfProcessedData'] == 0:
        ...         message.sections[1]['productionStatusOfProcessedData'] = 4
        >>> cubes = load_pairs_from_fields(cleaned_messages)
```

Or treat it like a continuation of the previous example
```
        >>> cleaned_messages = GribMessage.messages_from_filename(filename)
        >>> cleaned_messages = list(cleaned_messages)
        >>> for message in cleaned_messages:
        ...     if message.sections[1]['productionStatusOfProcessedData'] == 0:
        ...         message.sections[1]['productionStatusOfProcessedData'] = 4
        >>> cubes = load_pairs_from_fields(cleaned_messages)
```

marqh:
> Could you also remove this import? I think it just confuses things.

i think this is used in the example, so it is required?

lbdreyer:
> i think this is used in the example, so it is required?

Yes, but `GribMessage` and `filename` are also used in the example but not defined. However they are defined, along with `load_pairs_from_fields`, in the example on line 749, as I tried to explain in the [comment above](https://github.com/SciTools/iris/pull/2525#pullrequestreview-37317417)

marqh:
@lbdreyer I follow your point now, I had missed it the first time round.  Yes, we can treat example 2 as a continuation of example 1 and have them share teh imports

I have updated the example code to reflect this

thank you
mark

lbdreyer:
@marqh  I just noticed this is targeting the `dask` branch. Is this correct? If so I will remove the `v1.13.0` milestone. Otherwise please retarget at `master`

marqh:
hello @lbdreyer 

i agree that the 1.13 milestone was a mistake

this is targeting the dask branch, as this is where the GRIB code is in its updated form

thank you

bjlittle:
@marqh This PR should now target `master` rather than the `dask` feature branch, thanks.

PR #2526, opened 2017-05-04
Handle MaskedConstant in cube maths

djkirkham:
Numpy operations which result in a scalar masked array return an instance of `np.ma.core.MaskedConstant` that always has a dtype of `np.float64`. This means that in cube arithmetic the dtype information can be lost.

This PR is an attempt to rectify that. The idea is to run the operation on dummy arrays with the input dtypes and use the dtype of the result to fix up the dtype of the cube in the scalar masked array case.

This approach doesn't work for arbitrary functions (i.e. instances of `iris.analysis.maths.IFunc`), because (1) the dtype can depend on kwargs, and (2) the function may fail if the arguments don't meet particular requirements (e.g. shape), and it's not possible to determine these requirements for arbitrary functions. It might be worth handling the limited case where the IFunc wraps a Numpy ufunc, as this may cover the majority of use cases.

### DIFF
--- a/lib/iris/_data_manager.py
+++ b/lib/iris/_data_manager.py
@@ -205,8 +205,7 @@ class DataManager(object):
         assert state, emsg.format(self._realised_dtype)
 
         # Ensure validity of lazy data with realised dtype.
-        state = not (not self.has_lazy_data() and
-                     self._realised_dtype is not None)
+        state = self.has_lazy_data() or self._realised_dtype is None
         emsg = ('Unexpected real data with realised dtype, got '
                 'real data and realised {!r}.')
         assert state, emsg.format(self._realised_dtype)
--- a/lib/iris/analysis/maths.py
+++ b/lib/iris/analysis/maths.py
@@ -49,6 +49,8 @@ def _output_dtype(op, left_dtype, right_dtype=None, in_place=False):
     operand_dtypes = [left_dtype, right_dtype] if right_dtype is not None \
                       else [left_dtype]
     if in_place:
+        # Always return the first dtype, even if the operation would fail due
+        # to failure to cast the result.
         return operand_dtypes[0]
     key = (op, tuple(operand_dtypes))
     result = _output_dtype_cache.get(key, None)
--- a/lib/iris/analysis/maths.py
+++ b/lib/iris/analysis/maths.py
@@ -772,7 +772,7 @@ def _math_op_common(cube, operation_function, new_unit, new_dtype=None,
             and not new_cube.has_lazy_data() \
             and new_cube.data.shape == () \
             and ma.is_masked(new_cube.data):
-        new_cube.data = np.asanyarray(new_cube.data, dtype=new_dtype)
+        new_cube.data = ma.masked_array(0, 1, dtype=new_dtype)
 
     iris.analysis.clear_phenomenon_identity(new_cube)
     new_cube.units = new_unit
--- a/lib/iris/analysis/maths.py
+++ b/lib/iris/analysis/maths.py
@@ -29,6 +29,7 @@ import inspect
 
 import cf_units
 import numpy as np
+from numpy import ma
 
 from iris._deprecation import warn_deprecated
 import iris.analysis
@@ -39,6 +40,42 @@ import iris.util
 
 import dask.array as da
 from dask.array.core import broadcast_shapes
+
+
+_output_dtype_cache = {}
+
+
+def _output_dtype(op, operand_dtypes, in_place):
+    if in_place:
+        return operand_dtypes[0]
+    key = (op, tuple(operand_dtypes))
+    result = _output_dtype_cache.get(key, None)
+    if result is None:
+        arrays = [np.array([1], dtype=dtype) for dtype in operand_dtypes]
+        result = op(*arrays).dtype
+        _output_dtype_cache[key] = result
+    return result
+
+
+def _get_dtype(operand):
+    """
+    Get the numpy dtype corresponsing to the numeric data in the object
+    provided.
+
+    Args:
+
+    * operand:
+        An instance of :class:`iris.cube.Cube` or :class:`iris.coords.Coord`,
+        or a number or :class:`numpy.ndarray`.
+
+    Returns:
+        An instance of :class:`numpy.dtype`
+
+    """
+    if np.isscalar(operand):
+        operand = np.asanyarray(operand)
+
+    return operand.dtype
 
 
 def abs(cube, in_place=False):
@@ -59,8 +96,10 @@ def abs(cube, in_place=False):
         An instance of :class:`iris.cube.Cube`.
 
     """
+    _assert_is_cube(cube)
+    new_dtype = _output_dtype(np.abs, [cube.dtype], in_place)
     op = da.absolute if cube.has_lazy_data() else np.abs
-    return _math_op_common(cube, op, cube.units, in_place=in_place)
+    return _math_op_common(cube, op, cube.units, new_dtype, in_place=in_place)
 
 
 def intersection_of_cubes(cube, other_cube):
@@ -184,12 +223,15 @@ def add(cube, other, dim=None, ignore=True, in_place=False):
         An instance of :class:`iris.cube.Cube`.
 
     """
+    _assert_is_cube(cube)
+    new_dtype = _output_dtype(operator.add, [cube.dtype, _get_dtype(other)],
+                              in_place)
     if in_place:
         _inplace_common_checks(cube, other, 'addition')
         op = operator.iadd
     else:
         op = operator.add
-    return _add_subtract_common(op, 'add', cube, other, dim=dim,
+    return _add_subtract_common(op, 'add', cube, other, new_dtype, dim=dim,
                                 ignore=ignore, in_place=in_place)
 
 
@@ -224,17 +266,20 @@ def subtract(cube, other, dim=None, ignore=True, in_place=False):
         An instance of :class:`iris.cube.Cube`.
 
     """
+    _assert_is_cube(cube)
+    new_dtype = _output_dtype(operator.sub, [cube.dtype, _get_dtype(other)],
+                              in_place)
     if in_place:
         _inplace_common_checks(cube, other, 'subtraction')
         op = operator.isub
     else:
         op = operator.sub
-    return _add_subtract_common(op, 'subtract', cube, other,
+    return _add_subtract_common(op, 'subtract', cube, other, new_dtype,
                                 dim=dim, ignore=ignore, in_place=in_place)
 
 
 def _add_subtract_common(operation_function, operation_name, cube, other,
-                         dim=None, ignore=True, in_place=False):
+                         new_dtype, dim=None, ignore=True, in_place=False):
     """
     Function which shares common code between addition and subtraction
     of cubes.
@@ -246,6 +291,8 @@ def _add_subtract_common(operation_function, operation_name, cube, other,
                            to `operation_function`
     other                - the cube, coord, ndarray or number whose data is
                            used as the second argument
+    new_dtype            - the expected dtype of the output. Used in the
+                           case of scalar masked arrays
     dim                  - dimension along which to apply `other` if it's a
                            coordinate that is not found in `cube`
     ignore               - The value of this argument is ignored.
@@ -282,7 +329,7 @@ def _add_subtract_common(operation_function, operation_name, cube, other,
         coord_comp = None
 
     new_cube = _binary_op_common(operation_function, operation_name, cube,
-                                 other, cube.units, dim, in_place)
+                                 other, cube.units, new_dtype, dim, in_place)
 
     if coord_comp:
         # If a coordinate is to be ignored - remove it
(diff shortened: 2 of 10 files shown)


djkirkham:
@lbdreyer @bjlittle @dkillick

lib/iris/_data_manager.py
```diff
@@ -178,8 +178,7 @@ def _assert_axioms(self):
         assert state, emsg.format(self._realised_dtype)
 
         # Ensure validity of lazy data with realised dtype.
-        state = not (not self.has_lazy_data() and
-                     self._realised_dtype is not None)
+        state = self.has_lazy_data() or self._realised_dtype is None
```
djkirkham: I couldn't resist simplifying this logic

lib/iris/tests/unit/analysis/maths/__init__.py
```diff
@@ -159,3 +159,31 @@ def test_partial_mask_not_in_place(self):
...
+    def test_masked_constant_in_place(self):
+        # Cube in_place arithmetic operation.
+        dtype = np.int64
+        fill_value = 12345
+        dat = np.ma.masked_array(0, 1, dtype)
+        cube = Cube(dat, fill_value=fill_value)
+        res = self.cube_func(cube, 5, in_place=True)
+        self.assertMaskedArrayEqual(np.ma.masked_array(0, 1), res.data)
+        self.assertEqual(dtype, res.dtype)
+        # self.assertEqual(fill_value, res.fill_value)
```
djkirkham: Currently the fill value is being lost by maths operations, I don't know whether we need to rectify that?

bjlittle:
@djkirkham Can you rebase please.

djkirkham:
@bjlittle  Done, but the tests aren't running...

lib/iris/tests/unit/analysis/maths/__init__.py
```diff
@@ -159,3 +159,31 @@ def test_partial_mask_not_in_place(self):
...
+        self.assertIs(res, cube)
+
+    def test_masked_constant_not_in_place(self):
+        # Cube in_place arithmetic operation.
+        dtype = np.int64
+        fill_value = 12345
+        dat = np.ma.masked_array(0, 1, dtype)
```
lbdreyer: Could you rather ```import numpy.ma as ma``` then do ```ma.masked_array(0, 1, dtype)``` here?

djkirkham: OK, plus I'll change the existing usages in this file while I'm at it

bjlittle:
@djkirkham It may be an issue on travis-ci, although it's [status](https://www.traviscistatus.com/) seems fine ... however, this did just happen on my PR #2520

lib/iris/analysis/maths.py
```diff
@@ -41,6 +42,42 @@
...
+    if result is None:
+        arrays = [np.array([1], dtype=dtype) for dtype in operand_dtypes]
+        result = op(*arrays).dtype
+        _output_dtype_cache[key] = result
+    return result
+
+
+def _get_dtype(operand):
+    """
+    Get the numpy dtype corresponsing to the numeric data in the object
```
dkillick: Typo: "corresponsing" --> "corresponding".

lib/iris/analysis/maths.py
```diff
@@ -41,6 +42,42 @@
...
+
+    * operand:
+        An instance of :class:`iris.cube.Cube` or :class:`iris.coords.Coord`,
+        or a number or :class:`numpy.ndarray`.
+
+    Returns:
+        An instance of :class:`numpy.dtype`
+
+    """
+    if np.isscalar(operand):
```
dkillick: Have you checked this works with pesky edge cases such as masked constants?

lib/iris/tests/unit/analysis/maths/__init__.py
```diff
@@ -159,3 +160,31 @@ def test_partial_mask_not_in_place(self):
...
+    def test_masked_constant_in_place(self):
+        # Cube in_place arithmetic operation.
+        dtype = np.int64
+        fill_value = 12345
+        dat = ma.masked_array(0, 1, dtype)
+        cube = Cube(dat, fill_value=fill_value)
+        res = self.cube_func(cube, 5, in_place=True)
+        self.assertMaskedArrayEqual(ma.masked_array(0, 1), res.data)
+        self.assertEqual(dtype, res.dtype)
+        # self.assertEqual(fill_value, res.fill_value)
```
dkillick: Commented code: could it either be uncommented or removed, please?

lib/iris/tests/unit/analysis/maths/__init__.py
```diff
@@ -159,3 +160,31 @@ def test_partial_mask_not_in_place(self):
...
+    def test_masked_constant_not_in_place(self):
+        # Cube in_place arithmetic operation.
+        dtype = np.int64
+        fill_value = 12345
+        dat = ma.masked_array(0, 1, dtype)
+        cube = Cube(dat, fill_value=fill_value)
+        res = self.cube_func(cube, 5, in_place=False)
+        self.assertMaskedArrayEqual(ma.masked_array(0, 1), res.data)
+        self.assertEqual(dtype, res.dtype)
+        # self.assertEqual(fill_value, res.fill_value)
```
dkillick: Again.

lib/iris/analysis/maths.py
```diff
@@ -41,6 +42,42 @@
 from dask.array.core import broadcast_shapes
 
 
+_output_dtype_cache = {}
+
+
+def _output_dtype(op, operand_dtypes, in_place):
```
dkillick: It would be nice to see some unit testing of this new function.

lib/iris/analysis/maths.py
```diff
@@ -41,6 +42,42 @@
...
+    key = (op, tuple(operand_dtypes))
+    result = _output_dtype_cache.get(key, None)
+    if result is None:
+        arrays = [np.array([1], dtype=dtype) for dtype in operand_dtypes]
+        result = op(*arrays).dtype
+        _output_dtype_cache[key] = result
+    return result
+
+
+def _get_dtype(operand):
```
dkillick: It would be nice to see some unit testing of this new function.

lib/iris/analysis/maths.py
djkirkham: Yeah, but I'll add tests too

lib/iris/analysis/maths.py
djkirkham: Added.

lib/iris/analysis/maths.py
djkirkham: Added.

lib/iris/analysis/maths.py
djkirkham: Fixed.

lib/iris/tests/unit/analysis/maths/__init__.py
djkirkham: Removed.

lib/iris/tests/unit/analysis/maths/__init__.py
djkirkham: Removed.

lib/iris/analysis/maths.py
```diff
@@ -336,17 +387,12 @@ def _inplace_common_checks(cube, other, math_op):
...
-    if cube.dtype.kind in 'iu':
-        if hasattr(other, 'dtype'):
-            other_kind = other.dtype.kind
-        else:
-            other_kind = np.asanyarray(other).dtype.kind
-        if other_kind not in 'iu':
-            # Cannot coerce math op between int `cube` and float `other`
-            # back to int.
+    other_dtype = _get_dtype(other)
+    if not np.can_cast(other_dtype, cube.dtype, 'same_kind'):
```
djkirkham: I've modified this method to make it more general, and fix an issue where trying to do in-place operations on a cube with unsigned data, and signed data would fail silently.

lib/iris/tests/test_basic_maths.py
```diff
@@ -494,15 +494,6 @@ def test_multiplication_not_in_place(self):
         b = iris.analysis.maths.multiply(a, 5, in_place=False)
         self.assertIsNot(a, b)
 
-    def test_type_error(self):
```
dkillick: What happened to these tests?

lib/iris/tests/unit/analysis/maths/test__output_dtype.py
```diff
@@ -42,12 +42,17 @@ def setUp(self):
                                 operator.sub,
                                 operator.mul,
                                 operator.pow,
-                                operator.div,
+                                operator.floordiv,
```
dkillick: I think I would expect use of `truediv` here.

djkirkham: This is a list of operators which result in the same dtype as their arguments. `truediv` is in a list below.

lib/iris/analysis/maths.py
djkirkham: You're code calls the function with the arguments in the wrong order. The second argument is the one you're casting to (the cube's dtype in this case).

lib/iris/tests/test_basic_maths.py
djkirkham: Oh right. I was gonna make a comment. Personally, I don't think they're necessary. They're just saying 'don't pass the wrong type of argument to this function'. Well, you could have tests like that for any function, and that would force your to do instance checking at the top of every function. While we do instance check for the first argument to the function to insure it's a cube, we don't do it for the second argument, and the only reason these tests passed was because somewhere down the line it happened to raise a `TypeError`. My changes mean it now raises an `ArithmeticError`.

If you really want I'll add instance checks to make sure the tests pass, but I'm pretty against that idea. I suppose I would be willing to reinstate the first test, since we are doing the check for the first argument.

lib/iris/tests/unit/analysis/maths/test__get_dtype.py
```diff
@@ -0,0 +1,99 @@
...
+
+    def test_uint16(self):
+        n = 256
+        self._check_call(n, np.uint16)
+
+    def test_float16(self):
+        n = 60000.0
+        self._check_call(n, np.float16)
+
+    def test_float32(self):
```
djkirkham: OK, I'll add one.

lib/iris/tests/unit/analysis/maths/test__get_dtype.py
```diff
@@ -0,0 +1,99 @@
...
+        n = 65000.0
+        self._check_call(n, np.float32)
+
+    def test_scalar_demote(self):
+        n = np.int64(10)
+        self._check_call(n, np.uint8)
+
+    def test_array(self):
+        a = np.array([1, 2, 3], dtype=np.int16)
```
djkirkham: OK.

lib/iris/analysis/maths.py
dkillick: So it does, mystery solved. (Though I'm sure the ordering of args to `can_cast` is wrong in that case!)

lib/iris/analysis/maths.py
```diff
@@ -41,6 +42,44 @@
 from dask.array.core import broadcast_shapes
 
 
+_output_dtype_cache = {}
+
+
+def _output_dtype(op, left_dtype, right_dtype=None, in_place=False):
```
djkirkham: OK.

lib/iris/analysis/maths.py
```diff
@@ -41,6 +42,44 @@
...
+    operand_dtypes = [left_dtype, right_dtype] if right_dtype is not None \
+                      else [left_dtype]
+    if in_place:
+        # Always return the first dtype, even if the operation would fail due
+        # to failure to cast the result.
+        return operand_dtypes[0]
```
djkirkham: I agree. I'll change it.

lib/iris/tests/test_basic_maths.py
dkillick: I think the `TypeError` raised comes from NumPy in the end. But having taken a quick look I agree with your outlook that they're not necessary - if it were our own error case we would want to test it, but it is no surprise that if you try and mulitply a cube by a string it won't work!

lib/iris/tests/unit/analysis/maths/test__output_dtype.py
```diff
@@ -47,7 +47,7 @@ def setUp(self):
                                 np.subtract,
                                 np.multiply,
                                 np.power,
-                                np.divide]
+                                np.floor_divide]
```
djkirkham: I've changed this because in Python 2 `np.divide` is floor division, but in Python 3 it is true division. So in Python 3 it shouldn't be in this list, but in the one below.

I didn't want to do a Python version check to decide which list to put it in. It turns out that in Python 3, `np.divide` is the same object as `np.true_divide`, but I didn't want to rely on that either. So I've taken the easy way out and replaced it altogether.

lib/iris/analysis/maths.py
```diff
@@ -45,19 +45,50 @@
 _output_dtype_cache = {}
 
 
-def _output_dtype(op, left_dtype, right_dtype=None, in_place=False):
-    operand_dtypes = [left_dtype, right_dtype] if right_dtype is not None \
-                      else [left_dtype]
+def _output_dtype(op, first_dtype, second_dtype=None, in_place=False):
```
djkirkham: While writing the docstring I realised 'left' and 'right' didn't really make sense for the names of these arguments.

PR #2527, opened 2017-05-04
Integrate DataManager with Coord, DimCoord, and AuxCoord

dkillick:
This starts the process of integrating the `DataManager` with coord classes. Specifically, this adds the `DataManager` to the abstract coord class `iris.coords.Coord`. This is done as the initial step to fully integrating the `DataManager` with `DimCoord` and `AuxCoord`.

Still a work in progress:

 * I haven't touched `Coord.__getitem__` as there is still outstanding work in this area in #2519.
 * I've only done a light touch on `Coord.copy` as the copy code will need to be thought about in the context of `DimCoord` and `AuxCoord`.
 * I haven't even looked at the tests (and it is likely that many will now fail). I think the tests will be easier to deal with at the same time as tackling `DimCoord` and `AuxCoord`.

### DIFF
--- a/lib/iris/coords.py
+++ b/lib/iris/coords.py
@@ -1619,17 +1619,15 @@ class DimCoord(Coord):
             result += ', circular=%r' % self.circular
         return result
 
-    @property
-    def points(self):
-        """The local points values as a read-only NumPy array."""
-        return self._points_dm.data
+    def _new_points_requirements(self, points):
+        """
+        Confirm that a new set of coord points adheres to the requirements for
+        :class:`~iris.coords.DimCoord` points, being:
+            * points are 1D,
+            * points are numeric, and
+            * points are monotonic.
 
-    @points.setter
-    def points(self, points):
-        # We must realise points to check monotonicity.
-        points = as_concrete_data(points, result_dtype=self.dtype)
-
-        # Checks for 1d, numeric, monotonic
+        """
         if points.ndim != 1:
             raise ValueError('The points array must be 1-dimensional.')
         if not np.issubdtype(points.dtype, np.number):
@@ -1637,10 +1635,58 @@ class DimCoord(Coord):
         if len(points) > 1 and not iris.util.monotonic(points, strict=True):
             raise ValueError('The points array must be strictly monotonic.')
 
+    @property
+    def points(self):
+        """The local points values as a read-only NumPy array."""
+        return self._points_dm.data
+
+    @points.setter
+    def points(self, points):
+        # Checks for 1d, numeric, monotonic
+        self._new_points_requirements(points)
+
+        # We must realise points to check monotonicity.
+        points = as_concrete_data(points, result_dtype=self.dtype)
+
         # Make the array read-only.
         points.flags.writeable = False
 
         self._points_dm.data = points
+
+    def _new_bounds_requirements(self, bounds):
+        """
+        Confirm that a new set of coord points adheres to the requirements for
+        :class:`~iris.coords.DimCoord` points, being:
+            * points are 1D,
+            * points are numeric, and
+            * points are monotonic.
+
+        """
+        # Ensure the bounds are a compatible shape.
+        if self.shape != bounds.shape[:-1]:
+            raise ValueError(
+                "The shape of the bounds array should be "
+                "points.shape + (n_bounds,)")
+        # Checks for numeric and monotonic.
+        if not np.issubdtype(bounds.dtype, np.number):
+            raise ValueError('The bounds array must be numeric.')
+
+        n_bounds = bounds.shape[-1]
+        n_points = bounds.shape[0]
+        if n_points > 1:
+
+            directions = set()
+            for b_index in range(n_bounds):
+                monotonic, direction = iris.util.monotonic(
+                    bounds[:, b_index], strict=True, return_direction=True)
+                if not monotonic:
+                    raise ValueError('The bounds array must be strictly '
+                                     'monotonic.')
+                directions.add(direction)
+
+            if len(directions) != 1:
+                raise ValueError('The direction of monotonicity must be '
+                                 'consistent across all bounds')
 
     @property
     def bounds(self):
@@ -1659,31 +1705,8 @@ class DimCoord(Coord):
         if bounds is None:
             self._bounds_dm = None
         else:
-            # Ensure the bounds are a compatible shape.
-            if self.shape != bounds.shape[:-1]:
-                raise ValueError(
-                    "The shape of the bounds array should be "
-                    "points.shape + (n_bounds,)")
-            # Checks for numeric and monotonic.
-            if not np.issubdtype(bounds.dtype, np.number):
-                raise ValueError('The bounds array must be numeric.')
-
-            n_bounds = bounds.shape[-1]
-            n_points = bounds.shape[0]
-            if n_points > 1:
-
-                directions = set()
-                for b_index in range(n_bounds):
-                    monotonic, direction = iris.util.monotonic(
-                        bounds[:, b_index], strict=True, return_direction=True)
-                    if not monotonic:
-                        raise ValueError('The bounds array must be strictly '
-                                         'monotonic.')
-                    directions.add(direction)
-
-                if len(directions) != 1:
-                    raise ValueError('The direction of monotonicity must be '
-                                     'consistent across all bounds')
+            # Checks for appropriate values for the new bounds.
+            self._new_bounds_requirements(bounds)
 
             # Ensure the array is read-only.
             bounds.flags.writeable = False
@@ -1694,11 +1717,14 @@ class DimCoord(Coord):
 
     def replace_points(self, points, dtype=None, fill_value=None):
         # Cannot set fill_value for a `DimCoord`.
+        self._new_points_requirements(points)
         super(DimCoord, self).replace_points(points,
                                              dtype=dtype, fill_value=None)
 
     def replace_bounds(self, bounds=None, dtype=None, fill_value=None):
         # Cannot set fill_value for a `DimCoord`.
+        if bounds is not None:
+            self._new_bounds_requirements(bounds)
         super(DimCoord, self).replace_bounds(bounds=bounds,
                                              dtype=dtype, fill_value=None)
 
--- a/lib/iris/coords.py
+++ b/lib/iris/coords.py
@@ -37,7 +37,7 @@ import numpy as np
 import numpy.ma as ma
 
 from iris._data_manager import DataManager
-from iris._lazy_data import as_concrete_data, is_lazy_data
+from iris._lazy_data import as_concrete_data, is_lazy_data, multidim_lazy_stack
 import iris.aux_factory
 import iris.exceptions
 import iris.time
@@ -1054,14 +1054,21 @@ class Coord(six.with_metaclass(ABCMeta, CFVariableMixin)):
                 else self.core_points()
             lower, upper = item.min(), item.max()
             bounds_dtype = item.dtype
-            bounds = [lower, upper]
+            bounds = np.empty((2,), 'object')
+            bounds[0] = lower
+            bounds[1] = upper
             # Create points for the new collapsed coordinate.
             points_dtype = self.dtype
             points = [(lower + upper) * 0.5]
 
             # Create the new collapsed coordinate.
-            coord = self.copy(points=np.array(points, dtype=points_dtype),
-                              bounds=np.array(bounds, dtype=bounds_dtype))
+            if is_lazy_data(item):
+                bounds = multidim_lazy_stack(bounds)
+                coord = self.copy(points=points, bounds=bounds)
+            else:
+                bounds = np.concatenate(bounds)
+                coord = self.copy(points=np.array(points, dtype=points_dtype),
+                                  bounds=np.array(bounds, dtype=bounds_dtype))
         return coord
 
     def _guess_bounds(self, bound_position=0.5):
--- a/lib/iris/coords.py
+++ b/lib/iris/coords.py
@@ -443,18 +443,11 @@ class Coord(six.with_metaclass(ABCMeta, CFVariableMixin)):
         #: Relevant CoordSystem (if any).
         self.coord_system = coord_system
 
-        # TODO: how to set `realised_dtype`?
-        #   - Would like to use `np.asanyarray(points/bounds).dtype` - SLOW!
-        #   - Can't use `points/bounds.dtype` as might not be an array.
-        #   - Could not set realised_dtype? Or could pass it in.
-        # Contain points and bounds in `DataManager` objects.
-        self._points_dm =\
-            DataManager(points, realised_dtype=np.asanyarray(points).dtype)
-        if bounds is not None:
-            self._bounds_dm =\
-                DataManager(bounds, realised_dtype=np.asanyarray(bounds).dtype)
-        else:
-            self._bounds_dm = None
+        # Set up DataManager attributes and points and bounds values.
+        self._points_dm = None
+        self._bounds_dm = None
+        self.points = points
+        self.bounds = bounds
 
     def __getitem__(self, keys):
         """
@@ -470,7 +463,7 @@ class Coord(six.with_metaclass(ABCMeta, CFVariableMixin)):
         """
         # Fetch the points and bounds.
         points = self._points_dm.core_data()
-        if self._bounds_dm is not None:
+        if self.has_bounds():
             bounds = self._bounds_dm.core_data()
         else:
             bounds = None
@@ -492,8 +485,7 @@ class Coord(six.with_metaclass(ABCMeta, CFVariableMixin)):
         new_coord = self.copy(points=points, bounds=bounds)
         return new_coord
 
-    def copy(self, points=None, bounds=None,
-             points_dtype='none', bounds_dtype='none'):
+    def copy(self, points=None, bounds=None):
         """
         Returns a copy of this coordinate.
 
@@ -517,23 +509,14 @@ class Coord(six.with_metaclass(ABCMeta, CFVariableMixin)):
             raise ValueError('If bounds are specified, points must also be '
                              'specified')
 
-        if points is not None:
-            points_dm = self._points_dm.copy(data=points,
-                                             realised_dtype=points_dtype)
-        if bounds is not None:
-            bounds_dm = self._bounds_dm.copy(data=bounds,
-                                             realised_dtype=bounds_dtype)
-
         new_coord = copy.deepcopy(self)
         if points is not None:
-            new_coord.points = points_dm.core_data()
+            new_coord._points_dm = None
+            new_coord.points = points
             # Regardless of whether bounds are provided as an argument, new
             # points will result in new bounds, discarding those copied from
             # self.
-            if bounds is not None:
-                new_coord.bounds = bounds_dm.core_data()
-            else:
-                new_coord.bounds = None
+            new_coord.bounds = bounds
 
         return new_coord
 
@@ -577,7 +560,10 @@ class Coord(six.with_metaclass(ABCMeta, CFVariableMixin)):
             A lazy array, representing the coord bounds array.
 
         """
-        return self._bounds_dm.lazy_data()
+        lazy_bounds = None
+        if self.has_bounds():
+            lazy_bounds = self._bounds_dm.lazy_data()
+        return lazy_bounds
 
     def core_points(self):
         """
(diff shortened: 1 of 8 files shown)


djkirkham:
Are we allowing masked points now? They're not allowed in the current version are they?. In any case, I think they're only allowable to `AuxCoords`s, so `fill_value` should be a parameter to the `AuxCoord.__init__`, not the base class.

dkillick:
> Are we allowing masked coordinates now?

To catch the PR up with an offline discussion, I believe the answer is yes for `AuxCoord`s and no for `DimCoord`s.

Interestingly, in v1.12 land, if you pass a masked array to either constructor, the resultant coord in both cases has a points array composed of the masked array's data attribute as an `ndarray`; all evidence of the masked array is completely thrown away...

lib/iris/coords.py
```diff
@@ -442,31 +442,36 @@ def __init__(self, points, standard_name=None, long_name=None,
             integer or boolean masked bounds.
 
         """
-        #: CF standard name of the quantity that the coordinate represents.
+        # CF standard name of the quantity that the coordinate represents.
```
djkirkham: I think these colons make these comments get added to the docstring for the following attribute, so you might want to keep them

dkillick: Ah, could be. I just thought they looked daft :stuck_out_tongue_winking_eye:

lib/iris/coords.py
```diff
@@ -909,14 +917,25 @@ def is_compatible(self, other, ignore=None):
         return compatible
 
     @property
-    def dtype(self):
+    def points_dtype(self):
```
djkirkham: Don't we want to keep the `dtype` property, even if it returns the same as `points_dtype`, for backwards compatibility?

lib/iris/coords.py
```diff
@@ -909,14 +917,25 @@ def is_compatible(self, other, ignore=None):
...
         """
         Abstract property which returns the Numpy data type of the Coordinate.
```
djkirkham: This and the following property aren't abstract

dkillick: Ah yes, even though the docstring claims they are...

lib/iris/coords.py
dkillick: Not really, no... Do you have a good reason for wanting to maintain backward compatibility here?

lib/iris/coords.py
```diff
@@ -1604,12 +1701,21 @@ def bounds(self, bounds):
...
             # Ensure the array is read-only.
             bounds.flags.writeable = False
-
-        self._bounds = bounds
+            self._bounds_dm.data = bounds
 
     def is_monotonic(self):
         return True
 
+    def replace_points(self, points, dtype=None, fill_value=None):
```
djkirkham: This should have the same monotonicity as the `points` property. Ditto for `replace_bounds`

dkillick: Excellent points!

lib/iris/coords.py
djkirkham: Actually, yes, I've just thought of a reason. It's used [in the maths module](https://github.com/SciTools/iris/blob/dask/lib/iris/analysis/maths.py#L340) in a place where and object may be a cube, coord, numpy or scalar. There may be other places where it's used too

dkillick: So we'd have `coord.dtype` and `coord.bounds_dtype`, or `coord.dtype`, `coord.points_dtype` and `coord.bounds_dtype`? My preference is probably for only having two dtype attributes as `coord.dtype` and `coord.points_dtype` would surely always give the same answer...

lib/iris/coords.py
pp-mo: Can we maybe use this routine as a common gateway for those checks, i.e. as applied in `__init__`, `copy(points/bounds=..)` and the points/bounds setters ?

dkillick: > a common gateway for those checks

I've done something slightly different to this, which should be applicable in the same sort of manner. Let me know what you think about this approach!

pp-mo:
I can see that the proposed API extensions are a logical result of carrying over the cube data API extensions for Coord points and bounds.
However, I'm a bit horrified at the resulting proliferation of extra methods, properties and keywords !!

Discussed a bit offline with @dkillick and @bjlittle  : how can we keep this down ? ...

**One way** I proposed is simply to expose the DataManagers -- in both Cube and Coord -- and hang most of the functionality onto them.  This could mean much functionality does not need extended API at the Cube amd Coord level.
 * e.g. `cube.data_dm`, `coords.points_dm`, `coord.bounds_dm`. 
 * We would still want retain existing combination properties like `coord.fill_value`, `coord.dtype`, otherwise almost certainly too much change to existing code (cf. see the mess created in this PR by need to replace coord.dtype with coord.points_dtype).
   * for full-strength simplicity, we must add support for DataManager objects to be passed into `__init__` and `copy` inplace of the real/lazy arrays.  This means you don't need to take data out of DMs to pass it back into these calls.  It could also avoid the need for extra keywords, especially if fill_value becomes part of the DM.

 * **PRO:**
   * simple avoidance of wrapper methods +/or code duplication (DRY)
   * smaller APIs, much less clutter (especially in docs pages)
 * **CON:**
   * still need to resolve logic of constraints on Coord points and bounds, such as same dtype and fill_value
   * need to make DataManager itself public, which means it is then ***supported** ...
     * further changes are awkward
     * opens it up to mission creep for additional 'generic array' functions to be requested + added

**Another way** proposed by @bjlittle is a package of simplifying measures that apply some constraints :
  * enforce hard limitations on Coord points + bounds matching, so we only support single dtype and fill_value keywords, ***and*** insist on either both lazy or both real.  This allows :
    * single `dtype=None, fill_value=None` keywords in relevant calls (init + copy; "replace" if needed).
    * single dtype and fill_value properties   
  * `lazy_data()` and `core_data()` calls can always return (points, bounds)

pp-mo:
Both of the ideas make use of key limiting assumptions about the correspondence of points and bounds data in a coord : by dtype, fill_value and (potentially) masked/unmasked type.

All that we really ***know*** is needed right now is Coords where points and bounds have the same dtype, fill_value and masked status.
We already insist on matching shapes.

I think it is sufficiently improbable to be mixing these up that we should go ahead and assume they always match (else errors).
  * this makes the API much more manageable
  * it makes for simple consistency with existing code, e.g. it naturally still supports coord.dtype and coord.shape
  * it means code can simply ignore a lot of awkward possibilities, excluded by screening out when the data is assigned.

In principle we can change in future, if a real need to do more becomes apparent.

Clearly, even if CF were to spell out these limitations as you might expect, it would *still* be possible to have a valid netcdf file that breaks those rules. We should be careful that we can load ***something*** in those cases :
  * probably with appropriate warnings and not all data represented
  * but ***not*** falling over + failing to load anything

It would be nice if we could never actually *fail* to load any possible netcdf data, but at present we don't achieve that.
At present, already, the nc loader can be tripped up by some things that look a bit like CF constructs, but don't follow through properly.
( For instance, possibly if a bounds variable does not have the right shape ? )

pp-mo:
Clearly, even if CF were to spell out agreed limitations, it is still always *possible* to have a netcdf file that breaks the rules. 
Ideally, we should be able to load ***something*** in those cases :
  * probably with appropriate warnings and not all data represented
  * but ***not*** falling over + failing to load anything

It would be really nice if we would ***never*** actually fail to load valid netcdf files, but actually we already don't achieve that anyway :   The nc loader can be tripped up by some things that look a bit like CF constructs, but don't follow through properly.
( For instance, possibly if a bounds variable does not have the right shape ? )

dkillick:
> I'm a bit horrified at the resulting proliferation of extra methods, properties and keywords

I don't see this is particularly a problem. If a class requires a lot of methods and properties, it requires a lot of methods and properties and you have to sort of deal with it. I really don't see it is a problem when the properties are only made up of a few SLOC!

I'm really against exposing the DataManager. It was not the original intent of the DataManager (it's another thing for users to have to get their heads around rather than just being able to have a cube that appears to have data, or a coord that appears to have points). It'll likely also be more effort than it is worth...

I'm also not convinced that requiring that points and bounds have the same dtype is a good idea. Imagine if you have a coord with points as follows:

`points = np.array([0, 1, 2, 3, 4])`

What do we now do with bounds? Let's assume that the bounds are at the midpoints of each point:

 * If the coord does not have bounds then we're OK. But if `guess_bounds` is called on this coord what do we do?
 * If the coord does have bounds then the bounds _have_ to be floating-point values. So do we...
    * Cast the bounds to `int`s? This would change the values of all the bounds by flooring them.
    * Cast the points to `float`s? This would mean Iris fundamentally changed the coord, and would mean the coord would fail a round-trip test.

dkillick:
> API backwards compatibility

This is a good point in favour of minimising the number of API introductions/changes. We could change all instances of `points_<property>` to `<property>` and maintain `bounds_<property>`. Thus:

* `points_shape`, `bounds_shape` --> `shape`, `bounds_shape`;
* `points_dtype`, `bounds_dtype` --> `dtype`, `bounds_dtype`; and so on.

I've already had to do this for `dtype` as other parts of the Iris codebase expect a `coord` to have a `dtype` property, so I think this set of changes is pretty much a necessity.

lib/iris/coords.py
```diff
@@ -1513,48 +1696,38 @@ def bounds(self):
...
+            self._bounds_dm = None
+        else:
+            # Checks for appropriate values for the new bounds.
+            self._new_bounds_requirements(bounds)
 
             # Ensure the array is read-only.
             bounds.flags.writeable = False
-
-        self._bounds = bounds
+            self._bounds_dm.data = bounds
```
djkirkham: If this coord currently has no bounds, `self._bounds_dm` will be `None` and this will raise and error.

lib/iris/coords.py
```diff
@@ -1478,32 +1619,74 @@ def _repr_other_metadata(self):
...
+        self._new_points_requirements(points)
+
+        # We must realise points to check monotonicity.
+        points = as_concrete_data(points, result_dtype=self.dtype)
+
         # Make the array read-only.
         points.flags.writeable = False
 
-        self._points = points
+        self._points_dm.data = points
```
djkirkham: There's a lot of processing in the `DataManager.data` setter and I'm not sure if it guarantees that the array assigned will be the same object as the one passed in. It might be better to set `self._points_dm.data.writeable = False` after the assignment, rather than on the `points` array before.

lib/iris/coords.py
```diff
@@ -1478,32 +1619,74 @@ def _repr_other_metadata(self):
...
         return result
 
-    @property
-    def points(self):
-        """The local points values as a read-only NumPy array."""
-        return self._points
-
-    @points.setter
-    def points(self, points):
-        points = np.array(points, ndmin=1)
```
djkirkham: This line has been lost.

dkillick: Deliberately! We want new functionality (i.e. `as_concrete_data` and the `DataManager.data` setter) to ensure we have a real array rather than injecting a random call to a NumPy function...

djkirkham: Oh I see, but the problem is this will promote points to have at least one dimension, so for a coord of length one you can assign as single value. Without this line that will fail at line 1632

dkillick: Yup! It's being dealt with :+1:

djkirkham: Additionally, if the new value doesn't have attributes `ndim` and `dtype` (e.g., if it's a Python list) , `_new_points_requirements` will fail with an `AttributeError`.

lib/iris/coords.py
```diff
@@ -1513,44 +1641,30 @@ def bounds(self):
...
-                    directions.add(direction)
-
-                if len(directions) != 1:
-                    raise ValueError('The direction of monotonicity must be '
-                                     'consistent across all bounds')
+        if bounds is None:
+            self._bounds_dm = None
+        else:
+            # Ensure we have a realised array of new bounds values.
+            bounds = as_concrete_data(bounds, result_dtype=self.dtype)
```
djkirkham: Ditto my comment above regarding dtype

djkirkham:
There are still a few places where `self.points` and `self.bounds` are used where `core_points()` and `core_bounds()` could be used.

lib/iris/coords.py
```diff
@@ -1606,54 +1703,48 @@ def points(self, points):
...
 
     @bounds.setter
     def bounds(self, bounds):
         # Ensure the bounds are a compatible shape.
-        if bounds is not None:
-            if not is_lazy_data(bounds):
+        if bounds is None:
+            self._bounds_dm = None
+        else:
+            if not self.has_lazy_bounds():
```
pp-mo: I think there is a problem here that prevents us having lazy bounds at the moment
(in at least some cases ??)
You can only use `_sanitise_arrray` on *real* data, but this is applied to lazy data if existing bounds are real, or absent.

djkirkham: Yes, it should be `if not is_lazy_data(bounds):`

lib/iris/coords.py
```diff
@@ -1373,33 +1446,26 @@ def from_regular(cls, zeroth, step, count, standard_name=None,
...
-        coord.circular = circular
-
-        points = (zeroth+step) + step*np.arange(count, dtype=np.float32)
+        points = (zeroth + step) + step * np.arange(count, dtype=np.float32)
+        # Spoof `iris.util.is_regular`, which only works on coords (which we
+        # don't yet have).
+        diffs = np.diff(points)
+        if not np.allclose(diffs, np.mean(diffs), rtol=0.001):
+            points = (zeroth + step) + step * np.arange(count,
+                                                        dtype=np.float64)
```
djkirkham: Please don't copy and paste code from elsewhere. The structure of this method was fine before (as a class method).

dkillick: The code is not "copy and pasted". The code was **not** fine before, which is why I have fixed it.

dkillick:
Remove the **temporary** replication of code from `iris.util`, to save @djkirkham's delicate graces.

lib/iris/coords.py
djkirkham: Line 1453 is copy and pasted from `iris.util.regular_step`. The only thing I can see which was wrong with the code before was duplication of the attribute assignments, but you're duplicating code now anyway, and code which is more complex than simply a few assignments

dkillick: Perhaps I should have made it more explicit that that code was temporary to confirm it fixed the problem. Duplicating the code was also simpler in the short term than changing the behaviour of the public function `iris.util.regular_step`. 

> simply a few assignments

So, I am now not duplicating code (at all). The original state of this code was not a few assignments, it was a duplication of the functionality of `Coord.__init__`. When I changed the init this method no longer functioned, because it was duplicated code that was not keeping up with the API changes. This version of this method is more robust because we are using the constructor for constructing the coord rather than making a `new`, empty coord bucket and filling it by duplicating most of the functionality of the constructor. 

Further, the code is now much shorter and easier to follow. It also replicates the style of `from_coord`. And the tests pass, which is an improvement on things before I made these changes.

djkirkham: > The original state of this code was not a few assignments, it was a duplication of the functionality of `Coord.__init__`

The functionality of `Coord.__init__` _is_ just a few assignments, plus assignments to the points and bounds properties which can be avoided because we don't need the monotonitiy checks.

> When I changed the init this method no longer functioned

The reason this method no longer functioned was not because of changes to init (how could it be?) but because it was using the `_points` and `_bounds` attributes, which no longer exist.

> This version of this method is more robust because we are using the constructor for constructing the coord 

I agree with this, but this method is essentially a constructor by another name. You could have achieved the same robustness by factoring out a private method which does the duplicated functionality and which is called by both this method and the `__init__` method.

dkillick: > how could it be?

Mmm, fair point, which completely misses the more fundamental point that the problem would have been lessened if this method had just called the coord constructor in the first place, which remains the most reasonable way for this method to work.

> factoring out a private method which does the duplicated functionality and which is called by both this method and the `__init__` method

But what's the point?! You're still duplicating the functionality of the `__init__`. And worse, you're proposing adding another method (that we don't need), and getting the constructor calling methods rather than just setting instance attributes. [Let the constructor be the constructor](https://goo.gl/images/8icfDc).

djkirkham: > But what's the point?!

To avoid the monotonicity checks, which I assume was the original intent of whoever wrote this method. Although maybe they just didn't want to write a method like `is_regular` which worked for points.

> You're still duplicating the functionality of the `__init__`.

No I'm not. The `__init__` method would call that method, as I said.

I think the underlying problem here is a combination of two factors:
1) Python doesn't allow multiple `__init__` methods
2) Users can construct DimCoords directly.

Because of this, the `__init__` method _has_ to do the monotonicity checks, so if you want to avoid them you need a separate method.

djkirkham: By the way, don't let the fact that I'm continuing to discuss this let you think I'm not happy with the method as it now is. Unless anyone else has concerns I'm, happy for it to go in as-is.

dkillick: I figured we'd long since passed that point :wink: But I will be discussing this still-open query with @pp-mo when we come to look through it so that we can get a casting vote on the matter.

lib/iris/coords.py
```diff
@@ -977,12 +1050,13 @@ def serialize(x):
                 warnings.warn(msg.format(self.name()))
 
             # Create bounds for the new collapsed coordinate.
-            item = self.bounds if self.bounds is not None else self.points
-            lower, upper = np.min(item), np.max(item)
+            item = self.core_bounds() if self.has_bounds() \
+                else self.core_points()
+            lower, upper = item.min(), item.max()
             bounds_dtype = item.dtype
             bounds = [lower, upper]
```
pp-mo: This line is actually a bit broken now, in that you can't successfully create a Coord like "AuxCoord(... bounds=[lazy_lower, lazy_upper])", as that argument is neither lazy or convertible to an array (it just fails, I think).

I see this breaks no tests.  It will require a collapse on a lazy coord to trigger it, which can only be loaded from NetCDF, and we probably don't have such a case.

There are bigger pre-existing problems here though, which I would say :
  * (a) the construction [lower, upper] is actually making a shape "(2,)+points.shape", when it **should** be ""points.shape+(2,)".  So this code actually only works for 1-D coords.
  * (b) thinking a bit more, I don't think this [min, max] calculation is actually sensible for multidimensional coords.  CF isn't really clear on the meaning of bounds on multidimensional coords, so we should probably just not produce result bounds in that case.  See : http://cfconventions.org/cf-conventions/v1.6.0/cf-conventions.html#cell-boundaries
  * (c) likewise, we could have nbounds > 2 and again we have no clear interpretation, so again the result should not have bounds.

lib/iris/coords.py
```diff
@@ -977,12 +1050,13 @@ def serialize(x):
                 warnings.warn(msg.format(self.name()))
 
             # Create bounds for the new collapsed coordinate.
-            item = self.bounds if self.bounds is not None else self.points
-            lower, upper = np.min(item), np.max(item)
+            item = self.core_bounds() if self.has_bounds() \
+                else self.core_points()
```
pp-mo: :+1: for trying to make this lazy, but see potential problems, [below]() ...

lib/iris/coords.py
```diff
@@ -1338,17 +1423,18 @@ class DimCoord(Coord):
...
-                        units=coord.units, bounds=coord.bounds,
+                        units=coord.units, bounds=coord.core_bounds(),
                         attributes=coord.attributes,
                         coord_system=copy.deepcopy(coord.coord_system),
                         circular=getattr(coord, 'circular', False))
 
-    @classmethod
-    def from_regular(cls, zeroth, step, count, standard_name=None,
+    @staticmethod
+    def from_regular(zeroth, step, count, standard_name=None,
```
pp-mo: By contrast, I *agree* with going through the init so we don't bypass checks.
However, I think this should be a classmethod, and create the result by calling `cls`, as that makes it work for possible inheriting subclasses.

djkirkham: > so we don't bypass checks.

But the checks aren't needed!

dkillick:
Review actions after @pp-mo's review.

lib/iris/coords.py
```diff
@@ -544,15 +609,9 @@ def _repr_other_metadata(self):
...
-        try:
-            # With NumPy 1.7 we need to ask for 'str' formatting.
-            result = np.core.arrayprint.array2string(
-                date_obj_array, formatter={'numpystr': str}, **kwargs)
-        except TypeError:
-            # But in 1.6 we don't need to ask, and the option doesn't
-            # even exist!
-            result = np.core.arrayprint.array2string(date_obj_array, **kwargs)
-        return result
+        return np.core.arrayprint.array2string(date_obj_array,
```
dkillick: You know, I never bothered to look. I checked it worked in NumPy v1.11.1 and was happy with that.

dkillick:
:tada:

Issue #2528, opened 2017-05-04
[Dask] Dtype not changed on lazy masked integer dtype cubes


djkirkham:
Cubes with lazy masked integer data will keep their integer dtype when operations are performed on them which should result in a change of dtype (e.g. log). For example:

```python
from numpy import ma
from iris.cube import Cube
from iris.analysis.maths import log
from iris._lazy_data import as_lazy_data

a = ma.masked_array([10,20], [0,1])
la = as_lazy_data(a)
cube = Cube(la, dtype=int)
cube = log(cube)

print cube.dtype
print cube.data
```
yields:
```
int64
[2 --]
```

djkirkham:
Edit: Actually it's any lazy integer cube, not just masked ones

djkirkham:
Maybe not...

pp-mo:
Status: #2570 still open

PR #2530, opened 2017-05-09
Provide wrapper for nan_array to give lazy object correct dtype

djkirkham:
As I suggested yesterday, a wrapper class around an array to make dask think it has a particular dtype. This will make fixing the issue of getting the correct dtype for cube operations (https://github.com/SciTools/iris/issues/2528) significantly easier.

Currently the tests fail.

### DIFF
--- a/lib/iris/_lazy_data.py
+++ b/lib/iris/_lazy_data.py
@@ -29,6 +29,22 @@ import dask.context
 import numpy as np
 import numpy.ma as ma
 
+class _DtypeWrapper(object):
+    def __init__(self, array, dtype):
+        self.dtype = dtype
+        self.array = array
+
+    @property
+    def shape(self):
+        return self.array.shape
+
+    @property
+    def ndim(self):
+        return self.array.ndim
+
+    def __getitem__(self, item):
+        return self.array.__getitem__(item)
+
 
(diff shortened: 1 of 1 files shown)


djkirkham:
@pp-mo @dkillick @bjlittle

djkirkham:
This causes concatenating cubes to give incorrect results in some cases. Here's some code that deomonstrates what goes wrong at the dask level:
```
import dask.array as da
import numpy as np

class _DtypeWrapper(object):
    def __init__(self, array, dtype):
        self.dtype = dtype
        self.array = array

    @property
    def shape(self):
        return self.array.shape

    @property
    def ndim(self):
        return self.array.ndim

    def __getitem__(self, item):
        return self.array.__getitem__(item)


a = np.array([1,2])
b = np.array([3.,np.nan])
b_wrapped = _DtypeWrapper(b, dtype=int)
a_dask = da.from_array(a, 1)
b_dask = da.from_array(b_wrapped, 1)
c = da.concatenate([a_dask, b_dask])

print c.compute()
```
yields:
```
[                   1                    2                    3
 -9223372036854775808]
```
This is because the second array is being interpreted as integer (compare with `b.astype(int)`).

PR #2531, opened 2017-05-09
test for bug in aggregator

corinnebosley:
The method that iris currently uses to aggregate by months assumes that there will always be whole years to aggregate over, so when you try to aggregate by month over, for example, two years and two months, the aggregation becomes confused.  

When extra months are added, the groups over which to aggregate become uneven, and this affects the calculation which determines the order of the months.  Consequently, if you add month names to the new coordinate array, they are are returned in the wrong order (like this for the example described above):

['Jan' 'Feb' 'Sep' 'Oct' 'Nov' 'Dec' 'Jan' 'Feb' 'Mar' 'Apr' 'May' 'Jun']

This problem requires a new approach to the way we aggregate over months rather than the addition of a quick bug fix.

### DIFF
--- a/lib/iris/tests/unit/analysis/test_Aggregator.py
+++ b/lib/iris/tests/unit/analysis/test_Aggregator.py
@@ -348,9 +348,10 @@ class Test_aggregated_by(tests.IrisTest):
                                               name='month')
         agg_cube = self.time_cube.aggregated_by('month', MEAN)
         coord_categorisation.add_month(agg_cube, 'time', name='month_name')
-        self.assertEquals(agg_cube.coord('month_name').points[0], 'Jan')
-        self.assertEquals(agg_cube.coord('month_name').points[4], 'May')
-        self.assertEquals(agg_cube.coord('month_name').points[8], 'Sep')
+        month_order = ['Jan', 'Feb', 'Mar', 'Apr', 'May', 'Jun', 'Jul', 'Aug',
+                       'Sep', 'Oct', 'Nov', 'Dec']
+        for i, month in enumerate(month_order):
+            self.assertEquals(agg_cube.coord('month_name').points[i], month)
 
 
 if __name__ == "__main__":
--- a/lib/iris/tests/unit/analysis/test_Aggregator.py
+++ b/lib/iris/tests/unit/analysis/test_Aggregator.py
@@ -339,19 +339,23 @@ class Test_aggregated_by(tests.IrisTest):
                                                     (latitude, 1),
                                                     (longitude, 2)])
 
-    def test_month_name(self):
-        # This will test the ability of the aggregator to correctly calculate
-        # the order of months in the new coordinate array.
-        # The month names should appear in chronological order from January
-        # through to December.
+    def test_date_aggregator(self):
+        # Test the ability of the aggregator to correctly calculate
+        # the new points array of an aggregated time coordinate.
         coord_categorisation.add_month_number(self.time_cube, 'time',
                                               name='month')
         agg_cube = self.time_cube.aggregated_by('month', MEAN)
-        coord_categorisation.add_month(agg_cube, 'time', name='month_name')
-        month_order = ['Jan', 'Feb', 'Mar', 'Apr', 'May', 'Jun', 'Jul', 'Aug',
-                       'Sep', 'Oct', 'Nov', 'Dec']
-        for i, month in enumerate(month_order):
-            self.assertEquals(agg_cube.coord('month_name').points[i], month)
+        # Get a list of point values for the aggregated month coordinate (this
+        # should be 1 to 12):
+        month_numbers = agg_cube.coord('month').points
+        # Get a list of point values in number format for the month on the
+        # aggregated time coordinate (this should also be 1 to 12):
+        months = [agg_cube.coord('time').units.num2date(point).month
+                  for point in agg_cube.coord('time').points]
+        # Check that the months are the same for the aggregated time
+        # coordinate and the aggregated month coordinate:
+        for month_number, month in zip(month_numbers, months):
+            self.assertEquals(month_number, month)
 
 
 if __name__ == "__main__":
--- a/lib/iris/tests/unit/analysis/test_Aggregator.py
+++ b/lib/iris/tests/unit/analysis/test_Aggregator.py
@@ -321,15 +321,19 @@ class Test_lazy_aggregate(tests.IrisTest):
 
 class Test_aggregated_by(tests.IrisTest):
     def setUp(self):
+        # This sets up a cube with a time coordinate containing two years and
+        # two months.
+        # This will test the ability of the aggregator to correctly calculate
+        # the order of months in the new coordinate array.
         latitude = DimCoord(np.linspace(-90, 90, 4),
                             standard_name='latitude',
                             units='degrees')
         longitude = DimCoord(np.linspace(45, 360, 8),
                              standard_name='longitude',
                              units='degrees')
-        time_unit = unit.Unit('hours since epoch',
+        time_unit = unit.Unit('days since epoch',
                               calendar=unit.CALENDAR_360_DAY)
-        time = DimCoord(np.linspace(104040, 122040, 26),
+        time = DimCoord(np.linspace(0, 750, 26),
                         standard_name='time',
                         units=time_unit)
         self.time_cube = Cube(np.zeros((26, 4, 8), np.float32),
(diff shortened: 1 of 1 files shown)


corinnebosley:
I'm not actually sure that this test is doing what it's supposed to.  I have manually constructed these time coords to be exactly the same as those in the test data that I used in the original problem, but somehow they have turned out different in this unit test.  Will spend some time finding out what I've done differently here...

corinnebosley:
Worked it out.  Was aggregating by time instead of month.  Silly billy.

lib/iris/tests/unit/analysis/test_Aggregator.py
```diff
@@ -315,6 +319,32 @@ def test_kwarg_pass_through_combined_kwargs(self):
...
+    def setUp(self):
+        latitude = DimCoord(np.linspace(-90, 90, 4),
+                            standard_name='latitude',
+                            units='degrees')
+        longitude = DimCoord(np.linspace(45, 360, 8),
+                             standard_name='longitude',
+                             units='degrees')
+        time_unit = unit.Unit('hours since epoch',
+                              calendar=unit.CALENDAR_360_DAY)
+        time = DimCoord(np.linspace(104040, 122040, 26),
```
ajdawson: The definition of this time coordinate makes it difficult to assess on reading what the first (and subsequent times) are, which in turn makes it hard to tell what you are expecting the test result to be and why. Can you change it to something more readable like `days since ...` and have the first time coordinate value be `0` so it is clear what the coordinate looks like?

ajdawson:
Can you update the description with details of what goes wrong in place of *not executed correctly* please. It would be helpful to describe what is the expected outcome and what actually happens, and relate this to the test case you have provided.

corinnebosley:
@ajdawson Would you like a more detailed description of where and how this problem occurs?  It's quite a difficult thing to explain, but I am happy to try.

lib/iris/tests/unit/analysis/test_Aggregator.py
```diff
@@ -315,6 +319,37 @@ def test_kwarg_pass_through_combined_kwargs(self):
...
+        # two months.
+        # This will test the ability of the aggregator to correctly calculate
+        # the order of months in the new coordinate array.
+        latitude = DimCoord(np.linspace(-90, 90, 4),
+                            standard_name='latitude',
+                            units='degrees')
+        longitude = DimCoord(np.linspace(45, 360, 8),
+                             standard_name='longitude',
+                             units='degrees')
+        time_unit = unit.Unit('days since epoch',
```
ajdawson: Can you change the reference time also, for those who don't know what epoch is. Since you are testing months it would be helpful to make it clear as day what the first month actually is by using an explicit date.

lib/iris/tests/unit/analysis/test_Aggregator.py
corinnebosley: Yeah sure.  I have updated the description (although I can add more detail if necessary, but it will get long-winded I expect) and tried to simplify the test.  Do you want it simplified more?

lib/iris/tests/unit/analysis/test_Aggregator.py
corinnebosley: I don't know.  I will do my best, but it was hard enough trying to replicate this problem in the first place.  An explicit date would be lovely, wouldn't it?

ajdawson: I meant to hard code the date, so that when this test is read it is immediately clear what the start month is:

```python
>>> import cf_units
>>> cf_units.Unit('days since epoch', calendar='360_day')
Unit('days since 1970-01-01 00:00:00', calendar='360_day')
```

corinnebosley: I don't understand why you are showing me that snippet; that's the same as what is in the test.

ajdawson: I was attempting to show how you can get the actual date so you could copy paste it in place of epoch..

corinnebosley: Thanks AJ, I've done that quite a lot.  I was unsure whether the test would still work when setting an explicit date because that's how I first wrote it and it didn't work, but I've been through so many iterations of this test trying to achieve the failure mode that I was unsure which bits of syntax had which effect.

Anyway, I have added the explicit date and tried to make it a little more clear what the point of the test is.  What do you think, any better?

lib/iris/tests/unit/analysis/test_Aggregator.py
```diff
@@ -315,6 +319,39 @@ def test_kwarg_pass_through_combined_kwargs(self):
...
+    def test_month_name(self):
+        # This will test the ability of the aggregator to correctly calculate
+        # the order of months in the new coordinate array.
+        # The month names should appear in chronological order from January
+        # through to December.
+        coord_categorisation.add_month_number(self.time_cube, 'time',
+                                              name='month')
+        agg_cube = self.time_cube.aggregated_by('month', MEAN)
+        coord_categorisation.add_month(agg_cube, 'time', name='month_name')
+        self.assertEquals(agg_cube.coord('month_name').points[0], 'Jan')
```
ajdawson: You could test all 12 months in a loop here instead of just picking 3. I assume you chose to include May because it is the first one that fails though? Can you add a comment with the order the months end up in.

ajdawson: A github comment

corinnebosley: No I initially used March because that was the first one that fell through, then I realised that maybe I should spread out the checks a little across months in case I apply a fix which works for March but changes the order later on.  A loop sounds like a good idea though, I will implement that.

corinnebosley: I updated the first comment with the order the months end up in, I thought that was what you meant when you asked me to update the description.  They end up like this:

['Jan' 'Feb' 'Sep' 'Oct' 'Nov' 'Dec' 'Jan' 'Feb' 'Mar' 'Apr' 'May' 'Jun']

ajdawson: Ok I wasn't sure if that was correct since it isn't just an order problem, and you mentioned ordering in the description. It seems July and August are missing and Jan and Feb are repeated, which is really wrong for aggregation no?

corinnebosley: Well it is, yes.  The bit which decides how to group the months over which to aggregate depends on a mean calculation of the months.  I don't really understand why it is done like this or how this normally calculates the correct grouping as it is, but when there are an uneven number of months in the calculation it most definitely produces the wrong answer.

I'll try and explain that a bit better; the mean calculation takes each coord point for January and takes the average, then does the same for February, then March, etc.  In the case described here, there are three coord points for January, three for February, and only two for March and the rest of the months.  That means that when it takes the mean coord point from these groups, it looks like this:
[ 360.  390.  240.  270.  300.  330.  360.  390.  420.  450.  480.  510.]

The third point in this list is much lower than the first two, and is translated into a date coordinate as being in September of the previous year, which is clearly wrong, as are the rest of the data points determined here.

As I don't really understand how this whole section works anyway, I am really struggling with how to approach a fix for this.

djkirkham:
I don't know if this should be considered a bug. What it comes down to is how iris constructs coordinates on the same dimension as the one being aggregated over. Dates aren't considered a special case (and IMO they shouldn't be).

In the example in your test, when you aggregate over the `month` coordinate Iris tries to calculate a new `time` coordinate by grouping the points based on the value of `month` at that point (that is, the dates which all have the same month).

The bounds for each point on the new `time` coordinate will be the first lower bound in the group and the last upper bound in the group. The point is the average of these two values. For the months where you have three dates (Jan and Feb), the upper and lower bound are about two years apart, so their average (the point) is some time in the same month (of the middle year, 1983). For the points where you have only two dates, these dates are only one year apart, so the average is offset by about six months.

So when you add the month name after aggregating, the values appear 'out of order'. But the months would still be 'wrong' even if you had only two dates for each month: they'd all be offset by six months.

Without making dates as special case, I don't think there's a way to fix this without changing the way iris constructs these coordinates.

corinnebosley:
I had a feeling that might be the case.  If you want to get involved in a conversation about this, there is a meeting at 10:30am on Thursday in E2-3.  All welcome who are interested.

ajdawson:
I misread your test, I thought you were saying there was more than one result for the repeated months, but now I see you are adding a month name after aggregation. In general you should not expect this to behave in the described way, once you aggregate over times you can't expect the time coordinate (any any coordinate derived from it like month_name) to have a predictable order. Once you have aggregated over the month coordinate, the only coordinate you should be using to deal with time is the month coordinate.

Thanks for the detail @djkirkham, I am in agreement with you.

corinnebosley:
@ajdawson What do you mean you are in agreement with @djkirkham?  Agreement about what?

ajdawson:
That this isn't a bug.

PR #2534, opened 2017-05-09
Dtype concat fail

djkirkham:
Demonstrates the issue with `PPDataProxy` specifying the 'wrong' dtype. The data is reinterpreted as integer and the nans, and hence the masked points, are lost.

Even if this test is adopted it should not be merged as-is, since the test_data sha will need to be changed (see https://github.com/SciTools/iris-test-data/pull/50).

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -20,7 +20,7 @@ git:
   depth: 10000
 
 install:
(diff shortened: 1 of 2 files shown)


djkirkham:
@bjlittle

djkirkham:
Test now included in https://github.com/SciTools/iris/pull/2543

PR #2535, opened 2017-05-10
plotting NZ rotated pole extents

marqh:
this is a work around for a long standing bug related to New Zealand model output and Cartopy extents

Cartopy defines the extent of a RotatedPole as -180 <= lambda < 180 and this is immutable

New Zealand modellers use a rotated pole CRS with an extent from 0 <= lambda < 360 with the area of interest around 180

The result is a map with a small strip of data on each side and a massive missing section in the middle

This work around in the shared plotting code identifies this condition and requests an adapted projection space from Cartopy to ensure a sensible map extent is set

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -22,7 +22,7 @@ git:
   depth: 10000
 
 install:
-  - export IRIS_TEST_DATA_REF="942e6344d6e6d00c85d824dadd49b5d6f57f4b4a"
+  - export IRIS_TEST_DATA_REF="7c0e32c8812b464e467a9555bdc25dc1e0c5be0c"
   - export IRIS_TEST_DATA_SUFFIX=$(echo "${IRIS_TEST_DATA_REF}" | sed "s/^v//")
 
   # Install miniconda
--- a/lib/iris/plot.py
+++ b/lib/iris/plot.py
@@ -45,6 +45,7 @@ import numpy.ma as ma
 import iris.cube
 import iris.analysis.cartography as cartography
 import iris.coords
+import iris.coord_systems
 from iris.exceptions import IrisError
 # Importing iris.palette to register the brewer palettes.
 import iris.palette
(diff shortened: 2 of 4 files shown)


lib/iris/plot.py
```diff
@@ -662,7 +663,18 @@ def _ensure_cartopy_axes_and_determine_kwargs(x_coord, y_coord, kwargs):
     # Ensure the current axes are a cartopy.mpl.GeoAxes instance.
     axes = kwargs.get('axes')
     if axes is None:
-        _replace_axes_with_cartopy_axes(cartopy_proj)
+        if (isinstance(cs, iris.coord_systems.RotatedGeogCS) and
```
cpelley: What bout an `iris.coord_system.GeogCS` coordinate system??

Are we at risk of producing unexpected bespoke behaviour here?
Is there a means to provide a more generalised solution?

.travis.yml
```diff
@@ -22,7 +22,7 @@ git:
   depth: 10000
 
 install:
-  - export IRIS_TEST_DATA_REF="942e6344d6e6d00c85d824dadd49b5d6f57f4b4a"
+  - export IRIS_TEST_DATA_REF="7c0e32c8812b464e467a9555bdc25dc1e0c5be0c"
```
marqh: > Did we really need to add this new dataset?

we already have, so this is a little out of scope.  I am keen to have access to this data set for testing, and this test may be the first of a number.  Plotting these data sets has been a particular challenge for some of our potential users

lib/iris/plot.py
marqh: it is plausible that we could include GeogCS in this behaviour control.  It is less of an issue in our user community as the use of GeogCS for limited area models is unusual

in my view this is a corner case, it is not a generalised problem.  There are limitations in Cartopy's interfaces and usage patterns with particular data sets which conflict in this small set of cases, so I am looking for a limited scope fix, not a generalised solution

@cpelley Please may you clarify: are you explicitly asking for this behaviour to be provided for and tested for `GeogCS` as part of this change?

cpelley: > It is less of an issue in our user community as the use of GeogCS for limited area models is unusual

True, one of my motivations is that having a specific caveat in the code can make it difficult to understand the context at a later date.

> @cpelley Please may you clarify: are you explicitly asking for this behaviour to be provided for and tested for GeogCS as part of this change?

Not at all.  I don't want to get in the way of the clear benefit this provides as it is.

Would be good to think about the iris-cartopy relationship with regarding to coordinate systems in future.  Doesn't feel right.  That's for another time.

Issue #2536, opened 2017-05-10
pandas version


marqh:
i have pinned pandas to <0.20 due to test failures of the form:

```
======================================================================

ERROR: test_series_simple (iris.tests.test_pandas.TestSeriesAsCube)

----------------------------------------------------------------------

Traceback (most recent call last):

  File "/home/travis/miniconda/envs/test-environment/lib/python2.7/site-packages/Iris-1.12.0.dev0-py2.7-linux-x86_64.egg/iris/tests/test_pandas.py", line 309, in test_series_simple

    iris.pandas.as_cube(series),

  File "/home/travis/miniconda/envs/test-environment/lib/python2.7/site-packages/Iris-1.12.0.dev0-py2.7-linux-x86_64.egg/iris/pandas.py", line 111, in as_cube

    calendars.get(0, None))

  File "/home/travis/miniconda/envs/test-environment/lib/python2.7/site-packages/Iris-1.12.0.dev0-py2.7-linux-x86_64.egg/iris/pandas.py", line 52, in _add_iris_coord

    if isinstance(points, pandas.tseries.index.DatetimeIndex):

AttributeError: 'module' object has no attribute 'index'

----------------------------------------------------------------------

Ran 3970 tests in 342.779s

FAILED (SKIP=9, errors=14)
```

this needs further investigation to remove the pin

PR #2537, opened 2017-05-10
pin pandas #2536

marqh:
pin pandas version due to perceived incompatibility with pandas 0.20

### DIFF
--- a/conda-requirements.txt
+++ b/conda-requirements.txt
@@ -30,7 +30,7 @@ iris-grib
 esmpy>=7.0
 gdal
 libmo_unpack
-pandas
+pandas<0.20
 pyugrid
 mo_pack
 python-stratify


marqh:
#2536 is raised to identify this pinning and work to remedy it

ocefpaf:
@marqh #2538 should fix this. Iris is already pinned to an old `netcdf4-python` version. Pinning `pandas` too is a deal breaker :smile:

marqh:
Hi @ocefpaf 

#2538 looks like a limited scope fix and is certainly better than a pin, I agree

> Iris is already pinned to an old netcdf4-python version.

I don't think this is the case, iris is testing with the latest available netcdf4-python (currently 1.2.7) and passing its test suite

https://github.com/conda-forge/iris-feedstock/blob/master/recipe/meta.yaml#L41
has a pin, but this does not seem to be required, to me

ocefpaf:
> I don't think this is the case

See https://github.com/SciTools/iris/issues/2322

I guess that the tests did not catch the incompatibility that @dkillick found on #2322

marqh:
obsoleted by #2538

PR #2337, opened 2017-02-06
Integration test for PartialDateTime-based extraction

dkillick:
A new integration test that tests the interface to `netcdftime` when performing an extract from an Iris cube using a `PartialDateTime`.

Specifically, this tests a previously uncaught failure mode originally reported in https://github.com/SciTools/conda-recipes-scitools/pull/229, and addresses one of the review actions on that PR.

PR #2539, opened 2017-05-10
test partial datetime

marqh:
replaces #2337 which is failing to set up the test environment properly

### DIFF
--- a/lib/iris/tests/integration/test_PartialDateTime.py
+++ b/lib/iris/tests/integration/test_PartialDateTime.py
@@ -27,9 +27,6 @@ import iris
 from iris.time import PartialDateTime
 
 
-iris.FUTURE.cell_datetime_objects = True
-
-
 class Test(tests.IrisTest):
 
     @tests.skip_data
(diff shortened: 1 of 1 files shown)


corinnebosley:
@marqh I'm sorry, maybe I'm being daft, but how is this different from https://github.com/SciTools/iris/pull/2337?

marqh:
> I'm sorry, maybe I'm being daft, but how is this different from #2337?

it's remarkably similar, it is not you at all

however, #2337 doesn't seem to be able to run its tests, it was like the merge it had for the test was so far behind, it wouldn't run, so I pinched the test and made a new PR so I could iterate on the issue

ocefpaf:
I see [`netcdf4 1.2.7`](https://travis-ci.org/SciTools/iris/jobs/230807770#L337) and the new test is passing. I guess we can unpin `netcdf4-python` in the conda package then. Thanks @marqh!

@dkillick since you raised this initial can you give us the green to remove the pin?

dkillick:
> since you raised this initial can you give us the green to remove the pin?

@ocefpaf do you happen to know what it is that has been changed that means Iris now supports the netcdf4 v1.2.7?

Happy to merge this though :smile:

ocefpaf:
> @ocefpaf do you happen to know what it is that has been changed that means Iris now supports the netcdf4 v1.2.7?

I am guessing that these tests changes: https://github.com/SciTools/iris/pull/2440/files#diff-5cb5e61c3251e6b6db0c1383719cd552

marqh:
@ocefpaf I think that the pin is on the v1.12.0 build

I am hoping to expedite a 1.13 release, is it worth waiting for that, which will have the changes and this test as part of the release, and removing the pin in the updated recipe for v1.13.0?

ocefpaf:
> I am hoping to expedite a 1.13 release, is it worth waiting for that, which will have the changes and this test as part of the release, and removing the pin in the updated recipe for v1.13.0?

It is worth. Not that the the tests are important to the conda-forge package (we don't run them), and in theory we could unpin b/c there is no actual iris code change regarding netcdf4 compatibility. However, there are two patches in 1.12 conda package already, and a new release would be nice so we can drop them.

PR #2543, opened 2017-05-10
Ensure declared dtype of data proxy objects matches returned dtype

djkirkham:
If a Dask array is created from an array-like with a `dtype` attribute which is different from the dtype of an array returned by its `__getitem__`, this can cause problems, notably with `concatenate` and `stack`.

This PR ensures that the dtype attribute of `NetCDFDataProxy` and `PPDataProxy` objects can be promoted to from the actual dtype of the arrays returned by `__getitem__`, or that an error is raised if not. In practice this means that if the attribute is of kind int, uint or byte, and during a `__getitem__` call masked data is discovered (which would be convered to NaNs with a dtype of f8), an error is raised.

I've made changes so that when constructing a `NetCDFDataProxy`, the actual expected dtype of the arrays returned by `__getitem__` should be passed in. For `PPDataProxy`, the dtype property is derived from the one passed in. I'm not particularly happy about this difference, but it's hard to avoid.

When a `NetCDFDataProxy` is constructed for use in a `Cube`, the dtype  passed in is that of the nan array which would be returned if the variable contained masked data. The actual dtype of the data is stored by the cube's dtype. In other cases (i.e coordinates and cell measures), the dtype passed in is the dtype of the NetCDF variable itself. This shouldn't cause problems; masked data is not supported in coordinates, and integer data is not supported in cell measures.

`PPDataProxy` objects are only ever used for cube data, so no dtype information is lost as it is stored in `Cube.dtype`.

### DIFF
--- a/lib/iris/fileformats/grib/__init__.py
+++ b/lib/iris/fileformats/grib/__init__.py
@@ -139,7 +139,7 @@ class GribWrapper(object):
     def __init__(self, grib_message, grib_fh=None):
         """Store the grib message and compute our extra keys."""
         self.grib_message = grib_message
-        self.data_dtype = np.zeros(0).dtype
+        self.realised_dtype = np.zeros(0).dtype
 
         if self.edition != 1:
             emsg = 'GRIB edition {} is not supported by {!r}.'
@@ -178,7 +178,7 @@ class GribWrapper(object):
             # The byte offset requires to be reset back to the first byte
             # of this message. The file pointer offset is always at the end
             # of the current message due to the grib-api reading the message.
-            proxy = GribDataProxy(shape, self.data_dtype, grib_fh.name,
+            proxy = GribDataProxy(shape, self.realised_dtype, grib_fh.name,
                                   offset - message_length)
             self._data = as_lazy_data(proxy)
         else:
--- a/lib/iris/fileformats/grib/__init__.py
+++ b/lib/iris/fileformats/grib/__init__.py
@@ -139,6 +139,7 @@ class GribWrapper(object):
     def __init__(self, grib_message, grib_fh=None):
         """Store the grib message and compute our extra keys."""
         self.grib_message = grib_message
+        self.data_dtype = np.zeros(0).dtype
 
         if self.edition != 1:
             emsg = 'GRIB edition {} is not supported by {!r}.'
@@ -177,7 +178,7 @@ class GribWrapper(object):
             # The byte offset requires to be reset back to the first byte
             # of this message. The file pointer offset is always at the end
             # of the current message due to the grib-api reading the message.
-            proxy = GribDataProxy(shape, np.zeros(0).dtype, grib_fh.name,
+            proxy = GribDataProxy(shape, self.data_dtype, grib_fh.name,
                                   offset - message_length)
             self._data = as_lazy_data(proxy)
         else:
--- a/lib/iris/fileformats/grib/message.py
+++ b/lib/iris/fileformats/grib/message.py
@@ -120,6 +120,10 @@ class GribMessage(object):
         # Default for fill value is None.
         return None
 
+    @property
+    def data_dtype(self):
+        return np.dtype('f8')
+
     def core_data(self):
         return self.data
 
--- a/lib/iris/fileformats/grib/message.py
+++ b/lib/iris/fileformats/grib/message.py
@@ -164,7 +164,7 @@ class GribMessage(object):
                 shape = (grid_section['numberOfDataPoints'],)
             else:
                 shape = (grid_section['Nj'], grid_section['Ni'])
-            proxy = _DataProxy(shape, np.dtype('f8'), np.nan,
+            proxy = _DataProxy(shape, self.data_dtype, np.nan,
                                self._recreate_raw)
             data = as_lazy_data(proxy)
         else:
--- a/lib/iris/fileformats/grib/message.py
+++ b/lib/iris/fileformats/grib/message.py
@@ -121,7 +121,7 @@ class GribMessage(object):
         return None
 
     @property
-    def data_dtype(self):
+    def realised_dtype(self):
         return np.dtype('f8')
 
     def core_data(self):
@@ -164,7 +164,7 @@ class GribMessage(object):
                 shape = (grid_section['numberOfDataPoints'],)
             else:
                 shape = (grid_section['Nj'], grid_section['Ni'])
-            proxy = _DataProxy(shape, self.data_dtype, np.nan,
+            proxy = _DataProxy(shape, self.realised_dtype, np.nan,
                                self._recreate_raw)
             data = as_lazy_data(proxy)
         else:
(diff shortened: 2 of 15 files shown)


djkirkham:
This isn't currently working for pp files because the original dtype isn't passed into the cube `__init__`. The trouble is, the data proxy is created in a completely different place in the code to the cube, and by the time the cube is created the proxy is already wrapped in a dask array, so the dtype is lost

djkirkham:
I've updated the code to add a new property - `data_dtype` to the 'field' classes (e.g. PPField). This contains the dtype of the underlying data, and is used as the `dtype` argument when constructing cubes in `rules.py`.

I've also added a test for pp files. This relies on a new test data file which is added in https://github.com/SciTools/iris-test-data/pull/50. The sha in `.travis.yml` will need to be updated once that PR is merged.

lib/iris/fileformats/pp.py
```diff
@@ -1297,6 +1301,16 @@ def core_data(self):
         return self._data
 
     @property
+    def data_dtype(self):
+        return self._data.dtype \
+            if self._data_dtype is None \
+            else self._data_dtype
```
dkillick: Hang on, won't this always return `self._data_dtype`, whether it's `None` or not??

lib/iris/fileformats/rules.py
```diff
@@ -906,7 +906,7 @@ def _make_cube(field, converter):
                           cell_methods=metadata.cell_methods,
                           dim_coords_and_dims=metadata.dim_coords_and_dims,
                           aux_coords_and_dims=metadata.aux_coords_and_dims,
-                          fill_value=field.bmdi, dtype=field.core_data().dtype)
+                          fill_value=field.bmdi, dtype=field.data_dtype)
```
dkillick: Hm, I changed a test for this in #2557 -- we will need to look out for follow-on failures once the dust has settled between these two PRs.

lib/iris/fileformats/um/_fast_load_structured_fields.py
```diff
@@ -99,6 +99,10 @@ def core_data(self):
         return self.data
 
     @property
+    def data_dtype(self):
+        return np.result_type(*[field.data_dtype for field in self._fields])
```
dkillick: This would seem to cause a change in behaviour from how things were before. Will this introduce the possibilty casting of the fields' data attributes that don't have the "highest" dtype?

lib/iris/tests/integration/test_pp.py
```diff
@@ -711,5 +711,27 @@ def test_singular_longitude(self):
         self.assertTrue(cube.coord('longitude').has_bounds())
 
 
+@tests.skip_data
+class TestLoadPartialMask(tests.IrisTest):
+    def test_data(self):
+        # Ensure that fields merge correctly where one has a mask and one
+        # doesn't
```
dkillick: Perhaps we could have a full stop at the end of this second comment line. At the moment it looks like you forgot to finish the comment!

lib/iris/fileformats/grib/__init__.py
```diff
@@ -139,6 +139,7 @@ class GribWrapper(object):
     def __init__(self, grib_message, grib_fh=None):
         """Store the grib message and compute our extra keys."""
         self.grib_message = grib_message
+        self.data_dtype = np.zeros(0).dtype
```
djkirkham: I moved it here because I need an attribute called `data_dtype` one this object. I don't think the change would cause any issue . I assume `np.zeros(0).dtype` will always give the same result.

lib/iris/fileformats/grib/message.py
```diff
@@ -160,7 +164,7 @@ def data(self):
                 shape = (grid_section['numberOfDataPoints'],)
             else:
                 shape = (grid_section['Nj'], grid_section['Ni'])
-            proxy = _DataProxy(shape, np.dtype('f8'), np.nan,
+            proxy = _DataProxy(shape, self.data_dtype, np.nan,
```
djkirkham: Do you mean the `data_dtype` on `GribWrapper`? As I understand it `GribWrapper` is used for the now deprecated loader, and this is for the new one. I don't think they're ever used together.

dkillick: Seems a bit odd we have two different default `data_dtype` values across the two though...

lib/iris/fileformats/pp.py
djkirkham: Look closer. The first line is the `dtype` attribute of `self._data`. Maybe it's a bit confusing...

lib/iris/fileformats/um/_fast_load_structured_fields.py
djkirkham: I think it's ok. `data_dtype` is used in place of `core_data().dtype`. If you look at how that is derived, it's basically by stacking the underlying fields' data arrays together. The result of that stacking has the same dtype as `np.result_type(...)` gives.

lib/iris/tests/integration/test_pp.py
djkirkham: Hehe. Ok

djkirkham: then.

lib/iris/fileformats/pp.py
dkillick: > Look closer.

I did look closer! It still took three goes for me to spot the difference even _after_ you assured me there was a difference!

My initial thought on this method was that it is OK to set a temporary variable in some cases, such as where it would require fewer lines if done or improve legibility :wink:

lib/iris/fileformats/grib/message.py
djkirkham: Yeah... I just went with what was already there to be honest.

lib/iris/fileformats/pp.py
djkirkham: Well, I dunno if it would improve it. You'd still have the confusion because of the similarly named variables. Maybe I should just rename `_data_dtype` to something else. In fact, now that I think about it, the name might be confusing anyway, because the data already _has_ a dtype, but it's not the one we want. Maybe I should call it `realised_dtype`, like in `DataManager`.

dkillick: > Maybe I should call it `realised_dtype`

I'm a fan of calling this `realised_dtype` :+1:

lib/iris/fileformats/grib/message.py
```diff
@@ -152,6 +152,7 @@ def data(self):
                                    'unsupported quasi-regular grid.')
 
         if template in (0, 1, 5, 12, 20, 30, 40, 90):
+        if template in (0, 1, 5, 12, 20, 30, 40, 90):
```
dkillick: What happened here?

djkirkham: Whoops...

dkillick:
Thanks @djkirkham!

PR #2545, opened 2017-05-11
Update regular coord checking

dkillick:
The existing checks for whether a coord is regular in `iris.util` quite reasonably only works on coords. However in #2527 I need to check whether just a set of _points_ is regular, which the existing functionality could not achieve.

To deal with this I have added a new function, `iris.util.points_step`, which checks whether there is a regular step between a set of points (a NumPy array). This function represents the part of `iris.util.regular_step` that dealt with checking the incoming coord's points. I've also added tests for the `iris.util` functions `is_regular`, `regular_step`, and `points_step`.

Note: #2527 now depends on this PR.

### DIFF
--- a/lib/iris/tests/unit/util/test__coord_regular.py
+++ b/lib/iris/tests/unit/util/test__coord_regular.py
@@ -1,0 +1,116 @@
+# (C) British Crown Copyright 2017, Met Office
+#
+# This file is part of Iris.
+#
+# Iris is free software: you can redistribute it and/or modify it under
+# the terms of the GNU Lesser General Public License as published by the
+# Free Software Foundation, either version 3 of the License, or
+# (at your option) any later version.
+#
+# Iris is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU Lesser General Public License for more details.
+#
+# You should have received a copy of the GNU Lesser General Public License
+# along with Iris.  If not, see <http://www.gnu.org/licenses/>.
+"""
+Test elements of :mod:`iris.util` that deal with checking coord regularity.
+Specifically, this module tests the following functions:
+
+  * :func:`iris.util.is_regular`,
+  * :func:`iris.util.regular_step`, and
+  * :func:`iris.util.points_step`.
+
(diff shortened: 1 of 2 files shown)


lib/iris/tests/unit/util/test__coord_regular.py
```diff
@@ -0,0 +1,116 @@
...
+
+
+class Test_regular_step(tests.IrisTest):
+    def test_basic(self):
+        dtype = np.float64
+        points = np.arange(5, dtype=dtype)
+        coord = DimCoord(points)
+        expected = np.mean(np.diff(points))
+        result = regular_step(coord)
+        self.assertArrayEqual(expected, result)
```
djkirkham: The result should be a scalar so this could just be `self.assertEqual`

lib/iris/tests/unit/util/test__coord_regular.py
```diff
@@ -0,0 +1,116 @@
...
+        with self.assertRaisesRegexp(CoordinateNotRegularError, exp_emsg):
+            regular_step(coord)
+
+
+class Test_points_step(tests.IrisTest):
+    def test_regular_points(self):
+        regular_points = np.arange(5)
+        exp_avdiff = np.mean(np.diff(regular_points))
+        result_avdiff, result = points_step(regular_points)
+        self.assertArrayEqual(exp_avdiff, result_avdiff)
```
djkirkham: Ditto

lib/iris/tests/unit/util/test__coord_regular.py
dkillick: So it is. Silly me for expecting NumPy functionality to return an array...

lib/iris/tests/unit/util/test__coord_regular.py
```diff
@@ -0,0 +1,116 @@
...
+        self.assertTrue(result)
+
+    def test_irregular_points(self):
+        irregular_points = np.array([2, 5, 1, 4])
+        exp_avdiff = np.mean(np.diff(irregular_points))
+        result_avdiff, result = points_step(irregular_points)
+        self.assertArrayEqual(exp_avdiff, result_avdiff)
```
dkillick: @djkirkham and here (which I've also fixed).

dkillick:
Thanks @djkirkham!

PR #2547, opened 2017-05-11
More Coordinate tests for variously lazy/real points and bounds

pp-mo:
Replaces https://github.com/dkillick/iris/pull/11
Since re-target across repos is not allowed ?

### DIFF
--- a/lib/iris/coords.py
+++ b/lib/iris/coords.py
@@ -1697,15 +1697,25 @@ class AuxCoord(Coord):
         return new_coord
 
     def _sanitise_array(self, src, ndmin):
-        # Ensure the array is writeable.
-        # NB. Returns the *same object* if src is already writeable.
-        result = np.require(src, requirements='W')
-        # Ensure the array has enough dimensions.
-        # NB. Returns the *same object* if result.ndim >= ndmin
-        result = np.array(result, ndmin=ndmin, copy=False)
-        # We don't need to copy the data, but we do need to have our
-        # own view so we can control the shape, etc.
-        result = result.view()
+        if is_lazy_data(src):
+            # Lazy data : just ensure ndmin requirement.
+            ndims_missing = ndmin - src.ndim
+            if ndims_missing <= 0:
+                result = src
+            else:
+                extended_shape = tuple([1] * ndims_missing + list(src.shape))
+                result = src.reshape(extended_shape)
+        else:
+            # Real data : a few more things to do in this case.
+            # Ensure the array is writeable.
+            # NB. Returns the *same object* if src is already writeable.
+            result = np.require(src, requirements='W')
+            # Ensure the array has enough dimensions.
+            # NB. Returns the *same object* if result.ndim >= ndmin
+            result = np.array(result, ndmin=ndmin, copy=False)
+            # We don't need to copy the data, but we do need to have our
+            # own view so we can control the shape, etc.
+            result = result.view()
         return result
 
     @property
@@ -1720,11 +1730,7 @@ class AuxCoord(Coord):
         # Ensure points has an ndmin of 1 and is either a numpy or lazy array.
         # This will avoid Scalar coords with points of shape () rather
         # than the desired (1,).
-        if is_lazy_data(points):
-            if points.shape == ():
-                points = points.reshape(1,)
-        else:
-            points = self._sanitise_array(points, 1)
+        points = self._sanitise_array(points, 1)
 
         # Set or update DataManager.
         if self._points_dm is None:
@@ -1753,8 +1759,7 @@ class AuxCoord(Coord):
         if bounds is None:
             self._bounds_dm = None
         else:
-            if not is_lazy_data(bounds):
-                bounds = self._sanitise_array(bounds, 2)
+            bounds = self._sanitise_array(bounds, 2)
             if self.shape != bounds.shape[:-1]:
                 raise ValueError("Bounds shape must be compatible with points "
                                  "shape.")
--- a/lib/iris/tests/unit/coords/__init__.py
+++ b/lib/iris/tests/unit/coords/__init__.py
@@ -17,7 +17,9 @@
 """
 Unit tests for the :mod:`iris.coords` module.
 
-Provides test methods and classes common to test_AuxCoord and test_DimCoord.
+Provides test methods and classes common to
+:class:`~iris.tests.unit.coords.test_AuxCoord` and 
+:class:`~iris.tests.unit.coords.test_DimCoord`.
 
 """
 from __future__ import (absolute_import, division, print_function)
@@ -46,7 +48,7 @@ def setup_test_arrays(self, shape):
 
 
 def is_real_data(array):
-    # A parallel to 'is_lazy_data'.
+    # A parallel to :func:`iris._lazy_data.is_lazy_data`.
     # Not just "not lazy" : ensure it is a 'real' array (i.e. numpy).
     return isinstance(array, np.ndarray)
 
--- a/lib/iris/tests/unit/coords/__init__.py
+++ b/lib/iris/tests/unit/coords/__init__.py
@@ -18,7 +18,7 @@
 Unit tests for the :mod:`iris.coords` module.
 
 Provides test methods and classes common to
-:class:`~iris.tests.unit.coords.test_AuxCoord` and 
+:class:`~iris.tests.unit.coords.test_AuxCoord` and
 :class:`~iris.tests.unit.coords.test_DimCoord`.
 
 """
--- a/lib/iris/tests/unit/coords/__init__.py
+++ b/lib/iris/tests/unit/coords/__init__.py
@@ -1,4 +1,4 @@
-# (C) British Crown Copyright 2013 - 2015, Met Office
+# (C) British Crown Copyright 2013 - 2017, Met Office
 #
 # This file is part of Iris.
 #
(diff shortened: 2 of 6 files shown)


dkillick:
@pp-mo don't forget that I have made a number of comments on https://github.com/dkillick/iris/pull/11 that will need to be addressed.

lib/iris/coords.py
```diff
@@ -1697,15 +1697,25 @@ def from_coord(coord):
...
-        # Ensure the array is writeable.
-        # NB. Returns the *same object* if src is already writeable.
-        result = np.require(src, requirements='W')
-        # Ensure the array has enough dimensions.
-        # NB. Returns the *same object* if result.ndim >= ndmin
-        result = np.array(result, ndmin=ndmin, copy=False)
-        # We don't need to copy the data, but we do need to have our
-        # own view so we can control the shape, etc.
-        result = result.view()
+        if is_lazy_data(src):
```
dkillick: @pp-mo these changes do not feel like they belong in this PR: they're making changes other than adding test coverage to the `Coord` classes. The changes look good, but could they be raised in a separate PR perchance?

lib/iris/coords.py
```diff
@@ -1720,11 +1730,7 @@ def points(self, points):
         # Ensure points has an ndmin of 1 and is either a numpy or lazy array.
         # This will avoid Scalar coords with points of shape () rather
         # than the desired (1,).
-        if is_lazy_data(points):
-            if points.shape == ():
-                points = points.reshape(1,)
-        else:
-            points = self._sanitise_array(points, 1)
+        points = self._sanitise_array(points, 1)
```
dkillick: It feels like you should go the whole hog and integrate these changes into `DimCoord` too. But, as per my previous comment, I'd prefer if that were done in another PR.

lib/iris/tests/unit/coords/test_AuxCoord.py
```diff
@@ -0,0 +1,764 @@
...
+
+    def test_lazy_core_realise(self):
+        coord = AuxCoord(self.pts_real, bounds=self.bds_lazy)
+        coord.bounds
+        result = coord.has_lazy_bounds()
+        self.assertFalse(result)
+
+
+@skip_lazy_auxcoords
+class Test_bounds_dtype(tests.IrisTest):
```
dkillick: You could do with a similar class to check `coord.dtype` too, it would seem.

lib/iris/tests/unit/coords/test_AuxCoord.py
```diff
@@ -0,0 +1,764 @@
...
+    def test_i16(self):
+        test_dtype = np.int16
+        coord = AuxCoord([1], bounds=np.array([[0, 4]], dtype=test_dtype))
+        result = coord.bounds_dtype
+        self.assertTrue(result == test_dtype)
```
dkillick: This looks ripe for being re-cast to `self.assertEqual(result, test_dtype)`... is there a reason you chose not to?

lib/iris/tests/unit/coords/test_AuxCoord.py
```diff
@@ -0,0 +1,764 @@
...
+
+    def test_u16(self):
+        test_dtype = np.uint16
+        coord = AuxCoord([1], bounds=np.array([[0, 4]], dtype=test_dtype))
+        result = coord.bounds_dtype
+        self.assertTrue(result == test_dtype)
```
dkillick: Again.

lib/iris/tests/unit/coords/test_AuxCoord.py
```diff
@@ -0,0 +1,764 @@
...
+
+    def test_f16(self):
+        test_dtype = np.float16
+        coord = AuxCoord([1], bounds=np.array([[0, 4]], dtype=test_dtype))
+        result = coord.bounds_dtype
+        self.assertTrue(result == test_dtype)
```
dkillick: And once more.

lib/iris/tests/unit/coords/test_AuxCoord.py
```diff
@@ -0,0 +1,764 @@
...
+
+def coords_all_dtypes_and_lazynesses(self, dtypes=(np.float64, np.int16)):
+    # Generate coords with all possible types of points and bounds, and all
+    # of the given dtypes.
+    points_types = ['real']
+    bounds_types = ['no', 'real']
+    if _AUXCOORD_LAZY_API:
+        points_types.append('lazy')
+        bounds_types.append('lazy')
+    for dtype in (np.float64, np.int16):
```
dkillick: It feels like you could compress all these `for` loops neatly using functionality provided by `itertools`. I'm sure this has been done in tests recently, but I couldn't find an example quickly...

lib/iris/tests/unit/coords/test_AuxCoord.py
```diff
@@ -0,0 +1,764 @@
...
+                if bounds_type_name == 'lazy':
+                    bds = da.from_array(bds, bds.shape)
+                elif bounds_type_name == 'no':
+                    bds = None
+                coord = AuxCoord(pts, bounds=bds)
+                result = (coord, points_type_name, bounds_type_name)
+                if _DEBUG_ALL_TYPES:
+                    msg = '\nTest result type {}({} {})'
+                    print(msg.format(coord.dtype, result[1], result[2]))
+                yield result
```
dkillick: Although I guess if you went the `itertools` route then this bit would not work so nicely...

lib/iris/coords.py
pp-mo: I do intend to have a go at reorganising the code a bit, because I think there is a lot of common code in the points+bounds getters+setters for DimCoord+AuxCoord ( :sweat_smile: ).
I'm definitely leaving that to a separate PR -- it should be a strict refactor, and it might not work out as easily as I think -- in which case we can quite happily run with what we have.

lib/iris/coords.py
pp-mo: At present, lazy bounds aren't achievable, so I can't get the intended tests working without changing ***something***.

lib/iris/tests/unit/coords/test_AuxCoord.py
pp-mo: No, just a slip..

lib/iris/tests/unit/coords/test_AuxCoord.py
pp-mo: I think I recently suggested that @lbdreyer used `itertools.permutations`, but that is a more complicated case.  My inclination is that up to 3 for loops is just easier to read, as long the indentation doesn't get prohibitive...

lib/iris/tests/unit/coords/test_AuxCoord.py
```diff
@@ -0,0 +1,764 @@
...
+                self.assertEqual(
+                    sub_bounds.dtype, coord_dtype,
+                    msg.format(coord_dtype,
+                               points_type_name, bounds_type_name,
+                               'bounds', sub_points.dtype))
+
+    def test_lazyness(self):
+        # Index coords with all combinations of real+lazy points+bounds, and
+        # either an int or floating dtype.
+        # Check that laziness is preserved in all cases.
```
pp-mo: > default dtypes from the function signature

This was just a mistake -- now fixed !

pp-mo: > default dtypes from the function signature

This was just a mistake -- now fixed !

pp-mo: "laziness is preserved" is supposed to mean "lazy stays lazy + real stays real"
I will try + clarify...

lib/iris/tests/unit/coords/__init__.py
```diff
@@ -14,7 +14,115 @@
 #
 # You should have received a copy of the GNU Lesser General Public License
 # along with Iris.  If not, see <http://www.gnu.org/licenses/>.
-"""Unit tests for the :mod:`iris.coords` module."""
+"""
+Unit tests for the :mod:`iris.coords` module.
 
+Provides test methods and classes common to test_AuxCoord and test_DimCoord.
```
dkillick: For completeness, you could include sphinx directives around `test_AuxCoord` and `test_DimCoord`.

lib/iris/tests/unit/coords/__init__.py
```diff
@@ -14,7 +14,115 @@
...
+    upper = points + 2.0
+    bounds = np.stack((lower, upper), axis=-1)
+    self.pts_real = points
+    self.pts_lazy = da.from_array(points, points.shape)
+    self.bds_real = bounds
+    self.bds_lazy = da.from_array(bounds, bounds.shape)
+
+
+def is_real_data(array):
+    # A parallel to 'is_lazy_data'.
```
dkillick: That is, "A parallel to `iris._lazy_data.is_lazy_data`."...

lib/iris/tests/unit/coords/__init__.py
```diff
@@ -14,7 +14,115 @@
...
+                    bds = da.from_array(bds, bds.shape)
+                elif bounds_type_name == 'no':
+                    bds = None
+                coord = coord_class(pts, bounds=bds)
+                result = (coord, points_type_name, bounds_type_name)
+                yield result
+
+
+class CoordTestMixin(object):
+    def setupTestArrays(self, shape=(3,)):
```
dkillick: I feel duty-bound to make a fuss that you are using weak camelCase for Python _methods_. Methods should have all_lowercase_underscore_separated style names, innit.

lib/iris/tests/unit/fileformats/grib/grib1_load_rules/test_grib1_convert.py
```diff
@@ -64,7 +64,8 @@ def assert_bounded_message(self, **kwargs):
                       '_forecastTimeUnit': 'hours',
                       'phenomenon_bounds': lambda u: (80, 120),
                       '_phenomenonDateTime': -1,
-                      'table2Version': 9999}
+                      'table2Version': 9999,
+                      '_originatingCentre': 'xxx'}
```
pp-mo: Yes, the handling of a single data point has changed.
The converter makes a scalar coord from the '_originatingCentre' property, which was previously just another Mock.  But now that will fail within '_sanitise_array', so this one must be made explicit.
The existing explicit properties are mostly to get around similar Mock-y problems.
I've explained it a bit more in subsequent commits...

lib/iris/tests/unit/coords/test_AuxCoord.py
```diff
@@ -0,0 +1,764 @@
...
+        self.assertIsInstance(result, np.ndarray)
+        self.assertArrayEqual(result, self.pts_real)
+        self.assertTrue(arrays_share_data(result, self.pts_real),
+                        'Points do not share data with the assigned array.')
+
+    @skip_lazy_auxcoords
+    def test_lazy_points(self):
+        # Getting lazy points realises them.
+        coord = AuxCoord(self.pts_lazy)
+        self.assertTrue(coord.has_lazy_points())
```
pp-mo: See newer code, hopefully clearer..

lib/iris/tests/unit/coords/test_AuxCoord.py
```diff
@@ -0,0 +1,764 @@
...
+        self.assertArrayEqual(result, self.bds_real)
+        self.assertTrue(arrays_share_data(result, self.bds_real),
+                        'Bounds do not share data with the assigned array.')
+
+    @skip_lazy_auxcoords
+    def test_real_points_with_lazy_bounds(self):
+        # Getting real points does not touch lazy bounds.
+        coord = AuxCoord(self.pts_real, bounds=self.bds_lazy)
+        coord.points
+        result = coord.core_bounds()
```
pp-mo: See newer code, likewise..

lib/iris/tests/unit/coords/test_DimCoord.py
```diff
@@ -0,0 +1,620 @@
...
+        self.assertFalse(result)
+
+
+@skip_lazy_auxcoords
+class Test_bounds_dtype(tests.IrisTest):
+    def test_i16(self):
+        test_dtype = np.int16
+        coord = DimCoord([1], bounds=np.array([[0, 4]], dtype=test_dtype))
+        result = coord.bounds_dtype
+        self.assertTrue(result == test_dtype)
```
pp-mo: :ok:

lib/iris/tests/unit/coords/test_DimCoord.py
```diff
@@ -0,0 +1,620 @@
...
+
+
+def lazyness_string(data):
+    return 'lazy' if isinstance(data, da.Array) else 'real'
+
+
+_DEBUG_ALL_TYPES = _DEBUG
+
+
+def coords_all_dtypes_and_lazynesses(self, dtypes=(np.float64, np.int16)):
```
pp-mo: Now done

lib/iris/tests/unit/coords/test_DimCoord.py
```diff
@@ -0,0 +1,620 @@
...
+            self.assertEqual(
+                sub_points.dtype, coord_dtype,
+                msg.format(coord_dtype,
+                           points_type_name, bounds_type_name,
+                           'points', sub_points.dtype))
+
+            if bounds_type_name is not 'no':
+                sub_bounds = _core_bounds(sub_coord)
+                self.assertEqual(
+                    sub_bounds.dtype, coord_dtype,
```
pp-mo: Yes, good spot, this should be checking against main_coord.bounds_dtype instead.

We do support different types for points + bounds.
I'll have a think about where and how we should test for that ...

lib/iris/tests/unit/coords/__init__.py
pp-mo: Ok, it was intended for compatibility with unittest styling, i.e. it is like `setUp()`, but I think that is no very good reason...

lib/iris/tests/unit/coords/test_AuxCoord.py
pp-mo: I have now ditched the control arg.
It now always checks for each of a number of *pairs* of dtypes, for points, bounds (which can be different).

lib/iris/tests/unit/coords/__init__.py
pp-mo: > no very good reason.

On reflection + looking at the class, I think maybe it ***is*** good to be consistent with unittest.TestCase styles in this case.
@dkillick would you reconsider ?

pp-mo:
My roundup of outstanding comments that may still need addressing...
 * [ ] re: `test_AuxCoord.Test_bounds_dtype`
    * > You could do with a similar class to check coord.dtype too, it would seem
    * https://github.com/SciTools/iris/pull/2547#discussion_r116172021

 * [ ] re: `test_AuxCoord.Test_points__getter.test_lazy_points_with_lazy_bounds`
    * > For a test class dealing with points, you seem to be making a lot of checks on the state of the bounds here...
    * https://github.com/SciTools/iris/pull/2547#discussion_r116195890

 * [ ] re: `test_AuxCoord.Test_points__getter.test_fail_bad_shape`
    * > That is one terse expected error message!
    * https://github.com/SciTools/iris/pull/2547#discussion_r116196275

 * [ ] re: `test_DimCoord.py.Test__init__.test_real_points_with_lazy_bounds`
    * > I think this needs to become a comment line.
    * https://github.com/SciTools/iris/pull/2547#discussion_r116201691

lib/iris/tests/unit/coords/__init__.py
dkillick: > would you reconsider?

Sure. I was only making a fuss, after all! They can stay.

lib/iris/tests/test_pp_cf.py
```diff
@@ -81,7 +81,7 @@ def callback_aaxzc_n10r13xy_b_pp(cube, field, filename):
 
 
 # XXX: Issue with integer.b.pp and invalid bmdi fill_value for int32 dtype
-@tests.skip_biggus
+#@tests.skip_biggus
```
pp-mo: Change got in by mistake (hence test fail). 
Now removed this + pushed modified version...

lib/iris/tests/unit/coords/test_AuxCoord.py
```diff
@@ -0,0 +1,613 @@
...
+# GNU Lesser General Public License for more details.
+#
+# You should have received a copy of the GNU Lesser General Public License
+# along with Iris.  If not, see <http://www.gnu.org/licenses/>.
+"""
+Unit tests for the :class:`iris.coords.AuxCoord` class.
+
+Note: a lot of these methods are actually defined by the :class:`Coord` class,
+but can only be tested on concrete DimCoord/AuxCoord instances.
+In addition, the DimCoord class has little behaviour for some of these, as it
```
dkillick: Sorry! To clarify, I mean I don't think we need the comments on L22-23.

lib/iris/tests/unit/coords/test_AuxCoord.py
```diff
@@ -0,0 +1,613 @@
...
+        coord = AuxCoord(self.pts_real, bounds=self.bds_real)
+        result = coord.lazy_bounds()
+        self.assertEqualLazyArraysAndDtypes(result, self.bds_lazy)
+
+    def test_lazy_core(self):
+        coord = AuxCoord(self.pts_real, bounds=self.bds_lazy)
+        result = coord.lazy_bounds()
+        self.assertIs(result, self.bds_lazy)
+
+    def test_none(self):
```
pp-mo: Oops..

lib/iris/tests/unit/coords/test_AuxCoord.py
```diff
@@ -0,0 +1,613 @@
...
+from __future__ import (absolute_import, division, print_function)
+from six.moves import (filter, input, map, range, zip)  # noqa
+
+# Import iris.tests first so that some things can be initialised before
+# importing anything else.
+import iris.tests as tests
+
+import numpy as np
+
+from iris.tests.unit.coords import (CoordTestMixin,
```
pp-mo: I tried that out, but then I found you can't then run the test as a standalone script any more.
You get a message like ...
```
$ python lib/iris/tests/unit/coords/test_AuxCoord.py Test_points__setter
  File "lib/iris/tests/unit/coords/test_AuxCoord.py", line 34, in <module>
    from . import (CoordTestMixin, lazyness_string,
ValueError: Attempted relative import in non-package

$ 
```

So, probably best left as it is.

lib/iris/tests/unit/coords/test_AuxCoord.py
```diff
@@ -0,0 +1,613 @@
...
+
+            if bounds_type_name is not 'no':
+                sub_bounds_lazy = lazyness_string(sub_coord.core_bounds())
+                self.assertEqual(
+                    sub_bounds_lazy, bounds_type_name,
+                    msg.format(coord_dtype,
+                               points_type_name, bounds_type_name,
+                               'bounds', bounds_type_name, sub_bounds_lazy))
+
+    def test_real_data_copies(self):
```
pp-mo: > don't we want to run a similar test on lazy data?

I don't think that lazy data has any obvious way of checking that a distinct copied array is actually mapped to the same data.
Also, unlike with real data, copying should not create any significant performance problems.  So I think we don't need to check this.

lib/iris/tests/unit/coords/test_AuxCoord.py
dkillick: > probably best left as it is

Agreed!

PR #2550, opened 2017-05-15
Corrected use of deprecated numpy behaviour. Necessary for numpy 1.12

marqh:
proposed replacement for #2468 

targeting dask branch where the GRIB code is now maintained.


pp-mo:
Testing with numpy 1.12, I don't get any deprecation warning for this expression ?

TBH in this low-level code I think it is really ***not*** a question of what default type numpy selects for a zeros array, or for a Python float. either :  Instead, we know ***exactly*** what the dtype should be for data read from the file, and we should be saying so explicitly.
I'd go with `np.dtype(np.float64)`, consistent with usage elsewhere in this module.

bjlittle:
@marqh The `dask` feature branch has since moved on, and so this fix has gone stale and is no longer applicable. Thanks.

PR #2553, opened 2017-05-15
Coord points and bounds inheritance.

pp-mo:
This refactors the inheritance organisation of the `points` and `bounds` property code amongst Coord  / AuxCoord / DimCoord.

At present, `DimCoord` and `AuxCoord` both provide their own *independent* implementations of points+bounds getters+setters, but there is quite a lot of duplication in that code.
This refactor removes the code duplication, by ...
  * most of the relevant code is relocated by "pushing it up" from `AuxCoord` into `Coord`.
  * `AuxCoord` then becomes an almost trivial descendant of `Coord`, 
  * `DimCoord` is now a 'specialised' form that inherits + extends those implementations, adding realisation and DimCoord-specific validity testing.
  * the 'from_coord' routine has also been generalised in a similar way.

I think this structure makes more sense than the existing.
The knock-on effects are extremely minor : just that some error cases produce a slightly different message.
The recently-extended testing in `iris.tests.unit.coords` encourages me that nothing serious has been broken (!)

### DIFF
--- a/lib/iris/coords.py
+++ b/lib/iris/coords.py
@@ -520,13 +520,95 @@ class Coord(six.with_metaclass(ABCMeta, CFVariableMixin)):
 
         return new_coord
 
-    @abstractproperty
-    def points(self):
-        """Property containing the points values as a numpy array"""
+    @classmethod
+    def from_coord(cls, coord):
+        """Create a new Coord of this type, from the given coordinate."""
+        kwargs = {'points': coord.core_points(),
+                  'bounds': coord.core_bounds(),
+                  'standard_name': coord.standard_name,
+                  'long_name': coord.long_name,
+                  'var_name': coord.var_name,
+                  'units': coord.units,
+                  'attributes': coord.attributes,
+                  'coord_system': copy.deepcopy(coord.coord_system)}
+        if issubclass(cls, DimCoord):
+            # DimCoord introduces an extra constructor keyword.
+            kwargs['circular'] = getattr(coord, 'circular', False)
+        return cls(**kwargs)
 
-    @abstractproperty
-    def bounds(self):
-        """Property containing the bound values as a numpy array"""
+    def _sanitise_array(self, src, ndmin):
+        if is_lazy_data(src):
+            # Lazy data : just ensure ndmin requirement.
+            ndims_missing = ndmin - src.ndim
+            if ndims_missing <= 0:
+                result = src
+            else:
+                extended_shape = tuple([1] * ndims_missing + list(src.shape))
+                result = src.reshape(extended_shape)
+        else:
+            # Real data : a few more things to do in this case.
+            # Ensure the array is writeable.
+            # NB. Returns the *same object* if src is already writeable.
+            result = np.require(src, requirements='W')
+            # Ensure the array has enough dimensions.
+            # NB. Returns the *same object* if result.ndim >= ndmin
+            result = np.array(result, ndmin=ndmin, copy=False)
+            # We don't need to copy the data, but we do need to have our
+            # own view so we can control the shape, etc.
+            result = result.view()
+        return result
+
+    def _points_getter(self):
+        """The coordinate points values as a NumPy array."""
+        return self._points_dm.data
+
+    def _points_setter(self, points):
+        # Set the points to a new array - as long as it's the same shape.
+
+        # Ensure points has an ndmin of 1 and is either a numpy or lazy array.
+        # This will avoid Scalar coords with points of shape () rather
+        # than the desired (1,).
+        points = self._sanitise_array(points, 1)
+
+        # Set or update DataManager.
+        if self._points_dm is None:
+            self._points_dm = DataManager(points)
+        else:
+            self._points_dm.data = points
+
+    points = property(_points_getter, _points_setter)
+
+    def _bounds_getter(self):
+        """
+        The coordinate bounds values, as a NumPy array,
+        or None if no bound values are defined.
+
+        .. note:: The shape of the bound array should be: ``points.shape +
+            (n_bounds, )``.
+
+        """
+        bounds = None
+        if self.has_bounds():
+            bounds = self._bounds_dm.data
+        return bounds
+
+    def _bounds_setter(self, bounds):
+        # Ensure the bounds are a compatible shape.
+        if bounds is None:
+            self._bounds_dm = None
+        else:
+            bounds = self._sanitise_array(bounds, 2)
+            if self.shape != bounds.shape[:-1]:
+                raise ValueError("Bounds shape must be compatible with points "
+                                 "shape.")
+            if not self.has_bounds() \
+                    or self.core_bounds().shape != bounds.shape:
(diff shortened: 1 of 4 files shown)


lib/iris/coords.py
```diff
@@ -1684,92 +1745,6 @@ def xml_element(self, doc):
...
-            if not self.has_bounds() \
-                    or self.core_bounds().shape != bounds.shape:
-                # Construct a new bounds DataManager.
-                self._bounds_dm = DataManager(bounds)
-            else:
-                self._bounds_dm.data = bounds
-
     # This is necessary for merging, but probably shouldn't be used otherwise.
     # See #962 and #1772.
     def __hash__(self):
```
dkillick: I know it's out of the current extent of this PR's changes, but could you (/did you) look into whether we still really need this? If we don't we could practically get rid of `AuxCoord` entirely.

lib/iris/coords.py
```diff
@@ -1580,29 +1652,28 @@ def _new_points_requirements(self, points):
...
-        if self._points_dm is None:
-            self._points_dm = DataManager(points)
-        else:
-            self._points_dm.data = points
+        # Invoke the generic points setter.
+        super(DimCoord, self)._points_setter(points)
+
+        if self._points_dm is not None:
+            # Fetch the new core array, as the parent can change it.
+            points = self._points_dm._real_array
```
dkillick: I'm not at all a fan of accessing this private `DataManager` method. Surely there is a better way of doing this?

lib/iris/coords.py
```diff
@@ -1639,37 +1710,27 @@ def _new_bounds_requirements(self, bounds):
...
             bounds = as_concrete_data(bounds)
             bounds = np.array(bounds, copy=copy, ndmin=2)
-            # Checks for appropriate values for the new bounds.
+
+        # Invoke the generic bounds setter.
+        super(DimCoord, self)._bounds_setter(bounds)
+
+        if self._bounds_dm is not None:
+            # Fetch the new core array, as the parent can change it.
+            bounds = self._bounds_dm._real_array
```
dkillick: Again, accessing this private method is a bad idea.

lib/iris/coords.py
```diff
@@ -1639,37 +1710,27 @@ def _new_bounds_requirements(self, bounds):
...
-    def bounds(self, bounds):
-        if bounds is None:
-            self._bounds_dm = None
-        else:
+    def _bounds_setter(self, bounds):
+        if bounds is not None:
             # Ensure we have a realised array of new bounds values.
             copy = is_lazy_data(bounds)
             bounds = as_concrete_data(bounds)
             bounds = np.array(bounds, copy=copy, ndmin=2)
```
dkillick: You no longer need to set `ndmin` here (it'll be done by `_sanitise_points` in the parent).

lib/iris/coords.py
pp-mo: I looked into this.
The same code is present in `Cube.__hash__` (and a couple of others).

***In Python3***, the hash operation is actually disabled for user classes that define `__eq__`, which means you must provide `__hash__` to allow, for example, collecting the objects into sets, or using them as dictionary keys.
So it seems we ***do*** need this as otherwise we can't collect Coords into sets etc.

**The problem:**
This violates a general principle that
"objects which compare equal have the same hash value".
See: https://docs.python.org/2.7/reference/datamodel.html#object.__hash__

So, we should really fix these, 
but it probably means changing other areas such as merge, and possibly concatenate.

pp-mo: P.S. I see that the exact same code appears in DimCoord, so it's another thing that can be moved into Coord.  I've now done that.

lib/iris/coords.py
```diff
@@ -684,6 +766,13 @@ def _as_defn(self):
                          self.units, self.attributes, self.coord_system)
         return defn
 
+    # Must supply __hash__ as Python 3 does not enable it if __eq__ is defined.
+    # NOTE: Violates "objects which compare equal must have the same hash".
+    # Currently needed, but not really correct and should be fixed.
```
dkillick: I'd add "for merging" to this comment because otherwise it might not be clear _why_ it is needed (that is, what's using it).

lib/iris/coords.py
```diff
@@ -1580,29 +1654,29 @@ def _new_points_requirements(self, points):
...
 
-        # Checks for 1d, numeric, monotonic
-        self._new_points_requirements(self._points_dm.data)
+            # Check validity requirements for dimension-coordinate points.
+            self._new_points_requirements(points)
 
-        # Make the array read-only.
-        self._points_dm.data.flags.writeable = False
+            # Make the array read-only.
+            points.flags.writeable = False
```
dkillick: This will update the underlying array (i.e. `self._points_dm.core_data()`) as well, right?

lib/iris/cube.py
```diff
@@ -3081,8 +3081,9 @@ def __ne__(self, other):
             result = not result
         return result
 
-    # Must supply __hash__, Python 3 does not enable it if __eq__ is defined
-    # This is necessary for merging, but probably shouldn't be used otherwise.
+    # Must supply __hash__ as Python 3 does not enable it if __eq__ is defined.
+    # NOTE: Violates "objects which compare equal must have the same hash".
+    # Currently needed, but not really correct and should be fixed.
```
dkillick: Again, you could specify where it is needed.

lib/iris/coords.py
pp-mo: But in fact I tested what happens when you error its usage, and the resulting errors are ***not*** about merge after all.
See : https://github.com/SciTools/iris/pull/1772#issuecomment-302080133

? I could say e.g. "needed so we can use Coords as set members and dictionary keys" ?

lib/iris/coords.py
pp-mo: Yes, but we made sure we copied it.
As [here]()

pp-mo: Yes, but we made sure we are dealing with our own view, as stated just above (L1661).

lib/iris/coords.py
pp-mo: New commit coming that I hope makes this clearer...

lib/iris/coords.py
```diff
@@ -1536,10 +1617,13 @@ def __eq__(self, other):
...
-    # See #962 and #1772.
-    def __hash__(self):
-        return hash(id(self))
+    # For Python 3, we must explicitly re-implement the '__hash__' method, as
+    # defining an '__eq__' has blocked its inheritance.  See ...
+    # https://docs.python.org/3.1/reference/datamodel.html#object.__hash__
+    # "If a class that overrides __eq__() needs to retain the
+    # implementation of __hash__() from a parent class, the interpreter
+    # must be told this explicitly".
+    __hash__ = Coord.__hash__
```
pp-mo: I could use "super" here, if you think it would be preferable (it avoids saying who your parent is).
This way is more efficient though :wink:

PR #2554, opened 2017-05-15
Fix masked time bounds

corinnebosley:
In the problem that this PR is attempting to solve, a NetCDF file had a time coordinate whose points were unmasked, but its bounds were masked.  The coordinate could not be constructed because the masked bounds were being filled with the fill_value (9x10^36) which could not be interpreted as a date. 

This potential fix checks that if the bounds array for a coordinate is masked, the points array is also masked, otherwise the bounds array is filled with zeros instead of the cube fill value.  I consider this acceptable because I can't think of any reasonable situation in which a bounds array would be masked when the points array isn't.

This does not assign reasonable values to the bounds array, it just makes it able to be constructed in a case like the one in question so that the silly coordinate can at least be inspected and adjusted.

I am still not sure whether this is the best way to fix this, and would appreciate some opinions please.

### DIFF
--- a/lib/iris/fileformats/_pyke_rules/fc_rules_cf.krb
+++ b/lib/iris/fileformats/_pyke_rules/fc_rules_cf.krb
@@ -1,4 +1,4 @@
-# (C) British Crown Copyright 2010 - 2016, Met Office
+# (C) British Crown Copyright 2010 - 2017, Met Office
 #
 # This file is part of Iris.
 #
--- a/lib/iris/fileformats/_pyke_rules/fc_rules_cf.krb
+++ b/lib/iris/fileformats/_pyke_rules/fc_rules_cf.krb
@@ -1543,7 +1543,10 @@ fc_extras
             bounds_data = cf_bounds_var[:]
             # Gracefully fill bounds masked array.
             if ma.isMaskedArray(bounds_data):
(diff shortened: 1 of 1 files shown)


lib/iris/fileformats/_pyke_rules/fc_rules_cf.krb
```diff
@@ -1543,7 +1543,10 @@ fc_extras
             bounds_data = cf_bounds_var[:]
             # Gracefully fill bounds masked array.
             if ma.isMaskedArray(bounds_data):
-                bounds_data = ma.filled(bounds_data)
+                if not ma.isMaskedArray(points_data):
```
djkirkham: By this point in the function `points_data` has been filled, so `ma.isMaskedArray(points_data)` will always return `False`. You need to record the value of `ma.isMaskedArray(points_data)` before the array is filled.

djkirkham:
It seems like you may be assuming that the points remain masked, but this is not the case (see line 1536 in the file you've modified). 

The problem then, is that any value you choose to fill with could represent a valid value, so by the time the cube is loaded there's no way to know which values are valid.

djkirkham:
Actually, now that I think about it, it seems highly unlikely that any partially masked coordinate would be able to load successfully, due to the monotonicity constraints.

PR #2555, opened 2017-05-15
Fix pickle test

dkillick:
Fix a cube pickling roundtrip test that was not preserving lazy data and causing a spurious test failure. Bonus: test module now pep8 compliant!

Fixes part of #2398

### DIFF
--- a/lib/iris/tests/test_pickling.py
+++ b/lib/iris/tests/test_pickling.py
@@ -22,7 +22,8 @@ Test pickling of Iris objects.
 from __future__ import (absolute_import, division, print_function)
 from six.moves import (filter, input, map, range, zip)  # noqa
 
-# import iris tests first so that some things can be initialised before importing anything else
+# Import iris tests first so that some things can be initialised
+# before importing anything else.
 import iris.tests as tests
 
 import six.moves.cPickle as pickle
@@ -36,7 +37,10 @@ import iris
 
 class TestPickle(tests.IrisTest):
     def pickle_then_unpickle(self, obj):
-        """Returns a generator of ("cpickle protocol number", object) tuples."""
+        """
+        Returns a generator of ("cpickle protocol number", object) tuples.
+
+        """
         for protocol in range(1 + pickle.HIGHEST_PROTOCOL):
             bio = io.BytesIO()
             pickle.dump(obj, bio, protocol)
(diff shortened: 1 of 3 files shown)


lib/iris/tests/test_pickling.py
```diff
@@ -48,19 +52,22 @@ def pickle_then_unpickle(self, obj):
...
+                       checksum=False)
 
-        for _, recon_cube in self.pickle_then_unpickle(cube):
+        for p, recon_cube in self.pickle_then_unpickle(cube):
             self.assertTrue(recon_cube.has_lazy_data())
-            self.assertCML(recon_cube, ('cube_io', 'pickling', 'theta.cml'), checksum=False)
-            self.assertCubeData(cube, recon_cube)
+            self.assertCML(recon_cube, ('cube_io', 'pickling', 'theta.cml'),
+                           checksum=False)
+            self.assertCubeData(cube.copy(), recon_cube)
```
djkirkham: Why is this copied now?

dkillick: Think about it: `assertCubeData` _realises the cube's data_. This was causing the following lazy data assertion to always, incorrectly, fail.

djkirkham: I see. I was puzzled as to how this test ever passed. It turns out `assertCubeData` used to be defined in a way which didn't realise the data. I suppose it was changed in a previous PR.

dkillick: >  turns out `assertCubeData` used to be defined in a way which didn't realise the cube's data

Oh excellent. Do you think it would be more appropriate to revert `assertCubeData` to how it was before rather than copying the cube here?

djkirkham: Yeah, maybe. As it is, you're relying on `copy` working correctly. It might be confusing if this test ever failed because `copy` broke.

djkirkham: The way the method used to be defined was to use `cube.lazy_data().ndarray()`. I supposed we'd want to change it to:
```
as_concrete_data(cube.lazy_data(),
                 nans_replacement=cube.fill_value,
                 result_dtype=cube.dtype)
```
Also, do that for both arguments... urgh.

dkillick: It does make it a little like an integration test, I guess. However, it's not like we don't use `copy` elsewhere in our tests... and I'm not sure I can see a good way of checking the original cube and reconstructed cube data match without realising them.

dkillick: > do that for both arguments

We do have a helpful test method that we can mess around with that won't affect anyone else!

dkillick:
Review actions: no longer copying the original cube to maintain lazy data.

PR #2557, opened 2017-05-16
Fix rules test for _make_cube

### LINKED ISSUE #2398: Get all Iris tests working again (5 man days' effort)

dkillick:
Remove another biggus skipper, this time on a test that was causing a segfault. It appears the segfault was being caused when creating the cube with `core_data` described by an empty mock object.

I've also removed some needless mockism from the wider test that was being used to capture a raised warning. I've replaced the mockism with the warning capture and check code used widely elsewhere within Iris tests.

Part-fixes #2398.

### DIFF
--- a/lib/iris/tests/unit/fileformats/rules/test__make_cube.py
+++ b/lib/iris/tests/unit/fileformats/rules/test__make_cube.py
@@ -18,18 +18,20 @@
 
 from __future__ import (absolute_import, division, print_function)
 from six.moves import (filter, input, map, range, zip)  # noqa
+import six
 
 # Import iris.tests first so that some things can be initialised before
 # importing anything else.
 import iris.tests as tests
 
-from iris.fileformats.rules import _make_cube
-import iris.fileformats.rules
+import warnings
+
+from iris.fileformats.rules import ConversionMetadata, _make_cube
 from iris.tests import mock
+import numpy as np
 
 
 class Test(tests.IrisTest):
-    @tests.skip_biggus
     def test_invalid_units(self):
         # Mock converter() function that returns an invalid
         # units string amongst the collection of other elements.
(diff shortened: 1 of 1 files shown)


lib/iris/tests/unit/fileformats/rules/test__make_cube.py
```diff
@@ -18,18 +18,20 @@
...
 
 # Import iris.tests first so that some things can be initialised before
 # importing anything else.
 import iris.tests as tests
 
-from iris.fileformats.rules import _make_cube
-import iris.fileformats.rules
+import warnings
+
+from iris.fileformats.rules import ConversionMetadata, _make_cube
```
dkillick: It just seems a bit daft to duplicate an import line like that, especially when there doesn't seem that much value gained from the style...

lib/iris/tests/unit/fileformats/rules/test__make_cube.py
```diff
@@ -57,9 +61,9 @@ def test_invalid_units(self):
         self.assertEqual(cube.attributes, expected_attributes)
 
         # Check warning was raised.
-        self.assertEqual(warn.call_count, 1)
-        warning_msg = warn.call_args[0][0]
-        self.assertIn('invalid units', warning_msg)
+        self.assertEqual(len(warn), 1)
+        exp_emsg = 'invalid units {!r}'.format(units)
+        six.assertRegex(self, str(warn[0].message), exp_emsg)
```
dkillick: We could, but there are a [couple of](https://github.com/SciTools/iris/blob/dask/lib/iris/tests/unit/fileformats/grib/save_rules/test_set_time_range.py#L100) [other examples](https://github.com/SciTools/iris/blob/dask/lib/iris/tests/unit/fileformats/pp/test__field_gen.py#L76) of this test method being used when testing warning text like this (including the one I duplicated here). Admittedly there are actually more examples of `assertRegexpMatches` being used in such situations, so I'll change this.

dkillick:
@pp-mo I've addressed your suggestions above.

PR #2560, opened 2017-05-17
Purge spurious imports

dkillick:
Purges unused dask imports that were left behind after we improved the API for lazy data handling, plus any other unused imports in the affected modules.

I've also added a new function, `iris._lazy_data.lazy_concat`, which performs dask concatenation of lazy arrays. This means we can remove dask imports from `cube.py` and `_concatenate.py`.

### DIFF
--- a/lib/iris/_lazy_data.py
+++ b/lib/iris/_lazy_data.py
@@ -201,6 +201,25 @@ def multidim_lazy_stack(stack):
     return result
 
 
+def lazy_concat(data, axis):
+    """
+    Concatenate dask arrays along a given axis.
+
+    Args:
+
+    * data:
+        An iterable of dask arrays.
+
+    * axis:
+        The data dimension along which to perform the concatenation.
+
+    Returns:
+        The input data concatenated along the specified axis.s
+
+    """
+    return da.concatenate(data, axis)
+
+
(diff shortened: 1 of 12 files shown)


lib/iris/_lazy_data.py
```diff
@@ -201,6 +201,25 @@ def multidim_lazy_stack(stack):
     return result
 
 
+def lazy_concat(data, axis):
```
pp-mo: Not terribly convinced about the usefulness of this ...
I'm opposed to duplicating APIs unless you are really defining a complete abstraction layer with its own value.
-- for example, what if da.concatenate later acquires useful keywords, which then have to be added here...

In this case, I don't think abstracting *all* usage dask is obvioiusly helpful in itself, so I don't see a great need for this specific function.
So I think I would ditch this, just to make the PR tighter.

lib/iris/cube.py
```diff
@@ -2396,13 +2396,13 @@ def make_chunk(key):
         if len(chunks) == 1:
             result = chunks[0]
         else:
+            chunk_data = [chunk.core_data() for chunk in chunks]
```
pp-mo: :+1: for using core_data() here !

PR #2558, opened 2017-05-16
draft whats new

marqh:
draft what's new page

clean up of contributions files

PR #2561, opened 2017-05-17
Updated v1.13 whatsnew

lbdreyer:
As @marqh is unavailable I have grabbed his commit from #2558 and made some edits.

I will close #2558.


docs/iris/src/whatsnew/1.13.rst
```diff
@@ -1,34 +1,34 @@
...
+* An experimental link to python-stratify via :mod:`iris.experimental.stratify`.
 * Data arrays may be shared between cubes, and subsets of cubes, by using the :meth:`iris.cube.share_data` flag.
 
 
 Bug Fixes
 =========
 
 * The bounds are now set correctly on the longitude coordinate if a zonal mean diagnostic has been loaded from a PP file as per the CF Standard.
-* netCDF loading will now determine whether there is a string-valued scalar label, i.e.: a character variable that only has one dimension (the length of the string) and interpret this effectively.
+* NetCDF loading will now determine whether there is a string-valued scalar label, i.e. a character variable that only has one dimension (the length of the string), and interpret this correctly.
```
lbdreyer: 'effectively' doesn't feel like the right word to me, but that may just be my interpretation of the word.

docs/iris/src/whatsnew/1.13.rst
```diff
@@ -0,0 +1,37 @@
...
+* Data arrays may be shared between cubes, and subsets of cubes, by using the :meth:`iris.cube.share_data` flag.
+
+
+Bug Fixes
+=========
+
+* The bounds are now set correctly on the longitude coordinate if a zonal mean diagnostic has been loaded from a PP file as per the CF Standard.
+* NetCDF loading will now determine whether there is a string-valued scalar label, i.e. a character variable that only has one dimension (the length of the string), and interpret this correctly.
+* A line plot of geographic coordinates (e.g. drawing a trajectory) wraps around the edge of the map cleanly, rather than plotting a segment straight across the map.
+* Preserve lazy data in PP save-pairs generation, so you can work with a list of save cubes and/or fields without excessive memory requirements (as intended for saving).
```
corinnebosley: I find the sentence 'work with a list of save cubes' a bit confusing.  Does it mean 'save a list of cubes' or 'work with a list of saved cubes' or something else entirely?

docs/iris/src/whatsnew/1.13.rst
```diff
@@ -0,0 +1,37 @@
...
+* When two cubes have coordinates which do not match, an error should be raised if a user tries to perform arithmetic operations on them. These cases are now caught by the multiply and divide operators as well as the add and subtract operators, which already worked correctly.
```
corinnebosley: Is 'worked correctly' the right phrase? I would prefer something like 'has this behaviour', but it boils down to a matter of opinion.  Don't mind if you keep it as is.

docs/iris/src/whatsnew/1.13.rst
```diff
@@ -0,0 +1,37 @@
...
+* Limited area Rotated Pole datasets where the data range is ``0 <= lambda < 360``, for example as produced in New Zealand, are plotted over a sensible map extent by default.
+* Removed the potential for a RuntimeWarning: overflow encountered in ``int_scalars`` which was missed during collapsed calculations. This could trip up unwary users of limited data types, such as int32 for very large numbers (e.g. seconds since 1970).
+* The CF conventions state that certain ``formula_terms`` terms may be omitted and assumed to be zero (http://cfconventions.org/cf-conventions/v1.6.0/cf-conventions.html#dimensionless-v-coord) so Iris now allows factories to be constructed with missing terms.
+* In the User Guide's contour plot example, clabel inline is set to be False, so that it renders correctly, avoiding spurious horizontal lines across plots, although this does make labels a little harder to see.
```
corinnebosley: I would remove the comma in 'set to be False so that it renders correctly' but this is also minor and a matter of personal opinion.  Your call.

docs/iris/src/whatsnew/1.13.rst
lbdreyer: It looks like this came refers to #2452

docs/iris/src/whatsnew/1.13.rst
lbdreyer: I will change it. I can be a little comma-happy so I do usually need to be reigned in!

docs/iris/src/whatsnew/1.13.rst
corinnebosley: That ticket doesn't really clarify, but I would guess that it means 'save a list of cubes'...?

lbdreyer: I would agree that it means 'save a list of cubes'.

Although I don't think the 'PP save-pairs' makes much sense to users!

corinnebosley: On second thought I'm still not sure if that's right.  It feels like it means 'work with a list of cubes which will be saved at some point down the line' but obviously this sentence needs to be succinct.  Maybe it's best just left as it is?

docs/iris/src/whatsnew/1.13.rst
corinnebosley: 'An error is now raised if a user tries to perform an arithmetic operation on two cubes with mismatching coordinates'

corinnebosley: Or maybe 'An error should be raised if a user......'

The fact that it is now caught by divide and multiply is explained later.

docs/iris/src/whatsnew/1.13.rst
lbdreyer: 'When saving to PP, preserve lazy data when generating PP fields from cubes so that a list of cubes can be saved to PP without excessive memory requirements.'

What do you think about removing '(as intended for saving)'?

corinnebosley: Yep.  Remove it.

corinnebosley: And that sentence is better.

docs/iris/src/whatsnew/1.13.rst
lbdreyer: 'An error is now correctly raised if a user tries to perform an arithmetic operation on two cubes with mismatching coordinates. Previously these cases were caught by the add and subtract operators, and now it is also caught by the multiply and divide operators.'
?

lbdreyer:
Thanks @corinnebosley !

PR #2563, opened 2017-05-17
Mergeback 1.13.x branch

lbdreyer:
This merge back is incredibly light (since we only cut the `v1.13.x` branch so late) but I still think it's worth doing

I have set the new version to be `2.0.0-DEV`. Is this correct?


corinnebosley:
@lbdreyer Tests timed out.  Need a kick.

Issue #2507, opened 2017-04-24
PP load/save BMDI (5 man days' effort)


bjlittle:
Ensure that PP `BMDI` and `fill_value` is being handled correctly for PP load/save.

pp-mo:
Not necessarily ***all*** about BMDI, but setting out to fix outstanding load/save and test problems.

pp-mo:
Taking these skipped tests on from #2382 :
from [this comment](https://github.com/SciTools/iris/issues/2382#issuecomment-280377648)

`lib/iris/tests/test_cdm.py TestMaskedData.test_save_and_merge`
`lib/iris/tests/test_pp_module.py TestPackedPP.test_wgdos`
 ==> those both now fixed, according to #2398

`lib/iris/tests/test_pp_module.py TestPackedPP.test_rle`
  ==> **outstanding**

pp-mo:
Tests highlighted by, + taken out of #2398
`lib/iris/tests/test_cube_to_pp.py:236`
`lib/iris/tests/test_cube_to_pp.py:270`
`lib/iris/tests/test_pp_cf.py:84`
 ==> "pp/bmdi problem"

`lib/iris/tests/test_pp_module.py:245 test_rle`
 ==> "Test loads in pp file and checks the data/mask which is different to the expected"

`lib/iris/tests/integration/test_pp.py:51`
`lib/iris/tests/integration/test_pp.py:86`
`lib/iris/tests/integration/test_pp.py:122`
 ==> Seg fault

pp-mo:
Here's the PR : #2507

dkillick:
Closed now that #2564 has been merged.

PR #2564, opened 2017-05-17
Fix PP bmdi handling

pp-mo:
Unskip tests and fix problems relating to BMDI and fill value in PP save/load.
Addresses #2507.

This code is no longer allowing *any* usage of BMDI with integer fields.
There seems to be no official interpretation of what a missing data value might be in that case, and we don't seem to have any test data that requires it.
 * On load, we create cubes with no fill value.
 * On save, masked integer data will cause an error.
These changes are reflected in the results files.

The remaining test changes are almost all to do with providing mock test fields with a ".core_data()" that returns an array, or a mock array.


lib/iris/fileformats/pp.py
```diff
@@ -1398,8 +1396,14 @@ def save(self, file_handle):
         # Before we can actually write to file, we need to calculate the header
         # elements. First things first, make sure the data is big-endian
```
dkillick: It is outside of your changes admittedly, but it seems that this comment (specifically about the endianness) may now be in the wrong place given the code that immediately follows it.

lib/iris/fileformats/rules.py
```diff
@@ -897,13 +897,19 @@ def _make_cube(field, converter):
     # Convert the field to a Cube.
     metadata = converter(field)
 
-    cube = iris.cube.Cube(field.core_data(),
+    cube_data = field.core_data()
+    cube_dtype = cube_data.dtype
+    if cube_dtype.kind in 'bui':
```
dkillick: Right, everywhere else this is expressed as `'biu'`... could you perhaps maintain that consistency?

lib/iris/tests/test_cube_to_pp.py
```diff
@@ -230,10 +230,6 @@ def geog_cs(self):
 
 
 class TestPPSaveRules(tests.IrisTest, pp.PPTest):
-    # Skip this test, there appears to be a long standing bug in PP saving
-    # for int32, which is made worse by assigning the 'default' bmdi of
-    # 1e30 into int arrays
-    @tests.skip_biggus
```
dkillick: Nice to see these all being fixed! :100:

lib/iris/tests/unit/fileformats/test_rules.py
```diff
@@ -107,10 +109,16 @@ def test_simple_factory(self):
...
-        field.core_data = mock.Mock()
+
+        # Construct a mock content array.
+        core_data_array = mock.Mock()
         # Add a compute attribute so that the data will be treated as lazy.
-        field.core_data.compute = None
-        field.core_data.dtype = None
+        core_data_array.compute = None
+        # It must also have a recognisable dtype.
+        core_data_array.dtype = np.dtype('f4')
```
dkillick: You _could_ set these attributes inside the mock call... but you would lose the chance to comment on each attribute, I guess.

dkillick:
> On save, masked integer data will cause an error.

Could you provide the rationale for this decision?

dkillick:
@pp-mo you also have a couple of [failing tests on Travis](https://travis-ci.org/SciTools/iris/jobs/233330611#L5466-L5494). Looks like they're both to do with resultant dtype.

PR #2565, opened 2017-05-18
Use six.assertRegex to avoid Python 3 test method deprecation.

pp-mo:
In contrast to what I said [here](https://github.com/SciTools/iris/pull/2557#discussion_r117057815), there *is* a point to using six.assertRegex,
as otherwise assertRegexpMatches in Python 3 gives "DeprecationWarning: Please use assertRegex instead."


dkillick:
> In contrast to what I said [here](https://github.com/SciTools/iris/pull/2557#discussion_r117057815)

Ironically, you actually haven't reverted this change you requested in this PR :wink:

PR #2567, opened 2017-05-18
Get cell method interval from t2 - t1 no lbtim.ia when loading pp

lbdreyer:
Before now, when loading a pp file/fieldsfile, we have used the lbtim.ia value to set the interval of the cell method. This is incorrect.

 :open_mouth: 

From the f03 documentation:
> for time aggregated fields IA is the time interval in hours between the individual fields from which the mean was computed

We had thought "time interval" referred to the period over which time-processing is being done but that is not the case! "time interval" refers to the sampling frequency aka how frequently the data is being sampled to generate the mean/accumulation/etc.

### Example
Say the UM is outputting a pressure field every hour. Some of these fields are aggregated so that the maximum over 12 hours is calculated and the resultant field is output as part of a fieldsfile/pp file. In this resultant file the field would have the values:
`lbtim.ia = 1` <- because the maximum was calculated from hourly data
`t2-t1 = 12 hours` <- because the maximum was taken over a 12 hour time interval
Previously this example file would have been loaded in as 
```
     Cell methods:
          mean: time (1 hour)
```
But it should be:
```
     Cell methods:
          mean: time (12 hour)
```

### This PR
The cell_method time interval should be calculated from the difference between T1 and T2 so this PR adds that.

N.B.  In this PR I have removed all saving to lbtim.ia as we don't encode that information in the cube so we can't save it out correctly.

### DIFF
--- a/lib/iris/etc/pp_save_rules.txt
+++ b/lib/iris/etc/pp_save_rules.txt
@@ -79,7 +79,6 @@ IF
     scalar_coord(cm, 'forecast_period') is None
     scalar_coord(cm, 'forecast_reference_time') is None
 THEN
-    pp.lbtim.ia = 0
     pp.lbtim.ib = 0
     pp.t1 = scalar_coord(cm, 'time').units.num2date(scalar_coord(cm, 'time').points[0])
     pp.t2 = netcdftime.datetime(0, 0, 0)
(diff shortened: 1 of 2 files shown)


corinnebosley:
@lbdreyer I thought there was something else that was calculated from t2-t1.  We had support tickets about it because it was being calculated backwards (t1-t2).  Can't remember what that was...

lbdreyer:
The f03 documentation is being updated to say:
> for time aggregated fields IA is the time interval in hours between the individual fields from which the mean was computed. This has a minimum value of 1 when time aggregation is switched on, so sampling frequencies of less than one hour appear to be one hour. Note: This is **not** the period that the mean/accumulation is calculated over (which is defined by T1 and T2 as described above), rather it is the frequency with which the data in that period is sampled.

lbdreyer:
From discussions with @pp-mo it sounds like I may have misunderstood cf.

I am going to clarify this with the users who raised the issue but it look like I will close this.

Issue #2460, opened 2017-03-27
Ensure docs still build (2 man days)


dkillick:
Ensure the docs still build, fixing any problems

Doc build checks:
 * "main" documentation
 *  gallery
 * doctests

dkillick:
Building docs for iris/master (@ 538eed4): `build succeeded, 100 warnings.`

Building docs for iris/dask (@ 88db1d2): `build succeeded, 100 warnings.`

The good news is the docs still build and the number of docs-build warnings has not changed. I haven't checked yet whether the warnings being produced are the same or not. I'll do that soon...

Of course, almost all of the documentation that we've added is in private modules (I'm looking at you, `_lazy_data`) so this result shouldn't be much of a surprise.

dkillick:
Completed by #2576

PR #2580, opened 2017-05-24
Ensure cube.transpose functionality

### LINKED ISSUE #2579: [Dask] Cannot perform reordered multidim slice of lazy cube
In cases where a slice of a cube will require transposing the intermediate result (that is, when the requested slice dims are not in the same order as they are on the cube), the following error will occur:

```python
>>> cube = iris.load_cube(iris.sample_data_path('A1B_north_america.nc'))
>>> ht_slice = next(cube.slices(['latitude', 'time']))
  File "<input>", line 1, in <module>
  File "/.../git/iris/lib/iris/cube.py", line 3965, in __next__
    cube.transpose(self._mod_requested_dims)
  File "/.../git/iris/lib/iris/cube.py", line 2793, in transpose
    data = dm.core_data().transpose(new_order)
  File "/.../python2.7/site-packages/dask/array/core.py", line 1266, in transpose
    return transpose(self, axes=axes)
  File "/.../python2.7/site-packages/dask/array/core.py", line 2362, in transpose
    if axes:
ValueError: The truth value of an array with more than one element is ambiguous. Use a.any() or a.all()
```

This is because `iris.cube._SliceIterator` passes a NumPy array of reordering dims to `cube.transpose`, which dask cannot handle. This case was missed when `cube.transpose` was originally daskified, as was a test for this case that would have picked up on the error.

dkillick:
In some cases (notably when calling `cube.slices` with slice dims specified in a different order to their order on the cube), `cube.transpose` could be called with a NumPy array specifying the `new_order` kwarg. This is not supported by `dask.array.transpose`.

This PR ensures that `new_order` is **always** a list, which the previous checks did not ensure, and adds tests for this behaviour.

Fixes #2579 and required for #2460/#2576 to proceed.

### DIFF
--- a/lib/iris/tests/unit/cube/test_Cube.py
+++ b/lib/iris/tests/unit/cube/test_Cube.py
@@ -1926,6 +1926,5 @@ class Test_transpose(tests.IrisTest):
             self.cube.transpose([1])
 
 
-
 if __name__ == '__main__':
     tests.main()
--- a/lib/iris/tests/unit/cube/test_Cube.py
+++ b/lib/iris/tests/unit/cube/test_Cube.py
@@ -1880,20 +1880,51 @@ class TestCellMeasures(tests.IrisTest):
 
 
 class Test_transpose(tests.IrisTest):
+    def setUp(self):
+        self.data = np.arange(24).reshape(3, 2, 4)
+        self.cube = Cube(self.data)
+        self.lazy_cube = Cube(as_lazy_data(self.data))
+
     def test_lazy_data(self):
-        data = np.arange(12).reshape(3, 4)
-        cube = Cube(as_lazy_data(data))
+        cube = self.lazy_cube
         cube.transpose()
         self.assertTrue(cube.has_lazy_data())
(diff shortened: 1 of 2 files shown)


lib/iris/tests/unit/cube/test_Cube.py
```diff
@@ -1880,20 +1880,51 @@ def test_fail_cell_measure_dims(self):
...
+        cube.transpose(new_order)
+        self.assertTrue(cube.has_lazy_data())
+        self.assertArrayEqual(self.data.transpose(new_order), cube.data)
+
+    def test_bad_transpose_order(self):
+        exp_emsg = 'Incorrect number of dimensions'
+        with self.assertRaisesRegexp(ValueError, exp_emsg):
+            self.cube.transpose([1])
 
-    def test_not_lazy_data(self):
```
dkillick: Note this test method actually hasn't been removed! It has been changed a little and renamed to `test_real_data`.

dkillick:
Thanks @corinnebosley!

PR #2583, opened 2017-05-25
Dask docs, first draft

corinnebosley:
In this new page of the user guide, I have attempted to cover:
- What is real and lazy data?
- When does my lazy data become real?
- What is 'core data'?
- Different methods of changing a cube's data (cube maths, copy, replace) and their respective consequences
- Some dask processing options and what they might be useful for (very briefly of course, because I don't fully understand myself...)
- Some links to the dask and distributed docs.

### DIFF
--- a/docs/iris/src/userguide/real_and_lazy_data.rst
+++ b/docs/iris/src/userguide/real_and_lazy_data.rst
@@ -47,7 +47,7 @@ that is, the loading and manipulating of datasets that otherwise would not fit i
 You can check whether a cube has real data or lazy data by using the method
 :meth:`~iris.cube.Cube.has_lazy_data`. For example::
 
-    >>> cube = iris.load_cube(filename, 'air_temp.pp')
+    >>> cube = iris.load_cube(iris.sample_data_path('air_temp.pp'))
     >>> cube.has_lazy_data()
     True
     # Realise the lazy data.
--- a/docs/iris/src/userguide/real_and_lazy_data.rst
+++ b/docs/iris/src/userguide/real_and_lazy_data.rst
@@ -99,11 +99,13 @@ For example::
     >>> cube = iris.load_cube(iris.sample_data_path('air_temp.pp'))
     >>> cube.has_lazy_data()
     True
+
     >>> the_data = cube.core_data()
     >>> type(the_data)
     <class 'dask.array.core.Array'>
     >>> cube.has_lazy_data()
     True
+
     >>> cube.data
     >>> the_data = cube.core_data()
     >>> type(the_data)
@@ -133,6 +135,7 @@ coordinates' lazy points and bounds:
 .. doctest::
 
     >>> cube = iris.load_cube(iris.sample_data_path('hybrid_height.nc'))
+
     >>> dim_coord = cube.coord('model_level_number')
     >>> print dim_coord.has_lazy_points()
     False
@@ -140,6 +143,7 @@ coordinates' lazy points and bounds:
     False
     >>> print dim_coord.has_lazy_bounds()
     False
+
     >>> aux_coord = cube.coord('sigma')
     >>> print aux_coord.has_lazy_points()
     True
@@ -150,6 +154,8 @@ coordinates' lazy points and bounds:
     >>> points = aux_coord.points
     >>> print aux_coord.has_lazy_points()
     False
+
+    >>> derived_coord = cube.coord('altitude')
     >>> print derived_coord.has_lazy_points()
     True
     >>> print derived_coord.has_bounds()
--- a/docs/iris/src/userguide/real_and_lazy_data.rst
+++ b/docs/iris/src/userguide/real_and_lazy_data.rst
@@ -146,34 +146,34 @@ coordinates' lazy points and bounds:
     >>> cube = iris.load_cube(iris.sample_data_path('hybrid_height.nc'))
 
     >>> dim_coord = cube.coord('model_level_number')
-    >>> print dim_coord.has_lazy_points()
+    >>> print(dim_coord.has_lazy_points())
     False
-    >>> print dim_coord.has_bounds()
+    >>> print(dim_coord.has_bounds())
     False
-    >>> print dim_coord.has_lazy_bounds()
+    >>> print(dim_coord.has_lazy_bounds())
     False
 
     >>> aux_coord = cube.coord('sigma')
-    >>> print aux_coord.has_lazy_points()
+    >>> print(aux_coord.has_lazy_points())
     True
-    >>> print aux_coord.has_bounds()
+    >>> print(aux_coord.has_bounds())
     True
-    >>> print aux_coord.has_lazy_bounds()
+    >>> print(aux_coord.has_lazy_bounds())
     True
 
     # Realise the lazy points. This will **not** realise the lazy bounds.
     >>> points = aux_coord.points
-    >>> print aux_coord.has_lazy_points()
+    >>> print(aux_coord.has_lazy_points())
     False
-    >>> print aux_coord.has_lazy_bounds()
+    >>> print(aux_coord.has_lazy_bounds())
     True
 
     >>> derived_coord = cube.coord('altitude')
-    >>> print derived_coord.has_lazy_points()
+    >>> print(derived_coord.has_lazy_points())
     True
-    >>> print derived_coord.has_bounds()
+    >>> print(derived_coord.has_bounds())
     True
-    >>> print derived_coord.has_lazy_bounds()
+    >>> print(derived_coord.has_lazy_bounds())
     True
 
 .. note::
--- a/docs/iris/src/userguide/real_and_lazy_data.rst
+++ b/docs/iris/src/userguide/real_and_lazy_data.rst
@@ -13,10 +13,8 @@ Real and Lazy Data
 ==================
 
 We have seen in the :doc:`user_guide_introduction` section of the user guide that
-Iris cubes contain data and metadata about a phenomenon. The data attribute of a
-cube contains the actual numerical values recorded for the phenomenon the cube describes.
-The data element of a cube is always an array, but the array may be either
-"real" or "lazy".
+Iris cubes contain data and metadata about a phenomenon. The data element of a cube
+is always an array, but the array may be either "real" or "lazy".
 
 In this section of the user guide we will look specifically at the concepts of
 real and lazy data as they apply to the cube and other data structures in Iris.
@@ -59,10 +57,9 @@ When does my data become real?
 
 Most operations on data arrays can be run equivalently on both real and lazy data.
 If the data array is real then the operation will be run on the data array
-immediately with the results becoming available as soon as processing is completed.
-If the data array is lazy then the operation will be deferred until you request
-the result (such as when you call ``cube.data``). In this case the data array will
-remain lazy::
+immediately. The results of the operation will be available as soon as processing is completed.
+If the data array is lazy then the operation will be deferred and the data array will
+remain lazy until you request the result (such as when you call ``cube.data``)::
 
     >>> cube = iris.load_cube(iris.sample_data_path('air_temp.pp'))
     >>> cube.has_lazy_data()
@@ -71,27 +68,27 @@ remain lazy::
     >>> cube.has_lazy_data()
     True
 
-This is referred to as **lazy evaluation**.
+The process by which the operation is deferred until the result is requested is
+referred to as **lazy evaluation**.
 
 Certain operations, including regridding and plotting, can only be run on real data.
 Calling such operations on lazy data will automatically realise your lazy data.
 
 You can also realise (and so load into memory) your cube's lazy if you 'touch' the data.
-This means directly accessing the data by calling ``cube.data``, as in the previous example.
+To 'touch' the data means directly accessing the data by calling ``cube.data``,
+as in the previous example.
 
 Core data
 ^^^^^^^^^
 
 Cubes have the concept of "core data". This returns the cube's data in its
-current state. If a cube has lazy data, calling the cube's
-:meth:`~iris.cube.Cube.core_data` method will return the cube's lazy dask array.
-If the cube has real data, calling the cube's
-:meth:`~iris.cube.Cube.core_data` method will return the cube's real NumPy array.
+current state:
 
-Calling the cube's :meth:`~iris.cube.Cube.core_data` method will not change the
-state of the cube's data.  Thus, if the cube's data is lazy then calling
-:meth:`~iris.cube.Cube.core_data` will return the cube's lazy data and not
-realise it.
+ * If a cube has lazy data, calling the cube's :meth:`~iris.cube.Cube.core_data` method
+   will return the cube's lazy dask array. Calling the cube's
+   :meth:`~iris.cube.Cube.core_data` method **will not realise** the cube's data.
+ * If a cube has real data, calling the cube's :meth:`~iris.cube.Cube.core_data` method
+   will return the cube's real NumPy array.
 
 For example::
 
@@ -145,7 +142,7 @@ coordinates' lazy points and bounds:
     >>> print aux_coord.has_bounds()
     True
     >>> print aux_coord.has_lazy_bounds()
-    False
+    True
     >>> points = aux_coord.points
     >>> print aux_coord.has_lazy_points()
     False
@@ -154,12 +151,10 @@ coordinates' lazy points and bounds:
     >>> print derived_coord.has_bounds()
     True
     >>> print derived_coord.has_lazy_bounds()
-    False
+    True
 
 .. note::
     Printing a lazy :class:`~iris.coords.AuxCoord` will realise its points and bounds arrays!
-
-Derived coordinates (also called aux factories) .
 
 
 Dask processing options
@@ -169,10 +164,11 @@ As stated earlier in this user guide section, Iris uses dask to provide
 lazy data arrays for both Iris cubes and coordinates. Iris also uses dask
 functionality for processing deferred operations on lazy arrays.
 
-There are a wide range of dask processing options that can be adjusted to
-control how dask processes deferred operations on lazy arrays. You can make use
-of these dask processing options to control how lazy arrays within Iris are
-processed as well.
+Dask provides processing options to control how deferred operations on lazy arrays
+are computed. This is provided via the ``dask.set_options`` interface.
+We can make use of this functionality in Iris. This means we can
+control how dask arrays in Iris are processed, for example giving us power to
+run Iris processing in parallel.
 
 Iris by default applies a single dask processing option. This specifies that
 all dask processing in Iris should be run in serial (that is, without any
@@ -198,7 +194,7 @@ processing within Iris should use four workers in a thread pool::
     These dask processing options will last for the lifetime of the Python session
     and must be re-applied in other or subsequent sessions.
 
-See the
+Other dask processing options are also available. See the
 `dask documentation <http://dask.pydata.org/en/latest/scheduler-overview.html>`_
 for more information on setting dask processing options.
 
@@ -206,6 +202,7 @@ for more information on setting dask processing options.
 Further reading
 ---------------
 
-This section of the user guide has been designed to give a quick overview of the
-key concepts of real and lazy data within Iris. If you require more detail, we
-have produced a much more detailed whitepaper on the subject.
+This section of the user guide is intended as a quick overview of the
+key concepts of real and lazy data within Iris. For more detail and more in-depth
+discussions on the concepts introduced here and related concepts, see the related
+whitepaper.
--- a/docs/iris/src/userguide/real_and_lazy_data.rst
+++ b/docs/iris/src/userguide/real_and_lazy_data.rst
@@ -7,25 +7,30 @@ Real and Lazy Data
 What is Real and Lazy Data?
 ---------------------------
 
-Every Iris cube contains an n-dimensional  data array, which could be real or
+Every Iris cube contains an n-dimensional data array, which could be real or
 lazy.
 
-Real data is contained in an array which has a shape, a data type, some other
-useful information and many data points, each of which use up a small
-allocation of memory.  This generally takes the form of a numpy array.
+Real data is contained in a NumPy array which has a shape, a data type, some
+other useful information and many data points, each of which use up a small
+allocation of memory.
 
 Lazy data is contained in a conceptual array which retains the information
 about its real counterpart but has no actual data points, so its memory
 allocation is much smaller.  This will be in the form of a Dask array.
 
-Arrays in Iris can be converted flexibly(?) between their real and lazy states,
-although there are some limits to this process.  The advantage of using lazy
-data is that it has a small memory footprint, so certain operations
-(such as...?) can be much faster.  However, in order to perform other
-operations (such as calculations on actual data values) the real data must be
-realized.
+Arrays in Iris can be converted between their real and lazy states,
+although there are some limits to this process (*explain this better - what
+limits?*).
 
-* When a cube is loaded, are the data arrays always lazy to begin with? *
+The advantage of using lazy data is that it has a small memory footprint which
+enables the user to load and manipulate datasets that you would otherwise not
+be able to fit into memory.
+
+However, in order to execute certain operations (such as calculations on actual
+data values) the real data must be realized.  Using Dask, the operation will
+be deferred until you request the result, at which point it will be executed
+using Dask's parallel processing schedulers.  The combination of these two
+behaviours can offer a significant performance boost.
 
 You can check whether the data array on your cube is lazy using the Iris
 function 'has_lazy_data'.  For example:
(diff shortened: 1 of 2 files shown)


corinnebosley:
@dkillick @pp-mo @bjlittle Apologies for my previous PR, clearly not used to working on the dask branch.  

Feel free to tear into this draft of the docs, that's what it's there for.  Don't beat around the bush, I won't take anything personally.

docs/iris/src/userguide/real_and_lazy_data.rst
```diff
@@ -0,0 +1,225 @@
...
+allocation of memory.  This generally takes the form of a numpy array.
+
+Lazy data is contained in a conceptual array which retains the information
+about its real counterpart but has no actual data points, so its memory
+allocation is much smaller.  This will be in the form of a Dask array.
+
+Arrays in Iris can be converted flexibly(?) between their real and lazy states,
+although there are some limits to this process.  The advantage of using lazy
+data is that it has a small memory footprint, so certain operations
+(such as...?) can be much faster.  However, in order to perform other
```
dkillick: The advantage is quite right, but the application is not so accurate: lazy data allows you to load and manipulate datasets that you would otherwise not be able to fit into memory.

As it happens, it is likely that operations will be sped up, but this could be considered an indirect benefit:

* the calculations will not be performed immediately as you execute the lines of code that define the calculations. Instead they are **deferred** until you request a realised result (this happens when you call `compute` on a dask array)
* dask's parallel-processing schedulers allows the calculations to be performed in parallel when the realised result is requested, which can bring a significant performance improvement. However it is not _directly_ related to the smaller memory footprint.

I think both of these points are also worth adding here as they build the picture of the benefits of lazy data and its processing.

docs/iris/src/userguide/real_and_lazy_data.rst
```diff
@@ -0,0 +1,225 @@
...
+operations (such as calculations on actual data values) the real data must be
+realized.
+
+* When a cube is loaded, are the data arrays always lazy to begin with? *
```
dkillick: Usually but not for definite! Might be worth a section somewhere detailing this (but perhaps in the whitepaper as this is a more detail-oriented concept).

docs/iris/src/userguide/real_and_lazy_data.rst
```diff
@@ -0,0 +1,225 @@
...
+
+    >>> import iris
+    >>> cube = iris.load_cube(filename, 'air_temp.pp')
+    >>> cube.has_lazy_data()
+    True
+    >>> _ = cube.data
+    >>> cube.has_lazy_data()
+    False
+
+Assigning the data array to a variable causes the real data to be realized, at
```
dkillick: In fact, you could assign the lazy data array (e.g. `_ = cube.lazy_data()`) and this would not realise the data.

Perhaps the key thing here is instead "touching" the data, which you will need to define. If you access the data array of a cube you will realise the data and load it into memory.

docs/iris/src/userguide/real_and_lazy_data.rst
```diff
@@ -0,0 +1,225 @@
...
+lazy.  This can be used if you wish to refer to the data array but are
+indifferent to its current state.  If the cube's data is lazy, it will not be
+realized when you reference the core data attribute (?):
+
+.. doctest::
+
+    >>> cube = iris.load_cube(iris.sample_data_path('air_temp.pp'))
+    >>> cube.has_lazy_data()
+    True
+    >>> the_data = cube.core_data
```
dkillick: I believe `core_data()` is now a method that needs to be called.

docs/iris/src/userguide/real_and_lazy_data.rst
```diff
@@ -0,0 +1,225 @@
...
+    False
+    >>> new_cube.data = new_cube.lazy_data()
+    >>> new_cube.has_lazy_data()
+    True
+
+This method is necessary as Dask is currently unable to handle masked arrays.
+Please refer to the Whitepaper for further details.
+
+
+Dask Processing Options
```
dkillick: You know, I think we could do with a little section here on coords, too. They've been updated now as well, and while `DimCoords` can only have real data (because of the monotonicity checks that must be run), `AuxCoords` can now have lazy data, which is a bit of a big deal, I reckon.

docs/iris/src/userguide/real_and_lazy_data.rst
```diff
@@ -0,0 +1,225 @@
...
+
+Multiprocessing works well with strings, lists or custom Dask objects.
+
+
+You can choose to run all processes in serial (which is currently the Iris
+default):
+
+.. doctest::
+
+    >>> dask.set_options(get=dask.get)
```
dkillick: I think this should read `dask.async.get_sync`

docs/iris/src/userguide/real_and_lazy_data.rst
```diff
@@ -0,0 +1,225 @@
...
+Dask Processing Options
+-----------------------
+
+As well as Dask offering the benefit of a smaller memory footprint through the
+handling of lazy arrays, it can significantly speed up performance by allowing
+Iris to use multiprocessing.
+
+There are some default values which are set by Dask and passed through to Iris.
+If you wish to change these options, you can override them globally or using a
+context manager.
```
dkillick: Dask options are only controllable by calling `dask.set_options`, which isn't quite clear from this sentence. The changes you specify with this call will apply to all uses of dask for the lifetime of your "session" (be it a Python session or the extent of the context manager).

docs/iris/src/userguide/real_and_lazy_data.rst
```diff
@@ -0,0 +1,225 @@
...
+This method is necessary as Dask is currently unable to handle masked arrays.
+Please refer to the Whitepaper for further details.
+
+
+Dask Processing Options
+-----------------------
+
+As well as Dask offering the benefit of a smaller memory footprint through the
+handling of lazy arrays, it can significantly speed up performance by allowing
+Iris to use multiprocessing.
```
dkillick: I think I'd call this parallel processing rather than multiprocessing, which is one parallel processing option available to users.

docs/iris/src/userguide/real_and_lazy_data.rst
corinnebosley: Which bit?  Do you mean like this:
>>> dask.set_options(get=dask.async.get_sync)
?

corinnebosley: That was not how I expected that comment to look...

docs/iris/src/userguide/real_and_lazy_data.rst
corinnebosley: It seems to work both with and without the brackets.  That's pretty confusing.

docs/iris/src/userguide/real_and_lazy_data.rst
```diff
@@ -0,0 +1,246 @@
...
+    >>> cube.has_lazy_data()
+    True
+    >>> cube += 5
+    >>> cube.has_lazy_data()
+    True
+    >>> cube.data
+    >>> cube.has_lazy_data()
+    False
+
+You can also convert realized data back into a lazy array:
```
pp-mo: In my head "realized data" is maybe not good grammar, and simply "real data" is more accurate. :  so I'd prefer it to say that a cube or coord points/bounds can be "realized" -- the content  is just a "real array".

docs/iris/src/userguide/real_and_lazy_data.rst
```diff
@@ -0,0 +1,246 @@
...
+
+.. doctest::
+
+    >>> cube.has_lazy_data()
+    False
+    >>> cube.data = cube.lazy_data()
+    >>> cube.has_lazy_data()
+    True
+
+Core data refers to the current state of the cube's data, be it real or
```
pp-mo: Don't much like "state of the cube's data", it sounds like it is telling you what state it is on.
? Prefer "Core data returns the cube's current data array, of whichever type, be it real or lazy."

docs/iris/src/userguide/real_and_lazy_data.rst
```diff
@@ -0,0 +1,246 @@
...
+lazy.  This can be used if you wish to refer to the data array but are
+indifferent to its current state.  If the cube's data is lazy, it will not be
```
pp-mo: Again, don't much like this use of "state".
Maybe "This can be used when you want to use the data array in some way that works regardless of whether it is real or lazy, such as indexing it".

docs/iris/src/userguide/real_and_lazy_data.rst
```diff
@@ -0,0 +1,246 @@
...
+----------------------
+
+There are several methods of modifying a cube's data array, each one subtly
+different from the others.
+
+Maths
+^^^^^
+
+You can use :ref:`cube maths <cube_maths>` to make in-place modifications to
+each point in a cube's existing data array.  Provided you do not directly
```
pp-mo: I think "every" or "all" points is better than "each" -- I.E. it needs to be a whole-array operation.
The distinction is that you must avoid things like "cube.core_data()[:3] += 2.0",
\- which works fine on real data.

docs/iris/src/userguide/real_and_lazy_data.rst
```diff
@@ -0,0 +1,246 @@
...
+    >>> cube *= 10
+    >>> cube.has_lazy_data()
+    True
+
+Copy
+^^^^
+
+You can copy a cube and assign a completely new data array to the copy. All the
+original cube's metadata will be the same as the new cube's metadata.  However,
+the new cube's data array will not be lazy if you replace it with a real array:
```
pp-mo: You could make it clear here that you *can* provide a lazy array.
However, in certain cases you will need to use 'replace()' instead, described next...

docs/iris/src/userguide/real_and_lazy_data.rst
```diff
@@ -0,0 +1,246 @@
...
+    >>> data =np.ma.masked_values(values, 0)
+    >>> print(data)
+    [[-- -- -- ..., -- -- --]
+     [-- -- -- ..., -- -- --]
+     [-- -- -- ..., -- -- --]
+     ...,
+     [-- -- -- ..., -- -- --]
+     [-- -- -- ..., -- -- --]
+     [-- -- -- ..., -- -- --]]
+    >>> new_cube = cube.copy(data=data)
```
pp-mo: Needs a `replace` !

docs/iris/src/userguide/real_and_lazy_data.rst
```diff
@@ -0,0 +1,246 @@
...
+    True
+
+This method is necessary as Dask is currently unable to handle masked arrays.
+Please refer to the Whitepaper for further details.
+
+
+Coordinate Arrays
+-----------------
+
+Cubes possess coordinate arrays as well as data arrays, so these also benefit
```
pp-mo: I'd put this as "Like cubes, coordinates also contain data arrays (points and bounds)."

docs/iris/src/userguide/real_and_lazy_data.rst
```diff
@@ -0,0 +1,246 @@
...
+undergo monotonicity checks which cause the arrays to be realized upon
+construction, so they can only contain real arrays.
+
+
+Dask Processing Options
+-----------------------
+Dask applies some default values to certain aspects of the parallel processing
+that it offers with Iris. It is possible to change these values and override
+the defaults by using 'dask.set_options(option)' in your script.
+
```
pp-mo: worth describing exactly *what* Dask options Iris sets first, linking it directly to the account in the Dask docs.
It may also be worth explaining how Iris "steps aside" if you make your own set_options call.

docs/iris/src/userguide/real_and_lazy_data.rst
```diff
@@ -0,0 +1,246 @@
...
+You can use this as a global variable if you wish to use your chosen option for
```
pp-mo: I'm worried that, while maybe useful, all this remaining text is duplicating the Dask documentation.
Apart from the general problems with copying and coupling, I think we really ***want*** the user to refer to Dask docs at this point + not just copy our magic recipes without a deeper understanding.

docs/iris/src/userguide/real_and_lazy_data.rst
```diff
@@ -0,0 +1,246 @@
...
+Every Iris cube contains an n-dimensional data array, which could be real or
+lazy.
+
+Real data is contained in a NumPy array which has a shape, a data type, some
+other useful information and many data points, each of which use up a small
+allocation of memory.
+
+Lazy data is contained in a conceptual array which retains the information
+about its real counterpart but has no actual data points, so its memory
+allocation is much smaller.  This will be in the form of a Dask array.
```
bjlittle: @corinnebosley Perhaps format `Dask` as a link to the dask readthedocs, so the user can follow this if they want ...

docs/iris/src/userguide/real_and_lazy_data.rst
```diff
@@ -0,0 +1,246 @@
...
+
+Arrays in Iris can be converted between their real and lazy states,
+although there are some limits to this process (*explain this better - what
+limits?*).
+
+The advantage of using lazy data is that it has a small memory footprint which
+enables the user to load and manipulate datasets that you would otherwise not
+be able to fit into memory.
```
bjlittle: @corinnebosley Perhaps introduce the concepts of `deferred loading` i.e. keeping large data on disk and not in memory

docs/iris/src/userguide/real_and_lazy_data.rst
```diff
@@ -0,0 +1,246 @@
...
+
+However, in order to execute certain operations (such as calculations on actual
+data values) the real data must be realized.  Using Dask, the operation will
+be deferred until you request the result, at which point it will be executed
+using Dask's parallel processing schedulers.  The combination of these two
+behaviours can offer a significant performance boost.
```
bjlittle: @corinnebosley Introduce the concept (and use the term) `lazy evaluation` ...

docs/iris/src/userguide/real_and_lazy_data.rst
```diff
@@ -0,0 +1,246 @@
...
+
+    >>> import iris
+    >>> cube = iris.load_cube(filename, 'air_temp.pp')
+    >>> cube.has_lazy_data()
+    True
+    >>> cube.data
+    >>> cube.has_lazy_data()
+    False
+
+When does my Data Become Real?
```
bjlittle: `When does my data become real?`

docs/iris/src/userguide/real_and_lazy_data.rst
```diff
@@ -0,0 +1,246 @@
...
+This method is necessary as Dask is currently unable to handle masked arrays.
+Please refer to the Whitepaper for further details.
+
+
+Coordinate Arrays
+-----------------
+
+Cubes possess coordinate arrays as well as data arrays, so these also benefit
+from Dask's functionality, although there are some distinctions between how
+the different coordinate types are treated.
```
bjlittle: I think that we need to make a clear statement that coordinates etc are now `lazy`, just like the data of a cube, rather than mentioning that we use dask, which implies lazy behaviour is supported ... subtle, but perhaps important.

docs/iris/src/userguide/real_and_lazy_data.rst
dkillick: So "state" is my suggested terminology for `core_data`, although maybe its use here is a little clunky. I mean, the state of the cube's data is likely to change over the lifetime of the cube, between being in a lazy state and a real state. 

I'll have a go at using "state" differently here and if it still isn't liked then maybe we could go for something a little more like what @pp-mo has suggested.

docs/iris/src/userguide/real_and_lazy_data.rst
dkillick: Actually, this is written in title case, which is fine... other than the fact that the rest of the user guide doesn't appear to use title case.

docs/iris/src/userguide/real_and_lazy_data.rst
```diff
@@ -0,0 +1,246 @@
...
+However, in order to execute certain operations (such as calculations on actual
+data values) the real data must be realized.  Using Dask, the operation will
+be deferred until you request the result, at which point it will be executed
+using Dask's parallel processing schedulers.  The combination of these two
+behaviours can offer a significant performance boost.
+
+You can check whether the data array on your cube is lazy using the Iris
+function 'has_lazy_data'.  For example:
+
+.. doctest::
```
dkillick: We hardly use the `doctest` directive anywhere else in the userguide, and this implementation is incorrect (no accompanying `testsetup` directive). We could do with a decision on how much, if at all, we use `doctest` in this userguide chapter.

docs/iris/src/userguide/real_and_lazy_data.rst
```diff
@@ -0,0 +1,217 @@
...
+take up a large amount of memory.
+
+Conversely, we use the term **lazy data** to describe data that is not loaded into memory.
+(This is sometimes also referred to as **deferred data**.)
+In Iris, lazy data is provided as a
+`dask array <http://dask.pydata.org/en/latest/array-overview.html>`_.
+A dask array also has a shape and data type
+but typically the dask array's data points are not loaded into memory.
+Instead the data points are stored on disk and only loaded into memory in
+small chunks when absolutely necessary.
```
bjlittle: `small chunks` -> did we want a link here to the dask readthedocs section on chunks? if we use dask terms, we should point users off to their documentation.
`absolutely necessary` -> did we want to qualify this some how here i.e. explicitly loading or crunching numbers to do an actual computation (not a lazy computation)

docs/iris/src/userguide/real_and_lazy_data.rst
```diff
@@ -0,0 +1,217 @@
...
+
+The primary advantage of using lazy data is that it enables the loading and manipulating
+of datasets that would otherwise not fit into memory.
```
bjlittle: Reference `out-of-core processing` i.e. it's a real term that means exactly what we want e.g. see https://en.wikipedia.org/wiki/Out-of-core_algorithm

docs/iris/src/userguide/real_and_lazy_data.rst
```diff
@@ -0,0 +1,217 @@
...
+
+You can check whether a cube has real data or lazy data by using the method
+:meth:`~iris.cube.Cube.has_lazy_data`. For example::
+
+    >>> cube = iris.load_cube(filename, 'air_temp.pp')
+    >>> cube.has_lazy_data()
+    True
+    >>> cube.data
+    >>> cube.has_lazy_data()
+    False
```
bjlittle: Did we want to add a note, which reinforces that `cube.data` will make lazy data, real data by loading it into memory.

Did we want to introduce the term `realising the data` to mean making lazy data real? and use it through-out ?

bjlittle: Oh, that would just tee up the next section nicely ...

docs/iris/src/userguide/real_and_lazy_data.rst
```diff
@@ -0,0 +1,217 @@
...
+
+When does my data become real?
+------------------------------
+
+When you load a dataset using Iris the data array will almost always initially be
+a lazy array. This section details some operations that will realise lazy data
```
bjlittle: `details` -> `discusses` ?

docs/iris/src/userguide/real_and_lazy_data.rst
```diff
@@ -0,0 +1,217 @@
...
+as well as some operations that will maintain lazy data.
+
+Most operations on data arrays can be run equivalently on both real and lazy data.
+If the data array is real then the operation will be run on the data array
+immediately. The results of the operation will be available as soon as processing is completed.
+If the data array is lazy then the operation will be deferred and the data array will
```
bjlittle: highlight `deferred` as it's key to this sentence ? is there a link to an explanation in dask for deferred operations ? might be a good opportunity to cross-reference into the dask documentation ...

docs/iris/src/userguide/real_and_lazy_data.rst
```diff
@@ -0,0 +1,217 @@
...
+
+Core data
+^^^^^^^^^
+
+Cubes have the concept of "core data". This returns the cube's data in its
+current state:
+
+ * If a cube has lazy data, calling the cube's :meth:`~iris.cube.Cube.core_data` method
+   will return the cube's lazy dask array. Calling the cube's
+   :meth:`~iris.cube.Cube.core_data` method **will not realise** the cube's data.
```
bjlittle: `will not realise` -> `will never realise` is a stronger contract ...

docs/iris/src/userguide/real_and_lazy_data.rst
```diff
@@ -0,0 +1,217 @@
...
+ * If a cube has real data, calling the cube's :meth:`~iris.cube.Cube.core_data` method
+   will return the cube's real NumPy array.
```
bjlittle: `real` -> `in-memory` ? ... or `real` -> `real (or in-memory)` ...

docs/iris/src/userguide/real_and_lazy_data.rst
```diff
@@ -0,0 +1,217 @@
...
+    <type 'numpy.ndarray'>
+    >>> cube.has_lazy_data()
+    False
+
+
+Coordinates
+-----------
+
+In the same way that Iris cubes contain a data array, Iris coordinates contain
+points and bounds arrays. Coordinate points and bounds arrays can also be real or lazy:
```
bjlittle: `bounds` are optional ... don't know if "optional" is the right word, but something that implies that a coordinate doesn't need to have bounds but it must have points (you get my drift, right)

docs/iris/src/userguide/real_and_lazy_data.rst
```diff
@@ -0,0 +1,217 @@
...
+
+ * A :class:`~iris.coords.DimCoord` will only ever have **real** points and bounds
+   arrays because of monotonicity checks that realise lazy arrays.
+ * An :class:`~iris.coords.AuxCoord` can have **real or lazy** points and bounds.
+ * An :class:`~iris.aux_factory.AuxCoordFactory` (or derived coordinate)
+   can have **real or lazy** points and bounds. If all of the
+   :class:`~iris.coords.AuxCoord` instances that the coordinate is derived from have
+   real points and bounds then the derived coordinate will also have real points
+   and bounds, otherwise the derived coordinate will have lazy points and bounds.
```
bjlittle: `If all ...` Difficult sentence. Totally understand what you mean, just wonder if we can make it simpler ... e.g.

`Note that, a derived coordinate will only have real points and bounds if all of it's associated AuxCoord instances have real points and bounds.`

docs/iris/src/userguide/real_and_lazy_data.rst
```diff
@@ -0,0 +1,217 @@
...
+    >>> aux_coord = cube.coord('sigma')
+    >>> print aux_coord.has_lazy_points()
+    True
+    >>> print aux_coord.has_bounds()
+    True
+    >>> print aux_coord.has_lazy_bounds()
+    True
+    >>> points = aux_coord.points
+    >>> print aux_coord.has_lazy_points()
+    False
```
bjlittle: Would be nice to now show that touching the points doesn't effect the laziness of the coordinates bounds i.e. they are still lazy ... that's the case, right?

docs/iris/src/userguide/real_and_lazy_data.rst
```diff
@@ -0,0 +1,217 @@
...
+
+.. note::
+    Printing a lazy :class:`~iris.coords.AuxCoord` will realise its points and bounds arrays!
+
+
+Dask processing options
+-----------------------
+
+As stated earlier in this user guide section, Iris uses dask to provide
+lazy data arrays for both Iris cubes and coordinates. Iris also uses dask
```
bjlittle: Should we actually mention the real dask thing i.e. `dask.array` with a a link to dask-land?

bjlittle:
@corinnebosley and @dkillick Great work! :+1: 

We've got and outstanding issue with a doc-test failure in the new user guide. I vote note that we only merge this PR once we resolve that issue first (and rebase this PR with the fix)

docs/iris/src/userguide/real_and_lazy_data.rst
```diff
@@ -0,0 +1,217 @@
...
+Instead the data points are stored on disk and only loaded into memory in
+small chunks when absolutely necessary.
+
+The primary advantage of using lazy data is that it enables the loading and manipulating
+of datasets that would otherwise not fit into memory.
+
+You can check whether a cube has real data or lazy data by using the method
+:meth:`~iris.cube.Cube.has_lazy_data`. For example::
+
+    >>> cube = iris.load_cube(filename, 'air_temp.pp')
```
dkillick: Whoops, we should probably fix 'filename' in this line...

corinnebosley: Thanks @dkillick, good spot.

bjlittle:
@corinnebosley It's all lookin' good!

I forgot to merge #2590 before you done your rebase, duh! :roll_eyes: ... but now that I've done that, could you please rebase again? This should make the failing doc-test pass!

corinnebosley:
@bjlittle All passing now.

PR #2584, opened 2017-05-26
share data flag behaviour

marqh:
addressing concerns raised in

https://github.com/SciTools/iris/pull/2549#discussion_r118005791


cpelley:
Thanks @marqh

PR #2588, opened 2017-05-31
Maintain lazy bounds in Aux Factory derivation

dkillick:
If you print a cube with a bounded derived coordinate it realises the lazy bounds of the coordinates from which the derived coordinate is derived: :scream: 

```python
>>> cube = iris.load_cube(iris.sample_data_path('hybrid_height.nc'))
>>> aux_coord = cube.coord('sigma')
>>> print aux_coord.has_lazy_points()
True
>>> print aux_coord.has_lazy_bounds()
True

>>> print cube

>>> print aux_coord.has_lazy_points()
True
>>> print aux_coord.has_lazy_bounds()
False
```

More accurately, printing a cube with a derived (bounded) coordinate triggers the deferred construction of the derived coordinate. If this derived coordinate is bounded then an oversight in the code that calculates the derived coordinate from the dependency coordinates, such as `sigma` in the example above, would cause the bounds to be accessed directly and so realise the bounds.


bjlittle:
Closed by #2590

PR #2589, opened 2017-06-01
Add or improve docstrings for Cube

dkillick:
Adds docstrings to new `iris.cube.Cube` functionality that has been added as part of the dask integration work. Replaces one illegible docstring and tidies a couple of other existing docstrings.

### DIFF
--- a/lib/iris/cube.py
+++ b/lib/iris/cube.py
@@ -885,7 +885,7 @@ bound=(1994-12-01 00:00:00, 1998-12-01 00:00:00)
         self.units = unit
 
     def add_cell_method(self, cell_method):
-        """Add a CellMethod to the Cube."""
+        """Add a :class:`~iris.coords.CellMethod` to the Cube."""
         self.cell_methods += (cell_method, )
 
     def add_aux_coord(self, coord, data_dims=None):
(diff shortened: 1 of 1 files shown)


lib/iris/cube.py
```diff
@@ -1615,10 +1615,21 @@ def shape(self):
 
     @property
     def dtype(self):
+        """
+        The data type of the values in data array of this
+        :class:`~iris.cube.Cube`.
```
bjlittle: @dkillick Doesn't quite make sense ...

lib/iris/cube.py
```diff
@@ -1615,10 +1615,21 @@ def shape(self):
...
+
+        """
         return self._data_manager.dtype
 
     @property
     def fill_value(self):
+        """
+        The fill value for the masked data array of this
+        :class:`~iris.cube.Cube`. Will be ``None`` if the data array is
+        not masked.
```
bjlittle: @dkillick The data of the cube doesn't need to be masked for the cube to have a `fill_value`. This allows users to control any fill value that is saved to a file format, for example.

A `fill_value` of `None` simply means that no fill value has been explicitly specified. This means that if the data is masked, then the fill value of the `cube.data` returned to the user will be the default `numpy` fill value for the `dtype` of the data.

lib/iris/cube.py
```diff
@@ -1599,11 +1599,11 @@ def cell_methods(self, cell_methods):
...
         """
-        The data at the core of this cube.
-        May be a numpy array or a dask array.
-        In using this, you are buying into not caring about the
-        type of the result
-        to be decided: should this be public??
+        Retrieve the data array of this :class:`~iris.cube.Cube` in its
+        current state, which will be either real or lazy.
+
+        If this :class:`~iris.cube.Cube` has lazy data, accessing its data
```
pp-mo: I don't like that we lost the purpose description : i.e. it needs saying that the *point* of core_data is :
"so you can implement operations that work irrespective of the type of underlying data, and thus preserve laziness".
I mean of course, something like that but much shorter clearer and more intelligible :  Over to you :wink:

dkillick:
Closed in favour of #2625.

PR #2590, opened 2017-06-05
Lazy bounds test

corinnebosley:
This is a new PR to replace https://github.com/SciTools/iris/pull/2588.  This entails the one-line change which fixes the realization of derived coord lazy bounds upon printing the cube, along with a test to ensure its effectiveness.

### DIFF
--- a/lib/iris/tests/test_coord_api.py
+++ b/lib/iris/tests/test_coord_api.py
@@ -255,16 +255,6 @@ class TestAuxCoordCreation(tests.IrisTest):
                   ", standard_name='air_temperature', units=Unit('kelvin'))"
                   )
         self.assertEqual(result, str(a))
-
-    @tests.skip_data
-    def test_lazy_bounds(self):
-        # Make sure that bounds arrays loaded as lazy do not get realised
-        # when the cube is printed.
-        cube = iris.load_cube(tests.get_data_path(['NetCDF', 'testing',
-                                                   'small_theta_colpex.nc']))
-        self.assertTrue(cube.coord('sigma').has_lazy_bounds())
-        printed_cube = str(cube)
-        self.assertTrue(cube.coord('sigma').has_lazy_bounds())
         
     def test_string_coord_equality(self):
         b = iris.coords.AuxCoord(['Jan', 'Feb', 'March'], units='no_unit')
--- a/lib/iris/tests/test_coord_api.py
+++ b/lib/iris/tests/test_coord_api.py
@@ -256,6 +256,7 @@ class TestAuxCoordCreation(tests.IrisTest):
                   )
         self.assertEqual(result, str(a))
 
+    @tests.skip_data
     def test_lazy_bounds(self):
         # Make sure that bounds arrays loaded as lazy do not get realised
         # when the cube is printed.
--- a/lib/iris/tests/test_coord_api.py
+++ b/lib/iris/tests/test_coord_api.py
@@ -255,6 +255,13 @@ class TestAuxCoordCreation(tests.IrisTest):
                   ", standard_name='air_temperature', units=Unit('kelvin'))"
                   )
         self.assertEqual(result, str(a))
+
+    def test_lazy_bounds(self):
+        # Make sure that bounds arrays loaded as lazy do not get realised
+        # by the loading process.
+        cube = iris.load_cube(iris.sample_data_path('hybrid_height.nc'))
+        cube
+        self.assertTrue(cube.coord('sigma').has_lazy_bounds())
         
     def test_string_coord_equality(self):
         b = iris.coords.AuxCoord(['Jan', 'Feb', 'March'], units='no_unit')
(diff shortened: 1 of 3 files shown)


corinnebosley:
@dkillick @bjlittle Here is the test that I have popped in, I would appreciate opinions from both of you (if possible).

@dkillick has suggested the adapt or duplicate the test 'Test__nd_points' in iris/tests/unit/aux_factory/test_AuxCoordFactory.py.  I have my reservations about duplicating tests, but I could maybe adapt this class to test both points and bounds arrays.  What do you think?

corinnebosley:
@dkillick @bjlittle Bear in mind that adapting this test is not turning out as easy or straightforward as it first appeared...

dkillick:
@corinnebosley I like the test you've added here - it directly tests the problem behaviour that we're observing.

My suggestion for extra tests is that we add a new test class to test the behaviour of `AuxCoordFactory._nd_bounds`. I reckon that the tests for this method will be similar to the tests for `_nd_points`, which is why I suggested taking that route.

I definitely think we shouldn't update `Test__nd_points` - it is already doing a fair job of testing the `_nd_points` method.

corinnebosley:
@dkillick Also, and this is one for you specifically @dkillick, you mentioned the other day that there were two extra things that I had to look for in a PR; one was a test and we never actually got round to the other.  What was it?  A What's New entry? Or something else?

dkillick:
@corinnebosley ummm... I'll let you know if I remember what it was I was thinking of!

lib/iris/tests/test_coord_api.py
```diff
@@ -255,6 +255,15 @@ def test_bounded(self):
                   ", standard_name='air_temperature', units=Unit('kelvin'))"
                   )
         self.assertEqual(result, str(a))
+
```
corinnebosley: I need to add a test data skipper here.

corinnebosley:
@dkillick So if I want to replicate the _nd_points tests for _nd_bounds, there are lots of things which suddenly become complicated.  Like for example, first I have to either construct a bounds array to test the laziness of or I have to use existing test data.  Similarly, when I come to make the result using a line like this:  
```python
result = AuxCoordFactory._nd_bounds(coord, (3, 2), 5)
```
Python complains, telling me that I need to give it an AuxCoordFactory when I am actually giving it an Aux Coord instance.  Again, I could either spend a lot of time trying to construct a cube which can make derived coordinates, or I could use some existing test data.

In both cases, the former solution seems like more trouble than it's worth and the latter option seems to end in replication of the test I have already pushed up, so I am tempted to just keep the test as it is rather than adding extras.

Would appreciate your opinions though @dkillick and @bjlittle.

lib/iris/tests/unit/aux_factory/test_AuxCoordFactory.py
```diff
@@ -80,5 +80,82 @@ def test_lazy_complex(self):
...
+        expected = bounds
+        self.assertArrayEqual(result, expected)
+
+    def test_numpy_complex(self):
+        points = np.arange(12).reshape(4, 3)
+        bounds = np.arange(24).reshape(4, 3, 2)
+        coord = iris.coords.AuxCoord(points, bounds=bounds)
+        result = AuxCoordFactory._nd_bounds(coord, (3, 2), 5)
+        expected = bounds.T[np.newaxis, np.newaxis, ..., np.newaxis]
+        self.assertArrayEqual(result, expected)
```
corinnebosley: @bjlittle This test fails due to a mismatch between the resultant and expected order of coordinates after transpose.

lib/iris/tests/unit/aux_factory/test_AuxCoordFactory.py
```diff
@@ -80,5 +80,82 @@ def test_lazy_complex(self):
...
+        raw_bounds = np.arange(24).reshape(4, 3, 2)
+        bounds = as_lazy_data(raw_bounds, 1)
+        coord = iris.coords.AuxCoord(points, bounds=bounds)
+        self.assertTrue(is_lazy_data(coord.core_bounds()))
+        result = AuxCoordFactory._nd_bounds(coord, (3, 2), 5)
+        # Check we haven't triggered the loading of the coordinate values.
+        self.assertTrue(is_lazy_data(coord.core_bounds()))
+        self.assertTrue(is_lazy_data(result))
+        expected = raw_bounds.T[np.newaxis, np.newaxis, ..., np.newaxis]
+        self.assertArrayEqual(result, expected)
```
corinnebosley: @bjlittle This test fails due because of something about the number of chunks that dask wants to divide stuff up into.

bjlittle:
closes #2588

lib/iris/tests/unit/aux_factory/test_AuxCoordFactory.py
bjlittle: @corinnebosley Doing the `bounds.T[...]` isn't quite what we want here, I've addressed this in https://github.com/corinnebosley/iris/pull/10/files

lib/iris/tests/unit/aux_factory/test_AuxCoordFactory.py
bjlittle: @corinnebosley Turns out that for this test we were setting the `chunks` incorrectly. By default, lazy coordinates will have `chunks` set to the `shape` of the coordinate. See https://github.com/corinnebosley/iris/pull/10/files

bjlittle:
@corinnebosley and @dkillick Awesome, thanks :+1:

PR #2625, opened 2017-06-28
Add or improve docstrings for Cube

dkillick:
Adds docstrings to new `iris.cube.Cube` functionality that has been added during the `dask` integration work. Replaces one illegible docstring and tidies a few existing docstrings.

Replaces #2589 (targets master without the need for hefty rebase/retarget efforts).

### DIFF
--- a/lib/iris/cube.py
+++ b/lib/iris/cube.py
@@ -1603,7 +1603,9 @@ bound=(1994-12-01 00:00:00, 1998-12-01 00:00:00)
         current state, which will be either real or lazy.
 
         If this :class:`~iris.cube.Cube` has lazy data, accessing its data
-        array via this method **will not** realise the data array.
+        array via this method **will not** realise the data array. This means
+        you can perform operations using this method that work equivalently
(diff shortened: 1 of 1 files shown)


lib/iris/cube.py
```diff
@@ -1615,10 +1617,21 @@ def shape(self):
...
+        The data type of the values in the data array of this
+        :class:`~iris.cube.Cube`.
+
+        """
         return self._data_manager.dtype
 
     @property
     def fill_value(self):
+        """
+        The fill value for the masked data array of this
```
pp-mo: > I believe it is correct to say something like the following ..

..but I'm not actually suggesting these words : seems a bit too much ?

pp-mo:
> "I'm not actually suggesting these words :" ...

As @dkillick is about to be elsewhere for a while, I'm going to merge this + propose concrete improvements in a followup PR instead ...

pp-mo:
> followup PR

At : #2628

PR #2628, opened 2017-06-29
Tweaks to cube docstrings.

pp-mo:
Proposals to fix some issues I raised against #2625

### DIFF
--- a/lib/iris/cube.py
+++ b/lib/iris/cube.py
@@ -1600,7 +1600,7 @@ bound=(1994-12-01 00:00:00, 1998-12-01 00:00:00)
     def core_data(self):
         """
         Retrieve the data array of this :class:`~iris.cube.Cube` in its
-        current state, which will be either real or lazy.
+        current state, which may be either real or lazy.
 
         If this :class:`~iris.cube.Cube` has lazy data, accessing its data
         array via this method **will not** realise the data array. This means
@@ -1627,9 +1627,17 @@ bound=(1994-12-01 00:00:00, 1998-12-01 00:00:00)
     @property
     def fill_value(self):
         """
-        The fill value for the masked data array of this
-        :class:`~iris.cube.Cube`. If ``None``, the default NumPy fill value for
-        the dtype of the masked data array will be used.
+        A value suitable for filling in missing (i.e. masked) datapoints of
+        this :class:`~iris.cube.Cube` s data.
+
+        This can also be ``None``, meaning that no suitable fill value is
+        known.
+
+        .. Note::
+
+            Ideally, this value will not occur anywhere in `cube.data`. Thus,
+            many operations which change cube data will also set `fill_value`
+            to None.
 
         """
         return self._data_manager.fill_value
--- a/lib/iris/cube.py
+++ b/lib/iris/cube.py
@@ -1627,17 +1627,16 @@ bound=(1994-12-01 00:00:00, 1998-12-01 00:00:00)
     @property
     def fill_value(self):
         """
-        A value suitable for filling in missing (i.e. masked) datapoints of
-        this :class:`~iris.cube.Cube` s data.
+        A fill value for the data of the :class:`~iris.cube.Cube`.
 
-        This can also be ``None``, meaning that no suitable fill value is
-        known.
+        This is a value suitable for filling masked points in this cube's data.
+        It may be None, meaning that no suitable fill value is known.
 
         .. Note::
 
-            Ideally, this value will not occur anywhere in `cube.data`. Thus,
-            many operations which change cube data will also set `fill_value`
-            to None.
+            Ideally, this value will not occur anywhere in the cube data.
+            Thus, many operations that change cube data will also set
+            `fill_value` to None.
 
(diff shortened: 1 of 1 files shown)


lib/iris/cube.py
```diff
@@ -1627,9 +1627,17 @@ def dtype(self):
...
+        A value suitable for filling in missing (i.e. masked) datapoints of
+        this :class:`~iris.cube.Cube` s data.
+
+        This can also be ``None``, meaning that no suitable fill value is
+        known.
+
+        .. Note::
+
+            Ideally, this value will not occur anywhere in `cube.data`. Thus,
+            many operations which change cube data will also set `fill_value`
```
bjlittle: @pp-mo `operations that change` rather than `operations which change`

lib/iris/cube.py
```diff
@@ -1627,9 +1627,17 @@ def dtype(self):
...
-        the dtype of the masked data array will be used.
+        A value suitable for filling in missing (i.e. masked) datapoints of
+        this :class:`~iris.cube.Cube` s data.
+
+        This can also be ``None``, meaning that no suitable fill value is
+        known.
+
+        .. Note::
+
+            Ideally, this value will not occur anywhere in `cube.data`. Thus,
```
bjlittle: @pp-mo `... anywhere in the cube data`

lib/iris/cube.py
```diff
@@ -1600,7 +1600,7 @@ def cell_methods(self, cell_methods):
     def core_data(self):
         """
         Retrieve the data array of this :class:`~iris.cube.Cube` in its
-        current state, which will be either real or lazy.
+        current state, which may be either real or lazy.
```
dkillick: You're going to have to try quite hard to convince me that this is a necessary change. The core data of a cube will _either_ be real or lazy, which my original implies. Changing 'will' to 'may' reads more like 'it might be one of these, but it might be something else instead' 🍌 

We're in the realm of technical documentation here, where absolute clarity is key. Using 'will' gives us that clarity; I'd argue using 'may' does not.

lib/iris/cube.py
```diff
@@ -1627,9 +1627,17 @@ def dtype(self):
     @property
     def fill_value(self):
         """
-        The fill value for the masked data array of this
-        :class:`~iris.cube.Cube`. If ``None``, the default NumPy fill value for
-        the dtype of the masked data array will be used.
+        A value suitable for filling in missing (i.e. masked) datapoints of
+        this :class:`~iris.cube.Cube` s data.
```
dkillick: Could you please try and avoid forms like `:class:~iris.cube.Cube s` as it just looks a little bad...

lib/iris/cube.py
```diff
@@ -1627,9 +1627,17 @@ def dtype(self):
...
+
+        This can also be ``None``, meaning that no suitable fill value is
```
dkillick: Can you remove 'also' please, and add this to the former paragraph? I think this line is part of the same concept as the above sentence.

lib/iris/cube.py
pp-mo: This is not the general which / that argument, surely, because this "which" is not a pronoun ?

lib/iris/cube.py
dkillick: Maybe... 'will either be'?

lib/iris/cube.py
dkillick: Let's see if we can get to the situation of having a one-line description of this property and then also have most of the contents of these two paragraphs as one paragraph that follows it.

For example:
```
"""
A fill value for the data of the :class:cube.

This is a value suitable for filling masked points in this cube's data.
This may be None, meaning no suitable fill value has been defined...
"""
```

pp-mo:
Hope this addresses.
@bjlittle are you ok to go with this now ?

lib/iris/cube.py
```diff
@@ -1627,17 +1627,16 @@ def dtype(self):
...
     def fill_value(self):
         """
-        A value suitable for filling in missing (i.e. masked) datapoints of
-        this :class:`~iris.cube.Cube` s data.
+        A fill value for the data of the :class:`~iris.cube.Cube`.
 
-        This can also be ``None``, meaning that no suitable fill value is
-        known.
+        This is a value suitable for filling masked points in this cube's data.
+        It may be None, meaning that no suitable fill value is known.
```
dkillick: Best to put `None` in backticks IMO, like it was before.

lib/iris/cube.py
```diff
@@ -1627,17 +1627,16 @@ def dtype(self):
...
 
         .. Note::
 
-            Ideally, this value will not occur anywhere in `cube.data`. Thus,
-            many operations which change cube data will also set `fill_value`
-            to None.
+            Ideally, this value will not occur anywhere in the cube data.
+            Thus, many operations that change cube data will also set
+            `fill_value` to None.
```
dkillick: Double backticks for consistency? (And for making it display as pre-formatted text.) Sorry, should have picked up on this before now...

PR #2629, opened 2017-06-29
Added warnings of dask incompatibilities to whatsnew.

pp-mo:

PR #2627, opened 2017-06-29
Suggestions for improved release docs.

pp-mo:
Following on from #2620

Supposedly addresses all (my) outstanding comments on that, 
plus some other stuff.

PR #2631, opened 2017-06-30
Minor doc changes.

bjlittle:
Addresses the outstanding actions in #2627, #2628, and #2629.


pp-mo:
Checked against previous comments -- all seems good to me

PR #2633, opened 2017-07-03
WIP: Attempt to store conda package cache in Travis.

pp-mo:
Trial - **DO NOT MERGE**

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -43,7 +43,9 @@ install:
       wget ${CONDA_BASE}3-latest-Linux-x86_64.sh -O miniconda.sh;
     fi
   - bash miniconda.sh -b -p $HOME/miniconda
-  - cp $HOME/miniconda_pkgs_backup/* $HOME/miniconda/pkgs/
+  - if [[ -e $HOME/miniconda_pkgs_backup ]]; then
+      cp $HOME/miniconda_pkgs_backup/* $HOME/miniconda/pkgs/;
+    fi
   - export PATH="$HOME/miniconda/bin:$PATH"
 
   # Create the basic testing environment
@@ -77,6 +79,9 @@ install:
   - conda info -a
 
   # Create packages backup
+  - if [[ ! -e $HOME/miniconda_pkgs_backup ]]; then
+      mkdir $HOME/miniconda_pkgs_backup;
+    fi
   - cp $HOME/miniconda/pkgs/* $HOME/miniconda_pkgs_backup/
 
 # Pre-load Natural Earth data to avoid multiple, overlapping downloads.
--- a/.travis.yml
+++ b/.travis.yml
@@ -26,29 +26,35 @@ env:
 git:
   depth: 10000
 
+# Try to cache the conda installation, especially cached downloaded packages...
 cache:
   directories:
-  - $HOME/miniconda_pkgs_backup
+    - $HOME/miniconda
+
+# ...but remove the env so we make that afresh every time.
+before_cache:
+  - rm -rf $HOME/miniconda/envs/$ENV_NAME
+
+# Install conda if it is absent.
+# N.B. not sure why it needs doing before, presumably due to caching it ?
+before_install:
+  - if [ ! -f $HOME/miniconda/bin/conda ] ; then
+      echo "Fresh miniconda installation.";
+      export CONDA_BASE=https://repo.continuum.io/miniconda/Miniconda;
+      if [[ "$TRAVIS_PYTHON_VERSION" == 2* ]]; then
+        wget ${CONDA_BASE}2-latest-Linux-x86_64.sh -O miniconda.sh;
+      else
+        wget ${CONDA_BASE}3-latest-Linux-x86_64.sh -O miniconda.sh;
+      fi
+      wget $URL -O miniconda.sh;
+      rm -rf $HOME/miniconda;
+      bash miniconda.sh -b -p $HOME/miniconda;
+    fi
 
 install:
   - export IRIS_TEST_DATA_REF="2f3a6bcf25f81bd152b3d66223394074c9069a96"
   - export IRIS_TEST_DATA_SUFFIX=$(echo "${IRIS_TEST_DATA_REF}" | sed "s/^v//")
 
-  # Install miniconda
-  # -----------------
-  - export CONDA_BASE=https://repo.continuum.io/miniconda/Miniconda
-  - if [[ "$TRAVIS_PYTHON_VERSION" == 2* ]]; then
-      wget ${CONDA_BASE}2-latest-Linux-x86_64.sh -O miniconda.sh;
-    else
-      wget ${CONDA_BASE}3-latest-Linux-x86_64.sh -O miniconda.sh;
-    fi
-  - bash miniconda.sh -b -p $HOME/miniconda
-  - if [[ "$(ls $HOME/miniconda_pkgs_backup/* 2>/dev/null)" != "" ]]; then
-      echo "Existing Package backup was FOUND, copying ...";
-      cp -r $HOME/miniconda_pkgs_backup/* $HOME/miniconda/pkgs/;
-    else
-      echo "Package backup was NOT present.";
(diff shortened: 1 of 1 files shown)


ocefpaf:
@pp-mo you don't need to copy the packages and then copy them back. You can take advantage of the environment created for testing and conda's behavior by caching the whole miniconda directory and just removing the test environment before caching. That simplifies things a lot, here is an example:

```yaml
cache:
  directories:
    - $HOME/miniconda

# Removing the directory will remove the env but leave the cached packages
# at $HOME/miniconda/pkgs. That is a win-win because when re-creating the
# env we will download only the new packages.
before_cache:
  - rm -rf $HOME/miniconda/envs/$ENV_NAME

before_install:
  - |
    echo ""
    if [ ! -f $HOME/miniconda/bin/conda ] ; then
      echo "Fresh miniconda installation."
      wget $URL -O miniconda.sh
      rm -rf $HOME/miniconda
      bash miniconda.sh -b -p $HOME/miniconda
    fi
```

pp-mo:
> @ocefpaf  you don't need to copy the packages ...

Thanks for advice.  I had sort-of forgotten what a public this is to make your mistakes...
Do you have any experience of this, then ?

Unfortunately I'm very doubtful whether this will make any difference even if/when I do get it to work.
While it is clear that a large fraction of the test time is spent establishing a working environment, it hugely depends what is actually doing with that time : 
Travis docs explicitly warn that there is no benefit caching data that is merely downloaded, as it is still using a download to repopulate it.

ocefpaf:
> Travis docs explicitly warn that there is no benefit caching data that is merely downloaded, as it is still using a download to repopulate it.

You are probably referring to this:

    Large files that are quick to install but slow to download do not benefit from caching,
    as they take as long to download from the cache as from the original source:

I guess that the download of those packages from `anaconda.org` can be quite slow while, I believe, Travis-CI will be downloading the cache from their own servers. If that turns out to be the same we are at least saving the time for the actual installation of miniconda.

I did notice some improvements everywhere I implemented that. Here is a [before](https://travis-ci.org/ioos/notebooks_demos/builds/224367620) and [after](https://travis-ci.org/ioos/notebooks_demos/builds/228844386) example where I implemented that.

pp-mo:
Above @ f24c71b initially failed on 
Timings as follows:
```
--Python 2.7--
default     13:00
minimal      8:02
coding      10:52
examples    11:26
doctest     14:29
--Python 3.5--
default     16:22 FAIL
minimal      8:12
coding      11:26
examples    11:46
doctest     15:35
```

It is also then listing a cache against the PR.2633 of "2269.64MB".

pp-mo:
Newest using @ocefpaf ideas  = cache whole conda installation.

First viable version https://github.com/SciTools/iris/pull/2633/commits/b6b782e346c252ef57278b06a3061d9deefb38c0
Initial test results :
```
--Python 2.7--
default     16:57 TIMEOUT=FAIL
minimal      8:34
coding      10:16
examples    10:50
doctest     10:53
--Python 3.5--
default     17.28 TIMEOUT=FAIL
minimal      9:33
coding      11:26
examples    11:46
doctest     14.37
```

pp-mo:
Re-spun the python2.7 "default" that timed out
  `16:57 TIMEOUT=FAIL` --> `18:54 TIMEOUT=FAIL`
Re-examine the log ...
We find that the original has not created a conda cache, so the re-spin does not get one
But also **both** don't create one, apparently because it timed out writing the cache
> "running `casher push` took longer than 180 seconds and has been aborted"

However, the 'minimal' run takes about 120-130secs to write the cache.

pp-mo:
Latest https://github.com/SciTools/iris/pull/2633/commits/2bddeba5d1caa220603786461577103b54c21f56 aims to increase the cacher push timeout 
 \- from looking at https://github.com/travis-ci/casher/blob/9587798e0b7ce3609dfe472e6aa6c35ad37bf164/bin/casher#L43

pp-mo:
Result of that last is :
```
--Python 2.7--
default     15:30  TIMEOUT=FAIL
minimal      6:57
coding      12:05
examples    12:23
doctest     12:55
--Python 3.5--
default     17:00  TIMEOUT=FAIL
minimal      6:38
coding      10:38
examples    11:35
doctest     14:51


Log shows that the failed "py2.7 default" run 
has nevertheless now *succeeded* in  creating a cached conda
 \-- and the cache save took about 193 secs
Likewise, the failed "py3.5 default", and that cache-save took about 207 secs

**Comment by pp-mo** [comment]
Which of those runs successfully used a pre-existing cache ??

```
List of timing + whether it succeeded in using an exisiting installation:
--Python 2.7--
default     15:30  FAIL     no  [[but we know that no prior cache existed]]
minimal      6:57           yes
coding      12:05           no
examples    12:23           no
doctest     12:55           yes
--Python 3.5--
default     17:00  FAIL     no  [[none existed]]
minimal      6:38           yes
coding      10:38           yes
examples    11:35           yes
doctest     14:51           no
```

Interesting case : the "p2.7 doctest" times are like:
```
fetch cached python             ||   18.15
establish cache ~/miniconda     ||   46.62
install miniconda               ||    -
conda update                    ||    6.41
conda create                    ||    6.00
conda install                   ||   95.79
python setup.py build           ||    5.20
python setup.py install         ||    9.49
python -m iris.tests..          ||  232.46
"store build cache"             ||  251.80   [[ !! near timeout !! ]]
```
That is pretty slow on "conda install" and even slower to write the conda cache

**Comment by pp-mo** [comment]
Here's timings from that latest rerun (with the increased caching timeout=250sec)
```
Timings + whether it succeeded in using an exisiting installation
            TIME    ok  READ        INSTALL     SAVE
--Python 2.7--
default     13:50   NO   34.47       54.32      216.15
minimal      6:57        30.88       55.44      111.46
coding       8:05        49.77       55.64      196.87
examples    10:21         -         215.12      178.88 [[previous "took longer than 250 seconds and has been aborted"]]
doctest     14:28        60.29       95.55      279.71
--Python 3.5--
default     15:13   NO   46.57       51.08      201.72
minimal      6:55        29.01       42.29      125.23
coding      10:42        53.39       91.10      262.86
examples     8:30        46.72       46.48      184.38
doctest     12:34        47.97       52.38      214.62
```

Re-spins on the defaults...
```
p3-default   9:38        30.18       30.61      146.51
p2-default  14:xx .... (failed again)
p2-default
```

**Comment by pp-mo** [comment]
Overall timings at this point compared : "old" + "new":
```
            OLD     NEW
--Python 2.7--
default     11:40   13:50
minimal      6:21    6:57
coding       7:18    8:05
examples     7:34   10:21
doctest      7:21   14:28
--Python 3.5--
default      8:19   15:13
minimal      6:47    6:55
coding       7:15   10:42
examples     7:05    8:30
doctest     10:24   12:34
```

**Not that encouraging**
But perhaps something similar *not* involving the actual Travis caching could help ?

pp-mo:
> Perhaps something similar not involving the actual Travis caching could help ?

Status report:
If we could parachute in a pre-built conda installation from "somewhere",
then it looks to me like ...
 * we must still build an environment from the "travis.yml" each time ..
 * .. but we can potentially reduce "conda install" time from ~250 to ~50 seconds.
 * Fetching in zipped form + unzipping seems best, which is like the existing caching and will probably take another 30-40secs
 * we should avoid writing it back : with caching support, that is taking 100-250 secs ..
 * .. however, that time doesn't necessarily affect whether the tests time out

In principle, we can use a pre-built conda+packages to accelerate environment creation.
If the pre-built doesn't have all you need, the "install" step should still work, albeit maybe a bit slow.
However, it looks rather like the dependency resolution part is a significant part of the time cost, and that it may be mostly that part that gets significantly slower when Travis is busy.

The costs of pre-building like that is:
 * how you make it
 * where it is stored / hosted / served
 * how tricky it is to update or refresh it
  - i.e. make it include everything that the builds (now) need; exclude stuff no longer needed
 * how often it needs updating

PR #2682, opened 2017-07-12
np.percentile method as an alternative to scipy.mstats

bayliffe:
Introduction of a numpy.percentile method to the percentile aggregator for the purposes of providing a fast alternative (approx 50 times faster).

* Accessed with kwarg percentile_method="fast" passed to cube.collapsed method.
* Existing unit tests duplicated with call to fast method, where masked data will result in an error.
* Ran test cases and found fractional differences of order 1E-16 between two methods.
* scipy.stats.mstats.mquantiles method remains important for dealing with masked data.

- [x] added unit tests.

### DIFF
--- a/lib/iris/analysis/__init__.py
+++ b/lib/iris/analysis/__init__.py
@@ -1041,7 +1041,6 @@ def _percentile(data, axis, percent, **kwargs):
     if shape:
         data = data.reshape([np.prod(shape), data.shape[-1]])
     # Perform the percentile calculation.
-    print('PERCENTILE METHOD ', percentile_method)
     if percentile_method == 'fast':
         msg = 'Cannot use fast np.percentile method with masked array.'
         if ma.isMaskedArray(data):
--- a/lib/iris/analysis/__init__.py
+++ b/lib/iris/analysis/__init__.py
@@ -1017,7 +1017,8 @@ class WeightedAggregator(Aggregator):
         return result
 
 
-def _percentile(data, axis, percent, **kwargs):
+def _percentile(data, axis, percent, percentile_method='scipy_mquantiles',
+                **kwargs):
     """
     The percentile aggregator is an additive operation. This means that
     it *may* introduce a new dimension to the data for the statistic being
@@ -1029,11 +1030,13 @@ def _percentile(data, axis, percent, **kwargs):
     Kwargs:
 
     * percentile_method (string) :
-        Switch to opt between the scipy.stats.mstats.mquantiles method for
-        calculating percentiles, and the much faster np.percentiles method.
+        Switch to choose between methods for calculating percentiles.
+        Specify `percentile_method='fast'` to use the fast `numpy.percentiles`
+        method.
+        Do not specify or specify `percentile_method=None` to use the
+        scipy.mstats.mquantiles method.
 
     """
-    percentile_method = kwargs.get('percentile_method', None)
     # Ensure that the target axis is the last dimension.
     data = np.rollaxis(data, axis, start=data.ndim)
     shape = data.shape[:-1]
@@ -1041,15 +1044,18 @@ def _percentile(data, axis, percent, **kwargs):
     if shape:
         data = data.reshape([np.prod(shape), data.shape[-1]])
     # Perform the percentile calculation.
-    if percentile_method == 'fast':
+    if percentile_method == 'numpy_percentile':
         msg = 'Cannot use fast np.percentile method with masked array.'
         if ma.isMaskedArray(data):
             raise TypeError(msg)
         result = np.percentile(data, percent, axis=-1)
-    else:
+    elif percentile_method == 'scipy_mquantiles':
         quantiles = np.array(percent) / 100.
         result = scipy.stats.mstats.mquantiles(data, quantiles, axis=-1,
                                                **kwargs)
+    else:
+        msg = 'Unknown percentile_method {}.'
+        raise ValueError(msg.format(percentile_method))
     if not ma.isMaskedArray(data) and not ma.is_masked(result):
         result = np.asarray(result)
         if percentile_method == 'fast':
--- a/lib/iris/analysis/__init__.py
+++ b/lib/iris/analysis/__init__.py
@@ -1027,7 +1027,7 @@ def _percentile(data, axis, percent, **kwargs):
     dimension of the resulting percentile data payload.
 
     Kwargs:
-    
+
     * percentile_method (string) :
         Switch to opt between the scipy.stats.mstats.mquantiles method for
         calculating percentiles, and the much faster np.percentiles method.
--- a/lib/iris/analysis/__init__.py
+++ b/lib/iris/analysis/__init__.py
@@ -1026,18 +1026,36 @@ def _percentile(data, axis, percent, **kwargs):
     If a new additive dimension is formed, then it will always be the last
     dimension of the resulting percentile data payload.
 
+    Kwargs:
+    
+    * percentile_method (string) :
+        Switch to opt between the scipy.stats.mstats.mquantiles method for
+        calculating percentiles, and the much faster np.percentiles method.
+
     """
+    percentile_method = kwargs.get('percentile_method', None)
     # Ensure that the target axis is the last dimension.
     data = np.rollaxis(data, axis, start=data.ndim)
-    quantiles = np.array(percent) / 100.
     shape = data.shape[:-1]
     # Flatten any leading dimensions.
     if shape:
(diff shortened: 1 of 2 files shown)


lib/iris/analysis/__init__.py
```diff
@@ -1026,18 +1026,35 @@ def _percentile(data, axis, percent, **kwargs):
     If a new additive dimension is formed, then it will always be the last
     dimension of the resulting percentile data payload.
 
+    Kwargs:
+
+    * percentile_method (string) :
```
marqh: by providing a string input, you are enabling a further enhancement opportunity, specifying another implementation; which is no bad thing, this is just an observation.

Given the string input, I would like to see an enhanced docstring, describing the choice of inputs and their effect. for example

```
Switch to choose between methods for calculating percentiles.
Specify: `percentile_method='fast'` to use the fast `numpy.percentiles` method.
Do not specify or specify `percentile_method=None` to use the scipy.mstats.mquantiles method. 
```

lib/iris/analysis/__init__.py
```diff
@@ -1026,18 +1026,35 @@ def _percentile(data, axis, percent, **kwargs):
...
+        Switch to opt between the scipy.stats.mstats.mquantiles method for
+        calculating percentiles, and the much faster np.percentiles method.
+
     """
+    percentile_method = kwargs.get('percentile_method', None)
```
marqh: please may you consider specifying this in the function definition, i.e.:
```
def _percentile(data, axis, percent, percentile_method=None, **kwargs):
```

lib/iris/analysis/__init__.py
```diff
@@ -1026,18 +1026,35 @@ def _percentile(data, axis, percent, **kwargs):
...
     # Flatten any leading dimensions.
     if shape:
         data = data.reshape([np.prod(shape), data.shape[-1]])
     # Perform the percentile calculation.
-    result = scipy.stats.mstats.mquantiles(data, quantiles, axis=-1, **kwargs)
+    if percentile_method == 'fast':
+        msg = 'Cannot use fast np.percentile method with masked array.'
+        if ma.isMaskedArray(data):
+            raise TypeError(msg)
+        result = np.percentile(data, percent, axis=-1)
```
marqh: does `np.percentile` take the same `**kwargs` as `scipy.stats.mstats.mquantiles`?

if it is, then I think the kwargs should be passed down

if not, then some thought about how to handle kwargs differently for different implementations should be considered

lib/iris/tests/test_analysis.py
```diff
@@ -365,6 +365,26 @@ def test_percentile_1d(self):
...
                        checksum=False)
 
+    def test_fast_percentile_1d(self):
+        cube = tests.stock.simple_1d()
+
+        first_quartile = cube.collapsed('foo', iris.analysis.PERCENTILE,
+                                        percent=25, percentile_method='fast')
+        np.testing.assert_array_almost_equal(first_quartile.data,
+                                             np.array([2.5], dtype=np.float32))
+        self.assertCML(first_quartile, ('analysis',
```
marqh: here, and in other places, I would rather not run `self.assertCML` in these cases, it is an expensive and sometimes unstable test pattern

I would also prefer to see one assert per test case.  T=in this case, that would mean having 2 test functions, one for first_quartile and one for third_quartile

pp-mo:
@bayliffe 
Some of those test failures seem decidedly odd + I doubt they are "your fault".
Possibly something has changed in the Travis environment, like a new version of a package becoming available.
This happens from time to time -- I'll take a look...

lib/iris/analysis/__init__.py
```diff
@@ -1026,18 +1026,35 @@ def _percentile(data, axis, percent, **kwargs):
...
+    else:
+        quantiles = np.array(percent) / 100.
+        result = scipy.stats.mstats.mquantiles(data, quantiles, axis=-1,
+                                               **kwargs)
```
pp-mo: As **kwargs is passed here for extended args, you should really remove any 'percentile_method' key as it could otherwise cause errors.
BTW this is also a usage mode you don't test : `cube.collapsed(..., percent=x, percentile_method=None)` : 
As stating `percentile_method=None` is not the same as not providing one, this could cause problems with other routines calling it.
The usual way I've seen elsewhere is to use `kwargs.pop` in place of `kwargs.get`.

In fact, for choice, I'd prefer distinct allowed terms here, like "numpy_percentile" and "scipy_mquantiles", and not just "None".
I guess it would be neater if it was just a method object, but as they need handling differently, I'm happy that a string modal switch makes better sense.

lib/iris/tests/test_analysis.py
pp-mo: This is asking @bayliffe to improve on the existing code.
Which I'm not actually against, BTW :wink: 

Ideally, I'd prefer to see a common utility routine instead of all this duplicated code...
I wouldn't insist on it, but it would make these improvements simpler + more obvious.

pp-mo: > common utility routine

By which I mean, something like `self._check_percentile(cube, percents, expected_result)`

pp-mo: Also, I just noticed that nothing here is testing for multiple values passed to the percentile arg.

Actually TBH these are very old tests, as are any not under `iris/tests/unit` or `iris/tests/integration`.
I don't have time jst now to research a more definite suggestion, but I would definitely be thinking of adding new tests into `iris.tests.unit.analysis.test_PERCENTILE` or `iris.tests.unit.analysis.test_PercentileAggregator`.
There is testing there for multiple values, too.

Sorry to have been so slow seeing the obvious here !

lib/iris/analysis/__init__.py
pp-mo: > I think the kwargs should be passed down

:+1:
That makes good sense + from the docs you'd expect it to do that in both cases.

>  how to handle kwargs differently for different implementations

:-1:
I disagree : I don't think that's at all necessary or helpful **unless** your code is specifically using extra args by testing them, adjusting them or automatically generating them.
In my head, if it is clear that `percentile_method` switches between two underlying methods, and that is an explicit user choice, then any extra args are supplied by the same user code, so all the necessary information and responsibility is in the user's hands.  *This* routine knows nothing about those extra args, and doesn't need to, and certainly should not copy documentation relating to the underlying methods (which could change anyway).

lib/iris/analysis/__init__.py
pp-mo: :+1: 
Actually yes, that is a better way.

lib/iris/analysis/__init__.py
bayliffe: Afternoon. My affiliation is Met Office.
I have based the code on v1.13 as that is the version we are currently using. It is up-to-date with that branch, so a rebase does nothing. I can move it to v2 if that is desirable?

I have changed the comment and moved the keyword argument into the interface as suggested.

lib/iris/analysis/__init__.py
bayliffe: Done.

lib/iris/analysis/__init__.py
bayliffe: The np.percentile method does not except the same kwargs, hence their omission. I will leave this as is for now, pending any further suggestions.

ajdawson:
@bayliffe (cc @marqh) The merge target here is v1.13.x, which is a bugfix-only branch closed for new features. This contribution should be made to the master branch instead.

lib/iris/tests/test_analysis.py
bayliffe: I have rewritten the unit tests across the section of interest. I have (possibly unwisely) removed the contentious assertCML tests. Instead I have added a simple additional coordinate check on the resulting cubes as something approaching a replacement, though obviously not as rigorous and it could be expanded. I have moved the assert statements into functions and separated out the cases where the first and third percentiles were being checked in a single test.

bayliffe: This may not be what you were after, but it is a bit tidier. Feel free to suggest further changes, or the return of the CML statements as you prefer.

lib/iris/analysis/__init__.py
bayliffe: Changed to accept `numpy_percentile` or `scipy_mquantiles`, with the latter set as the keyword default. A ValueError is raised if an invalid method is set. I have added a unit test for this eventuality.

bayliffe:
Thanks @ajdawson, I will change this to Iris V2.

bayliffe:
Replaced with PR #2687 for implementation in Iris V2.

PR #2683, opened 2017-07-12
Add link to latest-docs build.

pp-mo:
Because it's useful, and I keep forgetting how to find it.


lbdreyer:
Adding a badge might be nice too

lbdreyer:
Update: I've put a pull request to do add the badge in case you think that is a good idea (see #2684)

PR #2684, opened 2017-07-12
Add docs badge to README

lbdreyer:
This adds a badge to the README that will take you to the latest build of the documentation for the master branch.


pp-mo:
I didn't think of that.
Does you think maybe the badge is good enough on its own ?

lbdreyer:
Thanks for merging @pp-mo 

I does think that it is fine to have both. We don't want our README to get too big, obviously, but it's fine for now. We can always take out one of the links at a later stage

PR #2699, opened 2017-07-31
Integrate dask masked array support (just code changes, no tests fix)

lbdreyer:
This is pointing at the `dask_mask_array` feature branch.

This contains just the code changes that I think require to be done to integrate the dask masked array support into Iris.

This will obviously break a bunch of tests but we can address that later.

### DIFF
--- a/lib/iris/_lazy_data.py
+++ b/lib/iris/_lazy_data.py
@@ -94,13 +94,12 @@ def as_lazy_data(data, chunks=_MAX_CHUNK_SIZE):
 
     """
     if not is_lazy_data(data):
-        if ma.isMaskedArray(data):
-            data = array_masked_to_nans(data)
-        data = da.from_array(data, chunks=chunks)
+        asarray = not ma.isMaskedArray(data)
+        data = da.from_array(data, chunks=chunks, asarray=asarray)
     return data
 
 
-def as_concrete_data(data, **kwargs):
+def as_concrete_data(data):
     """
     Return the actual content of a lazy array, as a numpy array.
     If the input data is a NumPy `ndarray` or masked array, return it
@@ -108,16 +107,10 @@ def as_concrete_data(data, **kwargs):
 
     If the input data is lazy, return the realised result.
 
-    Where lazy data contains NaNs these are translated by filling or converting
-    to masked data, using the :func:`~iris._lazy_data.convert_nans_array`
-    function.
-
     Args:
 
     * data:
         A dask array, NumPy `ndarray` or masked array
-
-    Kwargs are passed through to :func:`~iris._lazy_data.convert_nans_array`.
 
     Returns:
         A NumPy `ndarray` or masked array.
@@ -129,51 +122,8 @@ def as_concrete_data(data, **kwargs):
         # rather than a numpy.ndarray object.
         # Recorded in https://github.com/dask/dask/issues/2111.
         data = np.asanyarray(data.compute())
-        # Convert any missing data as requested.
-        data = convert_nans_array(data, **kwargs)
 
     return data
-
-
-def nan_array_type(dtype):
-    return np.dtype('f8') if dtype.kind in 'biu' else dtype
-
-
-def array_masked_to_nans(array):
-    """
-    Convert a masked array to a NumPy `ndarray` filled with NaN values. Input
-    NumPy arrays with no mask are returned unchanged.
-    This is used for dask integration, as dask does not support masked arrays.
-
-    Args:
-
-    * array:
-        A NumPy `ndarray` or masked array.
-
-    Returns:
-        A NumPy `ndarray`. This is the input array if unmasked, or an array
-        of floating-point values with NaN values where the mask was `True` if
-        the input array is masked.
-
-    .. note::
-        The fill value and mask of the input masked array will be lost.
-
-    .. note::
-        Integer masked arrays are cast to 8-byte floats because NaN is a
-        floating-point value.
-
-    """
-    if not ma.isMaskedArray(array):
-        result = array
-    else:
-        if ma.is_masked(array):
-            mask = array.mask
-            new_dtype = nan_array_type(array.data.dtype)
-            result = array.data.astype(new_dtype)
-            result[mask] = np.nan
-        else:
-            result = array.data
-    return result
 
 
 def multidim_lazy_stack(stack):
@@ -202,70 +152,3 @@ def multidim_lazy_stack(stack):
         result = da.stack([multidim_lazy_stack(subarray)
                            for subarray in stack])
     return result
-
-
-def convert_nans_array(array, nans_replacement=None, result_dtype=None):
-    """
-    Convert a :class:`~numpy.ndarray` that may contain one or more NaN values
-    to either a :class:`~numpy.ma.core.MaskedArray` or a
-    :class:`~numpy.ndarray` with the NaN values filled.
-
-    Args:
-
-    * array:
-        The :class:`~numpy.ndarray` to be converted.
-
-    Kwargs:
-
-    * nans_replacement:
-        If `nans_replacement` is None, then raise an exception if the `array`
-        contains any NaN values (default behaviour).
-        If `nans_replacement` is `numpy.ma.masked`, then convert the `array`
-        to a :class:`~numpy.ma.core.MaskedArray`.
-        Otherwise, use the specified `nans_replacement` value as the `array`
-        fill value.
-
-    * result_dtype:
-        Cast the resultant array to this target :class:`~numpy.dtype`.
-
-    Returns:
-        An :class:`numpy.ndarray`.
-
-    .. note::
-        An input array that is either a :class:`~numpy.ma.core.MaskedArray`
-        or has an integral dtype will be returned unaltered.
-
-    .. note::
-        In some cases, the input array is modified in-place.
-
-    """
-    if not ma.isMaskedArray(array) and array.dtype.kind == 'f':
-        # First, calculate the mask.
-        mask = np.isnan(array)
-        # Now, cast the dtype, if required.
-        if result_dtype is not None:
-            result_dtype = np.dtype(result_dtype)
-            if array.dtype != result_dtype:
-                array = array.astype(result_dtype)
-        # Finally, mask or fill the data, as required or raise an exception
-        # if we detect there are NaNs present and we didn't expect any.
-        if np.any(mask):
-            if nans_replacement is None:
-                emsg = 'Array contains unexpected NaNs.'
-                raise ValueError(emsg)
-            elif nans_replacement is ma.masked:
-                # Mask the array with the default fill_value.
-                array = ma.masked_array(array, mask=mask)
-            else:
-                # Check the fill value is appropriate for the
-                # result array dtype.
-                try:
-                    [fill_value] = np.asarray([nans_replacement],
-                                              dtype=array.dtype)
-                except OverflowError:
-                    emsg = 'Fill value of {!r} invalid for array result {!r}.'
-                    raise ValueError(emsg.format(nans_replacement,
-                                                 array.dtype))
-                # Fill the array.
-                array[mask] = fill_value
-    return array
--- a/lib/iris/_merge.py
+++ b/lib/iris/_merge.py
@@ -315,7 +315,7 @@ class _CoordSignature(namedtuple('CoordSignature',
 
 class _CubeSignature(namedtuple('CubeSignature',
                                 ['defn', 'data_shape', 'data_type',
-                                 'fill_value', 'cell_measures_and_dims'])):
+                                 'cell_measures_and_dims'])):
     """
     Criterion for identifying a specific type of :class:`iris.cube.Cube`
     based on its metadata.
@@ -330,9 +330,6 @@ class _CubeSignature(namedtuple('CubeSignature',
 
     * data_type:
         The data payload :class:`numpy.dtype` of a :class:`iris.cube.Cube`.
-
-    * fill_value:
-        The fill-value for the data payload of a :class:`iris.cube.Cube`.
 
     * cell_measures_and_dims:
         A list of cell_measures and dims for the cube.
@@ -1241,9 +1238,7 @@ class ProtoCube(object):
                 # All inputs were concrete, so turn the result back into a
                 # normal array.
                 dtype = self._cube_signature.data_type
-                merged_data = as_concrete_data(merged_data,
-                                               nans_replacement=ma.masked,
-                                               result_dtype=dtype)
+                merged_data = as_concrete_data(merged_data)
                 # Unmask the array if it has no masked points.
                 if (ma.isMaskedArray(merged_data) and
                         not ma.is_masked(merged_data)):
@@ -1283,14 +1278,6 @@ class ProtoCube(object):
         other = self._build_signature(cube)
         match = cube_signature.match(other, error_on_mismatch)
         if match:
-            # Determine whether the fill value requires to be demoted
-            # to the default value.
-            if cube_signature.fill_value is not None:
-                if cube_signature.fill_value != other.fill_value:
-                    # Demote the fill value to the default.
-                    signature = self._build_signature(self._source,
-                                                      default_fill_value=True)
-                    self._cube_signature = signature
             coord_payload = self._extract_coord_payload(cube)
             match = coord_payload.match_signature(self._coord_signature,
                                                   error_on_mismatch)
@@ -1504,8 +1491,6 @@ class ProtoCube(object):
                               dim_coords_and_dims=dim_coords_and_dims,
                               aux_coords_and_dims=aux_coords_and_dims,
                               cell_measures_and_dims=cms_and_dims,
-                              fill_value=signature.fill_value,
-                              dtype=signature.data_type,
                               **kwargs)
 
         # Add on any aux coord factories.
@@ -1611,7 +1596,7 @@ class ProtoCube(object):
                               self._vector_aux_coords_dims):
             aux_coords_and_dims.append(_CoordAndDims(item.coord, dims))
 
-    def _build_signature(self, cube, default_fill_value=False):
+    def _build_signature(self, cube):
         """
         Generate the signature that defines this cube.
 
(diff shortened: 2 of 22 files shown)


lbdreyer:
I'm pretty happy with the handling of fill value:

```
>>> x = ma.masked_array([1,2,3,4], mask=[0,1,0,0], fill_value=10)
>>> dx = da.from_array(x, chunks=(2,), asarray=False)
>>> cdx = dx.compute()
>>> print cdx.fill_value
10
```

Combining two daskified masked arrays was interesting though:
```
>>> x = ma.masked_array([1,2,3,4], mask=[0,1,0,0], fill_value=15)
>>> y = ma.masked_array([1,2,3,4], mask=[0,1,0,0], fill_value=32, dtype=np.float64)
>>> dx = da.from_array(x, chunks=(2,), asarray=False)
>>> dy = da.from_array(y, chunks=(2,), asarray=False)
>>> dxy = dx * dy
>>> cdxy = dxy.compute()
>>> print cdxy.fill_value
15.0
```

lib/iris/_lazy_data.py
```diff
@@ -94,31 +94,26 @@ def as_lazy_data(data, chunks=_MAX_CHUNK_SIZE):
 
     """
     if not is_lazy_data(data):
+        asarray = True
         if ma.isMaskedArray(data):
-            data = array_masked_to_nans(data)
-        data = da.from_array(data, chunks=chunks)
+            asarray = False
+        data = da.from_array(data, chunks=chunks, asarray=asarray)
```
bjlittle: @lbdreyer This change could be simply reduced to:
```python
asarray = not ma.isMaskedArray(data)
data = da.from_array(data, chunks=chunks, asarray=asarray)
```

bjlittle:
@lbdreyer Just noticed that we should be purging the use of the following `_lazy_data` functions:
- [ ] `nan_array_type`
    - [ ] `fileformats/netcdf.py` - imported but not used
    - [ ] `_lazy_data.py` - definition
- [ ] `array_masked_to_nans`
    - [ ] `fileformats/grib/message.py` - imported but not used
    - [ ] `fileformats/netcdf.py` - I don't think that we should be using this here
    - [ ] `fileformats/pp.py` - imported but not used
    - [ ] `_lazy_data.py` - definition
- [ ] `convert_nans_array`
    - [ ] `fileformats/grib/__init__.py` - revert usage back to 1.13 ?
    - [ ] `fileformats/netcdf.py` - imported by not used
    - [ ] `tests/unit/lazy_data/test_as_concrete_data.py` - we can nuke these tests
    - [ ] `tests/unit/lazy_data/test_convert_nans_array.py` - we can remove all these tests
    - [ ] `_lazy_data.py` - definition

bjlittle:
@lbdreyer We should remove the `DataManager.replace` method and associated `Cube.replace` method, and retrospectively use `datamanager.data = data` and `cube.data = data` instead ...
- [ ] `analysis/__init__.py`, line 566
- [ ] `cube.py`, lines 2912-2922
- [ ] `tests/test_analysis.py`, lines 132, 138, 147
- [ ] `tests/test_basic_maths.py`, line 42
- [ ] `tests/test_cdm.py`, line 1033, line 1039
- [ ] `tests/test_concatenate.py`, line 521, line 542
- [ ] `tests/test_interpolation.py`, line 46
- [ ] `tests/unit/analysis/regrid/test_RectilinearRegridder.py`, line 1266
- [ ] `tests/unit/cube/test_Cube.py`, assorted lines
- [ ] `tests/unit/fileformats/pp/test_as_pairs.py`, line 54
- [ ] `tests/unit/data_manager/test_DataManager.py`, assorted lines
- [ ] `_data_manager.py`, line 193, line 374

lib/iris/_merge.py
```diff
@@ -1620,22 +1605,14 @@ def _build_signature(self, cube, default_fill_value=False):
...
 
         """
-        fill_value = cube.fill_value
-        if default_fill_value:
-            fill_value = None
-        return _CubeSignature(cube.metadata, cube.shape, cube.dtype,
-                              fill_value, cube._cell_measures_and_dims)
+
+        return _CubeSignature(cube.metadata, cube.shape,
+                              cube.lazy_data().dtype,
```
bjlittle: @lbdreyer This should just be `cube.dtype` right? ...

lib/iris/coords.py
```diff
@@ -1866,12 +1866,7 @@ def data(self, data):
             raise ValueError('The data payload of a CellMeasure may not be '
                              'None; it must be a numpy array or equivalent.')
         if is_lazy_data(data) and data.dtype.kind in 'biu':
-            # Disallow lazy integral data, as it will cause problems with dask
-            # if it turns out to contain any masked points.
             # Non-floating cell measures are not valid up to CF v1.7 anyway,
```
bjlittle: @lbdreyer Minor, but you can remove the ` anyway,` from the end of this comment.

lib/iris/cube.py
```diff
@@ -1587,28 +1577,7 @@ def dtype(self):
...
-            Thus, many operations that change cube data will also set
-            ``fill_value`` to ``None``.
-
-        """
-        return self._data_manager.fill_value
-
-    @fill_value.setter
-    def fill_value(self, fill_value):
-        self._data_manager.fill_value = fill_value
+        return self._data_manager.core_data().dtype
```
bjlittle: @lbdreyer Isn't this just `self._data_manager.dtype` still? Why the need to go through the `core_data` function?

lib/iris/cube.py
```diff
@@ -1587,28 +1577,7 @@ def dtype(self):
...
-            Ideally, this value will not occur anywhere in the cube data.
-            Thus, many operations that change cube data will also set
-            ``fill_value`` to ``None``.
-
-        """
-        return self._data_manager.fill_value
-
-    @fill_value.setter
-    def fill_value(self, fill_value):
-        self._data_manager.fill_value = fill_value
```
bjlittle: @lbdreyer Yup, nice.

There's no point in even having a `fill_value` getter convenience, as it's not possible to get a lazy `fill_value` or indeed track whether the lazy array is an ndarray or masked without realising it (please tell me otherwise).

lib/iris/cube.py
```diff
@@ -2821,8 +2785,6 @@ def _xml_element(self, doc, checksum=False, order=True, byteorder=True):
         if self.var_name:
             cube_xml_element.setAttribute('var_name', self.var_name)
         cube_xml_element.setAttribute('units', str(self.units))
-        if self.fill_value is not None:
-            cube_xml_element.setAttribute('fill_value', str(self.fill_value))
```
bjlittle: Oh boy! CML carnage 😱

lib/iris/cube.py
lbdreyer: Addressed in the most recent commit

lib/iris/coords.py
lbdreyer: Addressed in the most recent commit

lib/iris/_merge.py
lbdreyer: Addressed in the most recent commit

bjlittle:
@lbdreyer Just missed one case of `replace` on `line 194` in `_data_manager.py`

lib/iris/cube.py
```diff
@@ -2939,40 +2901,15 @@ def _order(array):
             if order and ma.isMaskedArray(data):
                 data_xml_element.setAttribute('mask_order',
                                               _order(data.mask))
-        else:
-            dtype = self.lazy_data().dtype
-        data_xml_element.setAttribute('dtype', dtype.name)
+            else:
+                dtype = self.lazy_data().dtype
+            data_xml_element.setAttribute('dtype', dtype.name)
```
bjlittle: @lbdreyer I think you need to back this tab change out. 

Referring back to v1.13 the `else` on `line 2904` aligns with the `if` on `line2879`, and **not** the `if` on `line 2901`.

The previous logic makes sense, in that the `dtype` xml attribute is always set, and for non-lazy data extra xml metadata is set for the concrete data payload.

lib/iris/etc/pp_save_rules.txt
```diff
@@ -761,14 +761,14 @@ THEN
 
 #MDI
 IF
-    cm.fill_value is not None
+    isinstance(cm.data, ma.core.MaskedArray)
```
bjlittle: @lbdreyer Replace this with `ma.isMaskedArray(cm.data)` ? But this forces the loading of the data.

Slightly confusing is that for both of these PP save rule, in v1.13, there should be no change here. I don't quite understand that (at the moment) as there is no `fill_value` attribute on a `cm` (cube) ... so how did these save rule work in v1.13 ...

lib/iris/etc/pp_save_rules.txt
```diff
@@ -761,14 +761,14 @@ THEN
...
 THEN
-    pp.bmdi = cm.fill_value
+    pp.bmdi = cm.data.fill_value
 
 IF
-    cm.fill_value is None
+    not isinstance(cm.data, ma.core.MaskedArray)
 THEN
-    pp.bmdi = -1e30
+pp.bmdi = -1e30
```
bjlittle: You've lost the spacing indent ...

lib/iris/fileformats/grib/__init__.py
```diff
@@ -178,14 +177,11 @@ def __init__(self, grib_message, grib_fh=None):
             # The byte offset requires to be reset back to the first byte
             # of this message. The file pointer offset is always at the end
             # of the current message due to the grib-api reading the message.
-            proxy = GribDataProxy(shape, self.realised_dtype, grib_fh.name,
+            proxy = GribDataProxy(shape, np.array([0.]).dtype, grib_fh.name,
                                   offset - message_length)
             self._data = as_lazy_data(proxy)
         else:
             values_array = _message_values(grib_message, shape)
```
bjlittle: @lbdreyer Nothing is assigned back to `self.data` ?

lib/iris/fileformats/grib/__init__.py
```diff
@@ -692,6 +688,10 @@ def _message_values(grib_message, shape):
     data = gribapi.grib_get_double_array(grib_message, 'values')
     data = data.reshape(shape)
 
+    # Handle missing values in a sensible way.
+    mask = np.isnan(data)
+    if mask.any():
+        data = ma.array(data, mask=mask, fill_value=np.nan)
```
bjlittle: @lbdreyer Just confirming my understanding ... so this is change is making up for the removal of `convert_nans_array` to convert an array containing nans to a masked array, got it ... but you're explicitly setting the `fill_value` to be nan, which we didn't ever do before. Why so?

Also, this function `_message_values` is called from the `GribWrapper.__init__`, but also the `GribDataProxy.__getitem__` ... which is interesting as the GribDataProxy didn't ever then apply `convert_nans_array` ...

lib/iris/fileformats/grib/_save_rules.py
```diff
@@ -1160,17 +1160,20 @@ def product_definition_section(cube, grib):
 def data_section(cube, grib):
     # Masked data?
     if ma.isMaskedArray(cube.data):
-        fill_value = cube.fill_value
-        if fill_value is None or np.isnan(cube.fill_value):
+        if not np.isnan(cube.data.fill_value):
```
bjlittle: @lbdreyer Okay ... I think I see the connective tissue now. So is this the reason for setting the `fill_value` to nan in `_message_values` when a GRIB message has nan values in it's message data payload in order to control auto-selecting an appropriate `fill_value`?

lib/iris/fileformats/grib/message.py
```diff
@@ -272,7 +266,8 @@ def __getitem__(self, keys):
 
     def __repr__(self):
         msg = '<{self.__class__.__name__} shape={self.shape} ' \
-              'dtype={self.dtype!r} recreate_raw={self.recreate_raw!r} '
+              'dtype={self.dtype!r} fill_value={self.fill_value!r} ' \
```
bjlittle: @lbdreyer There is no `fill_value` ...

lib/iris/fileformats/grib/message.py
```diff
@@ -258,10 +253,9 @@ def __getitem__(self, keys):
                 # Only the non-masked values are included in codedValues.
                 _data = np.empty(shape=bitmap.shape)
                 _data[bitmap.astype(bool)] = data
-                # Use nan where input = 1, the opposite of the behaviour
-                # specified by the GRIB spec.
-                _data[np.logical_not(bitmap.astype(bool))] = np.nan
-                data = _data
+                # `ma.masked_array` masks where input = 1, the opposite of
+                # the behaviour specified by the GRIB spec.
+                data = ma.masked_array(_data, mask=np.logical_not(bitmap))
```
bjlittle: @lbdreyer So the deal is that `_DataProxy` will return a data payload that is masked, right? Rather than an array that is nan filled ... that might have testing fall-out, I guess.

That aside, shouldn't this be ...
```python
data = ma.masked_array(_data, mask=np.logical_not(bitmap.astype(bool)), fill_value=np.nan)
```
This would align with the strategy for auto-selecting the `fill_value` on save, right?

... Only problem is that using `fill_value=np.nan` means that the `dtype` of all GRIB data must be `float`, otherwise it's not possible to set `np.nan` as a `fill_value` for non-`float` data ...

lib/iris/fileformats/netcdf.py
```diff
@@ -510,11 +502,10 @@ def _load_cube(engine, cf, cf_var, filename):
     # Create cube with deferred data, but no metadata
     fill_value = getattr(cf_var.cf_data, '_FillValue', None)
 
-    dtype = nan_array_type(dummy_data.dtype)
-    proxy = NetCDFDataProxy(cf_var.shape, dtype,
+    proxy = NetCDFDataProxy(cf_var.shape, dummy_data.dtype,
```
bjlittle: @lbdreyer Calculation of the `fill_value` on `line 503` has the default value of `None`.

In v1.12 and v1.13 it is `netCDF4.default_fillvals[cf_var.dtype.str[1:]]` ... I think that this default behaviour should be reinstated.

lbdreyer:
We have agreed to keep this PR to just code changes so I am going to back out all the test fixes.

I am hoping to keep this PR to just code changes, and then when it gets merged, squash it to a single commit, so in the commit history there will be a single commit with the code changes and then subsequent commits with fixes to the tests that fall out from the code changes.

lib/iris/fileformats/netcdf.py
```diff
@@ -1958,7 +1949,11 @@ def set_packing_ncattrs(cfvar):
             # Explicitly assign the fill_value, which will be the type default
             # in the case of an unmasked array.
             if packing is None:
-                fill_value = cube.fill_value
+                if hasattr(cube.lazy_data(), 'fill_value'):
+                    fill_value = cube.lazy_data().fill_value
+                else:
+                    fill_value = None
+
```
bjlittle: @lbdreyer We can't get the `fill_value` if the data is lazy, so that changes the logic here some what ...

For lines 1951-1957, I propose the following:
```python
if packing is None:
    fill_value = None
    if not cube.has_lazy_data() and ma.isMaskedArray(cube.data):
        fill_value = cube.data.fill_value
    dtype = cube.dtype.newbyteorder('=')
```

lib/iris/fileformats/netcdf.py
```diff
@@ -1967,12 +1962,8 @@ def set_packing_ncattrs(cfvar):
...
             set_packing_ncattrs(cf_var)
 
-            # Now stream the cube data payload straight to the netCDF
-            # data variable within the netCDF file, where any NaN values
-            # are replaced with the specified cube fill_value.
-            data = da.map_blocks(convert_nans_array, cube.lazy_data(),
-                                 nans_replacement=cube.fill_value,
-                                 result_dtype=cube.dtype)
+            data = cube.lazy_data()
+
```
bjlittle: @lbdreyer We could tidy this now into a simple one liner (if you agree):
```python
da.store([cube.lazy_data()], [cf_var])
```

bjlittle: @lbdreyer We could tidy this now into a simple one liner (if you agree):
```python
da.store([cube.lazy_data()], [cf_var])
```

lib/iris/etc/pp_save_rules.txt
lbdreyer: It looks like @pp-mo added a `cube.fill_value` in #2452 

So this change will undo the benefit we got from that PR

bjlittle: @lbdreyer I believe that PR pretty much got removed as a result of the merge back of the original dask feature branch into master

lib/iris/fileformats/netcdf.py
lbdreyer: Agreed, a couple of tests were complaining about this

lib/iris/fileformats/netcdf.py
lbdreyer: Iris 1.13 has
```
            if packing is None:
                fill_value = cube.lazy_data().fill_value
                dtype = cube.lazy_data().dtype.newbyteorder('=')
```
Does that mean we could get the `fill_value` from cube.lazy_data() when we were still using biggus?

lbdreyer: I am concerned that in the example where we have lazy masked data, we aren't setting the fill_value as it would just be None. I am not sure what the fall out of this would be

lib/iris/fileformats/netcdf.py
bjlittle: Hoo-rah for test coverage. Win!

lib/iris/fileformats/netcdf.py
bjlittle: Yeah, that's behaviour that we just can't avoid, and it's concerning. If we're saving lazy masked data, then the fill-value is lost. Not good.

We could do a work around though ... we know that we can only get the fill-value from concrete or non-lazy masked data, so in the lazy masked case, we could slice the lazy data down to 1 data item, then compute it, then get the fill-value. That works, and it's similar to a trick we've used in netcdf to calculate the derived dtype from data that requires a scale + offset calculation.

As you mentioned, we could also apply this approach to PP to ensure that we don't lose the goodness of @pp-mo's #2452.

@lbdreyer How does that sound?

bjlittle: @lbdreyer Actually, we need to be slightly cautious how we do this ...
```python
>>> m = ma.masked_array([0,1,2,3], mask=[0,0,0,1], fill_value=123)
>>> dm = da.from_array(m, chunks=(1,), asarray=False)
>>> dm[0].compute()
0
>>> dm[-1].compute()
masked
>>> dm[:1].compute()
masked_array(data = [0],
             mask = [False],
       fill_value = 123)
```
The last example of slicing with `[:1]` gives us back a masked array with the fill-value, which is what we need

lbdreyer: See issues raised #2703 and #2704

lib/iris/etc/pp_save_rules.txt
lbdreyer: See issues raise #2704 to address this

lib/iris/etc/pp_save_rules.txt
```diff
@@ -761,14 +761,14 @@ THEN
 
 #MDI
 IF
-    isinstance(cm.data, ma.core.MaskedArray)
+    ma.isMasked(cm.data)
```
lbdreyer: Argh this should be `isMaskedArray` !
Hang on putting up a new commit

lbdreyer:
@bjlittle  I have now addressed all your comments (except for the fill_value at saving to pp and netcdf time which we have #2703 and #2704 to address)

bjlittle:
@lbdreyer Okay, I'll take a final look through the updates ...

PR #2702, opened 2017-08-01
Dask test api fixes + cml changes

lbdreyer:
Once #2699 is in, this is next on the list.

This PR adds two extra commits. One addresses the tests that now have to be removed or changed since the API has changed (e.g. replace() method has been removed etc) 
The second commit addresses the change of some of the cml to remove things like fill_value


lbdreyer:
I think this is going to confuse things so I will wait until #2699 is up before I put these changes up

Issue #2706, opened 2017-08-02
Remove usages of convert_nans_array, array_masked_to_nans and nans_array_type


lbdreyer:
`iris._lazy_data.convert_nans_array`, `iris._lazy_data.array_masked_to_nans` and `iris._lazy_data.nans_array_type` were removed in #2699 so we need to address the test fall out.

The bits that need removing are outlined in [this comment](https://github.com/SciTools/iris/pull/2699#issuecomment-319345179)

PR #2708, opened 2017-08-02
Remove usages of replace() method

### LINKED ISSUE #2705: Remove cube.replace() or data_manager.replace() from tests
As part of #2699 we removed the `replace` method on 
`iris.cube.Cube` and `iris._data_manager.DataManager` 

We need to address the test fall-out due to this change (largely outlined in [this comment](https://github.com/SciTools/iris/pull/2699#issuecomment-319369088))

lbdreyer:
Fixes #2705


lib/iris/tests/test_basic_maths.py
```diff
@@ -39,8 +39,7 @@
 class TestBasicMaths(tests.IrisTest):
     def setUp(self):
         self.cube = iris.tests.stock.global_pp()
-        self.cube.replace(self.cube.data - 260,
-                          fill_value=self.cube.fill_value)
+        self.cube.data = self.cube.data - 260)
```
lbdreyer: I'll remove this bracket!

lib/iris/tests/unit/analysis/regrid/test_RectilinearRegridder.py
```diff
@@ -1263,7 +1263,7 @@ def test_circular_src__masked_missingmask(self):
         # instead of being an array.
         src = self.src
         src.coord('longitude').circular = True
-        src.replace(ma.MaskedArray(src.data), fill_value=src.fill_value)
+        src.data  = ma.MaskedArray(src.data, fill_value=src.fill_value)
```
bjlittle: @lbdreyer Are there two spaces in `src.data  =` ?

lbdreyer: Sorry, now fixed.

bjlittle:
@lbdreyer All looks good. For #2706 I'm trying to retain some of the test tests, if I can, and they make sense.

I'm assuming that all the `replace` tests can just be purged?

lbdreyer:
> I'm assuming that all the replace tests can just be purged?

Which tests are you referring to?

bjlittle:
> Which tests are you referring to?

These https://github.com/SciTools/iris/pull/2708/files#diff-1a002eee61524000a58a8e337597c7d7L1262 ...

lbdreyer:
The `replace` methods on `iris.cube.Cube` and `iris._data_manager.DataManager`, which those tests were testing, no longer exists so we should just delete them

bjlittle:
One thing we did do was increase test coverage thru replacing biggus with dask.

For me `replace` is kinda equivalent to `cube.data = data` ... so I was just wondering whether there are any of those tests mentioned above, that could be re-purposed as such `cube.data = data` tests and live else where in another appropriate test class ... the answer is more likely not, but just thought I'd ask.

lbdreyer:
@bjlittle 
I now understand your point. I think it would be overkill to keep some cube.replace tests but possibly reworking the data_manager.replace tests might make sense

lib/iris/tests/unit/analysis/regrid/test_RectilinearRegridder.py
```diff
@@ -1263,7 +1263,7 @@ def test_circular_src__masked_missingmask(self):
         # instead of being an array.
         src = self.src
         src.coord('longitude').circular = True
-        src.replace(ma.MaskedArray(src.data), fill_value=src.fill_value)
+        src.data = ma.MaskedArray(src.data, fill_value=src.fill_value)
```
bjlittle: @lbdreyer But there is no `src.fill_value` ...

lbdreyer: In this test `src` was loaded in from a pp file, so the src.fill_value would presumably have been the bmdi value.

However, this test doesn't really care about the fill value so I have just let it use the numpy default fill_value

bjlittle:
@lbdreyer Thanks, this looks good to me 👍

PR #2722, opened 2017-08-03
Fix lib/iris/tests/unit/data_manager/test_DataManager.py tests

lbdreyer:
Fix all tests in lib/iris/tests/unit/data_manager/test_DataManager.py so that they all pass.

I have had to remove/change some tests that used the old API (e.g. cube.fill_value).
I have also had to make some fixes to handling of masked constants which hopefully makes sense

### DIFF
--- a/lib/iris/_data_manager.py
+++ b/lib/iris/_data_manager.py
@@ -187,6 +187,13 @@ class DataManager(object):
                     data = copy.deepcopy(self._lazy_array, memo)
                 else:
                     data = self._real_array.copy()
+            else:
+                # Check that the replacement data is valid relative to
+                # the currently managed data.
+                dm_check = DataManager(self.core_data())
+                dm_check.data = data
+                # If the replacement data is valid, then use it but
+                # without copying it.
             result = DataManager(data)
         except ValueError as error:
             emsg = 'Cannot copy {!r} - {}'
--- a/lib/iris/_lazy_data.py
+++ b/lib/iris/_lazy_data.py
@@ -93,6 +93,8 @@ def as_lazy_data(data, chunks=_MAX_CHUNK_SIZE):
         The input array converted to a dask array.
 
     """
+    if isinstance(data, ma.core.MaskedConstant):
+        data = ma.masked_array([data], mask=data.mask)
     if not is_lazy_data(data):
         asarray = not ma.isMaskedArray(data)
         data = da.from_array(data, chunks=chunks, asarray=asarray)
--- a/lib/iris/_lazy_data.py
+++ b/lib/iris/_lazy_data.py
@@ -94,7 +94,7 @@ def as_lazy_data(data, chunks=_MAX_CHUNK_SIZE):
 
     """
     if isinstance(data, ma.core.MaskedConstant):
-        data = ma.masked_array([data], mask=data.mask)
+        data = ma.masked_array(data.data, mask=data.mask)
     if not is_lazy_data(data):
         asarray = not ma.isMaskedArray(data)
         data = da.from_array(data, chunks=chunks, asarray=asarray)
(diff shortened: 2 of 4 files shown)


lib/iris/_data_manager.py
```diff
@@ -187,6 +187,13 @@ def _deepcopy(self, memo, data=None):
                     data = copy.deepcopy(self._lazy_array, memo)
                 else:
                     data = self._real_array.copy()
+            else:
+                # Check that the replacement data is valid relative to
+                # the currently managed data.
+                dm_check = DataManager(self.core_data())
+                dm_check.data = data
+                # If the replacement data is valid, then use it but
+                # without copying it.
```
lbdreyer: I am putting this back as this ensures you are doing a deepcopy with a data of the correct shape. 
Without this you can do:
```
>>> dm = DataManager(np.array([0,1,2]))
>>> dm._deepcopy(dict(), data=np.array(0))
DataManager(array(0))
```

Which shouldn't happen

lib/iris/_lazy_data.py
```diff
@@ -93,6 +93,8 @@ def as_lazy_data(data, chunks=_MAX_CHUNK_SIZE):
         The input array converted to a dask array.
 
     """
+    if isinstance(data, ma.core.MaskedConstant):
+        data = ma.masked_array([data], mask=data.mask)
```
bjlittle: @lbdreyer Doing this promotes the dimensionality of the data e.g.
```python
>>> a = ma.masked_array([1, 2, 3], mask=[0, 1, 0])
>>> v = a[1]
>>> v
masked
>>> r = ma.array([v.data], mask=v.mask)
>>> r
masked_array(data = [--],
             mask = [ True],
       fill_value = 1e+20)

>>> r.shape
(1,)
>>> r.ndim
1
>>> r = ma.array(v.data, mask=v.mask)
>>> r
masked_array(data = --,
             mask = True,
       fill_value = 1e+20)

>>> r.shape
()
>>> r.ndim
0
```

lbdreyer: np.asanyarray doesn't work
```
>>> a = ma.masked_array([8], mask=True)
>>> masked_constant = a[0]
>>> print type(masked_constant)
>>> print type(np.asanyarray(masked_constant))
```

lbdreyer: I also can't do 
```
>>> a = ma.masked_array(8, mask=True)
>>> print a[0]
  File "/usr/local/sci/lib/python2.7/site-packages/numpy/ma/core.py", line 2992, in __getitem__
    dout = ndarray.__getitem__(_data, indx)
IndexError: 0-d arrays can't be indexed
```

bjlittle: @lbdreyer Perhaps my point was lost in the example, but shouldn't line 97 be
```python
data = ma.masked_array(data, mask=data.mask)
```
instead of 
```python
data = ma.masked_array([data], mask=data.mask)
```

lbdreyer: Sorry my last comment makes no sense, that's actually relevant to the tests.

Can we change this line to be
`data = ma.masked_array(data, mask=data.mask)`
?

bjlittle: Yup, exactly!

lbdreyer: Nope

```
>>> orig = ma.masked_array([1,2,3], mask=[True, False, False])
>>> m_con = orig[0]
>>> data = ma.masked_array(m_con, mask=m_con.mask)
>>> print type(data)
```

😿

bjlittle: Apologies, that should be
```python
data = ma.masked_array(data.data, mask=data.mask)
```
so for your example above it's `m_con.data`

lbdreyer: Ah! You're right! That does indeed work.

I have now put up a new commit with that change in

PR #2723, opened 2017-08-03
Lazy masked fill_value retrieval for NC save 

### LINKED ISSUE #2703: Fix fill_value when saving to netCDF
If the data is lazy masked, we aren't correctly setting the fill_value (see [here](https://github.com/SciTools/iris/pull/2699/files#diff-e055d3568b9b3f9cfe397cec7e016700R1952))

We need to set the fill_value correctly, the detail of which is given in [this comment](https://github.com/SciTools/iris/pull/2699#discussion_r130849528)

dkillick:
Enables lazy masked `fill_value` retrieval for NetCDF save.

Dask exposes no mechanism for retrieving the fill value of a lazy masked array (indeed, arguably this is currently not a soluble problem). This PR proposes a workaround for this problem, exposes the workaround as a function in `iris._lazy_data` and integrates the function into the NetCDF saver.

Specifically, the workaround realises the smallest possible slice of the lazy masked data array that, when realised, contains the required fill value. Turns out the smallest possible slice is actually an empty slice:
```python
>>> m = np.ma.masked_array([0, 1, 2], mask=[1, 0, 1], fill_value=9999)
>>> dm = da.from_array(m, chunks=(3,), asarray=False)
>>> dm[:0].compute().fill_value
9999
```

Fixes #2703.

### DIFF
--- a/lib/iris/_lazy_data.py
+++ b/lib/iris/_lazy_data.py
@@ -126,6 +126,34 @@ def as_concrete_data(data):
     return data
 
 
+def lazy_masked_fill_value(data):
+    """
+    Return the fill value of a lazy masked array.
+
+    Args:
+
+    * data:
+        A dask array, NumPy `ndarray` or masked array
+
+    Returns:
+        The fill value of `data` if `data` represents a masked array, or None.
+
+    """
+    # If lazy, get the smallest slice of the data from which we can retrieve
+    # the fill_value.
+    if is_lazy_data(data):
+        inds = tuple([0] * (data.ndim-1))
+        smallest_slice = data[inds][:0]
+        data = as_concrete_data(smallest_slice)
+
+    # Now get the fill value.
+    fill_value = None
+    if ma.isMaskedArray(data):
+        fill_value = data.fill_value
+
+    return fill_value
+
+
 def multidim_lazy_stack(stack):
     """
     Recursively build a multidimensional stacked dask array.
--- a/lib/iris/tests/unit/lazy_data/test_lazy_masked_fill_value.py
+++ b/lib/iris/tests/unit/lazy_data/test_lazy_masked_fill_value.py
@@ -1,0 +1,86 @@
+# (C) British Crown Copyright 2017, Met Office
+#
+# This file is part of Iris.
+#
+# Iris is free software: you can redistribute it and/or modify it under
+# the terms of the GNU Lesser General Public License as published by the
+# Free Software Foundation, either version 3 of the License, or
+# (at your option) any later version.
+#
+# Iris is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU Lesser General Public License for more details.
+#
+# You should have received a copy of the GNU Lesser General Public License
+# along with Iris.  If not, see <http://www.gnu.org/licenses/>.
+"""Test the function :func:`iris._lazy data.lazy_masked_fill_value`."""
+
+from __future__ import (absolute_import, division, print_function)
+from six.moves import (filter, input, map, range, zip)  # noqa
+
+# Import iris.tests first so that some things can be initialised before
+# importing anything else.
+import iris.tests as tests
+
+import dask.array as da
+import numpy as np
+import numpy.ma as ma
+
+from iris._lazy_data import lazy_masked_fill_value, _MAX_CHUNK_SIZE
+
+
+class Test_as_lazy_data(tests.IrisTest):
+    def setUp(self):
+        shape = (2, 3, 4)
+        data = np.arange(24).reshape(shape)
+        mask = np.zeros(shape)
+        mask[data % 5 == 1] = 1
+        self.fill_value = 9999
+        self.m = ma.masked_array(data, mask=mask, fill_value=self.fill_value)
+        self.dm = da.from_array(self.m, asarray=False,
+                                chunks=_MAX_CHUNK_SIZE)
+
+    def test_lazy_masked_ND(self):
+        fill_value = lazy_masked_fill_value(self.dm)
+        self.assertEqual(fill_value, self.fill_value)
+
+    def test_lazy_masked_0D(self):
+        data = self.dm[0, 0, :1]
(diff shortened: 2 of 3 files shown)


dkillick:
Note: no tests yet! I thought I'd push the changes so that people could get eyes on the proposed functionality.

dkillick:
Now with unit tests! Still got a bunch of existing test failures to deal with, though.

lib/iris/_lazy_data.py
```diff
@@ -126,6 +126,34 @@ def as_concrete_data(data):
...
+    Returns:
+        The fill value of `data` if `data` represents a masked array, or None.
+
+    """
+    # If lazy, get the smallest slice of the data from which we can retrieve
+    # the fill_value.
+    if is_lazy_data(data):
+        inds = tuple([0] * (data.ndim-1))
+        smallest_slice = data[inds][:0]
+        data = as_concrete_data(smallest_slice)
```
bjlittle: @dkillick You might want coverage for 0-dim arrays and MaskedConstants ...

lib/iris/tests/unit/lazy_data/test_lazy_masked_fill_value.py
```diff
@@ -0,0 +1,86 @@
...
+import iris.tests as tests
+
+import dask.array as da
+import numpy as np
+import numpy.ma as ma
+
+from iris._lazy_data import lazy_masked_fill_value, _MAX_CHUNK_SIZE
+
+
+class Test_as_lazy_data(tests.IrisTest):
```
lbdreyer: This needs renaming

lib/iris/_lazy_data.py
```diff
@@ -126,6 +126,34 @@ def as_concrete_data(data):
     return data
 
 
+def lazy_masked_fill_value(data):
```
bjlittle: Would you but into a rebrand, say "lazy_fill_value" instead?

lib/iris/tests/unit/lazy_data/test_lazy_masked_fill_value.py
```diff
@@ -0,0 +1,86 @@
...
+    def setUp(self):
+        shape = (2, 3, 4)
+        data = np.arange(24).reshape(shape)
+        mask = np.zeros(shape)
+        mask[data % 5 == 1] = 1
+        self.fill_value = 9999
+        self.m = ma.masked_array(data, mask=mask, fill_value=self.fill_value)
+        self.dm = da.from_array(self.m, asarray=False,
+                                chunks=_MAX_CHUNK_SIZE)
```
lbdreyer: Could you not just call `iris._lazy_data.as_lazy_data`

lib/iris/_lazy_data.py
lbdreyer: Why calculate inds, why not just use [:0] 

i.e. 
```
    if is_lazy_data(data):
        smallest_slice = data[:0]
        data = as_concrete_data(smallest_slice)
```

dkillick: Should work; see [these tests](https://github.com/dkillick/iris/blob/68e2b6676d436728d3ca6b666bf356ad556e6b7d/lib/iris/tests/unit/lazy_data/test_lazy_masked_fill_value.py#L48-L56)... although I'm not convinced that covers the `MaskedConstant` case. Thoughts?

lib/iris/tests/unit/lazy_data/test_lazy_masked_fill_value.py
dkillick: Good spot! Thought I'd got away with copying an existing file...

lib/iris/_lazy_data.py
dkillick: You're going to have to try harder than that to convince me 😉 

I guess that...

* I only care about fill values if I have masked data,
* so 'masked' could be omitted in that sense, but
* clarity is a good thing – it's in the masked data case that I want to retrieve the fill value.

So I'm not currently buying into a rebrand! What do you see as the benefit of doing so?

lib/iris/tests/unit/lazy_data/test_lazy_masked_fill_value.py
dkillick: I could, but I don't see how that benefits me.

lib/iris/_lazy_data.py
dkillick: An interesting idea! I was going to say "Because then it won't be the smallest slice!", which turns out to not quite be accurate: 
```python
z = np.arange(24).reshape(2,3,4)
z[:0]
array([], shape=(0, 3, 4), dtype=int64)
```

Apparently (according to an offline conversation) doing the also causes the 0D case to not work properly. So, while I'm 👍 for reducing SLOC and code complexity, I think the extra line in this case adds some worthwhile resilience.

dkillick: How would one produce a `MaskedConstant` to test on? They seem to be things you can only make when you don't want to (i.e. I now want to and I can't manage to make one!) 😒

lbdreyer: The following works:
```
>>> a = ma.masked_array([4], mask=True)
>>> masked_constant = a[0]
>>> print type(masked_constant)
```

lib/iris/tests/unit/lazy_data/test_lazy_masked_fill_value.py
lbdreyer: Well that's just the argument we initially had of whether to include `as_lazy_data`.
We chose to put all our dealing with dask in a single module (`iris._lazy_data`) rather than having it dotted throughout the iris codebase

bjlittle:
@dkillick Awesome thanks 👍 

I'm going to bank this PR and follow-up with any/all rework in another PR, given your unavailability.

lib/iris/tests/unit/lazy_data/test_lazy_masked_fill_value.py
dkillick: That's fine... apart from the fact that it doesn't hold in the tests, which make widespread use of `da.from_array` (including the test module I duplicated to make this one). 

Either way, this intermediate step has now been banked, so do feel free to change this in a follow-up PR :smile:

PR #2726, opened 2017-08-03
Fix Cube unit tests.

lbdreyer:
Fill all the tests in  lib/iris/tests/unit/cube/test_Cube.py that were failing due to refactoring Iris to use dask mask array support.


lib/iris/tests/unit/cube/test_Cube.py
```diff
@@ -1457,255 +1348,127 @@ def test_lazy_data_masked__mask_set(self):
...
-    def test_lazy_data_masked_with_dtype_non_integral(self):
-        for dtype in self.dtypes:
-            data = ma.array([0, 1], dtype=dtype)
-            data[0] = ma.masked
-            emsg = 'Can only cast lazy data to an integer or boolean dtype'
-            with self.assertRaisesRegexp(ValueError, emsg):
-                Cube(as_lazy_data(data), dtype=np.complex)
-
 
 class Test_data_dtype_fillvalue(tests.IrisTest):
```
lbdreyer: I thought it might be worth retaining some of these tests, but I do wonder if it's a bit overkill. I sometimes felt like I was testing numpy rather.

dkillick: To catch up an offline conversation...

I think the tests updated here are valuable for the coverage they add, but perhaps this isn't the best place to keep them. Given that they're effectively testing the interface between Iris cubes (where the behaviour is finally observed), the Data Manager and NumPy functionality, I wonder if these would be best retained elsewhere as cube integration tests.

lib/iris/tests/unit/cube/test_Cube.py
```diff
@@ -1457,255 +1348,127 @@ def test_lazy_data_masked__mask_set(self):
...
-
-    def test_realdata___getitem__(self):
-        fill_value = 123
-        cube = Cube(np.arange(10), fill_value=fill_value)
-        subcube = cube[5:]
-        self.assertArrayAllClose(subcube.fill_value, fill_value)
-
-    def test_lazydata___getitem__(self):
-        fill_value = 123
+    def test_lazydata___getitem__dtype(self):
```
dkillick: Is there value in also checking the fill value following a getitem call?

lib/iris/tests/unit/cube/test_Cube.py
```diff
@@ -1457,255 +1348,127 @@ def test_lazy_data_masked__mask_set(self):
...
-            data[0] = ma.masked
-            emsg = 'Can only cast lazy data to an integer or boolean dtype'
-            with self.assertRaisesRegexp(ValueError, emsg):
-                Cube(as_lazy_data(data), dtype=np.complex)
-
 
 class Test_data_dtype_fillvalue(tests.IrisTest):
-    def _sample_data(self, dtype='f4', masked=False, lazy=False):
+    def setUp(self):
+        self.fill_value = 1234
```
dkillick: Looks like you only use this in two other places, so one could argue you don't really need this?

lib/iris/tests/unit/cube/test_Cube.py
```diff
@@ -1457,255 +1348,127 @@ def test_lazy_data_masked__mask_set(self):
...
+        # Check touching masked lazy data.
+        fill_value = 27.3
+        cube = self._sample_cube(masked=True, fill_value=fill_value, lazy=True)
         data = cube.data
-        self.assertTrue(ma.is_masked(data))
-        self.assertMaskedArrayEqual(data, masked_data_original)
-        self.assertArrayAllClose(cube.fill_value, fill_value)
+        self.assertIs(cube.core_data(), data)
+        self.assertEqual(cube.dtype, np.float32)
         self.assertArrayAllClose(data.fill_value, fill_value)
```
dkillick: I know you haven't changed this, but I guess that's why I'm asking about it... Two points:

* why is the fill value being tested as an array? Shouldn't there be a single fill value for all points in `data`?
* why is the value only _close_? Surely both values should be identical?

lib/iris/tests/unit/cube/test_Cube.py
lbdreyer: Yep I'll remove it.I had put it here previously as I had a few more tests that used it

lib/iris/tests/unit/cube/test_Cube.py
lbdreyer: I have now added a fill_value check

lib/iris/tests/unit/cube/test_Cube.py
lbdreyer: I have no idea. I thought it might be something about the data type being different, so comparing a float and an int would be "almost equal"

pp-mo: I believe I may have written this, if not I think I certainly "could have done".
I interpret it as just a way of doing a toleranced floating-point comparison
 \- because it is IMHO  "basically always wrong" to test floating-point values for exact equivalance.

pp-mo:
A bunch of the errors here are to do with Travis not correctly installing the latest Dask variant.
Now fixed on main feature branch, so rebase wanted.
Now closing this to replace with #2734 -- which is exactly just a rebased version of this.

PR #2728, opened 2017-08-04
PP + merge test fixes

bjlittle:
This PR addresses assorted PP save test failures.

I also ended up bleeding thru into merge and fixing those tests to confirm that merge was behaving correctly for merging masked and ndarray cube payload combos.

We also require to force an `asarray=False` default for `_lazy_data.as_lazy_data`, otherwise `dask` can cast masked data derived from a proxy as a ndarray.


lib/iris/tests/test_merge.py
```diff
@@ -187,7 +190,7 @@ def test__ndarray_ndarray(self):
             expected = self._make_data([0, 1], dtype=self.dtype)
             self.assertArrayEqual(result.data, expected)
             self.assertEqual(result.dtype, self.dtype)
-            self._check_fill_value(result, fill0, fill1)
+            self._check_fill_value(result)
```
lbdreyer: For tests including ndarray (e.g. `test__ndarray_ndarray`, `test__masked_ndarray`, `test__ndarray_masked`) it seems unnesccary to test the `fill_combos` since we already know they won't have fill_values

lbdreyer: Unless you want to check that after a merge, the resultant data isn't a masked array for some reason

lib/iris/fileformats/pp.py
```diff
@@ -1382,11 +1379,6 @@ def save(self, file_handle):
         # Get the actual data content.
         data = self.data
         if ma.is_masked(data):
-            # Fill missing data points with the MDI value from the header.
-            if data.dtype.kind in 'biu':
-                # Integer or Boolean data : No masking is supported.
-                msg = 'Non-floating masked data cannot be saved to PP.'
-                raise ValueError(msg)
```
lbdreyer: This was added in #2564 which has now been removed anyway

lbdreyer:
@bjlittle This all looks good. I just want to talk to you about the testing of fill_values when merging ndarrays but otherwise I think this is ready to merge in

PR #2747, opened 2017-08-22
Handle fill value on netCDF save

djkirkham:
No longer try to get a fill value from the cube's data array for saving to netCDF. Instead, allow the user to pass one in as a keyword argument or use the netCDF default.

Unresolved issues:

- [x] The behaviour of netCDF is to use a default fill value if none is specified, unless the dtype is a single byte in length, in which case the data is interpreted as being non-masked (on writing to a data set, the data is still filled). So the warning raised by my code doesn't apply in this case, and in fact perhaps a warning should be raised if the user tries to store masked byte data.
- [x] I noticed that that `packing` argument to save can be an iterable to allow different packing attributes for the different cubes. Perhaps we could do the same for `fill_value`.
- [x] Could probably reword some of the docstrings
- [x] Unit tests


lib/iris/fileformats/netcdf.py
```diff
@@ -902,11 +919,16 @@ def write(self, cube, local_keys=None, unlimited_dimensions=None,
             on `cube.data` and possible masking. For masked data, `fill_value`
             is taken from netCDF4.default_fillvals. For more control, pass a
```
pp-mo: I think this needs a tweak as it seems to say that if 'packing' is used then 'fill_value' would be ignored.
Whereas, the new note for 'fill_value' itself implies that it *will* be used for packing, if given, which I think is correct.

lib/iris/fileformats/netcdf.py
```diff
@@ -1858,54 +1881,61 @@ def _create_cf_data_variable(self, cube, dimension_names, local_keys=None,
             An interable of cube attribute keys. Any cube attributes
             with matching keys will become attributes on the data variable.
 
+        * packing (type or string or dict or list): A numpy integer datatype
```
pp-mo: It seems a bit fragile to repeat all this between `__init__` and `_create_cf_data_variable`.
Especially as this is not a public docstring for Sphinx, I think it would make more sense to cross-refer to the other here. How about just "`packing` and `fill_value` controls are applied to the data, as described in `__init__`."

djkirkham:
I've pushed changes so that the _FillValue attribute is removed from the data if it wasn't masked. Some things to note:

- The _FillValue attribute is always removed in the non-masked case, even if it was explicitly passed by the user
- The warning about clashes with the fill value in the data only occurs if the data is masked. This prevents warnings when writing byte data, which is never treated as masked by netCDF anyway, but we might want to have the warning for other data types.

lib/iris/tests/unit/fileformats/netcdf/test_Saver.py
```diff
@@ -158,12 +158,13 @@ def test_big_endian(self):
     def test_zlib(self):
         cube = self._simple_cube('>f4')
         with mock.patch('iris.fileformats.netcdf.netCDF4') as api:
+            api.default_fillvals = {'f4': 12345.}
```
djkirkham: I've had to modify this test because the `createVariable` call now passes the default fill value for the data type. Since the `netCDF4` module is mocked out, it turns out that the value passed for the `fill_value` argument is a numpy array. I'm guessing that this doesn't play well with `mock.ANY` (it probably tries to do an equality test somewhere along the line, which causes numpy to raise an exception), because the test fails unless I make sure there is an actual value passed by assigning this dict.

lib/iris/tests/unit/fileformats/netcdf/test_Saver.py
```diff
@@ -158,12 +158,13 @@ def test_big_endian(self):
...
             with Saver('/dummy/path', 'NETCDF4') as saver:
                 saver.write(cube, zlib=True)
         dataset = api.Dataset.return_value
         create_var_calls = mock.call.createVariable(
             'air_pressure_anomaly', np.dtype('float32'), ['dim0', 'dim1'],
-            fill_value=None, shuffle=True, least_significant_digit=None,
+            fill_value=mock.ANY, shuffle=True, least_significant_digit=None,
```
djkirkham: Unfortunately I've had to make this test less strict. I could check that there is a `delncattr` call to remove the `_FillValue`...

pp-mo:
Some initial thoughts on the behaviour, following your comments :
> I've pushed changes so that the _FillValue attribute is removed from the netCDF variable if the data wasn't masked. Some things to note:
> * The _FillValue attribute is always removed in the non-masked case, even if it was explicitly passed by the user
> * The warning about clashes with the fill value in the data only occurs if the data is masked. This prevents warnings when writing byte data, which is never treated as masked by netCDF anyway, but we might want to have the warning for other data types.

I think both those points differ from the conclusions of yesterday's discussion with @pelson 
Which I've tried to summarise in #2748 .
It's a shame we don't have @pelson to verify that right now, as the proposals changed several times in discussion + I might have mis-represented him.

So, for what it's worth, **I** think that outcome says ...
 * every nc variable is assigned a `_FillValue` attribute
    * ***except*** byte ones, where it removed if the data has no masked points
 * the check on clashes with the fill value, and possible resulting warning, is *always* done
 * there is also a blanket warning when saving masked bytes

pp-mo:
My own personal inclination would be to add a warning when saving masked int/uint32 **as well**...
I think this is potentially more awkward than the byte cases
 * present usage in netcdf/netCDF4-python mean byte types essentially don't allow masking
 * but short integer types (2 or 4 byes) can't *avoid* it, for some larger values.

The actual 'missing' values, from `netCDF4.default_fillvals`, are
 * u2 --> 65335 = 2^16 - 1 = 0xffff
   valid range is 0 .. 65534
 * i2 --> -32767 = 1 - 2^15
   valid range is -32766 ..+32767
 * u4 --> 4,294,967,295 = 2^32 - 1
   valid range is 0 .. 4,294,967,294
 * i4 --> -2,147,483,647 = 1 - 2^31
   valid range is -2,147,483,647 .. -2,147,483,646

Note that in the signed cases, the fill value is not "MININT", but "MININT+1", equal to _minus_ MAXINT,
so actually there are **_two_** values missing from the signed type ranges.

djkirkham:
@pp-mo 

>- every nc variable is assigned a _FillValue attribute
>    - except byte ones, where it removed if the data has no masked points

This would have the same effect as not assigning a _FillValue at all because of how the netCDF4 library works. Personally, I think we should avoid writing a fill value if we don't have to; it's what we tried to do before (<=1.13), and changing it means changing a lot of test results. Though I don't much mind either way.

>- the check on clashes with the fill value, and possible resulting warning, is always done

OK. I'm assuming byte data would be an exception here. I guess the reason I avoided this was because I didn't want to have to add branching based on the dtype.

>- there is also a blanket warning when saving masked bytes

OK, but what if the user passes their own fill value? Perhaps that should override the warning?

>My own personal inclination would be to add a warning when saving masked int/uint32 as well...

Assuming you mean a blanket warning like in the byte case, I disagree with this. I think it's enough to warn if there are clashes: round trips still work and maintain the mask, I don't see it causing problems if there are no clashes.

lib/iris/fileformats/netcdf.py
```diff
@@ -1973,19 +1968,26 @@ def store(data, cf_var, fill_value):
...
+            if is_masked:
+                warnings.warn("Cube '{}' contains masked byte data and will "
+                              "be interpreted as unmasked. To save as masked "
+                              "data please explicitly provide a fill value."
+                              .format(cube.name()))
+        elif contains_fill_value:
+            warnings.warn("Cube '{}' contains data points equal to the fill"
+                          "value {}. The points will be interpreted as being "
+                          "masked. Please provide a fill_value argument not "
+                          "equal to any data point.".format(cube.name(),
```
djkirkham: @pp-mo Please feel free to suggest a rewording of these warnings

pp-mo:
Looking much more sensible and consistent now, and the code looks neater too 👍 !
A few remaining points...
 * +1 for making 'fill_value' an iterable, like 'packing', for the save() call
   * and documenting it
 * some more tests for the new behaviour
 * remove the redundant [import of `iris._lazy_data.get_fill_value`](https://github.com/SciTools/iris/blob/d6f543be340513bd4816b068a9d294d42bbe6ab8/lib/iris/fileformats/netcdf.py#L59)
   * .. and then **that routine itself** as nothing else refers to it

lib/iris/fileformats/netcdf.py
```diff
@@ -1865,47 +1893,32 @@ def _create_cf_data_variable(self, cube, dimension_names, local_keys=None,
...
-                        fill_value = packing['fill_value']
+
+        if packing:
+            if isinstance(packing, dict):
+                if 'dtype' not in packing:
+                    msg = "The dtype attribute is required for packing."
+                    raise ValueError(msg)
+                dtype = np.dtype(packing['dtype'])
+                scale_factor = packing.get('scale_factor', None)
+                add_offset = packing.get('add_offset', None)
```
pelson: Not your code, but would you mind checking that there aren't other keys in there other than the ones handled?

lib/iris/fileformats/netcdf.py
```diff
@@ -1865,47 +1893,32 @@ def _create_cf_data_variable(self, cube, dimension_names, local_keys=None,
...
+            else:
+                masked = ma.isMaskedArray(cube.data)
```
pelson: This would trigger the data to be loaded. Is that what you intended?

pelson: Sorry not your code... Perhaps you wouldn't mind adding a comment though:
```
# We compute the scale_factor based on the min/max of the data. This requires the data to be loaded.
```

lib/iris/fileformats/netcdf.py
```diff
@@ -1924,36 +1937,55 @@ def set_packing_ncattrs(cfvar):
             self._dataset.file_format in ('NETCDF3_CLASSIC',
```
pelson: Do we still need the ``has_lazy_data`` here? It'd be nice to remove.

lib/iris/fileformats/netcdf.py
```diff
@@ -1924,36 +1937,55 @@ def set_packing_ncattrs(cfvar):
...
+
+        # Create the cube CF-netCDF data variable with data payload.
+        cf_var = self._dataset.createVariable(cf_name, dtype, dimension_names,
+                                              fill_value=fill_value,
+                                              **kwargs)
+        set_packing_ncattrs(cf_var)
+
+        # If packing attributes are specified, don't bother checking whether
+        # the fill value is in the data.
+        fill_value_to_check = None if packing else \
```
pelson: Let's unpack this line. Neat as it is, it really isn't very readable 😉

lib/iris/fileformats/netcdf.py
```diff
@@ -2244,13 +2284,28 @@ def is_valid_packspec(p):
                 raise ValueError(msg)
         packspecs = packing
 
+    if isinstance(fill_value, six.string_types):
```
pelson: This whole block is surprising in its data-validation. Is it consistent with other parts of this codebase? If not, what is the motivation for adding such complexity?

lib/iris/fileformats/netcdf.py
```diff
@@ -741,6 +741,26 @@ def _setncattr(variable, name, attribute):
...
+    def __init__(self, target, fill_value=None):
+        self.target = target
+        self.fill_value = fill_value
+        self.contains_value = False
+        self.is_masked = False
+
+    def __setitem__(self, keys, arr):
+        if self.fill_value is not None:
+            self.contains_value |= self.fill_value in arr
+        self.is_masked |= ma.isMaskedArray(arr)
```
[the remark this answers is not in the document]
djkirkham: I had assumed `|=` would short circuit but I guess not.

lib/iris/fileformats/netcdf.py
djkirkham: > This would trigger the data to be loaded. Is that what you intended?
I hadn't noticed this, it looks like it comes from the old code. It can be fixed, but it would require getting the min and max of the data while storing it. I can do that if you like, but it might be better to do it as a separate PR

lib/iris/fileformats/netcdf.py
djkirkham: It determines which `store` method gets defined, although I could move the `if` into that method if you like.

lib/iris/fileformats/netcdf.py
djkirkham: OK

lib/iris/fileformats/netcdf.py
djkirkham: I need to be handle a single value or iterable of values which may be strings. If it's a single value it needs to be converted to an iterable that can be `zip`ped with `cube` if `cube` is a list. I couldn't find a more succinct way of achieving that, though maybe there is.

lib/iris/fileformats/netcdf.py
pelson: No, just adding the comment is good.

lib/iris/fileformats/netcdf.py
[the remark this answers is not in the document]
pelson: you're going to need to double check that for me now. 😄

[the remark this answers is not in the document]
djkirkham: Yeah, I had already checked. It's because it's bitwise, and in fact it changes the data type of the left hand side.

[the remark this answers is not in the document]
pelson: Ok. Well, at least you can get away with the boolean operators...

```
>>> def foo(val):
...    print('called')
...    return val
... 

>>> False or foo(False)
called
False
>>> True or foo(False)
True
```

lib/iris/fileformats/netcdf.py
```diff
@@ -899,14 +919,18 @@ def write(self, cube, local_keys=None, unlimited_dimensions=None,
...
-            is taken from netCDF4.default_fillvals. For more control, pass a
-            dict with one or more of the following keys: `dtype` (required),
-            `scale_factor`, `add_offset`, and `fill_value`. Note that automatic
-            calculation of packing parameters will trigger loading of lazy
-            data; set them manually using a dict to avoid this. The default is
-            `None`, in which case the datatype is determined from the cube and
-            no packing will occur.
+            on `cube.data` and possible masking. For more control, pass a dict
+            with one or more of the following keys: `dtype` (required),
+            `scale_factor` and `add_offset`. Note that automatic calculation of
```
pp-mo: Isn't this the reference you were looking form, saying that  automatic packing realises the data ?

lib/iris/fileformats/netcdf.py
pp-mo: Doesn't the docstring at [this comment](https://github.com/SciTools/iris/pull/2747/files/5d8375a2cae41dd7041073f9e472fade533dc1ae#r135280518) cover it ?

lib/iris/fileformats/netcdf.py
pp-mo: Can you not use np.array(..., ndmin=1) for this ?
E.G.
```
>>> for thing in ('a', ('a', 'bc'), ['ab', 'c'], 1.0, [1, 2], np.array([1., 2])):
...   print thing, np.array(thing, ndmin=1)
... 
a ['a']
('a', 'bc') ['a' 'bc']
['ab', 'c'] ['ab' 'c']
1.0 [ 1.]
[1, 2] [1 2]
[ 1.  2.] [ 1.  2.]
>>> 
```
It does the special-casing of strings for you, and handles all the iterables uniformly.
I'm not sure about the repeat operation though.

lib/iris/fileformats/netcdf.py
pelson: Not really. It is about making the code readable. Doesn't need to be an essay.

lib/iris/fileformats/netcdf.py
pelson: Nice suggestion.

djkirkham: Yeah I was playing around with it but it doesn't work for generators.

docs/iris/src/whatsnew/contributions_2.0/incompatiblechange_2017-Aug-29_netCDF_packing_changes.txt
```diff
@@ -0,0 +1,3 @@
+* A 'fill_value' key can no longer be specified as part of the `packing` argument to `iris.save` when saving in netCDF format. Instead, a fill value should be specified as a separate `fill_value` argument if required.
```
pp-mo: Please add a note saying that multiple fill values may be supplied, and these are applied to output cubes in sequence.

docs/iris/src/whatsnew/contributions_2.0/incompatiblechange_2017-Aug-29_fill_value_save_arg.txt
```diff
@@ -0,0 +1 @@
+* When saving a cube or list of cubes, a fill value or list of fill values can be specified via a new `fill_value` argument. If a `fill_value` argument is not specified, the default fill value for the file format and the cube's data type will be used. Fill values are no longer taken from the cube's `data` attribute when it is a masked array.
```
pp-mo: Can we add a note explaining that multiple fill values are applied to output cubes in sequence.

djkirkham: OK

docs/iris/src/whatsnew/contributions_2.0/incompatiblechange_2017-Aug-29_netCDF_packing_changes.txt
djkirkham: How I about I just add "or list of fill values"? I think we can assume that anyone familiar with the packing attribute knows that if you specify a list of values they applies to each cube in turn. I will modify the other what's new entry to explain what happens when you pass a list of fill values.

lib/iris/fileformats/netcdf.py
```diff
@@ -741,6 +741,26 @@ def _setncattr(variable, name, attribute):
...
+    def __init__(self, target, fill_value=None):
+        self.target = target
+        self.fill_value = fill_value
+        self.contains_value = False
+        self.is_masked = False
+
+    def __setitem__(self, keys, arr):
+        if self.fill_value is not None:
+            self.contains_value = self.contains_value or self.fill_value in arr
+        self.is_masked = self.is_masked or ma.isMaskedArray(arr)
```
pp-mo: I think this should be using "ma.is_masked" instead of "ma.isMaskedArray".
As stated [here](https://github.com/SciTools/iris/issues/2748#issuecomment-324578818) "make no difference between masked + non-masked arrays"

lib/iris/fileformats/netcdf.py
```diff
@@ -1924,36 +1946,58 @@ def set_packing_ncattrs(cfvar):
...
-            # Create the cube CF-netCDF data variable with data payload.
-            cf_var = self._dataset.createVariable(
-                cf_name, dtype, dimension_names,
-                fill_value=fill_value, **kwargs)
-            set_packing_ncattrs(cf_var)
-            cf_var[:] = data
-
+            def store(data, cf_var, fill_value):
+                cf_var[:] = data
+                is_masked = ma.isMaskedArray(data)
```
pp-mo: Here too should be using "ma.is_masked" instead of "ma.isMaskedArray".
As stated [here](https://github.com/SciTools/iris/issues/2748#issuecomment-324578818) "make no difference between masked + non-masked arrays"

lib/iris/fileformats/netcdf.py
pp-mo: I've finally accepted that this *can't* be improved on, though it is painful !
However, could you add a comment to say exactly what all this is about ?
? Maybe something like ...
```
    # Make full-value(s) into an iterable over cubes.
    if isinstance(fill_value, six.string_types):
        # Strings are awkward -- handle separately.
        fill_values = repeat(fill_value)
    else:
        ...
```

lib/iris/tests/unit/fileformats/netcdf/test_Saver.py
```diff
@@ -24,10 +24,16 @@
 # importing anything else.
 import iris.tests as tests
 
+from contextlib import contextmanager
+import re
```
pp-mo: This import is now obsolete

lib/iris/tests/unit/fileformats/netcdf/test_Saver.py
```diff
@@ -24,10 +24,16 @@
...
+import warnings
+
 import netCDF4 as nc
 import numpy as np
+from numpy import ma
 
 import iris
```
pp-mo: This import is now obsolete

lib/iris/tests/unit/fileformats/netcdf/test_Saver.py
```diff
@@ -24,10 +24,16 @@
...
+import iris._lazy_data
```
pp-mo: You could just import `as_lazy_data` here, as that is all that is used.

lib/iris/tests/unit/fileformats/netcdf/test_Saver.py
pp-mo: Sorry my mistake, must have been looking at the wrong version.

pp-mo:
NOTE: errors are due to the special branch "jcrist/masked-array" no longer existing.
Meanwhile, we can't getting the newly-merged latest dask from conda because first we will need a distinct version string to identify it in channels.  https://github.com/dask/dask/pull/2654 addresses.

pp-mo:
> can't access the newly-merged latest dask code

We can now:
  * mask support went in at dask=0.15.3
  * currently version 0.15.4
  * this latest version has useful fixes, relevant to Iris tests (#2769)

@djkirkham can you please re-base this ?
I have been testing a prototype version in : #2780
(with other bits force-merged)

djkirkham:
@pp-mo Tests are passing now

lib/iris/fileformats/netcdf.py
```diff
@@ -899,14 +919,18 @@ def write(self, cube, local_keys=None, unlimited_dimensions=None,
             http://www.unidata.ucar.edu/software/netcdf/docs/BestPractices.html#bp_Packed-Data-Values
             If this argument is a type (or type string), appropriate values of
             scale_factor and add_offset will be automatically calculated based
-            on `cube.data` and possible masking. For masked data, `fill_value`
-            is taken from netCDF4.default_fillvals. For more control, pass a
-            dict with one or more of the following keys: `dtype` (required),
-            `scale_factor`, `add_offset`, and `fill_value`. Note that automatic
```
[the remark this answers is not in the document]
pp-mo: Agree with @pelson : this needs a whatsnew entry,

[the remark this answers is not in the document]
pp-mo: Also... 
the text that used to be here in 1.13 saying "or an iterable of such types, strings or dicts" appears to have got lost.
That concept now also applies to the `fill_value` keyword, where it also needs explaining.

lib/iris/fileformats/netcdf.py
pp-mo: I will consider this later, as I'm thinking to integrate this with the same requirements in the PP saver
( #2744, to be updated when this is done )

lib/iris/fileformats/netcdf.py
```diff
@@ -2244,13 +2284,28 @@ def is_valid_packspec(p):
...
         # Iterate through the cubelist.
-        for cube, packspec in zip(cubes, packspecs):
+        for cube, packspec, fill_value in zip(cubes, packspecs, fill_values):
             sman.write(cube, local_keys, unlimited_dimensions, zlib, complevel,
                        shuffle, fletcher32, contiguous, chunksizes, endian,
-                       least_significant_digit, packing=packspec)
+                       least_significant_digit, packing=packspec,
+                       fill_value=fill_value)
 
         if iris.config.netcdf.conventions_override:
```
pp-mo: My Eclipse says this needs an extra import to be truly righteous.

docs/iris/src/whatsnew/contributions_2.0/incompatiblechange_2017-Aug-29_fill_value_save_arg.txt
```diff
@@ -0,0 +1 @@
+* When saving a cube or list of cubes, a fill value or list of fill values can be specified via a new `fill_value` argument. If a list is supplied, each fill value will be applied to each cube in turn. If a `fill_value` argument is not specified, the default fill value for the file format and the cube's data type will be used. Fill values are no longer taken from the cube's `data` attribute when it is a masked array.
```
pp-mo: This isn't a general facility -- it only applies to certain save formats (netcdf only here, PP to be added shortly..)

lib/iris/tests/unit/fileformats/netcdf/test_Saver.py
```diff
@@ -325,6 +330,154 @@ def test_valid_max_saved(self):
...
+        # Test that masked lazy data saves correctly when given a fill value.
+        index = (1, 1)
+        fill_value = 12345.
+        cube = self._make_cube('>f4', masked_index=index, lazy=True)
+        with self._netCDF_var(cube, fill_value=fill_value) as var:
+            self.assertEqual(var._FillValue, fill_value)
+            self.assertTrue(var[index].mask)
+
+    def test_mask_lazy_default_fill_value(self):
+        # Test that masked lazy data saves correctly when given a fill value.
```
pp-mo: Should say "when ***not*** given a fill value" here.

docs/iris/src/whatsnew/contributions_2.0/incompatiblechange_2017-Aug-29_fill_value_save_arg.txt
djkirkham: ok, I'll add mentions for both file formats here.

lib/iris/tests/unit/fileformats/netcdf/test_Saver.py
```diff
@@ -325,6 +330,154 @@ def test_valid_max_saved(self):
...
+        # Test that no warning is raised if the data contains the fill_value at
+        # a masked point.
+        fill_value = 1
+        cube = self._make_cube('>f4', masked_value=fill_value)
+        with self._warning_check(
+                'contains data points equal to the fill value', False):
+            with self._netCDF_var(cube, fill_value=fill_value):
+                pass
+
+    def test_masked_byte_default_fill_value(self):
```
pp-mo: Possibly a missing testcase here ?
 = saving masked byte data _with a nonstandard fill-value_

It seems that the existing code this will _not_ generate a warning on save for this.
I'm not too clear what you would get when you read it back.
I'm not totally clear what it "ought" to do ?

djkirkham:
@pp-mo I think I've addressed all your comments and the tests are now passing.

pp-mo:
Thanks for all your patience @djkirkham !

pp-mo:
> My Eclipse says this needs an extra import to be truly righteous.

Sadly though, it still says that when the thing was fixed :-(

PR #2753, opened 2017-08-23
Fixes for GRIB unit tests

### LINKED ISSUE #2709: Investigate what to do with data in GribWrapper when deferred
See this comment https://github.com/SciTools/iris/pull/2699/files#r130800469

It appears we didn't do anything with `message_values` previously. Does this need changing

### LINKED ISSUE #2710: Fix Grib test failiures
Tests that need fixing are as follows:

AttributeError: Cannot find GRIB key data
 - [x] ERROR: test_grib1_known_standard_params (iris.tests.test_grib_load_translations.TestGrib1LoadPhenomenon)
- [x] ERROR: test_grib1_unknown_local_param (iris.tests.test_grib_load_translations.TestGrib1LoadPhenomenon)
- [ ] ERROR: test_grib1_unknown_standard_param (iris.tests.test_grib_load_translations.TestGrib1LoadPhenomenon)
- [x] ERROR: test_grib1_unknownparam (iris.tests.test_grib_load_translations.TestGrib1LoadPhenomenon)

mask (bitmap) is wrong
- [x] FAIL: test_bitmap_present_float_data (iris.tests.unit.fileformats.grib.message.test_GribMessage.Test_data__masked)
- [x] FAIL: test_bitmap_present_int_data (iris.tests.unit.fileformats.grib.message.test_GribMessage.Test_data__masked)

djkirkham:
Closes #2710
Closes #2709 

Relatively straightforward:
- Fixed the bug where the data array of a `GribWrapper` wasn't being assigned to, and unskipped relevant tests
- Removed uses of `nans_replacement` in other unit tests.
- Removed a duplicate unit test which only differed on the dtype of the data. It was testing the "realised dtype" and "nans_replacement" code.

### DIFF
--- a/lib/iris/tests/unit/fileformats/grib/message/test_GribMessage.py
+++ b/lib/iris/tests/unit/fileformats/grib/message/test_GribMessage.py
@@ -91,46 +91,21 @@ class Test_data__masked(tests.IrisTest):
         self.assertEqual(result.shape, self.shape)
         self.assertArrayEqual(result, expected)
 
-    @tests.skip_dask_mask
-    def test_bitmap_present_int_data(self):
+    def test_bitmap_present(self):
         # Test the behaviour where bitmap and codedValues shapes
-        # are not equal, and codedValues is integer data.
-        data_type = np.int64
+        # are not equal.
         input_values = np.arange(5)
-        output_values = np.array([-1, -1, 0, 1, -1, -1, -1, 2, -1, 3, -1, 4],
-                                 dtype=data_type)
+        output_values = np.array([-1, -1, 0, 1, -1, -1, -1, 2, -1, 3, -1, 4])
         message = _make_test_message({3: self._section_3,
                                       6: {'bitMapIndicator': 0,
                                           'bitmap': self.bitmap},
                                       7: {'codedValues': input_values}})
-        result = as_concrete_data(message.data, nans_replacement=ma.masked,
-                                  result_dtype=data_type)
-        expected = ma.masked_array(output_values,
-                                   np.logical_not(self.bitmap))
-        expected = expected.reshape(self.shape)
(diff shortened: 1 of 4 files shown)


djkirkham:
Now also fixes `_DataProxy` tests which were sending an extra argument to `__init__`

djkirkham:
There are still GRIB related tests failing based on handling of the fill value, in `lib/iris/tests/unit/fileformats/grib/save_rules/test_data_section.py`. I believe that the code currently takes the fill_value from the numpy array if it is masked, but the current strategy is to not use that attribute (see https://github.com/SciTools/iris/issues/2748). It's probably best to handle those changes on a separate PR.

lib/iris/tests/unit/fileformats/grib/message/test_GribMessage.py
```diff
@@ -91,46 +91,21 @@ def test_no_bitmap(self):
         self.assertEqual(result.shape, self.shape)
         self.assertArrayEqual(result, expected)
 
-    @tests.skip_dask_mask
-    def test_bitmap_present_int_data(self):
+    def test_bitmap_present(self):
```
lbdreyer: I'm missing something. Why don't we need to test for floats as well as integers?

lbdreyer: Sorry I realise you put a comment on this in the description

djkirkham: I suppose it wouldn't hurt to have them both, but I believe this test was split in two when we had the "nan array for masks" behaviour, which differed depending on the dtype. Now that we don't have that it seemed sensible just to collapse them back into a single test again.

lbdreyer:
Can you rebase this now that the PR to make it run serially (temporarily) is in

PR #2762, opened 2017-09-04
Use conda-forge for Dask in Travis (not special branch download) 

pp-mo:
**DO NOT MERGE UNTIL..**
https://github.com/dask/dask/pull/2654 is merged.

Because we were temporarily using a "jcrist/masked-array" branch of dask
 -- which is now merged.

See https://github.com/SciTools/iris/pull/2747#issuecomment-327003711

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -82,13 +82,6 @@ install:
       ln -s $(pwd)/iris-test-data-${IRIS_TEST_DATA_SUFFIX} iris-test-data;
     fi
 
-# Download + install dask from dask PR #2301
-  - conda uninstall --quiet dask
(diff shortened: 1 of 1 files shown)


pp-mo:
STATUS UPDATE:
 - PR was merged, but no version yet defined at that time
latest code has modded this to reference latest-master
 - that is still not the desired endpoint !

Later... there is now a tagged version 0.15.3, but unfortunately it introduces some new problems for us #2769

pp-mo:
STATUS UPDATE:

Latest Dask version (0.15.4) should now be suitable 
 -- let's do this ...

PR #2774, opened 2017-10-09
Bump travis python version 3.5 to 3.6

djkirkham:
Use Python 3.6 for the Python 3 Travis tests due to[ issues using python 3.5](https://github.com/SciTools/iris/pull/2769#issuecomment-335147830).

PR #2775, opened 2017-10-09
Use any latest Python 3 version in Travis matrix.

pp-mo:
In response to #2774,
let's just see what unpinning the minor version will do ...


pp-mo:
Didn't go well !
I don't think we have that capability : this a Travis thing, not conda.

PR #2785, opened 2017-10-11
Remove deprecated private interpolation modules

corinnebosley:
Removal of deprecated modules iris.analysis._interpolate_private and iris.analysis._interpolate_backdoor.

This PR will also require a check through the rest of the codebase to replace any remaining usages of the deprecated functions with their equivalents.

WIP.  Probably for quite a while.


lib/iris/analysis/_scipy_interpolate.py
```diff
@@ -43,8 +43,6 @@ def _ndim_coords_from_arrays(points, ndim=None):
     return points
 
 
-# source: https://github.com/scipy/scipy/blob/b94a5d5ccc08dddbc88453477ff2625\
-# 9aeaafb32/scipy/interpolate/interpolate.py#L1400
```
dkillick: Why have you removed this comment?

dkillick:
@corinnebosley I take it that in the cases where you have removed deleted imports from test module and not made any other changes in those modules that those imports weren't being used elsewhere in the test module?

lib/iris/analysis/_scipy_interpolate.py
corinnebosley: Because it seemed appropriate.  It's starting to get really difficult to work out what I'm looking at or doing in IntelliJ, so my strategy is: make some changes, push them up so I can see them properly, then figure out if that was a good idea or not.

I can see now that removing that line was not a good idea because it will still be relevant, so I'll pop that back in.

corinnebosley:
@dkillick I am in the process now of looking at the modules from which I have removed the imports and fixing up the things that don't work, which should just be changing some of the function calls to things that are still current.  This may be a slow process, and I am under no impression it will be easy.  Especially on this stupid tiny laptop.

corinnebosley:
Making a clean start with this one in a fresh PR.

PR #2790, opened 2017-10-12
Rename pp_rules to pp_load_rules

dkillick:
In preparation to convert PP save rules from a text file to Python, this PR renames `pp_rules` (that contains rules for loading PP files) to `pp_load_rules` (more descriptive, less likely to be confusing in the future).

PR #2795, opened 2017-10-13
Translate text PP save rules into Python

dkillick:
All the PP save rules were supplied by a text file `lib/iris/etc/pp_save_rules.txt` but the text rule processing classes in `lib/iris/fileformats/rules.py` are deprecated. This PR, therefore, takes the text rules and translates them into a new Python module `lib/iris/fileformats/pp_save_rules.py` and makes the necessary changes to PP save so that the new save rules are run. 

The old text rules have been updated and a very few test changes have been made – but, of course, this job will have been done correctly if the tests all pass as-is with the exception of the odd API change.


lib/iris/tests/unit/fileformats/pp/test_save.py
```diff
@@ -162,26 +163,42 @@ def test_um_version(self):
...
     def setUp(self):
         self.cube = stock.realistic_3d()
+        self.pp_field = mock.MagicMock(spec=pp.PPField3)
+        self.pp_field.HEADER_DEFN = pp.PPField3.HEADER_DEFN
+        self.patcher = mock.patch('iris.fileformats.pp.PPField3',
+                                  return_value=self.pp_field)
+        self.patcher.start()
+
+    def tearDown(self):
+        self.patcher.stop()
```
bjlittle: @dkillick Might want to take a peek at `iris.tests.patch` ...

lib/iris/fileformats/pp.py
```diff
@@ -221,31 +221,6 @@
                        'default': np.dtype('>f4'),
                        }
 
-# LBPROC codes and their English equivalents
```
pp-mo: I think moving these to iris.fileformats.rules is a mistake, as it the information is specific to PP (and FF), whereas iris.fileformats.rules "ought" to be the home for fileformat-independent rules processing.
They aren't specially tied to implementation of text rules anyway.

In any case, "LBPROC_PAIRS" and "lbproc_map" are perfectly reasonable public elements of the API, so you ***can't just move them*** anyway.

In an *ideal* world I think these would be in "iris.fileformats.um_cf_map", and also "lbroc_map" should be capitalised, but it's a bit late for that...

pp-mo: (from discussion with @dkillick)
The move was introduced because of a circularity problem when `iris.fileformats.pp` imports `iris.fileformats.pp_save_rules`, which imports `iris.fileformats.pp` ...
By putting 'lbproc_map' into `rules`, then `pp_save_rules` no longer needs to import `pp`.

New suggestion: Put LBPROC_PAIRS and `lbproc_map` inside pp_save_rules itself.
Then to avoid breaking existing API, publish them in 'pp' by importing them from 'pp_save_rules'.

pp-mo: More trouble...
I missed that "lbproc_map" is also used in "pp_rules" 
( i.e. "pp_load_rules", perhaps it should now be called that ? )

You can import it there from "pp_save_rules", but that is obviously a rather peculiar dependency ordering.
You cannot import it from 'pp', if 'pp' is to import 'pp_rules' (same circularity problem as before)

In my view, this info ***really*** belongs with the other similar stuff in 'um_cf_map'.
To do this *properly*, we should really put this info into metarelate, and modify`tools/gen_translations.py` to support it, but I really don't think we have time for that.
In which case, we should just host this translation info in a separate module.
See : https://github.com/dkillick/iris/pull/12 for worked suggestion ...

dkillick: > i.e. "pp_load_rules", perhaps it should now be called that ? 

See #2790.

>  we should just host this translation info in a separate module.

Agreed 👍 Thanks for spotting this, @pp-mo

LukeC92:
On [line 64 of pp.py](https://github.com/dkillick/iris/blob/6f8c2331a54fb56d3b1a08b0c4b5aaf1d75f781b/lib/iris/fileformats/pp.py#L64), _save_rules is set to none. The functionality of _save_rules was removed in a previous pull request and with the changes made in this PR it is now obsolete. It would probably be best to remove lines 64 (and maybe 63) now. Otherwise I see no issues with this PR and would approve it.

corinnebosley:
Finally!! Pats on the back all round for that one.

PR #2811, opened 2017-10-18
[DOC] whatsnew entry for experimental/ff removal

dkillick:


dkillick:
I've added "`* `" to the start of the whatsnew entry so that it will eventually render as an element in a bulleted list. I've also taken the liberty of updating the only other whatsnew entry currently in the whatsnew directory to reflect this change.

docs/iris/src/whatsnew/contributions_v2.0.0/incompatiblechange_2017-Oct-17_rename-pp-rules.txt
```diff
@@ -1,2 +1,2 @@
-`iris.fileformats.pp_rules` has been renamed to `iris.fileformats.pp_load_rules`.
-This has been done for the sake of clarity following the introduction of `iris.fileformats.pp_save_rules`.
\ No newline at end of file
+* ```iris.fileformats.pp_rules``` has been renamed to :mod:`iris.fileformats.pp_load_rules`.
```
djkirkham: For my own understanding, what's the difference between single and triple back-ticks here?

corinnebosley: @djkirkham I spoke to Pete about that earlier.  In most what's-news, we would use single back-ticks along with colons around the type of thing we have modified (like :func: or :mod:), and this would get rendered as a link to that thing.

In this case though, some of the things we are referring to have been removed, so we are using triple backticks instead to render them as code blocks.

PR #2844, opened 2017-10-23
Nc simple chunks

pp-mo:
This is a simple solution, supposed to address the problems in #2840

As the change is general in 'as_lazy_data' it can in principle apply elsewhere,
but in practice no single data fields ever get that big, so it only kicks in for large NetCDF variables.

~**Do not merge** until we can confirm that it definitely solves the known problem in  #2840 .~
#2839 tests the fix -- looking good.

### DIFF
--- a/lib/iris/_lazy_data.py
+++ b/lib/iris/_lazy_data.py
@@ -23,6 +23,8 @@ To avoid replicating implementation-dependent test and conversion code.
 from __future__ import (absolute_import, division, print_function)
 from six.moves import (filter, input, map, range, zip)  # noqa
 
+import atexit
+
 import dask
 import dask.array as da
 import dask.context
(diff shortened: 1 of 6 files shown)


lib/iris/_lazy_data.py
```diff
@@ -72,7 +72,24 @@ def is_lazy_data(data):
...
-def as_lazy_data(data, chunks=_MAX_CHUNK_SIZE, asarray=False):
+def _limited_shape(shape):
+    # Reduce a shape to less than a default overall number-of-points, reducing
+    # earlier dimensions preferentially.
+    # Note: this is only a heuristic, assuming that earlier dimensions are
+    # 'outer' storage dimensions -- not *always* true, even for NetCDF data.
+    shape = list(shape)
+    i_reduce = 0
+    while np.prod(shape) > _MAX_CHUNK_SIZE:
+        factor = np.ceil(np.prod(shape) / float(_MAX_CHUNK_SIZE))
```
djkirkham: You can remove the float cast since we have `from __future__ import division`.

lib/iris/_lazy_data.py
```diff
@@ -72,7 +72,24 @@ def is_lazy_data(data):
...
+        new_dim = int(np.floor(shape[i_reduce] / factor))
```
djkirkham: You can remove the `np.floor` since casting to `int` floors the argument.

PR #2849, opened 2017-10-24
Move whatsnew entries to best location

dkillick:
Raised in followup to #2700. In there, the whatsnew entry had been added directly to the compiled what's new rst document, meaning they had missed the whatsnew builder process, and were being recorded in the whatsnew for a release within which the changes they pointed to were not actually released (!). 

Rather than hold up #2700 any further, this PR removes them from the Iris v2.0a0 what's new document and adds them verbatim to a new whatsnew entry for the Iris v2.0 full release.


dkillick:
Thanks @pp-mo!

PR #2858, opened 2017-10-25
Deprecating Futures clip_latitudes

### LINKED ISSUE #2846: Change default behaviour of FUTURE option `clip_latitudes`
This involves completing the following work:

* [ ] set the default value of the FUTURE flag `clip_latitudes` to `True` in `iris.Future`.
* [ ] raise a deprecation warning iff a user sets a value for this FUTURE flag (that is, raise a warning if the flag is set by the user, but *don't* raise a warning when the default behaviour is set up by Iris).
* [ ] deprecate the old behaviour in `iris.coords.Coord()._guess_bounds()` and raise a deprecation warning if a user tries to use the old behaviour.
* [ ] add a deprecation note to the docstrings for `iris.coords.Coord().guess_bounds()` and `iris.coords.Coord()._guess_bounds()`.
* [ ] change all tests to expect the new default behaviour unless the test is directly testing the old behaviour.
* [ ] Add a deprecation directive to the description of `clip_latitudes` in the docstring for `iris.Future`.
* [ ] Update all doc examples to use the new default behaviour (there may not be any).
* [ ] Add whatsnew entry.

LukeC92:
This Closes #2846. I'm uncertain about some of my changes and fully expect there'll be things I'll need to alter.

### DIFF
--- a/lib/iris/__init__.py
+++ b/lib/iris/__init__.py
@@ -211,7 +211,7 @@ class Future(threading.local):
 
     deprecated_options = {'cell_datetime_objects': 'warning',
                           'netcdf_promote': 'error',
-		          'clip_latitudes': 'warning'}
+                          'clip_latitudes': 'warning'}
 
     def __setattr__(self, name, value):
         if name in self.deprecated_options:
--- a/lib/iris/__init__.py
+++ b/lib/iris/__init__.py
@@ -190,8 +190,8 @@ class Future(threading.local):
         .. deprecated:: 2.0.0
 
             The option `clip_latitudes` is deprecated and will be removed in a
-            future release and the deprecated code paths this option used to
-            toggle have been removed.
+            future release. `clip_latitudes` is set to True by default and
+            should not be altered.
 
             The option `clip_latitudes` controlled whether the
             :meth:`iris.coords.Coord.guess_bounds()` method would limit the
--- a/lib/iris/__init__.py
+++ b/lib/iris/__init__.py
@@ -187,9 +187,15 @@ class Future(threading.local):
         unlimited.  The current default is that the leading dimension is
         unlimited unless otherwise specified.
 
-        The option `clip_latitudes` controls whether the
-        :meth:`iris.coords.Coord.guess_bounds()` method limits the
-        guessed bounds to [-90, 90] for latitudes.
+        .. deprecated:: 2.0.0
+
+            The option `clip_latitudes` is deprecated and will be removed in a
+            future release and the deprecated code paths this option used to
+            toggle have been removed.
+
+            The option `clip_latitudes` controlled whether the
+            :meth:`iris.coords.Coord.guess_bounds()` method would limit the
+            guessed bounds to [-90, 90] for latitudes.
 
         """
         self.__dict__['cell_datetime_objects'] = cell_datetime_objects
--- a/lib/iris/__init__.py
+++ b/lib/iris/__init__.py
@@ -144,7 +144,7 @@ class Future(threading.local):
     """Run-time configuration controller."""
 
     def __init__(self, cell_datetime_objects=True, netcdf_promote=True,
-                 netcdf_no_unlimited=False, clip_latitudes=False):
+                 netcdf_no_unlimited=False, clip_latitudes=True):
         """
         A container for run-time options controls.
 
@@ -204,7 +204,8 @@ class Future(threading.local):
                           self.netcdf_no_unlimited, self.clip_latitudes)
 
     deprecated_options = {'cell_datetime_objects': 'warning',
-                          'netcdf_promote': 'error'}
+                          'netcdf_promote': 'error',
+		          'clip_latitudes': 'warning'}
 
     def __setattr__(self, name, value):
         if name in self.deprecated_options:
--- a/lib/iris/coords.py
+++ b/lib/iris/coords.py
@@ -1234,9 +1234,11 @@ class Coord(six.with_metaclass(ABCMeta, CFVariableMixin)):
             - all the points are in the range [-90, 90].
 
         .. deprecated:: 2.0.0
-            The 'iris.FUTURE.clip_latitudes' option is now deprecated
+
+            The `iris.FUTURE.clip_latitudes` option is now deprecated
             and is set to True by default. Please remove code which
-            uses 'iris.FUTURE.clip_latitudes'.
+            relies on coordinate bounds being outside the range
+            [-90, 90].
 
         """
         # XXX Consider moving into DimCoord
@@ -1279,7 +1281,6 @@ class Coord(six.with_metaclass(ABCMeta, CFVariableMixin)):
                 points = self.points
                 if (points >= -90).all() and (points <= 90).all():
                     np.clip(bounds, -90, 90, out=bounds)
-
         else:
             wmsg = ("guessing latitude bounds outside of [-90, 90] is "
                     "deprecated behaviour. "
@@ -1329,9 +1330,11 @@ class Coord(six.with_metaclass(ABCMeta, CFVariableMixin)):
             - all the points are in the range [-90, 90].
 
         .. deprecated:: 2.0.0
-            The 'iris.FUTURE.clip_latitudes' option is now deprecated
+
+            The `iris.FUTURE.clip_latitudes` option is now deprecated
             and is set to True by default. Please remove code which
-            uses 'iris.FUTURE.clip_latitudes'.
+            relies on coordinate bounds being outside the range
+            [-90, 90].
 
         """
         self.bounds = self._guess_bounds(bound_position)
--- a/lib/iris/coords.py
+++ b/lib/iris/coords.py
@@ -1233,6 +1233,11 @@ class Coord(six.with_metaclass(ABCMeta, CFVariableMixin)):
             - the units are degrees,
             - all the points are in the range [-90, 90].
 
+        .. deprecated:: 2.0.0
+            The 'iris.FUTURE.clip_latitudes' option is now deprecated 
+            and is set to True by default. Please remove code which
+            uses 'iris.FUTURE.clip_latitudes'.
+
         """
         # XXX Consider moving into DimCoord
         # ensure we have monotonic points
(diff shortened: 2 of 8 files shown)


lib/iris/__init__.py
```diff
@@ -204,7 +210,8 @@ def __repr__(self):
                           self.netcdf_no_unlimited, self.clip_latitudes)
 
     deprecated_options = {'cell_datetime_objects': 'warning',
-                          'netcdf_promote': 'error'}
+                          'netcdf_promote': 'error',
+		          'clip_latitudes': 'warning'}
```
djkirkham: This should be indented by one more space to line up with the line above

djkirkham:
I think I'm out of the loop here, but is this the correct procedure? Shouldn't we be retaining the default of false but issuing a warning that in future it will be true by default?

lib/iris/coords.py
```diff
@@ -1268,12 +1273,19 @@ def _guess_bounds(self, bound_position=0.5):
...
-            points = self.points
-            if (points >= -90).all() and (points <= 90).all():
-                np.clip(bounds, -90, 90, out=bounds)
+        if iris.FUTURE.clip_latitudes:
+            if (self.name() in ('latitude', 'grid_latitude') and
+                    self.units == 'degree'):
+                points = self.points
+                if (points >= -90).all() and (points <= 90).all():
+                    np.clip(bounds, -90, 90, out=bounds)
+
```
dkillick: Can you remove this blank line please.

lib/iris/coords.py
```diff
@@ -1233,6 +1233,11 @@ def _guess_bounds(self, bound_position=0.5):
             - the units are degrees,
             - all the points are in the range [-90, 90].
 
+        .. deprecated:: 2.0.0
```
dkillick: You need one blank line after this directive, as per the usage of other directives in this docstring (and elsewhere).

lib/iris/__init__.py
```diff
@@ -187,9 +187,15 @@ def __init__(self, cell_datetime_objects=True, netcdf_promote=True,
...
         unlimited unless otherwise specified.
 
-        The option `clip_latitudes` controls whether the
-        :meth:`iris.coords.Coord.guess_bounds()` method limits the
-        guessed bounds to [-90, 90] for latitudes.
+        .. deprecated:: 2.0.0
+
+            The option `clip_latitudes` is deprecated and will be removed in a
+            future release and the deprecated code paths this option used to
+            toggle have been removed.
```
djkirkham: This isn't true; it's still possible to toggle the different code paths, and neither of them were deprecated anyway.

lib/iris/tests/unit/coords/test_Coord.py
```diff
@@ -169,12 +169,12 @@ def test_all_inside(self):
...
         lat = DimCoord([-80, 0, 70], units='degree', standard_name='latitude')
         lat.guess_bounds()
-        self.assertArrayEqual(lat.bounds, [[-120, -40], [-40, 35], [35, 105]])
+        self.assertArrayEqual(lat.bounds, [[-90, -40], [-40, 35], [35, 90]])
 
     def test_points_to_edges_bounds_outside(self):
         lat = DimCoord([-90, 0, 90], units='degree', standard_name='latitude')
         lat.guess_bounds()
-        self.assertArrayEqual(lat.bounds, [[-135, -45], [-45, 45], [45, 135]])
+        self.assertArrayEqual(lat.bounds, [[-90, -45], [-45, 45], [45, 90]])
```
djkirkham: This test and the one above are no identical to their namesakes in `Test_guess_bounds__enabled_latitude_clipping` and can be removed.

djkirkham: Although, perhaps we can to keep the tests but toggle the flag the other way?

lib/iris/__init__.py
```diff
@@ -187,9 +187,15 @@ def __init__(self, cell_datetime_objects=True, netcdf_promote=True,
...
+
+            The option `clip_latitudes` controlled whether the
+            :meth:`iris.coords.Coord.guess_bounds()` method would limit the
+            guessed bounds to [-90, 90] for latitudes.
```
djkirkham: Given that the functionality is still available, this should probably be in the present tense,

lib/iris/tests/unit/test_Future.py
```diff
@@ -39,6 +39,9 @@ def test_valid_netcdf_no_unlimited(self):
     def test_valid_clip_latitudes(self):
         future = Future()
         new_value = not future.clip_latitudes
+        with warnings.catch_warnings(record=True) as warn:
+            warnings.simplefilter('always')
+            future.clip_latitudes = new_value
         future.clip_latitudes = new_value
```
djkirkham: You've got two identical assignments here.

lib/iris/coords.py
```diff
@@ -1233,6 +1233,11 @@ def _guess_bounds(self, bound_position=0.5):
...
+            The 'iris.FUTURE.clip_latitudes' option is now deprecated
+            and is set to True by default. Please remove code which
+            uses 'iris.FUTURE.clip_latitudes'.
```
dkillick: This deprecation warning needs to focus on the behaviour change in this method, not on the FUTURE flag.

The reference to the FUTURE needs to be surrounded with backticks \` and not apostrophes.

lib/iris/coords.py
```diff
@@ -1316,6 +1328,11 @@ def guess_bounds(self, bound_position=0.5):
...
```
dkillick: The same comments apply to this addition to the docstring as well.

lib/iris/__init__.py
dkillick: However use of the functionality is deprecated and should not be encouraged, so I reckon we should use the past tense, (to be clear: I started the trend elsewhere in this docstring of using the past tense).

lib/iris/tests/unit/coords/test_Coord.py
dkillick: Question is, do we want to test deprecated behaviour?

lib/iris/__init__.py
djkirkham: I noticed your comment elsewhere (with reference to cell datetimes, I think), but I thought in that case the behaviour had actually been removed?

lib/iris/tests/unit/coords/test_Coord.py
dkillick: If we did test the deprecated behaviour we'd need to capture the deprecation warning that would also be raised. Speaking from the convention we've set up when updating the other FUTURE flags, we've tended to update tests to test the new behaviour if the test is still necessary, and remove the test entirely if not.

djkirkham: I don't know, but one argument in favour of testing it is this: the reason it's deprecated rather than simply removed is so that users can still use it to give them a chance to change their code. If we don't test it and it breaks, we take away that grace period.

lib/iris/tests/unit/test_Future.py
```diff
@@ -39,6 +39,9 @@ def test_valid_netcdf_no_unlimited(self):
     def test_valid_clip_latitudes(self):
         future = Future()
         new_value = not future.clip_latitudes
+        with warnings.catch_warnings(record=True) as warn:
+            warnings.simplefilter('always')
+            future.clip_latitudes = new_value
```
dkillick: This test also needs to ensure the warning that's been captured is the correct warning. The code to do this can be found in various places throughout the tests, notably in the `cell_datetime_objects` change (see [L61-69](https://github.com/LukeC92/iris/blob/f3f75550ad7994f0a24552e93122298d8633fef7/lib/iris/tests/unit/test_Future.py#L61-L69)).

lib/iris/tests/experimental/regrid/test_regrid_area_weighted_rectilinear_src_and_grid.py
```diff
@@ -473,7 +473,7 @@ def test_global_data_reduce_res(self):
         src.coord('longitude').guess_bounds()
         dest = _resampled_grid(src, 0.4, 0.3)
         res = regrid_area_weighted(src, dest)
-        self.assertArrayShapeStats(res, (21, 38), 280.137698, 15.998966)
+        self.assertArrayShapeStats(res, (21, 38), 280.484932, 15.831545)
```
dkillick: What caused these value changes?

lib/iris/tests/experimental/regrid/test_regrid_area_weighted_rectilinear_src_and_grid.py
```diff
@@ -482,7 +482,7 @@ def test_global_data_increase_res(self):
         src.coord('longitude').guess_bounds()
         dest = _resampled_grid(src, 1.5, 1.5)
         res = regrid_area_weighted(src, dest)
-        self.assertArrayShapeStats(res, (109, 144), 279.955722, 16.313940)
+        self.assertArrayShapeStats(res, (109, 144), 280.349625, 16.073397)
```
dkillick: Again.

lib/iris/tests/unit/test_Future.py
djkirkham: I wrote a context manager for this, which @pp-mo put in the `iris.tests` module: https://github.com/SciTools/iris/blob/0ce7d5b0117856f9d6e3a2746dae364bad453a80/lib/iris/tests/__init__.py#L542

lib/iris/tests/experimental/regrid/test_regrid_area_weighted_rectilinear_src_and_grid.py
djkirkham: It's because the bounds at the edges of the source grid are constrained due to the clipping, which changes the areas.

lib/iris/tests/unit/test_Future.py
dkillick: @djkirkham that's rather nice 👍 (as is the shiny new previewer there! ⬆️). @LukeC92 feel free to use that context manager here.

PR #2864, opened 2017-10-26
Fix license headers for um files

djkirkham:
Fix the license headers currently causing the coding standards tests to fail.


dkillick:
Erroring tests are all timeouts... I'm merging this regardless.

PR #2866, opened 2017-10-26
Allow missing_value as a user set attribute

### LINKED ISSUE #1588: Allow users to set `missing_value` attribute
Currently users are prevented from setting an attribute named `missing_value` on their cubes. However, since Iris will never automatically create one, this means there is no way to save a cube to a netCDF file and include a `missing_value` attribute.

For indicating missing values iris will always save to netCDF with a `_FillValue` attribute. The CF conventions recommend that if both `_FillValue` and `missing_value` attributes are used then they should have the same value (sensible), but do not prohibit them from being different. Therefore I propose that `missing_value` becomes an attribute that is allowed to be set on cubes, since the effects of doing so do not conflict with CF conventions.

zklaus:
This allows the user to set missing_value as an attribute on a
cube while ensuring that it is saved as local attribute in the
netcdf file.

Fixes #1588

Signed-off-by: Klaus Zimmermann <klaus.zimmermann@smhi.se>

### DIFF
--- a/lib/iris/_cube_coord_common.py
+++ b/lib/iris/_cube_coord_common.py
@@ -1,4 +1,4 @@
-# (C) British Crown Copyright 2010 - 2016, Met Office
+# (C) British Crown Copyright 2010 - 2017, Met Office
 #
 # This file is part of Iris.
 #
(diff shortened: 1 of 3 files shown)


ajdawson:
Thanks @zklaus. There are some test failures to deal with. For the license tests you just need to change `2016` to `2017` on the first line of `_cube_coord_common.py`. I'm not sure why the doctests fail, doesn't look like your fault.

It would be nice to have this new functionality tested too, do you have an idea of how you might do that?

zklaus:
Fixed the license issue. I'll add a test while waiting for the answer to my cla.

QuLogic:
Is there any point to targeting the `v1.13.x` branch right now?

ajdawson:
No, this should be targeted on master.

zklaus:
I created pr #2892 with the same intention, but based on master.

PR #2881, opened 2017-10-27
Change behaviour of 'error' level

djkirkham:
Changed the behaviour of the 'error' level for `iris.FUTURE` options so that if a user sets them to `True` a warning is raised rather than an error.

(As an aside, I'm only able to test this behaviour because we currently have an option which is at 'error' level.)

### DIFF
--- a/lib/iris/__init__.py
+++ b/lib/iris/__init__.py
@@ -224,7 +224,7 @@ class Future(threading.local):
     def __setattr__(self, name, value):
         if name in self.deprecated_options:
             level = self.deprecated_options[name]
-            if level == 'error':
+            if level == 'error' and not value:
                 emsg = ("setting the 'Future' property {prop!r} has been "
                         "deprecated to be removed in a future release, and "
                         "deprecated {prop!r} behaviour has been removed. "
--- a/lib/iris/tests/unit/test_Future.py
+++ b/lib/iris/tests/unit/test_Future.py
@@ -57,15 +57,16 @@ class Test___setattr__(tests.IrisTest):
 
     def test_netcdf_no_unlimited(self):
         future = Future()
-        exp_emsg = "'Future' property 'netcdf_promote' is deprecated"
+        exp_emsg = "'Future' property 'netcdf_no_unlimited' is deprecated"
         with self.assertWarnsRegexp(exp_emsg):
-            future.netcdf_promote = True
+            future.netcdf_no_unlimited = True
 
     def test_invalid_netcdf_no_unlimited(self):
         future = Future()
-        exp_emsg = "'Future' property 'netcdf_promote' has been deprecated"
+        exp_emsg = \
+            "'Future' property 'netcdf_no_unlimited' has been deprecated"
         with self.assertRaisesRegexp(AttributeError, exp_emsg):
-            future.netcdf_promote = False
+            future.netcdf_no_unlimited = False
 
     def test_cell_datetime_objects(self):
         future = Future()
--- a/lib/iris/tests/unit/test_Future.py
+++ b/lib/iris/tests/unit/test_Future.py
@@ -43,13 +43,17 @@ class Test___setattr__(tests.IrisTest):
         with self.assertRaises(AttributeError):
             future.numberwang = 7
 
-    def test_invalid_netcdf_promote_attributes(self):
+    def test_netcdf_promote(self):
         future = Future()
-        states = [True, False]
-        exp_emsg = "deprecated 'netcdf_promote' behaviour has been removed"
-        for state in states:
-            with self.assertRaisesRegexp(AttributeError, exp_emsg):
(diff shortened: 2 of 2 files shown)


pelson:
```

ERROR: test_netcdf_promote (iris.tests.unit.test_Future.Test___setattr__)

  File "/home/travis/miniconda/envs/test-environment/lib/python2.7/site-packages/Iris-2.0a0-py2.7.egg/iris/tests/unit/test_Future.py", line 55, in test_netcdf_promote

    with self.assertGivesWarning(exp_emsg):

AttributeError: 'Test___setattr__' object has no attribute 'assertGivesWarning'
```

pelson:
Getting a:

```
FAIL: test_invalid_netcdf_no_unlimited_attributes (iris.tests.unit.test_Future.Test___setattr__)
  File "/home/travis/miniconda/envs/test-environment/lib/python2.7/site-packages/Iris-2.0a0-py2.7.egg/iris/tests/unit/test_Future.py", line 65, in test_invalid_netcdf_no_unlimited_attributes
    future.netcdf_no_unlimited = state
AssertionError: AttributeError not raised
```

lib/iris/__init__.py
```diff
@@ -224,7 +224,7 @@ def __repr__(self):
     def __setattr__(self, name, value):
         if name in self.deprecated_options:
             level = self.deprecated_options[name]
-            if level == 'error':
+            if level == 'error' and not value:
```
dkillick: I don't understand the purpose of this change. If `__setattr__` has been called, there will **always** be a `value`:

```python
>>> import iris
>>> # Specifically enable deprecated behaviour.
>>> iris.FUTURE.cell_datetime_objects = False  # key=cell_datetime_objects; value=False
DeprecationWarning...
>>>
```

This is why the test is failing, as it is testing precisely this behaviour, is being passed a value and is thus failing as this `if` block evaluates to False.

lib/iris/tests/unit/test_Future.py
```diff
@@ -43,13 +43,17 @@ def test_invalid_attribute(self):
...
-                future.netcdf_promote = state
+        exp_emsg = "'Future' property 'netcdf_promote' is deprecated"
+        with self.assertWarnsRegexp(exp_emsg):
+            future.netcdf_promote = True
+
+    def test_invalid_netcdf_promote(self):
+        future = Future()
+        exp_emsg = "'Future' property 'netcdf_promote' has been deprecated"
+        with self.assertRaisesRegexp(AttributeError, exp_emsg):
+            future.netcdf_promote = False
```
dkillick: These tests are now wrong. The `netcdf_promote` FUTURE has been removed, so if a user tries to set either `True` and `False` for this flag an error should be raised. Please reset these tests to their original behaviour.

lib/iris/__init__.py
djkirkham: `not False == True`...

djkirkham:
@dkillick Please see https://github.com/SciTools/iris/pull/2838#discussion_r147371426 for reasoning behind this PR.

lib/iris/__init__.py
djkirkham: I should have said `not True == False` which is the case when we don't want this `if` block to run.

dkillick:
@djkirkham thanks for the context. Problem is, the suggestion you're working to with this disagrees with the plan for Iris v2.0 that @bjlittle and @pelson put together, which we worked to when dealing with the updates to the FUTURE flags...

djkirkham:
Well, maybe it was a bad plan ¯\_(ツ)_/¯

dkillick:
@djkirkham it seems quite reasonable to me :wink:

lib/iris/__init__.py
dkillick: Of course it is. That always catches me out.

lib/iris/tests/unit/test_Future.py
```diff
@@ -43,22 +43,29 @@ def test_invalid_attribute(self):
...
-    def test_invalid_netcdf_no_unlimited_attributes(self):
+    def test_invalid_netcdf_promote(self):
         future = Future()
-        states = [True, False]
-        exp_emsg =\
-            "deprecated 'netcdf_no_unlimited' behaviour has been removed"
-        for state in states:
-            with self.assertRaisesRegexp(AttributeError, exp_emsg):
-                future.netcdf_no_unlimited = state
+        exp_emsg = "'Future' property 'netcdf_promote' has been deprecated"
```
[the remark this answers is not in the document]
djkirkham: This one's fine, well spotted on the other two though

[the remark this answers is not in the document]
dkillick: Copy-paste mistakes are my favourite mistakes to make :wink:

PR #2883, opened 2017-10-27
Fix doctest failing due to change in 'netcdf_promote' future behaviour

djkirkham:
Closes https://github.com/SciTools/iris/issues/2880

The new behaviour causes the reference surface to be loaded as a separate cube, so the `load_cube()` call fails.


djkirkham:
@LukeC92

djkirkham:
Closes https://github.com/SciTools/iris/issues/2880

LukeC92:
Based off of what you said to me in person these code changes make perfect sense to me.

PR #2886, opened 2017-10-27
Rename file to fit categories

djkirkham:
A What's New file was mistakenly added with a file name beginning with 'enhancement'. Since this doesn't fit into any of the categories, it causes a test to fail.

The change in question is a speed up, and doesn't really fit into any of the categories. I've renamed the file to begin with 'bugfix', which is obviously wrong.


djkirkham:
Superceded by https://github.com/SciTools/iris/pull/2884

PR #2915, opened 2017-11-13
Remove obsolete reference from Userguide.

pp-mo:
See reference here : https://github.com/SciTools/iris/pull/1911#issuecomment-215032215

The `iris.analysis.maths.add() / subtract()` routines used to have a keyword `ignore=True`, but this has been deprecated and non-functional since (allegedly) v0.8 (!)
The keyword is finally removed in v2.0 anyway.

This is what the Userguide reference is referring to, so I just removed it.


pp-mo:
Ok, re-targetted.
Thanks @pelson

pp-mo:
> You're also going to need to rebase
Thanks @pelson 
This ok now ?

PR #2918, opened 2017-11-15
Add _repr_html_ for Iris cubes

dkillick:
Add functionality to Iris cubes to produce a pretty, collapsible html representation of the cube with a jupyter notebook.

For example:

screen shot 2017-11-15 at 15 27 07

### DIFF
--- a/docs/iris/src/whatsnew/contributions_2.1/newfeature_2018-May-08_repr-html.txt
+++ b/docs/iris/src/whatsnew/contributions_2.1/newfeature_2018-May-08_repr-html.txt
@@ -1,0 +1,2 @@
+* Added ``repr_html`` functionality to the :class:`~iris.cube.Cube` to provide
+  a rich html representation of cubes in Jupyter notebooks.
--- a/lib/iris/experimental/representation.py
+++ b/lib/iris/experimental/representation.py
@@ -20,6 +20,9 @@ Definitions of how Iris objects should be represented.
 
 """
 
+from __future__ import (absolute_import, division, print_function)
+from six.moves import (filter, input, map, range, zip)  # noqa
+
 
 class CubeRepresentation(object):
     """
--- a/lib/iris/experimental/representation.py
+++ b/lib/iris/experimental/representation.py
@@ -98,7 +98,6 @@ class CubeRepresentation(object):
         self.cube_id = id(self.cube)
         self.cube_str = str(self.cube)
 
-        self.summary = self.cube.summary(True)
         self.str_headings = {
             'Dimension coordinates:': None,
             'Auxiliary coordinates:': None,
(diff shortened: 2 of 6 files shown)


dkillick:
Unused imports removed – thanks, @stickler-ci...

niallrobinson:
just had a go with this - its well cool. Top bang for buck

niallrobinson:
what's the deal with this? - the xarray guys in the pub said they've been wanting to do this and i said "OH IRIS ALREADY HAS IT!!" but it's just languishing here. I think it's a really easy win which people would find cooler than maybe you'd expect/they should

pelson:
Fully supportive of the concept.

@dkillick - my two requests would be: have the tests passing; separate the functionality out of cube.py (with ``_repr_html_`` being a simple import and wrapper).

Longer term, there is so much more that can be done with this, but it is a great start. 👍

dkillick:
This is quite a big change, moving from using `div` tags to lay out the `_repr_html_` content to putting it all in an html table. However, the table looks better and is far more compact, although (at the moment) we've lost the ability to dynamically expand and collapse the content. I've updated the tests to match the new and updated code.

Here's how the same cube from the description above looks now:

screen shot 2018-05-08 at 15 01 01

docs/iris/src/whatsnew/contributions_2.1/newfeature_2018-May-08_repr-html.txt
```diff
@@ -0,0 +1,2 @@
+* Added ``repr_html`` functionality to the :class:`~iris.cube.Cube` to provide
```
pelson: We will want this to have a screenshot when it comes to cutting the release.

pelson:
I haven't actually gone through the code yet, but found the following issue with a basic cube:

screen shot 2018-05-09 at 06 50 14

pelson:
The screenshot I added highlights my only major concern about the implementation: the ```_html_repr_``` is using existing reprs and parsing them. In terms of approach, I'd be more comfortable keying into the cube's internals than I would manipulating the cube's existing str representations. @dkillick - did you take a look at this option before, and if so, was there an issue that cropped up that made it harder than it needed to be?

docs/iris/src/whatsnew/contributions_2.1/newfeature_2018-May-08_repr-html.txt
dkillick: We could use the image above?

dkillick:
@pelson thanks for the bug report, you've evidently found a gap in the tests 😉 

> the `_html_repr_` is using existing reprs and parsing them

That it is, and the reason it does so is for simplicity: all the content we need to show is already contained within the existing repr. Keying into a cube's internals is _definitely_ a better way to produce the `_html_repr_`, but every time I thought about doing so I realised that I would literally just rewrite `cube.summary` in the context of the `_html_repr_`.

As such, a much better (and longer-term) approach here would probably be to make a generic representer object that can then be used by to construct the various different representations we want to make of a cube (such as `__repr__`, `__str__` and `_html_repr_`). This would sort out your quite pertinent comment on the current implementation; it's just that such an implementation would be quite a lot more work so didn't quite make the cut as an / proof-of-concept implementation.

lib/iris/experimental/representation.py
```diff
@@ -0,0 +1,308 @@
...
+      font-family: monaco, monospace;
+  }}
+  th.iris {{
+      background: #303f3f;
+      color: #e0e0e0;
+      border-left: 1px solid;
+      border-color: #9c9c9c;
+      font-size: 1.05em;
+      min-width: 50px;
+      max-width: 125px;
```
pelson: Can we increase this, or even remove altogether? The effect of this width is that we can get some really ugly column rendering (e.g. ```model_level_number``)

lib/iris/experimental/representation.py
```diff
@@ -0,0 +1,308 @@
...
+        # Retrieve info for each heading from the printout.
+        for i0, i1 in zip(start_inds[:-1], start_inds[1:]):
+            str_heading_name = bits[i0].strip()
+            if i1 != 0:
+                content = bits[i0 + 1: i1]
+            else:
+                content = bits[i0 + 1:]
+            self.str_headings[str_heading_name] = content
+
+    def _make_header(self):
```
pelson: Would be great to have column highlighting too. See https://stackoverflow.com/questions/16967366/how-do-i-highlight-an-html-table-column-with-bootstrap.

Note: not in this PR.

lib/iris/experimental/representation.py
```diff
@@ -0,0 +1,308 @@
...
+    This includes:
+
+    * ``_html_repr_``: a representation of the cube as an html object,
+      available in Jupyter notebooks. Specifically, this is presented as an
+      html table.
+
+    """
+
+    _template = """
+<style>
```
pelson: Ideally we wouldn't re-define the style each time we do a repr. I think there is (should be) a hook for this somewhere, but I did a similar hack for the matplotlib notebook backend and it does work effectively.

pelson:
Awesome. Thanks @dkillick. This will go into v2.1

PR #2924, opened 2017-12-15
Reduce exception message for loading missing files.

rcomer:
See #2032.

I liked @pelson's suggestion of a tick and a cross at the start of each line, but comments on [this stackoverflow question](https://stackoverflow.com/questions/16676101) suggest doing this is terminal dependent (I couldn't get the suggestions to work under my ipython anyway).  So I went for a more basic - or * bullet point difference.


rcomer:
I am confused by the conflict.  Perhaps my `git` is pointing to the wrong repository?

rcomer:
Never mind.  I think I did a stupid with my `git branch` command.

PR #2926, opened 2017-12-18
Reverse 2878

djkirkham:
A workaround for https://github.com/dask/dask/issues/2823 was added in https://github.com/SciTools/iris/pull/2878. The dask issue is fixed in the latest release, so this PR reverts the changes.

N.B. There is a test which will fail on master now that the Dask issue is fixed, so this needs resolving ASAP.

### DIFF
--- a/lib/iris/_lazy_data.py
+++ b/lib/iris/_lazy_data.py
@@ -65,15 +65,6 @@ def _limited_shape(shape):
     return tuple(shape)
 
 
-def _getall(a):
-    res = a[()]
-    if isinstance(res, ma.core.MaskedConstant):
-        res = ma.masked_array(res.data, mask=res.mask)
-    return res
-
-_getall_delayed = dask.delayed(_getall)
-
-
 def as_lazy_data(data, chunks=None, asarray=False):
     """
     Convert the input array `data` to a dask array.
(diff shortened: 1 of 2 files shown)


lib/iris/_lazy_data.py
```diff
@@ -104,15 +95,10 @@ def as_lazy_data(data, chunks=None, asarray=False):
         # but reduce it if larger than a default maximum size.
         chunks = _limited_shape(data.shape)
 
+    if isinstance(data, ma.core.MaskedConstant):
+        data = ma.masked_array(data.data, mask=data.mask)
```
pp-mo: Is this related, or fixing a different problem ?

lib/iris/_lazy_data.py
```diff
@@ -138,7 +124,10 @@ def as_concrete_data(data):
         # In some cases dask may return a scalar numpy.int/numpy.float object
         # rather than a numpy.ndarray object.
         # Recorded in https://github.com/dask/dask/issues/2111.
+        dtype = data.dtype
         data = np.asanyarray(data.compute())
+        if isinstance(data, ma.core.MaskedConstant):
```
pp-mo: Again, how is this tied up with the main point ?

lib/iris/_lazy_data.py
```diff
@@ -138,7 +124,10 @@ def as_concrete_data(data):
         # In some cases dask may return a scalar numpy.int/numpy.float object
         # rather than a numpy.ndarray object.
         # Recorded in https://github.com/dask/dask/issues/2111.
+        dtype = data.dtype
```
pp-mo: The implication here is that `array.dtype == array.compute().dtype` is not guaranteed.
Is that really true + if so why ??

lib/iris/_lazy_data.py
djkirkham: This was the way the code was before the original change. I had moved it into the `_getall` function, which I've now removed. I _think_ it's necessary because dask doesn't like `MaskedConstant`.

lib/iris/_lazy_data.py
djkirkham: If the result of the compute is a `MaskedConstant`, it will always have a float dtype

lib/iris/_lazy_data.py
djkirkham: When I reverted the code to how it was before I found that some of our tests which check that the result of realising lazy data is not a `MaskedConstant` were failing. It seems that Dask is returning a `MaskedConstant` in situations where it wasn't before.

pp-mo:
> the way the code was before the original change

Ah, thanks.  Stupidly, I hadn't thought to compare with the pre-change code.
That's all good, then.

PR #2977, opened 2018-03-13
Provide info on pp-field indices in the file for structured um loads.

pp-mo:
This intends to address problems discussed in [this comment](https://github.com/SciTools/iris/pull/2640#issuecomment-313665559) and follow-ons.

Implementation notes :
  * The information is attached to the 'field' arg in the load callback, i.e. in this case a FieldCollation.
  * The caller should be able to use this to annotate the multidimensional collated cubes as required :  I believe this is what the users's existing code does (using normal load to annotate each raw cube with an extra  "pseudolevel ordering" coordinate).
  * the old 'FieldCollation' class is now renamed 'BasicFieldCollation' : the new, public 'FieldCollation' is a derived class which adds the location-in-input-file info.  This preserves the original, cleaner concept.

Todo:
  * [x] user to confirm suitablility
  * [x] needs a whatsnew

### DIFF
--- a/lib/iris/fileformats/pp.py
+++ b/lib/iris/fileformats/pp.py
@@ -831,7 +831,7 @@ def _pp_attribute_names(header_defn):
     special_headers = list('_' + name for name in _SPECIAL_HEADERS)
     extra_data = list(EXTRA_DATA.values())
     special_attributes = ['_raw_header', 'raw_lbtim', 'raw_lbpack',
-                          'boundary_packing']
+                          'boundary_packing', '_index_in_structured_load_file']
     return normal_headers + special_headers + extra_data + special_attributes
 
 
@@ -864,6 +864,7 @@ class PPField(six.with_metaclass(abc.ABCMeta, object)):
         self.raw_lbtim = None
         self.raw_lbpack = None
         self.boundary_packing = None
+        self._index_in_structured_load_file = None
         if header is not None:
             self.raw_lbtim = header[self.HEADER_DICT['lbtim'][0]]
             self.raw_lbpack = header[self.HEADER_DICT['lbpack'][0]]
--- a/lib/iris/fileformats/um/_fast_load.py
+++ b/lib/iris/fileformats/um/_fast_load.py
@@ -1,4 +1,4 @@
-# (C) British Crown Copyright 2016 - 2017, Met Office
+# (C) British Crown Copyright 2016 - 2018, Met Office
 #
 # This file is part of Iris.
 #
@@ -40,17 +40,65 @@ from contextlib import contextmanager
 import threading
 import os.path
 
+import numpy as np
+
 # Be minimal about what we import from iris, to avoid circular imports.
 # Below, other parts of iris.fileformats are accessed via deferred imports.
 import iris
 from iris.coords import DimCoord
 from iris.cube import CubeList
 from iris.exceptions import TranslationError
-
+from iris.fileformats.um._fast_load_structured_fields import \
+    BasicFieldCollation, group_structured_fields
 
 # Strings to identify the PP and FF file format handler specs.
 _FF_SPEC_NAME = 'UM Fieldsfile'
 _PP_SPEC_NAME = 'UM Post Processing file'
+
+
+class FieldCollation(BasicFieldCollation):
+    # This class specialises the BasicFieldCollation by adding the file-index
+    # and file-path concepts.
+    # This preserves the more abstract scope of the original 'FieldCollation'
+    # class, now renamed 'BasicFieldCollation'.
+
+    def __init__(self, fields, filepath):
+        """
+        Args:
+
+        * fields (iterable of :class:`iris.fileformats.pp.PPField`):
+            The fields in the collation.
+
+        * filepath (string):
+            The path of the file the collation is loaded from.
+
+        """
+        super(FieldCollation, self).__init__(fields)
+        self._load_filepath = filepath
+
+    @property
+    def data_filepath(self):
+        return self._load_filepath
+
+    @property
+    def data_field_indices(self):
+        """
+        Field indices of the contained PPFields in the input file.
+
+        This records the original file location of the individual data fields
+        contained, within the input datafile.
+
+        Returns:
+            An integer array of shape `self.vector_dims_shape`.
+
+        """
+        # Get shape :  N.B. this calculates (and caches) the structure.
+        vector_dims_shape = self.vector_dims_shape
+        # Get index-in-file of each contained field.
+        indices = np.array([field._index_in_structured_load_file
+                            for field in self._fields],
+                           dtype=np.int64)
+        return indices.reshape(vector_dims_shape)
 
 
 def _basic_load_function(filename, pp_filter=None, **kwargs):
@@ -71,8 +119,6 @@ def _basic_load_function(filename, pp_filter=None, **kwargs):
     # Therefore, the actual loader will pass this as the 'pp_filter' keyword,
     # when it is present.
     # Additional load keywords are 'passed on' to the lower-level function.
-    from iris.fileformats.um._fast_load_structured_fields import \
-        group_structured_fields
 
     # Helper function to select the correct fields loader call.
     def _select_raw_fields_loader(fname):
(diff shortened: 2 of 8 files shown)


lib/iris/fileformats/pp.py
```diff
@@ -859,6 +859,7 @@ def __init__(self, header=None):
         self.raw_lbtim = None
         self.raw_lbpack = None
         self.boundary_packing = None
+        self._index_in_structured_load_file = None
```
pp-mo: This may not be necessary, could leave uninitialised.  Suspect the cost is very low.

lib/iris/fileformats/um/_fast_load.py
```diff
@@ -40,19 +40,67 @@
...
+
 # Be minimal about what we import from iris, to avoid circular imports.
 # Below, other parts of iris.fileformats are accessed via deferred imports.
 import iris
 from iris.coords import DimCoord
 from iris.cube import CubeList
 from iris.exceptions import TranslationError
-
+from iris.fileformats.um._fast_load_structured_fields import \
+    BasicFieldCollation, group_structured_fields
```
pp-mo: 'group_structured_fields' was originally a dynamic import.
Don't think that was necessary (seems to work).

pp-mo:
Remaining errors now look to be unrelated (= present business-as-usual problems)

pp-mo:
User confirmed it does work for them.
:heavy_check_mark:

pp-mo:
~DO NOT MERGE~
Now good to go AFAIK :crossed_fingers:

pp-mo:
> Is it so that the order is round-tripped

Exactly.
They are processing and re-saving data, and need to preserve the original file order for compatibility with existing applications.
To be exact, the input is fieldsfiles, for which field order is supposed to be defined by format specification, but there is an unfortunate loophole relating to 'pseudo-level' encoding, for which the occurrence order is, effectively, significant but not predictable.
Correct me if wrong @hdyson ?

>  technically the order isn't entirely lost ... we store a pointer to the data

That is true, but this enables the purpose to be achieved using public supported API instead of relying on hidden implementation detail.

pp-mo:
Rebased !
Just reviewing this morning, whether this can be resolved easily + signed off.
Is this still of interest @hdyson ?

hdyson:
> Is this still of interest @hdyson ?

It is, yes - we *need* to be able to preserve the pseudo level order, and would *like* to be able to use the structured loading, so this looks like it hits the sweet spot of giving us the structured loading benefit alongside retaining the existing functionality.

pp-mo:
Just to paraphrase + review @pelson [previous comments above](https://github.com/SciTools/iris/pull/2977#pullrequestreview-113065457).
I believe I already addressed most of these, as follows ...

  * **What is the usecase for needing the position in the file? Is it so that the order is round-tripped or is there a deeper meaning? ... I think we need to know that we are doing it for good reason ...**
    * the driving usecase seems specific, but the proposal just restores info available to standard load callbacks but missing in structure loads : i.e. which index in which file each field came from

  * **The changes look involved for a use-case that doesn't appear to have been fully clarified. In order to take on such churn, I think we need to know that we are doing it for good reason ...**
    * I honestly don't see that the changes are very involved.  An index property is added to each field; a new collation type stores the filepath, and adds access methods for the new info.

  * **users are going to need to dig deep into the Iris API anyway (to control the save order) ...**
    * not really, you can use the save_pairs interface effectively to do this.

  * **technically the order isn't entirely lost... we store a pointer to the data ...**
    * this is not available via public API


The remaining point is about performance, which I think is worthy of more comment ...

  * **You are touching performance sensitive code. Can you give some reproducible remarks (read: not a full-on performance assessment, more a quick sense of the implications) on what impact the changes have?**

From a performance PoV, nothing has changed in the structure identification step.  PPField gain an extra assignment in `__init__`, and loading -- for structure loading only -- has an additional iterator that assigns a value to each field .  I really can't see anything very costly in this.
The purpose of structured um loading is to avoid : (a) running rules and (b) creating cube objects for each PPField object, and (c) a raw-cube merging stage.
None of that changes, so the only real cost is a small slowdown in creating PPField objects themselves.
This does apply to 'normal' loads as well.  It is in principle avoidable (we don't need to always initialise the private '_index_in_structured_load_file'  property), but I think it is wiser to include the new thing in the `__init__`.  

Here's are some desktop measures of the timing differences, loading from `dataZoo/FF/ff_variants/ff_64.ff`, containing 1202 fields 
  * for making a list of PPFields with `fields = list(iris.fileformats.um.um_to_pp(path))` :
    * old Iris 2.1.0 : (0.533359, 0.528687, 0.527758, 0.556547, 0.529979) ~ 0.5353 seconds
    * new branch code : (0.533923, 0.528155, 0.526929, 0.526727, 0.530310) ~ 0.5292 seconds
  * a 'normal' load with `data = iris.load(path)` :
    * old : (5.352467, 5.332399, 5.365070, 5.373689, 5.333997) ~  5.3515 seconds
    * new : (5.367050, 5.273514, 5.282430, 5.262198, 5.349504) ~ 5.3069 seconds
  * a 'structured' load with `iris.fileformats.structured_loading` and `iris.load` :
    * old : (2.515778, 2.420102, 2.439936, 2.452089, 2.424549) ~  2.4505 seconds
    * new : (2.456725, 2.454413, 2.452883, 2.425538, 2.430701) ~ 2.4441 seconds

Not a great testcase, but nothing much to report.

lib/iris/tests/unit/fileformats/um/fast_load/test_FieldCollation.py
```diff
@@ -0,0 +1,82 @@
...
+                         self.test_filepath)
+        self.assertEqual(self.callback_collations[1].data_filepath,
+                         self.test_filepath)
+
+    def test_callback_collations_field_indices(self):
+        self.assertEqual(
+            self.callback_collations[0].data_field_indices.dtype, np.int64)
+        self.assertArrayEqual(
+            self.callback_collations[0].data_field_indices,
+            [[1, 3], [5, 7]])
```
dkillick: It seems a bit of a surprise to me that the first phenomenon ends up not at the first indices (that is, even indices from 0) in the resultant field indices.

lib/iris/fileformats/pp.py
pp-mo: Yes, on balance I do think it is worth it.

lib/iris/fileformats/um/_fast_load_structured_fields.py
```diff
@@ -285,8 +294,8 @@ def group_structured_fields(field_iterator):
        :func:`iris.fileformats.pp_load_rules._convert_time_coords`).
 
     Returns:
-        A generator of FieldCollation objects, each of which contains a single
-        collated group from the input fields.
+        A generator of BasicFieldCollation objects, each of which contains a
```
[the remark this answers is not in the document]
pp-mo: ok

lib/iris/tests/unit/fileformats/um/fast_load_structured_fields/test_BasicFieldCollation.py
```diff
@@ -31,7 +31,8 @@
 from cftime import datetime
 import numpy as np
 
-from iris.fileformats.um._fast_load_structured_fields import FieldCollation
+from iris.fileformats.um._fast_load_structured_fields \
+    import BasicFieldCollation as FieldCollation
```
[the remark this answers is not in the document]
pp-mo: ok

lib/iris/tests/unit/fileformats/um/fast_load/test_FieldCollation.py
pp-mo: After some investigation (!) ...

It's because structured load ***sorts by stashcode*** [here in fileformats/um/_fast_load_structured_fields](https://github.com/SciTools/iris/blob/7c97b2ee52eb778ca98955f3183351f80cb18aa2/lib/iris/fileformats/um/_fast_load_structured_fields.py#L242), 
and I defined a list of somewhat-arbitrary real phenomena for testing [here in tests/integration/fast_load/test_fast_load](https://github.com/SciTools/iris/blob/7c97b2ee52eb778ca98955f3183351f80cb18aa2/lib/iris/tests/integration/fast_load/test_fast_load.py#L152).
So it happens that p=01 delivers air-temp, then x-wind, which have stashcodes m01s01i004, then m01s00i002.

lib/iris/tests/unit/fileformats/um/fast_load_structured_fields/test_BasicFieldCollation.py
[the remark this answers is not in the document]
pp-mo: I have also changed the `BasicFieldCollation as FieldCollation` usage in this file, because I think that is unreasonably confusing.

lib/iris/tests/unit/fileformats/um/fast_load/test_FieldCollation.py
dkillick: @pp-mo nice investigation!

dkillick:
@pp-mo nice work! Thanks for pushing this through.

Issue #2968, opened 2018-03-07
Cubes with time units lose calendar information


djkirkham:
Probably not a common use case, but still:
```
>>> import iris
>>> from cf_units import Unit
>>> u = Unit('days since 1970-01-01', calendar='360_day')
>>> cube = iris.cube.Cube([1], standard_name='time', units=u)
>>> iris.save(cube, 'my.nc')
```
```
$ ncdump -h my.nc
netcdf my {
dimensions:
	dim0 = 1 ;
variables:
	int64 time(dim0) ;
		time:standard_name = "time" ;
		time:units = "days since 1970-01-01" ;

// global attributes:
		:Conventions = "CF-1.5" ;
}
```

PR #2984, opened 2018-03-15
Save calendar attribute for cube data

djkirkham:
Closes https://github.com/SciTools/iris/issues/2968


bjlittle:
@djkirkham Good spot!

PR #2986, opened 2018-03-17
Rename iris_sample_data to correct package name (iris-sample-data)

raybellwaves:
I got the 
```     
raise ImportError("Please install the 'iris_sample_data' package to "
                          "access sample data.")
```

Then did `conda install -c conda-forge iris_sample_data` to get:
```
PackageNotFoundError: Packages missing in current channels:
            
  - iris_sample_data
```

Renamed iris_sample_data to the correct package name of iris-sample-data

### DIFF
--- a/lib/iris/__init__.py
+++ b/lib/iris/__init__.py
@@ -478,7 +478,7 @@ def sample_data_path(*path_to_join):
     if iris_sample_data is not None:
         target = os.path.join(iris_sample_data.path, target)
     else:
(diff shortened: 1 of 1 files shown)


ajdawson:
This is a bit tricky because the name of the package (the name you import it with) really is `iris_sample_data`, see the lines above the line you changed.

However, since the installation name is probably more relevant than the package name for the sample data, let's go with this change to make the user experience better.

ajdawson:
Thanks @raybellwaves

PR #3005, opened 2018-05-01
Touch tests/unit/analysis/stats/test_pearsonr 

pp-mo:
Derived from #3004, but avoids license header date problem.
This one should check out ok...


rcomer:
See also #3006 :sunglasses: .

PR #3006, opened 2018-05-01
test_pearsonr: reflect new latitude guess_bounds behaviour

rcomer:
Following #2960, I've changed the reference data for `test_compatible_cubes_weighted`.  The difference is purely due to the change in default behaviour of `cube.coords('latitude').guess_bounds()` at Iris 2.0.0.


rcomer:
@pp-mo

pp-mo:
This also obsoletes #3005
Thanks!

PR #3013, opened 2018-05-08
Replace iris.co_realise_cubes with CubeList.realise_data.

### LINKED ISSUE #3012: Rename iris.co_realise_data to CubeList.realise_data
As per https://github.com/SciTools/iris/pull/2967#issuecomment-386780885, follow through with moving the new ```iris.co_realise_data``` into ```CubeList.realise_data```.

cc: @pp-mo

pp-mo:
Closes #3012 

I have to say though, that I find this much less obvious both in the usage and the naming.
The problem is that wrapping a bunch of cubes in a cubelist "just" in order to co-realise them seems clunky and non-obvious.  It means also that the documentation now appears in a much more obscure place.
The new name also does not sing for me : it does not highlight  what is **specifically useful** about the operation.
See the example docstring change here : https://github.com/pp-mo/iris/blob/9b24050b92559b467984bd98b57c7aeadcf2f39e/lib/iris/cube.py#L612
I think this will be typical usage, and it is a bit obscure and horrible IMO.

So frankly,  I still preferred it the way it was.
Please think on? @pelson @dkillick

### DIFF
--- a/lib/iris/coords.py
+++ b/lib/iris/coords.py
@@ -38,8 +38,10 @@ import numpy.ma as ma
 
 from iris._data_manager import DataManager
 from iris._deprecation import warn_deprecated
-from iris._lazy_data import (as_concrete_data, is_lazy_data,
-                             multidim_lazy_stack, lazy_elementwise)
+from iris._lazy_data import (as_concrete_data as _lazy_as_concrete_data,
+                             is_lazy_data as _lazy_is_lazy_data,
+                             multidim_lazy_stack as lazy_multidim_stack,
+                             lazy_elementwise as _lazy_elementwise)
 import iris.aux_factory
 import iris.exceptions
 import iris.time
@@ -554,7 +556,7 @@ class Coord(six.with_metaclass(ABCMeta, CFVariableMixin)):
         return cls(**kwargs)
 
     def _sanitise_array(self, src, ndmin):
-        if is_lazy_data(src):
+        if _lazy_is_lazy_data(src):
             # Lazy data : just ensure ndmin requirement.
             ndims_missing = ndmin - src.ndim
             if ndims_missing <= 0:
@@ -672,7 +674,7 @@ class Coord(six.with_metaclass(ABCMeta, CFVariableMixin)):
 
         """
         result = self._points_dm.core_data()
-        if not is_lazy_data(result):
+        if not _lazy_is_lazy_data(result):
             result = result.view()
         return result
 
@@ -685,7 +687,7 @@ class Coord(six.with_metaclass(ABCMeta, CFVariableMixin)):
         result = None
         if self.has_bounds():
             result = self._bounds_dm.core_data()
-            if not is_lazy_data(result):
+            if not _lazy_is_lazy_data(result):
                 result = result.view()
         return result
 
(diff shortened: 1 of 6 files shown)


lib/iris/cube.py
```diff
@@ -41,7 +41,7 @@
 import iris._concatenate
 import iris._constraints
 from iris._data_manager import DataManager
-from iris._lazy_data import lazy_elementwise
+from iris._lazy_data import lazy_elementwise, co_realise_cubes
```
pelson: Exposing this as ```iris.cube.co_realise_cubes``` isn't ideal.

pp-mo: It isn't exposed, because it's not in the `__all__`.
no?

ajdawson: Importing into the module namespace make it available there from outside, the `__all__` list only affects `from iris.cube import *`.

Instead of importing and calling `co_realise_cubes` you could take the few lines of code that define `co_realise_cubes` and put them in the `realise_cubes` method, and have that as the  only interface to the functionality (deleting the `co_realise_cubes` altogether instead of calling it from `realise_cubes`).

pp-mo: > @ajdawson can we import just "iris._lazy_data" here, instead of specific named components ?

New commit attempts to do this...

pp-mo:
> @ajdawson can we import just "iris._lazy_data" here, instead of specific named components ?

From https://github.com/SciTools/iris/pull/3013#discussion_r187965854
New commit attempts to do this...

lib/iris/coords.py
```diff
@@ -38,8 +38,10 @@
 
 from iris._data_manager import DataManager
 from iris._deprecation import warn_deprecated
-from iris._lazy_data import (as_concrete_data, is_lazy_data,
-                             multidim_lazy_stack, lazy_elementwise)
+from iris._lazy_data import (as_concrete_data as _lazy_as_concrete_data,
+                             is_lazy_data as _lazy_is_lazy_data,
+                             multidim_lazy_stack as lazy_multidim_stack,
```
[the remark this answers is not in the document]
pp-mo: > Is the missing _ intentional?

Oops nooo .... :grimacing:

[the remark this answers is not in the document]
pp-mo: > import iris._lazy_data as _lazy

Thanks, I really missed that :  nicely combines concise use with pre-existing names.

pp-mo:
Following https://github.com/SciTools/iris/pull/3013#discussion_r188306045
Latest looks neater + still addresses earlier concerns.

lib/iris/tests/unit/cube/test_CubeList.py
```diff
@@ -301,5 +302,18 @@ def test_summary_stash(self):
         self.assertEqual(str(self.cubes), expected)
 
 
+class TestRealiseData(tests.IrisTest):
+    def test_realise_data(self):
+        # Simply check that calling CubeList.realise_data is calling
+        # _lazy_data.co_realise_cubes.
+        mock_cubes_list = [mock.Mock(ident=count) for count in range(3)]
+        test_cubelist = CubeList(mock_cubes_list)
+        call_patch = self.patch('iris.cube._lazy.co_realise_cubes')
```
pelson: You might find that you can now mock the actual function ```iris._lazy_data.co_realise_cubes``` as a result of the changed import semantics.

pp-mo: Good spot :+1:

pelson:
@pp-mo - suspect this just needs a rebase.

pp-mo:
Rebased :crossed_fingers: ...

pelson:
Thanks @pp-mo!

PR #3029, opened 2018-05-24
Collapse mean points

duncanwp:
Follow up on #3008  - Adding support for partial collapse of multi-dimensional coordinates *and* now taking the mean of the points from the bounds

(I can rebase this once/if #3028 is merged).


duncanwp:
OK, rebased now. 

What's the process with breaking changes? Could we squeeze a warning into 2.1 that this will change in 2.2?

corinnebosley:
Hi @duncanwp, I'm afraid a warning won't let us make a breaking change in a minor version (aside from the fact that users get so many warnings they all get ignored anyway, it's just plain against the rules).

What I can do is put a label on this so that we are ready to pop it in the next major release and we don't forget about it.

It will need rebasing (again, sorry...), but it might be worth saving that until we're ready to merge it.

lib/iris/coords.py
```diff
@@ -1219,7 +1219,7 @@ def serialize(x):
             # Calculate the bounds and points along the right dims
             bounds = al.stack([item.min(axis=dims_to_collapse),
                                item.max(axis=dims_to_collapse)]).T
-            points = al.array(bounds.sum(axis=-1) * 0.5, dtype=self.dtype)
+            points = item.mean(axis=dims_to_collapse, dtype=self.dtype)
```
rcomer: There was a [question at #3008](https://github.com/SciTools/iris/pull/3008#issuecomment-385983949) about whether multiple collapses are commutative.  It looks like by averaging `item` here, which is the `bounds` of the input coord where available, we do  lose commutativity.  E.g.

```python
mycoord = iris.coords.AuxCoord([[1, 2], [10, 11], [100, 101]], long_name='foo')
print(mycoord.collapsed())
print(mycoord.collapsed(0).collapsed())
print(mycoord.collapsed(1).collapsed())
```

gives:
```
AuxCoord(array([37]), bounds=array([[  1, 101]]), standard_name=None, units=Unit('1'), long_name='foo')
AuxCoord(array([51]), bounds=array([[  1, 101]]), standard_name=None, units=Unit('1'), long_name='foo')
AuxCoord(array([37]), bounds=array([[  1, 101]]), standard_name=None, units=Unit('1'), long_name='foo')
```
Is this the intended behaviour, or should the new points be averaged from the input points?

rcomer:
Just some thoughts that occurred to me while thinking about this change.  Feel free to answer/argue/ignore me as you see fit.  I appreciate I've stumbled in somewhat late in the day!

- If we change the treatment of points in `collapsed`, should we also change it in `aggregated_by` for consistency (I don't know much about `rolling_window`, but maybe that too)?

- What happens if a weighted aggregator is passed?  Should the points averaging use a weighted mean?

bjlittle:
@duncanwp We're not going to able to get this in for `iris` 3.0 :-(

Our plan is to add this a into `iris` `v3.x` but as an opt-in future, which will then become the default behaviour for `iris` 4.0

bjlittle:
@duncanwp Sadly this PR has been stalled for almost 2 and a half years now. 

PR #3030, opened 2018-05-24
Adding a utility for flattening Aux Coords 

duncanwp:
e.g. for later use in extract calls, and other use cases (cf #2766).

It still needs a couple of tests and some docs but I'd like some feedback on the general usefulness and approach first.

See discussion here: https://groups.google.com/forum/#!topic/scitools-iris/Kni1_QO7H1M

Ping @rcomer who also has an implementation of this!

### DIFF
--- a/lib/iris/tests/unit/util/test_flatten_multidim_coord.py
+++ b/lib/iris/tests/unit/util/test_flatten_multidim_coord.py
@@ -1,4 +1,4 @@
-# (C) British Crown Copyright 2010 - 2018, Met Office
+# (C) British Crown Copyright 2010 - 2019, Met Office
 #
 # This file is part of Iris.
 #
--- a/lib/iris/tests/unit/util/test_flatten_multidim_coord.py
+++ b/lib/iris/tests/unit/util/test_flatten_multidim_coord.py
@@ -26,8 +26,7 @@ import iris.tests as tests
 import unittest
 
 import iris.tests.stock as stock
-from iris.coords import DimCoord, AuxCoord
-from iris.util import flatten_multidim_coord, flatten_cube
+from iris.util import flatten_multidim_coord
 
 
 class Test(tests.IrisTest):
@@ -47,48 +46,6 @@ class Test(tests.IrisTest):
         cube_b = flatten_multidim_coord(cube_a, coord)
         self.assertEqual(cube_b.dim_coords, (cube_a.coord('wibble'), ))
         self.assertEqual(cube_b.shape, (2, 12))
-
-    def test_simple_cube_flattened(self):
-        cube_a = stock.simple_2d()
-        cube_b = flatten_cube(cube_a)
-        self.assertEqual(cube_b.dim_coords, tuple())
-        self.assertEqual(cube_b.shape, (12, ))
-
-    def test_multiple_oned_dim_coords_flattened(self):
-        cube_a = stock.simple_3d_w_multidim_coords()
-        cube_a.add_dim_coord(DimCoord([0, 1, 2], var_name='blah'), (1,))
-        cube_a.add_dim_coord(DimCoord([0, 1, 2, 3], var_name='blah2'), (2,))
-        cube_b = flatten_cube(cube_a, (1, 2))
-        self.assertEqual(cube_b.dim_coords, (cube_a.coord('wibble'), ))
-        self.assertEqual(cube_b.coord('blah', dim_coords=False).shape, (12, ))
-        self.assertEqual(cube_b.coord('blah2', dim_coords=False).shape, (12,))
-        self.assertEqual(cube_b.shape, (2, 12))
-
-    def test_oned_aux_coord_flattened(self):
-        cube_a = stock.simple_3d_w_multidim_coords()
-        cube_a.add_aux_coord(AuxCoord([0, 1, 2], var_name='blah'), (1,))
-        cube_b = flatten_cube(cube_a, (1, 2))
-        self.assertEqual(cube_b.dim_coords, (cube_a.coord('wibble'), ))
-        self.assertEqual(cube_b.coord('blah', dim_coords=False).shape, (12, ))
-        self.assertEqual(cube_b.shape, (2, 12))
-
-    def test_aux_coords_leading(self):
-        # Check that it doensn't matter which dimensions the aux-coord spans
-        cube_a = stock.simple_3d_w_multidim_coords()
-        # Move the aux coord dims to the front of the cube
-        cube_a.transpose((1, 2, 0))
-        cube_b = flatten_cube(cube_a, (0, 1))
-        self.assertEqual(cube_b.dim_coords, (cube_a.coord('wibble'), ))
-        self.assertEqual(cube_b.shape, (12, 2))
-
-    def test_split_aux_coord_dims(self):
-        # Check that an aux-coord with 'split' (non-continuous) dimensions are
-        #  flattened correctly
-        cube_a = stock.simple_3d_w_multidim_coords()
-        cube_a.transpose((1, 0, 2))
-        cube_b = flatten_cube(cube_a, (0, 2))
-        self.assertEqual(cube_b.dim_coords, (cube_a.coord('wibble'), ))
-        self.assertEqual(cube_b.shape, (12, 2))
 
 
 if __name__ == '__main__':
--- a/lib/iris/tests/unit/util/test_flatten_multidim_coord.py
+++ b/lib/iris/tests/unit/util/test_flatten_multidim_coord.py
@@ -31,13 +31,17 @@ from iris.util import flatten_multidim_coord, flatten_cube
 
 
 class Test(tests.IrisTest):
-    def test_argument_is_basestring(self):
+    def test_coord_name_as_argument(self):
         cube_a = stock.simple_2d_w_multidim_coords()
-
         cube_b = flatten_multidim_coord(cube_a, 'bar')
         self.assertEqual(cube_b.shape, (12, ))
 
-    def test_argument_is_coord_instance(self):
+    def test_coord_instance_as_argument(self):
+        cube_a = stock.simple_2d_w_multidim_coords()
+        cube_b = flatten_multidim_coord(cube_a, cube_a.coord('bar').copy())
+        self.assertEqual(cube_b.shape, (12, ))
+
+    def test_flatten_2d_coord_on_3d_cube(self):
         cube_a = stock.simple_3d_w_multidim_coords()
         coord = cube_a.coord('bar').copy()
         cube_b = flatten_multidim_coord(cube_a, coord)
@@ -49,14 +53,6 @@ class Test(tests.IrisTest):
         cube_b = flatten_cube(cube_a)
         self.assertEqual(cube_b.dim_coords, tuple())
         self.assertEqual(cube_b.shape, (12, ))
-
-    def test_oned_dim_coord_flattened(self):
-        cube_a = stock.simple_3d_w_multidim_coords()
-        cube_a.add_dim_coord(DimCoord([0, 1, 2], var_name='blah'), (1,))
-        cube_b = flatten_cube(cube_a, (1, 2))
-        self.assertEqual(cube_b.dim_coords, (cube_a.coord('wibble'), ))
-        self.assertEqual(cube_b.coord('blah', dim_coords=False).shape, (12, ))
-        self.assertEqual(cube_b.shape, (2, 12))
 
     def test_multiple_oned_dim_coords_flattened(self):
         cube_a = stock.simple_3d_w_multidim_coords()
@@ -77,6 +73,7 @@ class Test(tests.IrisTest):
         self.assertEqual(cube_b.shape, (2, 12))
 
     def test_aux_coords_leading(self):
+        # Check that it doensn't matter which dimensions the aux-coord spans
         cube_a = stock.simple_3d_w_multidim_coords()
         # Move the aux coord dims to the front of the cube
         cube_a.transpose((1, 2, 0))
@@ -85,6 +82,8 @@ class Test(tests.IrisTest):
         self.assertEqual(cube_b.shape, (12, 2))
 
     def test_split_aux_coord_dims(self):
+        # Check that an aux-coord with 'split' (non-continuous) dimensions are
+        #  flattened correctly
         cube_a = stock.simple_3d_w_multidim_coords()
         cube_a.transpose((1, 0, 2))
         cube_b = flatten_cube(cube_a, (0, 2))
--- a/lib/iris/tests/unit/util/test_flatten_multidim_coord.py
+++ b/lib/iris/tests/unit/util/test_flatten_multidim_coord.py
@@ -1,0 +1,96 @@
+# (C) British Crown Copyright 2010 - 2018, Met Office
+#
+# This file is part of Iris.
+#
+# Iris is free software: you can redistribute it and/or modify it under
+# the terms of the GNU Lesser General Public License as published by the
+# Free Software Foundation, either version 3 of the License, or
+# (at your option) any later version.
+#
+# Iris is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU Lesser General Public License for more details.
+#
+# You should have received a copy of the GNU Lesser General Public License
+# along with Iris.  If not, see <http://www.gnu.org/licenses/>.
+"""Test function :func:`iris.util.flatten_multidim_coord`."""
+
+from __future__ import (absolute_import, division, print_function)
+from six.moves import (filter, input, map, range, zip)  # noqa
+
+# Import iris.tests first so that some things can be initialised before
+# importing anything else.
+import iris.tests as tests
+
+import unittest
+
+import iris.tests.stock as stock
+from iris.coords import DimCoord, AuxCoord
+from iris.util import flatten_multidim_coord, flatten_cube
+
+
+class Test(tests.IrisTest):
+    def test_argument_is_basestring(self):
+        cube_a = stock.simple_2d_w_multidim_coords()
+
+        cube_b = flatten_multidim_coord(cube_a, 'bar')
+        self.assertEqual(cube_b.shape, (12, ))
+
+    def test_argument_is_coord_instance(self):
+        cube_a = stock.simple_3d_w_multidim_coords()
+        coord = cube_a.coord('bar').copy()
+        cube_b = flatten_multidim_coord(cube_a, coord)
+        self.assertEqual(cube_b.dim_coords, (cube_a.coord('wibble'), ))
+        self.assertEqual(cube_b.shape, (2, 12))
+
+    def test_simple_cube_flattened(self):
+        cube_a = stock.simple_2d()
+        cube_b = flatten_cube(cube_a)
+        self.assertEqual(cube_b.dim_coords, tuple())
+        self.assertEqual(cube_b.shape, (12, ))
+
+    def test_oned_dim_coord_flattened(self):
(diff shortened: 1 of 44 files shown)


lib/iris/util.py
```diff
@@ -1507,6 +1507,98 @@ def demote_dim_coord_to_aux_coord(cube, name_or_coord):
...
+                           np.product(shape_array[list(coord_dims)])])
+
+    new_data = cube.data.reshape(new_shape)
+    new_aux_coords = other_aux_coords
+    for c in coords_to_flatten:
+        # Only flatten aux coords (we will drop DimCoords)
+        if not isinstance(c, DimCoord):
+            # The new (flat) AuxCoords are always the last dimension
+            new_aux_coords.append((c.copy(c.points.flat,
+                                          c.bounds.reshape(-1, 2)),
```
rcomer: I got an error here because `bounds` is None.

lib/iris/util.py
```diff
@@ -1507,6 +1507,98 @@ def demote_dim_coord_to_aux_coord(cube, name_or_coord):
...
+        coord = name_or_coord
+    else:
+        # Don't know how to handle this type
+        msg = ("Don't know how to handle coordinate of type {}. "
+               "Ensure all coordinates are of type six.string_types or "
+               "iris.coords.Coord.")
+        msg = msg.format(type(name_or_coord))
+        raise TypeError(msg)
+
+    coord_dims = cube.coord_dims(coord)
```
rcomer: Is it worth adding something like
```
if len(coord_dims) == 1:
    return cube
```

lib/iris/util.py
```diff
@@ -1507,6 +1507,98 @@ def demote_dim_coord_to_aux_coord(cube, name_or_coord):
...
+                        if c not in coords_to_flatten]
+
+    shape_array = np.asarray(cube.shape)
+    new_shape = np.hstack([shape_array[other_dims],
+                           np.product(shape_array[list(coord_dims)])])
+
+    new_data = cube.data.reshape(new_shape)
+    new_aux_coords = other_aux_coords
+    for c in coords_to_flatten:
+        # Only flatten aux coords (we will drop DimCoords)
```
rcomer: My preference would be for the DimCoords to be converted to AuxCoords, with their points and bounds repeated so they fit along the flattened axis.

lib/iris/util.py
```diff
@@ -1507,6 +1507,98 @@ def demote_dim_coord_to_aux_coord(cube, name_or_coord):
...
+                        if c not in coords_to_flatten]
+    other_aux_coords = [(c, cube.coord_dims(c)) for c in
+                        cube.coords(dim_coords=False)
+                        if c not in coords_to_flatten]
+
+    shape_array = np.asarray(cube.shape)
+    new_shape = np.hstack([shape_array[other_dims],
+                           np.product(shape_array[list(coord_dims)])])
+
+    new_data = cube.data.reshape(new_shape)
```
rcomer: Could this be done on the dask array if the cube has lazy data?

lib/iris/util.py
rcomer: I'm also not sure if this is robust to the cases where AuxCoord dimensions are not necessarily in the order you expect (#2606).

rcomer:
I'd definitely be pleased to see something like this included in Iris, and your implementation looks a lot cleaner than mine!

In case it helps motivate a merge, I'll explain my own use case:  I'm analysing GloSea hindcast ensembles.  Sometimes the cube looks like this (I added the `combined_realization` coord as a function of the `realization` and `forecast_reference_time` coords):
```
air_temperature / (K)               (realization: 4; -- : 108; latitude: 324; longitude: 432)
     Dimension coordinates:
          realization                           x       -              -               -
          latitude                              -       -              x               -
          longitude                             -       -              -               x
     Auxiliary coordinates:
          combined_realization                  x       x              -               -
          forecast_period                       -       x              -               -
          forecast_reference_time               -       x              -               -
          time                                  -       x              -               -
     Scalar coordinates:
          height: 1.5 m
     Attributes:
          STASH: m01s03i236
          um_version: 10.4
     Cell methods:
          mean: time (1 hour)
```

But if a member falls over it gets replaced with a higher numbered one, so the cube looks like this:

```
air_temperature / (K)               (-- : 432; latitude: 324; longitude: 432)
     Dimension coordinates:
          latitude                      -              x               -
          longitude                     -              -               x
     Auxiliary coordinates:
          forecast_period               x              -               -
          forecast_reference_time       x              -               -
          realization                   x              -               -
          time                          x              -               -
     Scalar coordinates:
          height: 1.5 m
     Attributes:
          STASH: m01s03i236
          um_version: 10.4
     Cell methods:
          mean: time (1 hour)
```

If I flatten out the `combined_realization` coord, then my cubes will always have the same shape.  This makes writing subsequent analysis code simpler!

lib/iris/util.py
rcomer: I'd like the 1D AuxCoords on these dimensions to be treated similarly.

lib/iris/util.py
duncanwp: Yes, data.reshape should work on both numpy and dask arrays

lib/iris/util.py
duncanwp: 1d AuxCoords should survive untouched

lib/iris/util.py
duncanwp: Good idea

lib/iris/util.py
duncanwp: I'll fix that

duncanwp:
Thanks for the useful feedback! 

The latest commit should include all of your suggestions, and I've added some tests. It would be great to include your use-case as a test though - do you have some code I could use to mock up a cube with those coordinate?

This still needs a whats-new if it's to be merged

duncanwp:
Note to self: it might be nice to factor out a method for flattening a cube along specific dimensions (which is what the meat of this function really does). I should also check that creating a new cube is necessary and desirable - and if so that the new coord points aren’t just views on the old ones.

lib/iris/util.py
rcomer: I've just twigged that the auxcoord of interest gets moved to the last dimension when flattened, so perhaps a `cube.transpose` would be more consistent.

lib/iris/util.py
```diff
@@ -1517,6 +1517,111 @@ def demote_dim_coord_to_aux_coord(cube, name_or_coord):
...
+
+    other_dim_coords = [(c, i) for i, c in
+                        enumerate(cube.coords(dim_coords=True))
+                        if c not in coords_to_flatten]
+    other_aux_coords = [(c, cube.coord_dims(c)) for c in
+                        cube.coords(dim_coords=False)
+                        if c not in coords_to_flatten
+                        and c not in coords_to_expand]
+
+    new_data = cube.data.reshape(new_shape)
```
rcomer: Think this needs `cube.core_data()` if it's to work on Dask arrays.

I also think the array might need transposing if the auxcoord didn't start off on the trailing dimensions.  Sorry, I missed that before.

lib/iris/util.py
```diff
@@ -1517,6 +1517,111 @@ def demote_dim_coord_to_aux_coord(cube, name_or_coord):
...
+    coords_to_expand = []
+    for d in coord_dims:
+        coords_to_expand.extend(cube.coords(dimensions=d))
+
+    # Expand the coords which need expanding first, then add to flatten list
+    for c in coords_to_expand:
+        new_points = broadcast_to_shape(c.points, coord.shape, (0,))
+        new_bounds = broadcast_to_shape(c.bounds, coord.bounds.shape, (0,)) \
+            if c.bounds is not None else None
+        coords_to_flatten.append(c.copy(new_points, new_bounds))
```
rcomer: Test is failing here because its trying to create a 2d DimCoord.  You can get around it with
```python
coords_to_flatten.append(
    iris.coords.AuxCoord.from_coord(c).copy(new_points, new_bounds))
```
though there may be a better way.

Then the test **still** fails when the cube is constructed because it claims the new 'blah' is only length 3.  I do not understand this at all.

lib/iris/util.py
```diff
@@ -1517,6 +1517,111 @@ def demote_dim_coord_to_aux_coord(cube, name_or_coord):
...
+    coords_to_flatten = cube.coords(dimensions=coord_dims)
+    # And any other coords which span any one of the dimensions
+    # These (1-D) coordinates are expanded before being flattened
+    coords_to_expand = []
+    for d in coord_dims:
+        coords_to_expand.extend(cube.coords(dimensions=d))
+
+    # Expand the coords which need expanding first, then add to flatten list
+    for c in coords_to_expand:
+        new_points = broadcast_to_shape(c.points, coord.shape, (0,))
```
rcomer: The `(0,)` here will need generalising I think.

rcomer:
> do you have some code I could use to mock up a cube with those coordinate?
I don't at the moment but will have a go when I have a minute.

lib/iris/util.py
duncanwp: Actually, although it looks a bit wierd the zero here is fine because c is must be one dimensional. I've added a comment to clarify

lib/iris/util.py
duncanwp: Yes, good catch thanks - I've fixed that now.

lib/iris/util.py
duncanwp: Yes, I need to get used to using `core_data()` now!

duncanwp: OK, I think it will now work with Aux Coords across any combination of dimensions, I've added a couple of tests too.

duncanwp:
Does anyone have a preference for if this should work on the Cube in-place? It would have to reach into the Cube internals to update the coordinates which then makes it feel more like a method on the Cube itself, but maybe it's OK?

lib/iris/util.py
rcomer: `c` is one dimensional, but `coord` is multidimensional, and the tuple indicates the dimension of `coord` that `c` is getting mapped to.  So, for your docstring example, you would have 
`broadcast_to_shape(c.points, (2, 2), (0,))` for the `x` coord, but 
`broadcast_to_shape(c.points, (2, 2), (1,))` for the `y` coord.

duncanwp: Ah, yes - sorry! :flushed:

Is there a nice way of getting the dimension of an Aux Coord for a given DimCoord?

rcomer: No specific method that I know of.  I'd be tempted to combine the loop with the one at line 1614.

```python
for dim_map, d in enumerate(coord_dims):
    for c in cube.coords(dimensions=d):
        new_points = broadcast_to_shape(c.points, coord.shape, (dim_map,))
        ...etc...
```

duncanwp: Good idea!

duncanwp:
OK, I don't think I'll be able to do this inlace since the DataManager explicitly doesn't allow these kind of reshapes (see `_data_manager.py:L260`). Is there a way around this? 

I had come to the conclusion that since we're just reshaping it would be nice not to have to copy the data...

duncanwp:
@pelson any ideas?

pp-mo:
An inplace operation should definitely be a Cube method, as at present the external API is designed to guarantee that any cube derived from another one is totally independent, i.e. no data sharing.
However, this is something we are intending to relax in future, though we are currently a bit blocked on that -- see #2681.

The guaranteed copying behaviour seemed like a great idea at outset of Iris, because of the total certainty given, and at least partly because many of our internal users were used to IDL which did make copies in all cases.
Going forward we can instead probably adopt the numpy attitude, which I'd summarise as 
"for efficiency, results can sometimes share data with the input; you just need to be careful".

rcomer:
I've finally got around to testing the latest revision with my data and can report that it's working nicely now:

```python
In [14]: print(cube)
air_temperature / (K)               (realization: 4; -- : 108; latitude: 324; longitude: 432)
     Dimension coordinates:
          realization                           x       -              -               -
          latitude                              -       -              x               -
          longitude                             -       -              -               x
     Auxiliary coordinates:
          combined_realization                  x       x              -               -
          forecast_period                       -       x              -               -
          forecast_reference_time               -       x              -               -
          time                                  -       x              -               -
     Scalar coordinates:
          height: 1.5 m
     Attributes:
          STASH: m01s03i236
          um_version: 10.4
     Cell methods:
          mean: time (1 hour)

In [15]: print(iris.util.flatten_multidim_coord(cube, 'combined_realization'))
air_temperature / (K)               (-- : 432; latitude: 324; longitude: 432)
     Dimension coordinates:
          latitude                      -              x               -
          longitude                     -              -               x
     Auxiliary coordinates:
          combined_realization          x              -               -
          forecast_period               x              -               -
          forecast_reference_time       x              -               -
          realization                   x              -               -
          time                          x              -               -
     Scalar coordinates:
          height: 1.5 m
     Attributes:
          STASH: m01s03i236
          um_version: 10.4
     Cell methods:
          mean: time (1 hour)

```

I did start writing some code to mock up a cube with similar metadata, but it got messy very quickly :worried:.

lib/iris/tests/unit/util/test_flatten_multidim_coord.py
```diff
@@ -0,0 +1,96 @@
...
+
+        cube_b = flatten_multidim_coord(cube_a, 'bar')
+        self.assertEqual(cube_b.shape, (12, ))
+
+    def test_argument_is_coord_instance(self):
+        cube_a = stock.simple_3d_w_multidim_coords()
+        coord = cube_a.coord('bar').copy()
+        cube_b = flatten_multidim_coord(cube_a, coord)
+        self.assertEqual(cube_b.dim_coords, (cube_a.coord('wibble'), ))
+        self.assertEqual(cube_b.shape, (2, 12))
```
corinnebosley: I also think that these tests are getting a bit jumbled.  Your top two tests here are `test_argument_is_basestring` and `test_argument_is_coord_instance`, which should have all the same parameters and checks (i.e. the same stock cube and the same asserts) to make sure that the behaviour is exactly the same when that one input type is changed.

Then it is appropriate to add another test with a 3d cube to make sure that this is also handled correctly, rather than adding those checks sneakily into a test for something else. 

Would you mind separating out these tests in that way please?

lib/iris/tests/unit/util/test_flatten_multidim_coord.py
```diff
@@ -0,0 +1,96 @@
...
+
+    def test_simple_cube_flattened(self):
+        cube_a = stock.simple_2d()
+        cube_b = flatten_cube(cube_a)
+        self.assertEqual(cube_b.dim_coords, tuple())
+        self.assertEqual(cube_b.shape, (12, ))
+
+    def test_oned_dim_coord_flattened(self):
```
corinnebosley: The odd comment with an explanation of the point of the test wouldn't go amiss.

lib/iris/tests/unit/util/test_flatten_multidim_coord.py
```diff
@@ -0,0 +1,96 @@
...
+        cube_a = stock.simple_3d_w_multidim_coords()
+        cube_a.add_dim_coord(DimCoord([0, 1, 2], var_name='blah'), (1,))
+        cube_b = flatten_cube(cube_a, (1, 2))
+        self.assertEqual(cube_b.dim_coords, (cube_a.coord('wibble'), ))
+        self.assertEqual(cube_b.coord('blah', dim_coords=False).shape, (12, ))
+        self.assertEqual(cube_b.shape, (2, 12))
+
+    def test_multiple_oned_dim_coords_flattened(self):
```
corinnebosley: I don't think this test name describes the test very well.  The word 'oned' seems inappropriate as one of your additional coords is actually twoed (or they are both just inserted).  

I'm also not sure what this test adds compared to that above, so I would recommend removing it (considering the heaving test suite that Iris already accommodates, any new tests added should be lean and purposeful).

lib/iris/tests/unit/util/test_flatten_multidim_coord.py
```diff
@@ -0,0 +1,96 @@
...
+        self.assertEqual(cube_b.dim_coords, (cube_a.coord('wibble'), ))
+        self.assertEqual(cube_b.shape, (12, 2))
+
+    def test_split_aux_coord_dims(self):
+        cube_a = stock.simple_3d_w_multidim_coords()
+        cube_a.transpose((1, 0, 2))
+        cube_b = flatten_cube(cube_a, (0, 2))
+        self.assertEqual(cube_b.dim_coords, (cube_a.coord('wibble'), ))
+        self.assertEqual(cube_b.shape, (12, 2))
+
```
corinnebosley: Again, I'm not sure if this test adds any value that hasn't already been added by `test_aux_coords_leading`, so I would suggest removing it.

duncanwp: This is useful because there is some special logic for dealing with these types of coordinates

lib/iris/tests/unit/util/test_flatten_multidim_coord.py
duncanwp: I've tried to rationalise these a bit now and added a comment to describe them

duncanwp:
Thanks for the detailed review @corinnebosley. I've finally managed to get around to making your proposed changes. Hopefully it's looking a bit tidier now.

rcomer:
Hi @duncanwp, I just had a look at your Travis failures, and some of them are to do with a `cftime` problem that was fixed at #3259.  Could you rebase?  It would be good to get this in: some recent posters on the Google Group would benefit from it 😀.

duncanwp:
Should be ready to go now :-)

rcomer:
It’s good to see the tests are now passing 👍 , but it looks like the PR now incorporates some of the `master` commit history.  Not sure what’s happened there.  Hopefully a `git` expert will be along soon with a suggestion...

duncanwp:
I had a conflict rebasing so perhaps that caused it. You should be able to squash it when you accept the pull request though I think…

duncanwp:
@bjlittle I consider this done now, rather than work in progress. Do you think there's something missing?

bjlittle:
Hey @duncanwp, awesome! I'll take a peek and start reviewing...

Before I do, would you consider it a feature enhancement that's not a breaking change? We're teeing up `iris` minor release `2.3.0` for the end of August, and `iris` major release `3.0.0` for November...

duncanwp:
It's just an extra feature and not breaking so if you can roll it into `2.3.0` that would be great!

pp-mo:
Hi @duncanwp
We're working on iris 2.3 and I'm just considering this PR, but the history has got into a nasty mess for some reason...

The whole of a 2.2.x mergeback to master (#3256) is appearing on your commit chain, and the main set of changes with your name appear twice in 2 different versions !
I think the [penultimate commit](https://github.com/SciTools/iris/pull/3030/commits/1b187a353568d73e9c84eefd0ce1a3020e0c12a4) "Merge remote-tracking branch 'origin/flatten_aux_coord' into flatten_aux_coord" is probably the main culprit, it sounds like a mistake to merge from your fork back into your local ?

So I rebuilt by just cherry-picking the final commits, which then looks like #3379.
Does that contain everything it should have, in your latest incarnation ??

Apart from that I think may be a few remaining things to pick up on, but I guess we need you to fix the PR, i.e. reconfirm what is proposed first.
**Can you find some time to fix this ?**

pp-mo:
Note my (failed!) attempt to expedite this : https://github.com/SciTools/iris/pull/3379

There are outstanding relevant comments there :
  * https://github.com/SciTools/iris/pull/3379#discussion_r317369357
  * https://github.com/SciTools/iris/pull/3379#pullrequestreview-280093029

duncanwp:
That's frustrating, I thought it was ready :-( 

I won't have a chance to look at this for a while but I'll pick it up when I have a chance.

pp-mo:
Thanks for your attention @duncanwp .
We'll wait !

bjlittle:
@duncanwp Sadly this PR has been stalled for almost 2 and a half years now. 

PR #3073, opened 2018-06-20
Fix docstring re structured text warnings

tv3141:
Related to #3065


lib/iris/fileformats/pp.py
```diff
@@ -1886,7 +1886,7 @@ def load_pairs_from_fields(pp_fields):
         An iterable of :class:`iris.fileformats.pp.PPField`.
 
     Returns:
-        An iterable of :class:`iris.cube.Cube`s.
+        An iterable of :class:`iris.cube.Cube`\ s.
```
pp-mo: Ooh, is that how you do it ?!?
I've been wanting + failing to find this trick for ages.
There are probably a lot more places which could benefit ...

PR #3086, opened 2018-07-13
Quick fix to get tests passing

dkillick:
Further pin the dask dependency version as a quick workaround for [this failing test](https://travis-ci.org/SciTools/iris/jobs/397436711#L5120-L5128). For clarification, the test passes at dask=0.18.0, but fails at dask=0.18.1.

I'm certain there's a proper code fix to this currently-failing test, but this change as a temporary measure at least lets us get the tests passing again.

@pelson @ajdawson do you have any space to offer a review?

PR #3088, opened 2018-07-16
Workaround for dask array copy bug

### LINKED ISSUE #3087: Lazy cube copy behaviour
Copying a cube with lazy data currently isn't behaving as expected:
```python
from iris.cube import Cube
from iris._lazy_data import as_lazy_data
import numpy as np

c = Cube(as_lazy_data(np.array([0, 1]), chunks=2,))
cc = c.copy()
c is cc
False
c.data is cc.data
True
```

There's current a [failing test](https://travis-ci.org/SciTools/iris/jobs/397436711#L5120-L5128) on Travis because of this behaviour ([it's this test that's failing](https://github.com/SciTools/iris/blob/master/lib/iris/tests/unit/cube/test_Cube.py#L1432-L1434)).

I think we're inheriting this behaviour from [dask 0.18.1](https://github.com/dask/dask/issues/3751), so a quick workaround for this is to [pin dask](#3086) to v0.18.0, which doesn't evidence this behaviour.

dkillick:
A code fix, rather than a library version pin, to fix [this failing test](https://travis-ci.org/SciTools/iris/jobs/397436711#L5120-L5128), while we await a dask release containing the [bugfix](https://github.com/dask/dask/pull/3754). The code fix implemented here was suggested as a workaround in the [dask bug report](https://github.com/dask/dask/issues/3751#issuecomment-404913371).

Backs out the library version pin temporarily introduced in #3086. 

Fixes #3087.

### DIFF
--- a/lib/iris/tests/unit/cube/test_Cube.py
+++ b/lib/iris/tests/unit/cube/test_Cube.py
@@ -1430,7 +1430,9 @@ class Test_copy(tests.IrisTest):
         self._check_copy(cube, cube.copy())
 
     def test__lazy(self):
-        cube = Cube(as_lazy_data(np.array([1, 0])))
+        # Note: multiple chunks added as a workaround suggested to dask#3751,
+        # which is fixed in dask#3754.
+        cube = Cube(as_lazy_data(np.array([1, 0]), chunks=(1, 1)))
         self._check_copy(cube, cube.copy())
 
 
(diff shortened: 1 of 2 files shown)


pp-mo:
Looks fine to me.
I think that workaround can just stay in place indefinitely, for compatibility with any Dask version.

PR #3092, opened 2018-07-23
add contact options to README

tv3141:
Add list of contact options to README, so that inquiries go to the most appropriate channel.

Gitter is blocked for me. Are you actively using it?


ajdawson:
Gitter is still available at that link, it was supposed to be for general development discussions. It was only used a handful of times between April-October 2017, although in principal it is still active...

PR #3094, opened 2018-07-23
Add stash mapping changes v2.0.0 -> v2.1.0 to changelog

### LINKED ISSUE #2414: Add stash to cf mapping changes to change log
A few names in the cf to stash mapping in `lib/iris/fileformats/um_cf_map.py ` were corrected in [Iris 1.11](https://github.com/SciTools/iris/compare/v1.10.x...v1.11.x#diff-1680206bdc5cfaa83e14428f5ba0f848) and [Iris 1.12](https://github.com/SciTools/iris/compare/v1.11.x...v1.12.x#diff-1680206bdc5cfaa83e14428f5ba0f848).

I think changes to cf-standard names should be part of future version change logs, as they break existing code.

tv3141:
Closes #2414.

I included only name changes, as referring to a cube by name is common.

I omitted unit changes and whether a STASH is translated into a standard name or a long_name.


dkillick:
Hi @tv3141, thanks for making this proposal to increase the detail in the Iris whatsnew. I do, however, have significant reservations about the change in its current form. The Iris whatsnew should detail changes made in Iris, but my concern is that the additions you're proposing here are **not** changes that have been made in Iris, merely changes that Iris inherits from metarelate and presents to users as Iris behaviour is affected by metarelate changes.

By means of an analogy, the change you're proposing here is like reproducing part of the netCDF4-python library's changelist because it impacts on how Iris behaves. Such a reproduction is unprecedented, and "not the done thing", beyond, perhaps, noting that Iris now uses a new version of netCDF4-python by default, which (minus the metarelate release version) is effectively what's happening here too.

To preempt a perhaps obvious issue with my argument, I'm aware that a change _has_ been made to Iris in the `um_cf_map` file. I'm well aware that this is the case, but even then, the file in question [makes it clear](https://github.com/SciTools/iris/blob/master/lib/iris/fileformats/um_cf_map.py#L18-L21) that the contents of this file are auto-generated from changed metarelate content. As such, this file is arguably not a part of Iris beyond needing to be included for the translations it provides.

Having said all that, I very much support the intention here if not the implementation, for the reasons detailed above. I think it's valuable for Iris users to be able to easily see the metadata translation changes that may impact them on release of a new Iris version with metadata translation updates. My proposal for doing this would be a single item in the whatsnew that looks something like the following:

``` markdown
## New Features
* ...
* This release of Iris contains a number of updated metadata translations. 
  See [this changelist](https://github.com/SciTools/iris/commit/69597eb3d8501ff16ee3d56aef1f7b8f1c2bb316#diff-1680206bdc5cfaa83e14428f5ba0f848) for further information.
* ...
```

So instead of laying out the changes in the whatsnew, we instead provide a link to the changes that have been made to metadata translations (either within Iris or directly from metarelate, I guess).

corinnebosley:
@bjlittle This WhatsNew entry is outdated now (since we have just released iris 2.2); is there still value in adding this entry retrospectively?

pp-mo:
>  is there still value in adding this entry retrospectively?

I'm with @dkillick on this : I don't think we should start doing this.

Unfortunately I think the key problem is the lack of versioning in https://github.com/metarelate/metOcean.
Ideally, we ***should*** be able to simply say  "Updated new UM file translations from metarelate/metOcean vX.Y", and point to some whatsnew-like summary of changes over a period there.
But as it is, metarelate/metOcean has no versioning, no releases and no changenotes :-(

pp-mo:
> metarelate/metOcean has no versioning

We can however just reference https://github.com/SciTools/iris/pull/3043, which shows the resulting changes in um_cf_map.py.

corinnebosley:
Boom! Thanks @tv3141, your time and effort is much appreciated.

pp-mo:
Oh yes, that is much nicer ! :bouquet:

dkillick:
@tv3141 nice! Thanks for submitting this whatsnew update, and thanks @corinnebosley for the merge 👍

dkillick:
@tv3141 nice! Thanks for submitting this whatsnew update, and thanks @corinnebosley for the merge 👍

PR #3095, opened 2018-07-24
2d bounds plotting with initial tests

corinnebosley:


corinnebosley:
@lbdreyer @pp-mo I just went ahead and added the changes that I made this morning, but I haven't yet reviewed and amended the tests.

I have left a lot of comments and notes in the discontiguity checking function, but I don't intend to keep them there.

lib/iris/plot.py
```diff
@@ -1045,6 +1140,9 @@ def pcolormesh(cube, *args, **kwargs):
     * axes: the :class:`matplotlib.axes.Axes` to use for drawing.
         Defaults to the current axes if none provided.
 
+    * two_dim_coord_contiguity_atol: absolute tolerance when checking for
```
lbdreyer: This needs updating to be `two_dim_contig_atol`

lib/iris/plot.py
```diff
@@ -286,7 +367,7 @@ def _draw_2d_from_bounds(draw_method_name, cube, *args, **kwargs):
         string_axes = {}
 
         for coord, axis_name, data_dim in zip([u_coord, v_coord],
-                                              ['xaxis', 'yaxis'],
+                                              ['plotxaxis', 'yaxis'],
```
lbdreyer: This should be changes back to `'xaxis'`
axis_name gets added to `string_axes` which is used in [_string_coord_axis_tick_labels](https://github.com/corinnebosley/iris/blob/1820e750315b07bdbdd6acadbe77fa23d4fca750/lib/iris/plot.py#L242) where it will try
```getattr(ax, 'plotxaxis')```
which doesn't exist. But
```gettattr(ax, 'axis')```
does

File: lib/iris/tests/stock.py [hunk 3]
```diff
@@ -46,12 +46,12 @@ def lat_lon_cube():
     """
     cube = Cube(np.arange(12, dtype=np.int32).reshape((3, 4)))
     cs = GeogCS(6371229)
-    coord = iris.coords.DimCoord(points=np.array([-1, 0, 1], dtype=np.int32),
+    coord = DimCoord(points=np.array([-1, 0, 1], dtype=np.int32),
```
lbdreyer: If you do this, you can just remove the imports:
```
import iris.coords
import iris.coords as icoords
```
(which as the same thing)  as they never get used

lib/iris/tests/stock.py
```diff
@@ -486,25 +495,37 @@ def realistic_4d():
...
                                    units='1', attributes={'positive': 'up'})
-    sigma = icoords.AuxCoord(sigma_pts, long_name='sigma', units='1', bounds=sigma_bnds)
-    orography = icoords.AuxCoord(orography, standard_name='surface_altitude', units='m')
-    time = icoords.DimCoord(time_pts, standard_name='time', units='hours since 1970-01-01 00:00:00')
-    forecast_period = icoords.DimCoord(forecast_period_pts, standard_name='forecast_period', units='hours')
+    sigma = icoords.AuxCoord(sigma_pts, long_name='sigma', units='1',
+                             bounds=sigma_bnds)
+    orography = icoords.AuxCoord(orography, standard_name='surface_altitude',
+                                 units='m')
+    time = icoords.DimCoord(time_pts, standard_name='time',
```
lbdreyer: Make these `DimCoord` and `AuxCoord`

lib/iris/coords.py
```diff
@@ -143,6 +143,147 @@ def __new__(cls, name_or_coord, minimum, maximum,
                                       'groupby_point, groupby_slice')
 
 
+def discontiguity_check_2d(coord, data, abs_tol=1e-4, transpose=False):
```
pp-mo: I have a concern here about the use of an absolute tolerance, because the code will not *always* be dealing with lats+lons, and maybe not always in degrees, so you can't assume a particular value scaling.
A relative tolerance might be better, or maybe we can somehow establish a sensible value value + use a fraction of that as an absolute threshold ?

lib/iris/coords.py
```diff
@@ -143,6 +143,147 @@ def __new__(cls, name_or_coord, minimum, maximum,
...
+        bds = coord.bounds
+    axis = guess_coord_axis(coord)
+
+    # TODO Rewrite this comment? The bounds indexes should be anticlockwise from bottom-left, not clockwise from top-right.
+    # Check ordering:
+    #         i    i+1
+    #    j    @0    @1
+    #  j+1    @3    @2
+    def mod360_diff(x1, x2):
+        diff = x1 - x2
```
pp-mo: Here again, the code is assuming the values must be angles in degrees. 
Can we guarantee that is the only thing that this deals with,
and if not can we either distinguish this case or generalise the operation ?

lib/iris/coords.py
pp-mo: I have a concern here about the use of an absolute tolerance, because the code will not always be dealing with lats+lons, and maybe not always in degrees, so you can't assume a particular value scaling.
A relative tolerance might be better, or maybe we can somehow establish a sensible value value + use a fraction of that as an absolute threshold ?

lib/iris/coords.py
pp-mo: Here again, the code is assuming the values must be angles in degrees.
Can we guarantee that is the only thing that this deals with,
and if not can we either distinguish this case or generalise the operation ?

lib/iris/coords.py
```diff
@@ -143,6 +143,147 @@ def __new__(cls, name_or_coord, minimum, maximum,
...
+        # If discontinuity occurs but not masked, any grid will be created
+        #  incorrectly, so raise a warning
+        if not_masked_at_discontinuity:
+            raise ValueError('The bounds of the {} coordinate are not'
+                             ' contiguous and data is not masked where the'
+                             ' discontiguity occurs. Not able to create a'
+                             ' suitable mesh to give to'
+                             ' Matplotlib'.format(coord.name()))
+        else:
+            warnings.warn('The bounds of the {} coordinate are not contiguous.'
```
pp-mo: I'm not keen on issuing a warning in these cases, as we have seen that some of the well-known testcase data will  *always* give this warning, and there is no practical change you can make to stop that. 
IMHO a warning is simply not useful unless there is an obvious code change that you can make (and should make) to acknowledge and/or fix a problem  -- ***ideally*** the warning text will suggest appropriate action. 
Otherwise it is just more noise !

You could potentially add a keyword here, which the user can use to acknowledge the problem + silence the warning.  Not really sure whether I think that is a good idea or not  in this case ...

PR #3096, opened 2018-07-26
Gridcell rotations

pp-mo:
~**WIP - DO NOT MERGE~ probably can now merge, once caveats are established
Here is a first look at the gridcell angles code.

The tests are currently a tangled mess of exploratory code + not fit for human consumption !
Eventually, I will have just the 2 new routines and proper unit tests for them.

There is a fair bit that still needs to be done, even in the main code :
  * improve documentation generally
  * provide for coords in alternative coord-systems, by projecting to ordinary lat-lon
  * review how coords are identified (axis rather than 'longitude' + 'latitude' names)

### DIFF
--- a/lib/iris/analysis/_grid_angles.py
+++ b/lib/iris/analysis/_grid_angles.py
@@ -1,3 +1,27 @@
+# (C) British Crown Copyright 2010 - 2018, Met Office
+#
+# This file is part of Iris.
+#
+# Iris is free software: you can redistribute it and/or modify it under
+# the terms of the GNU Lesser General Public License as published by the
+# Free Software Foundation, either version 3 of the License, or
+# (at your option) any later version.
+#
+# Iris is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU Lesser General Public License for more details.
+#
+# You should have received a copy of the GNU Lesser General Public License
+# along with Iris.  If not, see <http://www.gnu.org/licenses/>.
+"""
+Code to implement vector rotation by angles, and inferring gridcell angles
+from coordinate points and bounds.
+
+"""
+from __future__ import (absolute_import, division, print_function)
+from six.moves import (filter, input, map, range, zip)  # noqa
+
 import numpy as np
 
 import iris
@@ -125,23 +149,41 @@ def _angle(p, q, r):
 
 def gridcell_angles(x, y=None, cell_angle_boundpoints='mid-lhs, mid-rhs'):
     """
-    Calculate gridcell orientation angles.
+    Calculate gridcell orientations for an arbitrary 2-dimensional grid.
+
+    The input grid is defined by two 2-dimensional coordinate arrays with the
+    same dimensions (ny, nx), specifying the geolocations of a 2D mesh.
+
+    Input values may be coordinate points (ny, nx) or bounds (ny, nx, 4).
+    However, if points, the edges in the X direction are assumed to be
+    connected by wraparound.
+
+    Input can be either two arrays, two coordinates, or a single cube
+    containing two suitable coordinates identified with the 'x' and'y' axes.
 
     Args:
 
     The inputs (x [,y]) can be any of the folliwing :
 
     * x (:class:`~iris.cube.Cube`):
-        a grid cube with 2D longitude and latitude coordinates.
+        a grid cube with 2D X and Y coordinates, identified by 'axis'.
+        The coordinates must be 2-dimensional with the same shape.
+        The two dimensions represent grid dimensions in the order Y, then X.
 
     * x, y (:class:`~iris.coords.Coord`):
-        longitude and latitude coordinates.
+        X and Y coordinates, specifying grid locations on the globe.
+        The coordinates must be 2-dimensional with the same shape.
+        The two dimensions represent grid dimensions in the order Y, then X.
+        If there is no coordinate system, they are assumed to be true
+        longitudes and latitudes.  Units must convertible to 'degrees'.
 
     * x, y (2-dimensional arrays of same shape (ny, nx)):
         longitude and latitude cell center locations, in degrees.
+        The two dimensions represent grid dimensions in the order Y, then X.
 
     * x, y (3-dimensional arrays of same shape (ny, nx, 4)):
         longitude and latitude cell bounds, in degrees.
+        The first two dimensions are grid dimensions in the order Y, then X.
         The last index maps cell corners anticlockwise from bottom-left.
 
     Optional Args:
@@ -167,9 +209,11 @@ def gridcell_angles(x, y=None, cell_angle_boundpoints='mid-lhs, mid-rhs'):
             the average of the 4 bounds.
 
     """
-    if hasattr(x, 'core_data'):
-        # N.B. only "true" lats + longs will do : Cannot handle rotated !
-        x, y = x.coord('longitude'), x.coord('latitude')
+    cube = None
+    if hasattr(x, 'add_aux_coord'):
+        # Passed a cube : extract 'x' and ;'y' axis coordinates.
+        cube = x  # Save for later checking.
+        x, y = cube.coord(axis='x'), cube.coord(axis='y')
 
     # Now should have either 2 coords or 2 arrays.
     if not hasattr(x, 'shape') and hasattr(y, 'shape'):
@@ -178,7 +222,7 @@ def gridcell_angles(x, y=None, cell_angle_boundpoints='mid-lhs, mid-rhs'):
         raise ValueError(msg.format(type(x), type(y)))
 
     x_coord, y_coord = None, None
-    if isinstance(x, iris.coords.Coord) and isinstance(y, iris.coords.Coord):
+    if hasattr(x, 'bounds') and hasattr(y, 'bounds'):
         x_coord, y_coord = x.copy(), y.copy()
         x_coord.convert_units('degrees')
         y_coord.convert_units('degrees')
@@ -203,9 +247,10 @@ def gridcell_angles(x, y=None, cell_angle_boundpoints='mid-lhs, mid-rhs'):
         else:
             x, y = x_coord.points, y_coord.points
 
-    elif isinstance(x, iris.coords.Coord) or isinstance(y, iris.coords.Coord):
+    elif hasattr(x, 'bounds') or hasattr(y, 'bounds'):
+        # One was a Coord, and the other not ?
         is_and_not = ('x', 'y')
-        if isinstance(y, iris.coords.Coord):
+        if hasattr(y, 'bounds'):
             is_and_not = reversed(is_and_not)
         msg = 'Input {!r} is a Coordinate, but {!r} is not.'
         raise ValueError(*is_and_not)
@@ -282,6 +327,18 @@ def true_vectors_from_grid_vectors(u_cube, v_cube,
     """
     Rotate distance vectors from grid-oriented to true-latlon-oriented.
 
+    .. Note::
+
+        This operation overlaps somewhat in function with
+        :func:`iris.analysis.cartography.rotate_winds`.
+        However, that routine only rotates vectors according to transformations
+        between coordinate systems.
+        This function, by contrast, can rotate vectors by arbitrary angles.
+        Most commonly, the angles are estimated solely from grid sampling
+        points, using :func:`gridcell_angles` :  This allows operation on
+        complex meshes defined by two-dimensional coordinates, such as most
+        ocean grids.
+
     Args:
 
     * u_cube, v_cube : (cube)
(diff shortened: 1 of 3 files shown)


lib/iris/analysis/_grid_angles.py
```diff
@@ -0,0 +1,390 @@
...
+            y_coord = iris.coords.AuxCoord(x, standard_name='latitude',
+                                           units='degrees')
+            x_coord = iris.coords.AuxCoord(y, standard_name='longitude',
+                                           units='degrees')
+    else:
+        # Get lhs and rhs locations by averaging top+bottom each side.
+        # NOTE: so with bounds, we *don't* need full circular longitudes.
+        xyz = _3d_xyz_from_latlon(x, y)
+        angle_boundpoints_vals = {'mid-lhs, mid-rhs': '03_to_12',
+                                  'lower-left, lower-right': '0_to_1'}
```
lbdreyer: Kinda confusing names `03_to_12` and `0_to_1`. I initially thought it may have something to do with a clock.
I understand them now, but I will try to think up something informative

lib/iris/analysis/_grid_angles.py
```diff
@@ -0,0 +1,390 @@
...
+        bounds_pos = angle_boundpoints_vals.get(cell_angle_boundpoints)
+        if bounds_pos == '0_to_1':
+            lhs_xyz = xyz[..., 0]
+            rhs_xyz = xyz[..., 1]
+        elif bounds_pos == '03_to_12':
+            lhs_xyz = 0.5 * (xyz[..., 0] + xyz[..., 3])
+            rhs_xyz = 0.5 * (xyz[..., 1] + xyz[..., 2])
```
lbdreyer: is this sensible?

lib/iris/analysis/_grid_angles.py
```diff
@@ -0,0 +1,390 @@
...
+        else:
+            msg = ('unrecognised cell_angle_boundpoints of "{}", '
+                   'must be one of {}')
+            raise ValueError(msg.format(cell_angle_boundpoints,
+                                        list(angle_boundpoints_vals.keys())))
+        if not x_coord:
```
lbdreyer: Maybe better to do 
`if not x_coord and y_coord`
I'm not sure if you catch the example someone supplying giving `gridcell_angles` a coord and an array as arguments. So maybe it would, instead, be better to check that earlier?

lib/iris/analysis/_grid_angles.py
```diff
@@ -0,0 +1,390 @@
...
+        Valid values are 'lower-left, lower-right', which takes the angle from
+        the lower left to the lower right corner, and 'mid-lhs, mid-rhs' which
+        takes an angles between the average of the left-hand and right-hand
+        pairs of corners.  The default is 'mid-lhs, mid-rhs'.
+
+    Returns:
+
+        angles : (2-dimensional cube)
+
+            Cube of angles of grid-x vector from true Eastward direction for
```
lbdreyer: How about, if the input was a cube, the grid cell angles are an aux coord on the original cube? Or is that too confusing? 

It just seems wasteful to have a copy of the lat and lon coord somewhere else. I guess this is an example of where shared coordinates would be useful

lib/iris/analysis/_grid_angles.py
```diff
@@ -0,0 +1,390 @@
...
+"""
+from __future__ import (absolute_import, division, print_function)
+from six.moves import (filter, input, map, range, zip)  # noqa
+
+import numpy as np
+
+import iris
+
+
+def _3d_xyz_from_latlon(lon, lat):
```
lbdreyer: These functions (` _3d_xyz_from_latlon` and '_latlon_from_xyz') seems quite general purpose functions (not just specific to the case of rotating grid cells)... could we not put them somewhere else so they can be used by other things?

lib/iris/analysis/_grid_angles.py
```diff
@@ -0,0 +1,390 @@
...
+            each gridcell, in radians.
+            It also has longitude and latitude coordinates.  If coordinates
+            were input the output has identical ones :  If the input was 2d
+            arrays, the output coords have no bounds; or, if the input was 3d
+            arrays, the output coords have bounds and centrepoints which are
+            the average of the 4 bounds.
+
+    """
+    cube = None
+    if hasattr(x, 'add_aux_coord'):
```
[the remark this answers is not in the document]
pp-mo: Because duck typing is "Better" :tm:  !

lib/iris/analysis/_grid_angles.py
```diff
@@ -0,0 +1,390 @@
...
+        msg = 'Input {!r} is a Coordinate, but {!r} is not.'
+        raise ValueError(*is_and_not)
+
+    # Now have either 2 points arrays or 2 bounds arrays.
+    # Construct (lhs, mid, rhs) where these represent 3 adjacent points with
+    # increasing longitudes.
+    if x.ndim == 2:
+        # PROBLEM: we can't use this if data is not full-longitudes,
+        # i.e. rhs of array must connect to lhs (aka 'circular' coordinate).
+        # But we have no means of checking that ?
```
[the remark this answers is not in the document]
pp-mo: Yes, but that's how it is.
That's why I mentioned it in the docstring.

lib/iris/analysis/_grid_angles.py
pp-mo: The code checks for "one coord + one non-coord" and outlaws it, [here](https://github.com/pp-mo/iris/blob/6b9bfbbcbaac87b3ade9e050613d17536d9bfd9f/lib/iris/analysis/_grid_angles.py#L250).

lib/iris/analysis/_grid_angles.py
```diff
@@ -0,0 +1,390 @@
...
+        grid_angles_cube = gridcell_angles(u_cube, **grid_angles_kwargs)
+    gridangles = grid_angles_cube.copy()
+    gridangles.convert_units('radians')
+    uu, vv, aa = (cube.data for cube in (u_out, v_out, gridangles))
+    mags = np.sqrt(uu*uu + vv*vv)
+    angs = np.arctan2(vv, uu) + aa
+    uu, vv = mags * np.cos(angs), mags * np.sin(angs)
+
+    # Promote all to masked arrays, and also apply mask at bad (NaN) angles.
+    mask = np.isnan(aa)
```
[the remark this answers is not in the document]
pp-mo: I hadn't thought of that.
I _think_ it doesn't much matter, as the results would just be NaN at those points.

lib/iris/analysis/_grid_angles.py
```diff
@@ -0,0 +1,390 @@
...
+
+    Optional Args:
+
+    * cell_angle_boundpoints (string):
+        Controls which gridcell bounds locations are used to calculate angles,
+        if the inputs are bounds or bounded coordinates.
+        Valid values are 'lower-left, lower-right', which takes the angle from
+        the lower left to the lower right corner, and 'mid-lhs, mid-rhs' which
+        takes an angles between the average of the left-hand and right-hand
+        pairs of corners.  The default is 'mid-lhs, mid-rhs'.
```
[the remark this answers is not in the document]
pp-mo: The original code I was working from (ex-IDL) used only point values.
I extended it to use bounds because that is much more logical, and also works properly for _regional_ data.
What I found was that, while 'mid-lhs..mid-rhs' was a more obvious choice, 'bottom-left..bottom-right' was closer to the 'official' cell angles in the model output file (when included).
So I hedged my bets by making it switchable, but only supported 2 options = 'seems most logical' and 'best match to official values'.

We can perhaps standardise on just the 'mid-lhs..mid-rhs' option, if I could convince myself that the differences are not significant.  I don't really want to make 'bottom-left..bottom-right' the default, as it seems the poorer choice.

lib/iris/analysis/_grid_angles.py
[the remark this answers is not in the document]
pp-mo: ... however, now I'm wondering if we should return masked arrays in every case.
Iris is usually more intelligent than that.

lib/iris/analysis/_grid_angles.py
```diff
@@ -0,0 +1,390 @@
...
+
+    elif hasattr(x, 'bounds') or hasattr(y, 'bounds'):
+        # One was a Coord, and the other not ?
+        is_and_not = ('x', 'y')
+        if hasattr(y, 'bounds'):
+            is_and_not = reversed(is_and_not)
+        msg = 'Input {!r} is a Coordinate, but {!r} is not.'
+        raise ValueError(*is_and_not)
+
+    # Now have either 2 points arrays or 2 bounds arrays.
```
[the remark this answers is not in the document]
pp-mo: ok!

lib/iris/analysis/_grid_angles.py
pp-mo: We can have gridcell angles in the file, but as they are not a CF concept they never appear as an aux-coord, but instead just as a separate cube.

Talking of inefficient, of course the "angles" cube has another copy of the lats + lons with all their bounds!
We really do need to fix this ...

lib/iris/analysis/_grid_angles.py
```diff
@@ -0,0 +1,390 @@
...
+    lats = np.arctan2(xyz[2], axial_radii)
+    return np.array([lons, lats])
+
+
+def _angle(p, q, r):
+    """
+    Return angle (in _radians_) of grid wrt local east.
+    Anticlockwise +ve, as usual.
+    {P, Q, R} are consecutive points in the same row,
+    eg {v(i,j),f(i,j),v(i+1,j)}, or {T(i-1,j),T(i,j),T(i+1,j)}
```
lbdreyer: Should this not be:
`{v(i,j),v(i,j),v(i+1,j)}`

Why are `v` and `T` used here? 

Also, I assume these function docs strings are not quite finalised yet? So I will reserve comments on the wording till then

lib/iris/analysis/_grid_angles.py
```diff
@@ -0,0 +1,390 @@
...
+    Calculate dot product of PR with lambda_hat at Q.
+    This gives us cos(required angle).
+    Disciminate between +/- angles by comparing latitudes of P and R.
+    p, q, r, are all 2-element arrays [lon, lat] of angles in degrees.
+
+    """
+#    old_style = True
+    old_style = False
+    if old_style:
```
lbdreyer: How do you plan to determine whether or not old_style should be used?

lib/iris/analysis/_grid_angles.py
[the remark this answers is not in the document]
pp-mo: >  standardise on just the 'mid-lhs..mid-rhs' option, if I could convince myself that the differences are not significant

But unfortunately I checked, and the discrepancies *are* significant in some areas.  At least a few degrees for practical purposes.
Also the errors for quasi-ideal square-ish gridcells rotated at various angles seem to be ~linear in the size of the cells for the 'bottom-left..bottom-right' method while ~square for the 'mid-lhs..mid-rhs'.
So that means I'm not comfortable 'standardising' on either of these exclusively :disappointed:

lib/iris/analysis/_grid_angles.py
pp-mo: This is inherited docs + it does need improving.
Let's defer that :

lib/iris/analysis/_grid_angles.py
pp-mo: Sorry, this was an alternative calculation that I used for evaluation.
I have since removed it (cf. https://github.com/pp-mo/iris/pull/45)

Though... I accept that I ***still*** don't have a full understanding of this calculation :grimacing: .
We should probably make more effort to explain exactly what this is doing.
Put that also down to #3109 !

lib/iris/analysis/_grid_angles.py
corinnebosley: @pp-mo have you logged this suggestion in a labelled issue yet?  Might be better to merge first and think about where things should live later.

lib/iris/analysis/_grid_angles.py
```diff
@@ -0,0 +1,390 @@
...
+
+
+def _latlon_from_xyz(xyz):
+    """
+    Return arrays of lons+lats angles from xyz locations.
+
+    Args:
+
+    * xyz: (array)
+        positions array, of dims (3, <others>), where index 0 maps x/y/z.
```
corinnebosley: This is a bit confusing.  I know that the docs in general are not finished though, so I won't go on.

lib/iris/analysis/_grid_angles.py
```diff
@@ -0,0 +1,390 @@
...
+
+    Returns:
+
+        lonlat : (array)
+            spherical angles, of dims (2, <others>), in radians.
```
corinnebosley: so many commas...

pp-mo: We need to work out how to capture outstanding review issues on things we merged in advance.
Perhaps like this ... #3114 ?

PR #3097, opened 2018-07-26
2d coord plotting

corinnebosley:


lib/iris/coords.py
```diff
@@ -143,6 +143,85 @@ def __new__(cls, name_or_coord, minimum, maximum,
                                       'groupby_point, groupby_slice')
 
 
+def _discontinuity_in_2d_bounds(bds, abs_tol=1e-4):
```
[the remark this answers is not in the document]
corinnebosley: Yeah maybe.  Not sure if it needs to be private either.

lib/iris/plot.py
```diff
@@ -65,7 +65,7 @@ def names(coords):
             if isinstance(coord, int):
                 result.append('dim={}'.format(coord))
             else:
-                result.append(coord.name())
+                result.append(cube.coord.name())
```
lbdreyer: This should go back to 
`result.append(coord.name()`

lib/iris/plot.py
```diff
@@ -256,17 +274,87 @@ def _invert_yaxis(v_coord, axes=None):
...
-        plot_defn = _get_plot_defn_custom_coords_picked(cube, coords, mode)
+        plot_defn = _get_plot_defn_custom_coords_picked(cube, coords, mode,
+                                                        ndims=2)
     else:
         plot_defn = _get_plot_defn(cube, mode, ndims=2)
 
+    two_dim_contig_atol = kwargs.pop('two_dim_coord_contiguity_atol',
+                                     1e-4)
+    for coord in plot_defn.coords:
+        if coord.ndim == 2 and coord.has_bounds():
```
lbdreyer: coord isn't necessarily a `iris.coord.Coord` object at this point. It could instead be `None` or a string.
So first we need to check that it is a coord. I would suggest putting a line that does:
 `if isinstance(cube, iris.coords.Coord)`, 
before this line, but apparently it is better practice to do 'duck-typing'. So instead you could do a check such as:
'if hasattr(coord, 'has_bounds'):
if hasattr(coord, 'add_aux_coord')`

lib/iris/plot.py
```diff
@@ -256,17 +274,87 @@ def _invert_yaxis(v_coord, axes=None):
...
+        data:
+            data of the the cube we are plotting
+        abs_tol:
+            tolerance when checking the contiguity
+
+    """
+    if transpose:
+        bounds = coord.bounds.T
+    else:
+        bounds = coord.bounds
```
lbdreyer: I do find this really weird. I would have expected it to be the other way around!

PR #3099, opened 2018-07-30
Provide a test skipper for 2d coords WIP.

pp-mo:
Targetting new feature branch ["2d_coords"](https://github.com/SciTools/iris/tree/2d_coords).
Establishes a basis for 2d coords work, with broken tests skipped out while WIP.


pp-mo:
We must remove all the skips, and the skipper itself when we are done.
See : #3102

Issue #3114, opened 2018-07-31
Outstanding review issues from #3096 (gridcell rotations)


pp-mo:
Premature merge.
Outstanding comments :
  * lots of stickler complaints
  * https://github.com/SciTools/iris/pull/3096#pullrequestreview-141076488
  * maybe reconsider https://github.com/SciTools/iris/pull/3096#discussion_r205756026
  * ? https://github.com/SciTools/iris/pull/3096#discussion_r205767222
  * ? consider : combine with other code for 3d latlon/xyz transform ?  https://github.com/SciTools/iris/pull/3096#discussion_r205778183
  * inherited + other comments to clarify
    *  https://github.com/SciTools/iris/pull/3096#discussion_r206112480
    * https://github.com/SciTools/iris/pull/3096#pullrequestreview-141902153
    * https://github.com/SciTools/iris/pull/3096#discussion_r206483318
  * should remove make-discontiguous routine : see https://github.com/SciTools/iris/pull/3127#discussion_r207522639

---

Missing test aspects : originally from #3103

CODE CHANGES
   * ~angles routine should handle non-2d coords with meshgrid~ 
      * Too tricky : ABANDONED ? but [this reduces the generality of "rotate_grid_angles"](https://github.com/SciTools/iris/issues/3145)

ANGLES TEST ASPECTS : **testing of 'grid_angles'**
   * **THINK THIS IS ALL NOW COVERED BY #3120** (check when that is accepted)
   *  (cube)
   *  unrecognised inputs (no shapes)
   *  missing y + x not cube
   *  coords
      *    not 2d : (support?)
      *    different shapes : error ~(except 1d?)~
      *    different cube dims ~(2d : ok for 1d)~
      *    different cs : error
      *    no cs, but not angles, i.e. plain X+Y coords : ??
      *    X/Y = coord/not, ***and*** not/coord
      *    cell_angle_boundpoints unrecognised

ROTATE TEST ASPECTS : **testing of 'rotate_grid_vectors'**
   * u+v
   * u+v with angles supplied
   * u+v with angles derived from u
   * angles method key
   * check that magnitudes remain unchanged
   * check actual angles-rotation calculation (known vector + angles)
   * ~check that non-2d cubes work~ (not supported now ?)
   * check masking of invalid points
   * check behaviour with 0-vectors ?
   * Exceptions:
     * angles not angular (bad unit)
     * u+v not same shape (implied by arithmetic)
[...]

Issue #3109, opened 2018-07-30
Improve comments describing calculation in iris.analysis.cartography.grid_angles


pp-mo:
See [this comment](https://github.com/SciTools/iris/pull/3096#discussion_r206186346)
And [this one](https://github.com/SciTools/iris/pull/3096#discussion_r206189359)

pp-mo:
Dealt with in #3120

PR #3120, opened 2018-07-31
Vector plots 2

pp-mo:
**WIP : not yet ready**

Followon work from #3096 
  * more review issues : see #3114
  * PLUS : tidy up + add some proper tests

~**PLEASE NOTE**~
~contents of `...tests/unit/analysis/cartography/test_gridcell_angles.py` are still mostly junk~
~Please await more commits before commenting !~
(now more sensible, though more tests still wanted)

### DIFF
--- a/lib/iris/analysis/_grid_angles.py
+++ b/lib/iris/analysis/_grid_angles.py
@@ -24,6 +24,7 @@ from six.moves import (filter, input, map, range, zip)  # noqa
 
 import numpy as np
 
+import cartopy.crs as ccrs
 import iris
 
 
@@ -183,8 +184,6 @@ def gridcell_angles(x, y=None, cell_angle_boundpoints='mid-lhs, mid-rhs'):
     if hasattr(x, 'bounds') and hasattr(y, 'bounds'):
         # x and y are Coords.
         x_coord, y_coord = x.copy(), y.copy()
-        x_coord.convert_units('degrees')
-        y_coord.convert_units('degrees')
         if x_coord.ndim != 2 or y_coord.ndim != 2:
             msg = ('Coordinate inputs must have 2-dimensional shape. ',
                    'Got x-shape of {} and y-shape of {}.')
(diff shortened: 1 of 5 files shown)


lbdreyer:
@pp-mo I think you also pushed this branch directly to upstream as we have a upstream/vector_plots_2d branch. I assume this was a mistake?

pp-mo:
> I think you also pushed this branch directly to upstream as we have a upstream/vector_plots_2d branch. I assume this was a mistake?

I only did that when testing the auto-build of docs.
It actually wasn't this branch but something different with the same name !
(now removed).

So, not relevant to this.

pp-mo:
Note: now rebased onto 'pin_dask', so we can hopefully have working tests (for now!).

This should look a lot more sensible.

However, it is still lacking some tests of the `grid_angles` method, like failure methods and some features.
In principle, it can use a grid with a rotated or even non-latlon coord system, but that quite possibly does not even work yet, and could even instead be removed ? Undecided.

Also, no tests yet for `rotate_grid_vectors` function.

pp-mo:
hi @lbdreyer tests now passing : ready for a look-over ?

lib/iris/analysis/_grid_angles.py
```diff
@@ -24,6 +24,7 @@
 
 import numpy as np
 
+import cartopy.crs as ccrs
 import iris
```
lbdreyer: Potential changes to the `_3d_xyz_from_latlon` function
```
    Returns:
        xyz : (array, dtype=float64)
            cartesian coordinates on a unit sphere.  Dimension 0 maps x,y,z.
```
Clarify the last sentence? `Dimension 0 maps x,y,z`

`lon1` and `lat1` -> `lon_rad` and `lat_rad`? seems a little more informative

lib/iris/analysis/_grid_angles.py
```diff
@@ -65,13 +66,13 @@ def _latlon_from_xyz(xyz):
     Returns:
 
         lonlat : (array)
-            spherical angles, of dims (2, <others>), in radians.
+            spherical angles, of dims (2, <others>), in degrees.
             Dim 0 maps longitude, latitude.
```
lbdreyer: Here's this `Dim - maps longitude, latitde` again. I don't know what that means?

lbdreyer: (I do prefer that this is working in degrees)

lib/iris/analysis/_grid_angles.py
```diff
@@ -65,13 +66,13 @@ def _latlon_from_xyz(xyz):
     Returns:
 
         lonlat : (array)
-            spherical angles, of dims (2, <others>), in radians.
+            spherical angles, of dims (2, <others>), in degrees.
```
lbdreyer: I do prefer that this is working in degrees to be consistent with the `_3d_xyz_from_latlon` function

pp-mo: hi @lbdreyer 
I'll look into these now if you like, 
but in principle I had already deferred these "comments on the comments" to a follow-on : #3109, also linked from #3114.

I have been working on the follow-up, to address those remaining issues + add missing tests ...

pp-mo:
Latest tries to improve docstrings, 
Is probably overkill (!)

pp-mo:
Hi @lbdreyer :bouquet: :champagne: for checking the trigonometry code !
I'm hoping my latest addresses all your outstanding points.

lbdreyer:
@pp-mo Thanks for this! The changes you've added all look fine so I'll squash and merge this in

PR #3127, opened 2018-08-02
Pin Dask to avoid 0.18.2 bug with masked arrays.

pp-mo:
Basically, hacking various things to get the tests passing again.
It starts by pinning dask, but that uncovered multiple other problems which have crept in since we merged stuff with failing tests ...


lib/iris/tests/unit/plot/test_2d_coords.py
```diff
@@ -22,29 +22,34 @@
...
     def setUp(self):
         self.multidim_cube = cube_2dcoords()
         self.overspan_cube = cube3d_2dcoords()
 
         # latlon_2d is a cube with 2d coords, 4 bounds per point,
         # discontiguities in the bounds but masked data at the discontiguities.
-        self.latlon_2d = testdata.full2d_global()
-        testdata.make_bounds_discontiguous_at_point(self.latlon_2d, 2, 2)
+        self.latlon_2d = full2d_global()
+        make_bounds_discontiguous_at_point(self.latlon_2d, 2, 2)
```
pp-mo: Only remaining usage of this routine : see https://github.com/SciTools/iris/pull/3127#discussion_r207522639 
Can remove later

lib/iris/tests/stock/_stock_2d_latlons.py
```diff
@@ -0,0 +1,281 @@
...
+        else:
+            # "Just" rotate operation : add a coord-system to each coord.
+            cs = RotatedGeogCS(pole_lat, pole_lon)
+            for co in cube.coords():
+                co.coord_system = cs
+
+    return cube
+
+
+def make_bounds_discontiguous_at_point(cube, at_iy, at_ix):
```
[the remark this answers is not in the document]
pp-mo: This *is* still used within this version [here](https://github.com/SciTools/iris/pull/3127/files#r207562522)

I'm still trying for minimal changes here to get the main branch passing tests !
I agree we probably don't have a need for this now - we can+should remove it later.
Added to todo list in #3114

lib/iris/tests/stock/_stock_2d_latlons.py
```diff
@@ -0,0 +1,281 @@
...
+
+def grid_coords_2d_from_1d(x_coord_1d, y_coord_1d):
+    """
+    Calculate a pair of 2d X+Y coordinates from 1d ones, in a "meshgrid" style.
+    If the inputs are bounded, the outputs have 4 points per bounds in the
+    usual way, i.e. points 0,1,2,3 are the gridcell corners anticlockwise from
+    the bottom left.
+
+    """
+    for co in (x_coord_1d, y_coord_1d):
```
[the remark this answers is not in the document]
pp-mo: Ok

pp-mo:
@lbdreyer I hope this fixes all the issues raised.

I :trumpet: have now tracked down the key problem + raised it against Dask [here](https://github.com/dask/dask/issues/3848) .

lib/iris/tests/stock/_stock_2d_latlons.py
```diff
@@ -0,0 +1,340 @@
...
+         [ 74.197  60.258  51.325  51.016  59.446  73.268]]
+        >>>
+
+    """
+    def sample_cube(xargs, yargs):
+        # Arguments are to make coord points with np.linspace : start/stop/N.
+        x0, x1, nx = xargs
+        y0, y1, ny = yargs
+        data = np.zeros((ny, nx))
+        data.flat[:] = np.arange(ny * nx) % (nx + 2)  # seems to work ok
```
lbdreyer: Did you want to keep this comment?

lib/iris/tests/stock/_stock_2d_latlons.py
```diff
@@ -0,0 +1,340 @@
...
+    if regional:
+        # Extract small region.
+        cube = sample_cube(xargs=(150., 243.75, 6), yargs=(-10., 40., 5))
+
+        # Make contiguous bounds.
+        for ax in ('x', 'y'):
+            cube.coord(axis=ax).guess_bounds()
+    else:
+        # Global data, but drastically reduced resolution.
+
```
lbdreyer: Can you remove this line?

lib/iris/tests/stock/_stock_2d_latlons.py
```diff
@@ -0,0 +1,340 @@
...
+        if transformed:
+            # Reproject coordinate values from rotated to true lat-lons.
+            co_x, co_y = [cube.coord(axis=ax) for ax in ('x', 'y')]
+            # Unrotate points.
+            lons, lats = co_x.points, co_y.points
+            lons, lats = unrotate_pole(lons, lats, pole_lon, pole_lat)
+            co_x.points, co_y.points = lons, lats
+            # Unrotate bounds.
+            lons, lats = co_x.bounds, co_y.bounds
+            shape = lons.shape  # unrotate pole is restricted to 1- or 2-d ??
```
lbdreyer: If it doesn't support 2d it will be a problem on line 278 where the coordinates will already be 2d

lbdreyer: It sounds like unrotate_pole can handle 2d arrays. unrotate_pole calls transform_points (in Cartopy) which accepts 1d or 2d (https://github.com/SciTools/cartopy/blob/25c786ebe60f3e7c45c1cfb283a6b150c7c8a931/lib/cartopy/_crs.pyx#L326)

lbdreyer: Could you remove this comment?

lib/iris/tests/stock/_stock_2d_latlons.py
pp-mo: Ok, will clarify.

lib/iris/tests/stock/_stock_2d_latlons.py
pp-mo: ok!

lib/iris/tests/stock/_stock_2d_latlons.py
pp-mo: It's just that `unrotate_pole` makes an unnecessary check for 1- or 2-d array inputs, whereas it could just work elementwise + not worry about dimensions, which is what we want here.
So I'm flattening + then re-applying the shape afterwards, to support any shape of arrays.

Will try to clarify comments !

lbdreyer:
#3134  is in, can you rebase the `2d_coords` branch?

pp-mo:
> #3134 is in, can you rebase the 2d_coords branch?

We shouldn't rebase a feature branch if we can avoid it.
The correct thing to do is rebase ***this***, then merge it.
I'll see whether that works out ok ...

pp-mo:
Ok @lbdreyer now rebased to bring in the recent tests fix from #3134 
Also addressed your most recent comments (I hope).

Please re-review.

pp-mo:
> We shouldn't rebase a feature branch if we can avoid it.
The correct thing to do is rebase this, then merge it.

No, it ***is*** right to rebase the feature branch.
Otherwise it won't merge back cleanly in future.

Will come back here after doing that ...
**The current state here is a mistake.**

pp-mo:
Ok, now rebased onto the new 2d_coords branch, lately updated from master to include #3134
Please re-check @lbdreyer

lbdreyer:
Okay looks all fine :+1:

PR #3134, opened 2018-08-13
Temporary dask fix

pp-mo:
Pin dask + fix minor dask usage glitch, to get tests passing again ( :crossed_fingers: ) .

Changes are basically cherrypicked from #3127 

**NASTY PROBLEM :**
Recent numpy (1.15 ?) uses different formatting for printing arrays in Python 2 + Python 3.
Alternative print solutions make the sample code very messy (as discussed offline with @lbdreyer).

I've fixed this for now by adopting the Python 3 output, and ***disabling the doctests on Python 2***
However that is pretty wasteful, as a whole row of the Travis test matrix now does almost nothing.
I expect there is some way of controlling that, but for now I don't care enough.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -128,7 +128,7 @@ script:
   - INSTALL_DIR=$(pwd)
 
   - >
-    if [[ $TEST_TARGET == 'doctest' ]]; then
+    if [[ $TEST_TARGET == 'doctest' && ${TRAVIS_PYTHON_VERSION} == 3* ]]; then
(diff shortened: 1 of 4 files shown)


pp-mo:
Hi @lbdreyer this is finally passing again.
Can you check this out + maybe merge in :pray: ?

I will then replace #3127 with a version rebased onto this ...

lbdreyer:
I like the simplicity of this fix so I'm going to merge this.

lbdreyer:
I should have squashed the commits! Sorry!

PR #3135, opened 2018-08-13
Extending ORCA blockplotting (w/ tests)

lbdreyer:
This adds some changes to the testing of the 2d blockplotting (and updates some things that fell out from testing it.)

I have focused on tests for the changes to iris/coords.py so far but will add tests for the changes to iris/plot.py once I have them (and once we have decided on what to do about discontiguity checking)


lib/iris/tests/unit/plot/test_2d_coords.py
```diff
@@ -1,157 +0,0 @@
-# (C) British Crown Copyright 2018, Met Office
```
lbdreyer: I have removed this file as it had mixed unit tests for lots of different functions.
Unit tests should be separated so that each file tests a specific function/class (e.g. a single file for testing iris.plot._get_plot_defn)

I have moved the tests in this file to the appropriate files for *almost* all of them. Have not yet done it for the plot.py functions but will also push that up to this PR once that is done

lbdreyer:
@pp-mo @corinnebosley 
Could someone look at this?

I have changed:
* Docs strings for a couple functions
* added support to cubes that may be of shape (2,1) or (1,2) or (1) to the discontiguity check
* changes `contiguous bounds` to check that the bounds are contiguous (before it was assumed)*
* Added a check in plot.py that the coordinates have the same number of dimensions (to stop someone doing a 1D coord against a 2D coord
* Removed the ability to modify the absolute tolerance when calling pcolormesh (instead use default)
* Added unit tests for:
  * coord methods`contiguous_bounds`, `is_contiguous` and `sanity_check_bounds`
  * `coords._discontiguity_in_2d_bounds`
  * `coords._get_2d_coord_bound_grid`
  * `plot.get_plot_defn`
  * `plot.get_plot_defn_custom_coords_picked`
* Added integration test for pcolormesh - will also need to add the expected graphic for this 

Note I have not written any tests for _check_contiguity_and_bounds or modified how we call that (other than to remove the abs_tol thing) but it does need fixing as currently it isn't right (we're calling it too many times) but this may be covered by what @corinnebosley is doing so I thought I'd leave it.

* Doing this has meant that when checking for discontiguities we are using the default absolute tolerance value which is 1e-8 which may be too small (it is too small for our ORCA2 test data).

pp-mo:
> it is too small for our ORCA2 test data

I'm worried about the practical usability of this.
If I understand right, between this and the bounds-check-routine that @corinnebosley is working on, this means that :
  * some ORCA data will generate warnings when plotted, because it has quite large discontiguities at certain points
  * the new contig-check routine will let you check the bounds with adjustable tolerance
  * but the only provided way to *avoid* the plot warning is to mask out the problem points. Which in fact you don't necessarily want to do.

So, I think it is necessary for the plot routine to either have a much more generous default tolerance, or a way to modify it.

IMHO the only point of warnings is to encourage people to take action to avoid them, so there should always _be_ a way to avoid it, and ideally the warning tells you how :  If there isn't a way out, then you are just giving more encouragement to ignore them !

lib/iris/tests/unit/plot/test_2d_coords.py
corinnebosley: Cool, it was on my list of things to do.  In my own PR I have shuffled some tests around and added some more for the new utils, but these were always intended to be cleaned up anyway, it's just a bit easier to run tests when they're in the same place.  

I guess I can pull in your changes and then move the rest of the tests round accordingly, because none of them are in the right place.

lib/iris/coords.py
```diff
@@ -1043,9 +1066,9 @@ def _sanity_check_contiguous(self):
                                  'coordinates with 4 bounds.'.format
                                  (self.name(), self.nbounds))
         else:
-            raise ValueError('Invalid operation for {!r}. Contiguous bounds '
-                             'are not defined for coordinates with more than '
-                             '2 dimensions.'.format(self.name()))
+            raise ValueError('Invalid operation for {!r}. Not supported for '
+                             'bounds with more than 2 dimensions.'.format
+                             (self.name()))
```
corinnebosley: I'm not sure about this error message.  It seems slightly more confusing than it was previously.

lib/iris/coords.py
```diff
@@ -1064,28 +1087,25 @@ def is_contiguous(self, rtol=1e-05, atol=1e-08):
...
-                                                               abs_tol=atol)
+
+                contiguous, _, _ = _discontiguity_in_2d_bounds(
+                    self.bounds, atol=atol)
         else:
             contiguous = False
         return contiguous
 
     def contiguous_bounds(self):
         """
```
corinnebosley: Docstring is better (I like things which are concise) but I'm still not taken with the name.  'contiguous_bounds' is just a bit vague.

lib/iris/coords.py
```diff
@@ -1097,7 +1117,10 @@ def contiguous_bounds(self):
                           'contiguous bounds.'.format(self.name()))
             bounds = self._guess_bounds()
         else:
-            self._sanity_check_contiguous()
+            if not self.is_contiguous():
+                raise ValueError('Invalid operation. Bounds of coord {!r} '
+                                 'are not contiguous.'.format(self.name()))
+
```
corinnebosley: Nice catch.  Much better.

lib/iris/plot.py
```diff
@@ -102,10 +102,20 @@ def get_span(coord):
...
+                             'Cell-based plotting is only supported for'
                              'coordinates with one or two dimensions.'
                              .format(coord.name()), len(span))
 
+    # Check the coords have the same number of dimensions.
+    if ndims == 2:
+        if not (len(spans[0]) == len(spans[1])):
+            raise ValueError('The coordinate {!r} has {} dimensions and the '
+                             'coordinate {!r} has {} dimensions. Cell-based '
+                             'plotting is only supported for coordinates with '
```
corinnebosley: I'd like to discuss this with you, because I am a little unsure of what cell-based plotting means. Also whether it is clear to users what cell-based plotting is.  And whether it matters, I suppose.

lib/iris/plot.py
```diff
@@ -273,10 +283,11 @@ def _invert_yaxis(v_coord, axes=None):
 
 def _check_contiguity_and_bounds(coord, data, abs_tol=1e-4, transpose=False):
     """
-    Check that the discontinuous bounds occur where the data is masked.
+    Checks that any discontiguity in the bounds of the coordinate only occur
+    where the data is masked.
```
corinnebosley: This is neater and more accurate, but I have made some changes to the name and docstring for this function as well.  Probably worth agreeing on changes.

lib/iris/tests/unit/coords/test_Coord.py
```diff
@@ -355,6 +356,112 @@ def test_different_array_attrs_incompatible(self):
...
+
+    def test_1d_coord_discontiguous(self):
+        coord = iris.coords.DimCoord(self.points, bounds=[[0, 2], [3, 4]])
+        result = coord.is_contiguous()
+        self.assertFalse(result)
+
+    def test_1d_coord_contiguous_atol(self):
+        coord = iris.coords.DimCoord(self.points/1000, bounds=self.bounds/1000)
+        result = coord.is_contiguous(atol=1e-2)
+        self.assertTrue(result)
```
corinnebosley: Might need you to explain this one.

lib/iris/tests/unit/coords/test_Coord.py
```diff
@@ -355,6 +356,112 @@ def test_different_array_attrs_incompatible(self):
...
+
+
+class Test__sanity_check_bounds(tests.IrisTest):
+    def test_coord_1d_2_bounds(self):
+        coord = iris.coords.DimCoord([0, 1], standard_name='latitude',
+                                     bounds=np.array([[0, 1], [1, 2]]))
+        coord._sanity_check_bounds()
```
corinnebosley: This is just to check that no error is raised is it?

lib/iris/tests/unit/coords/test_Coord.py
```diff
@@ -355,6 +356,112 @@ def test_different_array_attrs_incompatible(self):
...
+               "bounds are only defined for 2D coordinates with 4 bounds."
+        with self.assertRaisesRegexp(ValueError, emsg):
+            coord._sanity_check_bounds()
+
+    def test_coord_2d_2_bounds(self):
+        coord = iris.coords.AuxCoord([0, 1], standard_name='latitude',
+                                     bounds=np.array([[0], [1]]))
+        emsg = "Invalid operation for 'latitude', with 1 bounds. Contiguous " \
+               "bounds are only defined for 1D coordinates with 2 bounds."
+        with self.assertRaisesRegexp(ValueError, emsg):
```
corinnebosley: Shouldn't this raise an error about 2d coordinates needing 4 bounds, rather than 1d coordinates needing 2 bounds?

lib/iris/tests/unit/coords/test__discontiguity_in_2d_bounds.py
```diff
@@ -0,0 +1,143 @@
...
+            _discontiguity_in_2d_bounds(bds)
+
+    def test_contiguous_both_dirs(self):
+        bds = self.lon_bounds_3by3
+        all_eq, diffs_along_x, diffs_along_y = _discontiguity_in_2d_bounds(bds)
+        self.assertTrue(all_eq)
+        self.assertTrue(not diffs_along_x.any())
+        self.assertTrue(not diffs_along_y.any())
+
+    def test_discontiguous_along_x(self):
```
corinnebosley: A short comment about what you are doing to the bounds array here might be useful.

lib/iris/tests/unit/coords/test__discontiguity_in_2d_bounds.py
```diff
@@ -0,0 +1,143 @@
...
+        bds = self.lon_bounds_3by3[:, :1, :]
+        all_eq, diffs_along_x, diffs_along_y = _discontiguity_in_2d_bounds(bds)
+        self.assertTrue(all_eq)
+        self.assertIsNone(diffs_along_y)
+        self.assertTrue(not diffs_along_x.any())
+
+    def test_contiguous_abs_tol(self):
+        bds = self.discontiguous_bds / 1000
+        x_exp_result = np.array([1, 0]).reshape(2, 1) / 1000
+        y_exp_result = np.array([1, 0]).reshape(1, 2) / 1000
```
corinnebosley: Don't really understand what's happening here.

lib/iris/coords.py
lbdreyer: I think I changed this because I wanted the method `_sanity_check_bounds` to be able to be used for other things, but the previous error message.
Saying that, I realise I have made a lot of breaking changes in this PR, the above being one of them so I am going to undo that and put the error message back

lib/iris/coords.py
lbdreyer: Unfortunately we don't have much say with changing the name of `contiguous_bounds`. It was already part of the API so changing it would be a breaking change.

lib/iris/coords.py
lbdreyer: Unfortunately this is a breaking change so I am going to have to put this back

lib/iris/tests/unit/coords/test_Coord.py
lbdreyer: Yep I'll add a comment explaining that

lbdreyer:
**Summary of changes**

Discontiguity checking:
* Moved `_discontiguity_in_2d_bounds` function to a method on the coord so it could easily call other coord methods
* Rewrote `_discontiguity_in_2d_bounds` -> `_discontiguity_in_bounds` so that it supports 1D and 2D coords, and calls np.allclose
* `coord.is_contiguous` now calls `coord._discontiguity_in_bounds`. 1D coordinates are still handled as before (so this doesn't introduce a breaking change) but the handling has now been moved into  `_discontiguity_in_bounds` 

2d coord plotting:
* Introduced `contiguity_tolerance` kwarg to pcolor and pcolormesh
  * If it **is set**, the bounds of both 1D and 2D coords are checked for discontiguities using an absolute toelrance as set
  * If it **isn't set**, 1D coords are not checked, but 2D corods are checked with the defaults
* If discontiguities are found, an error is raised

Missing from this PR:
* An integration test but that will be easier to include in a separate PR

lbdreyer:
@pp-mo The tests are now passing, do you mind having a look?

lib/iris/plot.py
```diff
@@ -271,54 +271,78 @@ def _invert_yaxis(v_coord, axes=None):
...
-            tolerance when checking the contiguity
+        coord: (iris.coord.Coord)
+            Coordinate the bounds of which will be checked for contiguity
+        data: (array)
+            Data of the the cube we are plotting
+        atol:
+            Absolute tolerance when checking the contiguity. Defaults to None.
+            If an absolute tolerance is not set, 1D coords are not checked (so
+            as to not introduce a breaking change without a major release) but
+            2D coords are checked with a relative tolerance.
```
pp-mo: > but 2D coords are checked with a relative tolerance.

This is not quite clear.  Maybe something like 
"but 2D coords are always checked, by calling `coord._discontiguity_in_bounds` with its default tolerance"
(If I understood it right)

lib/iris/coords.py
```diff
@@ -143,81 +143,43 @@ def __new__(cls, name_or_coord, minimum, maximum,
...
     e.g.
     # 0-0-0-0-1
     # 0-0-0-0-1
     # 3-3-3-3-2
 
-
     Args:
-        bounds: array of shape (X,Y,4)
+    * bounds: (array)
+        Coordinate bounds array of shape (X,Y,4)
```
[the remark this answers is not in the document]
lbdreyer: Ahh yeah sorry about that! I always get this the wrong way round!

lib/iris/coords.py
```diff
@@ -1064,40 +1090,33 @@ def is_contiguous(self, rtol=1e-05, atol=1e-08):
...
-        Assumes input is contiguous.
-
         .. note::
 
             If the coordinate does not have bounds, this method will
             return bounds positioned halfway between the coordinate's points.
 
+        .. note::
+
+            Assumes bounds are contiguous
```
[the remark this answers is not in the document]
lbdreyer: But if the 1d coord was already bounded it would assume the bounds are contiguous. I will clarify this point

lib/iris/tests/unit/coords/test_Coord.py
```diff
@@ -355,6 +356,279 @@ def test_different_array_attrs_incompatible(self):
...
+                         bounds=self.lon_bounds_3by3[:1, :1, :])
+        contiguous, diffs = coord._discontiguity_in_bounds()
+        diffs_along_x, diffs_along_y = diffs
+        expected_diffs = np.array([], dtype=np.int64)
+        self.assertTrue(contiguous)
+        self.assertArrayEqual(diffs_along_x, expected_diffs.reshape(1, 0))
+        self.assertArrayEqual(diffs_along_y, expected_diffs.reshape(0, 1))
+
+    def test_2d_one_cell_along_x(self):
+        # Test a 2D coord with a single cell along the x axis, where the coord
```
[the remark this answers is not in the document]
lbdreyer: yep! got confused about (y,x) ordering again

lib/iris/coords.py
```diff
@@ -1064,40 +1090,33 @@ def is_contiguous(self, rtol=1e-05, atol=1e-08):
...
         Returns the N+1 bound values for a contiguous bounded 1D coordinate
-        of length N.
-
-        Returns the (N+1, M+1) bound values for a contiguous bounded 2D
+        of length N, or the (N+1, M+1) bound values for a contiguous bounded 2D
         coordinate of shape (N, M).
 
-        Assumes input is contiguous.
-
         .. note::
```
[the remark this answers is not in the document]
lbdreyer: > Possibly worth stating that this fails for coords which are neither 1D nor 2D.

Will do!

> Or possibly make it return 'None', and say so.

Not keen on this. Is this not a breaking change.

lib/iris/tests/unit/coords/test_Coord.py
```diff
@@ -355,6 +356,279 @@ def test_different_array_attrs_incompatible(self):
...
+        expected = np.array([1, 3, 5, 7])
+        result = coord.contiguous_bounds()
+        self.assertArrayEqual(result, expected)
+
+    def test_1d_coord_discontiguous(self):
+        coord = iris.coords.DimCoord([2, 4, 6], standard_name='latitude',
+                                     bounds=[[1, 3], [4, 5], [5, 7]])
+        expected = np.array([1, 4, 5, 7])
+        result = coord.contiguous_bounds()
+        self.assertArrayEqual(result, expected)
```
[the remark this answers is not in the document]
lbdreyer: I didn't include tests for 2d here is because `is_contiguous` is effectively a wrapper for calling `_discontiguity_in_bounds` and has things were already being tested in `Test__discontiguity_in_bounds` it seemed repetitive to also have them here.

I added the 1d cases to show that the behaviour is as before (before 1d was handled in `_discontiguity_in_bounds`.

I will add a couple of simple 2d examples and hopefully that will be enough testing.

lib/iris/tests/unit/coords/test_Coord.py
```diff
@@ -355,6 +356,279 @@ def test_different_array_attrs_incompatible(self):
...
+                                     bounds=self.discontiguous_bounds)
+        result = coord.is_contiguous()
+        self.assertFalse(result)
+
+    def test_1d_coord_contiguous_atol(self):
+        # Test setting a very large tolerance
+        coord = iris.coords.DimCoord(self.points,
+                                     bounds=self.discontiguous_bounds)
+        result = coord.is_contiguous(atol=1e1)
+        self.assertTrue(result)
```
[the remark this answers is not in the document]
lbdreyer: By: 
> check that it uses '_discontinuity_in_bounds'

Do you mean to also remove the 1d cases and just have the test for non bounded coords and a test that it calls  `_discontinuity_in_bounds`?

lib/iris/plot.py
```diff
@@ -390,8 +408,8 @@ def _draw_2d_from_bounds(draw_method_name, cube, *args, **kwargs):
 
         u, v = plot_arrays
 
-        # If the data is tranposed, 2D coordinates will also need to be
-        # tranposed.
+        # If the data is transposed, 2D coordinates will also need to be
+        # transposed.
```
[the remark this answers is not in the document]
lbdreyer: > Are we now allowing a mix of 1D of and 2D coordinates?

Sort of. We are allowing `get_plot_defn` to pick up a mix of 1D of and 2D coordinates. `_map_common` also supports a mix, but this part of the code does not. I can add a check that the coords are the same dimensionality here if you like.

lbdreyer:
@pp-mo  I believe I have addressed all your review comments in the most recent commit.

pp-mo:
> @lbdreyer pp-mo I believe I have addressed all your review comments in the most recent commit.

Yes, I'm happy that we are all done here now.
Result :+1: 

I'll merge, and hopefully all will check out, since #3150 is now also merged.

PR #3141, opened 2018-08-14
Reinstate Python2 doctests with patched print function.

pp-mo:
A small improvement, if you like it @lbdreyer ?

### DIFF
--- a/docs/iris/src/userguide/interpolation_and_regridding.rst
+++ b/docs/iris/src/userguide/interpolation_and_regridding.rst
@@ -3,6 +3,7 @@
 
 .. testsetup:: *
 
+  from __future__ import print_function
   import numpy as np
   import iris
   # Mask out all warnings
--- a/docs/iris/src/userguide/interpolation_and_regridding.rst
+++ b/docs/iris/src/userguide/interpolation_and_regridding.rst
@@ -5,8 +5,23 @@
 
   import numpy as np
   import iris
+  # Mask out all warnings
   import warnings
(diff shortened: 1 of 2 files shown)


docs/iris/src/userguide/interpolation_and_regridding.rst
```diff
@@ -5,8 +5,23 @@
...
   import numpy as np
   import iris
+  # Mask out all warnings
   import warnings
   warnings.simplefilter('ignore')
+  # Patch print function for same numpy formatting in Python 2 and 3
+  from iris.util import _printopts_context
+  _old_print_func = print
+  def _new_print_func(x):
+      with _printopts_context(legacy='1.13'):
```
lbdreyer: I'm not so keen on the pinning to a version but at least this is consistent with the cml and ensures it's more stable :+1: 

I'll merge assuming the tests pass.

pp-mo: If you pass anything else in, it warns you that only '1.13' and None are currently accepted :frowning_face:

lbdreyer: I see, so it's not like there are multiple versions and we're pinning to one, rather we're enabling the legacy version of printing things which numpy has set to 1.13

pp-mo:
Tricky, it seems the print function patching does not work in Python2, but it does if I try it in an interpreter.  Maybe something subtly about the way Sphinx works ??
I'm giving this up for now.

Issue #3105, opened 2018-07-30
Improve layout and structure of contiguity checking


corinnebosley:
There are now several different parts to contiguity checking, starting from '_draw_2d_from_bounds' in plot.py, but they are a bit higgledy-piggledy and could do with tidying and simplifying (if possible).

PR #3144, opened 2018-08-15
Public contiguity checking

corinnebosley:
This PR contains:

plot.py:
- minor changes to function names, just for clarity
- if a user tries to plot an unmasked discontiguous array, they now get a warning with advice on how to fix this rather than an error

test_2d_coords.py:
- changed a class name to something more appropriate
- moved some tests from Test_2d_coords_bound_mode to new class TestContiguityChecking
- Added minimal set of tests to TestContiguityChecking

util.py:
- new function find_discontiguity()
- new_function mask_discontiguity()

Enjoy!


corinnebosley:
I can already see that I could do with making these changes:
- change test_2d_coords.py class name 'Test_2d_coords_bound_mode' to camel case
- add test for multiple discontiguities (as stated in TODO which is left in there)

lib/iris/plot.py
```diff
@@ -271,7 +273,7 @@ def _invert_yaxis(v_coord, axes=None):
             axes.invert_yaxis()
 
 
-def _check_contiguity_and_bounds(coord, data, abs_tol=1e-4, transpose=False):
+def _check_contiguity_and_mask(coord, data, abs_tol=1e-4, transpose=False):
```
lbdreyer: I agree with the inclusion of `mask` but by not mentioning `bounds` it isn't clear that that's what it is referring to.
How about `_check_bounds_contiguity_and_mask`?

lib/iris/plot.py
```diff
@@ -311,14 +313,18 @@ def _check_contiguity_and_bounds(coord, data, abs_tol=1e-4, transpose=False):
...
         if not_masked_at_discontinuity_along_x or \
                 not_masked_at_discontinuity_along_y:
-            raise ValueError('The bounds of the {} coordinate are not'
-                             ' contiguous and data is not masked where the'
-                             ' discontiguity occurs. Not able to create a'
-                             ' suitable mesh to give to'
-                             ' Matplotlib'.format(coord.name()))
+            discontiguity_warning = 'The bounds of the {} coordinate are ' \
+                                    'not contiguous and data is not masked ' \
+                                    'where the discontiguity occurs. The ' \
```
lbdreyer: Not a change you've made but now rereading through this intial sentence I see it doesn't make sense. We discontiguous twice: "The bounds of the {} coordinate are not contiguous ... where the discontiguity occurs.".

lbdreyer: Oh I see, this says "there is a discontiguity and the data is not masked where that discontiguity is"

corinnebosley:
@lbdreyer The changes I made for the last commit prompted numpy to give me this warning:
"FutureWarning: Using a non-tuple sequence for multidimensional indexing is deprecated; use `arr[tuple(seq)]` instead of `arr[seq]`. In the future this will be interpreted as an array index, `arr[np.array(seq)]`, which will result either in an error or a different result."

I'm not exactly sure what it's referring to, I have a horrible feeling it's something to do with the zip output, and that's been a really horrible nightmare to put into a user-friendly return format.

We will obviously have to address this at some point, but I could do with your help with it I think.

lib/iris/util.py
```diff
@@ -20,7 +20,8 @@
 """
 
 from __future__ import (absolute_import, division, print_function)
-from six.moves import (filter, input, map, range, zip)  # noqa
```
lbdreyer: We are using these two lines as boilerplate for all our modules to help with Python 3 compatibility. (you'll find them at the top of all files.)
I think it would be best to keep all of them (even if they aren't being used) as otherwise we may forget to include them.
In fact that is what the `#noqa` tag is there for - to stop flake8 complaining about unused imports

lib/iris/plot.py
corinnebosley: Yeah that'll do.  It's a bit of a mouthful but it's private anyway so it shouldn't matter.

lib/iris/util.py
corinnebosley: I didn't even realise it wasn't there.  I'll pop it in.

lib/iris/util.py
```diff
@@ -1554,3 +1555,100 @@ def _meshgrid(*xi, **kwargs):
...
+        coord:
+        Either
+
+        (a) An instance of :class:`iris.coords.Coord`
+
+        or
+
+        (b) the :attr:`standard_name`, :attr:`long_name`, or
+        :attr:`var_name` of an instance of an instance of
+        :class:`iris.coords.Coord`.
```
lbdreyer: If you are going to allow for a user to do this, you need to also add a line at the start that gets the coordinate.
This function starts by doing 
```
_, diffs_x, diffs_y = iris.coords._discontiguity_in_2d_bounds(coord.bounds,
                                                                  abs_tol)
```
so if that needs to be a coord instance by then

lib/iris/util.py
corinnebosley: No idea how I removed that...

lib/iris/util.py
corinnebosley: Thing is, if we do allow that user to use a name instead of a coord object, then we have to pass the cube in as well so that we can get the coord.  This seems a bit unnecessary, so it might make sense to only allow a coord object.

lib/iris/util.py
```diff
@@ -1554,3 +1554,91 @@ def _meshgrid(*xi, **kwargs):
...
+    Args:
+        coord:
+        An instance of :class:`iris.coords.Coord`
+
+    Kwargs:
+        abs_tol:
+        absolute tolerance in degrees
+
+    Returns:
+        Dictionary of values representing data array indices where
```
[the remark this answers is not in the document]
corinnebosley: Not really.  Not simpler.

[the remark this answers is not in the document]
corinnebosley: Also I don't see how that's any more use to the user.  If they have a particularly large dataset then they will have to search through it for places where it's True anyway, so what's the point?

And also, would we return the boolean array for the data or for the coord points?  What are you expecting the user to do with a boolean array?

lib/iris/util.py
```diff
@@ -1554,3 +1554,91 @@ def _meshgrid(*xi, **kwargs):
...
+        discontiguities['y-direction discontiguity at coord point(s)'] = \
+            discontiguity_array_y
+
+    try:
+        return discontiguities
+    except ValueError:
+        return None
+
+
+def mask_discontiguities(coord, data, abs_tol=1e-4):
```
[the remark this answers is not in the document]
corinnebosley: A boolean array would be fine for this because we don't need anything to be human-readable; we are just doing operations.  But I still disagree with creating a boolean array for the previous.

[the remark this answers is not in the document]
corinnebosley: Also what do you mean by 'this routine wouold not be needed?' 

This routine is pretty much as simple as it gets?  The whole point of having the two separate routines is that one allows you to check which points are discontiguous (as a human) and look at them and think about them, and the other allows you to just mask them without thought.  Those are both valid options and it seems stupid to just not offer one of them because that's simpler.

corinnebosley:
Changes following last review:

- find_discontiguities now returns a boolean array, not a dictionary (although it was not simple, thank you very much)
- make_bounds_discontiguous_at_point now broadcasts the discontiguity mask properly to handle cubes with more than 2 dimensions
- mask_discontiguities also broadcasts the 2d mask to data arrays with more than 2 dimensions
- temporary tests have been updated to check that the broadcasting works

corinnebosley:
@lbdreyer All tests passed.  What do you think?

lib/iris/tests/unit/plot/test_2d_coords.py
```diff
@@ -135,32 +114,120 @@ def test_map_common_2d(self):
...
-#              'However, data is masked where the discontiguity occurs so ' \
-#              'plotting anyway.'
-#        with self.assertWarnsRegexp(msg):
-#            iplt._check_contiguity_and_bounds(coord, cube.data)
+    def test_draw_2d_from_bounds(self):
+        # Test this function will not raise an error even with our most
+        # awkward but supported cube.
+        cube = self.latlon_2d
+        result = iplt._draw_2d_from_bounds('pcolormesh', cube)
+        self.assertTrue(result)
```
lbdreyer: `draw_2d_from_bounds` is too complicated to unit test. I suggest you just remove this test.
I am in the process of adding an integration test that will test `draw_2d_from_bounds` (which will at the same time check that it is not raising an error)

lib/iris/util.py
```diff
@@ -1554,3 +1554,73 @@ def _meshgrid(*xi, **kwargs):
...
+
+    Kwargs:
+        abs_tol:
+        absolute tolerance in degrees
+
+    Returns:
+        Copy of the data array, masked at points where bounds are
+        discontiguous.
+    """
+    mask = find_discontiguities(coord, abs_tol=abs_tol)
```
lbdreyer: `mask` isn't that clear what it means as it doesn't refer to the cube's original mask, it refers to the locations of the discontiguities and so the new places that should be masked.
How about something like `discontiguous_locations` or `discontiguous` or `discontiguous_bounds_locations`... I can't really think of a good name.

lbdreyer: Maybe just `new_mask` or `extra_mask`

lib/iris/tests/unit/plot/test_2d_coords.py
```diff
@@ -135,32 +114,120 @@ def test_map_common_2d(self):
...
+
+        # Check that a warning occurs when the contiguity check finds
         # discontiguous bounds but unmasked data.
         cube = self.latlon_2d
         cube.data.mask = ma.nomask
+        coord = self.lons
+
+        with warnings.catch_warnings():
+            warnings.simplefilter("error")
+            with self.assertRaises(UserWarning):
```
lbdreyer: This really needs a `assertRaisesRegexp` to check it's getting the right warning`

lib/iris/tests/stock/_stock_2d_latlons.py
```diff
@@ -342,5 +343,7 @@ def adjust_coord(coord):
     if not ma.isMaskedArray(data):
         # Promote to masked array, to avoid converting mask to NaN.
         data = ma.masked_array(data)
-    data[at_iy, at_ix] = ma.masked
+
+    # Mask point after discontiguity to mitigate plotting issues
+    data[..., at_iy, at_ix + 1] = ma.masked
```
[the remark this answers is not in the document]
lbdreyer: And due to this, the test `TestContiguityChecking` isn't right. In the `setUp`, you do:
```
        make_bounds_discontiguous_at_point(self.latlon_2d, 3, 2)
       ...
        self.masked[3, 3] = True
```

lib/iris/tests/unit/plot/test_2d_coords.py
```diff
@@ -135,32 +114,120 @@ def test_map_common_2d(self):
...
+        self.latlon_2d = full2d_global()
+        make_bounds_discontiguous_at_point(self.latlon_2d, 3, 2)
+        self.lons = self.latlon_2d.coord('longitude')
+
+        # Construct an array the size of the points array filled with 'False'
+        # to represent an empty discontiguity array
+        self.empty_bool = np.zeros(self.lons.points.shape,
+                                   dtype=bool)
+        self.masked = self.empty_bool.copy()
+        self.masked[3, 3] = True
```
lbdreyer: Why not just use the mask of `self.latlon_2d`?

lib/iris/tests/unit/plot/test_2d_coords.py
```diff
@@ -135,32 +114,120 @@ def test_map_common_2d(self):
...
+        cube = self.latlon_2d
+        result = iplt._draw_2d_from_bounds('pcolormesh', cube)
+        self.assertTrue(result)
+
+
+class TestContiguityChecking(tests.IrisTest):
+    def setUp(self):
+        self.latlon_2d = full2d_global()
+        make_bounds_discontiguous_at_point(self.latlon_2d, 3, 2)
+        self.lons = self.latlon_2d.coord('longitude')
```
lbdreyer: I'd prefer more descriptive names, it would make reading the test easier, so:
`self.latlon_2d_cube` and `self.lons_coord`

lib/iris/plot.py
```diff
@@ -311,14 +314,19 @@ def _check_contiguity_and_bounds(coord, data, abs_tol=1e-4, transpose=False):
...
+                                    'mesh created for this plot may cause a ' \
+                                    'misrepresentation of the input data.  ' \
+                                    'Please use ' \
+                                    'iris.contiguity_check(coord) to find ' \
+                                    'and mask your data ' \
+                                    'appropriately'.format(coord.name())
```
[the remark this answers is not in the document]
corinnebosley: Yep, this is leftover from before the full implementation.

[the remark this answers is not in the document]
corinnebosley: Is this going to end up a bit long and rambly?

lib/iris/plot.py
```diff
@@ -311,14 +314,19 @@ def _check_contiguity_and_bounds(coord, data, abs_tol=1e-4, transpose=False):
...
+            warnings.warn(discontiguity_warning)
```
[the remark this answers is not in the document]
corinnebosley: Ah.  Good decision.

lib/iris/util.py
```diff
@@ -1554,3 +1554,73 @@ def _meshgrid(*xi, **kwargs):
...
+                                              gaps_x)
+
+    # apply mask for y-direction discontiguities:
+    bad_points_boolean[1:, :] = np.logical_or(bad_points_boolean[1:, :],
+                                              gaps_y)
+
+    try:
+        return bad_points_boolean
+    except ValueError:
+        return None
```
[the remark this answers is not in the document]
corinnebosley: Yes I did.  I needed this for an earlier implementation but it will never return None now (unless it's broken), so you're quite right.

lib/iris/util.py
```diff
@@ -1554,3 +1554,73 @@ def _meshgrid(*xi, **kwargs):
...
+    Searches coord for discontiguities in the bounds array, returned as
+    booleans.
+
+    Args:
+        coord:
+        An instance of :class:`iris.coords.Coord`
+
+    Kwargs:
+        abs_tol:
+        absolute tolerance in degrees
```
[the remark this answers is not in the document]
corinnebosley: OK, but I do think it needs some kind of clarification, so I'll put 'in the coordinate units' or something.

lib/iris/util.py
```diff
@@ -1554,3 +1554,73 @@ def _meshgrid(*xi, **kwargs):
         if mxii.dtype != xii.dtype:
             mxi[i] = mxii.astype(xii.dtype)
     return mxi
+
+
+def find_discontiguities(coord, abs_tol=1e-4):
```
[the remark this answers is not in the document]
corinnebosley: That's very long.  If I was a user I'd find that a massive pain in the arse to type out.

lib/iris/util.py
```diff
@@ -1554,3 +1554,73 @@ def _meshgrid(*xi, **kwargs):
...
+    bad_points_boolean[1:, :] = np.logical_or(bad_points_boolean[1:, :],
+                                              gaps_y)
+
+    try:
+        return bad_points_boolean
+    except ValueError:
+        return None
+
+
+def mask_discontiguities(coord, data, abs_tol=1e-4):
```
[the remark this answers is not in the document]
corinnebosley: I don't think it should be any longer than mask_data_at_discontiguities, and even that's longer than I'd like it to be.

lib/iris/util.py
```diff
@@ -1554,3 +1554,73 @@ def _meshgrid(*xi, **kwargs):
...
+        the :attr:`data` of the cube from which the coord is taken.
+
+    Kwargs:
+        abs_tol:
+        absolute tolerance in degrees
+
+    Returns:
+        Copy of the data array, masked at points where bounds are
+        discontiguous.
+    """
```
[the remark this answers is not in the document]
corinnebosley: Yeah, I was a little torn about this.  That seems like a good option though.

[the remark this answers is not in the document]
corinnebosley: To be robust and flexible though, I'll have to accept either the name of the coord or the coord object, and that means I'll have to extend this a bit.

lib/iris/util.py
```diff
@@ -1554,3 +1554,73 @@ def _meshgrid(*xi, **kwargs):
...
+    mask = find_discontiguities(coord, abs_tol=abs_tol)
+
+    masked_data = ma.masked_array(data.copy())
+    # mask is 2d point, so expand this with an ellipsis to broadcast to
+    # full data shape
+    masked_data[..., mask] = ma.masked
```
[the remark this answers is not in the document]
corinnebosley: Damn, you're right.  I didn't really want this to get too complex but you are right.

lib/iris/tests/unit/plot/test_2d_coords.py
```diff
@@ -21,15 +21,23 @@
...
 import iris.tests as tests
 
+import numpy as np
 import numpy.ma as ma
+import warnings
 
 import iris.coords as coords
+import iris.util
+from iris.util import find_discontiguities
+from iris.util import mask_discontiguities
```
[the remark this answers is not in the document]
corinnebosley: Yeah, I think this happens when I refactor things in PyCharm.  It keeps happening with other imports as well.  Annoying.

lib/iris/tests/stock/_stock_2d_latlons.py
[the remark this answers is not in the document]
corinnebosley: We talked about this, and agreed that it was the cell after the disocintiguity that would be the problem.  That's why this whole thing got so complicated.  I could have done this in a day if that wasn't the case.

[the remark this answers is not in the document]
corinnebosley: This menas I have to change practically everything.  FFS

lib/iris/tests/unit/plot/test_2d_coords.py
corinnebosley: Yeah, this was meant as a placeholder really, I think it was one of the first tests I put in, just to help me think.  I'll take it out.

lib/iris/tests/unit/plot/test_2d_coords.py
corinnebosley: And unmask it first?  I suppose so.  Don't see that it makes a lot of difference but I'll pop it in anyway.

corinnebosley: Oh no wait a minute, I'm not quite sure what I'm actually trying to achieve here, maybe I need to think about this a bit more and add some comments.

corinnebosley: I think I will need the array of Falses that I construct here because that's the easiest way to do it, but there is no need to then mask the single point I have done here.  I can just use the cube's mask instead of constructing self.masked.

corinnebosley:
Had some trouble with `make_bounds_discontiguous_at_point` here.  If you use full2d_global (which is just `sample_2d_latlons(transformed=True)`, some of the points become miniscule and therefore the differences between bottom-left bound and bottom-right bound are also miniscule.

I have been finding that many coordinate points from `full2d_global` (such as (0, 0) and (3, 2)) are so small that the discontiguity created by `make_bounds_discontiguous_at_point` is well below the default absolute tolerance in `find_discontiguities_in_bounds`.

This is possibly a good test for a relative tolerance option though...

corinnebosley:
@lbdreyer I'm slightly worried about bits still being a problem, particularly test data.  But for now, I'm afraid this is the best you're gonna get.  Thanks for all your help and good luck getting to the finish with this.

pp-mo:
Hi @corinnebosley @lbdreyer 
:+1: for good stuff + progress made

However, discussing today with @lbdreyer I'm still not very sure about the API.
We could easily combine these two functions now.
Or, we could make the 'apply fixes' one into a more general "apply_mask" operation.  But in that case it should work with a general mask input, for which purpose a boolean array is not good because the dimensions to apply it to are unknown.  
So actually, ideally, I think that result should be wrapped as a cube containing coords, so that the 2d mask can be mapped to the dimensions it applies to.
It also seems to me that this should be automatically checking the x- and y-coords of a given *cube*, or every use will be calling it twice ??

Rather more seriously ... I'm wondering now where + why we need this ??
If we haven't an immediate use within the collected test data, then I'm not interested + this gets low priority.
I'll go looking ...

pp-mo:
> @pp-mo Rather more seriously ... I'm wondering now where + why we need this 

Forget that -- I found out why today !

dkillick:
... and consider #3105 while getting this fixed.

lib/iris/plot.py
```diff
@@ -23,7 +23,9 @@
 """
 
 from __future__ import (absolute_import, division, print_function)
-from six.moves import (filter, input, map, range, zip)  # noqa
+
+from six.moves import (map, zip)  # noqa
+import six
```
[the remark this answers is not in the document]
corinnebosley: I just got told off by Travis for not having 'future imports' (i.e. six) in here, and now I've put it back I'm getting told off by Sticklerbot.  I can't win!!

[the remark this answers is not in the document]
corinnebosley: ohhh, maybe I need a `# noqa`... 

I always wondered what that meant.

lib/iris/util.py
```diff
@@ -1554,3 +1554,110 @@ def _meshgrid(*xi, **kwargs):
...
+        An instance of :class:`iris.coords.Coord`
+
+    Kwargs:
+        abs_tol:
+        absolute tolerance in coordinate units
+
+    Returns:
+        Boolean array representing the truth value for
+        discontiguous cells in the coordinate points array.
+    """
```
pp-mo: Extra blank line here

lib/iris/util.py
```diff
@@ -1554,3 +1554,110 @@ def _meshgrid(*xi, **kwargs):
...
+        the :attr:`data` of the cube from which the coord is taken.
+
+    Kwargs:
+        abs_tol:
+        absolute tolerance in degrees
+
+    Returns:
+        A cube whose data array is masked at points where the chosen
+        coordinate bounds array is discontiguous.
+    """
```
pp-mo: Blank line again

corinnebosley:
Stuff still to do:
- Add a sentence to the `iris.plot._check_bounds_contiguity_and_mask` error message recommending usage of `iris.util.find_discontiguities`
- Check appropriateness of examples in docstring for `iris.util.find_discontiguities`
- Add test for `iris.util.mask_discontiguities` y-direction discontiguity
- Get rtol test passing in `iris.tests.unit.util.test_find_discontiguities`

lib/iris/tests/stock/_stock_2d_latlons.py
```diff
@@ -341,10 +341,28 @@ def adjust_coord(coord):
...
-    data = cube.data  # N.B. fetch all the data.
-    if not ma.isMaskedArray(data):
-        # Promote to masked array, to avoid converting mask to NaN.
-        data = ma.masked_array(data)
-    data[at_iy, at_ix] = ma.masked
-    cube.data = data
+
+    # Check which dimensions are spanned by each coordinate.
+    for coord in (x_coord, y_coord):
+        if isinstance(coord, int):
```
pp-mo: The 'coord is an int' option is _definitely_ not applicable here, as x_coord and y_coord are definitely coords in this usage.
I think the use of 'set' is also redundant.
I think it could just have ...
```
coord_dims = cube.coord_dims(coord)
if not coord_dims:
    ...
sliced_cube = cube.slices(coord_dims)
```
(needs checking though ?)

corinnebosley: I think you're probably right.  I remember having that conversation.

lib/iris/util.py
```diff
@@ -1554,3 +1554,104 @@ def _meshgrid(*xi, **kwargs):
...
     return mxi
+
+
+def find_discontiguities(cube, rel_tol=1e-5, abs_tol=1e-8):
+    """
+    Searches coord for discontiguities in the bounds array, returned as a
+    boolean array (True where discontiguities are present).
+
+    Args:
+        cube:
```
pp-mo: Sorry to be a pain, but can you reshape these docstrings a bit, to get them in the approved format, more like other things in this module for example ?
So e.g. after a bit of massage, this one might read :
```
    Args:
        cube (`iris.cube.Cube`):
            The cube to be checked for discontinuities in its 'x' and 'y' coordinates.

    Kwargs:
        rel_tol (float):
            The relative equality tolerance to apply in coordinate bounds checking.
        abs_tol (float):
            The absolute value tolerance to apply in coordinate bounds checking.

    Returns:
        result (`numpy.ndarray` of bool) :
            true/false map of which cells in the cube XY grid have discontiguities in the coordinate points array.
            This can be used as the input array for :func:`iris.util.mask_discontiguities`.

```

lib/iris/util.py
```diff
@@ -1554,3 +1554,104 @@ def _meshgrid(*xi, **kwargs):
...
+    dim_err_msg = 'Discontiguity searches are currently only supported for ' \
+                  '2-dimensional coordinates.'
+    if len(spatial_coords) != 2:
+        raise NotImplementedError(dim_err_msg)
+
+    # Check which dimensions are spanned by each coordinate.
+    for coord in spatial_coords:
+        if coord.ndim != 2:
+            raise NotImplementedError(dim_err_msg)
+        elif isinstance(coord, int):
```
pp-mo: The `coord` is definitely an actual Coord, as we got it above from the cube.aux_coords.
So we don't need to support any case where coord is an int, and this bit can be removed.

lib/iris/util.py
```diff
@@ -1554,3 +1554,104 @@ def _meshgrid(*xi, **kwargs):
...
+
+        bad_points_boolean[:, :-1] = np.logical_or(bad_points_boolean[:, :-1],
+                                                   gaps_x)
+        # apply mask for y-direction discontiguities:
+        bad_points_boolean[:-1, :] = np.logical_or(bad_points_boolean[:-1, :],
+                                                   gaps_y)
+    return bad_points_boolean
+
+
+def mask_discontiguities(cube_slice, discontiguous_points):
```
pp-mo: Not clear why this is called 'cube_slice'.  
If you want to make the point that we only support 2d cubes, I don't think this helps much.
Can we call it just "cube" ?

lib/iris/util.py
```diff
@@ -1554,3 +1554,104 @@ def _meshgrid(*xi, **kwargs):
...
+    """
+    Masks any cells in the data array which correspond to discontiguous
+    cells in the coordinate points array.
+
+    Args:
+        cube_slice:
+        A scalar instance of :class:`iris.cube.Cube` (i.e. 2-dimensional)
```
pp-mo: I don't think 'scalar instance' is right, that sounds like a scalar cube, i.e. "shape=(1,)".
I think it is just 2-dimensional.

We agreed that this is a reasonable limitation, at least for now.
It does means you have separately mask any 'slice' of a larger cube you want to plot.
Maybe we should make it more noticeable in the docs, somehow.  Not sure how ?

lib/iris/util.py
pp-mo: Likewise, I wouldn't call the array argument 'discontiguous_points', because it undermines the point of making this a separate function, which is to emphasize that the user is making a definite decision to mask off some points, whether those chosen by the 'find_' routine, or something else.

pp-mo: Actually the routine shouldn't be called 'mask_discontiguities' either !!!

pp-mo: `def mask_cube(cube, points_to_mask)`

lib/iris/util.py
corinnebosley: Examples in WhatsNew?

lib/iris/util.py
corinnebosley: Oh wow, since when did we write our docstrings like this?  I like them organised this way.

lib/iris/tests/unit/plot/test__check_bounds_contiguity_and_mask.py
```diff
@@ -88,20 +90,20 @@ def test_2d_contiguous_atol(self):
...
+            cube.coord('longitude'), cube.data))
 
     def test_2d_discontigous_unmasked(self):
         # Test a 2D coordinate which is discontiguous and unmasked at
         # discontiguities.
         cube = sample_2d_latlons()
         make_bounds_discontiguous_at_point(cube, 3, 4)
         msg = 'coordinate are not contiguous'
-        cube.data[3, 4] = ma.nomask
+        cube.data.mask[3, 4] = ma.nomask
```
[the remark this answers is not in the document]
corinnebosley: Yeah I was a little unsure about the specifics of all this.  It's definitely a masked array at line 106ish because `make_bounds_discontiguous_at_point` masks it, but there are some secret and confusing corners of masked array operations.

Anyway, I've reverted it to `cube.data[3, 4] = ma.nomask` now.  Thanks for the advice.

lib/iris/tests/unit/plot/test__check_bounds_contiguity_and_mask.py
```diff
@@ -69,10 +70,11 @@ def test_1d_discontigous_unmasked(self):
             iplt._check_bounds_contiguity_and_mask(coord, data, atol=1e-3)
 
     def test_2d_contiguous(self):
-        # Test a 2D coordinate which is contiguous.
+        # Test that a 2D coordinate which is contiguous does not throw
+        # an error.
         cube = sample_2d_latlons()
-        iplt._check_bounds_contiguity_and_mask(cube.coord('longitude'),
-                                               cube.data)
+        self.assertIsNone(iplt._check_bounds_contiguity_and_mask(
```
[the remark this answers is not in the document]
corinnebosley: Sorted.

lib/iris/tests/stock/_stock_2d_latlons.py
```diff
@@ -308,12 +308,12 @@ def make_bounds_discontiguous_at_point(cube, at_iy, at_ix):
 
     Changes the points and bounds of a single gridcell, so that it becomes
     discontinuous with the next gridcell to its right.
+
     Also masks the cube data at the given point.
```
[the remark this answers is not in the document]
corinnebosley: @pp-mo does it?  I thought this would work fine on any lat-lon cube?

lib/iris/tests/stock/_stock_2d_latlons.py
```diff
@@ -26,7 +26,7 @@
 import numpy.ma as ma
 
 from iris.analysis.cartography import unrotate_pole
-from iris.cube import Cube
+from iris.cube import Cube, CubeList
```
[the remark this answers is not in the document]
corinnebosley: We simplified `iris.util.mask_cube` (which had pretty much the same masking operations in) but not this.  I don't have a problem with keeping this as it is because it's only to generate test data, which means it won't be used on large data sets so it won't eat up lots of processing time slicing up cubes and masking them.

lib/iris/tests/stock/_stock_2d_latlons.py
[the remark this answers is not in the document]
corinnebosley: Not to worry, I was addressing changes bottom-to-top instead of top-to-bottom.  Confusion resolved.

corinnebosley:
@pp-mo Still have to remove all references to discontiguity from `iris.util.mask_cube`.  Sorry, forgot...

corinnebosley:
@pp-mo I have now removed all references to discontiguity in `iris.util.mask_cube`, but I'd like you to check the words in the docstring for me please because I'm not sure if they're quite right.

Thanks!

lib/iris/tests/unit/plot/test__check_bounds_contiguity_and_mask.py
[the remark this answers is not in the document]
pp-mo: !sorted! :+1:

lib/iris/tests/unit/util/test_mask_discontiguities.py
```diff
@@ -0,0 +1,63 @@
...
+from six.moves import (filter, input, map, range, zip)  # noqa
+
+# Import iris.tests first so that some things can be initialised before
+# importing anything else.
+import iris.tests as tests
+
+import numpy as np
+import numpy.ma as ma
+
+from iris.util import mask_discontiguities
```
pp-mo: Something is wrong here.
There is no longer any `iris.util.mask_discontiguities`.
So I don't get how the tests can be passing ??

lib/iris/util.py
pp-mo: Well it's set out in the developer guide https://scitools.org.uk/iris/docs/latest/developers_guide/documenting/docstrings.html#multi-line-docstrings

I believe "most" of Iris follows it, but there are certainly some exceptions in older code.

lib/iris/util.py
pp-mo: Don't think it's critical.  People will get the point.

pp-mo:
>  iris.util.mask_cube ... I'd like you to check the words in the docstring

Looks fine to me :smiley: !

lib/iris/util.py
```diff
@@ -1554,3 +1554,111 @@ def _meshgrid(*xi, **kwargs):
...
+    in the `points_to_mask` array.
+
+    Args:
+
+    * cube (`iris.cube.Cube`):
+        A 2-dimensional instance of :class:`iris.cube.Cube`.
+
+    * points_to_mask (`numpy.ndarray` of bool):
+        A 2d boolean array of Truth values representing points to mask in the
+        x and y arrays of the cube.
```
lbdreyer: Surely there's no reason the cube and mask *have* to be 2D. Could you not just say that the mask must match the cube's shape?

lib/iris/util.py
```diff
@@ -1554,3 +1554,111 @@ def _meshgrid(*xi, **kwargs):
...
+
+def find_discontiguities(cube, rel_tol=1e-5, abs_tol=1e-8):
+    """
+    Searches coord for discontiguities in the bounds array, returned as a
+    boolean array (True where discontiguities are present).
+
+    Args:
+
+    * cube (`iris.cube.Cube`):
+        The cube to be checked for discontinuities in its 'x' and 'y'
```
lbdreyer: Pernickety but...
Ideally I'd change `discontinuities` -> `discontiguities` as it is referred to in the first sentence of this doc string.

There are a few places in this PR where the word `discontinuous` is used, but others where `discontiguous` is used. It would be nice if we could remain consistent and my preference is for `discontiguous`

corinnebosley: @lbdreyer You're quite right.  Those are just ones I missed.  Thanks!

lib/iris/util.py
pp-mo: Actually it's disappointing how *hard* the broadcasting was to get right.
The obvious (to me!) thing, which would look a bit like 
`full_dim_slices = xy_slices[None, None, :, :]; cube[full_dim_slices] = ma.masked`, 
*does not work*.
So I think you need to do an "|=" to get it to broadcast, or something like that, but that means addressing the mask array directly, with all the usual problems that causes.  
So for the sake of speed, as there were still a whole bunch of tiresome issues + choices to be agreed, we gave it up instead :grin: .

PR #3148, opened 2018-08-30
Tests for rotate_grid_vectors

pp-mo:
Previously deferred work, as noted in #3114

### DIFF
--- a/lib/iris/tests/unit/analysis/cartography/test_rotate_grid_vectors.py
+++ b/lib/iris/tests/unit/analysis/cartography/test_rotate_grid_vectors.py
@@ -30,9 +30,9 @@ from mock import Mock, call as mock_call
 import numpy as np
 
 from iris.cube import Cube
+from iris.tests.stock import sample_2d_latlons
 
 from iris.analysis.cartography import rotate_grid_vectors
-from iris.tests.stock import sample_2d_latlons
 
 
 class TestRotateGridVectors(tests.IrisTest):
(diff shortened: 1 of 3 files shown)


lib/iris/tests/unit/analysis/cartography/test_rotate_grid_vectors.py
```diff
@@ -0,0 +1,148 @@
...
+        self.assertTrue(np.ma.isMaskedArray(out_v))
+        if nan_angles_mask is not None:
+            self.assertArrayEqual(out_u.mask, nan_angles_mask)
+            self.assertArrayEqual(out_v.mask, nan_angles_mask)
+
+    def test_angles_calculation(self):
+        self._check_angles_calculation()
+
+    def test_angles_in_radians(self):
+        self._check_angles_calculation(angles_in_degrees=True)
```
lbdreyer: This test is called `test_angles_in_radians` but it uses `angles_in_degrees=True`

lib/iris/tests/unit/analysis/cartography/test_rotate_grid_vectors.py
```diff
@@ -0,0 +1,148 @@
...
+# importing anything else.
+import iris.tests as tests
+
+from mock import Mock, call as mock_call
+import numpy as np
+
+from iris.cube import Cube
+
+from iris.analysis.cartography import rotate_grid_vectors
+from iris.tests.stock import sample_2d_latlons
```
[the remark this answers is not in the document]
pp-mo: The last bit is conventionally separate as it is the "Device Under Test"
But you're right, that the `sample_2d_latlons` needs to be integrated with the rest of the 'supporting' imports.  Will fix ..

lib/iris/tests/unit/analysis/cartography/test_rotate_grid_vectors.py
pp-mo: Whoops!  
Should obviously be 'False'. Will fix.

pp-mo:
Thanks @lbdreyer 
I hope we're done here now ?

I'm not self-merging, as there's no immediate need (it's only extra tests really)

PR #3151, opened 2018-08-31
2d coords proj4 mod

pp-mo:
Problems have arisen with Mollweide plots since Proj4 version 5.1 has become available
(though cartopy version is still 0.16, but this is thought to have known problems with Proj4 > 5). 

Changes initally disabled to confirm problem, then fixed.
The fix depends on https://github.com/SciTools/test-iris-imagehash/pull/17

NOTE: targetting 2d_coords feature branch, for now.


pp-mo:
Closed in favour of #3150 
 - since @pelson explained how to make that work

PR #3155, opened 2018-09-03
iris.util.reverse on cubes

rcomer:
The `iris.util.reverse` [documentation](https://scitools.org.uk/iris/docs/v2.1/iris/iris/util.html?highlight=util%20reverse#iris.util.reverse) states that it can be used for reversing an array, but it works equally well for reversing a cube.  Since I use it this way in my code, I'd like to make this usage official by:

- Updating the docstring to indicate the function works on a cube.
- Adding a test for reversing a cube.

I also moved the existing array tests into `tests/unit` and swapped out `np.testing.assert_array_equal` for `self.assertArrayEqual`.

Note that for numpy arrays there is also [`numpy.flip`](https://docs.scipy.org/doc/numpy/reference/generated/numpy.flip.html)

### DIFF
--- a/lib/iris/tests/unit/util/test_reverse.py
+++ b/lib/iris/tests/unit/util/test_reverse.py
@@ -170,9 +170,15 @@ class Test_cube(tests.IrisTest):
         with self.assertRaisesRegexp(ValueError, msg):
             reverse(self.cube1, [])
 
-        msg = 'coords_or_dims must be int, str, coordinate or *'
+        msg = ('coords_or_dims must be int, str, coordinate or sequence of '
+               'these.  Got cube.')
         with self.assertRaisesRegexp(TypeError, msg):
             reverse(self.cube1, self.cube1)
+
+        msg = ('coords_or_dims must be int, str, coordinate or sequence of '
+               'these.')
+        with self.assertRaisesRegexp(TypeError, msg):
+            reverse(self.cube1, 3.)
 
 
 if __name__ == '__main__':
--- a/lib/iris/tests/unit/util/test_reverse.py
+++ b/lib/iris/tests/unit/util/test_reverse.py
@@ -1,0 +1,100 @@
+# (C) British Crown Copyright 2010 - 2018, Met Office
+#
+# This file is part of Iris.
+#
+# Iris is free software: you can redistribute it and/or modify it under
+# the terms of the GNU Lesser General Public License as published by the
+# Free Software Foundation, either version 3 of the License, or
+# (at your option) any later version.
+#
+# Iris is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU Lesser General Public License for more details.
+#
+# You should have received a copy of the GNU Lesser General Public License
+# along with Iris.  If not, see <http://www.gnu.org/licenses/>.
+"""Test function :func:`iris.util.reverse`."""
+
+from __future__ import (absolute_import, division, print_function)
+from six.moves import (filter, input, map, range, zip)  # noqa
+
+# Import iris.tests first so that some things can be initialised before
+# importing anything else.
+import iris.tests as tests
+
+import unittest
+
+import iris
+import numpy as np
+
+
+class Test(tests.IrisTest):
+    def SetUpReversed(self):
+        # On this cube pair, the coordinates to perform operations on have
+        # matching long names but the points array on one cube is reversed
+        # with respect to that on the other.
+        data = np.zeros((3, 4))
+        a1 = iris.coords.DimCoord([1, 2, 3], long_name='a')
+        b1 = iris.coords.DimCoord([1, 2, 3, 4], long_name='b')
+        a2 = iris.coords.DimCoord([3, 2, 1], long_name='a')
+        b2 = iris.coords.DimCoord([4, 3, 2, 1], long_name='b')
+
+        reversed1 = iris.cube.Cube(
+            data, dim_coords_and_dims=[(a1, 0), (b1, 1)])
+        reversed2 = iris.cube.Cube(
+            data, dim_coords_and_dims=[(a2, 0), (b2, 1)])
+
+        return reversed1, reversed2
+
+    def test_simple_array(self):
+        a = np.arange(12).reshape(3, 4)
+        self.assertArrayEqual(a[::-1], iris.util.reverse(a, 0))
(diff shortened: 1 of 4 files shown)


lib/iris/tests/unit/util/test_reverse.py
```diff
@@ -0,0 +1,101 @@
...
+
+        self.assertArrayEqual(cube1.data[::-1, ::-1], cube1_reverse_both.data)
+        self.assertArrayEqual(cube2.coord('a').points,
+                              cube1_reverse_both.coord('a').points)
+        self.assertArrayEqual(cube2.coord('b').points,
+                              cube1_reverse_both.coord('b').points)
+
+
+if __name__ == '__main__':
+    unittest.main()
```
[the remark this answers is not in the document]
rcomer: I fixed this with `--amend`, so think it ought to be marked "outdated".

lib/iris/tests/unit/util/test_reverse.py
```diff
@@ -0,0 +1,100 @@
...
+from __future__ import (absolute_import, division, print_function)
+from six.moves import (filter, input, map, range, zip)  # noqa
+
+# Import iris.tests first so that some things can be initialised before
+# importing anything else.
+import iris.tests as tests
+
+import unittest
+
+import iris
```
dkillick: By convention in unit test files like this, the unit being tested is imported explicitly, which in this case would mean `from iris.util import reverse`. You can then just call `reverse`, rather than `iris.util.reverse`.

See, for example, [this usage in the `array_util` tests](https://github.com/SciTools/iris/blob/master/lib/iris/tests/unit/util/test_array_equal.py#L29).

lib/iris/tests/unit/util/test_reverse.py
```diff
@@ -0,0 +1,100 @@
...
+import numpy as np
+
+
+class Test(tests.IrisTest):
+    def SetUpReversed(self):
+        # On this cube pair, the coordinates to perform operations on have
+        # matching long names but the points array on one cube is reversed
+        # with respect to that on the other.
+        data = np.zeros((3, 4))
```
dkillick: Hm, this isn't a good test of whether the data is being reversed as all the points are the same! Better to use `np.arange(12).reshape(3, 4)` 😉

lib/iris/util.py
```diff
@@ -406,16 +406,16 @@ def between(lh, rh, lh_inclusive=True, rh_inclusive=True):
...
 
     Args:
 
-    * array
-        The array to reverse
-    * axes
-        A single value or array of values of axes to reverse
+    * cube_or_array: :class:`iris.cube.Cube` or :class:`numpy.ndarray`
+        The cube or array to reverse
+    * axes: int or sequence of ints
```
dkillick: It's great that `reverse` can be used on cubes or NumPy arrays. This does, however, introduce a usage issue in that typically if I pass a cube to functionality that operates on a dimension I can specify the dimension of interest as a coordinate name as well as an axis. 

It would be good if you could pass one or more coordinates or axes here, and have the function look up the axis from the specified coordinate (I'm pretty sure there are examples of doing such elsewhere in the codebase). Of course, with this you'll need to add more checking, for example if you pass a NumPy array and named dimensions... You'll also need to add testing for this functionality.

lib/iris/tests/unit/util/test_reverse.py
rcomer: I believe all I can say here is "ooops!"

rcomer:
Thanks for your comments @dkillick, particularly the pointers about testing conventions, as I'm pretty new to that.  I'll make all those changes when I get a minute.

lib/iris/util.py
```diff
@@ -406,16 +407,19 @@ def between(lh, rh, lh_inclusive=True, rh_inclusive=True):
...
 
     Args:
 
-    * array
-        The array to reverse
-    * axes
-        A single value or array of values of axes to reverse
+    * cube_or_array: :class:`iris.cube.Cube` or :class:`numpy.ndarray`
+        The cube or array to reverse.
+    * coords_or_dims: int, str, :class:`iris.coords.Coord` or sequence of these.
```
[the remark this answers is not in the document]
rcomer: If I put "these." on the next line to make this compliant, it looks wrong in the docs.  There must be a simple solution here...?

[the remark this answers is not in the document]
QuLogic: Remove the `.`; you're only one character over, and no other types end in periods.

rcomer:
Hi @dkillick, I think I've covered everything.  I just have one very noddy PEP8 issue that I'm not sure how to rearrange (see Stickler's comment).

I split the changes between two commits, so it's easy to find the new tests for the API change separately from the general test reorganisation.

lib/iris/util.py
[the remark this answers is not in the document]
rcomer: Thanks @QuLogic !

lib/iris/util.py
```diff
@@ -453,7 +453,8 @@ def reverse(cube_or_array, coords_or_dims):
     """
     index = [slice(None, None)] * cube_or_array.ndim
 
-    if iris.cube._is_single_item(coords_or_dims):
+    if (iris.cube._is_single_item(coords_or_dims) or
+            isinstance(coords_or_dims, iris.cube.Cube)):
```
rcomer: Explicitly check for a cube here, on the assumption that the issues in my main comment are specific to cubes.

lib/iris/util.py
```diff
@@ -471,7 +472,7 @@ def reverse(cube_or_array, coords_or_dims):
                                 'or sequence of these.')
 
     axes = np.array(list(axes), ndmin=1)
-    if axes.ndim != 1:
+    if axes.ndim != 1 or axes.size == 0:
```
rcomer: This is added because the tests that pass an empty string were failing at `np.min` (line 478 below) and the error message was less informative in context.

dkillick:
> Well this is awkward.

Um, yes 😳 I think what's happening is `collections.Iterable` determines if an object is iterable by looking for an `__iter__` method on the base class. Iris cubes [have an `__iter__` method](https://github.com/SciTools/iris/blob/master/lib/iris/cube.py#L2111-L2112) but it just raises an error. These two things combined make cubes look a bit silly, so I think I might raise an issue regarding this...

> how should I fix this?

I think my preferred fix for this situation would be a separate check specifically for `coords_or_dims` being an Iris cube, which raises its own special error for that case (and, of course, is tested). I had a quick think about whether we might want to be able to define a reverse operation based on a cube, but I think that's an operation that makes no sense so shouldn't be allowed.

rcomer:
> I think my preferred fix for this situation would be a separate check specifically for coords_or_dims being an Iris cube, which raises its own special error for that case (and, of course, is tested).

Done, I think.

PR #3158, opened 2018-09-04
Vector plots 4

pp-mo:
Integration tests for `iris.plot.quiver` and `streamplot`

Needs https://github.com/SciTools/test-iris-imagehash/pull/18 to pass

### DIFF
--- a/lib/iris/tests/integration/plot/test_vector_plots.py
+++ b/lib/iris/tests/integration/plot/test_vector_plots.py
@@ -88,7 +88,7 @@ class MixinVectorPlotCases(object):
         x, y, u, v = self._nonlatlon_xyuv()
         u_cube, v_cube = self._nonlatlon_uv_cubes(x, y, u, v)
         """Basic non-latlon, 1d coords testcase."""
-        self.plot('nonlatlon_1d', u_cube, v_cube)
+        self.plot('nonlatlon, 1-d coords', u_cube, v_cube)
         plt.xlim(x.min() - 1, x.max() + 2)
         plt.ylim(y.min() - 1, y.max() + 2)
         self.check_graphic()
--- a/lib/iris/tests/integration/plot/test_vector_plots.py
+++ b/lib/iris/tests/integration/plot/test_vector_plots.py
@@ -31,6 +31,7 @@ import numpy as np
 
 import cartopy.crs as ccrs
 from iris.coords import AuxCoord, DimCoord
+from iris.coord_systems import Mercator
 from iris.cube import Cube
 from iris.tests.stock import sample_2d_latlons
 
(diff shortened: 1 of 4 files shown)


lib/iris/tests/results/imagerepo.json
```diff
@@ -1,952 +1,982 @@
 {
     "example_tests.test_COP_1d_plot.TestCOP1DPlot.test_COP_1d_plot.0": [
-        "https://scitools.github.io/test-iris-imagehash/images/v4/baff589936602d8ec977334ae4dac9b61a6dc4d99532c86cc2913e36c4cc0f61.png",
+        "https://scitools.github.io/test-iris-imagehash/images/v4/baff589936602d8ec977334ae4dac9b61a6dc4d99532c86cc2913e36c4cc0f61.png",
```
lbdreyer: This must be an editor thing. My PR removed all of the extra whitespace at the end of each line and yours adds it back in!

lib/iris/tests/integration/plot/test_vector_plots.py
```diff
@@ -0,0 +1,184 @@
...
+        # Plot against simple 1D x and y coords.
+        x, y, u, v = self._nonlatlon_xyuv()
+        u_cube, v_cube = self._nonlatlon_uv_cubes(x, y, u, v)
+        """Basic non-latlon, 1d coords testcase."""
+        self.plot('nonlatlon, 1-d coords', u_cube, v_cube)
+        plt.xlim(x.min() - 1, x.max() + 2)
+        plt.ylim(y.min() - 1, y.max() + 2)
+        self.check_graphic()
+
+    def test_nonlatlon_2d_coords(self):
```
lbdreyer: Your naming is inconsistent 
above it is `test_non_latlon_1d_coords` but this is `test_nonlatlon_2d_coords`

lib/iris/tests/integration/plot/test_vector_plots.py
```diff
@@ -0,0 +1,184 @@
...
+from iris.cube import Cube
+from iris.tests.stock import sample_2d_latlons
+
+# Run tests in no graphics mode if matplotlib is not available.
+if tests.MPL_AVAILABLE:
+    import matplotlib.pyplot as plt
+    from iris.plot import quiver, streamplot
+
+    # TEMP-TEST
+    plt.switch_backend('tkagg')
```
lbdreyer: don't you want to remove this?

lib/iris/tests/integration/plot/test_vector_plots.py
```diff
@@ -0,0 +1,184 @@
...
+        plt.ylim(y.min() - 1, y.max() + 2)
+        self.check_graphic()
+
+    def test_nonlatlon_2d_coords(self):
+        # Plot against expanded 2D x and y coords.
+        x, y, u, v = self._nonlatlon_xyuv()
+        x, y = np.meshgrid(x, y)
+        u_cube, v_cube = self._nonlatlon_uv_cubes(x, y, u, v)
+        # Call plot : N.B. default gives wrong coords order.
+        self.plot('nonlatlon_2d', u_cube, v_cube, coords=('x', 'y'))
```
lbdreyer: The naming is inconsistent for the plot names as well, above it is `'nonlatlon, 1-d coords'` but here it is `'nonlatlon_2d'` but changing that would also mean changing the images in the other PR so I'm happy to leave it as they are.

lib/iris/tests/integration/plot/test_vector_plots.py
pp-mo: :+1: will fix

lib/iris/tests/integration/plot/test_vector_plots.py
pp-mo: Oops !
will fix

lib/iris/tests/integration/plot/test_vector_plots.py
```diff
@@ -0,0 +1,184 @@
...
+        v_cube = u_cube.copy()
+        v_cube.rename('v')
+        v_cube.data = v
+        return u_cube, v_cube
+
+    def test_non_latlon_1d_coords(self):
+        # Plot against simple 1D x and y coords.
+        x, y, u, v = self._nonlatlon_xyuv()
+        u_cube, v_cube = self._nonlatlon_uv_cubes(x, y, u, v)
+        """Basic non-latlon, 1d coords testcase."""
```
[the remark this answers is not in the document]
pp-mo: Ok

lib/iris/tests/integration/plot/test_vector_plots.py
pp-mo: I actually did this because otherwise the plots were too similar (just 1 title character different), and then they both got the same hash value !

lib/iris/tests/integration/plot/test_vector_plots.py
```diff
@@ -0,0 +1,184 @@
...
+        u_cube.rename('dx')
+        u_cube.units = 'ms-1'
+        v_cube = u_cube.copy()
+        v_cube.rename('dy')
+        ny, nx = u_cube.shape
+        nn = nx * ny
+        angles = np.arange(nn).reshape((ny, nx))
+        angles = (angles * 360.0 / 5.5) % 360.
+        u_cube.data = np.cos(np.deg2rad(angles))
+        v_cube.data = np.sin(np.deg2rad(angles))
```
lbdreyer: You could add a scaling factor, otherwise all the vectors have the same length.
Although I suppose you tested different lengths in the first two non-lat lon tests

lib/iris/tests/results/imagerepo.json
pp-mo: But I didn't use an editor at all.
It seems to be just what happens when you run tests with "IRIS_TEST_CREATE_MISSING=1" ?

lib/iris/tests/integration/plot/test_vector_plots.py
pp-mo: I am looking into this ...

pp-mo:
To my horror, on closer inspection I found that all this is completely wrong-headed because streamplot _does not actually work_ for 2d coords **or even irregular 1-d coords**.
  * in the docs : ["x, y : 1d arrays / an **evenly spaced grid**."](https://matplotlib.org/api/_as_gen/matplotlib.pyplot.streamplot.html#matplotlib.pyplot.streamplot)
  * in [this stackoverflow answer](https://stackoverflow.com/a/34712773/2615050)

In fact, a closer look at the existing plots, even 1-D ones, shows it is not working as anticipated
So for example, I added some length scaling (as suggested by @lbdreyer above), and then **overlaid the quiver arrows on the streamplot**, to get this plot for the `TestStreamplot.test_non_latlon_1d_coords` case :
streamplot_nonlatlon_1d

That shows that streamplot is not _even_ right for 1-D coords if they are not regular -- which is after all exactly what the docs say !

Note that the 'iris.plot' code is actually broadcasting the X and Y coord arrays anyway, but I checked that and it seems to make no difference :  The streamplot function must somehow "permit" 2d inputs, but still giving bad results for anything irregular.

**Looks like we will need to remove streamplot entirely for now**
Please wait for me to fix ...

pp-mo:
Much simpler, but a disappointing outcome :sob: 
Please re-review @lbdreyer

lbdreyer:
Now that https://github.com/SciTools/test-iris-imagehash/pull/18 is merged, I have respun the tests to make sure they pass...

lbdreyer:
Tests are passing! I think this is now ready for merging :tada:

pp-mo:
Phew, thanks @lbdreyer !

pp-mo:
> @lbdreyer Tests are passing! I think this is now ready for merging tada

Agreed, I did it !

PR #3170, opened 2018-09-16
Update module headers and installation instructions

znicholls:
This PR updates a few module headers so that the tests pass (at least locally).

It also adds a slightly clearer explanation of how to get started with running the tests locally.


dkillick:
@znicholls looks good! Thanks for making this change 👍

PR #3171, opened 2018-09-16
Improve gracefully filling warning

znicholls:
This PR changes the behaviour of `iris.load` so that it only issues warnings about gracefully filling data and bounds if it actually does so.

### DIFF
--- a/lib/iris/fileformats/_pyke_rules/fc_rules_cf.krb
+++ b/lib/iris/fileformats/_pyke_rules/fc_rules_cf.krb
@@ -1621,7 +1621,7 @@ fc_extras
         attr_units = get_attr_units(cf_coord_var, attributes)
         points_data = cf_coord_var[:]
         # Gracefully fill points masked array.
-        if (ma.isMaskedArray(points_data)) and (not (points_data.mask == False).all()):
+        if ma.is_masked(points_data):
             points_data = ma.filled(points_data)
             msg = 'Gracefully filling {!r} dimension coordinate masked points'
             warnings.warn(msg.format(str(cf_coord_var.cf_name)))
@@ -1631,7 +1631,7 @@ fc_extras
         if cf_bounds_var is not None:
             bounds_data = cf_bounds_var[:]
             # Gracefully fill bounds masked array.
-            if (ma.isMaskedArray(bounds_data)) and (not (bounds_data.mask == False).all()):
+            if ma.is_masked(bounds_data):
                 bounds_data = ma.filled(bounds_data)
                 msg = 'Gracefully filling {!r} dimension coordinate masked bounds'
                 warnings.warn(msg.format(str(cf_coord_var.cf_name)))
--- a/lib/iris/fileformats/_pyke_rules/fc_rules_cf.krb
+++ b/lib/iris/fileformats/_pyke_rules/fc_rules_cf.krb
@@ -40,7 +40,7 @@ fc_default
 
 #
 # Context:
-#   This rule will trigger iff a grid_mapping() case specific fact 
+#   This rule will trigger iff a grid_mapping() case specific fact
 #   has been asserted that refers to a rotated pole.
 #
 # Purpose:
@@ -120,7 +120,7 @@ fc_provides_grid_mapping_mercator
 
 #
 # Context:
-#   This rule will trigger iff a grid_mapping() case specific fact 
+#   This rule will trigger iff a grid_mapping() case specific fact
 #   has been asserted that refers to a stereographic.
 #
 # Purpose:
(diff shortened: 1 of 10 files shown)


dkillick:
Hi @znicholls – congratulations on your first Iris PRs, and thanks for taking the time to make the submission! There are just a couple of things we could do with from you...

* you'll see that one of the checks that's failing is the CLA (Contributor License Agreement) checker. We ask all contributors to Iris (and the parent "SciTools" organisation) to submit a CLA with their first PR. Details on this can be found at https://scitools.org.uk/organisation.html#governance, in the "Contributors" bullet point.
* It looks like you've pulled in the docs changes you submitted in #3170 into this PR. Would you mind backing those changes out? No reason to submit the same things twice! It'll also make the changes you're proposing in this PR easier to spot.

Thanks again!

znicholls:
thanks @dkillick. I've just sent the CLA through as advised on the governance page. 

I've just backed out the changes. My editor is setup to strip whitespace at the end of lines by default, if that's a problem I can remove all those changes too.

dkillick:
Typically we try to avoid mixing code changes and stylistic changes in the same commit or PR, because it can make it harder to follow what's going on in the PR. This can, however, lead to stylistic changes never getting done! (As seems to be the case with this file). So not to worry about the whitespace removal; and thanks in fact for tidying up that file!

We'll try and progress the CLA soon as well - thanks for submitting one.

lib/iris/tests/unit/fileformats/pyke_rules/compiled_krb/fc_rules_cf_fc/test_build_dimension_coordinate.py
```diff
@@ -112,6 +112,60 @@ def test_dim_coord_construction(self):
...
+            # loading.
+            with self.deferred_load_patch, self.get_cf_bounds_var_patch:
+                build_dimension_coordinate(self.engine, self.cf_coord_var)
+
+                # Test that expected coord is built and added to cube.
+                self.engine.cube.add_dim_coord.assert_called_with(
+                    expected_coord, [0])
+
+            # Assert no warning is raised
+            assert len(w) == 0
```
dkillick: These new tests look great – but would you mind adding testing that covers the cases where the incoming coordinate has masked **bounds**? Thanks!

lib/iris/fileformats/_pyke_rules/fc_rules_cf.krb
```diff
@@ -1621,7 +1621,7 @@ fc_extras
         attr_units = get_attr_units(cf_coord_var, attributes)
         points_data = cf_coord_var[:]
         # Gracefully fill points masked array.
-        if ma.isMaskedArray(points_data):
+        if (ma.isMaskedArray(points_data)) and (not (points_data.mask == False).all()):
```
dkillick: You know, I think you can achieve equivalent behaviour with `if ma.is_masked(points_data)`. Would you mind updating this line to use that masked array check instead?

lib/iris/fileformats/_pyke_rules/fc_rules_cf.krb
```diff
@@ -1631,7 +1631,7 @@ fc_extras
         if cf_bounds_var is not None:
             bounds_data = cf_bounds_var[:]
             # Gracefully fill bounds masked array.
-            if ma.isMaskedArray(bounds_data):
+            if (ma.isMaskedArray(bounds_data)) and (not (bounds_data.mask == False).all()):
```
dkillick: And again with this line.

lib/iris/fileformats/_pyke_rules/fc_rules_cf.krb
znicholls: super nice pick up

znicholls:
> Typically we try to avoid mixing code changes and stylistic changes in the same commit or PR, because it can make it harder to follow what's going on in the PR.

Makes sense. To get the tests passing locally I had to update a bunch of headers. I can remove those changes again if you'd like but the tests may fail?

lib/iris/tests/unit/fileformats/pyke_rules/compiled_krb/fc_rules_cf_fc/test_build_dimension_coordinate.py
znicholls: have a look at the latest commit, I think this is what you wanted?

dkillick:
> I had to update a bunch of headers

Yes that does happen sometimes. Don't worry about it!

There are also a bunch of Python 3 test failures. It very much looks like these aren't being caused by this PR, so (separately from this PR) some work will need to happen to get those tests all passing again...

dkillick:
> a bunch of Python 3 test failures

I think these should be fixed by #3175 being merged. I'm going to restart the tests and see where that leaves us (now that the CLA checker bot is happy!).

dkillick:
Congratulations @znicholls for getting your first Iris PR merged 🎉 The change here is much appreciated!

znicholls:
Thanks @dkillick :) Shall I close #3170 as I think it's redundant?

dkillick:
>Shall I close #3170 ... ?

Actually, I think the note on [how to run the tests](https://github.com/SciTools/iris/blob/b10861647963c3e818b6aee4c1e4c1896f1c90b5/INSTALL#L69-L82) in #3170 is quite useful and not replicated here, so I think there's still value to it.

Issue #3169, opened 2018-09-16
Cell measure not used by collapsed


znicholls:
Apologies if I've missed something in the documentation or the issues. I've read through everything I could find and still can't see how to address this and hence think it's an issue with the library (which is why this is here and not in the Google Group, happy to move if you'd like).

I'm trying to use information from a netCDF file with areacella information to take a weighted mean of some spatial data. When I do the weighted mean using `iris.analysis.cartography.area_weights` then it all works smoothly. However if I add a `cell_measure` to my cube, I get the warning

```python
UserWarning: Collapsing spatial coordinate 'latitude' without weighting
```

All help is greatly appreciated. I put the minimal code to explain the error here.

```python
from os.path import join
import iris

tas_file = join(
    "..",
    "tests",
    "test_data",
    "cmip5",
    "1pctCO2",
    "Amon",
    "tas",
    "CanESM2",
    "r1i1p1",
    "tas_Amon_CanESM2_1pctCO2_r1i1p1_185001-198912.nc",
)

tas_cube = iris.load_cube(tas_file)

areacealla_file = join(
    "..",
    "tests",
    "test_data",
    "cmip5",
    "1pctCO2",
    "fx",
    "areacella",
    "CanESM2",
    "r0i0p0",
    "areacella_fx_CanESM2_1pctCO2_r0i0p0.nc",
)

areacella_cube = iris.load_cube(areacealla_file)

areacella_measure = iris.coords.CellMeasure(
    areacella_cube.data, 
    standard_name=areacella_cube.standard_name, 
    long_name=areacella_cube.long_name, 
    var_name=areacella_cube.var_name, 
    units=areacella_cube.units, 
    attributes=areacella_cube.attributes, 
    measure='area'
)

lat_lon_dims = [
    i
    for i, a in enumerate(tas_cube.dim_coords)
    if a.standard_name in ['latitude', 'longitude']
]
tas_cube.add_cell_measure(
    areacella_measure,
    data_dims=lat_lon_dims
)
print(tas_cube)
```

```python
air_temperature / (K)               (time: 1680; latitude: 64; longitude: 128)
     Dimension coordinates:
          time                           x               -              -
          latitude                       -               x              -
          longitude                      -               -              x
     Cell Measures:
          cell_area                      -               x              x
     Scalar coordinates:
          height: 2.0 m
     Attributes:
          CCCma_data_licence: 1) GRANT OF LICENCE - The Government of Canada (Environment Canada) is...
[...]

PR #3173, opened 2018-09-17
Use cell area for weights if available

znicholls:
This PR addresses #3169, allowing the user to use the weights provided in the cell measure more easily.

As mentioned in #3169, at the moment it's not that clear how to use cell measures for these weights. This is my attempt at doing so. It has a number of problems which I need help to address

- [ ] copying of code from `iris.analysis.cartography.area_weights`
- [ ] where the decision to use the area weights should go (feels wrong in `iris.analysis.cartography.area_weights` but is awkward to put in `iris.cube`)
- [ ] how cell measure should behave after collapsing (should cell area still be there, should it be the total area, the weighted average area?)
- [ ] how cell measure should appear in representation (I've made a change but not sure if it's what is desired)
- [ ] other PR bits e.g. docs, changelog

### DIFF
--- a/lib/iris/cube.py
+++ b/lib/iris/cube.py
@@ -3210,8 +3210,13 @@ bound=(1994-12-01 00:00:00, 1998-12-01 00:00:00)
 
         if (isinstance(aggregator, iris.analysis.WeightedAggregator) and
                 not aggregator.uses_weighting(**kwargs)):
-
-            if "cell_area" in [c.standard_name for c in self.cell_measures()]:
+            cell_area = [
+                c for c in self.cell_measures()
+                if c.measure == 'area'
+            ]
+            assert len(cell_area) < 2, "more than one cell area measure"
+            cell_area = cell_area[0] if cell_area else None
+            if cell_area is not None:
                 lon, lat = iris.analysis.cartography._get_lon_lat_coords(self)
 
                 # would be ideal to resuse the below from
@@ -3227,7 +3232,7 @@ bound=(1994-12-01 00:00:00, 1998-12-01 00:00:00)
                 lon_dim = self.coord_dims(lon)
                 lon_dim = lon_dim[0] if lon_dim else None
 
-                ll_weights = self.cell_measure('cell_area').data
+                ll_weights = cell_area.data
 
                 # Now we create an array of weights for each cell. This
                 # process will handle adding the required extra dimensions and
--- a/lib/iris/cube.py
+++ b/lib/iris/cube.py
@@ -3210,12 +3210,57 @@ bound=(1994-12-01 00:00:00, 1998-12-01 00:00:00)
 
         if (isinstance(aggregator, iris.analysis.WeightedAggregator) and
                 not aggregator.uses_weighting(**kwargs)):
-            msg = "Collapsing spatial coordinate {!r} without weighting"
-            lat_match = [coord for coord in coords
-                         if 'latitude' in coord.name()]
-            if lat_match:
-                for coord in lat_match:
-                    warnings.warn(msg.format(coord.name()))
+
+            if "cell_area" in [c.standard_name for c in self.cell_measures()]:
+                lon, lat = iris.analysis.cartography._get_lon_lat_coords(self)
+
+                # would be ideal to resuse the below from
+                # iris.analysis.cartography.area_weights rather than copying...
+                if lat.ndim > 1:
+                    raise iris.exceptions.CoordinateMultiDimError(lat)
+                if lon.ndim > 1:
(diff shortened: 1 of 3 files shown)


znicholls:
@dkillick if you have a minute, it would be great to get a few pointers about where to go with this. At the moment it's a feature I'd love to have in iris (rather than just wrapping around as is my current workwaround) so I'm happy to put some time into getting it up to scratch, there's just a few areas where I really don't know how best to proceed.

pp-mo:
Sorry to be so slow.
This definitely is interesting -- we'll get round to it, promise !

bjlittle:
Hey @znicholls, how's it going? 

Just pinging you on this PR to check on the state of play. Is this still an active WIP? Where you at?

Thanks 😄

znicholls:
Hey @bjlittle, yes still happy to have a go at this. As I mentioned earlier in the discussion, I just need some pointers on how best to add this feature. I've only added one test at the moment, is this enough? (I think there's also some code in there related to making html representations of cubes behave a bit nicer which I should remove and put in a different PR).

znicholls:
hi @bjlittle, is it worth revisiting this or have things moved in Iris since? I'm fiddling around with iris 3.0 and the html representation issue is certainly fixed!

schlunma:
@znicholls any plans to continue with this?

I didn't notice this pull request earlier and started implemented my own slightly different solution for this [here](https://github.com/schlunma/iris/tree/smarter_weights). See also https://github.com/SciTools/iris/issues/5082.

znicholls:
No plans to continue @schlunma, your solution looks much more elegant so please go for it! If you want to take the test I wrote and see if your solution passes that too, that would be great but no pressure.

schlunma:
Awesome! I think it won't pass the test, since my solution still needs the user to explicitly set `weights='cell_area'`, whereas yours uses the cell measure if possible even without the `weights` keyword (as far as I can tell) :+1:

pp-mo:
> Awesome! I think it won't pass the test, since my solution still needs the user to explicitly set `weights='cell_area'`, whereas yours uses the cell measure if possible even without the `weights` keyword (as far as I can tell) 👍

TBH now I've seen it I rather **_prefer_** that explicit keyword approach.  
You could use [`Cube._dimensional_metadata`](https://github.com/SciTools/iris/blob/57647d2c413dc3b85ecdde566a6340c9e43ad83b/lib/iris/cube.py#L1021-L1043) to select a matching cell-measure/aux-coord/ancillary-variable by name, which possibly may be a useful generalisation.  This has the advantage of automatic dimension mapping between the cube and its component, as shown in the code here.
We should also probably allow passing in such a thing (i.e. any DimensionalMetadata), but in that case we will have to take on trust that its dimensions map to the cube in the expected way -- so effectively that case is "just an array".
By contrast, I **_don't_** really favour automatically _calculating_ area-weights.  In those cases, I think it is "more Iris" to leave it to the user to provide what they they think is right.

These are just opinions !

znicholls:
Explicit is better than implicit so no complaints from me

schlunma:
> TBH now I've seen it I rather **_prefer_** that explicit keyword approach.

Fully agree!

> You could use [`Cube._dimensional_metadata`](https://github.com/SciTools/iris/blob/57647d2c413dc3b85ecdde566a6340c9e43ad83b/lib/iris/cube.py#L1021-L1043) to select a matching cell-measure/aux-coord/ancillary-variable by name, which possibly may be a useful generalisation. This has the advantage of automatic dimension mapping between the cube and its component, as shown in the code here. We should also probably allow passing in such a thing (i.e. any DimensionalMetadata), but in that case we will have to take on trust that its dimensions map to the cube in the expected way -- so effectively that case is "just an array".

Thanks for the pointer to `Cube._dimensional_metadata`, that's really helpful! Is there also a similar function to access to associated dimensions mappings? I.e. something that generalizes `Cube.coord_dims`, `Cube.cell_measure_dims`, etc.?

> By contrast, I **_don't_** really favour automatically _calculating_ area-weights. In those cases, I think it is "more Iris" to leave it to the user to provide what they they think is right.

Fully agree as well!

pp-mo:
> Is there also a similar function to access to associated dimensions mappings? I.e. something that generalizes `Cube.coord_dims`

Yes, sort-of.  All of the coords/cell-measures/ancillary-variable classes provide a [cube_dim](https://scitools-iris.readthedocs.io/en/stable/generated/api/iris/coords.html?highlight=cube_dims#iris.coords.AuxCoord.cube_dims) method. 
( Its actually an abstract method of _DimensionalMetadata : i.e. they all have it, but there is no common implementation )

PR #3175, opened 2018-09-23
Attempt to fix Matplotlib issue by resticting to 2.x

tomgreen66:
Matplotlib 3 was released a few days ago. Seems pulling in latest install from conda breaks a lot of functionality due to Matplotlib 3 not being supported.  Cartopy seems to have the biggest issues but might be other issues elsewhere.

For now suggest restricting to 2.x version of Matplotlib.

### DIFF
--- a/requirements/core.txt
+++ b/requirements/core.txt
@@ -7,7 +7,7 @@ cartopy
 cf_units>=2
 cftime
 dask[array]==0.18.1  #conda: dask==0.18.1
-matplotlib>=2
+matplotlib>=2,<3
 netcdf4
 numpy>=1.14
 scipy


dkillick:
Hi @tomgreen66, thanks for submitting this PR and getting the Iris tests passing again! For us to be able to progress this PR we're going to need you to sign a CLA, for which there are [instructions here](https://scitools.org.uk/organisation.html#governance) (see the "Contributors" bullet point). Thanks!

tomgreen66:
Hi - thanks for the pointer.  Just completed CLA and sent it in.

Thinking about it the main issue was in Cartopy so maybe I should submit the change to them as well?  I assume the dependency analysis will work if set in Iris but seems neater to fix it in Cartopy as well.  I would love to be able to submit a fix to make it work with Matplotlib 3 but this is a good first fix.

dkillick:
🎉 

Thanks very much @tomgreen66 - this should be really helpful for moving lots of Iris PRs forward!

PR #3176, opened 2018-09-26
fix for pandas column object bug

corinnebosley:
@DavidSimonin pointed out that if a 2-dimensional pandas DataFrame contains string-type column headers, then it is not currently being handled correctly.

This change treats the list of column headers as an array of strings rather than a single object, which enables subsequent operations to run without error.

### DIFF
--- a/lib/iris/pandas.py
+++ b/lib/iris/pandas.py
@@ -115,7 +115,8 @@ def as_cube(pandas_array, copy=True, calendars=None):
     _add_iris_coord(cube, "index", pandas_array.index, 0,
                     calendars.get(0, None))
     if pandas_array.ndim == 2:
-        _add_iris_coord(cube, "columns", pandas_array.columns, 1,
+        _add_iris_coord(cube, "columns", np.array([k for k in
+                                                   pandas_array.columns]), 1,
                         calendars.get(1, None))
     return cube
 
--- a/lib/iris/pandas.py
+++ b/lib/iris/pandas.py
@@ -115,8 +115,7 @@ def as_cube(pandas_array, copy=True, calendars=None):
     _add_iris_coord(cube, "index", pandas_array.index, 0,
                     calendars.get(0, None))
     if pandas_array.ndim == 2:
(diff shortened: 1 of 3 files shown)


corinnebosley:
@pp-mo
Travis build failing due to GeoAxesSubplot no longer having the attribute '_hold' in matplotlib 3.0.

May have to address this issue separately and rebase this later.

corinnebosley:
@pp-mo @dkillick Can somebody take a look at this while the tests are passing please?

lib/iris/pandas.py
```diff
@@ -115,7 +115,8 @@ def as_cube(pandas_array, copy=True, calendars=None):
     _add_iris_coord(cube, "index", pandas_array.index, 0,
                     calendars.get(0, None))
     if pandas_array.ndim == 2:
-        _add_iris_coord(cube, "columns", pandas_array.columns, 1,
+        _add_iris_coord(cube, "columns", np.array([k for k in
```
dkillick: This would be better done as `pandas_array.columns.values`, which returns you a NumPy array of the column headers in one shot.

corinnebosley: good shout @dkillick, I'll get on it.

corinnebosley: Hmmm....

corinnebosley: @dkillick The problem here is that this method turns all the column headers into objects.  Now that seems fine to me, but @DavidSimonin seemed to think that they all should have been strings.  I'm not convinced that the headers should be strings at all, I think you can probably extract the string from the object, but I might have to dig a little deeper to determine the appropriate behaviour.

corinnebosley: @dkillick I reckon that should be it.  I removed and regenerated the cml file (self-serving as that seems...), and I'm banking on the tests passing and the fix being good now.

dkillick:
Thanks @corinnebosley! And an extra 👍 for the word bambooish.

PR #3179, opened 2018-10-01
re-added fix for netcdf datasets always returning as MaskedArrays

corinnebosley:
This fix caused errors previously with versions of something or other.  Fix has been rebased (effectively).

I wonder what will happen...

### DIFF
--- a/lib/iris/fileformats/cf.py
+++ b/lib/iris/fileformats/cf.py
@@ -936,6 +936,11 @@ class CFReader(object):
 
         self._dataset = netCDF4.Dataset(self._filename, mode='r')
 
+        # Return variable data as np.ma.MaskedArray *only* when masked points
+        # are found in the slice requested, otherwise as 'normal' np.ndarray.
+        # Note: this suits us better, and was standard up to netcdf 1.3.
+        self._dataset.set_always_mask(False)
+
         # Issue load optimisation warning.
         if warn and self._dataset.file_format in ['NETCDF3_CLASSIC', 'NETCDF3_64BIT']:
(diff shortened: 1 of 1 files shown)


corinnebosley:
@pp-mo It looks very much like the attribute `set_always_mask` doesn't exist in netcdf4 1.4 for python 2.7.

I don't know how that can be so, but it is.  Probably.

corinnebosley:
Raised the issue as a question really with netcdf4 people:
https://github.com/Unidata/netcdf4-python/issues/848

Don't really want to wait for them to fox it and release it before we fix this our end though.  But it may be the case that we can only apply this fix to Python 3.6 environments though.  How do you feel about that?

pp-mo:
The problem here is that netcdf gets DOWNgraded when we install the grib api !
See the output in the logfile following the command 
`$ if [[ "$TEST_MINIMAL" != true && ${TRAVIS_PYTHON_VERSION} == 2* ]]; then conda install --quiet -n $ENV_NAME python-ecmwf_grib; pip install git+https://github.com/SciTools/iris-grib.git@v0.11.0; fi`

Of course, that only happens in Python 2 because we have no Python 3 grib api ...

corinnebosley:
@pp-mo aah, I see... well I'm still a bit confused about the conda list of packages that constituted the environment, but that certainly explains a lot.  OK, I'll close this and we'll continue with https://github.com/SciTools/iris/pull/3181

pp-mo:
> a bit confused about the conda list of packages that constituted the environment

The output from the first "conda list" command is ok, but it turns out not the final word as if iris-grib gets installed (Python2 only), that is done separately _and later_, so that modifies the original conda environment.

PR #3180, opened 2018-10-01
remove pin, see what blows up

corinnebosley:
This is a PR to enable iris to be compatible with matplotlib 3.x


QuLogic:
You are going to run into all the issues from SciTools/cartopy#1130; best to wait for 3.0.1 at least.

PR #3031, opened 2018-05-29
Unpin netcdf4

ocefpaf:
This needs #3016 

PS: I'm not sure if this is the right way to fix these tests!

PR #3181, opened 2018-10-01
Fix for netcdf4 returning masked array with no masked points.

pp-mo:
Alternative solution to #3179 

Hi @corinnebosley 
Let's see how this fix works out ?

### DIFF
--- a/lib/iris/fileformats/netcdf.py
+++ b/lib/iris/fileformats/netcdf.py
@@ -393,6 +393,10 @@ class NetCDFDataProxy(object):
             variable = dataset.variables[self.variable_name]
             # Get the NetCDF variable data and slice.
             var = variable[keys]
+            # Make into a 'plain' array if no masked points,
+            # for compatibility with netcdf <= 1.3.
+            if isinstance(var, np.ma.MaskedArray) and not np.ma.is_masked(var):
+                var = var.data
         finally:
             dataset.close()
         return np.asanyarray(var)


pp-mo:
> Let's see how this fix works out ?

The doctests are failing, possibly because a new version of "something" has forced numpy back from 0.15.2 to 0.15.0 ?
Tiresome 😞.  I will try to look into this tomorrow.

lib/iris/fileformats/netcdf.py
```diff
@@ -393,6 +393,10 @@ def __getitem__(self, keys):
             variable = dataset.variables[self.variable_name]
             # Get the NetCDF variable data and slice.
             var = variable[keys]
+            # Make into a 'plain' array if no masked points,
+            # for compatibility with netcdf <= 1.3.
+            if isinstance(var, np.ma.MaskedArray) and not np.ma.is_masked(var):
```
corinnebosley: @pp-mo I assume that you know what you're doing, but just for due diligence can you please confirm that `np.ma.is_masked` will check whether there are any masked points, rather than whether the array has a mask?

corinnebosley:
@pp-mo You've got a bunch of CMLs to regenerate (which I don't see a problem with), but that's all I can see as far as test failures are concerned.

lib/iris/fileformats/netcdf.py
pp-mo: It does, I checked !  
For some absolute pedantic certainty ...
```
>>> test_arrays = (('normal', np.arange(3)),
...                ('mask_false', np.ma.masked_array([1, 2])),
...                ('mask_empty', np.ma.masked_array([1, 2], mask=[0, 0])),
...                ('masked_points', np.ma.masked_array([1, 2], mask=[0, 1])))
>>> for name, array in test_arrays:
...     mask = array.mask if isinstance(array, np.ma.MaskedArray) else None
...     print('{:15s} :  mask={!s:15s} isMaskedArray={!s:5s} is_masked={!s:5s}'.format(name, mask, np.ma.isMaskedArray(array), np.ma.is_masked(array)))
... 
normal          :  mask=None            isMaskedArray=False is_masked=False
mask_false      :  mask=False           isMaskedArray=True  is_masked=False
mask_empty      :  mask=[False False]   isMaskedArray=True  is_masked=False
masked_points   :  mask=[False  True]   isMaskedArray=True  is_masked=True 
>>> 
```

pp-mo:
> a bunch of CMLs to regenerate (which I don't see a problem with)

I will also first check those out, as they don't fit my understanding of what netCDF "used to do".

lib/iris/fileformats/netcdf.py
corinnebosley: Awesome.  I'm convinced.

pp-mo:
> a bunch of CMLs to regenerate ... check those out, as they don't fit my understanding of what netCDF "used to do".

Ok, I think I have grasped it now.
All these CDL files with `mask_checksum="no-masked-elements"` in them _already_ changed before, when we upgraded for netcdf 1.4 : that occurred in #3031 just before we cut iris 2.1.0.
So the results of these operations (whether you get a masked array) *already* changed between v2.0 and v2.1.
And now we are proposing to change them all *back*.

Now I just need to form a view on whether that is a good idea (as some users have requested), or not ...

pp-mo:
Discussed off-line with @dkillick @pelson @bjlittle.
Concluded : Probably a bad idea -- instead, let's follow netcdf behaviour + pursue performance improvements in other ways if+when required

Withdrawn !

PR #3182, opened 2018-10-02
Improved Aux factory error handling: better message, delivered earlier

hdyson:
As per conversation with @corinnebosley , we had an issue where we were creating an orography as a standalone coordinate that wasn't attached to a cube.  Trying to use this in a hybrid height factory resulted in a hard to interpret error message.  I'm proposing adding a check when adding a hybrid height factory or other aux factory to a cube to ensure that the coordinates are present on the cube; and if not raising a hopefully more helpful error.

### DIFF
--- a/lib/iris/tests/unit/cube/test_Cube.py
+++ b/lib/iris/tests/unit/cube/test_Cube.py
@@ -1558,7 +1558,7 @@ class Test_add_metadata(tests.IrisTest):
         self.assertIsNone(cube.add_aux_factory(factory))
 
     def test_error_for_add_invalid_aux_factory(self):
-        cube = Cube(np.arange(8).reshape(2, 2, 2))
+        cube = Cube(np.arange(8).reshape(2, 2, 2), long_name='bar')
         delta = AuxCoord(points=[0, 1], long_name='delta', units='m')
         sigma = AuxCoord(points=[0, 1], long_name='sigma')
         orog = AuxCoord(np.arange(4).reshape(2, 2), units='m', long_name='foo')
@@ -1566,7 +1566,8 @@ class Test_add_metadata(tests.IrisTest):
         cube.add_aux_coord(sigma, 0)
         # Note orography is not added to the cube here
         factory = HybridHeightFactory(delta=delta, sigma=sigma, orography=orog)
-        expected_error = "foo coordinate for factory is not present on cube"
+        expected_error = ("foo coordinate for factory is not present on cube "
+                          "bar")
         with self.assertRaisesRegexp(ValueError, expected_error):
             cube.add_aux_factory(factory)
 
--- a/lib/iris/tests/unit/cube/test_Cube.py
+++ b/lib/iris/tests/unit/cube/test_Cube.py
@@ -1561,12 +1561,12 @@ class Test_add_metadata(tests.IrisTest):
         cube = Cube(np.arange(8).reshape(2, 2, 2))
         delta = AuxCoord(points=[0, 1], long_name='delta', units='m')
         sigma = AuxCoord(points=[0, 1], long_name='sigma')
-        orog = AuxCoord(np.arange(4).reshape(2, 2), units='m')
+        orog = AuxCoord(np.arange(4).reshape(2, 2), units='m', long_name='foo')
         cube.add_aux_coord(delta, 0)
         cube.add_aux_coord(sigma, 0)
         # Note orography is not added to the cube here
         factory = HybridHeightFactory(delta=delta, sigma=sigma, orography=orog)
-        expected_error = "Coordinate for factory is not present on cube"
+        expected_error = "foo coordinate for factory is not present on cube"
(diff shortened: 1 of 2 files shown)


hdyson:
Note that this doesn't check for when someone adds an aux factory to a cube and then updates the aux factory - `update` doesn't have cube information so there's no good way (that I can see) to check that the updated dependencies are valid.

lib/iris/tests/unit/cube/test_Cube.py
```diff
@@ -1546,6 +1546,30 @@ def test_add_cell_measure(self):
...
+    def test_add_valid_aux_factory(self):
+        cube = Cube(np.arange(8).reshape(2, 2, 2))
+        delta = AuxCoord(points=[0, 1], long_name='delta', units='m')
+        sigma = AuxCoord(points=[0, 1], long_name='sigma')
+        orog = AuxCoord(np.arange(4).reshape(2, 2), units='m')
+        cube.add_aux_coord(delta, 0)
+        cube.add_aux_coord(sigma, 0)
+        cube.add_aux_coord(orog, (1, 2))
+        factory = HybridHeightFactory(delta=delta, sigma=sigma, orography=orog)
+        self.assertIsNone(cube.add_aux_factory(factory))
```
corinnebosley: Sorry @hdyson, I'm a bit confuse about what this test is for.  Can you maybe add a comment or two to explain please?  I'm probably being slow but I don't really understand what we are asserting here.

lib/iris/tests/unit/cube/test_Cube.py
```diff
@@ -1546,6 +1546,30 @@ def test_add_cell_measure(self):
...
+    def test_error_for_add_invalid_aux_factory(self):
+        cube = Cube(np.arange(8).reshape(2, 2, 2))
+        delta = AuxCoord(points=[0, 1], long_name='delta', units='m')
+        sigma = AuxCoord(points=[0, 1], long_name='sigma')
+        orog = AuxCoord(np.arange(4).reshape(2, 2), units='m', long_name='foo')
+        cube.add_aux_coord(delta, 0)
+        cube.add_aux_coord(sigma, 0)
+        # Note orography is not added to the cube here
+        factory = HybridHeightFactory(delta=delta, sigma=sigma, orography=orog)
+        expected_error = "foo coordinate for factory is not present on cube"
```
corinnebosley: Looking at your error message in L.1001 in `iris.cube`, I would expect the cube's name to appear at the end of this string.  Does it not?

lib/iris/tests/unit/cube/test_Cube.py
hdyson: I didn't see an existing test for adding a valid aux factory.  Since I wanted to test for a particular error with an invalid aux factory, it felt like a good idea to include that test to confirm I wasn't breaking existing behaviour.  There are tests for adding dim coords, aux coords and cell measures, I think the lack of an existing test for an aux factory may be an oversight?

lib/iris/tests/unit/cube/test_Cube.py
hdyson: I was being lazy here - there's no name on the test cube.  I'll add one, and include it in the test output.

lib/iris/tests/unit/cube/test_Cube.py
corinnebosley: You're quite right.  I often despair at the testing (or lack of), but let's not get into that.  Adding tests that should already be there is great.

I'm still not quite sure what the `assertIsNone` asserts there is none of though.  That last line is slightly information dense for my delicate and easily-confused taste.

hdyson: Ah, okay, now I understand.  I'm using `assertIsNone` as an analogue for `assertDoesNotRaise` (which doesn't exist).  Would a comment help, or just removing the assertion?  Personally, it always makes me nervous when I see a test with no assert in it, it always makes me wonder if it's actually been finished or not.

corinnebosley: I think just a comment explaining the assert (as you just have to me, thanks for that) is enough.

corinnebosley:
@hdyson All your tests are passing and you've got lots of green ticks.  

The only thing I'm waiting for is a comment in the test explaining itself to simple people, then I'm happy to merge when the green tick comes back.

corinnebosley:
@hdyson Actually don't worry about it.  It's a pattern I haven't come across before, but I'm now using it in some tests I am working on now, and if I put a comment explaining what it does in every test it appears in you'd never be able to see the tests for waffle.

I'm happy to merge as-is and consider this a learning experience.

corinnebosley:
Boom! Nice work @hdyson, many thanks extended.

hdyson:
> 
> 
> Boom! Nice work @hdyson, many thanks extended.

Thanks @corinnebosley , I appreciate you picking this up so quickly 👍

PR #3187, opened 2018-10-05
seperate whatsnew entries for discontiguity checking and masking

corinnebosley:
DO NOT MERGE YET.

Waiting for:
- `iris.util` contiguity functions and tests have been merged (#3144 )
- another PR demonstrating the alternative option of a combined whatsnew entry for public contiguity checking/masking (#3188)

PR #3188, opened 2018-10-05
combined whatsnew entry for discontiguity checking and masking

corinnebosley:
DO NOT MERGE YET.

Waiting for:

- iris.util contiguity functions and tests have been merged (#3144 )
- another PR demonstrating the alternative option of separated whatsnew entries for public contiguity checking/masking (#3187 )

### DIFF
--- a/docs/iris/src/whatsnew/contributions_2.2.0/newfeature_2018-Oct-05_contiguity-checking-and-masking.txt
+++ b/docs/iris/src/whatsnew/contributions_2.2.0/newfeature_2018-Oct-05_contiguity-checking-and-masking.txt
@@ -1,0 +1,5 @@
+* Users can now check their cubes for any discontiguities in bounds arrays
+using the function :func:`iris.util.find_discontiguities_in_bounds`.
+Additionally, users can choose to mask their data at discontiguous points
+with the function :func:`iris.util.mask_data_at_discontiguities`, which could
+prevent plotting errors in some cases.


corinnebosley:
Hmmm... it occurs to me now that I'm not sure whether to call this 'contiguity checking' or 'discontiguity checking'.

I'm thinking it should probably be 'discontiguity checking' because this can easily be extended to 'discontiguity checking and masking'.  I don't think that 'contiguity checking and masking' is as correct.

pp-mo:
:+1: preferred to #3187

dkillick:
Yup, I prefer the combined whatsnew entry too 👍

docs/iris/src/whatsnew/contributions_2.2.0/newfeature_2018-Oct-05_contiguity-checking-and-masking.txt
```diff
@@ -0,0 +1,5 @@
+* Users can now check their cubes for any discontiguities in bounds arrays
```
dkillick: I don't like the use of "users" here. It implies that (a) there's an 'us' (devs) and 'them' (users), and not only that but (b) the whatsnew is by devs and for devs, and that users should be a little uncomfortable to also be here.

That said, it's hard to write these things! You can avoid the 'us' vs 'them' problem by using "you" in place of "users", which is more friendly, or not cast the whatsnew in terms of a person at all:

```
The function "find_discontiguities_in_bounds" can be used to check for discontiguities in the bounds arrays of cube coordinates.
Additionally, discontiguous points in coordinates can be explicitly masked using "mask_data_at_discontiguities".
```

Or something.

pp-mo: Well, I didn't expect you to hear you promoting use of the passive voice @dkillick !

Not that I object to that, but I know other people do.

corinnebosley:
I've put this change (augmented by @dkillick's suggestions) straight into the WhatsNew (#3192) instead of here.

PR #3189, opened 2018-10-05
Integration test for iris.plot.contour with 2d coords.

### LINKED ISSUE #3167: Add tests for 2d coords contouring
At present (https://github.com/SciTools/iris/tree/5923aaa8e725831a1575678fd93c938bfad36e95), we test :
  * various tests for 2d coord handling functions in tests.unit.plot : `test__get_plot_defn[_custom_coords_picked]`, `test__check_bounds_contiguity_and_mask`
  * integration tests for pcolormesh in tests.integration.plot : `test_plot_2d_coords`.

We ought to ***also*** have an integration test proving 2d coords operation in contour/contourf 
Probably also belongs in  `test_plot_2d_coords`

pp-mo:
Needs https://github.com/SciTools/test-iris-imagehash/pull/21

Closes #3167

### DIFF
--- a/lib/iris/tests/integration/plot/test_plot_2d_coords.py
+++ b/lib/iris/tests/integration/plot/test_plot_2d_coords.py
@@ -26,11 +26,10 @@ from six.moves import (filter, input, map, range, zip)  # noqa
 # importing anything else
 import iris.tests as tests
 
+import cartopy.crs as ccrs
+import matplotlib.pyplot as plt
 import numpy as np
 
-import matplotlib.pyplot as plt
-
-import cartopy.crs as ccrs
 import iris
 from iris.analysis.cartography import unrotate_pole
 from iris.cube import Cube
@@ -70,7 +69,6 @@ class Test(tests.GraphicsTest):
 
 
 @tests.skip_plot
-@tests.skip_data
 class Test2dContour(tests.GraphicsTest):
     def test_2d_coords_contour(self):
         ny, nx = 4, 6
--- a/lib/iris/tests/integration/plot/test_plot_2d_coords.py
+++ b/lib/iris/tests/integration/plot/test_plot_2d_coords.py
@@ -26,10 +26,16 @@ from six.moves import (filter, input, map, range, zip)  # noqa
 # importing anything else
 import iris.tests as tests
 
-import cartopy.crs as ccrs
+import numpy as np
+
 import matplotlib.pyplot as plt
 
+import cartopy.crs as ccrs
 import iris
+from iris.analysis.cartography import unrotate_pole
+from iris.cube import Cube
+from iris.coords import AuxCoord
+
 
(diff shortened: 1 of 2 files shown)


lib/iris/tests/integration/plot/test_plot_2d_coords.py
```diff
@@ -26,10 +26,16 @@
 # importing anything else
 import iris.tests as tests
 
-import cartopy.crs as ccrs
+import numpy as np
```
dkillick: @pp-mo why is the `NumPy` import separate at the top? At minimum I feel it should be after the `pyplot` import. Preferably they would both be after the Iris imports.

lib/iris/tests/integration/plot/test_plot_2d_coords.py
```diff
@@ -63,5 +69,29 @@ def test_2d_coord_bounds_northpolarstereo(self):
         self.check_graphic()
 
 
+@tests.skip_plot
+@tests.skip_data
```
dkillick: @pp-mo you shouldn't need this decorator as you're not making use of external test data.

pp-mo: :+1: cut+paste error, I think.
Good spot !!

lib/iris/tests/integration/plot/test_plot_2d_coords.py
pp-mo: Presumably you mean "NumPy ... after the pyplot ... ***before*** the Iris imports" ??

My understanding of our guidelines for imports is to separate into 3 or 4 specific groups : 
  1. standard-library packages
  2. other 3rd party libraries
  3. "own" packages + modules
  4. (in tests only) test subjects

So here, numpy and pyplot should be in the same block, 
but before any Iris stuff 
(except iris.tests, which is required to happen first. :nauseated_face: yech ).

This agree with your understanding @dkillick ?

dkillick: > This agree with your understanding

Yes... and no, at the same time! Basically, I feel like we've slipped away from separating (2.) and (3.), even in test modules. It's not worth sweating about here, but I would still like to see the `NumPy` import at least after the `pyplot` import 😉

lib/iris/tests/integration/plot/test_plot_2d_coords.py
dkillick: Ah yes, that old chestnut 🌰

lib/iris/tests/integration/plot/test_plot_2d_coords.py
pp-mo: :+1:

dkillick:
Whoops, one of the Travis jobs had an error. I've restarted it, whcih hopefully that will solve the problem.

PR #3196, opened 2018-10-11
Revert "Public contiguity checking"

pp-mo:
Reverts SciTools/iris#3144

Note to self : do not put emojis in commit messages !!


QuLogic:
> Note to self : do not put emojis in commit messages !!

Hmm? Why's that?

pp-mo:
> Hmm? Why's that?

Ok, probably not important.
I mistook the commit message box for the comment box, and I thought we might get weird unicode characters in the log.  But I think it is ok.

PR #3197, opened 2018-10-11
Fix docstring re structured text warnings (#3073)

corinnebosley:


dkillick:
Note that this pulls `master` in line with `v2.1.x`.

PR #3194, opened 2018-10-11
bumped python 3.6 to python 3.7 in travis yaml

corinnebosley:
Following Travis test failures in Python 3.6, this is a little test to see whether bumping the yaml to use Python 3.7 just fixes that.

PR #3203, opened 2018-10-15
see if travis likes python 3.7-dev

corinnebosley:
In #3194 I tried upgrading the python matrix to look like this:
```
language: python
python:
  - 2.7
  - 3.7
sudo: false
```
This didn't work because Travis doesn't actually support Python 3.7 fully yet.

Suggested workaround is to try this instead:
```
language: python
python:
  - 2.7
  - 3.7-dev
sudo: false
```
So that's what this PR is.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -8,7 +8,7 @@
 language: python
 python:
   - 2.7
-  - 3.6
+  - 3.7-dev
 sudo: false
 
 env:


corinnebosley:
Travis does not yet support python 3.7 (unable to download from source).  Closing this down.

QuLogic:
You need to flip a few more bits to get 3.7 working: https://github.com/travis-ci/travis-ci/issues/9069#issuecomment-425720905

PR #3205, opened 2018-10-22
Remove specific NumPy pin

dkillick:
Remove the pin to not NumPy v1.15.2, as this is [now apparently fixed](https://github.com/conda-forge/iris-feedstock/pull/41#discussion_r224928363).

Now, the only question is, why are the checks not running?


dkillick:
Something weird is going on here. I'm going to try closing and re-opening, and see if that makes the weirdness disappear.

dkillick:
Nope, still weird. I'm just going to start again.

PR #3206, opened 2018-10-22
Remove specific NumPy pin

dkillick:
Remove the `not v1.15.2` NumPy pin from the Iris requirements, as the problem with this build has [apparently been fixed](https://github.com/conda-forge/iris-feedstock/pull/41#discussion_r224928363).

### DIFF
--- a/requirements/core.txt
+++ b/requirements/core.txt
@@ -10,5 +10,5 @@ cftime
 dask[array]  #conda: dask
 matplotlib>=2,<3
 netcdf4
-numpy>=1.14,!=1.15.2
+numpy>=1.14
 scipy


dkillick:
@QuLogic I'm going to remove this from the v2.3.0 milestone - we can benefit from this now; no need to wait for v2.3 to do so.

QuLogic:
Oh, I didn't realize there's no `v2.2.x` branch yet; shouldn't there be? The 2.2.0 release has been made already.

ajdawson:
>Oh, I didn't realize there's no v2.2.x branch yet; shouldn't there be?

This seems like an oversight, normally there should be a v2.2.x branch for maintenance fixes. @dkillick what's the plan here, are we doing something different this time?

dkillick:
> This seems like an oversight

Yup - it's exactly that. Will correct soon! Apols.

dkillick:
Right, the `v2.2.x` branch exists. Thanks for pointing out that it was missing @QuLogic @ajdawson.

PR #3215, opened 2018-11-06
pinned cftime to less than 1.0.2.1

corinnebosley:


bjlittle:
The latest release of `cftime` at `1.0.2.1` is breaking iris master in [travis-ci](https://travis-ci.org/SciTools/iris/builds/448959261).

Happy to have a **short-term** pin to maintain functioning tests, whilst we investigate the underlying issue with latest `cftime` version.

corinnebosley:
thanks @bjlittle

dkillick:
Y'know, looking at the [cftime releases](https://github.com/Unidata/cftime/releases), I wonder whether we should be using v1.0.0 - that seems to be the most recent "proper" release.

PR #3222, opened 2018-11-09
Use Python3.7 in CI

tv3141:
## Changes to `.travis.yaml`

For testing with Travis, Python is installed using miniconda and only this version of Python is used. Specifying `- python: 3.6` in .travis.yml sets the version of Python that comes installed in the Travis container.
Python3.7 caused some problems  travis-ci/travis-ci#9069, and requires a workaround using Ubuntu Xenial.

```yaml
python:
- 2.7

matrix:
  include:
    - python: 3.7
      dist: xenial
      sudo: true
```

Instead:

- use minimal travis container
- specify Python version as env var

- [x] get Travis to work with Python3.7

## Fix failing tests with Python3.7

- [ ] `iris.math.IFunc` should only accept ~~numpy ufuncs~~ functions accepting numpy arrays

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -5,18 +5,19 @@
 # we will use conda to give us a much faster setup time.
 
 
-language: python
-python:
-  - 2.7
-  - 3.6
+language: minimal
 sudo: false
 
 env:
   matrix:
-    - TEST_TARGET=default TEST_MINIMAL=true
-    - TEST_TARGET=default
-    - TEST_TARGET=example
-    - TEST_TARGET=doctest
+    - PYTHON_VERSION=2.7 TEST_TARGET=default TEST_MINIMAL=true
+    - PYTHON_VERSION=3.6 TEST_TARGET=default TEST_MINIMAL=true
+    - PYTHON_VERSION=2.7 TEST_TARGET=default
+    - PYTHON_VERSION=3.6 TEST_TARGET=default
+    - PYTHON_VERSION=2.7 TEST_TARGET=example
+    - PYTHON_VERSION=3.6 TEST_TARGET=example
+    - PYTHON_VERSION=3.6 TEST_TARGET=doctest
+    # doctests only with Python3, see https://github.com/SciTools/iris/pull/3134
 
   global:
     # The decryption key for the encrypted .github/deploy_key.scitools-docs.enc.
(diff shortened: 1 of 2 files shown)


.travis.yml
```diff
@@ -5,18 +5,19 @@
...
-    - TEST_TARGET=example
-    - TEST_TARGET=doctest
+    - PYTHON_VERSION=2.7 TEST_TARGET=default TEST_MINIMAL=true
+    - PYTHON_VERSION=3.7 TEST_TARGET=default TEST_MINIMAL=true
+    - PYTHON_VERSION=2.7 TEST_TARGET=default
+    - PYTHON_VERSION=3.7 TEST_TARGET=default
+    - PYTHON_VERSION=2.7 TEST_TARGET=example
+    - PYTHON_VERSION=3.7 TEST_TARGET=example
+    - PYTHON_VERSION=3.7 TEST_TARGET=doctest
+    # doctests only with Python3, see https://github.com/SciTools/iris/pull/3134
```
bjlittle: The downside, for me, is that really we should be covering py27, py36 and py37... At this point, we've lost the convenience of the matrix...

ocefpaf: The alternative that I know works with Travis-CI is not better than this. It would require something like:

```yaml
python:
- 2.7
- 3.6

matrix:
  include:
    - python: 3.7
      dist: xenial
      sudo: true
```

tv3141: Maybe some reordering into blocks by python version can make this more appealing:
```yaml
env:
  matrix:
    - PYTHON_VERSION=2.7 TEST_TARGET=default TEST_MINIMAL=true
    - PYTHON_VERSION=2.7 TEST_TARGET=default
    - PYTHON_VERSION=2.7 TEST_TARGET=example
    # doctests only with Python3, see https://github.com/SciTools/iris/pull/3134
    #- PYTHON_VERSION=2.7 TEST_TARGET=doctest

    - PYTHON_VERSION=3.6 TEST_TARGET=default TEST_MINIMAL=true
    - PYTHON_VERSION=3.6 TEST_TARGET=default
    - PYTHON_VERSION=3.6 TEST_TARGET=example
    - PYTHON_VERSION=3.6 TEST_TARGET=doctest

    - PYTHON_VERSION=3.7 TEST_TARGET=default TEST_MINIMAL=true
    - PYTHON_VERSION=3.7 TEST_TARGET=default
    - PYTHON_VERSION=3.7 TEST_TARGET=example
    - PYTHON_VERSION=3.7 TEST_TARGET=doctest
```

tv3141:
I removed the failing test, but I think the intention was that `iris.math.IFunc` only accepts numpy ufuncs, which is not the case anymore.

CAM-Gerlach:
FYI, the ``sudo:`` key is obsolete and can be removed, as the containerized infrastructure it refers to was removed from Travis years ago.

bjlittle:
Closed by https://github.com/SciTools/iris/pull/3350

Thanks @tv3141 :+1:

PR #3232, opened 2018-11-27
fix for mercator test failures

corinnebosley:
- Added `latitude_true_scale=0.0` args to expected result in mercator tests
- Pinned cftime to !=1.0.2.1 for 2.2.x branch


bjlittle:
Thanks @corinnebosley :+1:

PR #3233, opened 2018-11-29
Fix proportion aggregator attempt 2

corinnebosley:


bjlittle:
@corinnebosley Isn't this targeting the proportion aggregation (take II) rather then the mercator test (which got merged)?

corinnebosley:
@bjlittle Yes.  And I think that proves to myself that I should go home.  I'm not ok again yet if I can't read words and I don't know what I'm doing.

bjlittle:
@corinnebosley Thanks :+1: 

LGTM

PR #3240, opened 2018-12-13
Support wider range of climatologies on pp/fields file read

jkettleb:
Could you give me feedback on this pull request - it isn't for merging at this stage - I just want to gauge your view on it.

This ticket aims to interpret a wider range of climatology fields from pp and fields files. Currently iris only interprets multi year climatologies that are means of means (For instance, multi year means of monthly mean data, multi year means of daily data).  I think the wording of F03 suggests we can do more than this.

The F03 document says:
  IB = 3 if the field is a time mean from T1 to T2 for each year from LBYR to LBYRD.
  
This will always be 'mean over years' as a CF cell method.  The statistical processing within the year is given by lbproc.  So and lbproc of 128 is a mean within years, lbproc of 4096 is a minumum within the year. 

Do you agree?

### DIFF
--- a/lib/iris/fileformats/pp_load_rules.py
+++ b/lib/iris/fileformats/pp_load_rules.py
@@ -983,17 +983,12 @@ def _all_other_rules(f):
                                            coords='time',
                                            intervals=intervals))
             unhandled_lbproc = False
-        elif f.lbtim.ib == 3 and f.lbproc == 128:
-            # Aggregation over a period of time within a year, over a number
-            # of years.
-            # Only mean (lbproc of 128) is handled as the min/max
-            # interpretation is ambiguous e.g. decadal mean of daily max,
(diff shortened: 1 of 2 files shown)


jkettleb:
The original comment wasn't very clear - so I have tried to update.  Please contact me though if its still unclear.

pp-mo:
I have some problems with this, but I don't think we can discuss it here as the key format specifications aren't public property.
@jkettleb I'll contact you offline ...

jkettleb:
Thanks @pp-mo.  I have looked into this in more detail.  Tools that set lbtim are not all consistent with my interpretation of F03. lbtim=3 does not guarantee that the cell_methods should be 'mean over years'.  For instance there are processing tools that use lbtim=3 to reprensent fields that would have cell_methods 'variance over years'.  I'll close this PR.

PR #3241, opened 2018-12-13
[ENH] Concatenate usability improvements

DPeterK:
Improve the usability of concatenate by:

* allowing concatenate to operate on cubes with differing data / coordinate points / coordinate bounds dtypes
* producing more descriptive error messages on failing to concatenate to a single cube.

Here's an example of some of the improved error messages:

```python
cubes.concatenate_cube()
```
```python-traceback
  ...
iris.exceptions.ConcatenateError: failed to concatenate into a single cube.
  Dimension coordinate metadata differs: name (latitude, longitude, time != longitude)
  Auxiliary coordinate metadata differs: units (Unit('1') != Unit('m'))
  Data types differ: float32 != <U32
  Scalar coordinate metadata differs: name (height != None)
  Data dimensions differ: 3 != 1
  Scalar coordinate metadata differs: long_name (None != 'alice')
```

### DIFF
--- a/lib/iris/_concatenate.py
+++ b/lib/iris/_concatenate.py
@@ -346,8 +346,9 @@ class _CubeSignature(object):
         self.aux_metadata = []
         self.dim_coords = cube.dim_coords
         self.dim_metadata = []
-        self.ndim = cube.ndim
+        self.scalar_coords = []
         self.scalar_metadata = []
+        self.ndim = cube.ndim
         self.cell_measures_and_dims = cube._cell_measures_and_dims
         self.dim_mapping = []
 
@@ -391,6 +392,7 @@ class _CubeSignature(object):
             else:
                 metadata = _CoordMetaData(coord, None)
                 self.scalar_metadata.append(metadata)
+                self.scalar_coords.append(coord)
 
     @staticmethod
     def _coordinate_attribute_differences(name, coord_1, coord_2):
@@ -925,7 +927,7 @@ class _ProtoCube(object):
             aux_coords_and_dims.append((coord.copy(), dims))
 
         # Generate all the scalar coordinates for the new concatenated cube.
-        for coord in cube_signature.scalar_metadata:
+        for coord in cube_signature.scalar_coords:
             aux_coords_and_dims.append((coord.copy(), ()))
 
         return aux_coords_and_dims
--- a/lib/iris/_concatenate.py
+++ b/lib/iris/_concatenate.py
@@ -364,13 +364,43 @@ class _CubeSignature(object):
             else:
                 self.scalar_coords.append(coord)
 
-    def _can_and_will_cast(self, other, attr):
+    @staticmethod
+    def _coordinate_attribute_differences(coord_1, coord_2):
         """
-        Determine whether the dtypes of `self.attr` and `other.attr` can be
-        cast. If they can be, cast both to be of the same appropriate dtype.
+        Determine the coordinate attributes that differ between `coord_1`
+        and `coord_2`. Note that if the dtypes of the points and bounds
+        attributes can be cast safely, they count as not different.
+
+
+        Args:
+
+        * coord_1: (_CoordMetaData)
+            The first _CoordMetaData object to compare the attributes of.
+
+        * coord_2: (_CoordMetaData)
+            The second _CoordMetaData object to compare the attributes of.
+
+        Returns:
+            A list of the keys that differ between `coord_1` and
+            `coord_2`.
 
         """
-        pass
+        diffs = []
+        coord_1_defn = coord_1._asdict()
+        coord_2_defn = coord_2._asdict()
+        for coord_1_key, coord_1_value in coord_1_defn.items():
+            coord_2_value = coord_2_defn[coord_1_key]
+            if 'dtype' in coord_1_key:
+                values = (coord_1_value, coord_2_value)
+                dtype_is_none = any([v is None for v in values])
+                can_cast = any((np.can_cast(*values),
+                                np.can_cast(*values[::-1])))
+                if can_cast and not dtype_is_none:
+                    diffs.append(coord_1_key)
+            else:
+                if coord_1_value != coord_2_value:
+                    diffs.append(coord_1_key)
+        return diffs
 
     def _coordinate_differences(self, other, attr):
         """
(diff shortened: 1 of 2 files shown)


bjlittle:
@DPeterK Did you fancy rebasing? We recently done a merge back of `2.2.x` to `master`, which may resolve your current travis-ci issues...

DPeterK:
@bjlittle done!

lib/iris/_concatenate.py
```diff
@@ -360,8 +390,63 @@ def key_func(coord):
                 coord_and_dims = _CoordAndDims(coord, tuple(dims))
                 self.aux_coords_and_dims.append(coord_and_dims)
             else:
+                metadata = _CoordMetaData(coord, None)
```
pp-mo: Tiny style note:
This 'metadata = ' line is effectively the same as the call on line 388 (taken when 'if dims:' succeeds), so it might be clearer to do this once before the 'if dims:' test.

lib/iris/_concatenate.py
```diff
@@ -331,18 +359,20 @@ def __init__(self, cube):
         self.defn = cube.metadata
         self.data_type = cube.dtype
 
-        #
-        # Collate the dimension coordinate metadata.
-        #
+        # Populate dim- and aux-coord metadata.
+        self._collate_dim_metadata(cube)
```
pp-mo: Another finnicky style point ...
I can see the point of separating the complicated bits into separate functions, but the effect of these 2 lines is now opaque.
Could we maybe return the data and assign it to self here to make it clearer ??  
I *think* those lists appended to aren't actually modified anywhere else, so we can make all this much clearer.
E.G.
```
    self.dim_metadata, self.dim_mapping = self._collate_dim_metadata(cube)
    self.aux_metadata, self.aux_coords_and_dims = self._collate_aux_metadata(cube)
```
(and don't pre-init them to empty lists above, either).

lib/iris/_concatenate.py
```diff
@@ -377,33 +462,34 @@ def _coordinate_differences(self, other, attr):
...
-            self_dict = {'< None >': None}
+            self_dict = {'None': None}
         if len(other_dict) == 0:
-            other_dict = {'< None >': None}
+            other_dict = {'None': None}
         self_names = sorted(self_dict.keys())
         other_names = sorted(other_dict.keys())
 
         # Compare coord metadata.
         if len(self_names) != len(other_names) or self_names != other_names:
```
pp-mo: This test is unnecessarily complex : 
I think just `self_names != other_names` does it all ?

pp-mo:
RE: "I've checked the before + afters on the error messages for the testcases in TestMessages, and it's not all an improvement" ...

I hacked the test code to report all the actual error messages and ran new code against old.
The results, with my ratings ...

```
 FAIL: test_aux_coords_difference_message (__main__.TestMessages)
-  Auxiliary coordinates differ: foo, wibble != wibble"
+  Auxiliary coordinate metadata differs: name (foo, wibble != wibble)"
```
Verdict :-1: 
"metadata" seems redundant and unhelpful : I think we should _only_ use the word when it *is* the metadata of the coords which differs (e.g. attributes or units) , rather than the set of coords themselves.
Also, "name" seems positively unhelpful :  What is actually different is the list of aux coords (not really their names) .. see below for an idea about this ...

----
```
 FAIL: test_aux_coords_metadata_difference_message (__main__.TestMessages)
-  Auxiliary coordinates metadata differ: foo != foo"
+  Auxiliary coordinate metadata differs: 'foo' units (Unit('1') != Unit('m'))"
```
Verdict : :+1: definitely better :bouquet: !!


----
```
 FAIL: test_dimensions_difference_message (__main__.TestMessages)
-  Dimension coordinates differ: latitude, longitude, time != longitude, time"
+  Dimension coordinate metadata differs: name (latitude, longitude, time != longitude, time)"
```
Verdict : :-1:
as for "test_aux_coords_difference_message" above, dislike both "metadata" and "name"

----
```
 FAIL: test_dimensions_metadata_difference_message (__main__.TestMessages)
-  Dimension coordinates metadata differ: latitude != latitude"
+  Dimension coordinate metadata differs: 'latitude' long_name (None != 'bob')"
```
Verdict : :+1:

----
```
 FAIL: test_dims_covered_difference_message (__main__.TestMessages)
-  Auxiliary coordinates metadata differ: wibble != wibble"
+  Auxiliary coordinate metadata differs: 'wibble' dims ((0, 1, 2) != (0, 1))"
```
Verdict : :+1:

----
```
 FAIL: test_ndim_difference_message (__main__.TestMessages)
-  Dimension coordinates differ: latitude, longitude, time != longitude
-  Auxiliary coordinates differ: foo, wibble != < None >
-  Scalar coordinates differ: height != < None >
+  Dimension coordinate metadata differs: name (latitude, longitude, time != longitude)
+  Auxiliary coordinate metadata differs: name (foo, wibble != None)
+  Scalar coordinate metadata differs: name (height != None)
   Data dimensions differ: 3 != 1"
```
Verdict : :-1:
Same stuff again re 'metadata' and 'name'.

----
```
 FAIL: test_scalar_coords_difference_message (__main__.TestMessages)
-  Scalar coordinates differ: height != < None >"
+  Scalar coordinate metadata differs: name (height != None)"
```
Verdict : :-1:

----
```
 FAIL: test_scalar_coords_metadata_difference_message (__main__.TestMessages)
-  Scalar coordinates metadata differ: height != height"
+  Scalar coordinate metadata differs: 'height' long_name (None != 'alice')"
```
Verdict : :+1:

----
----
**A Thought:**
The "name" term appears to come from [a single common line in the code](https://github.com/SciTools/iris/blob/f683aebbd119bf2e6eb80c6544eba5a47328550f/lib/iris/_concatenate.py#L481)
I think in that case, it could be taken from the 'attr' keyword to replace 'name' with a more specific useful term, e.g. with the mapping
```
type_context = {
    'dim_metadata' : 'dimension coordinates',
    'aux_metadata':  'auxiliary coordinates',
    'scale_metadata':  'scalar coordinates'}[attr]
```

pp-mo:
@DPeterK What do you think ?
Hope this is not discouraging, there is a lot of good waiting to come out of this one !

DPeterK:
@pp-mo thanks for the thorough review! Not discouraging; this is a hard problem to solve and it would be good to get it right, or at least better than currently.

> t's a shame we've combined it with the more "cosmetic" changes

Yes. I think I started one thing (the dtype support, if memory serves) and it flowed naturally into making the cosmetic changes and I was lazy and kept going in the same PR rather than starting afresh...

> The "name" term appears to come from a single common line in the code

It's a long time ago that I implemented this now, but I think that the reason I adjusted the error message in the way I did is that I could use just one template (good on account of consistency and simplicity) that, I felt, improved the error message in the majority of use-cases. That said, having a mapping that expands `name` as per the suggestion may help solve this problem.

pp-mo:
@SciTools/peloton unfortunately it does look like this is effectively dead now, not least because the whole way we do metadata (and what that word means) has since changed.
Closing for now -- please reopen @DPeterK if you still think we can move this forward.

PR #3245, opened 2018-12-18
Handle no time coord when saving to PP

dkillick:
In cases where a cube without a time coordinate is saved to PP, the default value for a PP field header element (0) is not changed for the elements that make up the header's `T1` and `T2` elements. This now causes errors to surface in cftime, because the 0 values in the header elements are interpreted as a datetime value of `0000-00-00`, which cftime does not accept as a valid date. See, for example, [this failing Iris test](https://travis-ci.org/SciTools/iris/jobs/467504312#L7082), which I think originates from this.

### DIFF
--- a/lib/iris/fileformats/pp_save_rules.py
+++ b/lib/iris/fileformats/pp_save_rules.py
@@ -124,12 +124,19 @@ def _general_time_rules(cube, pp):
     cm_time_min = scalar_cell_method(cube, 'minimum', 'time')
     cm_time_max = scalar_cell_method(cube, 'maximum', 'time')
 
+    # Handle case of no time coordinate.
+    if time_coord is None and fp_coord is None and frt_coord is None:
+        # Deal with the cftime error:
+        # "date 0000-00-00 does not exist in the standard calendar".
+        pp.t1 = cftime.datetime(1, 1, 1)
+        pp.t2 = cftime.datetime(1, 1, 1)
+
     # No forecast.
     if time_coord is not None and fp_coord is None and frt_coord is None:
         pp.lbtim.ia = 0
         pp.lbtim.ib = 0
         pp.t1 = time_coord.units.num2date(time_coord.points[0])
-        pp.t2 = cftime.datetime(0, 0, 0)
+        pp.t2 = cftime.datetime(1, 1, 1)
 
     # Forecast.
     if (time_coord is not None and
--- a/lib/iris/tests/unit/fileformats/pp/test_save.py
+++ b/lib/iris/tests/unit/fileformats/pp/test_save.py
@@ -201,6 +201,44 @@ class Test_Save__LbprocProduction(tests.IrisTest):
         self.assertEqual(lbproc, 8192)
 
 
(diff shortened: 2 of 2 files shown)


dkillick:
Following an offline discussion with @pp-mo I wonder if the initial direction of this PR (altering header values on save) isn't the best approach:

* the problem surfaces at load time, not save
* what's being saved isn't _wrong_, per se, as `T1` and `T2` values set to 0 in the PP header have a defined meaning – it's just that `cftime` doesn't like the most obvious interpretation of the `T1` and `T2` values in this case.

A better solution might be to leave save alone and handle this case at load time instead. This brings a different set of questions on how to handle such cases when they're encountered at load time:

* catch such cases and edit the values given by `T1` and `T2` to produce a time point at the nearest valid date (`0001-01-01`) – meaning Iris would actively alter a file's metadata 
* assume that `T1` and `T2` values set to 0 mean that there _isn't_ a time coordinate to be derived from `T1` and/or `T2` and not make one – fine perhaps for PP datasets produced by Iris produced by Iris as that's effectively how such values are used, but potentially more unclear for PP files from other sources.

In both cases a load-save round-trip maintaining an identical result becomes impossible. I think, however, that Iris already cannot produce an identical round-trip when saving a cube from a NetCDF file (with no time coordinate) to a PP file and re-loading it, because the PP save step will introduce a new time coordinate that did not exist in the NetCDF file.

Neither of these are desirable in many ways as they both mean a change in behaviour for some cases. Arguably though the problem here is caused by Iris, which has never interpreted such cases correctly, and it's now actually starting to bite. My preference is to not create a time coordinate in such cases (the second option above) as that seems to be the less bad interpretation of such values for `T1` and `T2`, and also raise a warning to inform the user that this is happening.

I'm going to start re-implementing this along these lines, but other opinions are also welcome.

dkillick:
In fact, as that basically amounts to starting again, I'm going to do just that.

pp-mo:
> Iris... has never interpreted such cases correctly... My preference is to not create a time coordinate in such cases

:+1: to this

jkettleb:
For what it's worth year 0 often appear in ancillary files.  We are recommending that we phase out this practice and replace with a real date.  The year could be 'representative'.  https://code.metoffice.gov.uk/doc/ancil/ants/latest/appendixA.html#date-information

This makes sense for ancillaries derived from climatologies - the climatology will have real years attached - for instance 1800-1860 for preindustrial, 2008-2018 for present day.

It also makes sense for 'fixed' fields like land sea mask and orography - these can vary for paleo runs or runs with changing ice sheet thickness. I realise the use for fixed fields is less useful - and it opens up other cans of worms like support for paleoclimate  calendars - http://cfconventions.org/Data/cf-conventions/cf-conventions-1.7/cf-conventions.html#calendar).  It avoids the year=0 problem though.

pp-mo:
>  phase out this practice and replace with a real date

A very reasonable approach, but for the purposes of this discussion we surely need to remain focussed on supporting legacy files -- especially as what used to work just broke :cry:

Issue #3239, opened 2018-12-08
cftime again


rcomer:
It looks like [new versions of cftime have been released](
https://github.com/Unidata/cftime/releases) in the last week.  Travis tests are failing again with the same errors we were getting before #3215, which pinned cftime to not 1.0.2.1.  For example of failing tests, see #3238.

cpelley:
This effects a number of our users.
To that end, #3215 suggests that it is pinning cftime to a version less than 1.0.2.1
The change however, just specifically precludes version 1.0.2.1

Currently, doing a conda install of iris via conda-forge will lead to problems to our users due to cftime.

I'm going to propose that this be dealt with in two stages (the first, an immediate fix giving breathing room for the second):
- Stage:1 Return the pre-existing behaviour by pinning cftime to <1.0.2.1
- Stage:2 Look at iris supporting cftime >= 1.0.2.1
  - To be clear here.  Any change of behaviour here might have significant impact to users so needs careful consideration.  For example, a number of ancillaries in the wild have time information populated with 0.  For better or for worse, common/historical usage is of importance here.

pp-mo:
Re: 
> Stage:2 Look at iris supporting cftime >= 1.0.2.1 ...  ancillaries in the wild ... common/historical usage

#3245 mentions this.
@DPeterK initially proposed to adjust the iris saving
But we then agreed it is better to fix loading (too?)

This is now awaiting a more suitable solution.
#3248 refers

At bottom, we should really be loading data with '0 times' with ***no*** time coord.
We can then save such with '0 times', too.
This makes good sense, but it does involve behaviour change.

cpelley:
- Additional context: this issue doesn't appear to effect the usecase from #1926 (this is where there is a year 0 but non-zero month, day, ...).  Identified by the following tests passing:
`iris.tests.unit.fileformats.pp_load_rules.test__all_other_rules.TestLBTIMx2x_ZeroYears`
`iris.tests.unit.fileformats.pp_load_rules.test__convert_time_coords.TestLBTIMx2x_ZeroYear`

pp-mo:
For now, I'm proposing to "fix" this by wrapping the usage in `iris.fileformats.pp_load_rules` to reproduce the "old-style" results for a date-to-num conversion.
#3259

That's really just a patch "for now" : it doesn't address any of the better ideas raised here, 
But if problems are confined to PP/FF conversion, that may be good enough + we can close this down ??

pp-mo:
[...]

Issue #3225, opened 2018-11-12
Error loading ancillary file with 2.2.0


MartinDix:
With 2.1.0, there's no problem with non-time varying ancillary files. E.g.
```
>>> c = iris.load_cube('/PATH/umdir/ancil/atmos/n216e/land_sea_mask/igbp/v4/qrparm.mask')
>>> print(c)
land_binary_mask / (1)              (latitude: 324; longitude: 432)
     Dimension coordinates:
          latitude                           x               -
          longitude                          -               x
     Attributes:
          STASH: m01s00i030
          source: Data from Met Office Unified Model
          um_version: 9.2
```

However with 2.2.0 it seems to be trying to create a time axis and is upset by the year 0.
```
>>> c = iris.load_cube('/PATH/umdir/ancil/atmos/n216e/land_sea_mask/igbp/v4/qrparm.mask')
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "/PATH/lib/python3.6/site-packages/iris/__init__.py", line 373, in load_cube
    cubes = _load_collection(uris, constraints, callback)
  File "/PATH/lib/python3.6/site-packages/iris/__init__.py", line 313, in _load_collection
    result = iris.cube._CubeFilterCollection.from_cubes(cubes, constraints)
  File "/PATH/lib/python3.6/site-packages/iris/cube.py", line 145, in from_cubes
    for cube in cubes:
  File "/PATH/lib/python3.6/site-packages/iris/__init__.py", line 300, in _generate_cubes
    for cube in iris.io.load_files(part_names, callback, constraints):
  File "/PATH/lib/python3.6/site-packages/iris/io/__init__.py", line 207, in load_files
    constraints):
  File "/PATH/lib/python3.6/site-packages/iris/fileformats/rules.py", line 429, in load_cubes
    user_callback_wrapper=loadcubes_user_callback_wrapper):
  File "/PATH/lib/python3.6/site-packages/iris/fileformats/rules.py", line 350, in _load_pairs_from_fields_and_filenames
    cube, factories, references = _make_cube(field, converter)
  File "/PATH/lib/python3.6/site-packages/iris/fileformats/rules.py", line 293, in _make_cube
    metadata = converter(field)
  File "/PATH/lib/python3.6/site-packages/iris/fileformats/pp_load_rules.py", line 665, in convert
    t1=f.t1, t2=f.t2, lbft=f.lbft)
[...]

PR #3247, opened 2019-01-07
BLD: Pinned cftime 1.0.1

cpelley:
#3239, #3215, #3225

### DIFF
--- a/requirements/core.txt
+++ b/requirements/core.txt
@@ -6,7 +6,7 @@
 cartopy
 #conda: proj4<5
 cf_units>=2
-cftime!=1.0.2.1
+cftime==1.0.1
 dask[array]  #conda: dask
 matplotlib>=2,<3
 netcdf4


rcomer:
My user code is failing with cftime version 1.0.0b1 so I am not sure if `<1.0.2.1` will solve all problems (although getting Travis checks to pass is obviously desirable!)  What version would Travis be picking up here?  I'm sure that info must be in the output somewhere but I can't see it!

cpelley:
Travis is using cftime 1.0.1 (conda-forge)

pp-mo:
>  cftime version 1.0.0b1 ... What version would Travis be picking up here? I'm sure that info must be in the output somewhere but I can't see it!

You can find this from a Travis log : 
To view a log, first click on the tests icon by the commit in Github (green tick or red cross!),
navigate to the "continuous-integration/travis-ci/pr ... Details". 
Hit another tick (or cross) for the log of one testing "phase" 
(e.g. python 2.7, TEST_TARGET=default, TEST_MINIMAL=True). 

In the log, look for the line "$ conda list -n $ENV_NAME ....",
the initial wedge ">" opens the collapsed command output ...
which is the list of installed packages !!

In this case, "cftime 1.0.1"

rcomer:
Thanks both.  I hadn't even twigged that there was collapsed output in the Travis log!

requirements/core.txt
```diff
@@ -6,7 +6,7 @@
 cartopy
 #conda: proj4<5
 cf_units>=2
-cftime!=1.0.2.1
+cftime<1.0.2.1
```
bjlittle: Why not pin directly to `1.0.1`, just to make this explicit and clear.

We still need to fix `iris` to deal with 0 based dates, which are now illegal in the latest `cftime`.

So this pin would only be a temporary pin...

cpelley: > ...Why not pin directly...

Sure.

> We still need to fix iris to deal with 0 based dates...

Agreed.  Some relating discussion in #3239

bjlittle:
@cpelley Apologies. Sorry to be a pain...

But could you please target the `2.2.x` branch instead of `master`, and then I'll merge back the `2.2.x` branch to `master`...

cpelley:
NP :)

requirements/core.txt
```diff
@@ -6,7 +6,7 @@
 cartopy
 #conda: proj4<5
 cf_units>=2
-cftime!=1.0.2.1
+cftime>=1.0.1,<=1.0.1
```
bjlittle: @cpelley Surely `cftime==1.0.1` is what you mean, right? Doesn't that work?

bjlittle:
Thanks @cpelley :+1: 

Ping @pp-mo

PR #3250, opened 2019-01-11
BUG: Cube equality boolean data

cpelley:
Issue introduced in numpy 1.14 I think where subtracting two boolean arrays was deprecated.

```Python
  File "/path/to/iris/cube.py", line 3024, in __eq__
    result = np.all(np.abs(self.data - other.data) < 1e-8)
TypeError: numpy boolean subtract, the `-` operator, is deprecated, use the bitwise_xor, the `^` operator, or the logical_xor function instead.
```

This may be better placed as a more general numpy comparison operator in iris but I don't see there is anywhere suitable for such a thing to live currently.

### DIFF
--- a/lib/iris/cube.py
+++ b/lib/iris/cube.py
@@ -1,4 +1,4 @@
-# (C) British Crown Copyright 2010 - 2018, Met Office
+# (C) British Crown Copyright 2010 - 2019, Met Office
 #
 # This file is part of Iris.
 #
@@ -3018,11 +3018,10 @@ bound=(1994-12-01 00:00:00, 1998-12-01 00:00:00)
                 result = not (coord_comparison['not_equal'] or
                               coord_comparison['non_equal_data_dimension'])
 
-            # having checked everything else, check approximate data
-            # equality - loading the data if has not already been loaded.
+            # Having checked everything else, check approximate data equality.
             if result:
-                result = np.all(np.abs(self.data - other.data) < 1e-8)
-
+                result = da.allclose(self.core_data(),
+                                     other.core_data()).compute()
         return result
 
     # Must supply __ne__, Python does not defer to __eq__ for negative equality
--- a/docs/iris/src/whatsnew/2.2.rst
+++ b/docs/iris/src/whatsnew/2.2.rst
@@ -85,6 +85,8 @@ Iris 2.2 Dependency updates
 Bugs Fixed
 ==========
 
+* Cube equality of boolean data is now handled correctly.
+
 * The bug has been fixed that prevented printing time coordinates with bounds
   when the time coordinate was measured on a long interval (that is, ``months``
   or ``years``).


rcomer:
Why not just use `numpy.allclose`?

rcomer:
> Why not just use `numpy.allclose`?

Or even `dask.array.allclose`.

lib/iris/cube.py
```diff
@@ -3021,8 +3021,7 @@ def __eq__(self, other):
             # having checked everything else, check approximate data
             # equality - loading the data if has not already been loaded.
             if result:
-                result = np.all(np.abs(self.data - other.data) < 1e-8)
-
+                result = da.array.allclose(self.data, other.data)
```
rcomer: `da` is already `dask.array` and the function returns a dask array.  So I think you need:
```python
da.allclose(self.data, other.data).compute()
```

lib/iris/cube.py
```diff
@@ -3021,8 +3021,7 @@ def __eq__(self, other):
             # having checked everything else, check approximate data
             # equality - loading the data if has not already been loaded.
```
rcomer: If we're happy with using `da.allclose`, I guess this comment about loading the data can be removed?

pp-mo: If this is using 'self.data', 'other.data' then of course it is still realising the cubes.

In fact I don't think we can use dask here, as the result needs to be a concrete boolean, whereas da.allclose returns a lazy scalar value, even when the args are concrete ...
```
>>> da.allclose(np.array([1,2]), np.array([1,1]))
dask.array<all-aggregate, shape=(), dtype=bool, chunksize=()>
>>> _.compute()
False
>>> 
```

rcomer: > If this is using 'self.data', 'other.data' then of course it is still realising the cubes.

Ah yes of course, you'd want to use `core_data()` or `lazy_data()`

Definitely in learning mode here...

Is there no advantage to using `da.allclose` with the `compute` call?  I was thinking it would access the files in chunks and therefore avoid busting memory.

Aside: that thing you did with the underscore, I did not know was a thing :open_mouth: .

pp-mo: >  you'd want to use core_data() 

I think it makes sense to use 'core_data' as it involves the minimum of extra work.
My key point was : the routine itself *must* still return a regular boolean, not a lazy value.
... but I see now you have just added the ".compute()" above ! :+1: 

> the underscore

Yes, it's where the interpreter puts unassigned results (in addition to printing them).
Also very handy when you forgot to save the result of a calculation !
```
>>> long_and_complicated()
BigObject(...)
>>> keep_that = _
```

bjlittle:
@pp-mo Fancy taking this on?

@cpelley I think that you're going to have to re-base to fold in the latest Travis CI changes

pp-mo:
Thanks @bjlittle 
@cpelley I think this is good to go, once we have a green tick on it. 🥕 🙏

lbdreyer:
@cpelley Sorry about this! The travis tests are failing but I believe the fix for those tests went in already. If you rebase we should be able to get a green tick.

Do you mind rebasing and then this can go in?

lib/iris/cube.py
```diff
@@ -3018,11 +3018,10 @@ def __eq__(self, other):
...
 
-            # having checked everything else, check approximate data
-            # equality - loading the data if has not already been loaded.
+            # Having checked everything else, check approximate data equality.
             if result:
-                result = np.all(np.abs(self.data - other.data) < 1e-8)
-
+                result = da.allclose(self.core_data(),
+                                     other.core_data()).compute()
         return result
```
pp-mo: Here's my comment

pp-mo:
Hi @cpelley !
Jog : can you get this rebased in a day or so,  and maybe add a whatsnew item (not essential:  we can make something up) ?
Then it can go into Iris 2.3.

pp-mo:
#3431 replaced : closing.

PR #3255, opened 2019-01-31
Dask landsea masks bugfix

### LINKED ISSUE #3237: Don't nest compute calls
In some cases Iris is nesting a compute call within a compute call. This has the undesirable behaviour of generating a loop that increments the total number of threads being used by the realisation operation, which can cause various problems, particularly when running on a cluster:

*  in the less serious, running in parallel on the cluster takes a lot longer than running synchronously on the cluster (if you interrupt the job it looks like the majority of the time is taken up switching threads)
* in the more serious, the maximum number of allowed threads is hit and the job fails.

This is happening in a case where a cube loaded from a PP file with `lbpack=2` (packed on masked land/sea points) is realised. In this case compute is called twice, one call nested inside the other, for a single PP field - the original call to realise the field's data and the nested call to realise the attached land/sea mask that comes from a separate PP field.

Here's the full stacktrace:
```python
cube.data
```
```python-traceback
  File "/.../create_jules_ic_v2.py", line 279, in <module>
    main()                           
  File "/.../create_jules_ic_v2.py", line 274, in main
    nameoffile=create_jules_ic(Ddate,INDIR,UM_FILE_IN,OUTDIR)              
  File "/.../create_jules_ic_v2.py", line 152, in create_jules_ic
    print(type(cube_vegfrac.data <=0.0001))  
  File "/.../python2.7/site-packages/iris/cube.py", line 1644, in data
    return self._data_manager.data
  File "/.../python2.7/site-packages/iris/_data_manager.py", line 216, in data
    result = as_concrete_data(self._lazy_array)
  File "/.../python2.7/site-packages/iris/_lazy_data.py", line 128, in as_concrete_data
    data = np.asanyarray(data.compute())
  File "/.../python2.7/site-packages/dask/base.py", line 143, in compute
    (result,) = compute(self, traverse=False, **kwargs)
  File "/.../python2.7/site-packages/dask/base.py", line 392, in compute
    results = get(dsk, keys, **kwargs)
  File "/.../python2.7/site-packages/dask/threaded.py", line 75, in get
    pack_exception=pack_exception, **kwargs)
  File "/.../python2.7/site-packages/dask/local.py", line 521, in get_async
    raise_exception(exc, tb)
  File "/.../python2.7/site-packages/dask/local.py", line 290, in execute_task
    result = _execute_task(task, data)
  File "/.../python2.7/site-packages/dask/local.py", line 270, in _execute_task
    args2 = [_execute_task(a, cache) for a in args]
  File "/.../python2.7/site-packages/dask/local.py", line 271, in _execute_task
    return func(*args2)
  File "/.../python2.7/site-packages/dask/array/core.py", line 74, in getter
    c = a[b]
  File "/.../python2.7/site-packages/iris/fileformats/pp.py", line 647, in __getitem__
    self.mdi, self.mask)
  File "/.../python2.7/site-packages/iris/fileformats/pp.py", line 770, in _data_bytes_to_shaped_array
    land_mask = mask.data.astype(np.bool)
  File "/.../python2.7/site-packages/iris/fileformats/pp.py", line 1033, in data
    self._data = as_concrete_data(self._data)
  File "/.../python2.7/site-packages/iris/_lazy_data.py", line 128, in as_concrete_data
    data = np.asanyarray(data.compute())
  File "/.../python2.7/site-packages/dask/base.py", line 143, in compute
    (result,) = compute(self, traverse=False, **kwargs)
  File "/.../python2.7/site-packages/dask/base.py", line 392, in compute
    results = get(dsk, keys, **kwargs)
  File "/.../python2.7/site-packages/dask/threaded.py", line 70, in get
    pool = ThreadPool(num_workers)
  File "/.../python2.7/multiprocessing/pool.py", line 727, in __init__
    Pool.__init__(self, processes, initializer, initargs)
  File "/.../python2.7/multiprocessing/pool.py", line 159, in __init__
    self._repopulate_pool()
  File "/.../python2.7/multiprocessing/pool.py", line 223, in _repopulate_pool
    w.start()
  File "/.../python2.7/multiprocessing/dummy/__init__.py", line 75, in start
    threading.Thread.start(self)
  File "/.../python2.7/threading.py", line 736, in start
    _start_new_thread(self.__bootstrap, ())
thread.error: can't start new thread
```

To fix this we can make sure that the original field and the land/sea mask are realised together and not separately. Note that there might be other cases where such nested compute calls might be possible within Iris.

Originally raised as https://github.com/dask/dask/issues/4240, but re-raised here as it's now clear that the origin of the problem is in Iris.

pp-mo:
Fix the problem with landsea-masked PPFields, where the dask compute() calls PPDataProxy.\_\_getitem__, which can then make a nested 'compute()' call for the landsea-mask data.

Closes #3237

  * [x] Contains content from #3247 , so needs that --or equivalent-- merging, and then a rebase, before merge.
  * [x] Also needs https://github.com/SciTools/iris-test-data/pull/52~

### DIFF
--- a/lib/iris/fileformats/pp.py
+++ b/lib/iris/fileformats/pp.py
@@ -1642,7 +1642,7 @@ def _interpret_fields(fields):
 def _create_field_data(field, data_shape, land_mask_field=None):
     """
     Modifies a field's ``_data`` attribute either by:
-     * converting DeferredArrayBytes into a lazy array,
+     * converting a 'deferred array bytes' tuple into a lazy array,
      * converting LoadedArrayBytes into an actual numpy array.
 
     If 'land_mask_field' is passed (not None), then it contains the associated
@@ -1735,8 +1735,8 @@ def _field_gen(filename, read_data_bytes, little_ended=False):
 
     A field returned by the generator is only "half-formed" because its
     `_data` attribute represents a simple one-dimensional stream of
-    bytes. (Encoded as an instance of either LoadedArrayBytes or
-    DeferredArrayBytes, depending on the value of `read_data_bytes`.)
+    bytes. (Encoded either as an instance of LoadedArrayBytes or as a
+    'deferred array bytes' tuple, depending on the value of `read_data_bytes`.)
     This is because fields encoded with a land/sea mask do not contain
     sufficient information within the field to determine the final
     two-dimensional shape of the data.
@@ -1825,7 +1825,10 @@ def _field_gen(filename, read_data_bytes, little_ended=False):
                 pp_field.data = LoadedArrayBytes(pp_file.read(data_len),
                                                  dtype)
             else:
-                # Provide enough context to read the data bytes later on.
+                # Provide enough context to read the data bytes later on,
+                # as a 'deferred array bytes' tuple.
+                # N.B. this used to be a namedtuple called DeferredArrayBytes,
+                # but it no longer is. Possibly for performance reasons?
                 pp_field.data = (filename, pp_file.tell(), data_len, dtype)
                 # Seek over the actual data payload.
                 pp_file_seek(data_len, os.SEEK_CUR)
@@ -2242,8 +2245,6 @@ def save_fields(fields, target, append=False):
         of the file.
         Only applicable when target is a filename, not a file handle.
         Default is False.
-    * callback:
-        A modifier/filter function.
 
     See also :func:`iris.io.save`.
 
@@ -2272,11 +2273,9 @@ def save_fields(fields, target, append=False):
 
     if isinstance(target, six.string_types):
         pp_file = open(target, "ab" if append else "wb")
-        filename = target
     elif hasattr(target, "write"):
         if hasattr(target, "mode") and "b" not in target.mode:
             raise ValueError("Target not binary")
-        filename = target.name if hasattr(target, 'name') else None
         pp_file = target
     else:
         raise ValueError("Can only save pp to filename or writable")
--- a/lib/iris/fileformats/pp.py
+++ b/lib/iris/fileformats/pp.py
@@ -1588,22 +1588,20 @@ def _interpret_fields(fields):
     numpy arrays (via the _create_field_data) function.
 
     """
-    land_mask = None
+    land_mask_field = None
     landmask_compressed_fields = []
     for field in fields:
         # Store the first reference to a land mask, and use this as the
         # definitive mask for future fields in this generator.
-        if land_mask is None and field.lbuser[6] == 1 and \
+        if land_mask_field is None and field.lbuser[6] == 1 and \
                 (field.lbuser[3] // 1000) == 0 and \
                 (field.lbuser[3] % 1000) == 30:
-            land_mask = field
+            land_mask_field = field
 
-        apply_landmask = None
-
-        # Handle land compressed data payloads,
-        # when lbpack.n2 is 2.
+        # Handle land compressed data payloads, when lbpack.n2 is 2.
         if (field.raw_lbpack // 10 % 10) == 2:
-            if land_mask is None:
+            # Field uses land-mask packing, so needs a related land-mask field.
+            if land_mask_field is None:
                 landmask_compressed_fields.append(field)
                 # Land-masked fields have their size+shape defined by the
                 # reference landmask field, so we can't yield them if they
@@ -1612,42 +1610,45 @@ def _interpret_fields(fields):
                 # the end of the file.
                 continue
 
-            # Landmask-compressed fields don't have an lbrow and lbnpt.
-            field.lbrow, field.lbnpt = land_mask.lbrow, land_mask.lbnpt
-            # Construct a data array using the landmask field as a template.
-            apply_landmask = land_mask
+            # Land-mask compressed fields don't have an lbrow and lbnpt.
+            field.lbrow, field.lbnpt = \
+                land_mask_field.lbrow, land_mask_field.lbnpt
+            _create_field_data(field, (field.lbrow, field.lbnpt),
+                               land_mask_field=land_mask_field)
+        else:
+            # Field does not use land-mask packing.
+            _create_field_data(field, (field.lbrow, field.lbnpt))
 
-        _create_field_data(field, (field.lbrow, field.lbnpt),
-                           with_landmask_field=apply_landmask)
         yield field
 
     # At file end, return any land-masked fields that were deferred because
     # they were encountered before the landmask reference field.
     if landmask_compressed_fields:
-        if land_mask is None:
+        if land_mask_field is None:
             warnings.warn('Landmask compressed fields existed without a '
                           'landmask to decompress with. The data will have '
                           'a shape of (0, 0) and will not read.')
             mask_shape = (0, 0)
         else:
-            mask_shape = (land_mask.lbrow, land_mask.lbnpt)
+            mask_shape = (land_mask_field.lbrow, land_mask_field.lbnpt)
 
         for field in landmask_compressed_fields:
             field.lbrow, field.lbnpt = mask_shape
             _create_field_data(field, mask_shape,
-                               with_landmask_field=land_mask)
+                               land_mask_field=land_mask_field)
             yield field
 
 
-def _create_field_data(field, data_shape, with_landmask_field=None):
+def _create_field_data(field, data_shape, land_mask_field=None):
     """
     Modifies a field's ``_data`` attribute either by:
      * converting DeferredArrayBytes into a lazy array,
      * converting LoadedArrayBytes into an actual numpy array.
 
-    If 'with_landmask_field' is passed, it is another field :  The landmask
-    field's data is used as a template for this field's data, determining its
-    size, shape and the locations of valid (non-missing) datapoints.
+    If 'land_mask_field' is passed (not None), then it contains the associated
+    landmask, which is also a field :  Its data array is used as a template for
+    `field`'s data, determining its size, shape and the locations of all the
+    valid (non-missing) datapoints.
 
     """
     if isinstance(field.core_data(), LoadedArrayBytes):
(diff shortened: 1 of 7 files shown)


pp-mo:
Passes at last :champagne: 
Mergeable ? :wink:

bjlittle:
@pp-mo Can you target the `v2.2.x` branch instead?

It would be good to bank this fix in a `2.2.1` release :wink:

pp-mo:
For some reason it wouldn't rebase cleanly, so I skwished + cherry-picked it
( also backed out an unintended change to tests/test_pp_module )
So it had to be force-pushed, but there should be nothing you don't recognise from before.

But all passing so @bjlittle it's go !

lib/iris/fileformats/pp.py
```diff
@@ -1591,20 +1598,31 @@ def _interpret_fields(fields):
                 (field.lbuser[3] % 1000) == 30:
             land_mask = field
 
+        apply_landmask = None
```
bjlittle: @pp-mo Minor point - The PP loading logic is pretty heavy going (always has been). So to help the reader I think it would be make a stronger association that `apply_landmask = None` is the `else` of `if (field.raw_lbpack // 10 % 10) == 2:` ... would you mind doing that?

lib/iris/fileformats/pp.py
```diff
@@ -1591,20 +1598,31 @@ def _interpret_fields(fields):
...
         # Handle land compressed data payloads,
         # when lbpack.n2 is 2.
         if (field.raw_lbpack // 10 % 10) == 2:
             if land_mask is None:
                 landmask_compressed_fields.append(field)
+                # Land-masked fields have their size+shape defined by the
+                # reference landmask field, so we can't yield them if they
+                # are encountered *before* the landmask.
+                # In that case, defer them, and process them all afterwards at
+                # the end of the file.
```
bjlittle: @pp-mo Nice clarification, thanks!

lib/iris/fileformats/pp.py
```diff
@@ -1591,20 +1598,31 @@ def _interpret_fields(fields):
...
                 continue
 
-            # Land compressed fields don't have a lbrow and lbnpt.
+            # Landmask-compressed fields don't have an lbrow and lbnpt.
```
bjlittle: @pp-mo Minor point - Just to match the terminology of your previous [comment](https://github.com/SciTools/iris/pull/3255/files#diff-066f828420b27bfd2a03868477b98415R1608), could you change this to `Land-masked compressed fields..`?

lib/iris/fileformats/pp.py
bjlittle: Also change `apply_landmask` to `land_mask_field`, after all that's what it actually is. The use of `apply_landmask` subtly hints (to me) that it's a boolean, and its not.

lib/iris/fileformats/pp.py
```diff
@@ -1591,20 +1598,31 @@ def _interpret_fields(fields):
...
             field.lbrow, field.lbnpt = land_mask.lbrow, land_mask.lbnpt
+            # Construct a data array using the landmask field as a template.
+            apply_landmask = land_mask
 
-        data_shape = (field.lbrow, field.lbnpt)
-        _create_field_data(field, data_shape, land_mask)
+        _create_field_data(field, (field.lbrow, field.lbnpt),
+                           with_landmask_field=apply_landmask)
```
bjlittle: @pp-mo Could we change this to be:
```python
_create_field_data(field, (field.lbrow, field.lbnpt),
                   land_mask_field=land_mask_field)
```

lib/iris/fileformats/pp.py
```diff
@@ -1616,16 +1634,21 @@ def _interpret_fields(fields):
 
         for field in landmask_compressed_fields:
             field.lbrow, field.lbnpt = mask_shape
-            _create_field_data(field, (field.lbrow, field.lbnpt), land_mask)
+            _create_field_data(field, mask_shape,
+                               with_landmask_field=land_mask)
```
bjlittle: @pp-mo Again...
```python
_create_field_data(field, mask_shape,
                   land_mask_field=land_mask)
```

lib/iris/fileformats/pp.py
```diff
@@ -1616,16 +1634,21 @@ def _interpret_fields(fields):
...
             yield field
 
 
-def _create_field_data(field, data_shape, land_mask):
+def _create_field_data(field, data_shape, with_landmask_field=None):
```
bjlittle: @pp-mo Change `with_landmask_field` to `land_mask_field` ... if you agree to the previous comment requests :wink:

lib/iris/fileformats/pp.py
```diff
@@ -1616,16 +1634,21 @@ def _interpret_fields(fields):
...
     """
     Modifies a field's ``_data`` attribute either by:
      * converting DeferredArrayBytes into a lazy array,
      * converting LoadedArrayBytes into an actual numpy array.
 
+    If 'with_landmask_field' is passed, it is another field :  The landmask
+    field's data is used as a template for this field's data, determining its
+    size, shape and the locations of valid (non-missing) datapoints.
```
bjlittle: @pp-mo Minor point - Could we slight reword to something along the lines of...
```
If 'land_mask_field' is passed (not None), then it is the associated landmask of the `field`. 
The landmask, which is also a field, is used as a template for the `field` to determine its 
size, shape and the locations of valid (non-missing) data-points.
```

lib/iris/fileformats/pp.py
```diff
@@ -1634,7 +1657,8 @@ def _create_field_data(field, data_shape, land_mask):
                                                  field.boundary_packing,
                                                  data_shape,
                                                  loaded_bytes.dtype,
-                                                 field.bmdi, land_mask)
+                                                 field.bmdi,
+                                                 with_landmask_field)
```
bjlittle: @pp-mo Change to `land_mask_field`...

lib/iris/fileformats/pp.py
```diff
@@ -1643,9 +1667,59 @@ def _create_field_data(field, data_shape, land_mask):
                             fname, position, n_bytes,
                             field.raw_lbpack,
                             field.boundary_packing,
-                            field.bmdi, land_mask)
+                            field.bmdi)
         block_shape = data_shape if 0 not in data_shape else (1, 1)
-        field.data = as_lazy_data(proxy, chunks=block_shape)
+        if with_landmask_field is None:
```
bjlittle: @pp-mo Change to `if land_mask_field is None`...

lib/iris/fileformats/pp.py
```diff
@@ -1643,9 +1667,59 @@ def _create_field_data(field, data_shape, land_mask):
...
+            def calc_array(mask, values):
+                # Note: "mask" is True at *valid* points, not missing ones.
+                # First ensure the mask array is boolean (not int).
+                mask = mask.astype(bool)
+                result = ma.masked_all(mask.shape, dtype=dtype)
+                n_values = np.sum(mask)
+                if n_values > 0:
+                    # Note: data field can have excess values, but not fewer.
+                    result[mask] = values[:n_values]
+                return result
```
bjlittle: @pp-mo Nice. The only part that _seems_ to be missing compare to the functionality of `_data_bytes_to_shaped_array` is setting the `fill_value` of the `result`. Is that still relevant? I'm thinking yes....?

If so, then is it the `proxy.mdi` that needs to also be passed into the delayed `calc_array` function?

bjlittle:
@pp-mo Over to you... :wink:

lib/iris/fileformats/pp.py
pp-mo: Sometime back when I was writing this, I had concluded that the fill-value was not respected by dask anyway, so this wasn't necessary. 
I *think* I found that the `da.stack` operation does not respect fill values, so in most cases we fail to preserve them on load, i.e. the data arrays come out with default fill-values instead of the BMDI value.

But the suggestions feels appropriate + I think does no harm, so I will put it back in.

pp-mo:
Thanks for taking this on @bjlittle .   Brave  :trophy: !
I've now tried to address all your comments, and a tiny bit more tidying besides.
Please re-review.

lib/iris/fileformats/pp.py
```diff
@@ -1581,51 +1588,68 @@ def _interpret_fields(fields):
...
+                               land_mask_field=land_mask_field)
             yield field
 
 
-def _create_field_data(field, data_shape, land_mask):
+def _create_field_data(field, data_shape, land_mask_field=None):
     """
     Modifies a field's ``_data`` attribute either by:
-     * converting DeferredArrayBytes into a lazy array,
+     * converting a 'deferred array bytes' tuple into a lazy array,
```
bjlittle: @pp-mo Yup :+1:

lib/iris/fileformats/pp.py
```diff
@@ -1581,51 +1588,68 @@ def _interpret_fields(fields):
...
     landmask_compressed_fields = []
     for field in fields:
         # Store the first reference to a land mask, and use this as the
         # definitive mask for future fields in this generator.
-        if land_mask is None and field.lbuser[6] == 1 and \
+        if land_mask_field is None and field.lbuser[6] == 1 and \
                 (field.lbuser[3] // 1000) == 0 and \
                 (field.lbuser[3] % 1000) == 30:
-            land_mask = field
+            land_mask_field = field
```
bjlittle: @pp-mo We're assuming that the `land_mask_field` is constant, right? i.e. multiple occurrences are simply duplicates?

bjlittle:
@pp-mo Nice one!
Only one last outstanding query, otherwise LGTM :+1:

lib/iris/fileformats/pp.py
bjlittle: No problem here, discussed to clarify. Always snaps the first land/sea mask.

pp-mo:
Magic, thanks @bjlittle !
Huge relief 😅

PR #3263, opened 2019-02-07
Fix problem with factories in iris.util.new_axis().

pp-mo:
It was found that adding an axis to a cube with an aux-factory had a subtle bug, which then causes errors _when you copy the result_ :

Example:
```
>>> cube
<iris 'Cube' of mass_fraction_of_ozone_in_air / (kg kg-1) (model_level_number: 63; grid_latitude: 182; grid_longitude: 146)>
>>> 
>>> # Presumably, something here is wrong ...
... cube2 = iutil.new_axis(cube, 'time')
>>> cube2
<iris 'Cube' of mass_fraction_of_ozone_in_air / (kg kg-1) (time: 1; model_level_number: 63; grid_latitude: 182; grid_longitude: 146)>
>>> 
>>> # ... but this is what actually fails :
... cube3 = cube2.copy()
  File "<stdin>", line 2, in <module>
  File "/opt/scitools/environments/experimental/2019_01_31-1/lib/python3.6/site-packages/iris/cube.py", line 2967, in copy
    cube = self._deepcopy(memo, data=data)
  File "/opt/scitools/environments/experimental/2019_01_31-1/lib/python3.6/site-packages/iris/cube.py", line 3003, in _deepcopy
    new_cube.add_aux_factory(factory.updated(coord_mapping))
  File "/opt/scitools/environments/experimental/2019_01_31-1/lib/python3.6/site-packages/iris/aux_factory.py", line 156, in updated
    coord = new_coord_mapping[id(coord)]
KeyError: 140318647616680
>>> 
```

The new addition to the test code should demonstrate this (fails when run against old code).

### DIFF
--- a/lib/iris/tests/unit/util/test_new_axis.py
+++ b/lib/iris/tests/unit/util/test_new_axis.py
@@ -25,11 +25,14 @@ import iris.tests as tests
 import iris.tests.stock as stock
 
 import copy
-from iris._lazy_data import as_lazy_data
 import numpy as np
 import unittest
 
+import six
+
 import iris
+from iris._lazy_data import as_lazy_data
+
 from iris.util import new_axis
 
 
@@ -135,6 +138,17 @@ class Test(tests.IrisTest):
 
         self.assertEqual(res, com)
         self._assert_cube_notis(res, cube)
+
+        # Check that factory dependencies are actual coords within the cube.
+        # Addresses a former bug : see XXXX
+        factory, = list(res.aux_factories)
+        deps = factory.dependencies
+        for dep_name, dep_coord in six.iteritems(deps):
+            coord_name = dep_coord.name()
+            msg = ('Factory dependency {!r} is a coord named {!r}, '
+                   'but it is *not* the coord of that name in the new cube.')
+            self.assertIs(dep_coord, res.coord(coord_name),
+                          msg.format(dep_name, coord_name))
 
     def test_lazy_data(self):
         cube = iris.cube.Cube(as_lazy_data(self.data))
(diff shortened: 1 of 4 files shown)


pp-mo:
What was actually wrong is :
 - the dependencies within a factory must be ***the actual coords in the cube*** for cube.copy() to work.. 
 - .. because otherwise the factory dependencies don't show up in the old-new 'coord_mapping' that the copy operation constructs
 - .. whereas, "new_axis" returns a cube whose factory has _its own independent copies_ of the original coords :  so that won't copy properly

lib/iris/tests/unit/util/test_new_axis.py
```diff
@@ -136,6 +139,17 @@ def test_maint_factory(self):
         self.assertEqual(res, com)
         self._assert_cube_notis(res, cube)
 
+        # Check that factory dependencies are actual coords within the cube.
+        # Addresses a former bug : see XXXX
```
lbdreyer: I assume you'll update this with the issue number?

pp-mo:
Thanks @lbdreyer !
I will fix that shortly.
But I want to just hold off a little on this one, as I have since been made aware of a possibly-related problem in some user code, which seems to have a similar error ...

lib/iris/tests/unit/util/test_new_axis.py
pp-mo: Ok

pp-mo:
Following offline discussion : 
Bugfix release whatsnew contributions should go in a separate block in the minor-version whatsnew documentation.
So here, moved note from a contributions file into a 'Bugs fixed in v2.2.1' section in the existing v2.2 whatsnew.
This parallels what we see in, for example, the [v1.7 whatsnew bugfixes](https://scitools.org.uk/iris/docs/latest/whatsnew/1.7.html#bugs-fixed)

lbdreyer:
Great, thank @pp-mo

PR #3264, opened 2019-02-07
Replace '__new__' with '__init__' in CubeList creation.

pp-mo:
"…  and typecheck content."

Raised in the course of #3238 

This is intended to combine with that, so 
 - (a) no whatsnew, though it is also applying that newly strict "only Cubes in CubeLists" approach
 - (b) doesn't anticipate other stuff there, like possibly [a better 'is a Cube' test](https://github.com/SciTools/iris/pull/3238#discussion_r254772464)

### DIFF
--- a/lib/iris/cube.py
+++ b/lib/iris/cube.py
@@ -165,19 +165,25 @@ class CubeList(list):
 
     """
 
-    def __new__(cls, list_of_cubes=None):
-        """Given a :class:`list` of cubes, return a CubeList instance."""
-        cube_list = list.__new__(cls, list_of_cubes)
+    def __init__(self, iterable_of_cubes=None):
+        """Given an iterable of cubes, return a CubeList instance."""
+        if iterable_of_cubes is None:
+            iterable_of_cubes = []
+        # Run the iterable, and capture as a list.
+        list_of_cubes = list(iterable_of_cubes)
 
-        # Check that all items in the incoming list are cubes. Note that this
-        # checking does not guarantee that a CubeList instance *always* has
-        # just cubes in its list as the append & __getitem__ methods have not
-        # been overridden.
-        if not all([isinstance(cube, Cube) for cube in cube_list]):
+        # Check that all items in the incoming list (if any) are cubes.
+        # Note that this checking does not (yet) guarantee that a CubeList
+        # instance *always* contains only cubes, as the append & __setitem__ methods
+        # (for example), have not  been overridden.
+        if not all(isinstance(cube, Cube) for cube in list_of_cubes):
             raise ValueError(
-                "All items in list_of_cubes must be Cube " "instances."
+                "CubeList create arguments are not all Cube "
+                "instances : {}".format(list_of_cubes)
             )
-        return cube_list
+
+        # Initialise "as a" list.
+        super().__init__(list_of_cubes)
 
     def __str__(self):
         """Runs short :meth:`Cube.summary` on every cube."""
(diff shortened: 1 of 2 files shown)


bjlittle:
@pp-mo Would you mind giving me an executive summary for the context of this PR? What's the purpose and intent?

Much obliged, thanks 😄

pp-mo:
> @bjlittle executive summary ... purpose and intent?

The existing `__new__` method is clearly intended to ensure that CubeLists can only contains Cubes when initially created.
But it is broken, and doesn't do that.

For reasons I don't quite grasp, the line `cube_list = list.__new__(cls, list_of_cubes)` always returns a new *empty* list object. So the subsequent checking code does nothing, ever.

I'm not even clear why this ever had `__new__` rather than an `__init__` method, as none of the usual reasons to use a `__new__` method apply :  
It is usually used to let object creation return an existing object, or a different type etc. :  anything other than producing a new object of the expected class.  
This code goes [right back to the earliest GitHub commit](https://github.com/SciTools/iris/blob/61ac271d425eb6037a8e9d3be2458f02f7de111d/lib/iris/cube.py#L84).  The choice was possibly related to the really old [inability to subclass builtin container types prior to python 2.2](https://docs.python.org/2.7/library/userdict.html?highlight=userdict#module-UserList).

The proposed replacement both actually removes the unnecessary use of `__new__`, and does the job right, in disallowing CubeList creation with non-cube arguments.
However, substantially more work is required to [prevent all other possible ways of getting a non-cube in a CubeList](https://github.com/SciTools/iris/pull/3238).

rcomer:
> This is intended to combine with [#3238]

I gave up on #3238 as it had stalled for some time.  It was something of a sledgehammer to crack a nut so not obviously the right way to go anyway.

Probably still worth ripping out `CubeList.__new__` in the interests of tidying up though?

pp-mo:
Rebased.
Hoping it goes ...

lib/iris/cube.py
```diff
@@ -165,19 +165,25 @@ class CubeList(list):
...
+        if iterable_of_cubes is None:
+            iterable_of_cubes = []
+        # Run the iterable, and capture as a list.
+        list_of_cubes = list(iterable_of_cubes)
+
+        # Check that all items in the incoming list (if any) are cubes.
+        # Note that this checking does not (yet) guarantee that a CubeList
+        # instance *always* contains only cubes, as the append & __setitem__ methods
+        # (for example), have not  been overridden.
+        if not all(isinstance(cube, Cube) for cube in list_of_cubes):
```
rcomer, line 179: There was a lot of discussion at #3238 about allowing cube-like objects into the cubelist.  Probably we need to come to a conclusion about that before merging this.  Or just remove all checks, since Iris has effectively been without them up to now.

🦆🦆🦆

trexfeathers: We've discussed this in a @SciTools/peloton meeting and the consensus is that this can proceed without #3238

rcomer: Sorry, I think my above comment was ambiguous: I meant to say we need a decision about whether we want to allow cube-like (duck typed) objects into cubelists.  This line would obviously disallow them to be included at  instantiation.  Within #3238 both @pp-mo and @pelson said they wanted to allow duck-cubes into cubelists.

pp-mo: Well we've moved on a bit now , especially with [UCube coming](https://github.com/SciTools-incubator/iris-ugrid/blob/master/iris_ugrid/ucube.py#L13) 
( though that isn't fully decided + agreed as a final approach yet ).

So I think I'd be happy to broaden this to a duck test.
But we need to think about what.

The previously suggested `hasattr(cube, 'add_aux_coord')` is a bit specific, but I guess only one test is really quick.
Or we could aim to accept an array-like object with cube elements and phenomenon identity.
Maybe something like `hasattr(cube, 'shape') and hasattr(cube, 'standard_name')` ?

PR #3267, opened 2019-02-08
Avoid numpy FutureWarning about indexing with sequence.

pp-mo:
This has been cropping up in our tests, [e.g. here](https://travis-ci.org/SciTools/iris/jobs/489641793#L1634).

It also reportedly affects some uses of interpolation (? though not seen in tests ?).

PR #3273, opened 2019-02-22
Mergeback master to 2.2.x

lbdreyer:
Two PRs were incorrectly targeted and merged into master:
#3265 Test with newest cf-units: no longer called cf_units.
#3267 Avoid numpy FutureWarning about indexing with sequence

This cherrypicks those two commits and adds them to the `v2.2.x` branch


lbdreyer:
Closing: This approach may cause complications later on when trying to mergeback `v2.2.x` branch into `master`.

Instead, the plan is to revert the last two PRs and retarget them at `v2.2.2x`

PR #3275, opened 2019-02-22
Revert "Test with newest cf-units: no longer called cf_units."

pp-mo:
Reverts SciTools/iris#3265

Following #3274 (thanks @lbdreyer !)
Likewise part of routing these changes via v2.2.x branch instead of direct to master ...


lbdreyer:
Thanks @pp-mo !

PR #3281, opened 2019-02-25
Adds lazy version of aggregated_by routine to Cube.

### LINKED ISSUE #3280: Add lazy `aggregated_by`
At the moment, `collapsed` is the only lazy statistics method. It would be useful to also have `aggregated_by` in a lazy version, i.e. allow for the use of lazy aggregators.

This touches on #2418, and #2928.

zklaus:
Closes #3280

Signed-off-by: Klaus Zimmermann <klaus.zimmermann@smhi.se>

### DIFF
--- a/lib/iris/tests/unit/cube/test_Cube.py
+++ b/lib/iris/tests/unit/cube/test_Cube.py
@@ -1,4 +1,4 @@
-# (C) British Crown Copyright 2013 - 2018, Met Office
+# (C) British Crown Copyright 2013 - 2019, Met Office
 #
 # This file is part of Iris.
 #
--- a/lib/iris/tests/unit/cube/test_Cube.py
+++ b/lib/iris/tests/unit/cube/test_Cube.py
@@ -550,6 +550,93 @@ class Test_aggregated_by(tests.IrisTest):
                          AuxCoord(['a|a', 'a'], long_name='bar'))
 
 
+class Test_aggregated_by__lazy(tests.IrisTest):
+    def setUp(self):
+        self.data = np.arange(44).reshape(4, 11)
+        self.lazydata = as_lazy_data(self.data)
+        self.cube = Cube(self.lazydata)
+
+        val_coord = AuxCoord([0, 0, 0, 1, 1, 2, 0, 0, 2, 0, 1],
+                             long_name="val")
+        label_coord = AuxCoord(['alpha', 'alpha', 'beta',
+                                'beta', 'alpha', 'gamma',
+                                'alpha', 'alpha', 'alpha',
+                                'gamma', 'beta'],
+                               long_name='label', units='no_unit')
+        simple_agg_coord = AuxCoord([1, 1, 2, 2], long_name='simple_agg')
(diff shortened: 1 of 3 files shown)


pp-mo:
I haven't had time to check code in depth yet, 
but the idea is a really worthwhile + exciting ! :+1: :rocket: 
Some tests of new functionality will be wanted.

P.S. We just fixed the CI header tests, so sorry about that !

zklaus:
Thanks! Yes, I am going through the checklist now and will add tests and whatsnew and whatnot :).

In fact, this is the only reason why this is marked as WIP, the code itself is considered finished and of course works for me.

Maybe to ease reading it: I completely kept the code for the non-lazy version, just put it in the else branch of an if statement guarding the lazy version in the exact same way as  in the collapsed method.

A more aggressive version is possible were also the non-lazy part is implemented in the same way as the lazy part with only `lazy_aggregate` replaced by `aggregate` and `da.stack` replaced by `np.stack`.

zklaus:
Ok, @bjlittle, @pp-mo, I think this is ready.

lib/iris/tests/unit/cube/test_Cube.py
```diff
@@ -549,6 +550,93 @@ def test_single_string_aggregation(self):
...
+                                'beta', 'alpha', 'gamma',
+                                'alpha', 'alpha', 'alpha',
+                                'gamma', 'beta'],
+                               long_name='label', units='no_unit')
+        simple_agg_coord = AuxCoord([1, 1, 2, 2], long_name='simple_agg')
+        spanning_coord = AuxCoord(np.arange(44).reshape(4, 11),
+                                  long_name='spanning')
+        spanning_label_coord = AuxCoord(
+            np.arange(1, 441, 10).reshape(4, 11).astype(str),
+            long_name='span_label', units='no_unit')
```
lbdreyer: The spanning and span_label 2D coordinates are added to the cube but there isn't any testing of whether they are handled correctly during aggregation.
You could add an extra check that their points (and bounds for the numerical coordinate) are as expected.

zklaus: Thanks @lbdreyer! I've been mulling this over a little bit and I think I'd rather remove the spanning stuff from the setup. The reason is that this part tests the part of *what* to aggregate, whereas this pr only changes some part of *how* to perform it. The alternative would be a verbatim copy of the non lazy test, but I think it really wouldn't add much. Would that be ok?

pp-mo: Hi @zklaus sorry for delay.
I think it's quite good enough to just remove the spanning coords, and then I think we can merge this as it is.
**Can you manage that within about the next 2 days, so it can go into Iris 2.3 ?**
Thanks.

zklaus: Yes

zklaus:
@pp-mo, @lbdreyer thanks for your help. I removed the spanning coordinates.

pp-mo:
@zklaus great !
Thanks for this really neat improvement.

Issue #3282, opened 2019-02-25
NaNs in coords of regridded cubes


pp-mo:
If you regrid a cube with an aux-factory, "ordinary" aux-coords are discarded, but factory dependencies are regridded.  If it involves any extrapolation, those points get NaNs.

Although it seems this was a definite past decision, I'm now questioning whether this is correct ...

Example:
```
>>> import iris
>>> import iris.tests.stock as istk
>>> 
>>> cube = istk.realistic_4d()[0, 0, [8, 30], 50:53]
>>> for axis in ('x', 'y'):
...   cube.coord(axis=axis).bounds = None
... 
>>> print(cube.coord(axis='y').points)
[-0.1206 -0.1008]
>>> 
>>> gridcube = cube.copy()
>>> gridcube.coord(axis='y').points = [-0.125, -0.115]
>>> result = cube.regrid(gridcube, iris.analysis.Linear())
>>> 
>>> print(cube.coord('surface_altitude').points)
[[205.14995 204.45988 202.7499 ]
 [321.84033 334.86038 351.05035]]
>>> 
>>> print(result.coord('surface_altitude').points)
[[      nan       nan       nan]
 [238.15329 241.34084 244.69345]]
>>> 
```

pp-mo:
**Implementation Notes:**
The key code in question is about here : https://github.com/SciTools/iris/blob/v2.2.0/lib/iris/analysis/_regrid.py#L541
The question is, why does that explicitly set `extrapolation_mode` to `'nan'`, rather than using `self._extrapolation_mode`, as does the 'main' regrid call which handles the cube data.

The current code seems to emerge from [this very old discussion](https://github.com/SciTools/iris/pull/1160#issuecomment-45096290)
It seems to have been a crude way to avoid masked points in (aux)-coords, but we don't actually have that restriction any more, and I'm not sure that CF does either.
In any case, **_I don't follow why it is "better" to ban extrapolation in this case, even when the main operation allows it_** (default mode='linear' --> extrapolation_mode='extrapolate').
Maybe that was a mistake.  It does seem perverse in this case.

pp-mo:
> Although it seems this was a definite past decision, I'm now questioning whether this is correct ...

Your take @bjlittle ?

mo-g:
CF not only explicitly allows, but requires missing values in aux-coords, and the wording has certainly changed...

[1.6](http://cfconventions.org/cf-conventions/v1.6.0/cf-conventions.html#coordinate-types):

[...]

PR #3283, opened 2019-02-26
Coord nan equal

pp-mo:
Following the discovery #3282, 
found that saving regridded cubes would fail because when a regridded dependency coordinate (orography in this case), coord contains NaNs, then "coord != coord".

Changes here implement a nan-tolerant comparison, so a coord containing NaNs **_does_** compare equal to an identical one (or itself). 

Example of resulting save error :
```
>>> import iris
>>> import iris.tests.stock as istk
>>> cube = istk.realistic_4d()
>>> orog = cube.coord('surface_altitude').points
>>> orog[10,10] = np.nan
>>> cube.coord('surface_altitude').points = orog
>>> iris.save(cube, 'tmp.nc')
  File "/home/h05/itpp/git/iris/iris_main/lib/iris/io/__init__.py", line 408, in save
    saver(source, target, **kwargs)
  File "/home/h05/itpp/git/iris/iris_main/lib/iris/fileformats/netcdf.py", line 2352, in save
    fill_value=fill_value)
  File "/home/h05/itpp/git/iris/iris_main/lib/iris/fileformats/netcdf.py", line 1001, in write
    self._add_aux_factories(cube, cf_var_cube, dimension_names)
  File "/home/h05/itpp/git/iris/iris_main/lib/iris/fileformats/netcdf.py", line 1248, in _add_aux_factories
    key, coord in six.iteritems(factory.dependencies)}
  File "/home/h05/itpp/git/iris/iris_main/lib/iris/fileformats/netcdf.py", line 1248, in <dictcomp>
    key, coord in six.iteritems(factory.dependencies)}
  File "/home/h05/itpp/git/iris/iris_main/lib/iris/fileformats/netcdf.py", line 320, in name
    raise KeyError(msg)
KeyError: "Coordinate is not mapped, AuxCoord(array([[413.93686, 417.8168 , ...
 . . .
>>> 
```

### DIFF
--- a/docs/iris/src/whatsnew/2.2.rst
+++ b/docs/iris/src/whatsnew/2.2.rst
@@ -66,6 +66,9 @@ Iris 2.2 Features
   discontiguous points in coordinates can be explicitly masked
   using another new feature :func:`iris.util.mask_cube`.
 
+* :func:`iris.util.array_equal` now has a 'withnans' keyword, which provides
+  a NaN-tolerant array comparison.
+
 
 Iris 2.2 Dependency updates
 =============================
@@ -95,9 +98,15 @@ Bugs fixed in v2.2.1
 
 * Iris can now correctly unpack a column of header objects when saving a
   pandas DataFrame to a cube.
-
+  
 * fixed a bug in :meth:`iris.util.new_axis` : copying the resulting cube
   resulted in an exception, if it contained an aux-factory.
+
+* :class:`iris.coords.AuxCoord`_'s can now test as 'equal' even when the points
+  or bounds arrays contain NaN values, if values are the same at all points.
+  Previously this would fail, as conventionally "NaN != NaN" in normal
+  floating-point arithmetic.
+
 
 
 Documentation Changes
--- a/docs/iris/src/whatsnew/contributions_2.3.0/newfeature_2019-Feb-25_compare_nan.txt
+++ b/docs/iris/src/whatsnew/contributions_2.3.0/newfeature_2019-Feb-25_compare_nan.txt
@@ -1,2 +1,0 @@
-* :func:`iris.util.array_equal` now has a 'withnans' keyword, which provides
-  a NaN-tolerant array comparison.
--- a/docs/iris/src/whatsnew/contributions_2.3.0/newfeature_2019-Feb-25_compare_nan.txt
+++ b/docs/iris/src/whatsnew/contributions_2.3.0/newfeature_2019-Feb-25_compare_nan.txt
@@ -1,0 +1,2 @@
+* :func:`iris.util.array_equal` now has a 'withnans' keyword, which provides
+  a NaN-tolerant array comparison.
--- a/lib/iris/tests/unit/util/test_array_equal.py
+++ b/lib/iris/tests/unit/util/test_array_equal.py
@@ -1,4 +1,4 @@
-# (C) British Crown Copyright 2014 - 2015, Met Office
+# (C) British Crown Copyright 2014 - 2019, Met Office
 #
 # This file is part of Iris.
 #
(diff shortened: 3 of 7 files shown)


docs/iris/src/whatsnew/contributions_2.3.0/newfeature_2019-Feb-25_compare_nan.txt
```diff
@@ -0,0 +1,2 @@
+* :func:`iris.util.array_equal` now has a 'withnans' keyword, which provides
+  a NaN-tolerant array comparison.
```
lbdreyer: You've targetted this at the `v2.2.x` branch but included a whats new entry to the `2.3.0` whats new?

pp-mo: D'oh!

pp-mo: Addressed .. ?

pp-mo:
NOTE:

This might call into question whether cube-comparison should also be made nan-tolerant ??

It's harder, as 
  * cube comparison involves [a toleranced array difference (though really not a very solid one)](https://github.com/SciTools/iris/blob/v2.2.0/lib/iris/cube.py#L3024)
  * We might further extend `array_equal` to do toleranced comparison, but that seems over-extension

I think it's also _less_ likely to cause internal Iris errors, like this one
 \- because Iris doesn't deal in collections of cubes very much (but collections of Coords, quite a lot)

So, **I'm choosing to ignore this for now**.

lib/iris/tests/unit/util/test_array_equal.py
```diff
@@ -1,4 +1,4 @@
-# (C) British Crown Copyright 2014 - 2015, Met Office
+# (C) British Crown Copyright 2014 - 2019, Met Office
```
lbdreyer: Did you want to follow [your suggested approach](https://github.com/SciTools/iris/pull/3272#issuecomment-468235708) and remove the year

docs/iris/src/whatsnew/2.2.rst
```diff
@@ -66,6 +66,9 @@ Iris 2.2 Features
   discontiguous points in coordinates can be explicitly masked
   using another new feature :func:`iris.util.mask_cube`.
 
+* :func:`iris.util.array_equal` now has a 'withnans' keyword, which provides
+  a NaN-tolerant array comparison.
+
```
lbdreyer: I'm torn with this.
Part of me wants it in under a sub-heading, similar to the ["bugs fixed in 1.7.3" sub headings](https://scitools.org.uk/iris/docs/latest/whatsnew/1.7.html#bugs-fixed-in-v1-7-3) in the 1.7 what's new but I'm not sure...

lib/iris/tests/unit/util/test_array_equal.py
lbdreyer: The template for the new file headers is here:
https://github.com/SciTools/scitools.org.uk/blob/master/documents/scitools_licensing.md#source-preamble

pp-mo: I think I'd rather do it separately.

My reason is, I just realised, we are going to have to modify the licence-header check in `test_coding_standards.py` **_first_** ... see #3285

docs/iris/src/whatsnew/2.2.rst
pp-mo: Hmm. It's awkward because bugfix releases aren't _supposed_ to contain new features !

We could treat the new keyword as a private matter, and announce the new functionality with enormous fanfare in the nest minor release ??
I think that means putting the whatsnew contribution back into a textfile in a contributions_2.3.0 directory.
The actual code changes (+docstring) can stay though,

Would you favour that solution @lbdreyer  ?

lib/iris/tests/unit/util/test_array_equal.py
lbdreyer: Yep, that makes sense :+1:

docs/iris/src/whatsnew/2.2.rst
lbdreyer: I believe the outcome of the offline discussion was to keep what you have already done.
Admittedly the reasons for that decision are a little fuzzy we discussed this over a week ago, but I think I personally wanted to not keep it private, but not add confusion by adding a "features added in 2.2.1" section

So basically, I'm happy with this change as it is.

lbdreyer:
@pp-mo It looks like this PR is ready to go!

Do you mind rebasing and then it can be merged?

pp-mo:
Rebased !

lbdreyer:
I am going to restarted _all_ the travis ci jobs now that #3311 that fixes the tests is in.

lbdreyer:
Tests all pass :tada: 
I'm going to merge this in (finally!)

rcomer:
@pp-mo 
> cube comparison involves a toleranced array difference (though really not a very solid one)

Not sure if it's relevant to your point, but the line in question will change if/when #3250 is merged.

PR #3289, opened 2019-03-06
made docstrings for contiguity masking more accurate and clear

corinnebosley:
Tweaked wording and naming convention of contiguity checking and masking so that it's clearer what's going on and what functions you can use.

Unfortunately I think this is a breaking change so can't go in until next major release.


pp-mo:
I don't get this.
I thought the point was that 'mask_cube' replaced 'mask_discontiguities' precisely *because* it no longer has anything specifically to do with discontiguity checking.  It is now just an apply-mask-array-to-a-cube routine.
So it wasn't just a rename, and I think renaming it back is not appropriate.

Instead, I reckon what is wrong with the existing is just [the mis-quoted name in the find_discontiguities docstring](https://github.com/SciTools/iris/blob/45a18abc1b87a43a1373212ab4205713fe92277c/lib/iris/util.py#L1614)
Fix that + it's all good, no ??

bjlittle:
@pp-mo Would you care to take this on?

pp-mo:
> what is wrong with the existing is just the mis-quoted name in the find_discontiguities docstring
> Fix that + it's all good, no ??

Re-reading, there are quite a few more things wrong...
Mostly, the routines don't handle data with "normal" 1D x and y coords, and/or don't explain that they don't.
There is also real confusion over the use of atol and rtol here : https://github.com/SciTools/iris/blob/v2.2.1/lib/iris/plot.py#L311
So, there is no rtol for 1-D comparisons.

Likewise, [`utils.find_discontiguities`](https://github.com/SciTools/iris/blob/v2.2.1/lib/iris/util.py#L1612) seems to ***only*** work with 2D coords, but this is not checked or explained.

bjlittle:
@corinnebosley and @pp-mo Thanks, but I'm closing this PR. It's gone stale and there's been no progress for seven months. If someone cares enough, then please breath life into this PR and we can gladly take it from there :+1:

Issue #3081, opened 2018-07-05
Cannot remove cell measure referenced by name only


DPeterK:
If you call `cube.remove_cell_measure` and specify the cell measure with its name string only, the cell measure is not removed:

```python
my_cube.remove_cell_measure('cell_area')
print(my_cube)
```
```
sea_surface_temperature / (degree_C) (-- : 1207; -- : 1442)
     Auxiliary coordinates:
          latitude                       x          x
          longitude                      x          x
     Cell Measures:
          cell_area                      x          x
     Scalar coordinates:
          time: 2003-05-16 12:00:00, bound=(2003-05-01 00:00:00, 2003-06-01 00:00:00)
     Attributes:
          ...
     Cell methods:
          mean: time (1350 s)
```

If, however, I pass the actual cell measure the removal does work:
```python
my_cube.remove_cell_measure(cube.cell_measures()[0])
print(my_cube)
```
```
sea_surface_temperature / (degree_C) (-- : 1207; -- : 1442)
     Auxiliary coordinates:
          latitude                       x          x
          longitude                      x          x
     Scalar coordinates:
          time: 2003-05-16 12:00:00, bound=(2003-05-01 00:00:00, 2003-06-01 00:00:00)
     Attributes:
          ...
     Cell methods:
          mean: time (1350 s)
```

bjlittle:
Closed by #3295

PR #3295, opened 2019-03-08
Removal of a cube's cell measure by name (#3081)

alastair-gemmell:
Although labelled as a bug I don't think #3081 is a bug strictly speaking. As far as I can tell its behaviour is as expected from the reference docs at https://scitools.org.uk/iris/docs/latest/iris/iris/cube.html#iris.cube.Cube.remove_cell_measure.  In as much as the argument to the method is documented as having to be a CellMeasure instance (not a name of a cell measure).

Having said that it could be a bit misleading as other public Cube methods such as Cube.cell_measure take either a CellMeasure object OR the name of a cell measure (https://scitools.org.uk/iris/docs/latest/iris/iris/cube.html#iris.cube.Cube.cell_measure), and currently if you attempt to pass in the name of a cell measure to be removed it isn't removed, silently

So although maybe not a bug in the strict sense it seemed a good candidate for a minor new feature to accept a name of a cell measure when calling the Cube.remove_cell_measure method

Especially as I'm a newbie contributor let me know what you think, and if I've forgotten anything here (I've updated the docstring, added new tests and added a 'whats new' entry but I may have missed something?)

### DIFF
--- a/lib/iris/cube.py
+++ b/lib/iris/cube.py
@@ -1117,14 +1117,14 @@ bound=(1994-12-01 00:00:00, 1998-12-01 00:00:00)
         for factory in self.aux_factories:
             factory.update(coord)
 
-    def remove_cell_measure(self, name_or_cell_measure):
+    def remove_cell_measure(self, cell_measure):
         """
         Removes a cell measure from the cube.
 
         Args:
 
-        * name_or_cell_measure
-            The cell measure to remove from the cube. As either
+        * cell_measure (string or cell_measure)
+            The (name of the) cell measure to remove from the cube. As either
 
             (a) a :attr:`standard_name`, :attr:`long_name`, or
             :attr:`var_name`. Defaults to value of `default`
@@ -1145,7 +1145,7 @@ bound=(1994-12-01 00:00:00, 1998-12-01 00:00:00)
             :meth:`Cube.add_cell_measure()<iris.cube.Cube.add_cell_measure>`
 
         """
-        cell_measure = self.cell_measure(name_or_cell_measure)
+        cell_measure = self.cell_measure(cell_measure)
 
         self._cell_measures_and_dims = [[cell_measure_, dim] for cell_measure_,
                                         dim in self._cell_measures_and_dims
--- a/lib/iris/cube.py
+++ b/lib/iris/cube.py
@@ -1,4 +1,4 @@
-# (C) British Crown Copyright 2010 - 2018, Met Office
+# (C) British Crown Copyright 2010 - 2019, Met Office
 #
 # This file is part of Iris.
 #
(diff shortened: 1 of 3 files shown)


pp-mo:
Welcome to Iris-devs !!!

I think this *is* a bug :  The docstring says you can pass a name, but you can't. 
I think all it needs is a "cell_measure = self.cell_measure(cell_measure)" at the start of the function.
It should also be worth creating a bug-style 'whatsnew' entry
(  [as described here](https://scitools.org.uk/iris/docs/latest/developers_guide/documenting/whats_new_contributions.html), 
  but as usual easier to copy existing stuff ! )

Send us a PR, if you fancy having a go ??

mo-g:
Tests look ok at first glance, either way. Most important part of any PR. :wink:

lib/iris/cube.py
```diff
@@ -1117,19 +1117,36 @@ def remove_coord(self, coord):
         for factory in self.aux_factories:
             factory.update(coord)
 
-    def remove_cell_measure(self, cell_measure):
+    def remove_cell_measure(self, name_or_cell_measure):
```
bjlittle: @alastair-gemmell I think that keeping the `cell_measure` positional argument "as is" is good enough, and is in keeping with other `remove_` methods that allow a mixture of different types of inputs to match on i.e., see `iris.cube.Cube.remove_coord` for example.

lib/iris/cube.py
```diff
@@ -1117,19 +1117,36 @@ def remove_coord(self, coord):
...
-            The CellMeasure to remove from the cube.
+        * name_or_cell_measure
+            The cell measure to remove from the cube. As either
 
-        See also
-        :meth:`Cube.add_cell_measure()<iris.cube.Cube.add_cell_measure>`
+            (a) a :attr:`standard_name`, :attr:`long_name`, or
+            :attr:`var_name`. Defaults to value of `default`
+            (which itself defaults to `unknown`) as defined in
+            :class:`iris._cube_coord_common.CFVariableMixin`.
```
bjlittle: @alastair-gemmell A lovely piece of clarification, which explains that `cell_measure` can be an actual `CellMeasure` instance or an appropriate string name :smile:

lib/iris/cube.py
alastair-gemmell: Thanks @bjlittle .  I've made this change now and the tests have passed.

It's good to have this consistency with other `remove` methods, although there is already inconsistency because the `cell_measures` and `cell_measure` methods do have the argument listed as 'name_or_cell_measure' (which is what I based my original change on!)

bjlittle:
Awesome work, thanks @alastair-gemmell :+1:

PR #3296, opened 2019-03-20
Xarray arithmetic POC

corinnebosley:
DO NOT MERGE!

This is a proof of concept PR to investigate the possibility of utilising xarray's laid-back attitude with regard to cube arithmetic in certain cases.

This is a simple and non-breaking potential solution to cube arithmetic problems described here: https://github.com/SciTools/iris/issues/2599 with the use of long-hand maths calls and a 'strict=False' flag.  

For example, instead of this:

a = cube
b = other
c = cube - other

You would say this:

a = cube
b = other
iris.analysis.maths.subtract(cube, other, strict=False)

This method would be used in cases where the arithmetic you wanted to perform was non-standard (according to iris and its checks) but still either logical or necessary.

### DIFF
--- a/lib/iris/analysis/maths.py
+++ b/lib/iris/analysis/maths.py
@@ -226,7 +226,7 @@ def _assert_matching_units(cube, other, operation_name):
         raise iris.exceptions.NotYetImplementedError(msg)
 
 
-def add(cube, other, dim=None, in_place=False):
+def add(cube, other, dim=None, in_place=False, strict=True):
     """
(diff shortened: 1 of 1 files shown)


bjlittle:
@corinnebosley If this is a POC then why do this here? Isn't it better to do this in your own fork with travis-ci turned on?

corinnebosley:
I didn't realise it wasn't, or where I am actually.  I'll sort it tomorrow when my brain works, a bit.

PR #3300, opened 2019-03-29
UserGuide: remove note on partial collapse of auxcoords.

rcomer:
Since #3028, this line in the user guide is no longer true.


rcomer:
#3319 fixed this in the v2.2.x branch, so think this is now redundant.

Issue #3298, opened 2019-03-28
Partial collapse of multidimensional auxcoords with bounds


rcomer:
Since #3028, we can partially collapse a multidimesional auxcoord, which is great.  However, it goes wrong if that auxcoord has bounds.  Example:

```python
import iris
import numpy as np

print(iris.__version__)

cube = iris.cube.Cube(np.zeros((2, 3)))

coord1 = iris.coords.DimCoord(range(2), long_name='foo')
coord2 = iris.coords.DimCoord(range(3), long_name='bar')

spanning_bounds = np.stack([np.arange(6), np.arange(1, 7)], axis=-1)
spanning_points = np.mean(spanning_bounds, axis=-1)
spanning_coord = iris.coords.AuxCoord(
    spanning_points.reshape(2, 3), bounds=spanning_bounds.reshape(2, 3, 2),
    long_name='spanning')

for i, coord in enumerate([coord1, coord2]):
    cube.add_dim_coord(coord, i)

cube.add_aux_coord(spanning_coord, (0, 1))

print(cube.collapsed('foo', iris.analysis.MEAN))
```

Output:
```
2.2.0
[warnings about metadata]
Traceback (most recent call last):
  File "2d_coord_issue.py", line 24, in <module>
    print(cube.collapsed('foo', iris.analysis.MEAN))
  File "[site-packages-path]/iris/cube.py", line 3217, in collapsed
    collapsed_cube.replace_coord(coord.collapsed(local_dims))
  File "[site-packages-path]/iris/cube.py", line 1149, in replace_coord
    self.add_aux_coord(new_coord, dims)
  File "[site-packages-path]/iris/cube.py", line 949, in add_aux_coord
    self._add_unique_aux_coord(coord, data_dims)
  File "[site-packages-path]/iris/cube.py", line 981, in _add_unique_aux_coord
    data_dims = self._check_multi_dim_metadata(coord, data_dims)
  File "[site-packages-path]/iris/cube.py", line 973, in _check_multi_dim_metadata
    metadata.shape[i]))
ValueError: Unequal lengths. Cube dimension 0 => 3; metadata 'spanning' dimension 0 => 2.
```

rcomer:
Fixed at #3302.

PR #3302, opened 2019-04-03
Partial Collapse of Multidimensional Auxcoords with Bounds

rcomer:
Proposed fix for #3298.

### DIFF
--- a/lib/iris/coords.py
+++ b/lib/iris/coords.py
@@ -1356,16 +1356,22 @@ class Coord(six.with_metaclass(ABCMeta, CFVariableMixin)):
                     'Metadata may not be fully descriptive for {!r}.'
                 warnings.warn(msg.format(self.name()))
 
-            # Determine the array library for stacking
-            al = da if self.has_bounds() \
-                and _lazy.is_lazy_data(self.core_bounds()) else np
+            if self.has_bounds():
+                item = self.core_bounds()
+                if dims_to_collapse is not None:
+                    # Express main dims_to_collapse as non-negative integers
+                    # and add the last (bounds specific) dimension.
+                    dims_to_collapse = tuple(
+                        dim % self.ndim for dim in dims_to_collapse) + (-1,)
+            else:
+                item = self.core_points()
 
-            item = al.concatenate(self.core_bounds()) if self.has_bounds() \
-                else self.core_points()
+            # Determine the array library for stacking
+            al = da if _lazy.is_lazy_data(item) else np
 
             # Calculate the bounds and points along the right dims
             bounds = al.stack([item.min(axis=dims_to_collapse),
-                               item.max(axis=dims_to_collapse)]).T
+                               item.max(axis=dims_to_collapse)], axis=-1)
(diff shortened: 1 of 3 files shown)


lib/iris/coords.py
```diff
@@ -1356,16 +1356,22 @@ def serialize(x):
...
-            # Determine the array library for stacking
-            al = da if self.has_bounds() \
-                and _lazy.is_lazy_data(self.core_bounds()) else np
+            if self.has_bounds():
+                item = self.core_bounds()
+                if dims_to_collapse is not None:
+                    # Express main dims_to_collapse as non-negative integers
+                    # and add the last (bounds specific) dimension.
+                    dims_to_collapse = tuple(
+                        dim % self.ndim for dim in dims_to_collapse) + (-1,)
```
lbdreyer: Nice!

lbdreyer: Nice!

lib/iris/tests/unit/coords/test_Coord.py
```diff
@@ -303,6 +303,30 @@ def test_numeric_nd(self):
...
+                                                                [78, 112]]))
+
+        # ... and the other
+        collapsed_coord = coord.collapsed(0)
+        self.assertArrayEqual(
+            collapsed_coord.points, np.array([40, 50, 60, 70]))
+        self.assertArrayEqual(collapsed_coord.bounds, np.array([[-2, 82],
+                                                                [8,  92],
+                                                                [18, 102],
+                                                                [28, 112]]))
```
lbdreyer: Could you split this into 3 tests? i.e. one test for each collapse operation.

My concern with having them all in the one test is that if the first one fails we won't know the result of the other two collapses until we get the first one working.
I realise `test_numeric_nd()` that already exists in this file does the same thing as you have done here, but I do think it's better practice to split the tests up

lbdreyer: And perhaps add a test where the dimensions given are negative to test that handling is correct:
i.e. 
```
collapsed_coord = coord.collapsed(-1)
```

rcomer: Will do, and thanks for explaining why - it makes sense.  Should I do the same with the lazy equivalent below?

lbdreyer: > Should I do the same with the lazy equivalent below?

Yes please!

rcomer: Done.  Also moved the dask import to the top of the module.  This is consistent with other test modules that use dask.

Since I now have `self.setupTestArrays((3, 4))` in 8 of the tests, I wondered about adding a setUp method to contain that.  But there are a lot of tests on the class that don't need these arrays so not really sure what the better approach is.

rcomer:
This change also (at least partly) addresses #3363.  I've added an extra test at #3364 that should pass once this branch is merged.

lbdreyer:
All looks good! Thanks @rcomer !

I'm going to merge this in. 
Note the travis tests will fail due to the numpy issues that are fixed in #3366 but I don't want that to hold this PR up

PR #3308, opened 2019-05-03
Imagehash uris check uses file-get instead of github api

pp-mo:
**WIP: do not merge**

Refers to pp=-mo/test-iris-imagehash
And anyway may incorporate this into #3309 

This requires https://github.com/SciTools/test-iris-imagehash/pull/23


pp-mo:
Replacing this with a rebased PR against the v2.2.x branch : #3311
It also needs a licence header fix (done there)

PR #3297, opened 2019-03-22
Fix for small plot changes with freetype 2.10.

pp-mo:
HISTORY NOTE 
Used to contain a lot of trial code attempting to make 'test_image_json' work.
I've taken all that out again, as that solution now lies in #3308 

So this can now be merged ok.
https://github.com/SciTools/test-iris-imagehash/pull/22 relates, but that is now merged.

PR #3309, opened 2019-05-03
Image fixes 2

pp-mo:
**WIP: do not merge**
Because it still references the test branch in pp-mo/imagehash instead of the main repo.

Should replaces #3297 , as the history there has got very complicated.


pp-mo:
Reason for close : we can and should deal with these 2 things separately

PR #3311, opened 2019-05-09
Imagehash listfile for fixbranch

pp-mo:
An almost complete rewrite of `iris.tests.test_image-json`.
Instead of using the GitHub API to interrogate the test-iris-imagehash repo, we now just fetch an 'image listings file' which is basically the text of an `ls *.png`.
This is because the (unauthenticated) GitHub API access was proving very unreliable.

Note : relies on SciTools/test-iris-imagehash#23 , but that is already merged.

### DIFF
--- a/lib/iris/tests/test_image_json.py
+++ b/lib/iris/tests/test_image_json.py
@@ -39,7 +39,7 @@ class TestImageFile(tests.IrisTest):
             '/gh-pages/v4_files_listing.txt')
         req = requests.get(listingfile_uri)
         if req.status_code != 200:
-            raise ValueError('Get failed on image listings file: {}'.format(
+            raise ValueError('GET failed on image listings file: {}'.format(
                 listingfile_uri))
 
         listings_text = req.content.decode('utf-8')
--- a/lib/iris/tests/test_image_json.py
+++ b/lib/iris/tests/test_image_json.py
@@ -29,44 +29,45 @@ import itertools
 import json
 import os
 import requests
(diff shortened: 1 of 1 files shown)


pp-mo:
I think this is ready to go now.
However, there is still an outstanding problem with Sphinx doctests, addressed by #3310.
So we probably want that in first, and maybe rebase onto it.

lbdreyer:
I am going to respin the tests now that #3310 is merged

lib/iris/tests/test_image_json.py
```diff
@@ -29,55 +29,45 @@
...
-            remaining = rates.get('rate', {})
-            ghapi_remaining = remaining.get('remaining')
-        else:
-            ghapi_remaining = 0
+        listingfile_uri = (
+            'https://raw.githubusercontent.com/SciTools/test-iris-imagehash'
+            '/gh-pages/v4_files_listing.txt')
+        req = requests.get(listingfile_uri)
+        if req.status_code != 200:
+            raise ValueError('Get failed on image listings file: {}'.format(
```
lbdreyer: You could capitalise `GET` to make it clear it is a http request. Otherwise, the error reads a bit funnily out of context

pp-mo: Good idea, I see what you mean :grin:

lbdreyer:
Thanks for all your work on this @pp-mo!

It will be great to get this test running!

PR #3316, opened 2019-05-28
v2.2.x branch mergeback

lbdreyer:
There has been a fair amount of additions to the 2.2.x branch that should be merged back into master (especially since we just cut the 2.2.1 release).

To create this PR I did the following commands:

```$ git checkout upstream/v2.2.x -b 221_mergeback_rebase```
Then I did an interactive release to select the commits that were missing 
```$ git rebase -i --onto upstream/master HEAD~12```
Which were all the commits since `c3048cbf Fix for numpy 1v16. (#3257)` as this and `b0d5512b BLD: Pinning cftime to 1.0.1 (#3247)` were included in the last mergeback.

PR #3318, opened 2019-05-29
Update copyright years in license headers

lbdreyer:
The license header check doesn't run in this branch so we missed a few copyright years that need updating.

This is needed so before we can do the mergeback of this branch (see #3316 )

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -125,16 +125,19 @@ install:
     fi
 
 script:
-  - if [[ $TEST_TARGET == 'default' ]]; then
-      python -m iris.tests.runner --default-tests --system-tests --coding-tests --print-failed-images --num-processors=3;
-    fi
-  - if [[ $TEST_TARGET == 'example' ]]; then
-      python -m iris.tests.runner --example-tests --print-failed-images --num-processors=3;
-    fi
-
   # Capture install-dir: As a test command must be last for get Travis to check
   # the RC, so it's best to start each operation with an absolute cd.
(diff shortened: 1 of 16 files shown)


QuLogic:
License check is skipped because this branch is [missing](https://github.com/SciTools/iris/blob/v2.2.x/.travis.yml#L128) the setting of `IRIS_REPO_DIR`, which is there on [master](https://github.com/SciTools/iris/blob/master/.travis.yml#L133).

corinnebosley:
@lbdreyer What's the strategy here?  Are we bothered that some of the license headers are still wrong?  Surely the main thing here is that the checks are working.

What do you want to do?

lbdreyer:
> License check is skipped because this branch is [missing](https://github.com/SciTools/iris/blob/v2.2.x/.travis.yml#L128) the setting of `IRIS_REPO_DIR`, which is there on [master](https://github.com/SciTools/iris/blob/master/.travis.yml#L133).
> 
> If you re-add that, it should run here.

Thank you @QuLogic. I have cherry-picked the commit that made that change to this PR.

We will just need to skip that commit when we do the mergeback onto master!



> @lbdreyer What's the strategy here? Are we bothered that some of the license headers are still wrong? Surely the main thing here is that the checks are working.
> 
> What do you want to do?

I'm going to update the license headers that the license header check is complaing about so that this won't be an issue to further PRs to this branch.
I'll fix this shortly!

corinnebosley:
Thanks @lbdreyer!

PR #3320, opened 2019-05-31
Update in preparation for Python3.8 deprecation of importing ABCs from collections

znicholls:
Updates the use of collections in preparation for Python3.8, see https://github.com/python/cpython/commit/c66f9f8d3909f588c251957d499599a1680e2320 (and also avoids deprecation warnings when using iris in Python3.7).

### DIFF
--- a/lib/iris/_constraints.py
+++ b/lib/iris/_constraints.py
@@ -1,4 +1,4 @@
-# (C) British Crown Copyright 2010 - 2017, Met Office
+# (C) British Crown Copyright 2010 - 2019, Met Office
 #
 # This file is part of Iris.
 #
@@ -23,7 +23,10 @@ from __future__ import (absolute_import, division, print_function)
 from six.moves import (filter, input, map, range, zip)  # noqa
 import six
 
-import collections
+try:  # Python 3
+    from collections.abc import Iterable, Mapping
+except:  # Python 2.7
+    from collections import Iterable, Mapping
 import operator
 
 import numpy as np
@@ -98,7 +101,7 @@ class Constraint(object):
             raise TypeError('cube_func must be None or callable, got %r'
                             % cube_func)
         if not (coord_values is None or isinstance(coord_values,
-                                                   collections.Mapping)):
+                                                   Mapping)):
             raise TypeError('coord_values must be None or a '
                             'collections.Mapping, got %r' % coord_values)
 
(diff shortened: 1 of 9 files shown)


znicholls:
Test failures appear to all be plotting and docs related so I think this is blocked for now

znicholls:
@DPeterK is my understanding correct that this is failing for reasons other than the changes made in the PR?

rcomer:
@znicholls you are right, all recent PRs against `master` have these errors.  I believe these should be fixed once #3316 is merged, then we can all rebase!

znicholls:
@rcomer are you the right person to ask for a review?

rcomer:
> are you the right person to ask for a review?

@bjlittle will know the best person to nominate to review this.

znicholls:
@bjlittle are you able to review this one, I think it's pretty trivial?

lib/iris/coord_categorisation.py
```diff
@@ -32,7 +32,10 @@
 import six
 
 import calendar
-import collections
+try:  # Python 3
+    from collections.abc import Counter
```
ajdawson: `Counter` isn't an abc and is not in the `collections.abc` module, this import will always fail. I don't think this module needs any updates.

ajdawson:
This looks OK except for the change to `collections.Counter` which I believe is incorrect. Can you sort that out @znicholls and we can get this merged.

lib/iris/coord_categorisation.py
znicholls: Oops thanks for the pick up!

znicholls:
Should be sorted now cheers @ajdawson ~

ajdawson:
Thanks @znicholls, would you mind squashing into a single commit? In case you don't know, you can squash these commits by doing an interactive rebase onto the commit before your first one `git rebase -i b209642` and choose squash/fixup for every commit except the first. Let me know if you have questions.

znicholls:
Assuming I haven't mucked up the header handling I think this should be good now

pp-mo:
> would you mind squashing into a single commit?

Just intrigued - why do you want @znicholls to do that, when you can squash on merge yourself : is it just for the contributor to write the commit message ?

ajdawson:
>Just intrigued @ajdawson - why do you want @znicholls to do that, when you can squash on merge yourself : is it just for the contributor to write the commit message ?

Habit I suppose... I assume you are referring to the squash-and-merge button? In this case that would have been a sensible use of it.

pp-mo:
> @ajdawson Habit I suppose...

😁

Issue #3235, opened 2018-11-29
missing example in the user guide for time constraints


nhsavage:
To my mind the Iris user guide skips a step in discussing time constraints. After the example with cube_11 = cube_all.extract(hour_11), an example where you have 2 dates in datetime format and want to constrain the data to be between these two dates seems a really common use case. 

something like:
```

d1 = datetime.datetime.strptime('20070416T0000Z', '%Y%m%dT%H%MZ')
d2 = datetime.datetime.strptime('20070430T0000Z', '%Y%m%dT%H%MZ')

daterange = iris.Constraint(time=lambda cell: d1<= cell.point <= d2)
data_subset = long_ts.extract(daterange)
```
Do other people think that this would be a useful addition? I think it is probably a fairly common use case and I got a bit confused by the discussion of partial datetime which cool but not needed here. I am happy to give this a go if people want it.

bjlittle:
Thanks for taking the time to highlight this @nhsavage.

I think that this would make a nice addition to the user guide. It's a small example, but like you said, is a very common use case. So I think that many users would fine it both useful and helpful.

If you pushed up a PR, I'd merge it :smile:

bjlittle:
@stephenworsley Reference [2.3.2 Constraining](https://scitools.org.uk/iris/docs/latest/userguide/loading_iris_cubes.html#constraining-on-time) on Time from the Iris documentation.

lbdreyer:
A example for constraining to a datetime range has been added in #3352.

PR #3352, opened 2019-07-09
Adding date range time example to docs.

stephenworsley:
Adding the example mentioned in #3235 to the docs.

### DIFF
--- a/docs/iris/src/userguide/loading_iris_cubes.rst
+++ b/docs/iris/src/userguide/loading_iris_cubes.rst
@@ -280,6 +280,7 @@ every week for many years:
 
 .. testsetup:: timeseries_range
 
+    import datetime
     import numpy as np
     from iris.time import PartialDateTime
     long_ts = iris.cube.Cube(np.arange(150), long_name='data', units='1')
@@ -314,6 +315,7 @@ facilities.  An :class:`iris.time.PartialDateTime` object can be compared to
 objects such as :class:`datetime.datetime` instances, and this comparison will
 then test only those 'aspects' which the PartialDateTime instance defines:
 
+    >>> import datetime
     >>> from iris.time import PartialDateTime
     >>> dt = datetime.datetime(2011, 3, 7)
     >>> print(dt > PartialDateTime(year=2010, month=6))
--- a/docs/iris/src/userguide/loading_iris_cubes.rst
+++ b/docs/iris/src/userguide/loading_iris_cubes.rst
@@ -278,26 +278,36 @@ A common case is the selection of all points between two given dates. In the
 following example we construct a time sequence representing the first day of
 every week for many years:
 
-   >>> import datetime
-   >>> import numpy as np
-   >>> long_ts = iris.cube.Cube(np.arange(150), long_name='data', units='1')
-   >>> _mondays = iris.coords.DimCoord(7 * np.arange(150), standard_name='time', units='days since 2007-04-09')
-   >>> long_ts.add_dim_coord(_mondays, 0)
-   >>> print(long_ts.coord('time'))
-   DimCoord([2007-04-09 00:00:00, 2007-04-16 00:00:00, 2007-04-23 00:00:00,
-             ...
-             2010-02-01 00:00:00, 2010-02-08 00:00:00, 2010-02-15 00:00:00],
-            standard_name='time', calendar='gregorian')
+.. testsetup:: timeseries_range
+
+    import numpy as np
+    from iris.time import PartialDateTime
+    long_ts = iris.cube.Cube(np.arange(150), long_name='data', units='1')
+    _mondays = iris.coords.DimCoord(7 * np.arange(150), standard_name='time', units='days since 2007-04-09')
+    long_ts.add_dim_coord(_mondays, 0)
+
+
+.. doctest:: timeseries_range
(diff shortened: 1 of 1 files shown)


bjlittle:
@stephenworsley Awesome, thanks.

You've got some Travis CI failures. Could you re-base your branch against `master` and push this to your branch?

Hopefully this should resolve these issues.

bjlittle:
@stephenworsley Thanks :+1: 

I'll just take a peek at the rendered documentation changes, and take it from there...

bjlittle:
@stephenworsley Nice. Good job :smile: 

Since we're in this space, could I cheekily ask for an additional change, just to bed in the example of using a date range in a constraint **and** the use of `iris.time.PartialDateTime` objects.

Just after the section:
```python
>>> import datetime
>>> from iris.time import PartialDateTime
>>> dt = datetime.datetime(2011, 3, 7)
>>> print(dt > PartialDateTime(year=2010, month=6))
True
>>> print(dt > PartialDateTime(month=6))
False
>>>
```
```These two facilities can be combined to provide straightforward calendar-based time selections when loading or extracting data.```
Could we bring it all together with a further example along the lines of:
```python
>>> from iris.time import PartialDateTime
>>> pdt1 = PartialDateTime(year=2007, month=4, day=16)
>>> pdt2 = PartialDateTime(year=2007, month=4, day=30)
>>> daterange = iris.Constraint(time=lambda cell: d1<= cell.point <= d2)
>>> data_subset = long_ts.extract(daterange)
>>> print(data_subset.coord('time'))
```
which builds on your new example, but using `PartialDateTime` objects instead... what do you think?

Of couse, add some suitable introductory text, as you see fit.

It also means that we might need to slightly rephrase the ```The previous constraint example can now be written as:``` sentence, as it's slightly confusing as to which constraint it's referring to - clearly the `hour_11` constraint.

bjlittle:
@stephenworsley Also if you make your next commit message "closes #3235", this mean that once this PR is merge, GitHub will automatically close the associated issue, see https://help.github.com/en/articles/closing-issues-using-keywords

stephenworsley:
@bjlittle I believe there is already an example doing more or less the same thing with the `st_swithuns_daterange`, though it might make sense to modify the dates in the new example to match up. I'll see what I can do about clearing up any ambiguity in the 'previous' references.

bjlittle:
@stephenworsley Yeah, I think that it's the overall flow of that section is just not quite right... there are a mixture of dates, constraints and examples and for first time reader, I can only think that it's harder work that it should be.

It would be awesome for you could manage to smooth out the wrinkles and yet get across all the relevant points of time constrains.

See how you go, and we can always talk about it in person tomorrow if it's easier :smile:

docs/iris/src/userguide/loading_iris_cubes.rst
```diff
@@ -314,21 +315,53 @@ for many years:
...
+A more complex example might require selecting points over an annually repeating
+date range. We can select points within a certain part of the year, in this case
+between the 15th of July through to the 25th of August. By making use of
+PartialDateTime this becomes simple:
 
 .. doctest:: timeseries_range
 
     >>> st_swithuns_daterange = iris.Constraint(
-    ...     time=lambda cell: PartialDateTime(month=7, day=15) < cell < PartialDateTime(month=8, day=25))
+    ...     time=lambda cell: PartialDateTime(month=7, day=15) <= cell < PartialDateTime(month=8, day=25))
```
lbdreyer: Why this change?

docs/iris/src/userguide/loading_iris_cubes.rst
```diff
@@ -314,21 +315,53 @@ for many years:
...
+    >>> st_swithuns_daterange_07 = iris.Constraint(
+    ...     time=lambda cell: d1 <= cell.point < d2)
+    >>> within_st_swithuns_07 = long_ts.extract(st_swithuns_daterange_07)
+    >>> print(within_st_swithuns_07.coord('time'))
+    DimCoord([2007-07-16 00:00:00, 2007-07-23 00:00:00, 2007-07-30 00:00:00,
+              2007-08-06 00:00:00, 2007-08-13 00:00:00, 2007-08-20 00:00:00],
+             standard_name='time', calendar='gregorian')
+
+Alternatively, we may do the same thing with :class:`iris.time.PartialDateTime`
+objects.
```
lbdreyer: > Alternatively, we may do the same thing with ...

This phrasing feels a bit too informal.
Could you rephrase to be something like:
> Alternatively, we may rewrite this using PartialDateTime objects ...

docs/iris/src/userguide/loading_iris_cubes.rst
stephenworsley: So there's two reasons for this. The first is that I wanted to make it more obvious that PartialDateTime range was the same as the range defined by datetetime objects. The datetime objects measure time from the start of the day, so for them it is necessary to specify the first date in the range and the first date outside of the range. With PartialDateTime, this is not necessary but I made the change so that the dates matched up.
The other reason is that this example appears to be a reference to a proverb about St Swithun's day https://en.wikipedia.org/wiki/Swithun. I decided that an object called `st_swithuns_daterange` could reasonably be expected to contain St Swithun's day itself (this does mean that there are now 41 days in the daterange, though this still fits with some readings of the proverb).

PR #3353, opened 2019-07-10
Remove version pin on proj4.

pp-mo:
***DO NOT MERGE*** (yet)

Check to see if tests still fail when removing this constraint

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -68,11 +68,15 @@ install:
     echo 'Install Iris dependencies';
     CONDA_REQS_FLAGS="";
     CONDA_REQS_GROUPS="test";
+    CONDA_INSTALL_EXTRAS="";
     if [[ "${PYTHON_VERSION}" == 2* ]]; then
         CONDA_REQS_FLAGS="${CONDA_REQS_FLAGS} --py2";
     fi;
     if [[ "${TEST_MINIMAL}" != true ]]; then
         CONDA_REQS_GROUPS="${CONDA_REQS_GROUPS} all";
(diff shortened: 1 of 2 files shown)


pp-mo:
> if tests still fail when removing this constraint
No they don't !!

So I think this is obsolete :  can someone please check + ideally merge ?

This is stopping (making skip) all the tests of the iris-grib interface.
That requires additional work, but I think this is  crucial.  Fixing those will unblock #2608

bjlittle:
@pp-mo We're using `pip` and at the moment `eccodes-python` is only available on PyPI - this is the only version that works in Python3 at the moment. Offline we can help the ECMWF package with conda.

pp-mo:
> We're using pip and at the moment eccodes-python is only available on PyPI

Ok, I was hoping to simplify here, but that explains it.
I can probably modify this, copying the iris-grib testing approach, to get Python3 testing working also.

In the meantime, I see we now have a bunch of things that look like they aren't compatible between gribapi and eccodes -- but presumably not tested in grib-api itself.
Some of this testing probably still wants to be migrated out of  iris : We still have a heap of tech debt in iris-grib!!

bjlittle:
@pp-mo So what are you intentions for this particular PR? 

Seems like you are pulling on quite a few strings that reach across into `iris-grib`, `eccodes`,`python-eccodes` and (confusingly) `eccodes-python`.

pp-mo:
Let's kill this off
and replace with a PR that just unpins proj4 like it says on the box :wink:  
and create a separate issue to address getting iris-grib interface testing back up.

PR #1952, opened 2016-03-03
pep8 and docstring untruths

marqh:
non-functional change to pp.py

pep8 plus removal of claims about save and as_pairs, as_fields being able to take CubeLists as inputs

Issue #425, opened 2013-03-27
Misleading docs for iris.fileformats.pp.save


DPeterK:
The docs for iris.fileformats.pp.save (http://www.scitools.org.uk/iris/docs/latest/iris/iris/fileformats/pp.html?highlight=save%20cube#iris.fileformats.pp.save) should be updated to reflect iris.save being the preferred method of saving data into a cube. 

Further, the docs should be updated to state that 'cube' may only be a single cube &ndash; passing a iris.cube.CubeList or list of cubes causes iris.fileformats.pp.save to exit with an error:

```
>>> import iris.fileformats.pp as ipp
>>> type(cubelist)
<class 'iris.cube.CubeList'>
>>> ipp.save(cubelist, '/filepath/cubes.pp')
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "/../iris/lib/iris/fileformats/pp.py", line 1598, in save
    n_dims = len(cube.shape)
AttributeError: 'CubeList' object has no attribute 'shape'
```

And again:

```
>>> ipp.save([cubelist[0], cubelist[1]], '/filepath/cubes.pp')
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "/../iris/lib/iris/fileformats/pp.py", line 1598, in save
    n_dims = len(cube.shape)
AttributeError: 'list' object has no attribute 'shape'
```

jonnyhtw:
Hi there

Has there been a solution to this?

I am currently unable to save my `iris.cube.CubeList` to a NetCDF file using Iris version 2.1.0 ...

```
>>> type(all_cubes)

iris.cube.CubeList

>>> iris.save(all_cubes,outdir+'foo.pp')

---------------------------------------------------------------------------
AttributeError                            Traceback (most recent call last)
<ipython-input-25-5a29685d3771> in <module>()
----> 1 iris.save(all_cubes,outdir+'foo.pp')

/nesi/nobackup/niwa00013/williamsjh/miniconda2/envs/master/lib/python2.7/site-packages/iris/io/__init__.pyc in save(source, target, saver, **kwargs)
    425                 if i != 0:
    426                     kwargs['append'] = True
--> 427                 saver(cube, target, **kwargs)
    428         # Netcdf saver.
    429         else:

/nesi/nobackup/niwa00013/williamsjh/miniconda2/envs/master/lib/python2.7/site-packages/iris/fileformats/pp.pyc in save(cube, target, append, field_coords)
[...]

PR #3354, opened 2019-07-11
Save docs patch

stephenworsley:
This adds documentation requested in #425. The documentation that only Cubes can be passed was already added in #1952.


bjlittle:
@stephenworsley Awesome, thasks :+1:

PR #3355, opened 2019-07-11
Unpin dask

bjlittle:
This PR undoes the temporary workaround implemented by PR https://github.com/SciTools/iris/pull/3350, which pinned `dask<2`.

The latest version of `dask` version `2.1.0` (only a few days old now) resolves the outstanding `stack overflow` issues avoided by our pin. 

Also see https://github.com/ESMValGroup/ESMValTool/issues/1188 for further details.

### DIFF
--- a/lib/iris/tests/unit/fileformats/pp/test__create_field_data.py
+++ b/lib/iris/tests/unit/fileformats/pp/test__create_field_data.py
@@ -64,7 +64,7 @@ class Test__create_field_data(tests.IrisTest):
         field = mock.Mock(core_data=core_data)
         data_shape = (100, 120)
         proxy = mock.Mock(dtype=np.dtype('f4'), shape=data_shape,
-                          spec=pp.PPDataProxy)
+                          spec=pp.PPDataProxy, ndim=len(data_shape))
         # We can't directly inspect the concrete data source underlying
         # the dask array, so instead we patch the proxy creation and check it's
         # being created and invoked correctly.
--- a/lib/iris/tests/unit/lazy_data/test_co_realise_cubes.py
+++ b/lib/iris/tests/unit/lazy_data/test_co_realise_cubes.py
@@ -1,4 +1,4 @@
-# (C) British Crown Copyright 2018, Met Office
+# (C) British Crown Copyright 2018 - 2019, Met Office
 #
 # This file is part of Iris.
 #
@@ -36,6 +36,7 @@ class ArrayAccessCounter(object):
     def __init__(self, array):
         self.dtype = array.dtype
         self.shape = array.shape
+        self.ndim = array.ndim
         self._array = array
         self.access_count = 0
 
--- a/lib/iris/tests/unit/lazy_data/test_co_realise_cubes.py
+++ b/lib/iris/tests/unit/lazy_data/test_co_realise_cubes.py
@@ -72,15 +72,29 @@ class Test_co_realise_cubes(tests.IrisTest):
         self.assertTrue(cube_inner.has_lazy_data())
 
     def test_combined_access(self):
+        import dask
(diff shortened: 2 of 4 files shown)


lib/iris/tests/unit/fileformats/pp/test__create_field_data.py
```diff
@@ -64,7 +64,7 @@ def test_deferred_bytes(self):
         field = mock.Mock(core_data=core_data)
         data_shape = (100, 120)
         proxy = mock.Mock(dtype=np.dtype('f4'), shape=data_shape,
-                          spec=pp.PPDataProxy)
+                          spec=pp.PPDataProxy, ndim=len(data_shape))
```
pp-mo: OK, but I'm really surprised by this development.
Hitherto it was sufficient for a wrapped object to have just a `.dtype` and a `.shape`.
As ndim is a trivial consequence of shape, I wonder if this is a bit of a mistake in dask, and should be queried ??

lib/iris/tests/unit/lazy_data/test_co_realise_cubes.py
```diff
@@ -71,15 +72,29 @@ def test_multi(self):
...
         cube_b = Cube(derived_b)
-        co_realise_cubes(cube_a, cube_b)
-        # Though used twice, the source data should only get fetched once.
-        self.assertEqual(wrapped_array.access_count, 1)
+        cube_c = Cube(derived_c)
+        co_realise_cubes(cube_a, cube_b, cube_c)
+        # Though used more than once, the source data should only get fetched
+        # once by dask.
+        if Version(dask.__version__) < Version('2.0.0'):
+            self.assertEqual(wrapped_array.access_count, 1)
```
pp-mo: Do we really need this older form?
Instead we could just specify dask>=2 ??

bjlittle: @pp-mo `dask` version `2.+` is Python3 only. 

As you know, in Travis CI we currently test Python `2.7`, `3.6` and now `3.7` - so yes, we do need this to cover the Travis CI testing split between Python2 and Python3... for now.

I've created the issue https://github.com/SciTools/iris/issues/3356 to tidy this for `iris` version `3.0.0`.

lib/iris/tests/unit/fileformats/pp/test__create_field_data.py
bjlittle: @pp-mo From what I understand there has been a bit of a refactor for `dask` version `2.0.0`, so for whatever reason they have started to access the `ndim` attribute of `dask.array` like objects.

Perhaps raise your concerns with them in an issue, if you care, but I don't see this as a blocker to this PR. The horse has kinda bolted, as they have already cut `dask` version `2.0.0` and now `2.1.0`.

bjlittle:
@pp-mo Unless you've got other objections, I think this PR is good to go :wink: ... what do you think?

lib/iris/tests/unit/lazy_data/test_co_realise_cubes.py
pp-mo: Ok, understood
Shame though!

lib/iris/tests/unit/fileformats/pp/test__create_field_data.py
pp-mo: FWIW, raised https://github.com/dask/dask/issues/5106

bjlittle: @pp-mo I see that itch has been scratched 😜

pp-mo:
Respect for patience @bjlittle !

PR #3361, opened 2019-07-25
Chunk control

### LINKED ISSUE #3357: Reading netcdf files is slow if there are unlimited dimensions
Reading arrays from NetCDF files is slow if one dimension is unlimited.

An example: reading an array of shape `(5000, 50)` takes ~7 s if the first dimension is unlimited. If both dimensions are fixed, it takes ~0.02 s. This is a major bottleneck if many (100s) of such files need to be processed. Time dimension is often declared unlimited in files generated by circulation models.

Test case:
```
import iris
import time

f = 'example_dataset.nc'
var = 'sea_water_practical_salinity'

tic = time.process_time()
cube = iris.load_cube(f, var)
cube.data
duration = time.process_time() - tic
print('Duration {:.3f} s'.format(duration))
```

The input NetCDF file can be generated with:
```
import iris
import numpy
import datetime

ntime = 5000
nz = 50
dt = 600.
time = numpy.arange(ntime, dtype=float)*dt
date_zero = datetime.datetime(2000, 1, 1)
date_epoch = datetime.datetime.utcfromtimestamp(0)
time_epoch = time + (date_zero - date_epoch).total_seconds()
z = numpy.linspace(0, 10, nz)
values = 5*numpy.sin(time/(14*24*3600.))
values = numpy.tile(values, (nz, 1)).T

time_dim = iris.coords.DimCoord(time_epoch, standard_name='time',
                                units='seconds since 1970-01-01 00:00:00-00')
z_dim = iris.coords.DimCoord(z, standard_name='depth', units='m')
cube = iris.cube.Cube(values)
cube.standard_name = 'sea_water_practical_salinity'
cube.units = '1'
cube.add_dim_coord(time_dim, 0)
cube.add_dim_coord(z_dim, 1)
iris.fileformats.netcdf.save(cube, 'example_dataset.nc',
                             unlimited_dimensions=['time'])
```

Profiling suggest that in the unlimited case, each time slice is being read separately, i.e. `NetCDFDataProxy.__getitem__` is being called 5000 times.

Tested with: iris version 2.2.0, Anaconda3 2019.03

pp-mo:
Revised chunking policy, which will mostly affect loading from netcdf files.
Key change : allows multiplying up as well as dividing chunks, mainly to fix the #3357.

Closes #3357 #3362

### DIFF
--- a/lib/iris/_lazy_data.py
+++ b/lib/iris/_lazy_data.py
@@ -60,9 +60,9 @@ def is_lazy_data(data):
     return result
 
 
-def _optimise_chunksize(chunks, shape,
-                        limit=None,
-                        dtype=np.dtype('f4')):
+def _optimum_chunksize(chunks, shape,
+                       limit=None,
+                       dtype=np.dtype('f4')):
     """
     Reduce or increase an initial chunk shape to get close to a chosen ideal
     size, while prioritising the splitting of the earlier (outer) dimensions
@@ -177,7 +177,7 @@ def as_lazy_data(data, chunks=None, asarray=False):
         chunks = list(data.shape)
 
     # Expand or reduce the basic chunk shape to an optimum size.
-    chunks = _optimise_chunksize(chunks, shape=data.shape)
+    chunks = _optimum_chunksize(chunks, shape=data.shape, dtype=data.dtype)
 
     if isinstance(data, ma.core.MaskedConstant):
         data = ma.masked_array(data.data, mask=data.mask)
--- a/lib/iris/_lazy_data.py
+++ b/lib/iris/_lazy_data.py
@@ -60,19 +60,6 @@ def is_lazy_data(data):
     return result
 
 
-# A magic value, chosen to minimise chunk creation time and chunk processing
-# time within dask.
-_MAX_CHUNK_SIZE = 8 * 1024 * 1024 * 2
-
-# _MAX_CHUNK_SIZE = dask.config.get('array', {}).get('chunk-size', None)
-# if _MAX_CHUNK_SIZE is not None:
-#     # Convert to bytes
-#     _MAX_CHUNK_SIZE = da.core.parse_bytes(_MAX_CHUNK_SIZE)
-# else:
-#     # Fall back on our own "magic" value.
-#     _MAX_CHUNK_SIZE = 8 * 1024 * 1024 * 2
-
-
 def _optimise_chunksize(chunks, shape,
                         limit=None,
                         dtype=np.dtype('f4')):
@@ -88,7 +75,7 @@ def _optimise_chunksize(chunks, shape,
     * shape (tuple of int):
         The full array shape of the target data.
     * limit (int):
-        The 'ideal' target chunk size, in bytes.
+        The 'ideal' target chunk size, in bytes.  Default from dask.config.
     * dtype (np.dtype):
         Numpy dtype of target data.
 
@@ -97,18 +84,18 @@ def _optimise_chunksize(chunks, shape,
         The proposed shape of one full chunk.
 
     .. note::
-        The use of this is very similar to `dask.array.core.normalize_chunks`,
-        when called as
+        The purpose of this is very similar to
+        `dask.array.core.normalize_chunks`, when called as
         `(chunks='auto', shape, dtype=dtype, previous_chunks=chunks, ...)`.
-        Except : that the logic here is optimised for a specific to a 'c-like'
-        dimension order, i.e. outer dimensions first, as in netcdf variables.
+        Except, the operation here is optimised specifically for a 'c-like'
+        dimension order, i.e. outer dimensions first, as for netcdf variables.
         So if, in future, this policy can be implemented in dask, then we would
         prefer to replace this function with a call to that one.
         Accordingly, the arguments roughly match 'normalize_chunks', except
         that we don't support the alternative argument forms of that routine.
-        This routine also returns a single 'full chunk', rather
-        than a complete chunking scheme : so equivalent code usage would be
-        "chunks = [c[0] for c in normalise_chunks(chunks, ...)]".
+        The return value, however, is a single 'full chunk', rather than a
+        complete chunking scheme : so an equivalent code usage could be
+        "chunks = [c[0] for c in normalise_chunks('auto', ...)]".
 
     """
     # Return chunks unchanged, for types of invocation we don't comprehend.
@@ -119,9 +106,13 @@ def _optimise_chunksize(chunks, shape,
         # or if shape contains 0 or -1 (like raw landsea-mask data proxies).
         return chunks
 
-    # Calculate default chunksize limit.
+    # Set the chunksize limit.
     if limit is None:
-        limit = _MAX_CHUNK_SIZE * 4
+        # Fetch the default 'optimal' chunksize from the dask config.
+        limit = dask.config.get('array.chunk-size')
+        # Convert to bytes
+        limit = da.core.parse_bytes(limit)
+
     point_size_limit = limit / dtype.itemsize
 
     # Create result chunks, starting with a copy of the input.
--- a/lib/iris/_lazy_data.py
+++ b/lib/iris/_lazy_data.py
@@ -23,16 +23,12 @@ To avoid replicating implementation-dependent test and conversion code.
 from __future__ import (absolute_import, division, print_function)
 from six.moves import (filter, input, map, range, zip)  # noqa
 
-try:  # Python 3
-    from collections.abc import Iterable
-except ImportError:  # Python 2
-    from collections import Iterable
 from functools import wraps
 
 import dask
 import dask.array as da
-import dask.array.core
 import dask.config
+import dask.utils
 
 import numpy as np
 import numpy.ma as ma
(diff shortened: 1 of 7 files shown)


pp-mo:
I think I've finally got what I want out of this now.
Good to go @bjlittle ? :pray:

pp-mo:
Whoops. :alarm_clock: :warning: :bomb: 
Sorry guys, I didn't mean to do that ...

Hopefully fixed.

pp-mo:
Evidence for this fixing the original problem in #3357 ...
Running sample testcode from there.
**"Before"** examples (release v2.2.1) :
```
Duration 25.450 s
Duration 26.748 s
Duration 27.222 s
```
**"After"** examples :
```
Duration 0.219 s
Duration 0.222 s
Duration 0.232 s
```

tkarna:
Thanks! I confirm that this fixes the issue #3357.

bjlittle:
@pp-mo This is also applicable to #3362

lib/iris/_lazy_data.py
```diff
@@ -23,12 +23,14 @@
 from __future__ import (absolute_import, division, print_function)
 from six.moves import (filter, input, map, range, zip)  # noqa
 
+from collections import Iterable
```
bjlittle: @pp-mo See https://github.com/SciTools/iris/pull/3320 for context.

Currently in Python3.7, you get the following `DeprecationWarning`:
```python
>>> from collections import Iterable
__main__:1: DeprecationWarning: Using or importing the ABCs from 'collections' instead of from 'collections.abc' is deprecated, and in 3.8 it will stop working
```
Could you adopt the following pattern:
```python
try:  # Python 3
    from collections.abc import Iterable
except ImportError:  # Python 2
```

lib/iris/_lazy_data.py
```diff
@@ -106,9 +172,12 @@ def as_lazy_data(data, chunks=None, asarray=False):
```
lbdreyer: Can you update the doc string for this function. The "chunk" explanation" is out of date

lib/iris/_lazy_data.py
pp-mo: Looks good, will do !

pp-mo:
@lbdreyer I think this is only failing due to numpy 1.17,
thus since #3369, a rebase should get a clean pass. 
Are you ok for that, or are you still writing review comments on existing commits ?

lib/iris/_lazy_data.py
```diff
@@ -58,26 +63,90 @@ def is_lazy_data(data):
...
+        that we don't support the alternative argument forms of that routine.
+        The return value, however, is a single 'full chunk', rather than a
+        complete chunking scheme : so an equivalent code usage could be
+        "chunks = [c[0] for c in normalise_chunks('auto', ...)]".
 
+    """
+    # Return chunks unchanged, for types of invocation we don't comprehend.
+    if (any(elem <= 0 for elem in shape) or
+            not isinstance(chunks, Iterable) or
+            len(chunks) != len(shape)):
```
lbdreyer: You don't have explicit tests for this check.
I'm not sure how thorough we want to be with testing this. It does seem like a bit of overkill to add tests for 
`_optimum_chunksize((200,300), (1,200,300))`

lib/iris/_lazy_data.py
```diff
@@ -58,26 +63,90 @@ def is_lazy_data(data):
...
-        new_dim = int(shape[i_reduce] / factor)
-        if new_dim < 1:
-            new_dim = 1
-        shape[i_reduce] = new_dim
-        i_reduce += 1
-    return tuple(shape)
+    return tuple(result)
 
 
 def as_lazy_data(data, chunks=None, asarray=False):
```
lbdreyer: Could you modify the 'chunks' section in the docstring now that we are no longer using the Biggus magic number

lib/iris/_lazy_data.py
```diff
@@ -58,26 +60,90 @@ def is_lazy_data(data):
...
+        # Don't modify chunks for special values like -1, (0,), 'auto',
+        # or if shape contains 0 or -1 (like raw landsea-mask data proxies).
+        return chunks
+
+    # Set the chunksize limit.
+    if limit is None:
+        # Fetch the default 'optimal' chunksize from the dask config.
+        limit = dask.config.get('array.chunk-size')
+        # Convert to bytes
+        limit = da.core.parse_bytes(limit)
```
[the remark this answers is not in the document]
pp-mo: I think I copied this from the dask code.
I will fix it.

lib/iris/fileformats/netcdf.py
```diff
@@ -511,7 +511,7 @@ def _get_cf_var_data(cf_var, filename):
     proxy = NetCDFDataProxy(cf_var.shape, dtype, filename, cf_var.cf_name,
                             fill_value)
     chunks = cf_var.cf_data.chunking()
-    # Chunks can be an iterable, None, or `'contiguous'`.
+    # Chunks can be an iterable, or `'contiguous'`.
```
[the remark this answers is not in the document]
pp-mo: There are two different sets of conventions here.
  * The 'chunks' value that nc.Variable.data_chunking returns can *not* be None, I believe.
     * I don't know quite why it ever said that : it just seems wrong to me.
  * The 'chunks' keyword we pass to as_lazy_data *can* be None.  And it absolutely can't be 'contiguous', which is why we are converting here.

I will try to clarify in the comments.

lib/iris/_lazy_data.py
```diff
@@ -58,26 +60,90 @@ def is_lazy_data(data):
...
+
+    point_size_limit = limit / dtype.itemsize
+
+    # Create result chunks, starting with a copy of the input.
+    result = list(chunks)
+    if shape is None:
```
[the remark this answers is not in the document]
pp-mo: I think I was trying to mimic the API of `dask.array.core.normalize_chunks`, in case we can use that in the future.
Actually I haven't achieved that, and we never use this option, so I will remove it.

lib/iris/_lazy_data.py
```diff
@@ -58,26 +63,90 @@ def is_lazy_data(data):
...
+        # (i.e. inner) dims first.
+        i_expand = len(shape) - 1
+        while np.prod(result) < point_size_limit and i_expand >= 0:
+            factor = np.floor(point_size_limit * 1.0 / np.prod(result))
+            new_dim = result[i_expand] * int(factor)
+            # Clip to dim size : N.B. means it cannot exceed the original dims.
+            if new_dim > shape[i_expand]:
+                new_dim = shape[i_expand]
+            result[i_expand] = new_dim
+            i_expand -= 1
```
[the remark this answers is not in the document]
pp-mo: That's true, but I think it is ok.  Dask does this too, it does not attempt to be "neat" about it.

lib/iris/_lazy_data.py
pp-mo: This was an attempt to allow alternative, dask-type chunks arguments in 'as_lazy_data'.
Obviously we don't use anything like that at present.  The only immediate need is to skip shapes with a zero in them (see comment).

I now see this is wrong anyway, as that initial test clause assumes that shape is iterable !
I will simplify to just what we need, and add a testcase.

lib/iris/_lazy_data.py
pp-mo: Ok.

lib/iris/_lazy_data.py
pp-mo: ... after some thought, I have removed this to the caller + documented the behaviour there.

pp-mo:
Attempted to address review comments.
Please re-consider @lbdreyer .

Note : I didn't make any changes to address [the comment about the size of divided chunks](https://github.com/SciTools/iris/pull/3361/#discussion_r315683225).
We can still discuss that if wanted.

pp-mo:
Thanks @lbdreyer this fixes the outstanding errors.

I'm happy that those other uses in testing of as_lazy_data(.. chunks=X ..) are all now unnecessary, and we don't need to support any more complex usages : We accept that this 'chunks' keyword is not much like the dask one, and that is ok : the new docstring reflects this I hope.

Regarding the "lost" test, 
`test_as_lazy_data.Test__optimised_chunks.test_large_specific_chunk_passthrough`
I think that is now obsolete, because the earlier idea was that a specified chunks= should always pass through unchanged to the dask call, but that is definitely no longer the case.

pp-mo:
Meanwhile, though ...

I'm still wondering about [your comment "This does end with an array that is unequally split into chunks"](https://github.com/SciTools/iris/pull/3361/#discussion_r315683225)
There could still be an improvement that can be made here, but I'm still not clear exactly how it should work.

So far : a good practical example ...
```
>>> _optimum_chunksize((3, 300, 200), (111, 300, 2000))
(54, 300, 2000)
```
this will result in 3 chunks with initial dimensions (54, 54, 3), 
but you could have had (37, 37, 37) instead.
I will investigate how that can be calculated, taking account of what happens to slightly-smaller and slightly-larger cases (i.e. that don't divide equally).

pp-mo:
Interested ? @cpelley

pp-mo:
Hi @lbdreyer
Finally, I think I made sense of the point you raised above, regarding a better choice of chunk sizes.
Hope this makes sense, sorry it has proved so complicated to explain + test !

lib/iris/_lazy_data.py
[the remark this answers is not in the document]
lbdreyer: This has now been addressed in the last 3 commits

pp-mo:
Hi again @lbdreyer
As-per offline discussion, now simplified the testing a bit.

lbdreyer:
This is a really great change! :100:

Thanks for persisting with it @pp-mo!

Issue #3365, opened 2019-08-05
Ifunc, inspect and numpy 1.17


rcomer:
My latest PR has an unexpected [Travis failure](https://travis-ci.org/SciTools/iris/jobs/567844517#L5124 ) in `test_ifunc`:
```
ERROR: test_ifunc (iris.tests.test_basic_maths.TestIFunc)
----------------------------------------------------------------------
Traceback (most recent call last):
  File "/home/travis/miniconda/envs/test-environment/lib/python3.6/site-packages/scitools_iris-2.3.0.dev0-py3.6.egg/iris/tests/test_basic_maths.py", line 605, in test_ifunc
    lambda a: a.units
  File "/home/travis/miniconda/envs/test-environment/lib/python3.6/site-packages/scitools_iris-2.3.0.dev0-py3.6.egg/iris/analysis/maths.py", line 928, in __init__
    raise ValueError(msg.format(data_func.__name__, self.nin))
ValueError: cumsum requires 0 input data arrays, the IFunc class currently only supports functions requiring 1 or two data arrays as input.
```

  `Ifunc` uses either the `nin` attribute (which `numpy.cumsum` doesn't appear to have) or `inspect.getargspec` to [figure out how many arrays can be passed to the numpy function](https://github.com/SciTools/iris/blob/master/lib/iris/analysis/maths.py#L917 ).

I note that numpy 1.17 was released at the end of July, which probably explains why this failure has appeared now.  I think this is as far as my debugging skills can take me on this one though!

PR #3366, opened 2019-08-09
Dress naked excepts

bjlittle:
This PR fixes the `iris` travis-ci tests after moving to the latest `numpy` 1.17.0 - and it also follows-up from https://github.com/SciTools/iris/pull/3320 to dress the naked `try`...`except`s with an `ImportError`

### DIFF
--- a/lib/iris/analysis/maths.py
+++ b/lib/iris/analysis/maths.py
@@ -1,4 +1,4 @@
-# (C) British Crown Copyright 2010 - 2017, Met Office
+# (C) British Crown Copyright 2010 - 2019, Met Office
 #
 # This file is part of Iris.
 #
@@ -22,10 +22,11 @@ Basic mathematical and statistical operations.
 from __future__ import (absolute_import, division, print_function)
 from six.moves import (filter, input, map, range, zip)  # noqa
 
-import warnings
+import inspect
 import math
 import operator
-import inspect
+import six
+import warnings
 
 import cf_units
 import numpy as np
@@ -917,9 +918,15 @@ class IFunc(object):
         if hasattr(data_func, 'nin'):
             self.nin = data_func.nin
         else:
-            (args, varargs, keywords, defaults) = inspect.getargspec(data_func)
-            self.nin = len(args) - (
-                len(defaults) if defaults is not None else 0)
+            if six.PY2:
+                (args, _, _, defaults) = inspect.getargspec(data_func)
+                self.nin = len(args) - (
+                    len(defaults) if defaults is not None else 0)
+            else:
+                sig = inspect.signature(data_func)
+                args = [param for param in sig.parameters.values()
+                        if '=' not in str(param)]
+                self.nin = len(args)
 
         if self.nin not in [1, 2]:
             msg = ('{} requires {} input data arrays, the IFunc class '
--- a/lib/iris/_constraints.py
+++ b/lib/iris/_constraints.py
@@ -25,7 +25,7 @@ import six
 
 try:  # Python 3
     from collections.abc import Iterable, Mapping
-except:  # Python 2.7
+except ImportError:  # Python 2.7
     from collections import Iterable, Mapping
 import operator
 
--- a/lib/iris/analysis/__init__.py
+++ b/lib/iris/analysis/__init__.py
@@ -53,7 +53,7 @@ import six
 from collections import OrderedDict
 try:  # Python 3
     from collections.abc import Iterable
-except:  # Python 2.7
+except ImportError:  # Python 2.7
     from collections import Iterable
 from functools import wraps
 
--- a/lib/iris/coords.py
+++ b/lib/iris/coords.py
@@ -27,7 +27,7 @@ from abc import ABCMeta, abstractproperty
 from collections import namedtuple
 try:  # Python 3
     from collections.abc import Iterator
-except:  # Python 2.7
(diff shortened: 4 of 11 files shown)


lib/iris/analysis/maths.py
```diff
@@ -917,9 +918,15 @@ def ws_units_func(u_cube, v_cube):
...
-                len(defaults) if defaults is not None else 0)
+            if six.PY2:
+                (args, _, _, defaults) = inspect.getargspec(data_func)
+                self.nin = len(args) - (
+                    len(defaults) if defaults is not None else 0)
+            else:
+                sig = inspect.signature(data_func)
+                args = [param for param in sig.parameters.values()
+                        if '=' not in str(param)]
+                self.nin = len(args)
```
bjlittle: Avoid `DeprecationWarning: inspect.getargspec() is deprecated since Python 3.0, use inspect.signature() or inspect.getfullargspec()` in Python3.

Also, signature of `numpy.cumsum` from `1.17.0` has changed resulting in `inspect.getargspec` not working, hence the move to `inspect.signature` in order to get `iris.tests.test_basic_maths.TestIFunc.test_ifunc` to pass again.

bjlittle:
@rcomer Fancy taking this on?

bjlittle:
Closes https://github.com/SciTools/iris/issues/3365

lib/iris/analysis/maths.py
rcomer: I'm new to `inspect.Signature` and `inspect.Parameter`, but would this also work?
```python
args = [param for param in sig.parameters.values() if param.default is param.empty]
```
It seems a bit more direct than searching the string.

Not really important, but I figure I ought to find **something** to comment on :laughing:.

pp-mo: > I'm new to inspect.Signature and inspect.Parameter,

Me too, and it's more complicated than I expected...

Digging a little deeper, I think the above doesn't _quite_ work, because (I found), since Python 3 now supports keyword-only parameters, it is now possible to have a keyword without a default :   It becomes a _required_ keyword.
Like this : 
```
>>> def myfn(a, b, *, reqd_key):
...   return (a, b, reqd_key)
... 
>>> myfn(1,2)
TypeError: myfn() missing 1 required keyword-only argument: 'reqd_key'
>>> myfn(1,2,3)
TypeError: myfn() takes 2 positional arguments but 3 were given
>>> myfn(1,2, reqd_key=3)
(1, 2, 3)

>>> sig = inspect.Signature.from_callable(myfn)
>>> for name, par in sig.parameters.items():
...   print("'{}': kind={}, default={}".format(name, par.kind, par.default))
... 
'a': kind=1, default=<class 'inspect._empty'>
'b': kind=1, default=<class 'inspect._empty'>
'reqd_key': kind=3, default=<class 'inspect._empty'>
>>> 
```

Worse (?!?!), Python 3 since (I think) 3.3 now supports [the concept of *positional-only* parameters](https://docs.python.org/3/faq/programming.html#faq-positional-only-arguments)
  \- even though a Python syntax for defining them only exists [from Python 3.8 (I am really not making this up)](https://docs.python.org/3.8/whatsnew/3.8.html#positional-only-parameters).

So to be safe, I think we now need to use something like 
```
    [param for param in sig.parameters.values()
     if param.kind in (Parameter.POSITIONAL_ONLY, Parameter.POSITIONAL_OR_KEYWORD)]
```
Urrgh.
Anyone for going back to small+simple Python2 ??

rcomer: `POSITTIONAL_OR_KEYWORD` includes oldschool keywords, e.g. `axis` in `np.cumsum`, that we specifically don't want to count:
```
In [3]: sig = inspect.Signature.from_callable(np.cumsum)                                                                                                      

In [4]: for name, par in sig.parameters.items(): 
   ...:     print("'{}': kind={}, default={}".format(name, par.kind, par.default)) 
   ...:                                      

'a': kind=1, default=<class 'inspect._empty'>
'axis': kind=1, default=None
'dtype': kind=1, default=None
'out': kind=1, default=None
```

So I think we still want to check the defaults.  Maybe

```python
if param.kind != param.KEYWORD_ONLY and param.default is param.empty
```

?

In any case, we are still relying on `numpy` type functions sticking to a convention that arrays are "positional"  parameters and all other parameters are "keywords".  This is possibly not ideal, though seems to have served us OK so far.

pp-mo: Yes, I think you are exactly right @rcomer.

Any "normal, old-fashioned named parameter" could be required (== has no default) or optional (== has a default), but inspect still labels all those as 'POSITIONAL_OR_KEYWORD', so this "includes oldschool keywords, e.g. axis in np.cumsum" as you said.
We want to count the "required positional arguments", which means ignoring anything optional, i.e. with a default.  The only change is, since Python3, we now have keyword-onlys, which must also not appear on this count.

pp-mo:
Big Note:

We want this fixed ASAP now.  As @bjlittle is currently unavailable, I'm going to adopt this + replace with my own branch + PR.
I hope you can relate to that and ideally review+merge it for us @rcomer ?? 🙏

rcomer:
> I hope you can relate to that and ideally review+merge it for us

Sounds reasonable to me.

pp-mo:
Closed in favour of #3369

Issue #3340, opened 2019-06-21
iplt.quiver reshapes u but not v if longitude coordinate is circular


SimonPeatman:
Following a question raised on the Google Group: https://groups.google.com/forum/#!topic/scitools-iris/_w4QkS00wLU

If you do a quiver plot using Cubes with longitude coordinates which are "circular", `_map_common` copies the column of values at 0deg longitude and appends it to the end as 360deg longitude.  However, it does this for the u data but not the v data.  Therefore, the two arrays have different shapes so `_vector_component_args` then crashes.

e.g.:
```
# Imports
import numpy as np
import iris
import iris.plot as iplt
import matplotlib.pyplot as plt

# Create lat and lon coordinates
res = 5
lat = iris.coords.DimCoord(np.arange(-90, 91, res), 'latitude',
                           units='degrees_north')
lon = iris.coords.DimCoord(np.arange(0, 360, res), 'longitude',
                           units='degrees_east')

# Create u and v Cubes
nlat = len(lat.points)
nlon = len(lon.points)
u_arr = np.ones((nlat, nlon))
v_arr = np.ones((nlat, nlon))
u = iris.cube.Cube(u_arr, dim_coords_and_dims=[(lat, 0), (lon, 1)],
                   standard_name='eastward_wind')
v = iris.cube.Cube(v_arr, dim_coords_and_dims=[(lat, 0), (lon, 1)],
                   standard_name='northward_wind')

# Try plotting
ax1 = plt.subplot(211)
iplt.quiver(u, v)   # This one works

# Try making the longitude coordinate "circular", then plot again
lon.circular = True
ax2 = plt.subplot(212)
iplt.quiver(u, v)   # This one crashes
```

bjlittle:
Hey @SimonPeatman, thanks for taking the time to report this, and apologies for the delay in getting back to you.

Also, thanks for the example, which perfectly highlights and clarifies the issue. 

Would you be tempted to make a pull-request to resolve the issue?

lbdreyer:
A fix for this bug has been added in #3371 which I am about to merge in :)

PR #3371, opened 2019-08-21
Fix circular quiver plots

stephenworsley:
Fixes bug in #3340.

### DIFF
--- a/lib/iris/tests/integration/plot/test_vector_plots.py
+++ b/lib/iris/tests/integration/plot/test_vector_plots.py
@@ -167,6 +167,26 @@ class MixinVectorPlotCases(object):
             self.plot('2d_rotated', u_cube, v_cube,
                       coords=('longitude', 'latitude'))
 
+    def test_circular_longitude(self):
+        # Test circular longitude does not cause a crash.
+        res = 5
+        lat = DimCoord(np.arange(-90, 91, res), 'latitude',
+                                   units='degrees_north')
+        lon = DimCoord(np.arange(0, 360, res), 'longitude',
+                                   units='degrees_east')
+        nlat = len(lat.points)
+        nlon = len(lon.points)
+        u_arr = np.ones((nlat, nlon))
+        v_arr = np.ones((nlat, nlon))
+        u_cube = Cube(u_arr, dim_coords_and_dims=[(lat, 0), (lon, 1)],
+                           standard_name='eastward_wind')
+        v_cube = Cube(v_arr, dim_coords_and_dims=[(lat, 0), (lon, 1)],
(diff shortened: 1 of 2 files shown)


stephenworsley:
@trexfeathers thanks.
As far as asserting success goes, the success case is that no exceptions are thrown. From what I can tell, in this case asserting there are no exceptions would be redundant. I've scanned through a couple other unit tests and it seems like there are a couple other unit tests without assertions (in iris/tests/unit/cube/test_Cube.py the test test_tolerance_f4() seems to be an example) though I'm still not entirely confident that I know exactly how such cases should be handled.

trexfeathers:
> @trexfeathers thanks.
> As far as asserting success goes, the success case is that no exceptions are thrown. From what I can tell, in this case asserting there are no exceptions would be redundant. I've scanned through a couple other unit tests and it seems like there are a couple other unit tests without assertions (in iris/tests/unit/cube/test_Cube.py the test test_tolerance_f4() seems to be an example) though I'm still not entirely confident that I know exactly how such cases should be handled.

OK, sounds good to me 😊

trexfeathers:
@pp-mo / @lbdreyer could this be merged please?

lib/iris/tests/integration/plot/test_vector_plots.py
```diff
@@ -167,6 +167,26 @@ def test_fail_unsupported_coord_system(self):
...
+                       units='degrees_east')
+        nlat = len(lat.points)
+        nlon = len(lon.points)
+        u_arr = np.ones((nlat, nlon))
+        v_arr = np.ones((nlat, nlon))
+        u_cube = Cube(u_arr, dim_coords_and_dims=[(lat, 0), (lon, 1)],
+                      standard_name='eastward_wind')
+        v_cube = Cube(v_arr, dim_coords_and_dims=[(lat, 0), (lon, 1)],
+                      standard_name='northward_wind')
+        u_cube.coord('longitude').circular = True
```
lbdreyer: Could you also set `circular=True` for v_cube?
I know we don't explicitly do a compatibility check for the u_cube and v_cube as it is still a [TODO](https://github.com/SciTools/iris/blob/master/lib/iris/plot.py#L1301) but once that is fixed, it would throw up an error about the longitude coordinates being different on u_cube and v_cube.

Also, you could do it when you initialise the `lon`, i.e.:
```
lon = DimCoord(np.arange(0, 360, res), 'longitude',
                       units='degrees_east', circular=True)
```
As that would apply to both u_cube and v_cube

lbdreyer:
Great! Thanks @stephenworsley !

Issue #3154, opened 2018-08-31
Don't pin Proj4


pp-mo:
Remove the pin "proj4<5", introduced in #3150

Background:
#3150 fixed problems emerging with cartopy and Proj4 v 5.1
Apparently triggered by new builds on conda_forge (though also anaconda?).
A new build of cartopy 0.16, though no new version.

Needs fixing properly, but probably better done in cartopy ?
Possibly, cartopy may not work properly with Proj4 >= 5.
Which said, the newest build of cartopy 0.16 may be effectively invalid
  - [as here](https://travis-ci.org/SciTools/iris/builds/422933777?utm_source=github_status&utm_medium=notification) , lists as "py27h81b52dc_1  conda-forge",

pp-mo:
UPDATE:
despite relaxing the pin, we still have problem with this, since proj4 v6 came out + [causes a variety of problems]().

So for the future :  **we still need to remove the pin**
( but now it means fixing proj4 v6 compatibility )

trexfeathers:
### This is now a higher priority issue for fixing

#### Why

The proj4<6 pin now prevents Conda installing Iris with Cartopy>=0.18 - compare the installed packages for the [latest Travis build](https://travis-ci.org/github/SciTools/iris/jobs/711355125#L797) against this [experimental unpinned build](https://travis-ci.org/github/bjlittle/iris/jobs/711398545#L838).

Cartopy<0.18 also requires Matplotlib<3.3. So pins begetting more pins, each of which brings extra risk of future tech debt inflation (like what has already happened here).

It is also no longer possible for a pin of this sort to work - Iris' dependencies* are now referencing the [proj package](https://anaconda.org/conda-forge/proj) - this having replaced the [proj4 package](https://anaconda.org/conda-forge/proj4) - and proj cannot be pinned back to the desired point since it 'took over' where proj4 stopped.

#### How

- [x] Thankfully the previously mentioned problems with Cartopy's tests have since been fixed 🎉🏆
- [ ] Fix [TestMappingSubRegion](https://github.com/SciTools/iris/blob/ffcfad475e0593e1e40895453cf1df154e5f6f2c/lib/iris/tests/test_mapping.py#L104) test - [proj4>=6 broke this 2 years ago](https://travis-ci.org/github/SciTools/iris/jobs/422933781#L6721) and [proj still breaks it now](https://travis-ci.org/github/bjlittle/iris/jobs/711398545#L6977).
Might just be that the plot size has changed?
[...]

PR #3374, opened 2019-08-22
Replace version pin on proj4 to permit proj 4 v5.

pp-mo:
Simply replaces #3353, which was over-complicated by a different issue

Relates to (but no longer closes)  #3154 

There is a chance we'll have to exclude proj4 >= 6.0 (newly out I think).
But I think cartopy dependencies should sort that out for us.
Here goes ...

### DIFF
--- a/requirements/core.txt
+++ b/requirements/core.txt
@@ -4,6 +4,7 @@
 # Without these, iris won't even import.
 
 cartopy
+#conda: proj4<6
 cf-units>=2
 cftime
 dask[array]  #conda: dask
--- a/requirements/core.txt
+++ b/requirements/core.txt
@@ -4,7 +4,6 @@
 # Without these, iris won't even import.
 
 cartopy
-#conda: proj4<5
 cf-units>=2
 cftime
 dask[array]  #conda: dask


pp-mo:
**UPDATE**
Tests pass in Python2, where the proj4 version is 5.2.0, and fail in Python3 where proj4 is 6.1.1

Reviewing progress on cartopy, we have a problem.
Although they fixed an initial problem whereby cartopy would not install with Proj4 v6, they did not get the tests working again, [that is still outstanding](https://github.com/SciTools/cartopy/pull/1289#pullrequestreview-272774087)
There too, tests pass in Python2 + proj4 v5, but not Python3 with proj4 v6.
The above discussion implies that outputs have changed with the new proj4 : they are probably "real" graphical changes.

So it seems it is not appropriate to pin proj4 < 6 in cartopy, as I had hoped, which would have made for a quick fix for Iris' sake :disappointed:

pp-mo:
@lbdreyer in view of the above, and a desire to fix this for iris 2.3, do you think it is after all appropriate to pin proj4 to <6 for Iris alone ?

pp-mo:
> pin proj4 to <6 for Iris alone ?

Now done.
Ping @lbdreyer : are you in accord with this ?

Meanwhile I update #3154.  This will no longer close that one.

lbdreyer:
> > pin proj4 to <6 for Iris alone ?
> 
> Now done.
> Ping @lbdreyer : are you in accord with this ?

I am in accord with this approach. We really do, at some point, need to remove the pin, but there are higher priority items that we need to get started on, so I don't want to just yet sink too much time into fixing the tests to work with the latest proj4.
When we do unpin proj4, at the same time we should also get the cartopy tests passing(!) to give us assurance cartopy works with proj4 6.

Issue #3351, opened 2019-07-09
Iris causes ValueError for empty spaces in long_name 


JuliaKukulies:
When using `iris.load_cube(file, constraint) ` to read in a variable of a netCDF4 file which has an empty space in its long_name, iris raises `ValueError: 'substring not found'` when using the long_name as a constraint. The ValueError does not give any  clear message how to solve the problem, which in this case is straight forward: the long_name should not contain any empty spaces or special characters. To solve this problem, one can define a new long_name, e.g. with NCO: 

 `ncatted -a long_name,'varname',o,c,'new_long_name' in.nc out.nc`

When, however, using the variable name as a constraint in `iris.load_cube`,  no ValueError appears until you display the cube. In contrast to #3261, it worked to read in the data based on the variable name (instead of standard_name or long_name)m but any further operations with the cube object fail.

bjlittle:
@JuliaKukulies Awesome, thanks for raising this issue. Much appreciated! :smile: 

Could you provide an example of the constraint that you are using to load your example file, with the traceback, just to clarify.

Many thanks :+1:

JuliaKukulies:
Yes, of course! :) 

The ncdump -h output of the file is: 


`ncdump -h CNRR-GSISAT_hourly_rain_199706.nc4
netcdf CNRR-GSISAT_hourly_rain_199706 {
dimensions:
	time = UNLIMITED ; // (720 currently)
	west_east = 480 ;
	south_north = 360 ;
variables:
	double time(time) ;
		time:standard_name = "time" ;
		time:units = "hours since 1997-1-1 0:0:0" ;
		time:calendar = "standard" ;
		time:axis = "T" ;
	float prcp(time, south_north, west_east) ;
		prcp:long_name = "CNRR-GSISAT hourly rain, 1997" ;
		prcp:units = "mm/hr" ;
	float LAT(south_north, west_east) ;
		LAT:description = "LATITUDE, SOUTH IS NEGATIVE" ;
		LAT:units = "degree_north" ;
	float LON(south_north, west_east) ;
		LON:description = "LONGITUDE, WEST IS NEGATIVE" ;
		LON:units = "degree_east" ;

// global attributes:
		:CDI = "Climate Data Interface version ?? (http://mpimet.mpg.de/cdi)" ;
		:Conventions = "CF-1.6" ;
		:history = "Wed Jul 10 11:00:19 2019: ncks -A -v LON surface_20150112.nc 1997_mon06.nc4\n",
[...]

PR #3373, opened 2019-08-22
Fix bug with html_repr

stephenworsley:
Fixes the bug with #3351.

Previously, if a cube had an attribute which was a string containing '\n', attempting to get the html_repr of that cube would cause an exception.  This pull request would instead add what is in that new line to a new row of the html_repr.
A cube whose string representation looks like:
```
sea_water_potential_temperature / (degC) (-- : 148; -- : 180)
     Auxiliary coordinates:
          latitude                           x         x
          longitude                          x         x
     Scalar coordinates:
          depth: 4.999938 m, bound=(0.0, 10.0) m
          time: 0001-01-01 12:00:00
     Attributes:
          Conventions: CF-1.5
          history: Wed Jul 10 11:00:19 2019: ncks -A -v LON surface_20150112.nc 1997_mon06.nc4
...
     Cell methods:
          mean: time
```
will now have a html_repr which creates a table looking something like:
<table class="iris" id="140242688578000">
    <tr class="iris">
<th class="iris iris-word-cell">Sea Water Potential Temperature (degC)</th>
<th class="iris iris-word-cell">--</th>
<th class="iris iris-word-cell">--</th>
</tr>
    <tr class="iris">
<td class="iris-word-cell iris-subheading-cell">Shape</td>
<td class="iris iris-inclusion-cell">148</td>
<td class="iris iris-inclusion-cell">180</td>
</td>
    <tr class="iris">
    <td class="iris-title iris-word-cell">Auxiliary coordinates</td>
    <td class="iris-title"></td>
    <td class="iris-title"></td>
</tr>
<tr class="iris">
    <td class="iris-word-cell iris-subheading-cell">	latitude</td>
    <td class="iris-inclusion-cell">x</td>
    <td class="iris-inclusion-cell">x</td>
</tr>
<tr class="iris">
    <td class="iris-word-cell iris-subheading-cell">	longitude</td>
    <td class="iris-inclusion-cell">x</td>
    <td class="iris-inclusion-cell">x</td>
</tr>
<tr class="iris">
    <td class="iris-title iris-word-cell">Scalar coordinates</td>
    <td class="iris-title"></td>
    <td class="iris-title"></td>
</tr>
<tr class="iris">
    <td class="iris-word-cell iris-subheading-cell">	depth</td>
    <td class="iris-word-cell" colspan="2">4.999938 m, bound=(0.0, 10.0) m</td>
</tr>
<tr class="iris">
    <td class="iris-word-cell iris-subheading-cell">	time</td>
[...]

PR #3369, opened 2019-08-20
Dress naked excepts #2

pp-mo:
Replaces #3366,
just so we can resolve in @bjlittle absence

Original head
> This PR fixes the iris travis-ci tests after moving to the latest numpy 1.17.0 - and it also follows-up from #3320 to dress the naked try...excepts with an ImportError

Closes #3365  (ref [old comment](https://github.com/SciTools/iris/pull/3366#issuecomment-519952864))

PR #3376, opened 2019-08-23
Catch failed html_repr for cube

DPeterK:
In jupyter, if an enhanced repr (such as `_repr_html_`) fails, the exception is raised but the execution continues, also returning the standard repr. This is useful for noting that there's a problem with the `_repr_html_`, but very bad for UX (as reflected in #3351).

As such, here we catch exceptions raised by constructing the html repr and explicitly return `None` so that the default repr is returned instead.

### DIFF
--- a/lib/iris/cube.py
+++ b/lib/iris/cube.py
@@ -2561,7 +2561,14 @@ bound=(1994-12-01 00:00:00, 1998-12-01 00:00:00)
         from iris.experimental.representation import CubeRepresentation
 
         representer = CubeRepresentation(self)
-        return representer.repr_html()
+        # Catch exceptions in the html repr rather than passing them,
+        # and so fall back to the default cube repr.
+        try:
+            result = representer.repr_html()
+        except (IndexError, ValueError):
+            result = None
+        finally:
+            return result
 
     def __iter__(self):
         raise TypeError("Cube is not iterable")
--- a/lib/iris/tests/unit/cube/test_Cube.py
+++ b/lib/iris/tests/unit/cube/test_Cube.py
@@ -38,7 +38,6 @@ from iris.exceptions import (
     AncillaryVariableNotFoundError,
     UnitConversionError,
 )
-from iris.experimental.representation import CubeRepresentation
 from iris._lazy_data import as_lazy_data
 import iris.tests.stock as stock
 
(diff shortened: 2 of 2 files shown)


lib/iris/cube.py
```diff
@@ -2124,7 +2124,14 @@ def __repr__(self):
     def _repr_html_(self):
         from iris.experimental.representation import CubeRepresentation
         representer = CubeRepresentation(self)
-        return representer.repr_html()
+        # Catch exceptions in the html repr rather than passing them,
+        # and so fall back to the default cube repr.
+        try:
+            result = representer.repr_html()
+        except (IndexError, ValueError):
```
DPeterK: My original preference was to catch any exceptions, but I see that's [bad practice](https://github.com/SciTools/iris/pull/3369). Is this too broad a range of exceptions, too narrow, or just right?

lib/iris/cube.py
```diff
@@ -2124,7 +2124,14 @@ def __repr__(self):
...
+            result = None
```
DPeterK: Should we warn that something's gone wrong? It might be weird to not get an html repr of a cube unexpectedly (my preference is no - Iris is loud enough already).

lib/iris/tests/unit/cube/test_Cube.py
```diff
@@ -1794,5 +1795,21 @@ def test_preserves_lazy(self):
...
+class Test__repr_html(tests.IrisTest):
+    def test_bad_repr(self):
+        # If an html repr fails we don't want it to pass the error.
+        cube = stock.simple_3d()
+        bad_attr = {'history': 'First edit\nSecond edit'}
+        cube.attributes.update(bad_attr)
+        representer = CubeRepresentation(cube)
+
+        # Confirm we get an exception when we make the html repr...
+        with self.assertRaisesRegexp(ValueError, 'substring'):
```
DPeterK: This will stop erroring once #3373 is merged, so we will have to find / cause a different exception in the representer instead...

bjlittle:
@DPeterK Fancy doing a rebase?

trexfeathers:
This seems to have gone stale. We still have ambitions for better HTML, particularly given @pp-mo's work turning the `Cube` summary into actual objects, which would make for much better HTML creation. But that would need a different PR, so I'm going to close this one. @DPeterK feel free to re-open if you disagree.

Issue #3377, opened 2019-08-23
html_repr may include unsanitized html code


stephenworsley:
Relevant to #3313 and #3373.
It is currently possible to construct a cube whose attributes contain strings with arbitrary html in them (for example, `cube.attributes['bad html'] = "...</td><td>---"` would break the formatting of the table). Their html_repr is not currently sanitizing such text. It might be wise to call `html.escape()` on such text when it is used to construct html objects.

pp-mo:
Also fixes #3351
Hence I'll assign 2.3 milestone

pp-mo:
Hi @stephenworsley 
Although #3373 closed, this still needs doing.
Is that right ??

stephenworsley:
@pp-mo yes, this is addressed by #3378 and should be closed by it being merged (though as an issue it would still remain relevant to #3313). I mention #3313 and #3373 because there was a chance that they could introduce code that needed to be addressed by this issue. Though, with the way they are currently written, #3373 and #3378 ought to be compatible.

pp-mo:
Addressed by #3378

PR #3313, opened 2019-05-15
Add repr_html for CubeLists

DPeterK:
As a (long-delayed) follow-on to a previous PR to add cube html repr (see #2918), this adds html repr functionality to `CubeList` objects. The html repr is designed to be reasonably simple, retaining the content of the original `CubeList` repr while adding an accordion effect that, when a cube in the list is clicked, expands to reveal the html repr of that cube.

PR #3378, opened 2019-08-23
Sanitize text before making repr_html

stephenworsley:
Addresses #3377. May be relevant to #3313.
All text which is taken from a cube and put into its repr_html is now passed through `html.escape()` first.


lib/iris/experimental/representation.py
```diff
@@ -23,6 +23,7 @@
 from __future__ import (absolute_import, division, print_function)
 from six.moves import (filter, input, map, range, zip)  # noqa
 
+import html
```
DPeterK: Given we only need to use `escape` from this library, how about `from html import escape as html_escape` instead.

lib/iris/experimental/representation.py
```diff
@@ -206,7 +209,8 @@ def _make_shapes_row(self):
         cells = ['<tr class="iris">', title_cell]
         for shape in self.shapes:
             cells.append(
-                '<td class="iris iris-inclusion-cell">{}</td>'.format(shape))
+                '<td class="iris iris-inclusion-cell">{}</td>'.format(
+                    html.escape(shape)))
```
DPeterK: Definitely don't need to sanitize this.

lib/iris/experimental/representation.py
```diff
@@ -264,6 +268,8 @@ def _make_content(self):
             if v is not None:
                 # Add the sub-heading title.
                 elements.extend(self._make_row(k))
+                # Clean text for html representation
+                v = html.escape(v)
```
DPeterK: Can you instead sanitize the definition of `self.cube_str` in the init please.

lib/iris/experimental/representation.py
```diff
@@ -190,12 +191,14 @@ def _make_header(self):
         """
         # Header row.
         tlc_template = \
-            '<th class="iris iris-word-cell">{self.name} ({self.units})</th>'
-        top_left_cell = tlc_template.format(self=self)
+            '<th class="iris iris-word-cell">{name} ({units})</th>'
+        top_left_cell = tlc_template.format(name=html.escape(self.name),
+                                            units=html.escape(self.units))
```
DPeterK: I'd prefer to see both of these (and each `dim_name` below) sanitized in the init. That way if we end up using them somewhere else in the class in the future we won't have to duplicate these `escape` calls.

stephenworsley:
So it looks like the html module was added to python in version 3.2 so this is not compatible with 2.7. That said, the escape function isn't that complicated as it just involves a couple of string replace operations (In the form i'm using, it could be written in about 5 lines). 
I think I may rewrite the escape function instead of importing it.

DPeterK:
> I think I may rewrite the escape function instead of importing it.

👍 Support this decision (with the expectation that it will be reversed once Iris goes Python 3+ only!)

lbdreyer:
@stephenworsley I've moved this PR back into to the "In progress" column as it looks like this is still requiring tests

bjlittle:
@stephenworsley @lbdreyer Is this PR making the cut for `2.3.0`?

If so, it needs a bit of ❤️ and should now target the `2.3.x` branch, not `master`.

Otherwise, let's target `3.0.0`

lbdreyer:
> @stephenworsley @lbdreyer Is this PR making the cut for `2.3.0`?
> 
> If so, it needs a bit of heart and should now target the `2.3.x` branch, not `master`.
> 
> Otherwise, let's target `3.0.0`

I think it's a bit too late, unfortunately. And I'm a little unsure about the approach. 
I think it is better to wait for Iris 3

stephenworsley:
I've been having a look at how this interacts with #3313. From what I can tell, it's looking somewhat less trivial now, especially with the way the tests for CubeListRepresentation are set up. So I would agree with @lbdreyer here, it's probably one for 3.0.

lib/iris/tests/integration/experimental/test_CubeRepresentation.py
```diff
@@ -136,12 +137,12 @@ def test_identfication(self):
 
     def test_header__name(self):
         header = self.representer._make_header()
-        expected_name = self.cube.name().title().replace('_', ' ')
+        expected_name = escape(self.cube.name().title().replace('_', ' '))
```
lbdreyer: Why did you use `escape` here? self.cube is one we create and shouldn't have a name that needs sanitising

stephenworsley: The thinking here is that I want the process to be consistent with what is happening in the code it is testing, even if `escape` isn't doing anything in this instance.

pp-mo:
Hi @stephenworsley 
All looks good to me, 
though I'm not merging until I'm a bit happier with my own understanding of the original code, which is a bit obscure IMO.

However, I unfortunately found that **none of the tests in `iris.tests.unit.experimental.representation` have actually been running** .
Because it has been missing an `__init__.py` in the containing folder.
Which naturally means they do not pass :face_with_head_bandage: .
Arrrgh.

pp-mo:
Re: https://github.com/stephenworsley/iris/pull/5
I think that fixes the above : please review + merge if you think correct @stephenworsley 

OWOK, I think :+1:

PR #3379, opened 2019-08-23
Rebuilt flatten auxcoords

pp-mo:
**WIP: do not merge**

Cherry-picked the later versions of @duncanwp 's commits from #3030

### DIFF
--- a/lib/iris/util.py
+++ b/lib/iris/util.py
@@ -1545,7 +1545,7 @@ def demote_dim_coord_to_aux_coord(cube, name_or_coord):
         or
 
         (b) the :attr:`standard_name`, :attr:`long_name`, or
-        :attr:`var_name` of an instance of an instance of
+        :attr:`var_name` of an instance of
         :class:`iris.coords.DimCoord`.
 
     For example::
@@ -1596,7 +1596,6 @@ def demote_dim_coord_to_aux_coord(cube, name_or_coord):
 def flatten_multidim_coord(cube, name_or_coord):
     """
     Flatten the Aux coord and associated dimensions on the cube
-
 
     * cube
         An instance of :class:`iris.cube.Cube`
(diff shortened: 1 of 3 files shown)


lib/iris/util.py
```diff
@@ -1593,6 +1593,173 @@ def demote_dim_coord_to_aux_coord(cube, name_or_coord):
...
+    """
+    import numpy as np
+    from iris.cube import Cube
+    from iris.coords import AuxCoord
+
+    all_dims = range(cube.ndim)
+
+    dims = dims if dims is not None else all_dims
+
+    other_dims = list(set(all_dims) - set(dims))
```
pp-mo: This set of numbers needs to be sorted into order : not guaranteed.

pp-mo: This set of numbers needs to be sorted into order : it is not guaranteed.

lib/iris/util.py
```diff
@@ -1593,6 +1593,173 @@ def demote_dim_coord_to_aux_coord(cube, name_or_coord):
...
+    shape_array = np.asarray(cube.shape)
+    # The new (flat) dimension will be in the first dimension of the aux coord
```
pp-mo: not true if dims are not in order, hence dims[0] != min(dims)

lib/iris/util.py
```diff
@@ -1593,6 +1593,173 @@ def demote_dim_coord_to_aux_coord(cube, name_or_coord):
...
+    flat_dim = dims[0]
+    dim_shape = shape_array[list(dims)]
+    new_dim_shape = np.product(dim_shape)
+    new_shape = np.insert(shape_array[other_dims], flat_dim, new_dim_shape)
```
pp-mo: There are problems with this calc, even when the dims are in order.
Examples testing this part of the code, extracted into a standalone function ...
```
def newshape(shape, dims):
   all_dims = range(len(shape))
   shape_array = np.asarray(shape)
   flat_dim = dims[0]
   dim_shape = shape_array[list(dims)]
   new_dim_shape = np.product(dim_shape)
   new_shape = np.insert(shape_array[other_dims], flat_dim, new_dim_shape)
   return new_shape
```
Then
`newshape((10, 11, 12, 13), (0, 2)) --> [120,  11,  13]` is ok, but ..
`newshape((10, 11, 12, 13), (2, 0)) --> [ 11,  13, 120]` is wrong : (reverse dims order) : should be same as previous (??)
Likewise ...
`newshape((10, 11, 12, 13), (1, 2)) --> [ 11, 132,  13]` is wrong : should give  (10, 132, 12)

lib/iris/util.py
```diff
@@ -1593,6 +1593,173 @@ def demote_dim_coord_to_aux_coord(cube, name_or_coord):
...
+
+    new_data = cube.core_data().reshape(new_shape)
+    new_aux_coords = other_aux_coords
+    for c in coords_to_flatten:
+        # The new (flat) AuxCoords are always the last dimension
+        new_bounds = None if c.bounds is None else c.bounds.reshape(-1, 2)
+        # Any DimCoords spanning these dimensions have been demoted so
+        # there should only be AuxCoords left
+        new_aux_coords.append((c.copy(c.points.flat, new_bounds), flat_dim))
+    # TODO: Deal with Aux Factories and cell measures....
```
pp-mo: As stated, I think these aspects need dealing with, or at least to detect and issue an error.

pp-mo:
Given outstanding problems, I think we are better going back to the original PR + waiting for @duncanwp to respond

PR #3380, opened 2019-08-27
Cube HTML repr CSS improvements

DPeterK:
Use CSS variables defined in Jupyterlab for styling custom elements of the cube repr table. The primary advantage of doing this is that the cube html repr table will correctly follow different Jupyterlab themes:

_Light theme:_
Screenshot 2019-08-27 at 12 00 58
_Dark theme:_
Screenshot 2019-08-27 at 12 00 40

Also removed `max-width` attribute of table columns.

### DIFF
--- a/lib/iris/experimental/representation.py
+++ b/lib/iris/experimental/representation.py
@@ -285,7 +285,7 @@ class CubeRepresentation(object):
                 for line in v:
                     # Add every other row in the sub-heading.
                     if k in self.dim_desc_coords:
-                        body = re.findall(r'[\w-]+', line)
+                        body = re.findall(r"[\w-]+", line)
                         title = body.pop(0)
                         colspan = 0
                     elif k in self.two_cell_headers:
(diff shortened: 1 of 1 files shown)


stephenworsley:
If I understand correctly, this would make the _html_repr_ better integrated into Jupyter but would make the _html_repr_ less well integrated into other environments without access to --jp-* css variables. Is there anywhere other than Jupyter where we could reasonably expect _html_repr_ to be used?

DPeterK:
@stephenworsley correct, this is explicitly targetting Jupyter environments, which is the only place that `_repr_html_` could conceivably be called on a Python object, as it's part of the broader Jupyter API. I'm pretty sure this means that alternate notebook environments, such as nteract, won't be impacted by this change.

DPeterK:
Rebase...

trexfeathers:
This seems to have gone stale. We still have ambitions for better HTML, particularly given @pp-mo's work turning the `Cube` summary into actual objects, which would make for much better HTML creation. But that would need a different PR, so I'm going to close this one. @DPeterK feel free to re-open if you disagree.

PR #3384, opened 2019-08-29
Fix naming of cubes from abf/abl files

lbdreyer:
The abf/abl file loader was setting the name of the cubes incorrectly.

On [this page](http://wap.sciencenet.cn/home.php?mod=space&mobile=1&do=blog&id=889834&id=889834) explaining the LAI3g & FPAR3g datasets it states that:
abl <- LAI3g
abf <- FPAR3g
as opposed to what we currently have

### DIFF
--- a/lib/iris/fileformats/abf.py
+++ b/lib/iris/fileformats/abf.py
@@ -1,4 +1,4 @@
-# (C) British Crown Copyright 2012 - 2015, Met Office
+# (C) British Crown Copyright 2012 - 2019, Met Office
 #
 # This file is part of Iris.
 #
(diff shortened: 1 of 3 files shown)


lbdreyer:
@pp-mo, do you mind having a look at this?

lib/iris/fileformats/abf.py
```diff
@@ -118,9 +118,9 @@ def to_cube(self):
 
         # Name.
         if self.format.lower() == "abf":
-            cube.rename("leaf_area_index")
-        elif self.format.lower() == "abl":
             cube.rename("FAPAR")
```
pp-mo: FWIW it took me some searching to convince myself that FPAR and FAPAR are names for the same thing.
But I think I can confirm that now, as [this paper](https://www.frontiersin.org/articles/10.3389/fpls.2018.00776/full) references [this one](https://www.sciencedirect.com/science/article/pii/S0034425798000509?via%3Dihub), apparently assuming they are the same.

PR #3386, opened 2019-08-30
Support volcello loading even if volcello referenced in cell_measures

### LINKED ISSUE #3367: Can't read "volcello" files
As far as I can tell, iris can't read any "volcello" files e.g.

```python
>>> import iris

# e.g. file (I've also tried other CMIP6 volcello files and get the same behaviour)
>>> volcello_path = "CMIP6/CMIP/NCAR/CESM2-WACCM/piControl/r1i1p1f1/Ofx/volcello/gn/v20190320/volcello_Ofx_CESM2-WACCM_piControl_r1i1p1f1_gn.nc"
>>> iris.load_cube(volcello_path)
.../miniconda3/envs/netcdf-scm/lib/python3.7/site-packages/iris/fileformats/cf.py:798: UserWarning: Missing CF-netCDF measure variable 'areacello', referenced by netCDF variable 'volcello'
  warnings.warn(message % (variable_name, nc_var_name))
  File ".../miniconda3/envs/netcdf-scm/lib/python3.7/site-packages/iris/__init__.py", line 377, in load_cube
    cube = cubes.merge_cube()
  File ".../miniconda3/envs/netcdf-scm/lib/python3.7/site-packages/iris/cube.py", line 374, in merge_cube
    raise ValueError("can't merge an empty CubeList")
ValueError: can't merge an empty CubeList

During handling of the above exception, another exception occurred:

  File ".../miniconda3/envs/netcdf-scm/lib/python3.7/site-packages/iris/__init__.py", line 381, in load_cube
    raise iris.exceptions.ConstraintMismatchError('no cubes found')
iris.exceptions.ConstraintMismatchError: no cubes found
```

I did some digging and I think the reason is that "volcello" is being identified as a `iris.fileformats.cf.CFMeasureVariable` and hence does not appear in `iris.fileformats.cf.CFGroup.data_variables` (which matters here https://github.com/SciTools/iris/blob/5d4976614d39a93c12e640ff5a2ec1ccdd6e8380/lib/iris/fileformats/netcdf.py#L709)

If someone can suggest a fix, I'm happy to try and implement it (alongside a test). I just have no idea what the logic sitting underneath all the CF stuff is so can't work out how to start.

znicholls:
Closes #3367

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -33,7 +33,7 @@ git:
 
 install:
   - >
-    export IRIS_TEST_DATA_REF="5d4eab1b4b3f6ef9ede5c0d24853c902f3613fea";
+    export IRIS_TEST_DATA_REF="1696ac3a823a06b95f430670f285ee97671d2cf2";
     export IRIS_TEST_DATA_SUFFIX=$(echo "${IRIS_TEST_DATA_REF}" | sed "s/^v//");
 
   # Install miniconda
--- a/.travis.yml
+++ b/.travis.yml
@@ -33,7 +33,7 @@ git:
 
 install:
   - >
-    export IRIS_TEST_DATA_REF="dba47566a9147645fea586f94a138e0a8d45a48e";
+    export IRIS_TEST_DATA_REF="5d4eab1b4b3f6ef9ede5c0d24853c902f3613fea";
     export IRIS_TEST_DATA_SUFFIX=$(echo "${IRIS_TEST_DATA_REF}" | sed "s/^v//");
 
   # Install miniconda
--- a/lib/iris/tests/integration/test_netcdf.py
+++ b/lib/iris/tests/integration/test_netcdf.py
@@ -276,6 +276,33 @@ class TestCellMeasures(tests.IrisTest):
 
 class TestCMIP6VolcelloLoad(tests.IrisTest):
     def setUp(self):
+        self.fname = tests.get_data_path(
+            ('NetCDF', 'volcello',
+             'volcello_Ofx_CESM2_deforest-globe_r1i1p1f1_gn.nc'))
+
+    def test_cmip6_volcello_load_issue_3367(self):
+        # Ensure that reading a file which references itself in
+        # `cell_measures` can be read. At the same time, ensure that we
+        # still receive a warning about other variables mentioned in
+        # `cell_measures` i.e. a warning should be raised about missing
+        # areacello.
+        areacello_str = "areacello" if six.PY3 else u"areacello"
+        volcello_str = "volcello" if six.PY3 else u"volcello"
+        expected_msg = "Missing CF-netCDF measure variable %r, " \
+                       "referenced by netCDF variable %r" \
+                       % (areacello_str, volcello_str)
+
+        with mock.patch('warnings.warn') as warn:
+            # ensure file loads without failure
+            cube = iris.load_cube(self.fname)
+            warn.assert_has_calls([mock.call(expected_msg)])
+
+        # extra check to ensure correct variable was found
+        assert cube.standard_name == 'ocean_volume'
+
+
+class TestSelfReferencingVarLoad(tests.IrisTest):
+    def setUp(self):
         self.temp_dir_path = os.path.join(tempfile.mkdtemp(),
                                           'issue_3367_volcello_test_file.nc')
         dataset = nc.Dataset(self.temp_dir_path, 'w')
@@ -317,7 +344,7 @@ class TestCMIP6VolcelloLoad(tests.IrisTest):
 
         dataset.close()
 
-    def test_cmip6_volcello_load_issue_3367(self):
+    def test_self_referencing_load_issue_3367(self):
         # Ensure that reading a file which references itself in
         # `cell_measures` can be read. At the same time, ensure that we
         # still receive a warning about other variables mentioned in
--- a/lib/iris/tests/integration/test_netcdf.py
+++ b/lib/iris/tests/integration/test_netcdf.py
@@ -274,6 +274,7 @@ class TestCellMeasures(tests.IrisTest):
                          '    x         x') in printed)
 
 
+@tests.skip_data
 class TestCMIP6VolcelloLoad(tests.IrisTest):
     def setUp(self):
         self.fname = tests.get_data_path(
--- a/lib/iris/tests/integration/test_netcdf.py
+++ b/lib/iris/tests/integration/test_netcdf.py
@@ -18,6 +18,7 @@
 
 from __future__ import (absolute_import, division, print_function)
 from six.moves import (filter, input, map, range, zip)  # noqa
+import six
 
 # Import iris.tests first so that some things can be initialised before
 # importing anything else.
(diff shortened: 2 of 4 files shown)


znicholls:
@pp-mo i think your solution has worked! If you have a minute to have a quick look that’d be great, I’ll add test files during the week

lib/iris/tests/test_netcdf.py
```diff
@@ -910,6 +910,65 @@ def test_std_name_lookup_fail(self):
         self.assertFalse('phenomenon_time' in iris.std_names.STD_NAMES)
 
 
+class TestCMIP6VolcelloLoad(tests.IrisTest):
+    def setUp(self):
+        dataset = nc.Dataset('issue_3367_volcello_test_file.nc', 'w')
```
pp-mo: I like the self-generated testfile,
but I think it's worth creating the file in a "standard" temporary file location, rather than current directory.
That works in case you are running tests from a protected directory (like a system install).

If you pre-create a cube (instead of a file) in the setUp then you can use 
`with self.temp_filename('.nc'):`   (that is, `IrisTest.temp_filename`), and then just save it to file in the test function before reading it back.
I guess that's ok if there's only one testcase, but wasteful if there are more, because the file is always destroyed afterward.  
Another way is to explicitly 'enter' and 'exit' the context manager, instead of the 'with', which can then be done in setUp and tearDown, like the present code.
Or, you can just use `tempfile.mkdtemp()` instead.  [Like this case](https://github.com/SciTools/iris/blob/master/lib/iris/tests/integration/fast_load/test_fast_load.py#L57)

Do whatever seems nicer to you.

lib/iris/tests/test_netcdf.py
```diff
@@ -910,6 +910,65 @@ def test_std_name_lookup_fail(self):
...
+    def test_cmip6_volcello_load_issue_3367(self):
+        # a warning should be raised about missing areacello
+        areacello_str = "areacello" if six.PY3 else u"areacello"
+        volcello_str = "volcello" if six.PY3 else u"volcello"
+        expected_msg = "Missing CF-netCDF measure variable %r, " \
+                       "referenced by netCDF variable %r" \
+                       % (areacello_str, volcello_str)
+
+        with mock.patch('warnings.warn') as warn:
+            cube = iris.load_cube('issue_3367_volcello_test_file.nc')
```
pp-mo: It's at least worth a comment here what the basic point of the test is
-- because the basic point is that the code did not fail, which is never that obvious !

Especially as the warning you're handling here is not the main point : I'm not even sure it's worth doing that, you could just let it happen, or maybe just mask it with warnings.catch_warnings / warnings.filterwarnings.

lib/iris/tests/test_netcdf.py
```diff
@@ -1,4 +1,4 @@
-# (C) British Crown Copyright 2010 - 2018, Met Office
+# (C) British Crown Copyright 2010 - 2019, Met Office
```
pp-mo: After some discussion with @lbdreyer , I think we're agreed that we'd rather not add more to this sourcefile.
The reason is that the files `lib/iris/tests/test*.py` are all basically ancient legacy code, and we'd rather have all testing either under `lib/iris/tests/integration`, or in appropriate sub-directories of `lib/iris/tests/unit`.
( ultimately it would be nice to tidy up the entire structure, but we never seem to find the time! )

In this case, I think you could usefully put this code in `lib/iris/tests/integration/test_netcdf.py` instead.
Is that ok ?   I couldn't see any technical difficulty with it.

znicholls:
Thanks for the feedback @pp-mo ! I think all that is addressed now. When you're happy, I can squash and rebase against master.

znicholls:
@pp-mo if you get a minute to look at this that would be awesome

pp-mo:
> I'll get moving on adding a real test file now (should I add that in this PR or in a separate one and how do I work out the updated iris test data SHA).

I assume we're talking about https://github.com/SciTools/iris-test-data/pull/53
The process is...
  * [ ] get https://github.com/SciTools/iris-test-data/pull/53 merged
  * [ ] get the SHA which is then the latest commit in the master branch of iris-test-data
    * (like the current latest commit is  https://github.com/SciTools/iris-test-data/commit/dba47566a9147645fea586f94a138e0a8d45a48e, so the SHA is "dba47566a9147645fea586f94a138e0a8d45a48e")
  * [ ] change the SHA where it include [in the .travis.yml](https://github.com/SciTools/iris/blob/master/.travis.yml#L36) on this branch ...
    * ... the Travis re-test ~should~ will then pass all the tests !
  * [ ] then I'll merge this too.

*But I'm puzzled ... *
***Why is there no test here that requires that new file in iris-test-data (and should therefore be failing) ??***

pp-mo:
Also... tiny point, but [the above stickler comment](https://github.com/SciTools/iris/pull/3386/files#r319347299) points out that [this line in cf.py](https://github.com/SciTools/iris/blob/52aa986ab3cf052f04dc85e9ff028f7e11ba3c3e/lib/iris/fileformats/cf.py#L798) is still over-length.
I know this file is excluded from the Iris codestyle check, but it's nicer to enable the big green tick on the PR.

znicholls:
> I assume we're talking about [SciTools/iris-test-data#53](https://github.com/SciTools/iris-test-data/pull/53)

Yep so merging that would be great.

> Why is there no test here that requires that new file in iris-test-data (and should therefore be failing) ??

I hadn't got my head around how to make the test data machinery work yet so just left things without that file to make sure everything else got the green Travis tick. Will add that now then should be good to go.

znicholls:
I've added the new test so travis is failing because it can't find the file it needs. I've verified locally that it works so I think we should be more or less good to go.

lib/iris/tests/integration/test_netcdf.py
```diff
@@ -273,6 +274,100 @@ def test_print(self):
                          '    x         x') in printed)
 
 
+class TestCMIP6VolcelloLoad(tests.IrisTest):
```
pp-mo: These tests need to be qualified with `@tests.skip_data`
So, any tests that get test data (via tests.get_data_path) *must* have `@skip_data'.

This is why the 'TEST_MINIMAL' entries in the Travis matrix are failing
 - which should never happen, as 'minimal' tests are a strict subset of all tests
 - the idea is basically just that the 'minimal' tests run quicker (and part of that is not installing iris-test-data)

.travis.yml
```diff
@@ -33,7 +33,7 @@ git:
 
 install:
   - >
-    export IRIS_TEST_DATA_REF="dba47566a9147645fea586f94a138e0a8d45a48e";
+    export IRIS_TEST_DATA_REF="5d4eab1b4b3f6ef9ede5c0d24853c902f3613fea";
```
pp-mo: N.B. ok, but this will need changing *again* when we merge https://github.com/SciTools/iris-test-data/pull/53,
because this commit is on your branch.

Stop press: 
https://github.com/SciTools/iris-test-data/pull/53 **now merged !**
New master commit SHA  is `1696ac3a823a06b95f430670f285ee97671d2cf2`
So that's what is now needed here ...

lib/iris/tests/integration/test_netcdf.py
znicholls: smart, fixed now

pp-mo:
Brilliant, thanks @znicholls
Sorry it's been so fiddly. That nearly always happens !

znicholls:
No worries thanks for the help!

PR #3389, opened 2019-09-02
Support Standard Name Modifiers

lbdreyer:
CF conventions support a set of standard_name modifiers.  See [3.3. Standard Name](http://cfconventions.org/Data/cf-conventions/cf-conventions-1.7/cf-conventions.html#standard-name) for more details, specifically:

> A standard name is associated with a variable via the attribute standard_name which takes a string value comprised of a standard name optionally followed by one or more blanks and a standard name modifier (a string value from Appendix C, Standard Name Modifiers).

This PR modifies the CFVariableMixin standard_name setter (and the pyke rules) to support standard name modifiers. 

I found there wasn't much testing of cube/coord names so I have split this PR into two commits:
1. The first commit has no functional changes. It just adds testing of the pyke rules to show that they don't break later when I add the functional changes
2. The second commit adds the changes to include the standard_name modifiers

I hope that is clear!

### DIFF
--- a/lib/iris/fileformats/_pyke_rules/fc_rules_cf.krb
+++ b/lib/iris/fileformats/_pyke_rules/fc_rules_cf.krb
@@ -1075,6 +1075,7 @@ fc_extras
     import iris.exceptions
     import iris.std_names
     import iris.util
+    from iris._cube_coord_common import get_valid_standard_name
     from iris._lazy_data import as_lazy_data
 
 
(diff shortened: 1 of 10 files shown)


lib/iris/fileformats/cf.py
```diff
@@ -74,6 +74,12 @@
                        ocean_s_coordinate_g1=['eta', 'depth'],
                        ocean_s_coordinate_g2=['eta', 'depth'])
 
+# Supported standard name modifiers and their associated units. Ref: [CF]
+# Appendix E.
+STD_NAME_MODIFIERS_UNITS = {'detection_minimum': 'u',
+                            'number_of_observations': '1',
+                            'standard_error': 'u',
+                            'status_flag': '1'}
```
lbdreyer: I'm unsure about this unit. The [StandardModifier table in the cf conventions](http://cfconventions.org/cf-conventions/cf-conventions.html#standard-name-modifiers) suggests there shouldn't be a unit. However, it also states that the `status_flag` modifier has been deprecated and should be replaced with a standard_name of status_flag which does have a unit of '1' (search for "status_flag" in the [cf-standard-name-table.xml](https://raw.githubusercontent.com/SciTools/iris/master/etc/cf-standard-name-table.xml))

lib/iris/fileformats/_pyke_rules/fc_rules_cf.krb
```diff
@@ -1536,7 +1536,6 @@ fc_extras
                         standard_name = coord_name
                     else:
                         standard_name = None
-                        long_name = standard_name
```
lbdreyer: long_name is already None so this line doesn't do anything - so I removed to make this code mode clear

lbdreyer:
Closing as the handling of units is a little confused! Some variables with std name modifiers need units but some don't (e.g. status_flags). Also the testing needs some work
I'll reopen this change as part of a different PR

PR #3393, opened 2019-09-09
Support Standard Name Modifers (not units!)

lbdreyer:
CF conventions support a set of standard_name modifiers.  See [3.3. Standard Name](http://cfconventions.org/Data/cf-conventions/cf-conventions-1.7/cf-conventions.html#standard-name) for more details, specifically:

> A standard name is associated with a variable via the attribute standard_name which takes a string value comprised of a standard name optionally followed by one or more blanks and a standard name modifier (a string value from Appendix C, Standard Name Modifiers).

This PR modifies the CFVariableMixin standard_name setter (and the pyke rules) to support standard name modifiers. 

I found there wasn't much testing of cube/coord names so I have split this PR into two commits:
1. The first commit has no functional changes. It just adds testing of the pyke rules to show that they don't break later when I add the functional changes
2. The second commit adds the changes to include the standard_name modifiers

Note that standard name modifiers have associated units (See [CF Appendix C](http://cfconventions.org/Data/cf-conventions/cf-conventions-1.7/cf-conventions.html#standard-name-modifiers)) however I have not included that in this PR as I am only modifying the standard_name setter which at the moment doesn't have any requirement on the unit

### DIFF
--- a/lib/iris/_cube_coord_common.py
+++ b/lib/iris/_cube_coord_common.py
@@ -34,22 +34,30 @@ _TOKEN_PARSE = re.compile(r'''^[a-zA-Z0-9][\w\.\+\-@]*$''')
 def get_valid_standard_name(name):
     # Standard names are optionally followed by a standard name
     # modifier, separated by one or more blank spaces
-    name_parts = name.split(' ')
-    std_name, std_name_extension = name_parts[0], name_parts[1:]
 
-    error_msg = '{!r} is not a valid standard_name'.format(name)
+    if name is not None:
+        name_is_valid = False
+        # Supported standard name modifiers. Ref: [CF] Appendix C.
+        valid_std_name_modifiers = ['detection_minimum',
+                                    'number_of_observations',
+                                    'standard_error',
+                                    'status_flag']
 
-    if std_name in iris.std_names.STD_NAMES:
-        if std_name_extension:
-            standard_name_modifier = std_name_extension[-1]
-            # Check only blank spaces before valid standard_name_modifier
-            if (any(item for item in std_name_extension[:-1]) or
-                    standard_name_modifier not in
-                    iris.fileformats.cf.STD_NAME_MODIFIERS):
-                raise ValueError(error_msg)
-        return name
-    else:
-        raise ValueError(error_msg)
+        valid_name_pattern = re.compile(r'''^([a-zA-Z_]+)( *)([a-zA-Z_]*)$''')
+        name_groups = valid_name_pattern.match(name)
+
+        if name_groups:
+            std_name, whitespace, std_name_modifier = name_groups.groups()
+            if (std_name in iris.std_names.STD_NAMES) and (
+                bool(whitespace) == (std_name_modifier in
+                                     valid_std_name_modifiers)):
+                name_is_valid = True
+
+        if name_is_valid is False:
+            raise ValueError('{!r} is not a valid standard_name'.format(
+                    name))
+
+    return name
 
 
 class LimitedAttributeDict(dict):
@@ -195,10 +203,7 @@ class CFVariableMixin(object):
 
     @standard_name.setter
     def standard_name(self, name):
-        if name is None:
-            self._standard_name = name
-        else:
-            self._standard_name = get_valid_standard_name(name)
+        self._standard_name = get_valid_standard_name(name)
 
     @property
     def units(self):
--- a/lib/iris/_cube_coord_common.py
+++ b/lib/iris/_cube_coord_common.py
@@ -31,6 +31,27 @@ import iris.std_names
 _TOKEN_PARSE = re.compile(r'''^[a-zA-Z0-9][\w\.\+\-@]*$''')
 
 
+def get_valid_standard_name(name):
+    # Standard names are optionally followed by a standard name
+    # modifier, separated by one or more blank spaces
+    name_parts = name.split(' ')
+    std_name, std_name_extension = name_parts[0], name_parts[1:]
+
+    error_msg = '{!r} is not a valid standard_name'.format(name)
+
+    if std_name in iris.std_names.STD_NAMES:
+        if std_name_extension:
+            standard_name_modifier = std_name_extension[-1]
+            # Check only blank spaces before valid standard_name_modifier
+            if (any(item for item in std_name_extension[:-1]) or
+                    standard_name_modifier not in
+                    iris.fileformats.cf.STD_NAME_MODIFIERS):
+                raise ValueError(error_msg)
+        return name
+    else:
+        raise ValueError(error_msg)
+
+
 class LimitedAttributeDict(dict):
     _forbidden_keys = ('standard_name', 'long_name', 'units', 'bounds', 'axis',
                        'calendar', 'leap_month', 'leap_year', 'month_lengths',
@@ -174,10 +195,10 @@ class CFVariableMixin(object):
 
     @standard_name.setter
     def standard_name(self, name):
-        if name is None or name in iris.std_names.STD_NAMES:
+        if name is None:
             self._standard_name = name
         else:
-            raise ValueError('%r is not a valid standard_name' % name)
+            self._standard_name = get_valid_standard_name(name)
 
     @property
     def units(self):
--- a/lib/iris/fileformats/_pyke_rules/fc_rules_cf.krb
+++ b/lib/iris/fileformats/_pyke_rules/fc_rules_cf.krb
@@ -1075,6 +1075,7 @@ fc_extras
     import iris.exceptions
     import iris.std_names
     import iris.util
+    from iris._cube_coord_common import get_valid_standard_name
     from iris._lazy_data import as_lazy_data
 
 
@@ -1186,13 +1187,16 @@ fc_extras
         cube.long_name = long_name
 
         if standard_name is not None:
-            if standard_name in iris.std_names.STD_NAMES:
-                cube.standard_name = standard_name
-            else:
-                if cube.long_name is not None:
-                    cube.attributes['invalid_standard_name'] = standard_name
+            try:
+                cube.standard_name = get_valid_standard_name(standard_name)
+            except ValueError as e:
+                if str(e) != '{!r} is not a valid standard_name'.format(standard_name):
+                    raise
                 else:
-                    cube.long_name = standard_name
+                    if cube.long_name is not None:
+                        cube.attributes['invalid_standard_name'] = standard_name
+                    else:
+                        cube.long_name = standard_name
 
         # Determine the cube units.
         attr_units = get_attr_units(cf_var, cube.attributes)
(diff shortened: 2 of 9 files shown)


lbdreyer:
Ping @bjlittle !

bjlittle:
@lbdreyer Okay, I think that I'm pretty much there...

As it happens, there wasn't any unit test coverage of `CellMethods` (there is for parsing, but not construction and usage), but I've added the minimal test coverage for the changes in this PR.

Being a good citizen, I'll add test coverage in a separate PR.

I've also not updated the docs either, just so that this PR can get banked sooner rather than later, but happy to make a follow-up PR, if you think that's appropriate... :smile:

lib/iris/fileformats/_pyke_rules/fc_rules_cf.krb
```diff
@@ -1523,20 +1527,25 @@ fc_extras
...
+                        else:
+                            standard_name = None
                     else:
-                        standard_name = None
-                        long_name = standard_name
+                        if coord_name is not None:
+                            attributes['invalid_standard_name'] = standard_name
+                            standard_name = coord_name
+                        else:
+                            standard_name = None
```
bjlittle: The previous behaviour also set the `long_name = standard_name` here i.e.,
```python
else:
    standard_name = None
    long_name = standard_name
```
This is still relevant, right?

lib/iris/tests/integration/test_netcdf.py
```diff
@@ -429,5 +429,15 @@ def test_scalar_cube_save_load(self):
...
 
 
+class TestStandardName(tests.IrisTest):
+    def test_standard_name_roundtrip(self):
+        standard_name = 'air_temperature detection_minimum'
+        cube = iris.cube.Cube(1, standard_name=standard_name)
+        with self.temp_filename(suffix='.nc') as fout:
+            iris.save(cube, fout)
+            detection_limit_cube = iris.load_cube(fout)
+            self.assertEqual(detection_limit_cube.standard_name, standard_name)
```
bjlittle: Nice :+1:

lib/iris/tests/unit/cube_coord_common/test_CFVariableMixin.py
```diff
@@ -0,0 +1,48 @@
...
+# importing anything else.
+import iris.tests as tests
+
+from iris._cube_coord_common import CFVariableMixin
+
+
+class Test_standard_name__setter(tests.IrisTest):
+    def test_valid_standard_name(self):
+        cf_var = CFVariableMixin()
+        cf_var.standard_name = 'air_temperature'
```
bjlittle: Shouldn't you then `self.assertEqual(cf_var.standard_name, 'air_temperature')` to ensure that we get back what we set?

lib/iris/tests/unit/cube_coord_common/test_CFVariableMixin.py
```diff
@@ -0,0 +1,48 @@
...
+
+    def test_invalid_standard_name(self):
+        cf_var = CFVariableMixin()
+        emsg = "'not_a_standard_name' is not a valid standard_name"
+        with self.assertRaisesRegexp(ValueError, emsg):
+            cf_var.standard_name = 'not_a_standard_name'
+
+    def test_none_standard_name(self):
+        cf_var = CFVariableMixin()
+        cf_var.standard_name = None
```
bjlittle: Also, include a `self.assertIsNone(cf_var.standard_name)`?

lib/iris/tests/unit/fileformats/pyke_rules/compiled_krb/fc_rules_cf_fc/test_build_cube_metadata.py
```diff
@@ -35,33 +35,33 @@
...
-            cell_methods=None,
-            cf_group=cf_group)
+    engine = mock.Mock(
+        cube=Cube([23]),
+        cf_var=cf_var)
 
-        engine = mock.Mock(
-            cube=Cube([23]),
-            cf_var=cf_var)
+    return engine
```
bjlittle: Wondered why you pulled this out of `TestInvalidGlobalAttributes`, but you're also using it in the new test class `TestCubeName`... nice :+1:

bjlittle:
@lbdreyer What's the strategy for checking the `units` of a `standard_name` with a standard name modifier?

Separate PR?

lib/iris/tests/unit/fileformats/pyke_rules/compiled_krb/fc_rules_cf_fc/test_get_names.py
```diff
@@ -0,0 +1,258 @@
...
+        self.check_names(inp, exp)
+
+    def test_std_name_invalid_long_name_set_var_name_invalid_coord_name_set(
+            self):
+        # standard_name is not a valid standard name, long_name is not None,
+        # var_name is not a valid standard name, coord_name is set.
+        inp = ('latitude_coord', 'lat_long_name', 'lat_var_name', 'latitude')
+        exp = ('latitude', 'lat_long_name', 'lat_var_name',
+               {'invalid_standard_name': 'latitude_coord'})
+        self.check_names(inp, exp)
```
bjlittle: @lbdreyer Whoa... nice test coverage :tada:

lib/iris/_cube_coord_common.py
```diff
@@ -28,6 +28,27 @@
 import iris.std_names
 
 
+def get_valid_standard_name(name):
+    # Standard names are optionally followed by a standard name
+    # modifier, separated by one or more blank spaces
+    name_parts = name.split(' ')
```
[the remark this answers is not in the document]
lbdreyer: If `name` has leading or trailing white space it should cause issues, I think, as then the name is not valid.
```
bad_name = ' air_temperature'
>>> bad_name.split(' ')
['', 'air_temperature']
```
should raise an error, as should 
```
>>> bad_name = 'air_temperature '
>>> bad_name.split(' ')
['air_temperature', '']
```

lib/iris/fileformats/_pyke_rules/fc_rules_cf.krb
```diff
@@ -1186,13 +1187,16 @@ fc_extras
...
+                cube.standard_name = get_valid_standard_name(standard_name)
+            except ValueError as e:
+                if str(e) != '{!r} is not a valid standard_name'.format(standard_name):
+                    raise
                 else:
-                    cube.long_name = standard_name
+                    if cube.long_name is not None:
+                        cube.attributes['invalid_standard_name'] = standard_name
+                    else:
+                        cube.long_name = standard_name
```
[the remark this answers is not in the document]
lbdreyer: No? The diff that is being displayed is makign this a bit hard to read, but basically what was previously lines 1527-1538 
https://github.com/SciTools/iris/blob/ec46449b9705789268b77089a2161ff2c1d605c5/lib/iris/fileformats/_pyke_rules/fc_rules_cf.krb#L1527-L1539
https://github.com/SciTools/iris/blob/b6775634eccbe29a45519c85a0da99a36dbc4886/lib/iris/fileformats/_pyke_rules/fc_rules_cf.krb#L1536-L1547
But they have been indented out once more because they are in the else of the except block.

One change I have made is I have removed line 1539 in the old code (see the last line in the first snippet above) because this line seemed pointless? It was setting long_name to be None even though it already was None?

lib/iris/fileformats/_pyke_rules/fc_rules_cf.krb
lbdreyer: I removed it as I don't see the point in that line.
Just looking at the old code (you have to scroll the below snippet to see the last line):
https://github.com/SciTools/iris/blob/ec46449b9705789268b77089a2161ff2c1d605c5/lib/iris/fileformats/_pyke_rules/fc_rules_cf.krb#L1527-L1539
One line 1527 it says
`if long_name is not None`
so line 1539 is within the `else` of that statement, implying that `long_name=None`. Which means that line 1539  was setting long_name to be None, but it already was None...

lib/iris/fileformats/_pyke_rules/fc_rules_cf.krb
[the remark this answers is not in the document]
lbdreyer: I don't believe so?
What were lines 1192-1195 in the old code
https://github.com/SciTools/iris/blob/ec46449b9705789268b77089a2161ff2c1d605c5/lib/iris/fileformats/_pyke_rules/fc_rules_cf.krb#L1192-L1195
Are now lines 1196-1199 in the new code
https://github.com/SciTools/iris/blob/b6775634eccbe29a45519c85a0da99a36dbc4886/lib/iris/fileformats/_pyke_rules/fc_rules_cf.krb#L1196-L1199

so no change, just an extra indentation due to being in the `else` of the `if`statement in the `except` part

lib/iris/tests/unit/cube_coord_common/test_CFVariableMixin.py
lbdreyer: I guess my concern with that was that we are then testing the standard_name getter and setter at the same time, which doesn't feel very unit-testy but I guess we should test not just that setting the standard doesn't raise errors, but also that it is done correctly and it's hard to do that in isolation from testing the getter.

lib/iris/_cube_coord_common.py
```diff
@@ -28,6 +28,27 @@
...
+
+    if std_name in iris.std_names.STD_NAMES:
+        if std_name_extension:
+            standard_name_modifier = std_name_extension[-1]
+            # Check only blank spaces before valid standard_name_modifier
+            if (any(item for item in std_name_extension[:-1]) or
+                    standard_name_modifier not in
+                    iris.fileformats.cf.STD_NAME_MODIFIERS):
+                raise ValueError(error_msg)
+        return name
```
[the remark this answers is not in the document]
lbdreyer: Yeah I agree, it would be nicer.I just couldn't quite think through how to do that, but I will give it another look!

lib/iris/fileformats/_pyke_rules/fc_rules_cf.krb
bjlittle: Yup, agreed :+1:

lib/iris/_cube_coord_common.py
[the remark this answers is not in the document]
bjlittle: @lbdreyer I think that all you need to do is simply move the `return name` outside the `if ... else ...`, as your not changing the contents of `name`, the output of the function is just the same as the input.

If the input `std_name` (from `name`) is valid, then it will pass through the `if ...`, and then if the `standard_name_modifier` is valid it will simply drop through and `return name`. On the otherhand, if the modifier is not valid an exception is raised and the `return name` is never be reached.

If the input `name` is invalid, then it will pass through to the `else ...` where an exception is raised and the `return name` will never be reached.

Should just work, right? Check my logic is right, though...

lbdreyer:
> @lbdreyer What's the strategy for checking the `units` of a `standard_name` with a standard name modifier?
> 
> Separate PR?

I have raised this in a separate issue: https://github.com/SciTools/iris/issues/3408

lib/iris/fileformats/_pyke_rules/fc_rules_cf.krb
```diff
@@ -1186,13 +1187,16 @@ fc_extras
...
-            if standard_name in iris.std_names.STD_NAMES:
-                cube.standard_name = standard_name
-            else:
-                if cube.long_name is not None:
-                    cube.attributes['invalid_standard_name'] = standard_name
+            try:
+                cube.standard_name = get_valid_standard_name(standard_name)
+            except ValueError as e:
+                if str(e) != '{!r} is not a valid standard_name'.format(standard_name):
+                    raise
```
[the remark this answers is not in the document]
lbdreyer: My concern is that with the `try ... except`_any_ ValueError would be ignored. However, I suppose that any issues with setting the standard_name, e.g an issue with the regexp matching (in the upcoming commit "review actions") should be skipped and go on to try to set the long name instead

lbdreyer:
@bjlittle I believe I have addressed your review comments in the third commit ["review actions"](https://github.com/SciTools/iris/pull/3393/commits/101dc5cb2f2a9a35721eb119757ddc623ab167dd)

Note: I had to rebase the first two commits due a conflict that was caused when #3399 was merged, but those commits should be the same as they were before.

bjlittle:
@lbdreyer Awesome, thanks! :tada:

Issue #3323, opened 2019-06-05
Support climatological statistics


bjlittle:
Provide end-to-end support for [climatological statistics](http://cfconventions.org/Data/cf-conventions/cf-conventions-1.7/cf-conventions.html#climatological-statistics) within `iris`.

Reference:
- https://github.com/SciTools/iris/issues/2904
- https://github.com/SciTools/iris/pull/2905

bjlittle:
Ping @zklaus :smile:

abooton:
Related to https://github.com/SciTools/iris/issues/1438

trexfeathers:
Minimum implementation will be to allow Iris to load and save climatological statistics in NetCDF, even if there is not yet any specific functionality while the data is in a cube. Can achieve this by:
- adding a new parameter to the coordinate class specifying if the bounds are climatological or not
- adding more code to specify the type during the creation of bounds
- get_cf_bounds_var needs to return the bound type
- build_dimension_coordinate needs to supply the bound type when using the coordinate class

Will also need a 'round trip' test. Have a look at the AssertCDL test

pp-mo:
> Some sample metadata to look at

E.G. [this "gridfile...Daily_Climatology_1982_2008_DayNightCombined-v02.0-fv01.0.cdl"](https://data.nodc.noaa.gov/testdata/netCDFTemplateExamples/grid/00000110200000-NODC-L4_GHRSST-SSTskin-AVHRR_Pathfinder-PFV5.0_Daily_Climatology_1982_2008_DayNightCombined-v02.0-fv01.0.cdl)
Yes?

bjlittle:
Closed by https://github.com/SciTools/iris/pull/3395

PR #3395, opened 2019-09-11
Support climatological bounds for NetCDF Save and Load

trexfeathers:
Addresses #3323


trexfeathers:
Still need to write a what's new for this

trexfeathers:
Ready for us to merge, @pp-mo

jkettleb:
Great to see this beeing worked on.  Thanks for finding time to do it.

lbdreyer:
Just had a quick check through this to see if there would be any conflicts with the redesign of the Cube dimensional metadata structure, but it looks like it should all be fine.

I did however notice that `bounds_are_climatological` is not mentioned in the xml_element which seems a bit strange as it is included in `Coord._repr_other_metadata()`. Was this done for a particular reason?

trexfeathers:
> Just had a quick check through this to see if there would be any conflicts with the redesign of the Cube dimensional metadata structure, but it looks like it should all be fine.
> I did however notice that bounds_are_climatological is not mentioned in the xml_element which seems a bit strange as it is included in Coord._repr_other_metadata(). Was this done for a particular reason?

No particular reason, this is poor work on my part. I wasn't aware of xml_element and that I might need to add to it. Seems like it should be done, I'll take a look now

lib/iris/coords.py
```diff
@@ -55,7 +55,8 @@
 class CoordDefn(namedtuple('CoordDefn',
                            ['standard_name', 'long_name',
                             'var_name', 'units',
-                            'attributes', 'coord_system'])):
+                            'attributes', 'coord_system',
+                            'bounds_are_climatological'])):
```
bjlittle: @trexfeathers A small thing, but a big thing...

The implication of supporting the CF `climatology` attribute simply means that the `bounds` are climatological - that's it.

Therefore, I don't think that this needs to be explicitly qualified here. Would you consider changing `bounds_are_climatological` to simply `climatology`?

If `climatology` is `True`, therefore the `bounds` are *special*, otherwise not. I find it a hard sell to force users to use `bounds_are_climatological`

trexfeathers:
Latest change now needs fixes to various CML files

lib/iris/coords.py
```diff
@@ -85,6 +86,8 @@ def _sort_key(defn):
                     defn.long_name is not None, defn.long_name,
                     defn.var_name is not None, defn.var_name,
                     defn.units is not None, defn.units,
+                    # Should attributes be included here?
```
bjlittle: @trexfeathers Rather than have a comment here, can you make an associated issue? Thanks

lib/iris/coords.py
```diff
@@ -85,6 +86,8 @@ def _sort_key(defn):
...
+                    # Should bounds_are_climatological be included here?
```
bjlittle: @trexfeathers What behavioural difference does it make if it is included?

Note that, this is Python2 specific, so this will be a non-issue come Python3 and `iris` 3.0.0 :tada:

lib/iris/coords.py
```diff
@@ -671,6 +679,32 @@ def _bounds_setter(self, bounds):
 
     bounds = property(_bounds_getter, _bounds_setter)
 
+    def _climatology_getter(self):
```
bjlittle: @trexfeathers It's like your subconscious is trying to tell you that `climatology` is the proper name of the attribute and not `bounds_are_climatological` :stuck_out_tongue_winking_eye:

lib/iris/coords.py
```diff
@@ -671,6 +679,32 @@ def _bounds_setter(self, bounds):
...
+
+        Always reads as False if there are no bounds.
+        On set, the input value is cast to a boolean, and an exception is
+        raised if there are no bounds.
+
+        """
+        result = False
+        if self.has_bounds():
+            result = self._bounds_are_climatological
+        return result
```
bjlittle: @trexfeathers This could all neatly boil down to:
```python
return self._climatology if self.has_bounds() else False
```

docs/iris/src/whatsnew/contributions_2.3.0/newfeature_2019-Sep-11_climatological_bounds.txt
```diff
@@ -0,0 +1,3 @@
+* support for loading, storing and saving of NetCDF climatological bounds.
```
bjlittle: `support` -> `Support`

lib/iris/fileformats/netcdf.py
```diff
@@ -1450,13 +1450,21 @@ def _create_cf_bounds(self, coord, cf_var, cf_name):
...
+            else:
+                property_name = 'bounds'
+                varname_extra = 'bnds'
+
             if bounds_dimension_name not in self._dataset.dimensions:
                 # Create the bounds dimension with the appropriate extent.
                 self._dataset.createDimension(bounds_dimension_name, n_bounds)
 
-            _setncattr(cf_var, 'bounds', cf_name + '_bnds')
+            boundsvar_name = cf_name + '_' + varname_extra
```
bjlittle: Splitting hairs...
```python
boundsvar_name = '{}_{}'.format(cf_name, varname_extra)
```

trexfeathers:
@abooton suggests the documentation needs to include the word 'climatology_bounds' since users could feasibly be looking for this within their data

bjlittle:
@trexfeathers Don't quite understand what you mean by `climatology_bounds` ? Only in the documentation or the actual name of the property?

lib/iris/coords.py
trexfeathers: `bounds_are_climatological` is a mouthful I agree. But I am worried about a word that could have too broad implications, such as `climatology`. For me the greatest importance is clarity on the limited scope of the property, and so that the name we use needs to begin with `bounds_are_`. I suggest `bounds_are_climatology`, which should be a _little_ less clunky than the current name.

@abooton please could you relay what you have discussed with me, so that we can get every side of this recorded in the PR?

lib/iris/coords.py
trexfeathers: I struggled to follow the purpose of this section when talking about it with @pp-mo . Given it's just an uncertainty and doesn't break any tests, I'm in favour of leaving this until Iris 3 solves the issue anyway. I'll remove the comment

lib/iris/coords.py
trexfeathers: Stop, my mind is feeling more dualistic by the minute 🤢

lib/iris/coords.py
```diff
@@ -671,6 +679,32 @@ def _bounds_setter(self, bounds):
...
+    def _climatology_setter(self, value):
+        # Ensure the bounds are a compatible shape.
+        value = bool(value)
+        if value and not self.has_bounds():
+            raise ValueError("Cannot set bound climatology to True: no "
+                             "bounds exist.")
+        self._bounds_are_climatological = value
+
+    bounds_are_climatological = property(_climatology_getter,
+                                         _climatology_setter)
```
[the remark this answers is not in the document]
trexfeathers: This would create inconsistent behaviour from _bounds_getter and _bounds_setter. Should I change these too?

lib/iris/coords.py
bjlittle: Okay, how about I throw I throw this curve ball into the mix...

Reference [Appendix A: Attrributes](http://cfconventions.org/Data/cf-conventions/cf-conventions-1.7/cf-conventions.html#attribute-appendix) in **Table A.1. Attributes**, the `climatology` attribute has the **Description** of `Identifies a climatology variable`.

So call the property `climatology_variable` because according to CF, that's what it is :wink:

lib/iris/fileformats/netcdf.py
trexfeathers: 👍

docs/iris/src/whatsnew/contributions_2.3.0/newfeature_2019-Sep-11_climatological_bounds.txt
trexfeathers: 👍

lib/iris/coords.py
trexfeathers: 👍

trexfeathers:
> @trexfeathers Don't quite understand what you mean by climatology_bounds ? Only in the documentation or the actual name of the property?
> If climatology_bounds was the proposed property name, then to me it suggests that it is actually the bounds of the climatology.

To clarify this was purely referring to the documentation

trexfeathers:
> @trexfeathers Awesome PR. I've reviewed all the code changes and added comments to be addressed - I'll spin thought the tests separately.
> The only part that's also missing apart from updating the xml_element is addressing the case where you have a coordinate that does indeed have climatology bounds, but the user then manually removes the bounds. We need the connective tissue to automatically reset the climatology boolean flag to False.

This should be achieved by the following section of _bounds_setter:
```
    def _bounds_setter(self, bounds):
        # Ensure the bounds are a compatible shape.
        if bounds is None:
            self._bounds_dm = None
            self._bounds_are_climatological = False
```

lib/iris/coords.py
bjlittle: Actually, re-reading the CF conventions again, the climatology variable is the actual cube that the time axis coordinate is attached to, where the time axis is deemed a climatological time axis defined within a climatological time coordinate variable (well, from my understanding).

That said, the following statement from the CF conventions [7.4. Climatological Statistics](http://cfconventions.org/Data/cf-conventions/cf-conventions-1.7/cf-conventions.html#climatological-statistics) is relevant here:

> _"... climatological time coordinate variable does not have a `bounds` attribute. Instead, it has a `climatology` attribute ..."_

The difficulty here is that we're in the business of implementing the conventions and we're overloading (for convenience and the common sense balance between effort and benefit) the use of the `bounds` property, and need a simple means to discriminate it's overloaded use between _normal_ and _climatological_ time coordinates.

In the pure CF conventions world `bounds` and `climatology` as controlled attributes are mutually exclusive, so you could also argue that introducing a property of `climatology` is really incorrect here. However, having a property of `climatological` (as much as I hate typing that word) make sense in this context.

Given that we're advocating that the data model of `iris` is based on the CF conventions (with a healthy peppering of pragmatism due to conventions vs implementations dilemmas) I believe it's fine to borrow from the CF conventions vernacular without being accused of being file format specific. 

That said, and given the discussion so far, I'd recommend the use of `climatological` as a property name, but on the proviso that this can **only** be set on a **time coordinate**, which implies enforcing that the `iris.Coord.units.is_time_reference()` is checked **before** setting the `climatological` property to `True`. Otherwise, it's not valid to set the `climatological` property on just any old coordinate.

lib/iris/coords.py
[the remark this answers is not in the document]
bjlittle: @trexfeathers Just go with this pattern here for the `climatological` getter/setter and nothing more.

I'll raise an issue to cover the use of `x = property(_x_getter, _x_setter)` in the rest of the code base, as a technical debt issue that we need to address.

lib/iris/coords.py
trexfeathers: Sounds ideal @bjlittle I'll get on this now

abooton: I have two thoughts on this...
1) I think it would be clearer from a user perspective to have the flag called `climatological_bounds`.  (The use of "climatology" confuses me) I usually use the term ‘climatology’ to refer to the data phenomenon averaged over a number of years.  Everything else is "climatological" as it is associated with the climatology.  Incidentally JULES has a "is_climatological" flag for this.
2) What behaviour are we after?  From the users perspective I think using a "bounds_are_climatological" flag rather than something akin to "is_climatological" indicates different behaviour, in that the former I expect the bound info will be reset on save, the later hints that on save the entire coord and associated attributes will be "climatological".  i.e. the cell_methods name would get updated upon save too.

bjlittle: @abooton Thanks. 

We're simply going with `climatological` as a property name. 

It is a simple flag and only applies to a time coordinate, and is in keeping with the behaviour alluded to in the CF conventions [7.4. Climatological Statistics](http://cfconventions.org/Data/cf-conventions/cf-conventions-1.7/cf-conventions.html#climatological-statistics) i.e., this is a climatological time coordinate. The implementation implication for us is that the `bounds` of the coordinate are `climatological` - in this regard the use of `iris.coord.Coord.bounds` is overloaded and I deem the effort to subclass in this instance not worth the end benefit. We're playing the pragmatic card.

From a conventions purist perspective, this whole issue could be circumvented by having a `ClimatologicalCoord` sub-classed from `DimCoord`, which has a `climatology` rather than `bounds`, but the fall-out within `iris` would be unreasonable.

There may be further connective tissue with regards to enforcing certain behaviour for cell methods given a climatological time coordinate and performing climatological statistics, but for the sake of progress I'd argue that is outside the scope of this particular PR.

abooton: `climatological` is good.  As long as the doc string accurately indicates the current behaviour I think it is fine. We will have a nicely named variable from which to extent the "climatological" scope in the future as required.

trexfeathers:
@bjlittle testing has been fixed and added to, that's me done for the day

bjlittle:
@trexfeathers Testing is kinda broken now... I'll take a peek :eyes:

bjlittle:
@trexfeathers Thanks for hanging in there, awesome PR, thanks :+1:

pp-mo:
Great work @bjlittle @trexfeathers !

PR #3397, opened 2019-09-12
Auto branch docs

### LINKED ISSUE #3396: Auto build of docs for each branch
The documentation used to be automatically built each time each branch of iris was updated. 

This has now broken (sometime since last summer). 

See https://github.com/SciTools/iris/issues/3341#issuecomment-504438195

pp-mo:
Closes #3396 

NOTE: I have made version changes (2.2 --> 2.3)  to the .RSTs in 3 places :
  * `docs/iris/src/_templates/index.html`
  * `docs/iris/src/_templates/layout.html`
  * `docs/iris/src/whatsnew/index.rst`

In future, to make this scheme work properly (so branch docs are properly readable, especially the whats-new), we really need to move these on _just after a new release, when we bump the version in master_.
That is, exactly as we do to the iris version string.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -13,7 +13,19 @@ env:
     # The decryption key for the encrypted .github/deploy_key.scitools-docs.enc.
     - secure: "N9/qBUT5CqfC7KQBDy5mIWZcGNuUJk3e/qmKJpotWYV+zwOI4GghJsRce6nFnlRiwl65l5oBEcvf3+sBvUfbZqh7U0MdHpw2tHhr2FSCmMB3bkvARZblh9M37f4da9G9VmRkqnyBM5G5TImXtoq4dusvNWKvLW0qETciaipq7ws="
   matrix:
-    - PYTHON_VERSION=3.7 TEST_TARGET=doctest
+    - PYTHON_VERSION=2.7 TEST_TARGET=default TEST_MINIMAL=true
+    - PYTHON_VERSION=2.7 TEST_TARGET=default
+    - PYTHON_VERSION=2.7 TEST_TARGET=example
+
+    - PYTHON_VERSION=3.6 TEST_TARGET=default TEST_MINIMAL=true
+    - PYTHON_VERSION=3.6 TEST_TARGET=default
+    - PYTHON_VERSION=3.6 TEST_TARGET=example
+    - PYTHON_VERSION=3.6 TEST_TARGET=doctest
+
+    - PYTHON_VERSION=3.7 TEST_TARGET=default TEST_MINIMAL=true
+    - PYTHON_VERSION=3.7 TEST_TARGET=default
+    - PYTHON_VERSION=3.7 TEST_TARGET=example
+    - PYTHON_VERSION=3.7 TEST_TARGET=doctest PUSH_BUILT_DOCS=true
 
 git:
   # We need a deep clone so that we can compute the age of the files using their git history.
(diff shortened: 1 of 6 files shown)


.travis.yml
```diff
@@ -135,6 +135,25 @@ script:
...
+    fi
+
+  # When pushing built docs, attempt to make a preliminary whatsnew with 'aggregate_directory.py', before the build.
+  - >
+    if [[ ${PUSH_BUILT_DOCS} == 'true' ]]; then
+      cd ${INSTALL_DIR}/docs/iris/src/whatsnew;
+      WHATSNEW=$(ls -d contributions_* 2>/dev/null);
+      if [[ "$WHATSNEW" != "" ]]; then
+        VERSION=$(echo $WHATSNEW | grep -o "[[:digit:].]*");
+        python aggregate_directory.py $VERSION;
```
lbdreyer: I think aggregate_directory should also take an optional flag to set the release date to be something like "Unreleased" or "TBC"
Currently, it [uses the current date](https://github.com/SciTools/iris/blob/7419f8839510d8b784bcca9567fe5ac9a9aaa68a/docs/iris/src/whatsnew/aggregate_directory.py#L180) so if you look at the [Iris 2.3 whats new](https://scitools-docs.github.io/iris/auto_branch_docs/whatsnew/2.3.html) that is generated, it claims that it has already been released...
I think this will just lead to confusion

.travis.yml
```diff
@@ -135,6 +135,25 @@ script:
...
     fi
 
+  # A call to check "whatsnew" contributions are valid, because the Iris test
+  # for it needs a *developer* install to be able to find the docs.
+  - if [[ ${TEST_TARGET} == 'doctest' ]]; then
+      cd ${INSTALL_DIR}/docs/iris/src/whatsnew;
+      python aggregate_directory.py --checkonly;
+    fi
+
+  # When pushing built docs, attempt to make a preliminary whatsnew with 'aggregate_directory.py', before the build.
```
lbdreyer: This line is a bit long. Could you split it over two lines to make it more readable

README.md
```diff
@@ -85,6 +85,11 @@ use of standard NumPy/dask arrays as its underlying data storage.
 The documentation for Iris is available at <https://scitools.org.uk/iris/docs/latest>,
 including a user guide, example code, and gallery.
 
+## Work in progress
+
+A documentation build for the latest development code is visible at
+<https://scitools-docs.github.io/iris/master/index.html>.
+
```
lbdreyer: In the past we also had a badge in the [README that pointed to the docs built from master](https://github.com/SciTools/iris/blob/962f0159513e6eb08ed1a76fbc8485a31dd584bb/README.md). Could you readd it?

README.md
```diff
@@ -85,6 +85,11 @@ use of standard NumPy/dask arrays as its underlying data storage.
 The documentation for Iris is available at <https://scitools.org.uk/iris/docs/latest>,
 including a user guide, example code, and gallery.
 
+## Work in progress
```
lbdreyer: Could we remove this title?
To me, it reads as if this section is a WIP in the sense that at some point we hope for it not to be "in progress anymore" and to actually finalise on the contents. But that's not what will happen. It will always be "in progress".

I think it would be fine if 

> A documentation build for the latest development code is visible at https://scitools-docs.github.io/iris/master/index.html.

is part of the section above

pp-mo: ok!

.travis.yml
pp-mo: ok

README.md
pp-mo: ok!

pp-mo:
Thanks @lbdreyer , I hope I've addressed those points now.
Mostly : I have added a 'special' mode to the aggregation as suggested, and it produces a temporary whatsnew page with a temporary non-version name.
See : https://scitools-docs.github.io/iris/auto_branch_docs/whatsnew/index.html

README.md
lbdreyer: @pp-mo It looks like this title is still yet to be removed.

pp-mo: Sorry, that change got lost when I reverted something.
Will reinstate...

lbdreyer:
Looking good! Thanks @pp-mo !

Issue #3360, opened 2019-07-23
Cell methods are treated inconsistently after saving and re-loading


schlunma:
Another issue which may prevent the concatenation of cubes:

```python
import iris

# Create 1D cube with coordinate which has a long name with a space
coord = iris.coords.DimCoord([0, 1], long_name='long name')
cube = iris.cube.Cube([1, 2], dim_coords_and_dims=[(coord, 0)])

# Collapse it and print cell_methods
cube = cube.collapsed('long_name', iris.analysis.MEAN)
print(cube.cell_methods)  # gives (CellMethod(method='mean', coord_names=('long name',), intervals=(), comments=()),)

# Save it and check cell methods again
iris.save(cube, 'test.nc')
cube_new = iris.load_cube('test.nc')
print(cube_new.cell_methods)  # gives (CellMethod(method='mean', coord_names=('name',), intervals=(), comments=()),)
```

Would it be possible to come up with a consistent solution that doesn't change the cell methods? This prevents `merge_cube` and `concatenate_cube` operations.

bjlittle:
@schlunma Thanks for taking the time to report this... l completely agree that this is inconsistent behaviour, and looks like a bug to me.

Let me just confirm, with some further details of the underlying issue and we'll take it from there

lbdreyer:
@schlunma thanks for reporting this!

Iris follows CF conventions which [specify that the name of the cell method should be](http://cfconventions.org/Data/cf-conventions/cf-conventions-1.7/cf-conventions.html#cell-methods):

> name can be a dimension of the variable, a scalar coordinate variable, a valid standard name, or the word "area".

where, `dimension of the variable` refers to the name of a dimension in the netCDF file that the cube saves out to; `scalar coordinate variable` refers to the name of the scalar coordinate variable in the netCDF file ; and `valid standard_name` refers to the list of approved [CF standard names](http://cfconventions.org/Data/cf-standard-names/68/build/cf-standard-name-table.html). 

Currently, Iris effectively uses what is returned by the `name()` method on the cube to set the name of the generated cell method. This is fine if the coordinate has a standard_name as `name()` will return the standard_name. However, as you have found, it causes issues if the coordinate has no standard_name and only contains a long_name, in which case `name()` returns the long_name as there is no standard_name. 
Iris is wrong to be using the long name to set the name of the cell method, as a long name is not approved by the CF conventions and may contain spaces.

[...]

PR #3399, opened 2019-09-13
Support name tokens

bjlittle:
This PR is the outcome of a side discussion regarding issue https://github.com/SciTools/iris/issues/3360.

In short, in `iris` we've allowed the use of `long_name` from a coordinate to be used in the construction of the `CellMethod`, but that causes issues when the `long_name` contains one or more space (" ") characters, as it then quickly becomes ambiguous to parse the grammar suggested in the CF Conventions for a `cell_methods` attribute on loading after saving the `CellMethod` to NetCDF. That's not a fault of the CF Conventions, more our interpretation and implementation of it (from a file format agnostic point of view).

Reference, [Cell Methods](http://cfconventions.org/Data/cf-conventions/cf-conventions-1.7/cf-conventions.html#cell-methods), 2nd paragraph:

> In the specification of this attribute, name can be a dimension of the variable, a scalar coordinate variable, a valid standard name, or the word "area"

That said, this PR attempts to resolve this issue by ensuring that the `coord_names` and `method` used within the construction of a `CellMethod` are [valid NetCDF name tokens](https://www.unidata.ucar.edu/software/netcdf/docs/netcdf_data_set_components.html#object_name), and thus can be parsed within the rules suggested in the CF Conventions without ambiguity (or over-engineering :wink:).


bjlittle:
@lbdreyer I still need to add some test coverage to the PR... just pushed this so that you can see the approach I'm taking :wink:

lib/iris/_cube_coord_common.py
```diff
@@ -19,15 +19,18 @@
 from six.moves import (filter, input, map, range, zip)  # noqa
 import six
 
-# TODO: Is this a mixin or a base class?
-
```
bjlittle: Horse. Bolted.

bjlittle:
@lbdreyer Okay, I think that I'm pretty much there...

As it happens, there wasn't any unit test coverage of `CellMethods` (there is for parsing, but not construction and usage), but I've added the minimal test coverage for the changes in this PR.

Being a good citizen, I'll add test coverage in a separate PR.

I've also not updated the docs either, just so that this PR can get banked sooner rather than later, but happy to make a follow-up PR, if you think that's appropriate... :smile:

lib/iris/tests/unit/cube_coord_common/test_CFVariableMixin.py
```diff
@@ -0,0 +1,155 @@
...
+
+    def test_var_name(self):
+        token = 'var_name'
+        self.cf_var.var_name = token
+        result = self.cf_var.name()
+        self.assertEqual(result, token)
+
+    def test_stash(self):
+        token = 'stash'
+        self.cf_var.var_name = token
```
lbdreyer: This doesn't test stash, this repeats the var_name test? I'm guessing this was a copy and paste error?

lib/iris/tests/unit/cube_coord_common/test_CFVariableMixin.py
```diff
@@ -0,0 +1,155 @@
...
+    def test_token_default(self):
+        result = self.cf_var.name(token=True)
+        self.assertEqual(result, self.default)
+
+    def test_fail_token_default(self):
+        cf_var = CFVariableMixin()
+        cf_var.standard_name = None
+        cf_var.long_name = None
+        cf_var.var_name = None
+        cf_var.attributes = {}
```
lbdreyer: Is this necessary? I can't see a difference between `cf_var` and `self.cf_var`

lib/iris/_cube_coord_common.py
```diff
@@ -144,12 +196,10 @@ def var_name(self):
...
-            if not name:
-                raise ValueError('An empty string is not a valid netCDF '
-                                 'variable name.')
-            elif set(name).intersection(string.whitespace):
-                raise ValueError('{!r} is not a valid netCDF variable name '
-                                 'as it contains whitespace.'.format(name))
+            result = self.token(name)
+            if result is None or not name:
+                emsg = '{!r} is not a valid NetCDF variable name.'
+                raise ValueError(emsg.format(name))
```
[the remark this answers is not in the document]
bjlittle: Not sure either. But let's err on the side of this change upsetting someone... so I'll add an entry into the `whatsnew` 👍

lib/iris/tests/unit/cube_coord_common/test_CFVariableMixin.py
bjlittle: Yup, good spot 👀

lib/iris/tests/unit/cube_coord_common/test_CFVariableMixin.py
bjlittle: Yup, I can purge the use of `cf_var`, thanks!

bjlittle:
@lbdreyer I've actioned all your review comments, `travis-ci` is happy, over to you... 😺

lbdreyer:
All good :+1: :+1: Thanks @bjlittle !

Issue #3382, opened 2019-08-27
Support for NetCDF files on Geostationary Projections


lbdreyer:
The [Geostationary grid mapping](http://cfconventions.org/Data/cf-conventions/cf-conventions-1.7/cf-conventions.html#_geostationary_projection) was added in CF conventions 1.7.

This has been raised by a number of our users in the satellite community. Increasingly satellite products on this type of projection are becoming available in the satellite meteorology community so it would be very helpful to support the Geostationary grid_mapping correctly. 

For more background on the discussions adding the projection to CF 1.7, you can view the original discussion of its inclusion [here](https://cf-trac.llnl.gov/trac/ticket/72).

(Let me know if you need help getting hold of example netCDF files on a geostationary projection!)

lbdreyer:
Currently Iris has `iris.coord_system.VerticalPerspective`  coord system which points off to the `ccrs.Geostationary`, which is a `geos` [proj4 projection](https://github.com/SciTools/cartopy/blob/master/lib/cartopy/crs.py#L2056).
 
The cf conventions suggest that the Vertical Perspective projection corresponds to the `nsper` proj4 

>     A general description of vertical perspective projection is given in [Snyder], pages 169-181.
>     The corresponding projection in PROJ.4 is nsper. This should not be confused with the PROJ.4 geos projection.
(see [Vertical Perspective](http://cfconventions.org/Data/cf-conventions/cf-conventions-1.7/cf-conventions.html#vertical-perspective))

I think Iris needs to be updated so that the `iris.coord_system.VerticalPerspective` uses `ccrs.NearsidePerspective` (which is `nsper`) and we add a `iris.coord_system.Geostationary` that uses `ccrs.Geostationary` (which is `geos`)

lbdreyer:
Closed by #3406

PR #3406, opened 2019-09-19
Correct handling of NetCDF VerticalPerspective and Geostationary coordinate systems

trexfeathers:
Addresses #3382

### DIFF
--- a/docs/iris/src/whatsnew/contributions_2.3.0/newfeature_2019-Sep-20_geostationary_coordsystem.txt
+++ b/docs/iris/src/whatsnew/contributions_2.3.0/newfeature_2019-Sep-20_geostationary_coordsystem.txt
@@ -1,0 +1,7 @@
+* New coordinate system: :class:`iris.coord_systems.Geostationary`, based on
+the `CF Geostationary projection definition <http://cfconventions
+.org/cf-conventions/cf-conventions.html#_geostationary_projection>`_.
+* :class:`iris.coord_systems.VerticalPerspective` coordinate system now uses
+the `CF Vertical perspective definition <http://cfconventions
+.org/cf-conventions/cf-conventions.html#vertical-perspective>`_; had been
+erroneously using Geostationary.
--- a/docs/iris/src/whatsnew/contributions_2.3.0/newfeature_2019-Sep-20_geostationary_coordsystem.txt
+++ b/docs/iris/src/whatsnew/contributions_2.3.0/newfeature_2019-Sep-20_geostationary_coordsystem.txt
@@ -1,7 +1,5 @@
 * New coordinate system: :class:`iris.coord_systems.Geostationary`, based on
 the `CF Geostationary projection definition <http://cfconventions
 .org/cf-conventions/cf-conventions.html#_geostationary_projection>`_.
-* :class:`iris.coord_systems.VerticalPerspective` coordinate system now uses
-the `CF Vertical perspective definition <http://cfconventions
-.org/cf-conventions/cf-conventions.html#vertical-perspective>`_; had been
-erroneously using Geostationary.
+* :class:`iris.coord_systems.VerticalPerspective` can now be saved to and
+loaded from NetCDF files.
--- a/lib/iris/coord_systems.py
+++ b/lib/iris/coord_systems.py
@@ -555,7 +555,7 @@ class Orthographic(CoordSystem):
 
 class VerticalPerspective(CoordSystem):
     """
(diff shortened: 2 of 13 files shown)


trexfeathers:
Still needs a What's New

lib/iris/tests/unit/fileformats/netcdf/test_Saver.py
```diff
@@ -44,6 +45,10 @@
 
 
 class Test_write(tests.IrisTest):
+    # -------------------------------------------------------------------------
+    # It is not considered necessary to have integration tests for saving
```
stephenworsley: This refers to integration tests, but this is in iris/tests/unit . Can these still be described as integration tests?

trexfeathers: You're right, I hadn't considered how oddly this reads. But the tests below (e.g. test_mercator) definitely don't fit the definition of unit test in my opinion since NetCDF saving relies on a large number of steps. Could you suggest a better word than integration, which still makes it clear that these aren't unit tests?

pp-mo: Taking a **very** narrow view, just omit the word "integration" : job done!

stephenworsley: Yeah, having read through the tests I think I agree that these are closer to integration tests than unit tests. If you want to avoid confusion, it could be worth considering just getting rid of the word "integration", though I'm more convinced now that this ought to be be fine as is.

docs/iris/src/whatsnew/contributions_2.3.0/newfeature_2019-Sep-20_geostationary_coordsystem.txt
```diff
@@ -0,0 +1,7 @@
+* New coordinate system: :class:`iris.coord_systems.Geostationary`, based on
+the `CF Geostationary projection definition <http://cfconventions
+.org/cf-conventions/cf-conventions.html#_geostationary_projection>`_.
+* :class:`iris.coord_systems.VerticalPerspective` coordinate system now uses
+the `CF Vertical perspective definition <http://cfconventions
+.org/cf-conventions/cf-conventions.html#vertical-perspective>`_; had been
+erroneously using Geostationary.
```
lbdreyer: Should this not be a bug fix rather?

lib/iris/coord_systems.py
```diff
@@ -645,6 +639,104 @@ def as_cartopy_projection(self):
         return self.as_cartopy_crs()
 
 
+class Geostationary(CoordSystem):
+    """
+    An geostationary satellite image map projection.
```
lbdreyer: `An geostrationary` -> `A geostationary`

lib/iris/coord_systems.py
```diff
@@ -553,7 +555,7 @@ def as_cartopy_projection(self):
 
 class VerticalPerspective(CoordSystem):
     """
-    An geostationary satellite image map projection.
+    An vertical/near-side perspective satellite image map projection.
```
lbdreyer: `An vertical..` -> `A vertical`

lib/iris/coord_systems.py
```diff
@@ -563,7 +565,7 @@ def __init__(self, latitude_of_projection_origin,
                  longitude_of_projection_origin, perspective_point_height,
                  false_easting=0, false_northing=0, ellipsoid=None):
         """
-        Constructs an Vertical Perspective Geostationary coord system.
+        Constructs an Vertical Perspective coord system.
```
lbdreyer: `an` -> `a`

docs/iris/src/whatsnew/contributions_2.3.0/newfeature_2019-Sep-20_geostationary_coordsystem.txt
trexfeathers: 👍

lib/iris/coord_systems.py
trexfeathers: Without reading I copied from vertical perspective, which had been copied without reading from another coord system, so I managed to double the error! I feel appropriately ashamed 😖

pp-mo:
Needs rebase.  Can you do @trexfeathers ?

trexfeathers:
Note that an Iris Geostationary coordinate system will only successfully plot using Cartopy >=0.17

PR #3420, opened 2019-09-25
Fix _repr_html_, close tags appropriately

stephenworsley:
Made the `_make_shapes_row` generate more proper html, the initial `<tr ...>` is now closed by `</tr>` rather than `</td>`.


lib/iris/experimental/representation.py
```diff
@@ -207,7 +207,7 @@ def _make_shapes_row(self):
         for shape in self.shapes:
             cells.append(
                 '<td class="iris iris-inclusion-cell">{}</td>'.format(shape))
-        cells.append('</td>')
+        cells.append('</tr>')
```
lbdreyer: Great spot @stephenworsley ! :eyes:

PR #3421, opened 2019-09-25
use @property everywhere

bjlittle:
This PR follows-up from the review of #3395 and more specifically this [comment](https://github.com/SciTools/iris/pull/3395#discussion_r327686889), to move away from using this property pattern:
```python
def _thing_getter(self):
    ...
def _thing_setter(self, ...):
    ...
thing = property(_thing_getter, _thing_setter)
```
to using this property pattern:
```python
@property
def thing(self):
    ...
@thing.setter
def thing(self, ...):
    ...
```
for reasons discussed in the #3395 [comment](https://github.com/SciTools/iris/pull/3395#discussion_r327686889).

### DIFF
--- a/lib/iris/coords.py
+++ b/lib/iris/coords.py
@@ -631,13 +631,11 @@ class Coord(six.with_metaclass(ABCMeta, CFVariableMixin)):
             result = result.view()
         return result
 
-    @property
-    def points(self):
+    def _points_getter(self):
         """The coordinate points values as a NumPy array."""
         return self._points_dm.data.view()
 
-    @points.setter
-    def points(self, points):
+    def _points_setter(self, points):
         # Set the points to a new array - as long as it's the same shape.
 
         # Ensure points has an ndmin of 1 and is either a numpy or lazy array.
@@ -651,8 +649,11 @@ class Coord(six.with_metaclass(ABCMeta, CFVariableMixin)):
         else:
             self._points_dm.data = points
 
-    @property
-    def bounds(self):
+    # Require to use this property pattern rather than @property due to
+    # subclass specialisation of _points_setter method
+    points = property(_points_getter, _points_setter)
+
+    def _bounds_getter(self):
         """
         The coordinate bounds values, as a NumPy array,
         or None if no bound values are defined.
@@ -666,8 +667,7 @@ class Coord(six.with_metaclass(ABCMeta, CFVariableMixin)):
             bounds = self._bounds_dm.data.view()
         return bounds
 
-    @bounds.setter
-    def bounds(self, bounds):
+    def _bounds_setter(self, bounds):
         # Ensure the bounds are a compatible shape.
         if bounds is None:
             self._bounds_dm = None
@@ -714,6 +714,10 @@ class Coord(six.with_metaclass(ABCMeta, CFVariableMixin)):
                 raise ValueError(emsg)
 
         self._climatological = value
+
+    # Require to use this property pattern rather than @property due to
+    # subclass specialisation on _bounds_setter method
+    bounds = property(_bounds_getter, _bounds_setter)
 
     def lazy_points(self):
         """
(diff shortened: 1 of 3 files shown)


QuLogic:
What is this about subclass specialisation? You should be able to do `@SuperClass.property.setter` on the subclass.

pp-mo:
I reckon there are similar changes due in `iris.fileformats.pp`
See : https://github.com/bjlittle/iris/pull/22

lbdreyer:
@bjlittle Is this ready to be reviewed? (I've been holding off as you previously said this wasn't ready just yet so I thought I'd check)

bjlittle:
@lbdreyer It's good to go as far as I'm concerned, go for it if your happy :smile:

lib/iris/coords.py
```diff
@@ -1940,7 +1940,13 @@ def _new_points_requirements(self, points):
         if points.size > 1 and not iris.util.monotonic(points, strict=True):
             raise ValueError('The points array must be strictly monotonic.')
 
-    def _points_setter(self, points):
+    @property
+    def points(self):
+        # Call the parent points getter.
```
pp-mo: I'm now finding that the `points` and `bounds` properties of DimCoords now don't have a useable docstring, whereas previously they did, because we were only replacing the *setter*, as in `xxx = property(...Coord._xxx_getter, _xxx_setter)`.

By my experiment, the same occurs with `help()` and `print(coord.xxx.__doc__)` : 

Results before:
```
>>> print(iris.coords.Coord.points.__doc__) 
The coordinate points values as a NumPy array.
>>> print(iris.coords.Coord.bounds.__doc__)

        The coordinate bounds values, as a NumPy array,
        or None if no bound values are defined.

        .. note:: The shape of the bound array should be: ``points.shape +
            (n_bounds, )``.

        
>>> print(iris.coords.DimCoord.points.__doc__)
The coordinate points values as a NumPy array.
>>> 
>>> iris.coords.DimCoord.bounds.__doc__ is iris.coords.Coord.bounds.__doc__
True
>>> iris.coords.DimCoord.points.__doc__ is iris.coords.Coord.points.__doc__
True
>>> 
```

Results new:
```
>>> print(iris.coords.Coord.points.__doc__)
The coordinate points values as a NumPy array.
>>> print(iris.coords.Coord.bounds.__doc__)

        The coordinate bounds values, as a NumPy array,
        or None if no bound values are defined.

        .. note:: The shape of the bound array should be: ``points.shape +
            (n_bounds, )``.

        
>>> print(iris.coords.DimCoord.points.__doc__)
None
>>> print(iris.coords.DimCoord.bounds.__doc__)
None
>>> 
```

I don't know if you can achieve this with a wrap, or what ...

pp-mo: Actually, I also goofed ...
```
>>> print(iris.fileformats.pp.PPField2.t1.__doc__)
A cftime.datetime object consisting of the lbyr, lbmon,
>>> 
```

As written, there are 2 strings.  It obviously doesn't merge them, but only takes the first one as the docstring.  PPField3, by constrast, is okay.  I will fix ...

bjlittle:
@pp-mo My last change fixes the `points` and `bounds` doc-strings on the `DimCoord`, see https://github.com/SciTools/iris/pull/3421/commits/7ec1872987012d72efa060890a05e5c48bddfdd8

pp-mo:
> fix t1/t2 doc-strings https://github.com/SciTools/iris/pull/3421/commits/ecb79ca1a801d7c06ae2cbfa2a5cfb06030fc1be

Thanks, saves my effort !!

pp-mo:
> last change fixes the points and bounds doc-strings on the DimCoord, see https://github.com/SciTools/iris/pull/3421/commits/7ec1872987012d72efa060890a05e5c48bddfdd8

Okay this looks nice.
I'm just a little suspicious what it really means to say "@Coord.points.setter" within the child class.
But obviously it does seem to work ...

pp-mo:
> But obviously it does seem to work ...

Ok now I ***believe*** :pray: :clap: :church: !!
It makes sense from [this in Python docs](https://docs.python.org/3.3/library/functions.html?highlight=property#property), that `@Coord.points.setter` returns a copy of the parent property, with a modified setter function.  
Job done ! :candy:

bjlittle:
Nice one! Thanks @pp-mo :beers:

Issue #3419, opened 2019-09-25
Coord masked points


rcomer:
Stumbled across this when trying to understand #3402.  If you instantiate a `Coord` with a masked array, the mask gets stripped off:

```
In [47]: time_coord = iris.coords.AuxCoord(ma.array(range(5), mask=[False, True, False, True, False]),
                                           units='days since 2001-05-01', standard_name='time')

In [48]: time_coord.points
Out[48]: array([0, 1, 2, 3, 4])
```

```
In [75]: time_coord = iris.coords.DimCoord(ma.array(range(5), mask=ma.nomask),
                                           units='days since 2001-05-01', standard_name='time')
In [76]: time_coord.points
Out[76]: array([0, 1, 2, 3, 4])
```

This is not the case for a dask masked array:

```
In [7]: time_coord = iris.coords.AuxCoord(da.ma.masked_array(da.arange(5, chunks=5)),
   ...:                                   units='days since 2001-05-01', standard_name='time')
In [8]: time_coord.has_lazy_points()
Out[8]: True

In [9]: time_coord.points
Out[9]:
masked_array(data=[0, 1, 2, 3, 4],
             mask=False,
       fill_value=999999)
```

I think [this may be the line](https://github.com/SciTools/iris/blob/534b4088bcdd7d8d0c03279aef2b36a09462e90e/lib/iris/coords.py#L616) that converts a numpy masked array into a plain numpy array.  There is nothing in the comments to suggest that this was a deliberate effect, and it seems to me that it shouldn't happen because you are probably unmasking nonsense values.

The [cf conventions](http://cfconventions.org/Data/cf-conventions/cf-conventions-1.7/cf-conventions.html#missing-data) say that "Missing data is allowed in data variables and auxiliary coordinate variables."  So I think what should happen is:

* Modify the `Coord._sanitise_array` method to return a masked array if it's given a masked array. [Edit: or just throw an exception if we don't want to support that]
* Throw an exception if you try to assign a masked array to `DimCoord.points`.
* Also consider what happens to bounds.

pp-mo:
As described, definitely a bug !!

> I think this may be the line that converts a numpy masked array into a plain numpy array. 

I think so, yes.  Therefore the comment on the preceding line is also wrong !!

bjlittle:
@rcomer @pp-mo I'm working on a fix for this that I'd like to get included in the imminent `iris` 2.3.0 release :wink:

rcomer:
Fixed by #3427

PR #3422, opened 2019-09-26
Include AncillaryVariables and restructure Cube metadata.

lbdreyer:
This includes the functionality to represent CF Ancillary variables on a cube.

This PR also includes a restructuring of dimensional metadata objects (coords, cell measures and ancillary dataset) in a hope to avoid code duplication.

Below is a rough diagram of the new structure:
IrisCoordsUML

### DIFF
--- a/lib/iris/coords.py
+++ b/lib/iris/coords.py
@@ -92,6 +92,13 @@ class _DimensionalMetadata(six.with_metaclass(ABCMeta, CFVariableMixin)):
             A dictionary containing other cf and user-defined attributes.
 
         """
+        # Note: this class includes bounds handling code for convenience, but
+        # this can only run within instances which are also Coords, because
+        # only they may actually have bounds.  This parent class has no
+        # bounds-related getter/setter properties, and no bounds keywords in
+        # its __init__ or __copy__ methods.  The only bounds-related behaviour
+        # it provides is a 'has_bounds()' method, which always returns False.
+
         #: CF standard name of the quantity that the metadata represents.
         self.standard_name = standard_name
 
@@ -111,6 +118,7 @@ class _DimensionalMetadata(six.with_metaclass(ABCMeta, CFVariableMixin)):
         # Set up DataManager attributes and values.
         self._values_dm = None
         self._values = values
+        self._bounds_dm = None  # Only ever set on Coord-derived instances.
 
     def __getitem__(self, keys):
         """
@@ -124,6 +132,9 @@ class _DimensionalMetadata(six.with_metaclass(ABCMeta, CFVariableMixin)):
             indexing.
 
         """
+        # Note: this method includes bounds handling code, but it only runs
+        # within Coord type instances, as only these allow bounds to be set.
+
         # Fetch the values.
         values = self._values_dm.core_data()
 
@@ -137,20 +148,15 @@ class _DimensionalMetadata(six.with_metaclass(ABCMeta, CFVariableMixin)):
 
         # If the metadata is a coordinate and it has bounds, repeat the above
         # with the bounds.
-        if isinstance(self, Coord):
-            if self.has_bounds():
-                bounds = self._bounds_dm.core_data()
-                _, bounds = iris.util._slice_data_with_keys(bounds, keys)
-                bounds = bounds.copy()
-            else:
-                bounds = None
-
-            copy_args = dict(points=values, bounds=bounds)
-        else:
-            copy_args = dict(values=values)
+        copy_args = {}
+        if self.has_bounds():
+            bounds = self._bounds_dm.core_data()
+            _, bounds = iris.util._slice_data_with_keys(bounds, keys)
+            # Pass into the copy method : for Coords, it has a 'bounds' key.
+            copy_args['bounds'] = bounds.copy()
 
         # The new metadata is a copy of the old one with replaced content.
-        new_metadata = self.copy(**copy_args)
+        new_metadata = self.copy(values, **copy_args)
 
         return new_metadata
 
@@ -166,6 +172,8 @@ class _DimensionalMetadata(six.with_metaclass(ABCMeta, CFVariableMixin)):
             copied.
 
         """
+        # Note: this is overridden in Coord subclasses, to add bounds handling
+        # and a 'bounds' keyword.
         new_metadata = copy.deepcopy(self)
         if values is not None:
             new_metadata._values_dm = None
@@ -262,6 +270,8 @@ class _DimensionalMetadata(six.with_metaclass(ABCMeta, CFVariableMixin)):
                                                **kwargs)
 
     def __str__(self):
+        # Note: this method includes bounds handling code, but it only runs
+        # within Coord type instances, as only these allow bounds to be set.
         if self.units.is_time_reference():
             fmt = '{cls}({values}{bounds}' \
                   ', standard_name={self.standard_name!r}' \
@@ -289,6 +299,8 @@ class _DimensionalMetadata(six.with_metaclass(ABCMeta, CFVariableMixin)):
         return result
 
     def __repr__(self):
+        # Note: this method includes bounds handling code, but it only runs
+        # within Coord type instances, as only these allow bounds to be set.
         fmt = '{cls}({self._values!r}{bounds}' \
               ', standard_name={self.standard_name!r}, units={self.units!r}' \
               '{other_metadata})'
@@ -302,6 +314,9 @@ class _DimensionalMetadata(six.with_metaclass(ABCMeta, CFVariableMixin)):
         return result
 
     def __eq__(self, other):
+        # Note: this method includes bounds handling code, but it only runs
+        # within Coord type instances, as only these allow bounds to be set.
+
         eq = NotImplemented
         # If the other object has a means of getting its definition, then do
         # the  comparison, otherwise return a NotImplemented to let Python try
@@ -310,9 +325,19 @@ class _DimensionalMetadata(six.with_metaclass(ABCMeta, CFVariableMixin)):
             # metadata comparison
             eq = self._as_defn() == other._as_defn()
             # data values comparison
-            if eq:
+            if eq and eq is not NotImplemented:
                 eq = iris.util.array_equal(self._values, other._values,
                                            withnans=True)
+
+            # Also consider bounds, if we have them.
+            # (N.B. though only Coords can ever actually *have* bounds).
+            if eq and eq is not NotImplemented:
+                if self.has_bounds() and other.has_bounds():
+                    eq = iris.util.array_equal(self.bounds, other.bounds,
+                                               withnans=True)
+                else:
+                    eq = not self.has_bounds() and not other.has_bounds()
+
         return eq
 
     def __ne__(self, other):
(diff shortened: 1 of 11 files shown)


lib/iris/tests/unit/coords/test_CellMeasure.py
```diff
@@ -72,24 +72,6 @@ def test_set_data__lazy(self):
...
 
-    def test_set_data__int__lazy(self):
-        new_vals = as_lazy_data(np.array((1, 2, 3, 4), dtype=np.int32))
-        exp_emsg = "Cannot create cell measure with lazy data of type int32"
-        with self.assertRaisesRegexp(ValueError, exp_emsg):
-            self.measure.data = new_vals
-
-    def test_set_data__uint__lazy(self):
-        new_vals = as_lazy_data(np.array((1, 2, 3, 4), dtype=np.uint32))
-        exp_emsg = "Cannot create cell measure with lazy data of type uint32"
```
lbdreyer: In CF 1.7 it is allowable to create cell measures with integers

lib/iris/tests/unit/coords/test_CellMeasure.py
```diff
@@ -72,24 +72,6 @@ def test_set_data__lazy(self):
...
-        with self.assertRaisesRegexp(ValueError, exp_emsg):
-            self.measure.data = new_vals
-
-    def test_data_different_shape(self):
-        new_vals = np.array((1., 2., 3.))
-        msg = 'New data shape must match existing data shape.'
-        with self.assertRaisesRegexp(ValueError, msg):
-            self.measure.data = new_vals
-
```
lbdreyer: I removed this when I replaced the CellMeasure specific `_data_setter` with the new `_DimensionalMetadata._values_setter` which does not force the new shape to match the old shape. It seemed a bit of a strange thing to enforce, especially considering we don't enforce this requirement on Auxiliary coordinates. 
You could, however, argue that this is a breaking change. In which case I will reinstate this test and add a `data` property to `CellMeasure` that does the shape check as well as calling `_DimensionalMetadata._values_setter`

lib/iris/coords.py
```diff
@@ -52,6 +52,702 @@
...
+        _, values = iris.util._slice_data_with_keys(
+            values, keys)
+
+        # Copy values after indexing to avoid making metadata that is a
+        # view on another metadata.
+        values = values.copy()
+
+        # If the metadata is a coordinate and it has bounds, repeat the above
+        # with the bounds.
+        if isinstance(self, Coord):
```
lbdreyer: Not sure if you will agree with this approach, but it avoided the need for abstract methods.

lib/iris/coords.py
```diff
@@ -52,6 +52,702 @@
...
+                extended_shape = tuple([1] * ndims_missing + list(src.shape))
+                result = src.reshape(extended_shape)
+        else:
+            # Real data : a few more things to do in this case.
+            # Ensure the array is writeable.
+            # NB. Returns the *same object* if src is already writeable.
+            result = np.require(src, requirements='W')
+            # Ensure the array has enough dimensions.
+            # NB. Returns the *same object* if result.ndim >= ndmin
+            # Coords do not preserve masks of input points.
```
lbdreyer: As raised in #3419, coordinates for some reason throw away the mask. I kept that behaviour in case it was intentional. But if we decided it is a "bug" I could fix it here?

pp-mo: Just noticed this.  As noted [here in #3419](https://github.com/SciTools/iris/issues/3419#issue-498188507), masked (aux) coordinates definitely are _allowed_ in CF.
So I think it *is* a bug.

pp-mo:
Hi @lbdreyer 
Just visiting, I know this isn't ready for review yet.
But as I'm reviewing #3378, I'm seeing that changing the cube print function means you will also need to tweak the content handling in `iris.experimental.representation` :  This provides html output for the Jupyter `_repr_html_` handler, but it does it by  which reprocessing the `cube.__str__()` output.

BTW maybe that shouldn't be in `iris.experimental` anymore ?

lbdreyer:
Thanks for raising this @pp-mo. I also noticed that CellMeasure don't seem to be handled either?

I am happy for this work to be included in a separate PR, and have raised an issue to address it: #3467

lbdreyer:
@pp-mo Are you available to review this?

lib/iris/coords.py
```diff
@@ -52,6 +52,697 @@
...
+        * var_name:
+            The netCDF variable name for the ancillary variable.
+        * units
+            The :class:`~cf_units.Unit` of the ancillary variable's values.
+            Can be a string, which will be converted to a Unit object.
+        * attributes
+            A dictionary containing other cf and user-defined attributes.
+
+        """
+        super(AncillaryVariable, self).__init__(
```
pp-mo: An awful lot of your 'super' calls can be replaced by Python 3 argumentless ones, I think.
I.E. `super(SelfClass, self)` --> `super(self.__class__, self)` == `super()`.
Much nicer I think !
Applies a lot here, and in Coord.

lib/iris/coords.py
```diff
@@ -52,6 +52,697 @@
...
+        # with the bounds.
+        if isinstance(self, Coord):
+            if self.has_bounds():
+                bounds = self._bounds_dm.core_data()
+                _, bounds = iris.util._slice_data_with_keys(bounds, keys)
+                bounds = bounds.copy()
+            else:
+                bounds = None
+
+            copy_args = dict(points=values, bounds=bounds)
```
pp-mo: Just as this code section is the *only* place where you allow the _DimensionalMetadata code to refer to _bounds_dm,  so it is the only branch that sets up to pass a 'bounds' keyword to the copy function.
This reads oddly, as the local 'copy' definition, immediately below, does not support a bounds keyword.
So I think it is worth an extra comment here, to make it clear that 'self' here _must_ be a Coord, so that makes the difference.

pp-mo: Additional... this is ***not*** the only place where _DimensionMetadata deals with bounds : it also happens in `convert_units`.
But all the other methods, like `__getitem__`, seem to use an "override, call parent + extend" strategy instead where bounds must be handled.
Not sure what I think of this yet.

lib/iris/cube.py
```diff
@@ -2023,6 +2196,16 @@ def vector_summary(vector_coords, cube_header, max_line_offset,
...
                 summary += '\n'.join(cell_measure_summary)
 
+            #
+            # Generate summary of cube ancillary variables attribute
+            #
+            if vector_ancillary_variables:
+                ancillary_variable_summary, cube_header = vector_summary(
+                    [], cube_header, max_line_offset,
+                    ancillary_variables=vector_ancillary_variables)
+                summary += '\n     Ancillary Datasets:\n'
```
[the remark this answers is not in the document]
lbdreyer: I'd argue it should go the other way; we should update the other section headings (`Dimension coordinates -> Dimension Coordinates`) since they are headings and so should be capitalised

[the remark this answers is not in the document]
pp-mo: FWIW I lean toward @lbdreyer here : "All Title Case" makes sense to me.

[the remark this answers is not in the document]
stephenworsley: Should this say `Ancillary Variables?`.

pp-mo:
>  this method ...  pass a 'bounds' keyword to the copy function.  This reads oddly, as the local 'copy' definition, immediately below, does not support a bounds keyword.
... But all the other _DimensionMetadata methods, ...  use an "override, call parent + extend" strategy instead.
... Hmmmm. Still not quite sure what I think of this yet.

After some extensive discussion !

I also observed that the worst of the 'mixed' approach is that, where a Coord method "extends" the _DimensionalMetadata method to do "the same thing to the bounds", we have code which is effectively copied from one class to the other, which not only breaks DRY but puts the 'copy' a long way away where it is not obvious !  E.G. the worst case is probably the `__binary_operator__` method.

I think we should *either* 
  1. kick all the bounds-handling code out of the _DimensionalMetadata class, and put it all in Coord 
    * ( see the idea in https://github.com/lbdreyer/iris/pull/6 )
    *  ***or ..***
  2. include all the bounds-handling code in _DimensionalMetadata, but only support setting the bounds in the Coord class
    * ( see the idea in https://github.com/lbdreyer/iris/pull/7 )

Actually I think https://github.com/lbdreyer/iris/pull/7 is probably slicker, because then several methods don't need to be overriden in Coord at all.  But it is a bit weaselly, and takes some explaining...

lib/iris/coords.py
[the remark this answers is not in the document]
pp-mo: P.S. as seen in code change alternative suggestions [below](https://github.com/SciTools/iris/pull/3422#issuecomment-545081965), it may be easier to leave the points as a _positional_ argument, so you don't have to specify a 'points' or 'values' keyword depending on the instance type.

lib/iris/cube.py
```diff
@@ -1981,6 +2140,20 @@ def vector_summary(vector_coords, cube_header, max_line_offset,
...
                     vectors = vectors + cell_measures
+                if ancillary_variables:
+                    # Generate full textual summary for each vector ancillary
+                    # variable - WITH dimension markers.
+                    for index, av in enumerate(ancillary_variables):
+                        dims = self.ancillary_variable_dims(av)
+
+                        for dim in range(len(self.shape)):
+                            width = alignment[dim] - len(vector_summary[index])
+                            char = 'x'
```
lbdreyer: This line is incorrect. It should be 
```suggestion
                            char = 'x' if dim in dims else '-'
```

pp-mo:
Hi @lbdreyer, and anyone else watching.
I finally got round to doing some more practical checks, and checking out the testing provision.
It's now rather late to make new objections, as we really want to merge this so we can move on to subsequent tasks for the following 2-week development cycle.

So, I'm happy that there are at least _some_ tests to exercise all the new methods.  
However, I did some across some problems, and I have identified some missing testing that could usefully be included here :

  1. cube.summary test showing ancillary variable with correct dimension mapping
  1. merge example showing mismatched ancillary variables preventing merge
  1. merge example identical ancillary variables shared in result cube
  1. also concatenate equivalents of those 2 merge tests

----

The problem of cube summary getting the dimension mappings wrong I demonstrated like this:
```
        >>> cube = Cube(np.zeros((2,2,3)))
        >>> cube.add_dim_coord(DimCoord(np.arange(2), long_name='t'), 0)
        >>> cube.add_dim_coord(DimCoord(np.arange(2), long_name='x'), 1)
        >>> cube.add_dim_coord(DimCoord(np.arange(3), long_name='y'), 2)
        >>> cube.add_ancillary_variable(AncillaryVariable(np.zeros((2,3))), (1,2))
        >>> print(cube)
        unknown / (unknown)                 (t: 2; x: 2; y: 3)
             Dimension coordinates:
                  t                           x     -     -
                  x                           -     x     -
                  y                           -     -     x
             Ancillary Datasets:
                  unknown                     x     x     x
```
Something like this should be in the above test.
However, this problem is already being addressed anyway : see ["Fix cube summary of ancil vars" commit above](https://github.com/SciTools/iris/pull/3422/commits/0e5eac0e45e05899433d26f42c360dd990584594).  Nevertheless, a test exercising this will be useful..

pp-mo:
> However, I did some across some problems

It seems a considerable number of cube operations don't handle ancillary variables as they should do, of which the most worrying are ...
  * indexing ('__getitem__') discards AVs
  * cube copy discards AVs
  * equality testing (`__eq__`) ignores them

So right now, if a cube contains AVS, cube.copy() is clearly different (and prints differently) but it tests equal !

But none of this will affect cubes that *don't * contain AVs, so in the interests of progress I think we'd better ignore this for now + defer all those problems to [a new issue]() to tidy up the cube operations later.

lbdreyer:
@pp-mo I have added a test for ancillary variables in cube.summary (Note! we have very minimal testing for cube.summary in place! we should address, possibly if we also get a change to improve the html repr)

The tests for concatenate/merge made me realise that it wasn't being handled correctly (and consequently CellMeasures are also handled incorrectly when merging/concatenating). Unfortunately the work to get this done seems substantial and not something we will achieve in the next few hours. I have decided to back out the changes I previously made to _concatenate.py and _merge.py and instead fit that work into #3483

pp-mo:
I think what's done here is ok, though a lot remains to be dealt with.
I won't be happy if we don't fix the obvious problems in the next few days! #3483

PR #3423, opened 2019-09-26
Property getter setter usage extras

pp-mo:
***WIP*** testing


pp-mo:
(Just for testing stickler)

Issue #3408, opened 2019-09-20
Decide what to do about standard name modifiers units


lbdreyer:
In #3393 we included support for standard name modifiers. However, that PR, for simplicity, did not include in any special handling of the units.

In the [CF Conventions Appendix C](http://cfconventions.org/Data/cf-conventions/cf-conventions-1.7/cf-conventions.html#standard-name-modifiers) it lists the accepted Standard Name modifiers including the associated units. 

Standard name modifier | unit 
------------ | -------------
detection minimum | u 
number_of_observations | 1 
standard_error | u 
status_flag |  

`u` means that the unit it is equivalent to the unit of the unmodified standard_name (e.g. `northward_wind` has units of `m s-1` so `northward_wind standard_error` would also have units of `m s-1`).

Note only `status_flag` and `number_of_observations` have specified units. Also note that both these are _deprecated_ (See [this trac ticket](https://cf-trac.llnl.gov/trac/ticket/74#comment:52)), although only the deprecation of `status_flag` appears in the actual documentation. I think `number_of_observations` has been missed off?

The question is how to handle the units?
Should we enforce the behaviour such that if someone renames the cube from 
`air_temperature` -> `air_temperature number_of_observations`
 that it would also change the units? 

I'm concerned this may trip some people up:
Currently with standard_name, if some renames the cube, it doesn't have any impact on the units. Below we take a wind cube and rename the cube to `air_temperature` but the units stay as `m s-1` rather than converting to the canonical units of `air_temperature`
```
>>> c
<iris 'Cube' of northward_wind / (m s-1) (-- : 1)>
>>> c.rename('air_temperature')
>>> c
<iris 'Cube' of air_temperature / (m s-1) (-- : 1)>
```

pp-mo:
> Currently with standard_name, if some renames the cube, it doesn't have any impact on the units. Below we take a wind cube and rename the cube to air_temperature but the units stay as m s-1 rather than converting to the canonical units of air_temperature

From a practical point of view, the most likely usage is when a cube is copied and modified to form a template for a new cube.
So I don't think we shouldn't assign the canonical units, the correct units could be just anything compatible with the canonical unit.
How about making rename check the old + new canonical units, and if they are not compatible that set the units to None (aka 'unknown').
so for example (as possible future behaviour) ....
```
>>> cube = Cube([0.], standard_name='air_temperature', units='K')
>>> cube.rename('x_wind')
>>> cube
[...]

PR #3424, opened 2019-09-26
enforce climatological time coordinate units

bjlittle:
This PR follows-up from #3395, and enforces that the `units` of a `climatological` time coordinate must be a time reference.

### DIFF
--- a/lib/iris/tests/unit/coords/test_Coord.py
+++ b/lib/iris/tests/unit/coords/test_Coord.py
@@ -865,7 +865,7 @@ class TestClimatology(tests.IrisTest):
                          units='days since 1970-01-01',
                          climatological=True)
         emsg = 'Cannot change units of climatological time coordinate'
-        with self.assertRaisesRegex(TypeError, emsg):
+        with six.assertRaisesRegex(self, TypeError, emsg):
             coord.units = 'm'
 
     def test_units__change_valid_not_climatological(self):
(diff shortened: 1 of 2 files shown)


lib/iris/tests/unit/coords/test_Coord.py
```diff
@@ -852,6 +852,31 @@ def test_remove_bounds(self):
...
+        new_units = 'hours since 1970-01-01'
+        coord.units = new_units
+        self.assertEqual(str(coord.units), new_units)
+
+    def test_units__change_invalid_climatological(self):
+        coord = DimCoord(points=2, bounds=[[1, 3]],
+                         units='days since 1970-01-01',
+                         climatological=True)
+        emsg = 'Cannot change units of climatological time coordinate'
+        with self.assertRaisesRegex(TypeError, emsg):
```
trexfeathers: @bjlittle until we're exclusively Py3 we need to use **six**.assertRaisesRegex

lbdreyer:
As discussed offline, I do think we need to be a bit careful before we commit to an approach like this (as with #3408).
I'm concerned that this strictness will just become a pain point for users and force them to work in a certain way. I'd prefer for Iris to be more flexible when you're within Iris, but, so as to ensure we write CF compliant files, to enforce the strictness at save time.
It also seems a bit of a strange move since we're considering to make cube.arithmetic less strict, at the same to go the other way on things like this.

PR #3425, opened 2019-09-26
Climatological xml

bjlittle:
Follow-up from #3395, adding changes to support the `climatological` property in our `cml` test output.

### DIFF
--- a/lib/iris/coords.py
+++ b/lib/iris/coords.py
@@ -1712,6 +1712,8 @@ class Coord(six.with_metaclass(ABCMeta, CFVariableMixin)):
         if self.var_name:
             element.setAttribute('var_name', str(self.var_name))
         element.setAttribute('units', repr(self.units))
+        if self.climatological:
+            element.setAttribute('climatological', str(self.climatological))
 
         if self.attributes:
             attributes_element = doc.createElement('attributes')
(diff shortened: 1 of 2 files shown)


lib/iris/coords.py
```diff
@@ -1712,6 +1712,8 @@ def xml_element(self, doc):
         if self.var_name:
             element.setAttribute('var_name', str(self.var_name))
         element.setAttribute('units', repr(self.units))
+        if self.climatological:
+            element.setAttribute('climatological', str(self.climatological))
```
lbdreyer: Why put this here and not near the addition of the `coord_system` attribute one [line 1729](https://github.com/SciTools/iris/blob/ede76caa7fec8b664df1e7a32dd9d48eeee6e06a/lib/iris/coords.py#L1729)
As it so happens it doesn't matter as the cml is alphabetical and not that it should be a blocker to this PR going in but I just found that a bit strange.

lbdreyer: I now understand: This `climatological` is being added as an attribute, whereas the coordinate system is a separate child.
Usefully with the alphabetical order it means that `climatological` comes right after `bounds`

bjlittle:
@lbdreyer All good this side, go for it :+1:

PR #3426, opened 2019-09-30
only build docs for latest python

bjlittle:
We're building the docs twice once for `Python3.6` and once for `Python3.7`.

Don't see the point in this... so let's default to building the docs only for the latest Python.


pp-mo:
Fully :+1: with this.  Just waiting for Travis to check it out.

bjlittle:
Nice one, thanks @pp-mo :tada:

PR #3427, opened 2019-09-30
enforce non-masked dimension coordinate points/bounds

bjlittle:
This PR addresses #3419.

### DIFF
--- a/docs/iris/src/userguide/interpolation_and_regridding.rst
+++ b/docs/iris/src/userguide/interpolation_and_regridding.rst
@@ -185,8 +185,8 @@ For example, to mask values that lie beyond the range of the original data:
    >>> scheme = iris.analysis.Linear(extrapolation_mode='mask')
    >>> new_column = column.interpolate(sample_points, scheme)
    >>> print(new_column.coord('altitude').points)
-   [       nan  494.44452  588.8889   683.33325  777.77783  872.2222
-     966.66675 1061.1111  1155.5554         nan]
+   [-- 494.44451904296875 588.888916015625 683.333251953125 777.77783203125
+    872.2222290039062 966.666748046875 1061.111083984375 1155.555419921875 --]
 
 
 .. _caching_an_interpolator:
--- a/lib/iris/coords.py
+++ b/lib/iris/coords.py
@@ -625,7 +625,8 @@ class Coord(six.with_metaclass(ABCMeta, CFVariableMixin)):
             result = np.require(src, requirements='W')
             # Ensure the array has enough dimensions.
             # NB. Returns the *same object* if result.ndim >= ndmin
-            result = np.array(result, ndmin=ndmin, copy=False)
+            func = ma.array if ma.isMaskedArray(result) else np.array
+            result = func(result, ndmin=ndmin, copy=False)
             # We don't need to copy the data, but we do need to have our
             # own view so we can control the shape, etc.
             result = result.view()
@@ -1930,17 +1931,23 @@ class DimCoord(Coord):
         Confirm that a new set of coord points adheres to the requirements for
         :class:`~iris.coords.DimCoord` points, being:
             * points are scalar or 1D,
-            * points are numeric, and
+            * points are numeric,
+            * points are not masked, and
             * points are monotonic.
 
         """
         if points.ndim not in (0, 1):
-            raise ValueError(
-                'The points array must be scalar or 1-dimensional.')
+            emsg = 'The {!r} {} points array must be scalar or 1-dimensional.'
+            raise ValueError(emsg.format(self.name(), self.__class__.__name__))
         if not np.issubdtype(points.dtype, np.number):
-            raise ValueError('The points array must be numeric.')
+            emsg = 'The {!r} {} points array must be numeric.'
+            raise ValueError(emsg.format(self.name(), self.__class__.__name__))
+        if ma.isMaskedArray(points) and ma.count_masked(points):
+            emsg = 'A {!r} {} points array must not be masked.'
+            raise TypeError(emsg.format(self.name(), self.__class__.__name__))
         if points.size > 1 and not iris.util.monotonic(points, strict=True):
-            raise ValueError('The points array must be strictly monotonic.')
+            emsg = 'The {!r} {} points array must be strictly monotonic.'
+            raise ValueError(emsg.format(self.name(), self.__class__.__name__))
 
     @Coord.points.setter
     def points(self, points):
@@ -1948,10 +1955,13 @@ class DimCoord(Coord):
         # Ensure it is an actual array, and also make our own copy so that we
         # can make it read-only.
         points = _lazy.as_concrete_data(points)
-        points = np.array(points)
+        # Make sure that we have an array (any type of array).
+        points = np.asanyarray(points)
 
         # Check validity requirements for dimension-coordinate points.
         self._new_points_requirements(points)
+        # Cast to a numpy array for masked arrays with no mask.
+        points = np.array(points)
 
         # Call the parent points setter.
         super(DimCoord, self.__class__).points.fset(self, points)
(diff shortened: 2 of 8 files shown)


bjlittle:
Happy to add more tests as a review action...

lib/iris/coords.py
```diff
@@ -625,7 +625,8 @@ def _sanitise_array(self, src, ndmin):
             result = np.require(src, requirements='W')
             # Ensure the array has enough dimensions.
             # NB. Returns the *same object* if result.ndim >= ndmin
-            result = np.array(result, ndmin=ndmin, copy=False)
+            func = ma.array if ma.isMaskedArray(result) else np.array
+            result = func(result, ndmin=ndmin, copy=False)
```
bjlittle: As @rcomer points out, this is indeed the nub of the issue...

lib/iris/coords.py
```diff
@@ -1930,28 +1931,37 @@ def _new_points_requirements(self, points):
...
-                'The points array must be scalar or 1-dimensional.')
+            emsg = 'The {!r} {} points array must be scalar or 1-dimensional.'
+            raise ValueError(emsg.format(self.name(), self.__class__.__name__))
         if not np.issubdtype(points.dtype, np.number):
-            raise ValueError('The points array must be numeric.')
+            emsg = 'The {!r} {} points array must be numeric.'
+            raise ValueError(emsg.format(self.name(), self.__class__.__name__))
+        if ma.isMaskedArray(points) and ma.count_masked(points):
+            emsg = 'A {!r} {} points array must not be masked.'
+            raise TypeError(emsg.format(self.name(), self.__class__.__name__))
```
bjlittle: Raise a `TypeError` exceptions here to explicitly catch those masked points. Hence the need for the change [here](https://github.com/SciTools/iris/pull/3427/files#diff-07b1c9cb6b22e14b7f314da7af652aafR394)

docs/iris/src/userguide/interpolation_and_regridding.rst
```diff
@@ -185,8 +185,8 @@ For example, to mask values that lie beyond the range of the original data:
    >>> scheme = iris.analysis.Linear(extrapolation_mode='mask')
    >>> new_column = column.interpolate(sample_points, scheme)
    >>> print(new_column.coord('altitude').points)
-   [       nan  494.44452  588.8889   683.33325  777.77783  872.2222
-     966.66675 1061.1111  1155.5554         nan]
+   [-- 494.44451904296875 588.888916015625 683.333251953125 777.77783203125
+    872.2222290039062 966.666748046875 1061.111083984375 1155.555419921875 --]
```
bjlittle: This really makes sense to me now...

We've asked for an `extrapolation_mode=mask`, so now we get a `masked` array for the `altitude` derived coordinate for the extrapolated points, and not just an array with `nan`s injected.

If the `extrapolation_mode=nan`, then we (quite rightly) get:
```
[       nan  494.44452  588.8889   683.33325  777.77783  872.2222
  966.66675 1061.1111  1155.5554         nan]
```

bjlittle:
@pp-mo Yup, I definitely want to add some extra test coverage. I think that our masked array test coverage (in gerneral) is pretty poor (at least an after thought).

I just want to convince myself that I've not broken anything here for `DimCoord`s... as it's pretty fundamental to the core of `iris` :worried:

lib/iris/coords.py
```diff
@@ -1930,28 +1932,37 @@ def _new_points_requirements(self, points):
...
         if points.ndim not in (0, 1):
-            raise ValueError(
-                'The points array must be scalar or 1-dimensional.')
+            emsg = 'The {!r} {} points array must be scalar or 1-dimensional.'
+            raise ValueError(emsg.format(self.name(), self.__class__.__name__))
         if not np.issubdtype(points.dtype, np.number):
-            raise ValueError('The points array must be numeric.')
+            emsg = 'The {!r} {} points array must be numeric.'
+            raise ValueError(emsg.format(self.name(), self.__class__.__name__))
+        if ma.isMaskedArray(points) and ma.count_masked(points):
```
pp-mo: Should probably be using 'is_masked' for this : https://docs.scipy.org/doc/numpy/reference/generated/numpy.ma.is_masked.html

lib/iris/coords.py
```diff
@@ -1969,20 +1980,27 @@ def _new_bounds_requirements(self, bounds):
...
+                    'points.shape + (n_bounds)')
+            raise ValueError(emsg.format(self.name(), self.__class__.__name__))
+        # Checks for numeric.
         if not np.issubdtype(bounds.dtype, np.number):
-            raise ValueError('The bounds array must be numeric.')
-
+            emsg = 'The {!r} {} bounds array must be numeric.'
+            raise ValueError(emsg.format(self.name(), self.__class__.__name__))
+        # Check not masked.
+        if ma.isMaskedArray(bounds) and ma.count_masked(bounds):
```
pp-mo: As above, use 'is_masked'

rcomer:
One thought that springs to mind: do methods such as `cube.extract` and `cube.aggregated_by` behave sensibly if presented with masked points?

lib/iris/coords.py
pp-mo: This bit will applies to AuxCoords rather than DimCoords, which was the point of the original issue.
I'll try and remember to edit the squashed commit for that.

pp-mo:
> One thought that springs to mind: do methods such as cube.extract and cube.aggregated_by behave sensibly if presented with masked points?

I think we have considered masking in most operations now, but it is not all tested.
Specifically, extract should be fine, aggregation does work (I just checked), but it is not tested.

lib/iris/coords.py
bjlittle: Yup, thanks @pp-mo ... I always forget that pattern, and it gives me the same behaviour that we want, namely:
```python
>>> import numpy as np
>>> import numpy.ma as ma
>>> a = np.arange(5)
>>> b = ma.arange(5)
>>> c = ma.arange(5)   
>>> c[0] = ma.masked
>>> ma.is_masked(a)
False
>>> ma.is_masked(b)
False
>>> ma.is_masked(c)
True
```

bjlittle: @pp-mo Done 👍

lib/iris/coords.py
bjlittle: @pp-mo Done 👍

bjlittle:
> All makes good sense, but clearly we have no testing for these behaviours : e.g.
> 
>     * errors for bad creation / assignment
> 
>     * successful AuxCoord creation with masked data
> 
>     * error making DimCoord with masked points
> 
>     * success making DimCoord from masked array with no masked points.
> 
> 
> Maybe we should address these, just as regression checks ?

@pp-mo After pushing commits https://github.com/SciTools/iris/pull/3427/commits/9f9a553b3936467d5a2271ea68fe4ed739cd6a4e and https://github.com/SciTools/iris/pull/3427/commits/91ceae91c7a618f26c9ad9404c723f66a2df0640, I'm pretty satified that we've got reasonable test coverage for both `DimCoord`s and `AuxCoord`s given the areas you highlighted.

Did we want to extend this further?

bjlittle:
This need a `whatsnew` entry...

lib/iris/tests/unit/coords/__init__.py
```diff
@@ -27,11 +27,12 @@
...
+import numpy.ma as ma
 
 from iris._lazy_data import is_lazy_data
 
 
-def setup_test_arrays(self, shape):
+def setup_test_arrays(self, shape, masked=False):
     # Create standard coordinate points and bounds test arrays,
     # given a desired coord shape.
     # Also create lazy versions, and save all on the 'self' object.
```
pp-mo: I think I'd like an extra comment here saying just something like 
```
  # If 'masked', also add masked arrays with some or no masked points,
  # for both points and bounds, lazy and real.
```

lib/iris/tests/unit/coords/__init__.py
```diff
@@ -27,11 +27,12 @@
 
 import dask.array as da
 import numpy as np
+import numpy.ma as ma
 
 from iris._lazy_data import is_lazy_data
 
 
-def setup_test_arrays(self, shape):
```
pp-mo: It's questionable whether we need to bother switching 'masked' on + off at all.
We could just *always* do the masked stuff ...
On my desktop, tested with `python -m unittest discover -v iris.tests.unit.coords` 
 * with 'masked' switch takes 0.560 - 0.570
 * with switching disabled (mask==True), takes 0.635-0.650 secs

So the cost is jolly small !!

lib/iris/tests/unit/coords/test_AuxCoord.py
```diff
@@ -639,5 +640,86 @@ def test_nanbounds_eq_self(self):
         self.assertEqual(co1, co1)
 
 
+class Test_masked__creation(tests.IrisTest, CoordTestMixin):
```
pp-mo: I think strictly, all this should go in the _existing_ `Test__init__`, at the top of the file.
That is the 'standard' place, and they would be easier to find.
Though if so, a comment introducing these ones + separating them off would also be handy.

But, did you actively choose _not_ to do that, for other purposes ??

lib/iris/tests/unit/coords/test_DimCoord.py
```diff
@@ -482,5 +484,82 @@ def test_copy_array(self):
         self.assertEqual(coord.bounds[1, 1], 5)
 
 
+class Test_masked__creation(tests.IrisTest, CoordTestMixin):
```
pp-mo: As in `test_AuxCoord.py` (see previous note) ...
I think this could all go in the existing `Test__init__`.

lib/iris/tests/unit/coords/__init__.py
bjlittle: Yup get your point, but I'm just going to leave it as is, if you don't mind :stuck_out_tongue_winking_eye:

lib/iris/tests/unit/coords/test_AuxCoord.py
bjlittle: @pp-mo I did actively choose not to include it in the `Test__init__` class, simply because it was dealing with masking specifically... but I don't want to split hairs here, so I'm totally happy to combine `Test_masked__creation` into `Test__init__`

bjlittle:
@pp-mo Over to you :+1:

pp-mo:
Nice @bjlittle 
:+1: for minimal fuss over this one !

PR #3428, opened 2019-10-01
Allow non-float cell measures

lbdreyer:
Currently, we do not allow the creation of cell measures with non-float data as it is claimed that this is not supported until CF1.7

As of #3406 we are now committed to CF1.7 so we can now support this.

It is worth mentioning that I struggled to find reference to the 
> Non-floating cell measures are not valid up to CF v1.7

in the CF documenstation. But regardless, we can remove this restriction now.

### DIFF
--- a/lib/iris/coords.py
+++ b/lib/iris/coords.py
@@ -2077,13 +2077,13 @@ class CellMeasure(six.with_metaclass(ABCMeta, CFVariableMixin)):
         Kwargs:
 
         * standard_name:
-            CF standard name of the coordinate.
+            CF standard name of the cell measure.
         * long_name:
-            Descriptive name of the coordinate.
+            Descriptive name of the cell measure.
         * var_name:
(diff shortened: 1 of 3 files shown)


lib/iris/coords.py
```diff
@@ -2092,16 +2092,16 @@ def __init__(self, data, standard_name=None, long_name=None,
...
-        #: Descriptive name of the coordinate.
+        #: Descriptive name of the cell measure.
         self.long_name = long_name
 
-        #: The netCDF variable name for the coordinate.
+        #: The netCDF variable name for the cell measure.
         self.var_name = var_name
 
-        #: Unit of the quantity that the coordinate represents.
+        #: Unit of the quantity that the cell measure represents.
```
bjlittle: @lbdreyer Nice spot... clearly cut'n paste syndrom 🤣

bjlittle:
@lbdreyer Nice one, looks good to me 👍

PR #3431, opened 2019-10-01
ENH: Working cube data equality

cpelley:
Replacing #3250 but targeting master.


pp-mo:
Hi @cpelley 
Regarding : #3433
It looks like this needs to ask for a later version of dask.
I have no idea why that doesn't happen automatically, but it seems we can fix it.

Meanwhile, this branch also needs to cut out the addition to `whatsnew/2.2.rst` as that is no longer
the appropriate place for it.

Rather than steal more of your time, I'm going to just merge this and retrospectively fix those things.
Thanks for the input !

cpelley:
Thanks @pp-mo

PR #3433, opened 2019-10-01
Daskversionfix 1v2

pp-mo:
**WIP** do not merge

### DIFF
--- a/docs/iris/src/whatsnew/2.2.rst
+++ b/docs/iris/src/whatsnew/2.2.rst
@@ -85,6 +85,8 @@ Iris 2.2 Dependency updates
 Bugs Fixed
 ==========
 
+* Cube equality of boolean data is now handled correctly.
+
 * The bug has been fixed that prevented printing time coordinates with bounds
   when the time coordinate was measured on a long interval (that is, ``months``
   or ``years``).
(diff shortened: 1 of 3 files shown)


lbdreyer:
It works ... So I wonder why #3431 doesn't pick up dask 1.2 ? :confused:

pp-mo:
>  I wonder why #3431 doesn't pick up dask 1.2 ?

😕 indeed.  N.B. the commit is exactly the same as the first one here.

However, it _does_ seem to be an adequate fix ...

pp-mo:
> why #3431 doesn't pick up dask 1.2 ?

Comparing the `conda list` output, the ***only*** differences are ...

| package | "OLD" pin dask>=1.1 | | | // |  "NEW" pin dask>=1.2  | | | |
| ---- | - | - | - | - | ---- | - | - | - |
| dask (AND dask-core) | 1.1.4 | py27_1 | defaults | //  | 1.2.2 | py_3 | conda-forge |
| pyproj | 1.9.6 | py27h14380d9_0  | defaults |  // | 1.9.6 | py27h516909a_1002 | conda-forge |

In fact that's so odd I don't believe it.
Maybe the underlying channels have  just shifted ??
I will re-spin the original checks ...

pp-mo:
> I will re-spin the original checks ...

Nope, still does the same.
:sigh:

PR #3434, opened 2019-10-01
Fix handling of cell measures during transposes of cubes

lbdreyer:
Currently cell measures are not handled when you do a transpose of a cube. 
```
>>> c = Cube(np.arange(6).reshape(2,3), 'x_wind')
>>> cm = CellMeasure(np.arange(6).reshape(2,3), measure='area')
>>> c.add_cell_measure(cm, (0,1))
>>> print(c._cell_measures_and_dims)
[[CellMeasure(array([[0, 1, 2],
       [3, 4, 5]]), measure=area, standard_name=None, units=Unit('1')), (0, 1)]]
>>> c.transpose()
>>> print(c._cell_measures_and_dims)
[[CellMeasure(array([[0, 1, 2],
       [3, 4, 5]]), measure=area, standard_name=None, units=Unit('1')), (0, 1)]]
```
Where I would expect the last line to return:

```
>>> print(c._cell_measures_and_dims)
[[CellMeasure(array([[0, 1, 2],
       [3, 4, 5]]), measure=area, standard_name=None, units=Unit('1')), (1, 0)]]
```
i.e. the dims swapped, as it does with auxiliary coords.

This PR includes the handling of the above, plus tests.

### DIFF
--- a/lib/iris/cube.py
+++ b/lib/iris/cube.py
@@ -2819,17 +2819,22 @@ bound=(1994-12-01 00:00:00, 1998-12-01 00:00:00)
 
         dim_mapping = {src: dest for dest, src in enumerate(new_order)}
 
-        def remap_dim_coord(coord_and_dim):
-            coord, dim = coord_and_dim
-            return coord, dim_mapping[dim]
-        self._dim_coords_and_dims = list(map(remap_dim_coord,
-                                             self._dim_coords_and_dims))
+        # Remap all cube dimensional metadata (dim and aux coords and cell
+        # measures).
+        def remap_cube_metadata(metadata_and_dims):
+            metadata, dims = metadata_and_dims
+            if isinstance(dims, Iterable):
+                dims = tuple(dim_mapping[dim] for dim in dims)
+            else:
+                dims = dim_mapping[dims]
+            return metadata, dims
 
-        def remap_aux_coord(coord_and_dims):
-            coord, dims = coord_and_dims
-            return coord, tuple(dim_mapping[dim] for dim in dims)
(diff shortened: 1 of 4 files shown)


lib/iris/cube.py
```diff
@@ -2825,11 +2825,15 @@ def remap_dim_coord(coord_and_dim):
         self._dim_coords_and_dims = list(map(remap_dim_coord,
                                              self._dim_coords_and_dims))
 
-        def remap_aux_coord(coord_and_dims):
+        # Remaps all cube dimensional metadata that may have multiple
+        # dimensions.
+        def remap_cube_metadata(coord_and_dims):
```
bjlittle: @lbdreyer Is there an opportunity here to take a step further and combine `remap_dim_coord` and `remap_cube_metadata`? e.g., some thing along the lines of:
```python
...
def remap_cube_metadata(coord_and_dims):
    coord, dims = coord_and_dims
    if isinstance(dims, Iterable):
        dims = tuple(dim_mapping[dim] for dim in dims)
    else:
        dims = dim_mapping[dims]
    return coord, dims
```
Then deriving both `self._dim_coords_and_dims`, `self._aux_coords_and_dims` and `self._cell_measures_and_dims` all call the same helper function?

bjlittle:
@lbdreyer Minor comment to think about...

Also, `travis-ci` is failing when it attempts to build the `whatsnew`. The reason (I think) is that you need to add a `.txt` to your new whatsnew contribution file.

bjlittle:
@lbdreyer Awesome, thanks :+1:

PR #3435, opened 2019-10-02
Boolean data compare - additional fixes

pp-mo:
Resolve some problems resulting from #3431 ...

 * whatsnew entry moved
 * needs dask>=1.2 (see #3433)
 * tests for new data-equality behaviour
 * whats new for behaviour change

### DIFF
--- a/docs/iris/src/whatsnew/contributions_2.3.0/newfeature_2019-Oct-02_cube_equality_tolerance.txt
+++ b/docs/iris/src/whatsnew/contributions_2.3.0/newfeature_2019-Oct-02_cube_equality_tolerance.txt
@@ -1,0 +1,3 @@
+* Cube data equality testing (and hence cube equality) now uses a more relaxed
+  tolerance, thus some cubes may now test 'equal' that previously did not.
+  The new behaviour is the default for :func:`numpy.allclose`.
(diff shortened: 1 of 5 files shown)


lib/iris/tests/unit/cube/test_Cube.py
```diff
@@ -1905,5 +1905,57 @@ def test_preserves_lazy(self):
         self.assertArrayAllClose(cube.data, real_data_ft)
 
 
+class Test__eq__data(tests.IrisTest):
+    """Partial cube equality testing, for data type only."""
+    def _check_data_eq(self, data1, data2):
```
lbdreyer: I can't see where this is being used?

lib/iris/tests/unit/cube/test_Cube.py
```diff
@@ -1905,5 +1905,57 @@ def test_preserves_lazy(self):
...
+        cube1 = Cube(data1)
+        cube2 = Cube(data2)
+        self.assertEqual(cube1 == cube2,
+                         np.all(data1 == data2))
```
lbdreyer: I can't see where this is being used?

docs/iris/src/whatsnew/contributions_2.3.0/newfeature_2019-Oct-02_cube_equality_tolerance.txt
```diff
@@ -0,0 +1,3 @@
+* Cube data equality testing (and hence cube equality) now uses a more relaxed
+  tolerance, thus some cubes may now test 'equal' that previously did not.
+  The new behaviour is the default for :func:`numpy.allclose`.
```
lbdreyer: If we are choosing to go down this path of relaxing the tolerance, I think it would helpful to our users to explain more specifically what has changed.
As a user, I would like to know what the behaviour was before (i.e. only absolute tolerance of 1e-8) and what the behaviour is now (as you have done with linking of to allclose

pp-mo: Ok!

lbdreyer:
Looks good! Thanks @pp-mo 
Once travis is happy I will merge

PR #3442, opened 2019-10-03
Additional whatsnew contributions for v2.3 release.

pp-mo:
Some extra 'whatsnew' information for the 2.3 release.

Mostly covering changes that were merged without including a whatsnew contribution.
:disappointed: we should work on this !

PR #3444, opened 2019-10-03
List whitepapers

pp-mo:
**WIP do not merge**
(currently based on top of #3442 commits).
Just a placeholder, for now **(treat this as an issue)**. 

Thought this might be a useful idea to promote the technical discussions (of which we could usefuly do more ?)
The intended effect is simply to list the individual titles under the mainpage section.
This probably isn't at all the right way to do it though.


bjlittle:
@pp-mo This PR has unfortunately gone stale. What's the plan here to move this forward into something that we can merge?

pp-mo:
> What's the plan here

It's just an idea/suggestion, not a proposal.  I don't really understand why it contains extra commits that don't belong here.  Will close + replace with an issue for now.

bjlittle:
@pp-mo Cool, thanks.

PR #3446, opened 2019-10-03
disable python2 travis-ci

bjlittle:
This PR removes Python2 `travis-ci` testing :tada: 

As a part of technical debt, the appropriate changes are now in place within the `.travis.yml` to commence `iris-grib` tests.... however, there are 5 errors and 10 failures that require to be addressed.

I'll raise a separate issue to cover this, see https://github.com/SciTools/iris/issues/3447.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -13,10 +13,6 @@ env:
     # The decryption key for the encrypted .github/deploy_key.scitools-docs.enc.
     - secure: "N9/qBUT5CqfC7KQBDy5mIWZcGNuUJk3e/qmKJpotWYV+zwOI4GghJsRce6nFnlRiwl65l5oBEcvf3+sBvUfbZqh7U0MdHpw2tHhr2FSCmMB3bkvARZblh9M37f4da9G9VmRkqnyBM5G5TImXtoq4dusvNWKvLW0qETciaipq7ws="
   matrix:
-    - PYTHON_VERSION=2.7 TEST_TARGET=default TEST_MINIMAL=true
-    - PYTHON_VERSION=2.7 TEST_TARGET=default
-    - PYTHON_VERSION=2.7 TEST_TARGET=example
-
     - PYTHON_VERSION=3.6 TEST_TARGET=default TEST_MINIMAL=true
     - PYTHON_VERSION=3.6 TEST_TARGET=default
     - PYTHON_VERSION=3.6 TEST_TARGET=example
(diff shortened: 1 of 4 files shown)


.travis.yml
```diff
@@ -112,12 +100,12 @@ install:
...
   # TODO : remove when iris doesn't do an integration test requiring iris-grib.
-  - if [[ "${TEST_MINIMAL}" != true && ${PYTHON_VERSION} == 2* ]]; then
-        conda install --quiet -n ${ENV_NAME} python-ecmwf_grib;
-        pip install git+https://github.com/SciTools/iris-grib.git@v0.11.0;
-    fi
+# TODO: uncomment and address the 5 failures and 10 errors in iris-grib.
+#  - if [[ "${TEST_MINIMAL}" != true ]]; then
+#        conda install --quiet -n ${ENV_NAME} python-eccodes;
+#        conda install --quiet -n ${ENV_NAME} --no-deps iris-grib;
+#    fi
```
bjlittle: Pull the pin and fix these `iris-grib` related issues...

pp-mo:
Hi @bjlittle Just to check I've understood this...

To actually re-enable the iris-grib tests, we need to also insert a mention of iris-grib into the `requirements/all.txt` file, 
as `" all"`  is what gets added to `$CONDA_REQS_GROUPS` within `.travis.yml`, when it runs without the `TEST_MINIMAL` flag .

***Yes?***

pp-mo:
**P.S.**

To install iris-grib under Python3 for testing, we will *also* need to install eccodes-python explicitly with pip, as iris-grib currently does with [this conda requirement](https://github.com/SciTools/iris-grib/blob/v0.14.1/.travis.yml#L22) and [this pip install](https://github.com/SciTools/iris-grib/blob/v0.14.1/.travis.yml#L31)

***Yes?***

bjlittle:
@pp-mo This is a bit of a mess, and we need to sort this out, but I'd highly recommend that is done outside this PR...

Anyways, in an attempt to clarify the `iris-grib` dependency actually lives in the `requirements/extensions.txt`. However, the `extensions.txt` is **not** referenced anywhere. It may have been at some point previously, either implicitly baked into the `requirements/gen_conda_requirements.py` or within the `.travis.yml`, but certainly not now. Infact, within the `extensions.txt` there is an empty `#conda:` directive, which means that the `requirements/gen_conda_requirements.py` will skip this for `conda`, but not `pip`.

This leads me to believe that `iris-grib` was dealt with in a bespoke way within the `.travis.yml`, as it is kinda now. The other factor is that if we use `conda` to install `iris-grib` then we need to do so with `--no-deps` otherwise the PR version of `iris` that `travis-ci` is meant to be testing is clobbered by the `conda-forge` version of `iris` pulled in by `iris-grib` - and the fun factor there is that the `conda-forge` version of `iris` has all the `iris.tests` and `iris.tests.results` ripped out - hence all the test fail! 

Anyways, I can see a way forwards using either using `pip` or `conda` for `iris-grib` installation within the `.travis.yml`. This is probably all better highlighted in a concrete PR rather than described in imprecise prose - at least there would be a concrete thing to poke at and ponder over. Happy to take the first step and iterate with you... if you're game?

Otherwise, I'd suggest we leave the whole debate for later outside this PR. Are you trusting enough to believe me that there is nothing to worry about here? :smiling_imp: 

So `iris-grib` trauma aside, is this PR good to go?

pp-mo:
@bjlittle thanks for the clarity.


> PR good to go?

Yes, given that we all understand that grib testing is still missing, and that's the way we want it !

bjlittle:
@pp-mo Awesome, thanks :+1:

PR #3449, opened 2019-10-04
Undo deprecation warning - should be regular warning?

lbdreyer:
Whilst looking for deprecations that needed removing in the Iris 3 release, I came across what looks like something that shouldn't be deprecated?

I'm not sure! So I thought I would raise it.

It was originally a regular warning, but #1993 changed it to a deprecation warning

### DIFF
--- a/lib/iris/analysis/calculus.py
+++ b/lib/iris/analysis/calculus.py
@@ -26,11 +26,11 @@ from six.moves import (filter, input, map, range, zip)  # noqa
 import six
 
 import re
+import warnings
 
 import cf_units
 import numpy as np
 
-from iris._deprecation import warn_deprecated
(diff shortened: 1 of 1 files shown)


bjlittle:
@pp-mo Unfortunately, your name is against #1993... care to cast your mind back to 2016 and comment, if humanly possible?

pp-mo:
Nice spot @lbdreyer !
I think this is a plain error from the original PR #1993.  
Probably from an over-zealous use of search+replace.

I don't think this has anything to do with a deprecation.
It's also a pretty useless message, as-is, as it doesn't provide any context for what is wrong and only makes sense in reference to the API of a private internal routine.
 :missing emoji for cupping brow with closed eyes:

PR #3452, opened 2019-10-04
Improve warning and error messages for old routine.

pp-mo:
This is a minimal improvement to the problems identified in #3449 with the error + warning messages produced by `iris.analysis.calculus._construct_midpoint_coord`.

Frankly there's a lot more that's wrong with this + I almost wish I hadn't bothered :  
  * the use of this private routine as a utility in the tests for another routine (_construct_delta_coord) is pretty horrible
  * the optional use of the 'circular' keyword is *only* used by the test code, not the main iris codebase : So it could, and  should be much simpler !
  * the routine raises an error if called with circular=True when coord.circular does not exist, but it only emits a *warning* if coord.circular==False  ((arrrrgh!))

### DIFF
--- a/lib/iris/analysis/calculus.py
+++ b/lib/iris/analysis/calculus.py
@@ -85,14 +85,16 @@ def _construct_midpoint_coord(coord, circular=None):
 
     """
     if circular and not hasattr(coord, 'circular'):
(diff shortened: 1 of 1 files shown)


bjlittle:
omg 🤪

Some corners of `iris` are just distressing...

PR #3454, opened 2019-10-07
Mergeback v2.3.x branch onto master

lbdreyer:
Now release candidate is cut, this PR is intended to mergeback the changes (as of yet) in the v2.3.x branch onto master.

This includes removing the `docs/iris/src/whatsnew/contributions_2.3.0` directory and replaces it with the `2.3.rst` 

This requires being followed up by a PR that includes a `docs/iris/src/whatsnew/contributions_3.0.0` directory


lbdreyer:
Note the SHAs are different! 
I wasn't sure how to do this in a way that also keeps the SHAs, since they are based on the parent commit which differs between `master` and `v2.3.x` (which each had PR to change the version string)

PR #3455, opened 2019-10-07
Make PointInCell a public member of iris.analysis.

pp-mo:
This fixes a problem I found with the 2.3.0rc0 release candidate ...

The new public PointInCell scheme did not appear in the docs (and the link in the whatsnew was broken), because it is not public : 
Because `iris.analysis.__init__` defines an '__all__', the PointInCell must be added to that list too, to be "public".  
As Sphinx also respects that, it didn't appear in the docs either.


lbdreyer:
Thanks @pp-mo 
I think that may explain a few issues we were having with mock testing!

PR #3458, opened 2019-10-10
tidy requirements and site.cfg

bjlittle:
This PR tidies the dependencies of `iris` now that it is exclusively Python3.

### DIFF
--- a/lib/iris/etc/site.cfg.template
+++ b/lib/iris/etc/site.cfg.template
@@ -1,8 +1,6 @@
 [System]
-udunits2_path = /path/to/libudunits2.so
 dot_path = /path/to/dot # see iris.fileformats.dot
 
 [Resources]
 sample_data_dir = /path/to/iris/resources/sample_data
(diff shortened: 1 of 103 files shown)


lib/iris/etc/site.cfg.template
```diff
@@ -1,8 +1,6 @@
 [System]
-udunits2_path = /path/to/libudunits2.so
```
bjlittle: This is now no longer used.

lib/iris/tests/__init__.py
```diff
@@ -53,16 +53,12 @@
...
+from unittest import mock
 import threading
 import warnings
 import xml.dom.minidom
 import zlib
 
-try:
-    from unittest import mock
-except ImportError:
-    import mock
```
bjlittle: Now use `unittest.mock` from Python3

lbdreyer:
Thanks @bjlittle !

Issue #3443, opened 2019-10-03
Address deprecations


abooton:
* Deprecations to be identified and listed
* Then deprecated as appropriate.

bjlittle:
@abooton thanks for raising this issue to collate deprecations... but what does `[M]` mean?

lbdreyer:
List of deprecations that need removing:
- [ ] iris.Future.cell_datetime_objects (deprecated Iris 2.0.0)
- [ ] iris.Future.netcdf_promote (deprecated Iris 2.0.0)
- [ ] iris.Future.netcdf_no_unlimited (deprecated Iris 2.0.0)
- [ ] iris.Future.clip_latitudes (deprecated Iris 2.0.0)
- [ ] iris.fileformats.pp._LBProc.flags attribute (deprecated 1.7.2? #1415)

stephenworsley:
Removing all these flags will mean the Future class has no flags. This will break some of the tests in test_Future.py until another flag is added. The sollution I'm currently working on is to keep these in but to skip these tests so that we retain the structure of the tests.

lbdreyer:
All deprecations have been addressed in #3459 and #3464 and #3461 
so this can now be closed

PR #3459, opened 2019-10-11
deprecated future flags

stephenworsley:
Removes the deprecated flags from iris.Future (while keeping the skeleton of Future as a template).

This addresses several of the deprecations in #3443.


lbdreyer:
@stephenworsley Could you add a whatsnew item for the removal of these deprecations

pp-mo:
Thanks @stephenworsley all looks good :100: :bouquet: !

Issue #3356, opened 2019-07-11
Tidy iris.tests.unit.lazy_data.test_co_realise_cubes


bjlittle:
Reference https://github.com/SciTools/iris/pull/3355

Require to purge the `dask<2` split in the test_co_realise_cubes

pp-mo:
#3460 fixes

PR #3460, opened 2019-10-11
Drop Dask 1 (tidy up test_co_realise_cubes)

stephenworsley:
As per #3356, iris.tests.unit.lazy_data.test_co_realise_cubes no longer tests on versions of dask less than 2. dask>=2 is added as a requirement.

### DIFF
--- a/lib/iris/tests/unit/lazy_data/test_co_realise_cubes.py
+++ b/lib/iris/tests/unit/lazy_data/test_co_realise_cubes.py
@@ -72,9 +72,6 @@ class Test_co_realise_cubes(tests.IrisTest):
         self.assertTrue(cube_inner.has_lazy_data())
 
     def test_combined_access(self):
-        import dask
-        from distutils.version import StrictVersion as Version
-
         wrapped_array = ArrayAccessCounter(np.arange(3.))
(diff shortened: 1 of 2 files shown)


lib/iris/tests/unit/lazy_data/test_co_realise_cubes.py
```diff
@@ -84,17 +81,7 @@ def test_combined_access(self):
         cube_b = Cube(derived_b)
         cube_c = Cube(derived_c)
         co_realise_cubes(cube_a, cube_b, cube_c)
-        # Though used more than once, the source data should only get fetched
```
pp-mo: Having removed all the comments, I think it is no longer clear why "2" is the correct and appropriate answer !
(i.e. why didn't it just fetch once, isn't that exactly the point ??)
Could you add a comment about that ?
Also, to make it really clear, you could add in more simultaneous statistics, to show e.g. "I calculated 5 derived stats but it only fetched twice".

pp-mo:
> no longer clear why "2" is the correct

But otherwise all fine.
Thanks @stephenworsley  :+1:

pp-mo:
Great stuff @stephenworsley !

PR #3461, opened 2019-10-11
Removed deprecated lbproc flag attributes

stephenworsley:
As per #3443, adresses the deprecation of the iris.fileformats.pp._LBProc.flag{} attributes.


pp-mo:
All reasonable, but I don't think it goes far enough.

We've have now removed the deprecated flag attributes, but with those gone the `class _LBProc` is now just a `SplittableInt`, extended to support some ordinary integer arithmetic operations.

But when [`field.lbproc` is created](https://github.com/SciTools/iris/blob/563426c1f74522015ec7985cfae5557c070a901c/lib/iris/fileformats/pp.py#L1042) , it is as a `SplittableInt`  with ***no named fields*** :  That makes the SplittableInt nature of it entirely pointless, so it might just as well be a plain integer !!

So... I ***think*** we should be able to discard "class _LBProc" altogether, and make field.lbproc a simple integer.

stephenworsley:
The _LBProc class has now been removed

pp-mo:
HI @stephenworsley, thanks for trying that.  I'm happy that if the tests pass, it is ok.

Given that, I would now merge, but ..
  1. we seem to have a merge problem.
  1. there should be something in the whats-new about PPField.lbproc now being a plain int

Can you fix those ?

stephenworsley:
@pp-mo 2 has been fixed, I believe I had already rebased in order to fix 1, GitHub is no longer showing any conflict on my end.

pp-mo:
Lovely!
Thanks for patience @stephenworsley

Issue #3447, opened 2019-10-04
Resolve iris-grib test failures


bjlittle:
See #3446, in particular this [comment](https://github.com/SciTools/iris/pull/3446/files#r331299450).

Results in the following errors=10, failures=5 for `iris-grib` tests, see this associated [travis-ci log](https://travis-ci.org/bjlittle/iris/builds/593293384).

pp-mo:
A couple of related issues that this ought to close :
  * #3375 
  * #3070

bjlittle:
This is also related:
- https://github.com/SciTools/iris/pull/2608

lbdreyer:
FYI:
When getting iris_grib (and python-eccodes) to correctly install in the travis job, you can use the [installation in the iris-grib repo](https://github.com/SciTools/iris-grib/blob/master/.travis.yml#L39-L48) as a reference. But note that [python-eccodes is now available from conda-forge](https://github.com/regro-cf-autotick-bot/python-eccodes-feedstock/blob/0b723587c91a83dba217c9e3b17b44f2783657e5/recipe/meta.yaml), so we should install it from there rather than a pip install as is done in iris-grib

bjlittle:
When installing from conda-forge note the approach used in [travis-ci](https://github.com/SciTools/iris/blob/master/.travis.yml#L103-L108), which ensures that installing `iris-grib` doesn't install `iris` as it's dependency - which could replace the development version of `iris` being tested.

lbdreyer:
#3469 has addressed most of the test failures, skipping the three that cannot be fixed in iris. 

I have raised an new PR to target the work of moving these tests to iris-grib where the necessary fixes can be applied

PR #3469, opened 2019-10-16
Have Travis test with iris-grib, remove problem tests

stephenworsley:
As per #3447, this allows Travis to test all grib tests except the ones currently failing/erroring.

### DIFF
--- a/lib/iris/tests/integration/test_grib2.py
+++ b/lib/iris/tests/integration/test_grib2.py
@@ -1,4 +1,4 @@
-# (C) British Crown Copyright 2014 - 2017, Met Office
+# (C) British Crown Copyright 2014 - 2019, Met Office
 #
 # This file is part of Iris.
 #
--- a/lib/iris/tests/integration/test_grib2.py
+++ b/lib/iris/tests/integration/test_grib2.py
@@ -22,7 +22,6 @@ from six.moves import (filter, input, map, range, zip)  # noqa
 # Import iris.tests first so that some things can be initialised before
 # importing anything else.
 import iris.tests as tests
-from unittest import skip
 
 from cf_units import Unit
 import numpy.ma as ma
@@ -50,7 +49,7 @@ class TestImport(tests.IrisTest):
         cube = load_cube(path)
         self.assertCMLApproxData(cube)
 
-    @skip('iris-grib is currently causing errors')
+    @tests.skip_grib_fail
     def test_gdt90_with_bitmap(self):
         path = tests.get_data_path(('GRIB', 'umukv', 'ukv_chan9.grib2'))
         cube = load_cube(path)
@@ -282,7 +281,7 @@ class TestGDT40(tests.IrisTest):
         cube = load_cube(path)
         self.assertCMLApproxData(cube)
 
-    @skip('iris-grib is currently causing failures')
+    @tests.skip_grib_fail
     def test_reduced(self):
         path = tests.get_data_path(('GRIB', 'reduced', 'reduced_gg.grib2'))
         cube = load_cube(path)
--- a/lib/iris/tests/integration/test_grib2.py
+++ b/lib/iris/tests/integration/test_grib2.py
@@ -22,6 +22,7 @@ from six.moves import (filter, input, map, range, zip)  # noqa
 # Import iris.tests first so that some things can be initialised before
 # importing anything else.
 import iris.tests as tests
+from unittest import skip
 
 from cf_units import Unit
 import numpy.ma as ma
@@ -49,6 +50,7 @@ class TestImport(tests.IrisTest):
         cube = load_cube(path)
         self.assertCMLApproxData(cube)
 
+    @skip('iris-grib is currently causing errors')
     def test_gdt90_with_bitmap(self):
         path = tests.get_data_path(('GRIB', 'umukv', 'ukv_chan9.grib2'))
         cube = load_cube(path)
@@ -280,6 +282,7 @@ class TestGDT40(tests.IrisTest):
         cube = load_cube(path)
         self.assertCMLApproxData(cube)
 
+    @skip('iris-grib is currently causing failures')
     def test_reduced(self):
         path = tests.get_data_path(('GRIB', 'reduced', 'reduced_gg.grib2'))
         cube = load_cube(path)
--- a/lib/iris/tests/integration/test_grib_load.py
+++ b/lib/iris/tests/integration/test_grib_load.py
@@ -1,4 +1,4 @@
-# (C) British Crown Copyright 2010 - 2017, Met Office
+# (C) British Crown Copyright 2010 - 2019, Met Office
 #
 # This file is part of Iris.
 #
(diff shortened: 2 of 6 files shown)


lib/iris/tests/integration/test_grib2.py
```diff
@@ -49,6 +50,7 @@ def test_gdt1(self):
         cube = load_cube(path)
         self.assertCMLApproxData(cube)
 
+    @skip('iris-grib is currently causing errors')
```
stephenworsley: I believe this is related (and possibly solved) by [this pull request](https://github.com/SciTools/iris-grib/pull/146).

lib/iris/tests/integration/test_grib_load.py
```diff
@@ -64,6 +65,7 @@ def test_load_time_processed(self):
                                                "time_bound.grib2")))
         self.assertCML(cubes, ("grib_load", "time_bound_grib2.cml"))
 
+    @skip('iris-grib is currently causing errors')
     def test_load_3_layer(self):
```
stephenworsley: I've done a bit of experimenting with this test. The error seems to be thrown when trying to realise one of the cubes in `cubes`. We get the error:
```ValueError: cannot reshape array of size 103680 into shape (360,600)```
Its worth noting that while `cubes[1]` has shape `(360, 600)` (where 360\*600 = 12960), `cubes[2]` has shape `(360,288)` (where 360\*288 = 103680).

It suggests that perhaps the cubes are at some point interfering with each other.
I tried changing the order of the cubelist to
```cubes = iris.cube.CubeList([cubes[2], cubes[0], cubes[1]])```
and I got the error:
```RuntimeError: Invalid GRIB message: /home/travis/build/stephenworsley/iris/iris-test-data/test_data/GRIB/3_layer_viz/3_layer.grib2 @ 671744```
Traceback can be found [here](https://travis-ci.com/stephenworsley/iris/jobs/246337089).
I'm still not sure if this is an Iris or Iris-grib issue.

File: lib/iris/tests/test_grib_load_translations.py [hunk 3]
```diff
@@ -60,7 +60,7 @@ def _mock_gribapi_fetch(message, key):
     if key in message:
         return message[key]
     else:
-        raise _mock_gribapi.GribInternalError
+        raise _mock_gribapi.errors.GribInternalError
```
stephenworsley: gribapi has changed where GribInternalError is stored. Fixing this solves 8 of the 10 errors in the grib tests mentioned in #3447. I've decided not to handle the possibility of both cases since the updated version of iris-grib is the only one which is compatible with python3.

lib/iris/tests/integration/test_grib_load.py
```diff
@@ -131,6 +133,7 @@ def test_reduced_ll(self):
             ("GRIB", "reduced", "reduced_ll.grib1")))
         self.assertCML(cube, ("grib_load", "reduced_ll_grib1.cml"))
 
+    @skip('iris-grib is currently causing failures')
     def test_reduced_gg(self):
```
stephenworsley: This failure seems to have been introduced by the latest verion of iris-grib. The points of one of the dimensions is reading [0.] rather than [~2.] .

lib/iris/tests/integration/test_grib2.py
```diff
@@ -280,6 +282,7 @@ def test_regular(self):
         cube = load_cube(path)
         self.assertCMLApproxData(cube)
 
+    @skip('iris-grib is currently causing failures')
     def test_reduced(self):
```
stephenworsley: This failure seems to have been introduced by the latest verion of iris-grib. The points of one of the dimensions is reading [0.] rather than [~2.] .

lib/iris/tests/test_grib_save.py
```diff
@@ -44,6 +45,7 @@ class TestLoadSave(tests.TestGribMessage):
     def setUp(self):
         self.skip_keys = []
 
+    @skip('iris-grib is currently causing failures')
     def test_latlon_forecast_plev(self):
```
stephenworsley: I suspect the following three tests are failing due to a change introduced by [this commit](https://github.com/SciTools/iris-grib/commit/c4de733f331b12207d493ace0946202ff3dc050c). It may be appropriate to change the `expected_diffs`.

lib/iris/tests/integration/test_grib_load.py
stephenworsley: Interestingly, if the `CubeList` isn't reordered then the test simply fails without errors.

stephenworsley: When I reordered the Cubelist as:
```cubes = iris.cube.CubeList([cubes[0], cubes[2], cubes[1]])```
I got the error:
```gribapi.errors.PrematureEndOfFileError: End of resource reached when reading message```

The traceback can be found [here](https://travis-ci.com/stephenworsley/iris/jobs/246714091).

Comment (stephenworsley) [reply lib/iris/tests/integration/test_grib_load.py]: It seems that if all the data is realised first, the test succeeds. The data must, however, be realised in the correct order:
```
cubes[0].data
cubes[1].data
cubes[2].data
```
where the data in `cubes` is realised the step before it is reordered.
If `cubes[1].data` is realised first, it will cause an error (similarly for `cubes[2]`).

Comment (stephenworsley) [reply lib/iris/tests/integration/test_grib_load.py]: So for some reason, the most recent version of python-ecodes is 0.9.3 rather than 2.12.3 . When the travis file specifies `python-eccodes=0.9.3` then this test no longer errors.

File: lib/iris/tests/integration/test_grib2.py (hunk 5 above)
Comment (stephenworsley) [reply lib/iris/tests/integration/test_grib2.py]: When loading this file I got the warning:
```
UserWarning: Unable to create instance of HybridPressureFactory.
The source data contains no field(s) for 'ref_surface_pressure'.
```
Possibly related to this commit: https://github.com/SciTools/iris-grib/commit/9f26e7aa23d15e93356598b4c231db887abdcd46

lib/iris/tests/integration/test_grib_load.py
stephenworsley: For reference, this issue is discussed here: https://github.com/conda-forge/python-eccodes-feedstock/issues/49
It looks like in future, a python interface will be added to eccodes so that it will replace python-eccodes and solve this version problem: https://github.com/conda-forge/staged-recipes/pull/9731

lib/iris/tests/integration/test_grib2.py
stephenworsley: This is actually related to this issue: https://github.com/SciTools/iris-grib/issues/150
When the suggested changes are applied this test passes again.

lib/iris/tests/integration/test_grib_load.py
stephenworsley: This is fixed by the changes suggested in https://github.com/SciTools/iris-grib/issues/150

lbdreyer:
@stephenworsley in the past, when we have skipped tests in Iris we have created a skip decorator in iris.tests.__init__ that we then use on each test.

The benefit of this approach is that we only need to write the message once (in the decorator), and it easier to find other skipped tests that are of the same "type"

lbdreyer:
Thanks @stephenworsley !

I have made a new issue to cover the work of moving the tests and ensuring that they are passing (See #3476)

Issue #3467, opened 2019-10-15
Include CellMeasures and AncillaryVariables in repr_html


lbdreyer:
Raised as part of this [review comment](https://github.com/SciTools/iris/pull/3422#issuecomment-541785295) 

It appears the CellMeasure and, once #3422 is in, AncillaryVariables are not handled in the `iris.experimental.representation` content handling

lbdreyer:
CellMeasures and AncillaryVariables should have their own headings, see: https://github.com/SciTools/iris/blob/master/lib/iris/experimental/representation.py#L101-L108

stephenworsley:
Just to add to this, it seems that `'Invalid coordinates'` and `'Scalar cell measures'` are also not currently hadled.

stephenworsley:
It looks like there are actually a few things that repr_html doesn't handle. Another edge case is when two coordinates share a name and this line is invoked: https://github.com/SciTools/iris/blob/0e362814bb87d4bda89f253b2342b41a1578b9bd/lib/iris/cube.py#L1793
This may or may not belong in the scope of this issue, but it is probably worth being aware of since it seems the addition of these headers may not be entirely trivial.

bjlittle:
Closed by #3475

PR #3475, opened 2019-10-22
Improve repr_html coverage

stephenworsley:
Addresses some of the comments made in #3467 . repr_html should work better in a few more cases now (recognising more headers, handling headers which aren't formatted with a colon etc.).


lib/iris/experimental/representation.py
```diff
@@ -102,13 +102,22 @@ def __init__(self, cube):
             'Dimension coordinates:': None,
             'Auxiliary coordinates:': None,
             'Derived coordinates:': None,
+            'Cell Measures:': None,
+            'Ancillary Datasets:': None,
```
lbdreyer: We have been throwing around a few different names for this (Ancillary Data, Ancillary Dataset, Ancillary Variable), so I think that's where some of the confusion has come from, but have agreed to call these `Ancillary Variables` so this will need updating

stephenworsley: Ancillary Datasets is what is currently being written in the summary here https://github.com/SciTools/iris/blob/77e4b3f7df83eb20a56957e236b7dc9073ce5458/lib/iris/cube.py#L2206 . Should I take it that this ought to be changed?

lbdreyer: Good spot! That does need changing!
I had originally called in "Ancillary Datasets" but that was pushed back so I needed to rename everything and clearly missed that name change!

stephenworsley:
Due to #3496 it may well be impossible to test `Invalid coordinates`. I think I'll probably remove `Invalid coordinates` from `str_headings` until #3496 is resolved.

bjlittle:
LGTM Thanks @stephenworsley :+1:

PR #3479, opened 2019-10-22
Fix test timings for Python 3.

pp-mo:
The [facility to time tests that are run](https://github.com/SciTools/iris/blob/v2.3.0rc0/lib/iris/tests/__init__.py#L944) was broken by changes to introspection in Python 3.


bjlittle:
@pp-mo I'd vote for totally removing this functionality in preparation for ASV.

Just purge it. Is there a good reason to keep it?

PR #3485, opened 2019-10-23
Coord as cube

pp-mo:
**WIP: incomplete, for discussion**

Some ideas (incomplete) for extracting a coordinate as a cube.
And the reverse operation : putting a cube inside another cube (with matching dims) as an aux-coord.

### DIFF
--- a/lib/iris/util.py
+++ b/lib/iris/util.py
@@ -1718,3 +1718,183 @@ def mask_cube(cube, points_to_mask):
     cube.data = ma.masked_array(cube.data)
     cube.data[points_to_mask] = ma.masked
     return cube
+
+
+def cubelike_array_as_cube(
+        item, cube, cube_dimensions,
+        name=None, standard_name=None, long_name=None, var_name=None,
+        units=None):
+    """
+    Make a new cube from data mapped over an existing cube's dimensions.
+
+    Args:
+
+    * item (array-like or cube- or coord-like):
+        An object which is either array-like object, or cube-like or
+        coordinate-like.
+        It provides array-like data, and must have a shape.
+        It may also optionally define a phenomemon identity (names), units and
+        attributes.
+        Cube data is drawn from the item 'points' or 'data' properties if it
+        has them, or the item itself.  Any of these must then be array-like.
+        Masked data is also supported.
+        Units, name properties and attributes are taken from the item if it has
+        them, or set with additional keywords, or may remain unspecified.
+
+    * cube (:class:`~iris.cube.Cube`):
+        A reference cube providing reference coordinate information for the
+        result.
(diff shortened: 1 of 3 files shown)


lib/iris/util.py
```diff
@@ -1718,3 +1718,254 @@ def mask_cube(cube, points_to_mask):
...
+                contains_dimension=i_dim_next)
+        if not daughter_coords:
+            msg = ("Could not find any coords in 'daughter_cube' for "
+                   "dimension {}.")
+            raise ValueError(msg.format(i_dim_next))
+        # If several coords, just take the first.
+        daughter_coord = daughter_coords[0]
+
+        # Find a unique matching coord in the parent.
+        parent_coords = parent_cube.coord(daughter_coord)
```
pp-mo: this should be ".coords()" (plural) !

bjlittle:
@pp-mo I'm not quite sure what the status of this PR is?

Are you just sharing the concept? Or do you want it reviewed to be merged?

Would you be able to clarify?

pp-mo:
> Are you just sharing the concept? Or do you want it reviewed to be merged?

Just sharing.  It is really intended as more of a an "issue" : A proposal that we should provide a mechanism to convert such cube component to-and-from independent cubes.

I wrote this stuff ages ago, but was reminded of it because it relates to the recent proliferation of "dimension metadata" components, in the effort to support Ancillary variables.
I.E. the changes in #3422

So it would make good sense, if doing this now, to extend the idea to work with any of coords, ancillary variables and cell measures.

I will have a go at re-casting this as a new-fangled

pp-mo:
Tried to change this to a newstyle non-mergeable PR, but it seems you don't get that choice.
I will try to re-form it as an issue somehow ....

bjlittle:
@pp-mo Would it make sense to close this PR and capture it as an issue? 

I'm all for having WIP PRs but it's confusing as a reviewer to determine what's a WIP PR that's actually aiming to deliver an agreed deliverable, what's a POC PR, and what's a "just sharing" for discussion PR.

To me it makes sense to discuss as much as possible via an issue, then commit to doing the work in confidence that there is agreement and buy-in that the PR will be banked (hopefully) on review.

What do you think?

pp-mo:
> What do you think?

I think the problem is that I'm using a PR format to present proof-of-concept code.
I will close this, convert to an issue, and reference a PR in my *own* repo for the PoC code...

pp-mo:
> I will close this, convert to an issue, and reference a PR in my own repo for the PoC code...

so
  - [ ] close this
  - [ ] [convert to an issue]()
  - ref a [PR in my own repo](https://github.com/pp-mo/iris/pull/61)

PR #3486, opened 2019-10-23
Remove comment about deprecated and deleted iris.Future

lbdreyer:
Since #3459, iris.FUTURE.cell_datetime_objects is not only deprecated but also removed, and so we should remove references to it.


lbdreyer:
@pp-mo could you look at this?

PR #3491, opened 2019-10-28
A couple of edits to the 2.3 whats new

lbdreyer:
Just a couple of changes to the Iris 2.3 what's new before we do the release

### DIFF
--- a/docs/iris/src/whatsnew/2.3.rst
+++ b/docs/iris/src/whatsnew/2.3.rst
@@ -11,7 +11,7 @@ Iris 2.3.0 Features
 ===================
 .. _showcase:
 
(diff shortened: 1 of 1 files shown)


docs/iris/src/whatsnew/2.3.rst
```diff
@@ -24,7 +24,7 @@ Iris 2.3.0 Features
   You can read more about each of these below.
 
   Additionally, the conventions attribute, added by Iris when saving to
-  NetCDF, has been updated to "CF-1.7", accordingly.
+  NetCDF, has been updated to ``CF-1.7``, accordingly.
```
lbdreyer: So it renders more nicely

PR #3492, opened 2019-10-29
Fix AncillaryVariable doc-string

bjlittle:
This simple PR aligns the `iris.coords.AncillaryVariable` doc-string with it's `__init__` formal arguments i.e., use `data` not `values`.


bjlittle:
@trexfeathers Awesome! Thanks :+1:

Issue #3457, opened 2019-10-10
Strip out Python 2 from Iris src code


lbdreyer:
Iris 3 will be a Python 3 only release so we need to remove all Python 2 support.

This will include removing all the `six` code and any other remaining Python 2 handling.

Note, we have already dropped testing with Python 2 (see #3446)

Note, this could be split into multiple smaller issues.

---
### Acceptance Criteria:
- [x] Remove all `six` code - @trexfeathers #3497 
- [x] Remove unnecessary `__future__` imports & address `iris.tests.test_coding_standards.TestFutureImports` - @stephenworsley #3505
- [x] Remove unnecessary `super` arguments - @stephenworsley #3506
- [x] Remove/replace all Python2 specific code  @stephenworsley #3507 
- [x] Address all code that deals with Python3 as optionally supported @stephenworsley #3507 
- [x] Remove Python2 class references to `object`, see [here](https://realpython.com/python3-object-oriented-programming/#how-to-define-a-class-in-python) for an example @stephenworsley #3511 
- [ ] Merge feature branch `py2_strip` to `master`
- [ ] Delete feature branch `py2_strip`

bjlittle:
Remove the `iris.tests.test_coding_standards.TestFutureImports` class, see [here](https://github.com/SciTools/iris/blob/master/lib/iris/tests/test_coding_standards.py#L346-L403).

bjlittle:
Also, grep through the code base and inspect hits for:

- `grep -ri "python3"` and `grep -ri "python 3"`
- `grep -ri "python2"` and `grep -ri "python 2"`

stephenworsley:
For Python 3 relevent `__future__` imports, see https://docs.python.org/3/library/__future__.html

stephenworsley:
Since it is conceivable that we might introduce `__future__` imports at a later date, it might be worthwhile skipping this test, removing Python 2/3 specific references, and leaving its skeleton for future reference.

trexfeathers:
An example of some of the 'Python 2 specific code' we might be looking for:
```python
        with self.assertRaisesRegex(TypeError,
                                    # exactly == Py2, positional == Py3
                                    'takes (exactly )?1 (positional )?'
                                    'argument '):
```

stephenworsley:
[...]

PR #3503, opened 2019-11-01
Remove unnecessary super arguments

stephenworsley:
As requested in #3457, `super` use has been simplified.


stephenworsley:
This is now being addressed by #3506, which is pulling into a branch of master for the sake of safely merging with other similar pull requests.

PR #3504, opened 2019-11-01
Removed __future__ imports

stephenworsley:
Following #3457, all imports from `__future__` have been removed, as have all tests for these imports and six imports.


stephenworsley:
This is now being addressed by #3505, which is pulling into a branch of master for the sake of safely merging with other similar pull requests.

PR #3505, opened 2019-11-01
Removed __future__ imports

stephenworsley:
Following #3457, all imports from __future__ have been removed, as have all tests for these imports and six imports.


stephenworsley:
@trexfeathers Good catch.

trexfeathers:
All good, thanks @stephenworsley 😎

PR #3506, opened 2019-11-01
Remove unnecessary super arguments

stephenworsley:
As requested in #3457, `super` use has been simplified.


bjlittle:
@stephenworsley Lookin' good!

The only `super` statement that might still need looking at is:
```script
1003:       return super(Cell, self).__hash__()
./coords.py

```

bjlittle:
@stephenworsley Awesome, thanks! :+1:

PR #3507, opened 2019-11-04
Remove python 2 specific code

stephenworsley:
As requested in #3457, Python 2 specific code has been removed.

### DIFF
--- a/lib/iris/tests/unit/analysis/maths/test_divide.py
+++ b/lib/iris/tests/unit/analysis/maths/test_divide.py
@@ -28,10 +28,7 @@ class TestBroadcasting(tests.IrisTest_nometa,
                        CubeArithmeticBroadcastingTestMixin):
     @property
     def data_op(self):
-        try:
-            return operator.div
-        except AttributeError:
-            return operator.truediv
+        return operator.truediv
 
     @property
     def cube_func(self):
@@ -42,10 +39,7 @@ class TestBroadcasting(tests.IrisTest_nometa,
 class TestMasking(tests.IrisTest_nometa, CubeArithmeticMaskingTestMixin):
     @property
     def data_op(self):
-        try:
-            return operator.div
-        except AttributeError:
-            return operator.truediv
+        return operator.truediv
 
     @property
     def cube_func(self):
--- a/lib/iris/_constraints.py
+++ b/lib/iris/_constraints.py
@@ -12,10 +12,7 @@ from __future__ import (absolute_import, division, print_function)
 from six.moves import (filter, input, map, range, zip)  # noqa
 import six
 
-try:  # Python 3
-    from collections.abc import Iterable, Mapping
-except ImportError:  # Python 2.7
-    from collections import Iterable, Mapping
+from collections.abc import Iterable, Mapping
 import operator
 
 import numpy as np
--- a/lib/iris/analysis/__init__.py
+++ b/lib/iris/analysis/__init__.py
@@ -40,10 +40,7 @@ from six.moves import (filter, input, map, range, zip)  # noqa
 import six
 
 from collections import OrderedDict
-try:  # Python 3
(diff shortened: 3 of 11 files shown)


lib/iris/tests/unit/analysis/maths/test_divide.py
```diff
@@ -28,10 +28,7 @@ class TestBroadcasting(tests.IrisTest_nometa,
                        CubeArithmeticBroadcastingTestMixin):
     @property
     def data_op(self):
-        try:
-            return operator.div
-        except AttributeError:
-            return operator.truediv
+        return operator.truediv
```
bjlittle: @stephenworsley It seems that the whole point of the `data_op` property was convenience i.e., to do the right thing across Python2 and Python wrt `div` and `truediv`.

Now that we're Python3, I'm just questioning the need for the existence of the `data_op` property. 

Can't we just use `operator.truediv` or `/` directly inline for the tests? i.e.,

```python
    def test_unmasked_div_zero(self):
        # Ensure cube behaviour matches numpy operator behaviour for the
        # handling of arrays containing 0.
        dat_a = np.array([0., 0., 0., 0.])
        dat_b = np.array([2., 2., 2., 2.])

        cube_a = Cube(dat_a)
        cube_b = Cube(dat_b)

        com = dat_b / dat_a
        res = self.cube_func(cube_b, cube_a).data

        self.assertArrayEqual(com, res)
```
Also drop the `import operator` as well :wink:

lib/iris/tests/unit/analysis/maths/test_divide.py
```diff
@@ -42,10 +39,7 @@ def cube_func(self):
 class TestMasking(tests.IrisTest_nometa, CubeArithmeticMaskingTestMixin):
     @property
     def data_op(self):
-        try:
-            return operator.div
-        except AttributeError:
-            return operator.truediv
+        return operator.truediv
```
bjlittle: As the [above comment](https://github.com/SciTools/iris/pull/3507/files#r342056420).

lib/iris/tests/unit/analysis/maths/test_divide.py
stephenworsley: So it looks like what's going on here is that these classes are actually inheriting a bunch of tests from https://github.com/SciTools/iris/blob/master/lib/iris/tests/unit/analysis/maths/__init__.py

`data_op` is an abstract property in the parent class where it is used in other tests https://github.com/SciTools/iris/blob/66423f8104f25cbea71bc470b5769eb6da876265/lib/iris/tests/unit/analysis/maths/__init__.py#L28-L31 It seems like the current usage makes sense in this context.

bjlittle: @stephenworsley Cool. 

I guess what I'm pushing is that the whole `data_op` property as a concept is now defunct and should be removed. For me it's simply technical debt that should be purged.

stephenworsley: I don't think it is defunct, it allows code reuse for the tests in [iris.tests.unit.maths.\_\_init__](https://github.com/SciTools/iris/blob/master/lib/iris/tests/unit/analysis/maths/__init__.py) over multiple different operations, see also addition for example https://github.com/SciTools/iris/blob/master/lib/iris/tests/unit/analysis/maths/test_add.py

bjlittle: Okay, so it's generic and not specific, thanks for clarifying :+1:

bjlittle:
@stephenworsley So is this good to go? Do you have more changes in-bound?

stephenworsley:
@bjlittle Thats all I can find for the moment, If we find any others they can go in another pull request.

bjlittle:
@stephenworsley Great detective work, thanks :smile:

Issue #3410, opened 2019-09-23
The iris.coords.Coord class should be abstract


bjlittle:
The `iris.coords.Coord` class should be private i.e., `iris.coords._Coord`, as it is not intended to be used to create instances, but rather it's purpose is to act as the base class in the class hierarchy of `DimCoord` and `AuxCoord`.

---
### Acceptance Criteria
- [x] A user is no longer able to instantiate a Coord

pp-mo:
> as it is not intended to be used to create instances

IMHO that means it should be abstract, not private.
It doesn't sound wise to me to make a parent of a public class *private* :  it makes it hard to work out what public API you can rely on, if some of it comes from a private.

For instance, it should remain possible to say `isinstance(thing, Coord)`, for which Coord needs to remain a public thing.  This is at least one thing that it is good for.

bjlittle:
ABC works also for me :+1:

bjlittle:
Closed by #3514

PR #3514, opened 2019-11-08
Make Coord abstract

stephenworsley:
This addresses #3410 by making `__init__` in `Coord` an `abstractmethod`.

### DIFF
--- a/lib/iris/coords.py
+++ b/lib/iris/coords.py
@@ -8,7 +8,7 @@ Definitions of coordinates and other dimensional metadata.
 
 """
 
-from abc import ABCMeta
+from abc import ABCMeta, abstractmethod
 from collections import namedtuple
 from collections.abc import Iterator
 import copy
@@ -1282,6 +1282,7 @@ class Coord(_DimensionalMetadata):
 
     """
 
+    @abstractmethod
     def __init__(
         self,
         points,
@@ -2565,6 +2566,8 @@ class AuxCoord(Coord):
 
     """
 
+    def __init__(self, *args, **kwargs):
+        super().__init__(*args, **kwargs)
     # Logically, :class:`Coord` is an abstract class and all actual coords must
     # be members of some concrete subclass, i.e. an :class:`AuxCoord` or
     # a :class:`DimCoord`.
--- a/lib/iris/coords.py
+++ b/lib/iris/coords.py
@@ -2567,7 +2567,9 @@ class AuxCoord(Coord):
     """
 
     def __init__(self, *args, **kwargs):
+        self.__doc__ = super().__init__.__doc__
         super().__init__(*args, **kwargs)
+
     # Logically, :class:`Coord` is an abstract class and all actual coords must
     # be members of some concrete subclass, i.e. an :class:`AuxCoord` or
     # a :class:`DimCoord`.
--- a/lib/iris/coords.py
+++ b/lib/iris/coords.py
@@ -12,6 +12,7 @@ from abc import ABCMeta, abstractmethod
 from collections import namedtuple
 from collections.abc import Iterator
 import copy
+from functools import wraps
 from itertools import chain, zip_longest
 import operator
 import warnings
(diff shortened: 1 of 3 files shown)


stephenworsley:
One thing to bear in mind about this approach is that the  `__doc__` of `AuxCoord.__init__` will be overwritten. Though `inspect.getdoc(AuxCoord.__init__)` still inherits the docstring properly, so this may be enough for things like Sphinx.

stephenworsley:
It might also make sense to make `_DimensionalMetadata` properly abstract. It currently derives from `ABCMeta` but has no abstract methods, so it can still be instantiated.

bjlittle:
@stephenworsley I played with a couple of options for the `AuxCoord` doc-string.

```python
from functools import wraps
...
class AuxCoord(Coord):
    @wraps(Coord.__init__)
    def __init__(self, *args, **kwargs):
        super().__init__(*args, **kwargs)
```
Works fine, but has a minor performance overhead:
```python
timeit('AuxCoord(1)', setup='from iris.coords import AuxCoord')
Out[10]: 11.958185139999841
timeit('AuxCoord(1)', setup='from iris.coords import AuxCoord')
Out[11]: 11.816572594994796
timeit('AuxCoord(1)', setup='from iris.coords import AuxCoord')
Out[12]: 11.72259231901262
timeit('AuxCoord(1)', setup='from iris.coords import AuxCoord')
Out[13]: 11.927083795992075
```
This is compared with `master`:
```python
timeit('AuxCoord(1)', setup='from iris.coords import AuxCoord')
Out[5]: 11.874402218993055
timeit('AuxCoord(1)', setup='from iris.coords import AuxCoord')
Out[6]: 11.801809267999488
timeit('AuxCoord(1)', setup='from iris.coords import AuxCoord')
Out[7]: 11.369123923010193
timeit('AuxCoord(1)', setup='from iris.coords import AuxCoord')
Out[8]: 11.455261625000276
```
An alternative is:
```python
class AuxCoord(Coord):
    def __init__(self, *args, **kwargs):
        self.__doc__ = super().__init__.__doc__
        super().__init__(*args, **kwargs)
```
which has the following behaviour:
```python
timeit('AuxCoord(1)', setup='from iris.coords import AuxCoord')
Out[7]: 12.073399219996645
timeit('AuxCoord(1)', setup='from iris.coords import AuxCoord')
Out[8]: 12.17960434599081
timeit('AuxCoord(1)', setup='from iris.coords import AuxCoord')
Out[9]: 12.184167061001062
timeit('AuxCoord(1)', setup='from iris.coords import AuxCoord')
Out[10]: 12.351267233985709
```
Or indeed, there is this:
```python
class AuxCoord(Coord):
    def __init__(self, *args, **kwargs):
        self.__doc__ = Coord.__init__.__doc__
        super().__init__(*args, **kwargs)
```
which has the following behaviour:
```python
timeit('AuxCoord(1)', setup='from iris.coords import AuxCoord')
Out[10]: 11.896369188994868
timeit('AuxCoord(1)', setup='from iris.coords import AuxCoord')
Out[11]: 12.091242638998665
timeit('AuxCoord(1)', setup='from iris.coords import AuxCoord')
Out[12]: 12.14877037498809
timeit('AuxCoord(1)', setup='from iris.coords import AuxCoord')
Out[13]: 12.003538096003467
```
Note that `timeit` is executing the `AuxCoord(1)` statement `1,000,000` times by default.

Given the above, I'd be happy to opt for the first option using `@wraps`

bjlittle:
@stephenworsley Could you please make an issue for `_DimensionalMetadata`, thanks :+1:

stephenworsley:
It looks like using `@wraps` causes `@abstractmethod` to carry over as well.

lib/iris/coords.py
```diff
@@ -2565,6 +2567,10 @@ class AuxCoord(Coord):
 
     """
 
+    @wraps(Coord.__init__, assigned=("__doc__",), updated=())
```
bjlittle: @stephenworsley Very nice indeed :wink:

PR #3519, opened 2019-11-11
_regrid_area_weighted_array: Axes creation before loop

abooton:
In _regrid_area_weighted_array:
Move `axes` creation before the main for loop.
`x_dim` and `y_dim` values do not change, thus the axis tuple should remain constant.
Therefore it can be calculated prior to the main computation loop.


bjlittle:
@abooton Thanks, nice spot :eyes:

Issue #3495, opened 2019-10-31
ASV - Airspeed Velocity Performance Tests - AVD-1325


alastair-gemmell:

trexfeathers:
All associated GitHub and internal pull requests have been reviewed and completed. Future work has been captured in further GitHub and internal issues

PR #3526, opened 2019-11-12
Airspeed Velocity Set Up

trexfeathers:
This pull request enables testing with [Airspeed Velocity](https://asv.readthedocs.io/en/stable/index.html) 🐦

Any user with an ASV install can run the suite of tests held within ``lib/iris/tests/benchmarking`` using the shell command ``asv run``, while in the root directory. See the ASV documentation linked above for more details on what the software can do.

Points for future discussion:
- What tests should be included in the suite?
- What will be the standard arguments for ``asv run``, and how should we store these?
  - Documentation?
  - A small text file that could be referenced in the building of scripts?


trexfeathers:
Covers some of #3495

asv.conf.json
```diff
@@ -0,0 +1,39 @@
...
+        "requests": [],
+    },
+
+    "benchmark_dir": "lib/iris/tests/benchmarking",
+    "env_dir": ".asv/env",
+    "results_dir": ".asv/results",
+    "html_dir": ".asv/html",
+
+    // See https://asv.readthedocs.io/en/stable/asv.conf.json.html for
+    // details on what can be included in this file.
```
bjlittle: @trexfeathers Perhaps move this comment to the top of the file? 🤔

lib/iris/tests/benchmarking/regridding.py
```diff
@@ -0,0 +1,34 @@
...
+performance testing. it is expected to be expanded/replaced in future.
+
+"""
+
+# import iris tests first so that some things can be initialised before
+# importing anything else.
+from iris import tests
+
+import iris
+
```
bjlittle: @trexfeathers No blank line necessary here.

lib/iris/tests/benchmarking/regridding.py
```diff
@@ -0,0 +1,34 @@
...
+from iris import analysis
```
bjlittle: Instead, perhaps use:
```python
from iris.analysis import AreaWeighted
```

lib/iris/tests/benchmarking/regridding.py
```diff
@@ -0,0 +1,34 @@
...
+
+
+@tests.skip_data
+class RegriddingTests:
+    def setup(self):
+        # Prepare a cube and a regridding scheme.
+        file_path = tests.get_data_path(
+            ["NetCDF", "global", "xyt", "SMALL_hires_wind_u_for_ipcc4.nc"]
+        )
+        self.cube = iris.load(file_path)[0]
```
bjlittle: @trexfeathers This test data only contains one `cube`, so instead use:
```python
self.cube = iris.load_cube(file_path)
```

lib/iris/tests/benchmarking/regridding.py
```diff
@@ -0,0 +1,34 @@
...
+        self.scheme_area_w = analysis.AreaWeighted()
```
bjlittle: @trexfeathers There's no real advantage to having this as part of the `setUp`. As `AreaWeighted` can simply be used directly within the test itself...

lib/iris/tests/benchmarking/regridding.py
```diff
@@ -0,0 +1,34 @@
...
+
+    def time_regrid_area_w(self):
+        # Regrid the cube onto itself.
+        self.cube.regrid(self.cube, self.scheme_area_w)
```
bjlittle: @trexfeathers ... i.e., make this:
```python
self.cube.regrid(self.cube, AreaWeighted())
```

bjlittle:
@trexfeathers Do we need to add the `asv` package to the `requirements/test.txt` to make it a dependency?

asv.conf.json
```diff
@@ -0,0 +1,39 @@
+{
+    "version": 1,
+    "project": "iris",
+    "project_url": "https://github.com/SciTools/iris",
+    "repo": ".",
+    "environment_type": "conda",
+    "show_commit_url": "http://github.com/scitools/iris/commit/",
+    //"pythons": ["2.7", "3.6", "3.7"],
```
bjlittle: @trexfeathers This is `virtualenv` specific, right? If that's the case, how do we control the python versions per `conda` environment? 

Apologies, I'm an `asv` noob 😄

bjlittle:
@trexfeathers Perhaps not in this PR, but we should also provide documentation for the `iris` developer on how to use `asv` etc

lib/iris/tests/benchmarking/regridding.py
trexfeathers: @bjlittle _any_ expense at all seems like something to avoid when it comes to benchmark testing?

lib/iris/tests/benchmarking/regridding.py
trexfeathers: https://github.com/SciTools/iris/pull/3526#discussion_r345729344

trexfeathers:
> @trexfeathers Do we need to add the asv package to the requirements/test.txt to make it a dependency?

@bjlittle nice catch. The existing automated tests would not break without this, but it's still clearer to have asv as a dependency, so I've added it 😊

asv.conf.json
trexfeathers: 👍

lib/iris/tests/benchmarking/regridding.py
trexfeathers: 👍 looks like I misinterpreted that part of pep8

lib/iris/tests/benchmarking/regridding.py
trexfeathers: 👍  I need to break the habit of unecessary future-proofing

lib/iris/tests/benchmarking/regridding.py
trexfeathers: 👍  I need to break the habit of unnecessary future-proofing

asv.conf.json
trexfeathers: @bjlittle ASV handles this with conda just fine - it just makes a different environment for each. I expect this to be something we might want to use but it seemed unnecessary for a minimal demonstration.

trexfeathers:
> @trexfeathers Perhaps not in this PR, but we should also provide documentation for the iris developer on how to use asv etc

@bjlittle 100% with you, once we have an agreed way of using ASV that we can document. I'm particularly interested in standardising the `asv run` arguments in some way.

asv.conf.json
bjlittle: @trexfeathers Okay, cool... but if we're not actively using the `pythons` key, then would you be adverse to removing it?

bjlittle:
> @bjlittle 100% with you, once we have an agreed way of using ASV that we can document. I'm particularly interested in standardising the `asv run` arguments in some way.

Nice. How do we capture this intent moving forwards? An issue? Elsewhere? Or have you got that covered? 😉

asv.conf.json
trexfeathers: Keeping commented code lines is another bad habit. I'll remove it.

bjlittle: Between your expertise and the `asv` docs, I think we'll have it covered. Nice one, thanks!

trexfeathers:
> @bjlittle 100% with you, once we have an agreed way of using ASV that we can document. I'm particularly interested in standardising the asv run arguments in some way.
> 
> Nice. How do we capture this intent moving forwards? An issue? Elsewhere? Or have you got that covered? 😉

I'm not intending to map out next steps until I have finished experimenting with local routine testing schedules as this will inform further discussion.

lib/iris/tests/benchmarking/regridding.py
bjlittle: @trexfeathers Seems like a healthy benchmarking philosophy to adopt - we should roll like that moving forwards for other `asv` benchmarks 👍

bjlittle:
@trexfeathers `asv` cleared for take-off runway `iris` 🎉

PR #3531, opened 2019-11-13
Make _DimensionalMetadata abstract

### LINKED ISSUE #3521: Decide abstract status of _DimensionalMetadata
Uncovered by work on #3514. While `_DimensionalMetadata` inherits from `ABCMeta`, it contains no abstract methods. Therefore, it can still be instatiated. It should be decided, (and then implemented) If this is how we want `_DimensionalMetadata` to behave.

bjlittle:
Closes #3521

### DIFF
--- a/lib/iris/tests/unit/coords/test__DimensionalMetadata.py
+++ b/lib/iris/tests/unit/coords/test__DimensionalMetadata.py
@@ -1,0 +1,26 @@
+# Copyright Iris contributors
+#
+# This file is part of Iris and is released under the LGPL license.
+# See COPYING and COPYING.LESSER in the root of the repository for full
+# licensing details.
+"""Unit tests for the :class:`iris.coords._DimensionalMetadata` class."""
+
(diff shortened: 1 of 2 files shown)


lib/iris/tests/unit/coords/test__DimensionalMetadata.py
```diff
@@ -0,0 +1,26 @@
...
+
+# Import iris.tests first so that some things can be initialised before
+# importing anything else.
+import iris.tests as tests
+
+from iris.coords import _DimensionalMetadata
+
+
+class Test___init____abstractmethod(tests.IrisTest):
+    def test(self):
```
stephenworsley: I feel like this method ought to have a more descriptive name.

stephenworsley: Actually, having had a look at other examples in the tests, I can see there is precedent for this.

bjlittle:
@stephenworsley Awesome, thanks :+1:

PR #3533, opened 2019-11-14
CellMeasure.measure with valid default

bjlittle:
This PR addresses the issue that `iris.coords.CellMeasure` currently has `measure` as a keyword argument with a default value of `None`, which is invalid:
```python
>>> from iris.coords import CellMeasure
>>> CellMeasure(1)
...
ValueError: measure must be 'area' or 'volume', not None
```
An `iris.coords.CellMeasure.measure` can only either be `area` or `volume`.

Having an invalid default value is confusing, and suggests that `measure` must be a positional argument to be provided by the user.

---
This has been revised to maintain `measure` as a keyword argument, but with a valid default value of `area` instead.


bjlittle:
This may be controversial as is... one could argue that it's a change to the public API.

Less controversial is to keep `measure` as a keyword argument, but with a valid default i.e., `area` instead of `None`

trexfeathers:
Checked that all Iris' existing uses of `CellMeasure` play nicely with this change - all fine.

@bjlittle definitely agree with your choice to keep the API consistent if possible. I'm assuming not having an actual default argument of 'area' is because of the philosophy espoused [here](https://graysonkoonce.com/always-use-none-for-default-args-in-python/)?

trexfeathers:
LGTM ✔

bjlittle:
@trexfeathers Nice one, thanks 🤓

Issue #3529, opened 2019-11-13
Fix cube equality for ancillary variables


stephenworsley:
A subtask for #3422. Fixing this first should make testing for the rest of #3422 easier.

Cube equality `(__eq__)` is ignoring AVs (and maybe CellMethods) but not AuxCoords.

---
### Acceptance Criteria
 - [x] Fix cube equality
 - [x] Write tests for equality
 - [x] Raise issues for any failing tests (tests may be skipped until a further pull request)

bjlittle:
@stephenworsley Just checking... is this also applicable to `iris.coords.CellMeasure`s?

pp-mo:
Just about the naming of this issue ...   It seems to be mixture of `auxiliary` and `ancillary` ?
It's confusing enough already !!

stephenworsley:
@bjlittle I believe so.

bjlittle:
@SciTools/iris-devs Are we proposing that `iris.cube.Cube` equality includes comparing the associated `AncillaryVariable` and `CellMeasure` metadata **and** data?

Just guessing, but that sounds **mighty expensive**, particularly for large cubes 💣 

Aren't we glad that #3526 landed...

pp-mo:
>  sounds mighty expensive

But we are doing it for coords, including possibly huge multidimensional AuxCoords,
so where's the difference ??

bjlittle:
> so where's the difference ??

@pp-mo The *difference* is that we could also potentially be comparing between 1..N `AncillaryVariable` data payloads of the same dimensionality as the `iris.cube.Cube.data`, in addition to the `iris.cube.Cube.data`, plus perhaps a `CellMeasure`, and then as you mentioned *possibly huge multidimensional AuxCoords* 🤒

I'm just raising the concern that this just might not scale well, and be not at all practical or performant for the user. Perhaps we should consider options for the user with large data **and** large metadata, rather than offering Hobson's choice `cube` equality; where absolutely everything is always compared.

Just sayin'...

bjlittle:
@stephenworsley BTW `Cube.__eq__` uses the `Cube.metadata` during equality, which includes `Cube.cell_methods`, see [here ](https://github.com/SciTools/iris/blob/v2.3.0rc0/lib/iris/cube.py#L3051)and [here](https://github.com/SciTools/iris/blob/v2.3.0rc0/lib/iris/cube.py#L832). HTH, so I guess you should change the description of this issue.

[...]

Issue #3530, opened 2019-11-13
Cube.is_compatible requires cell_methods to have the same order


stephenworsley:
On this line:  https://github.com/SciTools/iris/blob/b2ce2a34e2eef7e3d6203c77ada7ed4ce89e3145/lib/iris/cube.py#L918 cell methods is compared in a way which seems like it cares about the order of the cell methods. It may be more appropriate to check equality in a way which ignores order (perhaps by comparing sets).

bjlittle:
@stephenworsley The `iris.coords.CellMethod` is not very rich in functionality, even through it's been around for a while.

This certainly seems like a reasonable approach to take to make the equality more robust i.e., independent of order - which shouldn't make a difference.

rcomer:
While you’re in there, I have a question about this method at #2992.  Perhaps the docstring can be improved for clarification?

rhattersley:
In general, I believe cell methods are not commutable, e.g. taking the mean over time of a standard deviation over area vs. a standard deviation over area of a mean over time. Hence, the order of cell methods is important.

stephenworsley:
@rhattersley Thanks, that's good to know.

PR #3536, opened 2019-11-14
Fix equality

stephenworsley:
This pull request currently shows an approach to solving #3529 which, while failed, I think is still instructive.

I tried comparing `_ancillary_variables_and_dims` and `_cell_measures_and_dims` in a way which would ignore their order as lists. This first attempt involved simply turning these lists into sets. This failed because the objects inside the lists are not viable elements of a set.

One idea I am coming around to, if such a comparison is deemed necessary, is to use the code in this function: https://github.com/SciTools/iris/blob/a05424137f1f19c26b815c104808f4f563f88812/lib/iris/analysis/__init__.py#L189 perhaps creating a generic version which compares generic aspects of a cube, then having `coord_comparison` and something like `ancil_comparison` wrap this generic function. I'm leaning towards this idea since it seems as though `coord_comparison` is designed to be more efficient than a more naive, brute force approach. If we needed to compare many ancillary variables (which seems like a thing we would probably want to do) it seems likely we would end up writing something similar to what already exists in this code anyway.

### DIFF
--- a/lib/iris/cube.py
+++ b/lib/iris/cube.py
@@ -1044,7 +1044,7 @@ bound=(1994-12-01 00:00:00, 1998-12-01 00:00:00)
 
     def _add_unique_aux_coord(self, coord, data_dims):
         data_dims = self._check_multi_dim_metadata(coord, data_dims)
-        self._aux_coords_and_dims.append([coord, data_dims])
+        self._aux_coords_and_dims.append((coord, data_dims))
 
     def add_aux_factory(self, aux_factory):
         """
@@ -1095,7 +1095,7 @@ bound=(1994-12-01 00:00:00, 1998-12-01 00:00:00)
         if self.cell_measures(cell_measure):
             raise ValueError("Duplicate cell_measures are not permitted.")
         data_dims = self._check_multi_dim_metadata(cell_measure, data_dims)
-        self._cell_measures_and_dims.append([cell_measure, data_dims])
+        self._cell_measures_and_dims.append((cell_measure, data_dims))
         self._cell_measures_and_dims.sort(
             key=lambda cm_dims: (cm_dims[0]._as_defn(), cm_dims[1])
         )
@@ -1125,7 +1125,7 @@ bound=(1994-12-01 00:00:00, 1998-12-01 00:00:00)
             ancillary_variable, data_dims
         )
         self._ancillary_variables_and_dims.append(
-            [ancillary_variable, data_dims]
+            (ancillary_variable, data_dims)
         )
         self._ancillary_variables_and_dims.sort(
             key=lambda av_dims: (av_dims[0]._as_defn(), av_dims[1])
@@ -1195,7 +1195,7 @@ bound=(1994-12-01 00:00:00, 1998-12-01 00:00:00)
                 )
             )
 
-        self._dim_coords_and_dims.append([dim_coord, int(data_dim)])
+        self._dim_coords_and_dims.append((dim_coord, int(data_dim)))
 
     def remove_aux_factory(self, aux_factory):
         """Removes the given auxiliary coordinate factory from the cube."""
@@ -1263,7 +1263,7 @@ bound=(1994-12-01 00:00:00, 1998-12-01 00:00:00)
         cell_measure = self.cell_measure(cell_measure)
 
         self._cell_measures_and_dims = [
-            [cell_measure_, dim]
+            (cell_measure_, dim)
             for cell_measure_, dim in self._cell_measures_and_dims
             if cell_measure_ is not cell_measure
         ]
@@ -1280,7 +1280,7 @@ bound=(1994-12-01 00:00:00, 1998-12-01 00:00:00)
 
         """
         self._ancillary_variables_and_dims = [
-            [ancillary_variable_, dim]
+            (ancillary_variable_, dim)
             for ancillary_variable_, dim in self._ancillary_variables_and_dims
             if ancillary_variable_ is not ancillary_variable
         ]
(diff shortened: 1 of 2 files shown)


stephenworsley:
After having a bit more of a look at `coord_comparison`, this may not be the most efficient way to determine equality either, since dimensions within the cube are checked for equality after the data in the coords are checked for equality. There's still a case to make a generic function that checks equality of various `*_and dims` properties.

pp-mo:
> This first attempt involved simply turning these lists into sets. This failed because the objects inside the lists are not viable elements of a set.

I think you've diagnosed this wrong.
The errors with set creation occur because the elements of `cube._ancillary_variables_and_dims` are of the form [ancil-var-object, dims-tuple].
I think that is just a mistake.  The entries in that list should be 2-tuples (var, dims), not lists.
So the problem is just [this line](https://github.com/SciTools/iris/blob/dd6ae54657428b59ff614e73c8f23fa163050043/lib/iris/cube.py#L1128), which I reckon should be making a 2-tuples instead of a 2-element list.

If you fix that, I think the set approach will work.
I like this approach because it provides all the needed logic, based only on the equality testing.

pp-mo:
**WARNING**
@bjlittle / @rhattersley  pointed out that the ordering of cell-methods **is** significant.
So in that case we can have simple list equality.

stephenworsley:
@pp-mo For consistency, I have made the same change to `dim_coords_and_dims` and `aux_coords_and_dims`. Also, I have added a test to make sure reordering cell methods causes equality to fail.

lib/iris/cube.py
```diff
@@ -3583,6 +3583,15 @@ def __eq__(self, other):
                     or coord_comparison["non_equal_data_dimension"]
                 )
 
+            if result:
+                anc_diff = set(self._ancillary_variables_and_dims) == set(
+                    other._ancillary_variables_and_dims
+                )
+                cm_diff = set(self._cell_measures_and_dims) == set(
```
pp-mo: It might be worth "shortcutting" again, by putting both of these checks into an "if result:" block.
For choice, I think we should check cell-measures first, as being typically X/Y (geo-grid) based, they will commonly have fewer dimensions than ancillary variables -- which could often be as big as the whole main data array.
I think it is also slightly neater for all the checks to work the same way : = "test result and if not already False set it again", that is instead of grouping test results with 'and'.

lib/iris/tests/unit/cube/test_Cube.py
```diff
@@ -2206,5 +2212,68 @@ def test_data_bool_not_eq(self):
         self.assertFalse(cube1 == cube2)
 
 
+class Test__eq__meta(tests.IrisTest):
+    def test_ancillary_fail(self):
+        cube1 = Cube([0, 1])
+        cube2 = Cube([0, 1])
+        avr = AncillaryVariable([2, 3], long_name="foo")
+        cube2.add_ancillary_variable(avr, 0)
+        self.assertFalse(cube1 == cube2)
```
pp-mo: I think it might be worth checking that the values in ancillaries count : I.e. cubes with the same ancillaries, but some values different, will not match.
You could extend that to attributes and units, if you want to be really thorough.

You may find, as I did in #3550, that some of this could be simpler + clearer if we can get [cube.copy support for ancillaries](https://github.com/SciTools/iris/pull/3546) in first.

lib/iris/tests/unit/cube/test_Cube.py
```diff
@@ -2206,5 +2212,68 @@ def test_data_bool_not_eq(self):
...
+        cube2 = Cube([0, 1])
+        cmth1 = CellMethod("mean", "time", "6hr")
+        cmth2 = CellMethod("mean", "time", "12hr")
+        # Add the same cell method to cube1 and cube2 in
+        # opposite orders.
+        cube1.add_cell_method(cmth1)
+        cube1.add_cell_method(cmth2)
+        cube2.add_cell_method(cmth2)
+        cube2.add_cell_method(cmth1)
+        self.assertFalse(cube1 == cube2)
```
pp-mo: Should we have a **successful** comparison with cell-methods here, or is that covered somewhere else ?

lib/iris/tests/unit/cube/test_Cube.py
stephenworsley: The tests in #3550, as I have written them, use equality. At the moment, those tests aren't working properly so there isn't a full level of confidence (with respect to travis at least) in that pull request. I'm not sure if having equality testing rely upon copy and copy testing rely upon equality is a great idea anyway, so I think at least one of those tests should be done the long way.

lib/iris/tests/unit/cube/test_Cube.py
stephenworsley: This is capturing the behaviour mentioned in the comments on #3530. Adding cell_methods in different orders _should_ cause equality to fail.

lib/iris/cube.py
stephenworsley: Good call.

lib/iris/tests/unit/cube/test_Cube.py
pp-mo: Misunderstanding I think !
What I meant to say was :  do we need another testcase showing that equality testing succeeds, when the cubes have cell methods, but they *do* match.

stephenworsley: ah, yeah, that sounds appropriate.

pp-mo:
Looking very close, I think I'd still push for a test of successful cell-method comparisons, OWOK ! 😁

pp-mo:
Good stuff 👍

stephenworsley:
Closes #3529.

PR #3538, opened 2019-11-15
Use functools.lru_cache for maths dtype caching

bjlittle:
This PR adopts the use of `functools.lru_cache` instead of our own memoize approach for maths operation `dtype` caching.

### DIFF
--- a/lib/iris/analysis/maths.py
+++ b/lib/iris/analysis/maths.py
@@ -8,6 +8,7 @@ Basic mathematical and statistical operations.
 
 """
 
+from functools import lru_cache
 import inspect
 import math
 import operator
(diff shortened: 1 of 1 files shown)


lib/iris/analysis/maths.py
```diff
@@ -27,9 +28,7 @@
 from dask.array.core import broadcast_shapes
 
 
-_output_dtype_cache = {}
-
-
+@lru_cache(maxsize=128, typed=True)
```
trexfeathers: Since 128 is the default value ([docs](https://docs.python.org/3/library/functools.html)), can this be:
```python
@lru_cache(typed=True)
```

bjlittle: @trexfeathers Reasonable request, but I'd rather be explicit here, thanks.

trexfeathers:
> Looks good.  How do we know maxsize=128 is ok?
Worth noting this is the default setting. The source code shows there is some [specific handling](https://github.com/python/cpython/blob/3.6/Lib/functools.py#l551) for when maxsize is reached - the first entry is overwritten. This should be fine

trexfeathers:
Looks good to me and @abooton ✔

bjlittle:

Issue #3483, opened 2019-10-23
[PI] Cube operations support for Ancillary Variables


pp-mo:
Content update 2019-12-09
The notes below identify various problem areas, which leads to the following division of sub-tickets 
(Todo list) :
  * [x] `Cube.copy` and `Cube.__eq__` : #3546
  * [x] cube indexing = `Cube.__getitem__` : #3548
  * [x] concatenate and merge : #3552
  * [x] cube arithmetic : (cf. #3478)
  * [x] statistics = collapse/aggregated_by/rolling_window : #3549
  * [x] "all other cube operations" : ref #3554 . Includes :
    * 'other' indexing-like cube methods : `Cube.subset` and `Cube.intersection`
    * regrid/interpolate : `Cube.regid`, `Cube.interpolate` and all schemes code
    * `Cube.xml()`

----
**Original summary**

Followon from #3422 

A bunch of cube operations are not respecting Ancillary Variables.
We are deferring these issues from #3422.
See [this comment to \#3422](https://github.com/SciTools/iris/pull/3422#issuecomment-545421796).

In principle, for most of these things the handling should be the same as for AuxCoords and (probably) CellMeasures.  Unfortunately, it seems that there are some nasty outstanding bugs relating to CellMeasures anyway -- e.g. they don't survive cube.copy() !.

At least these operations need fixing :
  * ~cube indexing (`__getitem__`) is discarding AVs, even when they don't map to the indexed dimension~
  * ~cube.copy() is discarding AVs~
    * ~and it is also discarding CellMethods !~
  * ~cube equality (`__eq__`) is ignoring AVs~
    * ~but not AuxCoords.~
    * ~CellMethods I'm not sure ?~
  * **updated 2019-11-22** above core cube ops are now fixed
    * [copy](https://github.com/SciTools/iris/pull/3546) : [indexing](https://github.com/SciTools/iris/pull/3548) : [equality](https://github.com/SciTools/iris/pull/3536)/[equality#2](https://github.com/SciTools/iris/pull/3546)
    * however fixing those probably has knock-ons elsewhere within this list of things to fix ..
  * cube arithmetic is discarding AVs
    * but not AuxCoords.
    * CellMethods I'm not sure ?
    * ( note : probably fixed by fixing 'copy' ? )
  * cube.collapsed() is discarding AVs
    * even when they don't map to collapsed dimensions
    * for AuxCoords it applies the same aggregator as the main data
    * for CellMethods I'm not sure
[...]

PR #3546, opened 2019-11-18
Fix copy to include Ancillary variables and Cell measures

stephenworsley:
Fixes another issue in #3483.

### DIFF
--- a/lib/iris/analysis/__init__.py
+++ b/lib/iris/analysis/__init__.py
@@ -186,7 +186,7 @@ class _CoordGroup:
         return any(self.matches(predicate))
 
 
-def coord_comparison(*cubes):
+def coord_comparison(*cubes, object_type="coord"):
     """
     Convenience function to help compare coordinates on one or more cubes
     by their metadata.
@@ -240,7 +240,14 @@ def coord_comparison(*cubes):
         print('All equal coordinates: ', result['equal'])
 
     """
-    all_coords = [cube.coords() for cube in cubes]
+    if object_type == "coord":
+        all_coords = [cube.coords() for cube in cubes]
+    elif object_type == "cell measure":
+        all_coords = [cube.cell_measures() for cube in cubes]
+    elif object_type == "ancillary variable":
+        all_coords = [cube.ancillary_variables() for cube in cubes]
+    else:
+        raise Exception
     grouped_coords = []
 
     # set of coordinates id()s of coordinates which have been processed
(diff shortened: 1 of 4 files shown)


stephenworsley:
This should be merged after rebasing with #3536 so that equality works for testing.

stephenworsley:
This has now been rebased with #3536.

lib/iris/tests/unit/cube/test_Cube.py
```diff
@@ -1710,6 +1710,18 @@ def test(self):
...
 
+    def test_copy_ancillary_variables(self):
+        cube = stock.simple_3d()
+        avr = AncillaryVariable([2, 3], long_name="foo")
+        cube.add_ancillary_variable(avr, 0)
+        self._check_copy(cube, cube.copy())
+
+    def test_copy_cell_measures(self):
+        cube = stock.simple_3d()
+        cms = CellMeasure([2, 3], measure="area", long_name="foo")
```
trexfeathers: Since #3533 "area" is now the default for `measure` and this keyword argument has therefore been removed from all tests

trexfeathers:
While AV's are a new feature, CM's are not. So I think there should be a what's new bugfix entry for the fact that CM's are now copied

stephenworsley:
I have patched in a rough proof of concept fix for equality. This ought to be improvable once #3551 is merged.

trexfeathers:
Notes on the effects on all uses of `cube.copy`. See below for a full list of uses. For those where a problem may be caused I have created #3554, as the problems are relatively minor and can be handled separately, so that this more important fix can proceed now.

- `analysis.PercentileAggregator.post_process` ✔
    - used in `collapsed`, which is also being modified to retain AV's in #3549 so these should be released together
    - allows the end result of a percentile aggregation collapse to retain AV's, CM's
    - for this there are no considerations needed about retention/discarding depending on the collapsed dimension
- `analysis.maths._math_op_common` ✔
    - allows AV's and CM's to be retained after performing arithmetic on a cube - the first cube if multiple cubes are involved
    - no foreseen issues
- `analysis.calculus.cube_delta` ✔
    - actually needs this `copy` fix now that `__getitem__` has been fixed in #3548 - these should both be fixed together
    - (otherwise behaviour is different is coord(s) are circular or not)
- `analysis.calculus._curl_subtract` ✔
    - used in the production of curl cubes
    - original dimensions are retained in the output curl cubes, so no unusual consequences expected if the output cubes now include AV's/CM's
- `analysis._regrid.CurvilinearRegridder` ❌
    - see #3554
- `analysis.trajectory.UnstructuredNearestNeigbourRegridder`, also `.__call__` method ❌
    - see #3554
- `analysis._grid_angles.rotate_grid_vectors` ✔
    - original dimensions are retained in the output cubes, so no unusual consequences expected in the output cubes now include AV's/CM's
- `analysis._grid_angles.rotate_winds` ✔
    - see points on `analysis._grid_angles.rotate_grid_vectors`
- `experimental.stratify.relevel` ❌
    - used for interpolation over a dimension denoted with levels
    - see #3554

trexfeathers:
Thanks @stephenworsley, LGTM ✔

lib/iris/analysis/__init__.py
```diff
@@ -240,7 +240,12 @@ def coord_comparison(*cubes):
         print('All equal coordinates: ', result['equal'])
 
     """
-    all_coords = [cube.coords() for cube in cubes]
+    if object_get is None:
```
pp-mo: This is a public function, so you really need to extend the docstring to cover the changes.
It might also be possible to rename some things here, since e.g. "coords" is no longer always what it says.

Another way is to create an enhanced, private inner routine + leave the public one as a thin wrapper for coords alone.

stephenworsley: I've opened #3558 to address this.

PR #3548, opened 2019-11-18
__getitem__ gets ancillary variables

stephenworsley:
Fixes another issue in #3483.


trexfeathers:
Checked that new AncillaryVariable methods replicated existing CellMeasure methods to the correct extent.
Confirmed that indexed cubes now retain AncillaryVariables when they didn't before.
Checked that new information wouldn't interact badly with other existing code

LGTM ✔

PR #3549, opened 2019-11-18
PI-3483: Ensure collapsed/aggregated_by/rolling_window treat ancillary variables properly

stephenworsley:
Fixes a few issues with #3483. Rellies upon #3548 and #3546 being merged to work properly.


stephenworsley:
This has now been rebased to include #3548 and #3546.

lib/iris/tests/unit/cube/test_Cube.py
```diff
@@ -760,6 +792,20 @@ def test_kwargs(self):
...
+    def test_ancillary_variables_kept(self):
+        cube = self.multi_dim_cube.copy()
+        avr = AncillaryVariable([0, 1], long_name="foo")
+        cube.add_ancillary_variable(avr, 1)
+        res_cube = cube.rolling_window("val", self.mock_agg, 3)
+        self.assertEquals(res_cube.ancillary_variables()[0], avr)
+
+    def test_ancillary_variables_removed(self):
+        cube = self.multi_dim_cube.copy()
+        avr = AncillaryVariable([0, 1, 2, 0, 1, 2], long_name="foo")
```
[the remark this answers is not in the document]
stephenworsley: So the difference in behaviour in this case is caused by defining ancillary variables on different axes, one of which is the same as the axis of the `rolling_window`. I suppose the Ancillary variable could be defined the same for both tests if I set up a 6 by 6 cube, but I wanted to keep its size minimal. Either way, I suppose I could see a case for putting in a few comments to clear things up.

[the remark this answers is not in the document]
trexfeathers: I see, thanks 😊  Defining a common cube will help keep the tests minimal if it only needs creating once, which becomes doubly important when also testing cell measure handling.

trexfeathers:
Also, since this is a change in behaviour for cell measures, there will need to be a what's new entry

lib/iris/cube.py
```diff
@@ -3650,6 +3650,23 @@ def __itruediv__(self, other):
     __pow__ = iris.analysis.maths.exponentiate
     # END OPERATOR OVERLOADS
 
+    def _trim_ancils(self, dims_to_collapse):
+        reduced_cube = self.copy()
+
+        # Collapse any ancillary variables that span the dimension(s) being collapsed
```
pp-mo: "Collapse any ancillary variables that span ..."
Should read "**Remove** any ancillary variables that span ..."

_+1 similar, immediately below_

lib/iris/cube.py
```diff
@@ -3650,6 +3650,23 @@ def __itruediv__(self, other):
     __pow__ = iris.analysis.maths.exponentiate
     # END OPERATOR OVERLOADS
 
+    def _trim_ancils(self, dims_to_collapse):
```
pp-mo: Frankly I'm not keen on the implicit use of 'self.cube', and the fact this is now the only way in which that will be accessed.  Maybe just a more transparent naming can fix that, and/or a docstring/comment of purpose/usage.
Best I can think of on the fly would be something like..
```
    def _trimmed_cube(self, dims_to_collapse):
        # Get a copy of self.cube, minus any ancillary-variables or cell-measures that map the collapse dimensions.
```

pp-mo: **Alternatively**
If we are going to be using, and maybe copying this operation elsewhere, then make it standalone general function that accepts the original cube as an argument, e.g. 
```
def _remove_ancils_and_cms_mapping_dims(cube, dimensions_affected):
    """
    Remove ancillary variables and cell measures that map to specific dimensions.

    To be used by operations that modify or remove dimensions.
    Note: does nothing to (aux)-coordinates.  We expect those will be handled explicitly in the calling operation.

    """

File: lib/iris/util.py [hunk 4]
```diff
@@ -1828,3 +1828,30 @@ def equalise_attributes(cubes):
...
+    Returns a cube copy with (possibly) some cell-measures and ancillary variables removed.
+
+    To be used by operations that modify or remove dimensions.
+    Note: does nothing to (aux)-coordinates.  Those would be handled explicitly by the calling operation.
+
+    """
+    reduced_cube = cube.copy()
+
+    # Remove any ancillary variables that span the dimension(s) being collapsed
+    for ancil in reduced_cube.ancillary_variables():
```
pp-mo: You could iterate over `(reduced_cube.ancillary_variables() + reduced_cube.cell_measures())` here, to reduce this to a single loop.  I think that should be ok, as the docstrings specify that these methods do return _lists_ (not iterables).

stephenworsley: I think doing so would require some kind of generic `cube_dims` and `remove_from_cube` methods on all dimensional metadata. Perhaps this is worthwhile anyway, but it seems non-trivial. Perhaps it would be worth a separate issue adding a bunch of methods to all dimensional metadata objects?

pp-mo: Sorry, I had glossed over the internal differences in the loops.
We do have the cube_dims function now, but the generic "cube.remove" is I think a step too far.
So having seen that, I withdraw the suggestion :  I don't think we can justify the further complication !

lib/iris/util.py
```diff
@@ -1828,3 +1828,30 @@ def equalise_attributes(cubes):
         for key in list(cube.attributes.keys()):
             if key not in common_keys:
                 del cube.attributes[key]
+
+
+def _remove_ancils_and_cms_mapping_dims(cube, dims):
```
pp-mo: After some consideration, I guess this is a reasonable place for it, as we already have other general-purpose private routines here.
I know it was my proposal, but can you think of a handier short name ?
(Anticipating complaint from @bjlittle [as on this one](https://github.com/SciTools/iris/pull/3562#discussion_r351185418) !)

lib/iris/tests/unit/cube/test_Cube.py
```diff
@@ -283,6 +283,10 @@ def setUp(self):
             npts = cube.shape[i_dim]
             coord = DimCoord(np.arange(npts), long_name=name)
             cube.add_dim_coord(coord, i_dim)
+        self.ancillary_variable = AncillaryVariable([0, 1], long_name="foo")
```
pp-mo: I think adding this to a testclass called specifically `Test_collapsed__lazy` is a bit cheeky !
The actual (pre-existing) testcases show that this class is _specifically_ to test operation with lazy data.

So I think we should instead add a new testclass to be the home for these new tests.
As we don't have a "Test_collapsed", and we're not even aiming to do that here+now, I would just call it something like 'Test_collapsed__cellmeasure_ancils` + keep it minimal for this purpose.

By contrast, I think the changes to rolling-window and aggregated-by classes below are fine as they are.

pp-mo:
Hi @stephenworsley, reviewed what is left to do, so here is my list of what's outstanding :
  * [ ] [@trexfeathers request to add a whatsnew](https://github.com/SciTools/iris/pull/3549#issuecomment-557012731)
  * [ ] [correct comment](https://github.com/SciTools/iris/pull/3549#discussion_r349582280)
  * [ ] [more concise routine name](https://github.com/SciTools/iris/pull/3549/files#r351227847)
  * [ ] [move 'collapsed' tests to own class](https://github.com/SciTools/iris/pull/3549#pullrequestreview-323650392)

pp-mo:
Nice work @stephenworsley, thanks for patience !!

PR #3550, opened 2019-11-18
PI-3473: Netcdf saving of ancillary variables

pp-mo:
Saving ancillary variables, with some testcases based on CDL comparisons.

~There is some unfinished business here where the tests might be improved,~
~if and when they can use other aspects of ancillary support :~
**UPDATE:**
FIxed that.
Nearly good to go?

### DIFF
--- a/lib/iris/fileformats/netcdf.py
+++ b/lib/iris/fileformats/netcdf.py
@@ -1256,6 +1256,10 @@ class Saver:
                 cf_name = self._name_coord_map.name(element)
 
             if cf_name is not None:
+                if role_attribute_name == "cell_measures":
+                    # In the case of cell-measures, the attribute entries are not just
+                    # a var_name, but each have the form "<measure>: <varname>".
+                    cf_name = "{}: {}".format(element.measure, cf_name)
                 element_names.append(cf_name)
 
         # Add CF-netCDF references to the primary data variable.
--- a/lib/iris/fileformats/netcdf.py
+++ b/lib/iris/fileformats/netcdf.py
@@ -1269,33 +1269,11 @@ class Saver:
             Names associated with the dimensions of the cube.
 
         """
+        return self._add_inner_related_vars(
+            cube, cf_var_cube, dimension_names,
+            cube.aux_coords, self._create_cf_coord_variable, 'coordinates')
 
-#         return self._add_inner_related_vars(
-#             cube, cf_var_cube, dimension_names,
-#             cube.aux_coords, self._create_cf_coord_variable, 'coordinates')
-
-        auxiliary_coordinate_names = []
-        # Add CF-netCDF variables for the associated auxiliary coordinates.
-        for coord in sorted(cube.aux_coords, key=lambda coord: coord.name()):
-            # Create the associated coordinate CF-netCDF variable.
-            if coord not in self._name_coord_map.coords:
-                cf_name = self._create_cf_coord_variable(
-                    cube, dimension_names, coord
-                )
-                self._name_coord_map.append(cf_name, coord)
-            else:
-                cf_name = self._name_coord_map.name(coord)
-
-            if cf_name is not None:
-                auxiliary_coordinate_names.append(cf_name)
-
-        # Add CF-netCDF auxiliary coordinate variable references to the
-        # CF-netCDF data variable.
-        if auxiliary_coordinate_names:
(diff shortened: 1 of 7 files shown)


pp-mo:
Hi @stephenworsley 
Now rebased, should be GTG ! :pray:

lib/iris/tests/test_netcdf.py
```diff
@@ -964,6 +964,111 @@ def test_no_global_attributes(self):
...
+        testcube_1 = stock.realistic_3d()
+        ancil = iris.coords.AncillaryVariable(
+            np.zeros(testcube_1[0].shape), long_name="latlon_refs", units="s"
+        )
+        testcube_1.add_ancillary_variable(ancil, (1, 2))
+
+        testcube_2 = testcube_1.copy()
+        testcube_2.units = "m"
+        testcube_2.rename("alternate_data")
+        # TODO: this should not be necessary, when cube.copy() handles ancils.
```
stephenworsley: Now that #3546 is merged, does rebasing and modifying this code work as expected?

lib/iris/tests/test_netcdf.py
```diff
@@ -964,6 +964,111 @@ def test_no_global_attributes(self):
...
+        testcube_2.rename("alternate_data")
+
+        ancil1 = iris.coords.AncillaryVariable(
+            np.zeros(testcube_1[0].shape), long_name="latlon_refs", units="s"
+        )
+        testcube_1.add_ancillary_variable(ancil1, (1, 2))
+
+        ancil2 = ancil1.copy()
+        ancil2.data[0, 0] += 1.0
+        # TODO: needs modifying, when cube.copy() handles ancils.
```
stephenworsley: As above, this should be possible to address now #3546 is merged.

lib/iris/fileformats/netcdf.py
```diff
@@ -1093,14 +1093,18 @@ def write(
...
+        # Add the auxiliary coordinate variables and associate the data
         # variable to them
         self._add_aux_coords(cube, cf_var_cube, dimension_names)
 
-        # Add the cell_measures variable names and associate the data
+        # Add the cell_measures variables and associate the data
         # variable to them
         self._add_cell_measures(cube, cf_var_cube, dimension_names)
 
+        # Add the ancillary_variables variables and associate the data variable
```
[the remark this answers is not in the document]
pp-mo: "Variable names" is really _wrong_, I think.
What all these "_add" functions do is to is create variables in the datafile.

The line above just says "Add coordinate variables", which I think is the right way to describe it.

lib/iris/fileformats/netcdf.py
```diff
@@ -1714,26 +1760,79 @@ def _create_cf_cell_measure_variable(
...
+        Returns:
+            The string name of the associated CF-netCDF variable saved.
+
+        """
+        # Note: currently shares variable creation code with cell-measures.
+        return self._inner_create_cf_cellmeasure_or_ancil_variable(
+            cube, dimension_names, ancillary_variable
+        )
+
+    # TODO: this was renamed from "_create_cf_variable" .
```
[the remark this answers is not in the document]
pp-mo: Not that I can think of.
Good spot! 😀

lib/iris/tests/test_netcdf.py
pp-mo: Well, it probably *ought* to fail then, I think, complaining that the ancillary to be added is identical to one already there.
I will investigate  this ...

lib/iris/tests/test_netcdf.py
pp-mo: :+1:

pp-mo:
> it probably ought to fail then, I think, complaining that the ancillary to be added is identical to one already there.

Rebased, and it did fail as predicted :+1:
Now fixed.

lib/iris/tests/test_netcdf.py
pp-mo: Note: this actually did *not* need code changing.
The cube is copied before an ancil is attached, so no collision between the two different versions occurs when adding to the second cube.
The two are stored separately (with disambiguated var_names), _because_ they contain different data.  
All good :+1: !

lib/iris/fileformats/netcdf.py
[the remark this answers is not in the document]
stephenworsley: Ah, okay I think I mixed up the red and green lines above here.

PR #3551, opened 2019-11-19
Generic 'cube_dims' function for all _DimensionMetadata objects.

pp-mo:
The idea of this is to enable writing code that handles _DimensionalMetadata objects in a generic way.
For example, it will avoid the need for [the extra method keyword here](https://github.com/SciTools/iris/blob/c0c715849602f2d98c7b23f5f2f36304389a6989/lib/iris/fileformats/netcdf.py#L1695).

### DIFF
--- a/lib/iris/coords.py
+++ b/lib/iris/coords.py
@@ -173,6 +173,22 @@ class _DimensionalMetadata(CFVariableMixin, metaclass=ABCMeta):
 
         return new_metadata
 
+    @abstractmethod
+    def cube_dims(self, cube):
+        """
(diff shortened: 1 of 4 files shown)


pp-mo:
@stephenworsley I believe we were discussing that this might contribute to #3546,
by making it easier to generalise the operation of `iris.analysis.coord_comparison` to other _DimensionalMetadata objects.

pp-mo:
Thanks @stephenworsley 
Got Black going now.
Please re-check!

pp-mo:
Also rebased #3550 onto this, as it helps.

Issue #3415, opened 2019-09-24
Replace pyke package with other solutions for netcdf loading


pp-mo:
I think Pyke is near-obsolete, at best static with no development since 2010 as fas as I can tell
This is a bit worrying, to say the least.

We could replace it with another Python logic programming solution, but they all seem in a similar state IMHO (check it out).

Instead, I reckon that the actual rules operation is really quite simple and not hard to rewrite in ordinary Python.  Which would certainly make the rules easier to maintain.

BTW, I don't think we are really making ideal usage of the logic programming concept in the loader anyway :  A rules-type language would be a really *good* solution to assigning netcdf variables to specific CF categories.  This is currently handled by the `iris.fileformats.cf` wher the `CfReader` using the `identify` methods in the various subclasses of `CFVariable`. 
Either that is a missed opportunity or, possibly, it just shows that pure Python solutions are actually perfectly adequate !

pp-mo:
Detail : state of play on possible logic programming packages.

[Pyke](https://sourceforge.net/projects/pyke/)
 * Still on sourceforge, now pretty ancient in itself !
  * latest, and only public, release in 2010
  * 7 comments in last 2 years

Best of others seem to be [logpy(or 'kanren' ?)](https://github.com/logpy/logpy) and [pyswip](https://github.com/yuce/pyswip)
Both on github.  Both have better syntax. But neither is seeing much action...
  * [logpy releases](https://github.com/logpy/logpy/releases) 0.2.1 and 0.2.2, both on 1 Dec 2015
  * [pyswip releases](https://github.com/yuce/pyswip/releases)
    * 0.2.5/6/7 May-Jun 2018
    * 0.2.8 Feb 20 2019 : doc-changes only, nothing functional

bjlittle:
@pp-mo Guilty as charged.

I've been meaning to purge `pyke` for a long time now. In fact, I've even got a branch with it half done... just not had the time or capacity to finish the job :cry: 

I ended up taking the pure Python route, but still building on the `iris.fileformats.cf` layer.

It would be very healthy of us to purge ourselves of the `pyke` dependency, as it is effectively a dead package, and I'd be able to sleep better at night. However, it would be great to get ASV in place beforehand, so that we can measure the performance impact...

Nudge @trexfeathers :wink:

bjlittle:
That said, I'd even float the concept of making `iris.fileformats.cf` a standalone package, e.g., akin to `cf-units`. 

[...]

PR #3556, opened 2019-11-20
PI-3473: Netcdf loading ancillary variables

pp-mo:
Load ancils for netcdf.
**STATUS: why this targets the 'launch_ancils' feature branch**
We aren't yet sure that we will be ready to "release" ancillary variables in Iris 3.0
(mostly because there are a lot of places where specific handling is needed to avoid creating bad cubes).
Hence this is the key PR on the [launch_ancils feature branch](https://github.com/SciTools/iris/tree/launch_ancils).

In practice, some support has already been merged to master.
Especially #3422, which also already makes the cube API for ancils visible.
However.. bad things don't happen until cubes contain ancils, which will generally *only happen to users when we merge this PR*, following which they can be generated when loading.
Likewise, we are also "holding back" on other PRs, targetting the feature branch instead of master), e.g. [documentation additions](https://github.com/SciTools/iris/pull/3557)

---

NOTES: 
  * also silently fixes an hitherto unknown problem, whereby attributes of cell-measures were not loaded.
    * possibly there should be an extra, specific test for that ?  But I think it is not really needed.
  * only one test, at present


lib/iris/tests/test_netcdf.py
```diff
@@ -240,6 +250,52 @@ def test_cell_methods(self):
 
         self.assertCML(cubes, ("netcdf", "netcdf_cell_methods.cml"))
 
+    def test_ancillary_variables(self):
```
stephenworsley: It might also be worthwhile also testing for cell measures. As far as I can tell, they aren't currently being tested for.

lib/iris/fileformats/_pyke_rules/fc_rules_cf.krb
```diff
@@ -1941,36 +1957,36 @@ fc_extras
...
         cf_var = engine.cf_var
         cube = engine.cube
         attributes = {}
 
         # Get units
-        attr_units = get_attr_units(cf_cm_attr, attributes)
+        attr_units = get_attr_units(cf_cm_var, attributes)
 
-        data = _get_cf_var_data(cf_cm_attr, engine.filename)
+        data = _get_cf_var_data(cf_cm_var, engine.filename)
```
stephenworsley: For consistency, should this have a comment similar to this line? https://github.com/SciTools/iris/blob/1178de727b54e29efef7bc178feabfec1d8b164e/lib/iris/fileformats/_pyke_rules/fc_rules_cf.krb#L2018-L2019

lib/iris/fileformats/netcdf.py
```diff
@@ -597,31 +606,37 @@ def _load_cube(engine, cf, cf_var, filename):
...
+
+    # Also populate attributes of the top-level cube itself.
+    add_unused_attributes(cube, cf_var)
+
+    # Work out reference names for all the coords.
     names = {
         coord.var_name: coord.standard_name or coord.var_name or "unknown"
         for coord in cube.coords()
     }
-
```
stephenworsley: Should there still be whitespace here?

lib/iris/tests/test_netcdf.py
pp-mo: Ok..

lib/iris/fileformats/_pyke_rules/fc_rules_cf.krb
pp-mo: Ok, but "consistency" is a bit of a lost cause in this code IMHO.  Not least because we have far too much code duplication code, as here !
( N.B. We do still hope to replace all of this with Python code #3415  )

lib/iris/fileformats/netcdf.py
pp-mo: Ok, I suppose so.
I might have hoped that  Black would take care of this, but clearly the results are not totally standardised ..

pp-mo:
Thanks @stephenworsley 
Hopefully addressed your previous comments now.
Please re-review.

lib/iris/tests/test_netcdf.py
```diff
@@ -240,6 +250,104 @@ def test_cell_methods(self):
...
+                int64 axv(axv) ;
+                    axv:units = "1" ;
+                    axv:long_name = "x" ;
+                double my_areas(ayv, axv) ;
+                    my_areas:units = "m2" ;
+                    my_areas:long_name = "standardised cell areas" ;
+                    my_areas:custom = "extra-attribute";
+            data:
+                axv = 11, 12, 13;
+                ayv = 21, 22, 23;
```
stephenworsley: `ayv` should be length 2.

pp-mo: Thanks.
I'm surprised it didn't error that!

pp-mo:
Hi @stephenworsley 
Fixed ?

stephenworsley:
@pp-mo  Looks good. I think this just needs rebasing to make Travis pass

PR #3557, opened 2019-11-21
PI-3473: Whatsnews relating to ancillary load + save.

pp-mo:
Covers changes from #3550 and #3556

**NOTE: why this targets the 'launch_ancils' feature branch** 
We are still reserving the right to ***not*** announce ancillary variables in Iris 3.0, in which case this will be deferred from merging to master :  see [comment here](https://github.com/SciTools/iris/pull/3556#issue-343470988).

### DIFF
--- a/docs/iris/src/whatsnew/contributions_3.0.0/bugfix_2019-Nov-21_cell_measure_attributes.txt
+++ b/docs/iris/src/whatsnew/contributions_3.0.0/bugfix_2019-Nov-21_cell_measure_attributes.txt
@@ -1,0 +1,2 @@
+* Fixed a bug where the attributes of cell methods in netcdf-CF files were discarded on
+  loading.  They now appear on the CellMethod in the loaded cube.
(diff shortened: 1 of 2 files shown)


docs/iris/src/whatsnew/contributions_3.0.0/bugfix_2019-Nov-21_cell_measure_attributes.txt
```diff
@@ -0,0 +1,2 @@
+* Fixed a bug where the attributes of cell methods in netcdf-CF files were discarded on
+  loading.  They now appear on the CellMethod in the loaded cube.
```
stephenworsley: I believe this should refer to cell measures/CellMeasure rather than cell methods/CellMethod.

pp-mo: Whoops! 
I can't believe I'm _still_ getting these names mixed up.
Will fix...

PR #3558, opened 2019-11-21
PI-3483: Make changes to coord_comparison private

stephenworsley:
This addresses comments on #3546. At a minimum, this causes `coord_comparison` to have the same behaviour as it did previously. The internals of the function still refer to coords rather than, say, ancillary variables. Ideally there would be a more generic name to use as keys for the generic function (preferably one less cumbersome than 'dimensional metadata').

### DIFF
--- a/lib/iris/analysis/__init__.py
+++ b/lib/iris/analysis/__init__.py
@@ -187,6 +187,14 @@ class _CoordGroup:
 
 
 def _dimensional_metadata_comparison(*cubes, object_get=None):
+    """
+    This is a generalised form of :func:`coord_comparison`. See that function
+    for a more detailed description.
+
+    Additionally, this function can compare over different types of dimensional
+    metadata with the argument object_get. For example, :func:`coord_comparison`
+    effectively has object_get=:meth:`iris.cube.Cube.coords`.
+    """
     if object_get is None:
         from iris.cube import Cube
 
--- a/lib/iris/analysis/__init__.py
+++ b/lib/iris/analysis/__init__.py
@@ -194,6 +194,7 @@ def _dimensional_metadata_comparison(*cubes, object_get=None):
     Additionally, this function can compare over different types of dimensional
     metadata with the argument object_get. For example, :func:`coord_comparison`
     effectively has object_get=:meth:`iris.cube.Cube.coords`.
+
     """
     if object_get is None:
         from iris.cube import Cube
--- a/lib/iris/analysis/__init__.py
+++ b/lib/iris/analysis/__init__.py
@@ -186,60 +186,7 @@ class _CoordGroup:
         return any(self.matches(predicate))
 
 
-def coord_comparison(*cubes, object_get=None):
-    """
-    Convenience function to help compare coordinates on one or more cubes
-    by their metadata.
-
-    Return a dictionary where the key represents the statement,
-    "Given these cubes list the coordinates which,
-    when grouped by metadata, are/have..."
-
-    Keys:
-
-    * grouped_coords
-       A list of coordinate groups of all the coordinates grouped together
-       by their coordinate definition
-    * ungroupable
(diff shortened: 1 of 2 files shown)


lib/iris/analysis/__init__.py
```diff
@@ -186,60 +186,7 @@ def matches_any(self, predicate):
...
-    * transposable
-       A list of non equal, same data dimensioned, non scalar coordinate groups
-
-    Example usage::
-
-        result = coord_comparison(cube1, cube2)
-        print('All equal coordinates: ', result['equal'])
-
-    """
+def _dimensional_metadata_comparison(*cubes, object_get=None):
```
pp-mo: Just add a comment or docstring to explain 
  * this is a 'generalised form' of the coord_comparison function
  * see there for a detailed behaviour description (especially, what it returns)
  * the 'object_get' is a fetch-all function for cube _DimensionalMetadata contents, like `Cube.coords()` (the default.

bjlittle: @stephenworsley I don't quite understand the reason for making this private?
Wouldn't there be general utility in making this public and exposing this behaviour?

lib/iris/cube.py
```diff
@@ -3575,23 +3575,23 @@ def __eq__(self, other):
                 )
 
             if result:
-                coord_comparison = iris.analysis.coord_comparison(
+                comparison = iris.analysis._dimensional_metadata_comparison(
```
bjlittle: @stephenworsley Seems like we're baking in some indirection here, where the behaviour of our public API is now depending on the private behaviour of `_dimensional_metadata_comparison`. 

This means that indirectly we can change the public API behaviour of `iris.cube.Cube.__eq__` by changing `iris.analysis.__init__._dimensional_metadata_comparison`... that seems a tad dangerous to me.

lib/iris/analysis/__init__.py
stephenworsley: @bjlittle My thinking behind making it private is that this should always serve as the most generic version of comparison for dimensional metadata. If it needs to change in the future to make it more generic, there is room to do so. To expose the new comparison behaviour, would it make sense to create specific public functions `cell_measure_comparison` and `ancillary_variable_comparison` instead?

stephenworsley: @bjlittle With that said, much of the output form `_dimensional_metadata_comparison` reads as coordinate specific. If this function were to be made public somehow, I think this would need to be addressed.

lib/iris/cube.py
pp-mo: > we can change the public API behaviour of iris.cube.Cube.__eq__ by changing iris.analysis.__init__._dimensional_metadata_comparison

I don't see why this is a particular concern.  It should be obvious that we *wouldn't* change the behaviour of the public function without very careful considering + management, and I think the public API dependence on the private function should be clear enough to any potential developer that it wouldn't happen by accident.

Of course, if we make the more general private function public all such problems would go away.  But IMHO that seems like a good deal more work that we just don't really need to commit to at this stage.

lib/iris/analysis/__init__.py
pp-mo: Personally, I'm not convinced that we need public functionality for this.
If we _do_ want that, then I don't like ***either*** the existing form with the magic 'object_get' keyword, ***or*** the prospect of additional specific routines for `cell_measure_comparison` and `ancillary_variable_comparison`.  
Instead, I would push for a general-purpose `dimensional_metadata_comparison` routine with a rather nicer like an 'element_type' switch.

But in the end, I think we can just not do that, for now.

pp-mo:
#3562 aims to replace, following the above discussions.

Can we close this ?

stephenworsley:
@pp-mo Sounds fair to me.

PR #3562, opened 2019-11-26
Remove 'iris.analysis.coord_comparison' from public API.

pp-mo:
Replaces #3558 
Following discussions there, I'm proposing to *remove* the original, public 'coord_comparison' function.  Replaced with an augmented, private version for internal use only.

In offline discussions with @bjlittle @lbdreyer , we agreed this was always rather over-complicated and under-explained for a user public method.  Plus, the key `_CoordGroup` class it relies on was always private anyway.
Since it was practically impossible to understand or use with confidence, without also reading the sourcecode, we decided to simply retire it rather than providing an enhanced public function (or functions).

### DIFF
--- a/lib/iris/analysis/__init__.py
+++ b/lib/iris/analysis/__init__.py
@@ -190,22 +190,48 @@ def _dimensional_metadata_comparison(*cubes, object_get=None):
     Convenience function to help compare coordinates, cell-measures or
     ancillary-variables, on one or more cubes, by their metadata.
 
-    Return a dictionary where the key represents the statement,
-    "Given these cubes list the elements which,
-    when grouped by metadata, are/have..."
-
     .. Note::
 
         Up to Iris 2.x, this _used_ to be the public API method
         "iris.analysis.coord_comparison".
         It has since been generalised, and made private.
-        However, the cube elements handled are still referred to as 'coords' /
-        'coordinates' throughout, for backwards compatibility :  In practice, these can
-        be either `iris.coords.Coord`, or `iris.coords.CellMeasure` or
+        However, the cube elements handled are still mostly referred to as 'coords' /
+        'coordinates' throughout, for simplicity :  In fact, they will all be either
+        `iris.coords.Coord`, `iris.coords.CellMeasure` or
         `iris.coords.AncillaryVariable`, the cube element type being controlled by the
         'object_get' keyword.
 
-    Keys:
+    Args:
+
+    * cubes (iterable of `iris.cube.Cube`):
+        a set of cubes whose coordinates, cell-measures or ancillary-variables are to
+        be compared.
+
+    Kwargs:
+
+    * object_get (callable(cube) or None):
+        If not None, this must be a cube method returning a list of all cube elements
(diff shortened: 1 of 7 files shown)


lib/iris/analysis/__init__.py
```diff
@@ -186,15 +185,26 @@ def matches_any(self, predicate):
...
-    by their metadata.
+    Convenience function to help compare coordinates, cell-measures or
+    ancillary-variables, on one or more cubes, by their metadata.
 
     Return a dictionary where the key represents the statement,
-    "Given these cubes list the coordinates which,
+    "Given these cubes list the elements which,
     when grouped by metadata, are/have..."
 
+    .. Note::
```
stephenworsley: Placing this note between: `when grouped by metadata, are/have..."` and `Keys:` somewhat breaks up the flow of the docstring the `are/have` relates to the keys listed below. This note might be better placed elsewhere.

pp-mo: Well I didn't want to leave it until after this long long long list of keys in the returned data
-- Not least because, that info should now  be read in the light of this note (i.e. it assumes everything is a coord).

Also, none of this gets rendered, now it is private, so I don't think style is such a key factor.
( The existing is not in full alignment with the style anyway, I reckon )

pp-mo:
Thanks @stephenworsley .
Made a change, hoping to address your issue.
Do you have more changes to request ?

stephenworsley:
@pp-mo Looks good now, I reckon this is good to merge.

lib/iris/analysis/__init__.py
```diff
@@ -186,16 +185,53 @@ def matches_any(self, predicate):
         return any(self.matches(predicate))
 
 
-def coord_comparison(*cubes, object_get=None):
+def _dimensional_metadata_comparison(*cubes, object_get=None):
```
bjlittle: @pp-mo Awesome! 👍 

My only suggestion here, before we commit to this, is questioning the name of this private function (all 30+ characters of it).

I'm just making the explicit point that it can be anything that we want - it's private. I'm all for meaningful, appropriate names, rather than those that are cryptic, abstract or obfuscated, but now is probably the time to suggest shorter alternatives? It is a bit of a mouthful, right? 😮 

Any suggestions? Can I start the bun-fight with simply `_metadata_compare`?

pp-mo: > @bjlittle questioning the name of this private function

We just missed the boat on this.

I get the point, but I couldn't think of anything more concise and still clear.
Probably the term DimensionalMetadata is at fault here, also being a bit of a mouthful.  

I guess we could have `_dm_comparison`.
Do you want to propose a post-fix in separate issue ?

bjlittle: @pp-mo @stephenworsley No worries. Naming a thing is the hardest part, strangely. 

I recently asked @lbdreyer for an alternative name to `_DimensionalMetadata`.... but we came up blank 😕 

This isn't a high priority, by any means, just wanted to raise it as an obvious point to consider. We can stick with what we've got, and change at a later date in a follow-up PR, if we have an epiphany 💡

PR #3564, opened 2019-11-27
_regrid_area_weighted_array:  move indices variables

abooton:
Move the indices variable nearer to its use within the main code, making it easier to read


pp-mo:
LGTM! :+1: 
Is this your first @abooton  ??

Issue #3552, opened 2019-11-20
Ancillary Variables - Fix concatenate, merge


trexfeathers:
See #3422  [this comment](https://github.com/SciTools/iris/pull/3422#issuecomment-545480015) and #3483  [this comment](https://github.com/SciTools/iris/issues/3483#issuecomment-545503186)

Ancillary variable support should not be released without ensuring `concatenate` and `merge` handle AV's correctly

---
### Acceptance Criteria
All to be shown working with tests
  - [x] cubes with matching ancillary variables can merge to a result containing the ancillary
  - [x] likewise.., cubes with matching ancillary variables (not along the axis of concatenation) can concatenate to a result containing the ancillary
  - [x] a cube with an ancillary variables and one without (a matching one) will not merge;
     * ... and a descriptive error message is provided, similar to other merge failures
     * ... and the special extra message provided by "merge_cube" describes the precise difficulty
  - [x] cubes with differing ancillary variables (i.e. same names but some difference) will not merge;
     * ... likewise, message provisions as above
  - [x] likewise.., cubes with/without ancillary variables will not concatenate, and produce appropriate messages
  - [x] likewise.., cube with differing ancillary variables will not concatenate, and produce appropriate messages
  - [x] cubes with similar ancillary variables which span the axis of concatenation, will concatenate to form a new ancillary variable which spans the concatenated cube
 - [x] as all above with cell measures

In short, ancillary variables and cell measures will behave the same as (non-scalar) aux coords.

----
(@pp-mo 2019-11-22) Notes on changes I just made to this ...
  * added the "experience-needed: High" label because the existing merge+concatenate code is quite complex, so getting it + proposing changes with confidence is hard !
  * added some initial thoughts on acceptance criteria : not necessarily complete; I just updated this while working on something else

PR #3566, opened 2019-11-28
PI-3552: Add concatenate support for Ancilary Variables and Cell Measures

stephenworsley:
Addresses the concatenate half of #3552.

PR #3569, opened 2019-11-29
PI-3473: Add Ancillary Variable support to merge

stephenworsley:
Addresses half of #3552. Now `merge` treats Ancillary Variables the same as Cell Measures.

### DIFF
--- a/lib/iris/_merge.py
+++ b/lib/iris/_merge.py
@@ -328,7 +328,13 @@ class _CoordSignature(
 class _CubeSignature(
     namedtuple(
         "CubeSignature",
-        ["defn", "data_shape", "data_type", "cell_measures_and_dims"],
+        [
+            "defn",
+            "data_shape",
+            "data_type",
+            "cell_measures_and_dims",
+            "ancillary_variables_and_dims",
+        ],
     )
 ):
     """
@@ -348,6 +354,9 @@ class _CubeSignature(
 
     * cell_measures_and_dims:
         A list of cell_measures and dims for the cube.
+
+    * ancillary_variables_and_dims:
+        A list of ancillary variables and dims for the cube.
 
     """
 
@@ -441,6 +450,11 @@ class _CubeSignature(
             msgs.append(msg.format(self.data_type, other.data_type))
         if self.cell_measures_and_dims != other.cell_measures_and_dims:
             msgs.append("cube.cell_measures differ")
+        if (
+            self.ancillary_variables_and_dims
+            != other.ancillary_variables_and_dims
+        ):
+            msgs.append("cube.ancillary_variables differ")
 
         match = not bool(msgs)
         if error_on_mismatch and not match:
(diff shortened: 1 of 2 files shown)


lib/iris/_merge.py
```diff
@@ -441,6 +450,11 @@ def match(self, other, error_on_mismatch):
             msgs.append(msg.format(self.data_type, other.data_type))
         if self.cell_measures_and_dims != other.cell_measures_and_dims:
             msgs.append("cube.cell_measures differ")
+        if (
+            self.ancillary_variables_and_dims
+            != other.ancillary_variables_and_dims
```
pp-mo: I think it could be worth emphasizing here that although it looks like these two av-and-dims lists might differ only in their occurrence order, in fact this is ok as they are always stored in a cube in a specific sorted order, through [this code](https://github.com/SciTools/iris/blob/479b2cf80fc75c7b327c01a0f6a2b139ea6e9896/lib/iris/cube.py#L1130)

It caused me some thought, anyway !

pp-mo:
> Now merge treats Ancillary Variables the same as Cell Measures.

Unfortunately I think this approach is probably wrong, as discussed offline with @bjlittle ...

We think that ***not*** merging (stacking) cell-measures makes ok sense, as these generally describe the "data spatial sampling grid"  in dims X, Y and possibly Z.  
So we expect that grid to be probably complete within each source file, thus the grid will not extend over "extra" dimensions.

However in contrast, Auxiliary Variables can easily map over any and all dimensions of the data, particularly when they are used as status flags : they could have a value at every single datapoint.
So these should be stacked up by merging in a manner analogous to the data, or more exactly like AuxCoords.

----

On the other hand ...

I see this was ***not*** in the original Acceptance Criteria on #3552, as opposed with concatenation :
>  cubes with similar ancillary variables which span the axis of concatenation, will concatenate to form a new ancillary variable which spans the concatenated cube
> cubes with differing ancillary variables (i.e. same names but some difference) will not merge;

So, **Having seen that we can't currently do this with AuxCoords either, I wonder if we can leave this as extended functionality for later delivery ?**

@stephenworsley has shown that :
  * we *can* concatenate in these cases.
  * .. using first a `iris.util.new_axis` ..
  * .. then a fairly short utility to "promote" Ancillaries (?and AuxCoords) to mapping the 'new axis',
    * ( replace with reshaped version with extra length-1 dimension, and new-dim added into dims-map )

stephenworsley:
> **can we leave this as extended functionality for later delivery ?**

@pp-mo Does this mean that iris 3.0 should discard/ignore ancillary variables on merge? If this is the case, perhaps a warning should be given when this happens.
Or are you suggesting we keep the current functionality of the pull request with a view that this can be changed/extended in the future?

pp-mo:
> @stephenworsley Does this mean that iris 3.0 should discard/ignore ancillary variables on merge? ... Or are you suggesting we keep the current functionality 

The latter I think: That is, not discard but only carry over matching versions. A statement of that would be useful, somewhere but frankly I'm not sure where. Suggest if you like..

I still need to check this out further, including test coverage, but I will need a bit longer. Sorry to be so slow today.

pp-mo:
Ok, after long thinks about the concatenate #3566, I've convinced myself that this one really is much simpler.  Thanks @stephenworsley !

PR #3571, opened 2019-12-02
Tweak area weighting regrid: reorder variables

abooton:
Reorder some of the variables to be nearer their use in the code.
This is preparatory work for refactoring weighting calculations.


pp-mo:
Thanks @SciTools/mo-read.  
Slightly more logical, but very small change.  Awaiting the big + tricky one !

PR #3577, opened 2019-12-04
Fix problems with export and echo command.

### LINKED ISSUE #3542: Black is invisible
There was no whatsnew for #3518 to publicise what's been done.
I think it's worth it just to inform people that the code format testing has been updated.

Meanwhile, even if you go looking in the developers' guide, the new information doesn't appear for some odd reason.

So [this line](https://github.com/bjlittle/iris/blame/15e7f2c8fe2c24c90f41ee7600e949535fb4e772/docs/iris/src/developers_guide/index.rst#L15) says that the new "code_format.rst" page should now appear on the devguide menu, but [here in the built docs](https://scitools-docs.github.io/iris/master/developers_guide/index.html) it mysteriously does not.

I couldn't work out what's wrong here, but if we get a fix to give this some publicity, then obviously we'll fix that too...

---
Acceptance Criteria:
- [x] code_format.html appears in the developer's guide in the "latest" documentation (fixed in #3577 and #3578)
- [x] Add whats new docs_change to mention code_format.html page (see #3581)

pp-mo:
This intends to re-instate automatic docs build for the Github branches
Closes #3542

I think this is now correct.
I have left in a debug line that echoes the control variables, including the 'ORG' organisation identity (from the Travis SLUG), as I think this is generally useful to see in the logs.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -162,7 +162,11 @@ script:
     fi
 
   # Split the organisation out of the slug. See https://stackoverflow.com/a/5257398/741316 for description.
-  - export ORG=(${TRAVIS_REPO_SLUG//\// })
+  - >
+    ORG=$(echo ${TRAVIS_REPO_SLUG} | cut -d/ -f1);
+    export ORG
+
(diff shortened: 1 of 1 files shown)


bjlittle:
LGTM 👍

PR #3578, opened 2019-12-04
Revert to single-line command for doctr invocation.

pp-mo:
Another go at fixing #3542 
Since #3577 seems to have been bugged.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -171,8 +171,7 @@ script:
   # When we merge a change to SciTools/iris, we can push docs to github pages.
   # At present, only the Python 3.7 "doctest" job does this.
   # Results appear at https://scitools-docs.github.io/iris/<<branchname>>/index.html
-  - >
-    if [[ "${ORG}" == 'SciTools' && "${TRAVIS_EVENT_TYPE}" == 'push' && "${PUSH_BUILT_DOCS}" == 'true' ]]; then
+  - if [[ "${ORG}" == 'SciTools' && "${TRAVIS_EVENT_TYPE}" == 'push' && "${PUSH_BUILT_DOCS}" == 'true' ]]; then
       cd ${INSTALL_DIR};
       pip install doctr;
       doctr deploy --deploy-repo SciTools-docs/iris --built-docs docs/iris/build/html


.travis.yml
```diff
@@ -171,8 +171,7 @@ script:
   # When we merge a change to SciTools/iris, we can push docs to github pages.
   # At present, only the Python 3.7 "doctest" job does this.
   # Results appear at https://scitools-docs.github.io/iris/<<branchname>>/index.html
-  - >
-    if [[ "${ORG}" == 'SciTools' && "${TRAVIS_EVENT_TYPE}" == 'push' && "${PUSH_BUILT_DOCS}" == 'true' ]]; then
+  - if [[ "${ORG}" == 'SciTools' && "${TRAVIS_EVENT_TYPE}" == 'push' && "${PUSH_BUILT_DOCS}" == 'true' ]]; then
```
lbdreyer: It seems strange the `>` doesn't work. 
Have you tested that this line would work?

pp-mo: The ">" form seems to leaves newlines in it (or possibly carriage-returns, if there is a lined problem??).
So testing with the yaml package, I get...
```
>>> with open('.travis.yml') as fo:
...   data = yaml.safe_load(fo)
... 
>>> print(data['script'][-1])
if [[ "${ORG}" == 'SciTools' && "${TRAVIS_EVENT_TYPE}" == 'push' && "${PUSH_BUILT_DOCS}" == 'true' ]]; then
  cd ${INSTALL_DIR};
  pip install doctr;
  doctr deploy --deploy-repo SciTools-docs/iris --built-docs docs/iris/build/html
    --key-path .github/deploy_key.scitools-docs.enc
    --no-require-master
    ${TRAVIS_BRANCH:-${TRAVIS_TAG}};
fi

>>> 
```

I don't really get this. The ">" form is supposed to "fold" lines + remove newlines, but here we definitely got embedded newlines.

pp-mo: > Have you tested that this line would now work?

No. It's hard to do unless it's running in the right repository.
In principle, merging to a feature branch might work (which is what Scitools/iris/pushdocs_test was for).  I may try that ...

pp-mo:
Hi @lbdreyer 
Restarted build, which crashed for some reason?

Tested on separate "feature" branch : it works at last !! :rocket: 
See ['pushdocs_trial' branch docs here](https://scitools-docs.github.io/iris/pushdocs_trial/index.html)

PR #3583, opened 2019-12-05
PI-3478: Common metadata API

bjlittle:
This PR introduces a common metadata API for cubes, coordinates and auxiliary factories, and also addresses a lot of connective tissue technical debt. This PR underpins the forthcoming changes for cube arithmetic.

Essentially, this PR proposes a `metadata` interface, which provides a unified, consistent treatment of metadata within `iris`:
- the `_as_defn` interface has been replaced with the `metadata` interface
- metadata state now lives with the metadata behaviour
- the `CFVariableMixin` methods have been rationalised
- there is a new `iris.common` module, for common, non-core `iris` behaviour:
    - the `_cube_coord_common` has been rationalised here, see `iris.common.mixin`
    - the `metadata` has been rationalised here, see `iris.common.metadata`
- the mixin `name` method behaviour is now appropriate and specific to the container
- the recently introduced mixin `names` property behaviour is now private and specific to an `iris.cube.Cube`
- For `iris.aux_factory.AuxCoordFactory`:
    - the `units` are no longer a read-only property, but part of the containers `metadata`
    - `climatological` is now a read-only property of `False`

Still to do:
- [x] Test coverage for new `iris.common.metadata.MetadataFactory`
- [x] Test coverage for new `iris.common.metadata._BaseMeta`
- [x] Test coverage for existing `iris.common.mixin.LimitedAttributeDict`

### DIFF
--- a/lib/iris/common/metadata.py
+++ b/lib/iris/common/metadata.py
@@ -61,7 +61,10 @@ class _BaseMeta(ABCMeta):
 
 
 class BaseMetadata(metaclass=_BaseMeta):
-    """Container for common metadata."""
+    """
+    Container for common metadata.
+
+    """
 
     _names = (
         "standard_name",
@@ -74,6 +77,7 @@ class BaseMetadata(metaclass=_BaseMeta):
     __slots__ = ()
 
     def name(self, default="unknown"):
+
         """
         Returns a human-readable name.
 
@@ -103,7 +107,10 @@ class BaseMetadata(metaclass=_BaseMeta):
 
 
 class CellMeasureMetadata(BaseMetadata):
-    """Metadata for a :class:`~iris.coords.CellMeasure`."""
+    """
+    Metadata container for a :class:`~iris.coords.CellMeasure`.
+
+    """
 
     _names = "measure"
 
@@ -111,7 +118,10 @@ class CellMeasureMetadata(BaseMetadata):
 
 
 class CoordMetadata(BaseMetadata):
-    """Metadata for a :class:`~iris.coords.Coord`."""
+    """
+    Metadata container for a :class:`~iris.coords.Coord`.
+
+    """
 
     _names = ("coord_system", "climatological")
 
@@ -119,7 +129,10 @@ class CoordMetadata(BaseMetadata):
 
 
 class CubeMetadata(BaseMetadata):
-    """Metadata for a :class:`~iris.cube.Cube`."""
+    """
+    Metadata container for a :class:`~iris.cube.Cube`.
+
+    """
 
     _names = "cell_methods"
 
--- a/lib/iris/common/metadata.py
+++ b/lib/iris/common/metadata.py
@@ -17,7 +17,7 @@ __all__ = [
     "CellMeasureMetadata",
     "CoordMetadata",
     "CubeMetadata",
-    "MetadataFactory",
+    "MetadataManagerFactory",
 ]
 
 
@@ -25,7 +25,7 @@ __all__ = [
 _TOKEN_PARSE = re.compile(r"""^[a-zA-Z0-9][\w\.\+\-@]*$""")
 
 
-class _BaseMeta(ABCMeta):
+class _NamedTupleMeta(ABCMeta):
     """
     Meta-class to support the convenience of creating a namedtuple from
     names/members of the metadata class hierarchy.
@@ -71,7 +71,7 @@ class _BaseMeta(ABCMeta):
         return super().__new__(mcs, name, bases, namespace)
 
 
-class BaseMetadata(metaclass=_BaseMeta):
+class BaseMetadata(metaclass=_NamedTupleMeta):
     """
     Container for common metadata.
 
@@ -154,7 +154,8 @@ class BaseMetadata(metaclass=_BaseMeta):
     def __lt__(self, other):
         #
         # Support Python2 behaviour for a "<" operation involving a
-        # "NoneType" operand.
+        # "NoneType" operand. Require to at least implement this comparison
+        # operator to support sorting of instances.
         #
         if not isinstance(other, self.__class__):
             return NotImplemented
@@ -242,7 +243,7 @@ class CubeMetadata(BaseMetadata):
     @property
     def _names(self):
         """
-        A tuple containing the value of each name participating in the identify
+        A tuple containing the value of each name participating in the identity
         of a :class:`iris.cube.Cube`. This includes the standard name,
         long name, NetCDF variable name, and the STASH from the attributes
         dictionary.
@@ -267,7 +268,7 @@ class CubeMetadata(BaseMetadata):
         return (standard_name, long_name, var_name, stash_name)
 
 
-def MetadataFactory(cls, **kwargs):
+def MetadataManagerFactory(cls, **kwargs):
     """
     A class instance factory function responsible for manufacturing
     metadata instances dynamically at runtime.
@@ -333,7 +334,7 @@ def MetadataFactory(cls, **kwargs):
         instance, and dump and load instance state successfully.
 
         """
-        return (MetadataFactory, (self.cls,), self.__getstate__())
+        return (MetadataManagerFactory, (self.cls,), self.__getstate__())
 
     def __repr__(self):
         args = ", ".join(
@@ -373,7 +374,7 @@ def MetadataFactory(cls, **kwargs):
             raise ValueError(emsg.format(cls.__name__, bad))
 
     # Define the name, (inheritance) bases and namespace of the dynamic class.
-    name = "Metadata"
+    name = "MetadataManager"
     bases = ()
     namespace = {
         "DEFAULT_NAME": cls.DEFAULT_NAME,
--- a/lib/iris/common/metadata.py
+++ b/lib/iris/common/metadata.py
@@ -1,0 +1,126 @@
+# Copyright Iris contributors
+#
+# This file is part of Iris and is released under the LGPL license.
+# See COPYING and COPYING.LESSER in the root of the repository for full
+# licensing details.
+
+from abc import ABCMeta
+from collections import namedtuple
+from collections.abc import Iterable
+
+
+__all__ = [
+    "BaseMetadata",
+    "CellMeasureMetadata",
+    "CoordMetadata",
+    "CubeMetadata",
+]
+
+
+class _BaseMeta(ABCMeta):
+    """
+    Meta-class to support the convenience of creating a namedtuple from
+    names/members of the metadata class hierarchy.
+
+    """
+
+    def __new__(mcs, name, bases, namespace):
+        if "_names" in namespace and not getattr(
+            namespace["_names"], "__isabstractmethod__", False
+        ):
+            namespace_names = namespace["_names"]
+            names = []
+            for base in bases:
+                if hasattr(base, "_names"):
+                    base_names = base._names
+                    is_abstract = getattr(
+                        base_names, "__isabstractmethod__", False
+                    )
+                    if not is_abstract:
+                        if (
+                            not isinstance(base_names, Iterable)
+                        ) or isinstance(base_names, str):
+                            base_names = (base_names,)
+                        names.extend(base_names)
+
+            if (not isinstance(namespace_names, Iterable)) or isinstance(
+                namespace_names, str
+            ):
+                namespace_names = (namespace_names,)
+
+            names.extend(namespace_names)
+
+            if names:
+                item = namedtuple(f"{name}Namedtuple", names)
+                bases = list(bases)
+                # Influence the appropriate MRO.
+                bases.insert(0, item)
+                bases = tuple(bases)
+
+        return super().__new__(mcs, name, bases, namespace)
+
+
+class BaseMetadata(metaclass=_BaseMeta):
+    """Container for common metadata."""
+
+    _names = (
+        "standard_name",
+        "long_name",
+        "var_name",
+        "units",
+        "attributes",
+    )
+
+    __slots__ = ()
+
+    def name(self, default="unknown"):
+        """
+        Returns a human-readable name.
+
+        First it tries self.standard_name, then it tries the 'long_name'
+        attribute, then the 'var_name' attribute, before falling back to
+        the value of `default` (which itself defaults to 'unknown').
+
+        """
+        return self.standard_name or self.long_name or self.var_name or default
+
+    def __lt__(self, other):
+        #
+        # Support Python2 behaviour for a "<" operation involving a
+        # "NoneType" operand.
+        #
+        if not isinstance(other, self.__class__):
+            return NotImplemented
+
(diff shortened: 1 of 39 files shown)


lib/iris/common/metadata.py
```diff
@@ -0,0 +1,402 @@
...
+# https://www.unidata.ucar.edu/software/netcdf/docs/netcdf_data_set_components.html#object_name
+_TOKEN_PARSE = re.compile(r"""^[a-zA-Z0-9][\w\.\+\-@]*$""")
+
+
+class _BaseMeta(ABCMeta):
+    """
+    Meta-class to support the convenience of creating a namedtuple from
+    names/members of the metadata class hierarchy.
+
+    """
```
bjlittle: Reference:
- https://docs.python.org/3/reference/datamodel.html?highlight=namespace#object.__new__
- https://docs.python.org/3/reference/datamodel.html?highlight=namespace#metaclasses
- https://docs.python.org/3/reference/datamodel.html?highlight=namespace#creating-the-class-object

lib/iris/common/metadata.py
```diff
@@ -0,0 +1,402 @@
...
+
+    _members = (
+        "standard_name",
+        "long_name",
+        "var_name",
+        "units",
+        "attributes",
+    )
+
+    __slots__ = ()
```
bjlittle: Reference:
- https://docs.python.org/3/reference/datamodel.html?highlight=__slots__#object.__slots__

lib/iris/common/metadata.py
```diff
@@ -0,0 +1,402 @@
...
+                raise AttributeError(emsg.format(self.__class__.__name__))
+
+        stash_name = self.attributes.get("STASH")
+        if stash_name is not None:
+            stash_name = str(stash_name)
+
+        return (standard_name, long_name, var_name, stash_name)
+
+
+def MetadataFactory(cls, **kwargs):
```
bjlittle: Note that, this is a **function**, not a **class**...

lib/iris/common/metadata.py
```diff
@@ -0,0 +1,402 @@
...
+
+    def __eq__(self, other):
+        if not hasattr(other, "cls"):
+            return NotImplemented
+        match = self.cls is other.cls
+        if match:
+            match = self.values == other.values
+        return match
+
+    def __getstate__(self):
```
bjlittle: Reference:
- https://docs.python.org/3/library/pickle.html?highlight=__getstate__#object.__getstate__

lib/iris/common/metadata.py
```diff
@@ -0,0 +1,402 @@
...
+        """Return the instance state to be pickled."""
+        return {field: getattr(self, field) for field in self.fields}
+
+    def __ne__(self, other):
+        match = self.__eq__(other)
+        if match is not NotImplemented:
+            match = not match
+        return match
+
+    def __reduce__(self):
```
bjlittle: Reference:
- https://docs.python.org/3/library/pickle.html?highlight=__getstate__#object.__reduce__

lib/iris/common/metadata.py
```diff
@@ -0,0 +1,402 @@
...
+    def __repr__(self):
+        args = ", ".join(
+            [
+                "{}={!r}".format(field, getattr(self, field))
+                for field in self.fields
+            ]
+        )
+        return "{}({})".format(self.__class__.__name__, args)
+
+    def __setstate__(self, state):
```
bjlittle: Reference:
- https://docs.python.org/3/library/pickle.html?highlight=__getstate__#object.__setstate__

lib/iris/common/metadata.py
```diff
@@ -0,0 +1,402 @@
...
+        "token": cls.token,
+        "values": values,
+    }
+
+    # Account for additional "CubeMetadata" specialised class behaviour.
+    if cls is CubeMetadata:
+        namespace["_names"] = cls._names
+
+    # Dynamically create the class.
+    Metadata = type(name, bases, namespace)
```
bjlittle: Reference:
- https://docs.python.org/3/library/functions.html?highlight=type#type

lib/iris/common/metadata.py
```diff
@@ -0,0 +1,402 @@
...
+        if match is not NotImplemented:
+            match = not match
+        return match
+
+    def __reduce__(self):
+        """
+        Dynamically created classes at runtime cannot be pickled, due to not
+        being defined at the top level of a module. As a result, we require to
+        use the __reduce__ interface to allow 'pickle' to recreate this class
+        instance, and dump and load instance state successfully.
```
bjlittle: Reference:
- https://docs.python.org/3/library/pickle.html#what-can-be-pickled-and-unpickled

lib/iris/common/metadata.py
```diff
@@ -0,0 +1,402 @@
...
+    # Check whether kwargs have valid fields for the specified metadata.
+    if kwargs:
+        extra = [field for field in kwargs.keys() if field not in cls._fields]
+        if extra:
+            bad = ", ".join(map(lambda field: "{!r}".format(field), extra))
+            emsg = "Invalid {!r} field parameters, got {}."
+            raise ValueError(emsg.format(cls.__name__, bad))
+
+    # Define the name, (inheritance) bases and namespace of the dynamic class.
+    name = "Metadata"
```
lbdreyer: Could we call this `MetadataManager` instead?

And perhaps call this function `MetadataManagerFactory` instead of `MetadataFactory`?

lib/iris/cube.py
```diff
@@ -775,6 +747,9 @@ def __init__(
         if isinstance(data, str):
             raise TypeError("Invalid data type: {!r}.".format(data))
 
+        # Configure the metadata manager.
+        self._metadata = MetadataFactory(CubeMetadata)
```
lbdreyer: Could we call these `self._metadata_manager`?

Currently we have:
`self.metadata `- returns an instance of BaseMetadata which is a souped up namedtuple
`self._metadata `- returns the instance that MetadataFactory returns.
I find that a bit confusing.

Another good thing about `self._metadata_manager` is follows a similar naming pattern to `self._data_manager`

lib/iris/coords.py
```diff
@@ -1502,8 +1454,13 @@ def climatological(self):
...
         On set, the input value is cast to a boolean, exceptions raised
         if units are not time units or if there are no bounds.
+
         """
-        return self._climatological if self.has_bounds() else False
+        if not self.has_bounds():
+            self._metadata.climatological = False
+        if not self.units.is_time_reference():
+            self._metadata.climatological = False
+        return self._metadata.climatological
```
lbdreyer: I have come around to the idea of this. 
We do a `is_time_reference` check in the setter, so we are being careful with what is being set as having climatological bounds.
There is a risk is if someone modifies an already existing climatological coordinate to  something like a latitude coordinate, without updating the climatological property. In these cases, I would prefer we ensure we save things out correctly (i.e. pick up that this coordinate shouldn't be climatological). It would only be a problem if they were to access coord._metadata.climatological directly, but they shouldn't be doing that
.
So this just needs some testing

lib/iris/common/metadata.py
```diff
@@ -0,0 +1,402 @@
...
+            emsg = "Cannot retrieve a valid name token from {!r}"
+            raise ValueError(emsg.format(self))
+
+        return result
+
+    def __lt__(self, other):
+        #
+        # Support Python2 behaviour for a "<" operation involving a
+        # "NoneType" operand.
+        #
```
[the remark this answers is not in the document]
bjlittle: @lbdreyer The reason is the behaviour of Python has changed between 2.x and 3.x, and we want to support the old behaviour.
For example, in Python2 we have:
```python
Python 2.7.15 |Anaconda, Inc.| (default, Dec 14 2018, 19:04:19) 
[GCC 7.3.0] on linux2
Type "help", "copyright", "credits" or "license" for more information.
>>> from collections import namedtuple
>>> N = namedtuple("N", ["a", "b"])
>>> a = N(1, 2)
>>> b = N(1, None)
>>> a < b
False
>>> b < a
True
```
However, in Python3 we have:
```python
Python 3.6.8 |Anaconda, Inc.| (default, Dec 30 2018, 01:22:34) 
[GCC 7.3.0] on linux
Type "help", "copyright", "credits" or "license" for more information.
>>> from collections import namedtuple
>>> N = namedtuple("N", ["a", "b"])
>>> a = N(1, 2)
>>> b = N(1, None)
>>> a < b
TypeError: '<' not supported between instances of 'int' and 'NoneType'
```
The `BaseMetadata.__lt__` resolves this issue:
```python
>>> class M(N):       
...     def __lt__(self, other):
...         def _sort_key(item):
...             keys = []
...             for field in item._fields:
...                 value = getattr(item, field)
...                 keys.extend((value is not None, value))
...             return tuple(keys)
...         return _sort_key(self) < _sort_key(other)
... 
>>> a = M(1, 2)
>>> b = M(1, None)
>>> a < b
False
>>> b < a
True
```

lib/iris/common/metadata.py
```diff
@@ -0,0 +1,402 @@
...
+        if token and result is None:
+            emsg = "Cannot retrieve a valid name token from {!r}"
+            raise ValueError(emsg.format(self))
+
+        return result
+
+    @property
+    def _names(self):
+        """
+        A tuple containing the value of each name participating in the identify
```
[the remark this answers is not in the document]
bjlittle: Yup, good spot :eyes:

lib/iris/tests/unit/common/mixin/test_get_valid_standard_name.py
```diff
@@ -4,15 +4,15 @@
 # See COPYING and COPYING.LESSER in the root of the repository for full
 # licensing details.
 """
-Unit tests for the :func:`iris._cube_coord_common.get_valid_standard_name`.
+Unit tests for the :func:`iris.common.mixin.get_valid_standard_name`.
```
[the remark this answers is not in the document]
bjlittle: @lbdreyer Yup. Originally it wasn't a private function. I ported it across into `iris.common.mixin` wrote the tests, then later changed it to be a private function. Forgot to also change the associated comment and test file name. Another good spot!

lib/iris/cube.py
```diff
@@ -841,43 +816,15 @@ def __init__(
...
+    def _names(self):
         """
-        An instance of :class:`CubeMetadata` describing the phenomenon.
-
-        This property can be updated with any of:
-         - another :class:`CubeMetadata` instance,
-         - a tuple/dict which can be used to make a :class:`CubeMetadata`,
-         - or any object providing the attributes exposed by
-           :class:`CubeMetadata`.
+        A tuple containing the value of each name participating in the identify
```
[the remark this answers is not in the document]
bjlittle: @lbdreyer Cut'n'paste carnage. Guilty as charged :wink:

lib/iris/coords.py
```diff
@@ -2674,19 +2619,20 @@ def __init__(self, method, coords=None, intervals=None, comments=None):
                 "'method' must be a string - got a '%s'" % type(method)
             )
 
-        default_name = CFVariableMixin._DEFAULT_NAME
+        default_name = BaseMetadata.DEFAULT_NAME
```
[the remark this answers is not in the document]
bjlittle: @lbdreyer It should be public, as the `iris.coords.CellMethod` requires access to it. 

If it was private, then the side-effect is that as a developer I might want to change the value of `BaseMetadata.DEFAULT_NAME` thinking that it's totally safe to do so, and yet it would directly impact the behaviour of `iris.coords.CellMethod`.

If `DEFAULT_NAME` was only used internally within `BaseMetadata` then I'd suggest that it's safe to make it private - but that's not the case. Hence the change.

lib/iris/common/mixin.py
```diff
@@ -244,40 +144,99 @@ def rename(self, name):
...
     def attributes(self, attributes):
-        self._attributes = LimitedAttributeDict(attributes or {})
+        self._metadata.attributes = LimitedAttributeDict(attributes or {})
+
+    @property
+    def metadata(self):
+        return self._metadata.values
+
+    @metadata.setter
+    def metadata(self, metadata):
```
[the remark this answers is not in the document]
bjlittle: @lbdreyer Nope, not that I know of. The doc-string will be taken from the property `getter` of `CFVariableMixin.metadata` and not the `setter` :cry:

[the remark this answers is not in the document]
bjlittle: As it stands, this doc-string behaviour is the same as `master` as it stands...

lib/iris/common/metadata.py
```diff
@@ -0,0 +1,402 @@
...
+    "CubeMetadata",
+    "MetadataFactory",
+]
+
+
+# https://www.unidata.ucar.edu/software/netcdf/docs/netcdf_data_set_components.html#object_name
+_TOKEN_PARSE = re.compile(r"""^[a-zA-Z0-9][\w\.\+\-@]*$""")
+
+
+class _BaseMeta(ABCMeta):
```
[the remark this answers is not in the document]
bjlittle: @lbdreyer Yup, on reflection I do agree. It is confusing :confused: 

After a bit of thought (and mutual agreement) we're going with `_NamedTupleMeta` instead.

lib/iris/common/mixin.py
```diff
@@ -5,43 +5,20 @@
 # licensing details.
```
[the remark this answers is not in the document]
bjlittle: @lbdreyer My intent here was simple, as in, it's a namespace for common mixin classes to live.

It might just be the case that `CFVariableMixin` will be our only mixin class, but at least we have scope for collecting such classes together in one place, rather than having them sprinkled around the codebase. HTH :wink:

lib/iris/common/metadata.py
bjlittle: @lbdreyer Agreed. Let's do both for clarity.

lib/iris/coords.py
bjlittle: @lbdreyer Thanks for taking the time to think about this one, much appreciated. 

It was a bit cheeky of me to include this in this PR, but since I was in the space, I thought I'd try to address something that reeked of a potential issue for users.

I'll add some tests :smile:

lib/iris/tests/unit/common/metadata/test_BaseMetadata.py
```diff
@@ -0,0 +1,208 @@
...
+            "standard_name",
+            "long_name",
+            "var_name",
+            "units",
+            "attributes",
+        )
+        self.assertEqual(BaseMetadata._fields, expected)
+
+
+class Test___le__(tests.IrisTest):
```
lbdreyer:
```suggestion
class Test___lt__(tests.IrisTest):
```

bjlittle:
@lbdreyer and @alastair-gemmell Awesome, thanks! :beers: :tada:

PR #3587, opened 2019-12-09
_regrid_area_weighted_array: Set order y_dim, x_dim axis to be last dimensions

abooton:
Re-order the axis to be  [..., y_dim, x_dim].  The weights.T then becomes redundant.
Changes are in preparation for optimising the averaging.


lib/iris/experimental/regrid.py
```diff
@@ -491,6 +491,31 @@ def _regrid_area_weighted_array(
...
+        # test cross_section along line longitude
+        src_data = np.moveaxis(src_data, x_dim, -1)
+        x_dim = src_data.ndim - 1
+    elif y_dim is not None and x_dim is None:
+        # test cross_section along line latitude
+        src_data = np.moveaxis(src_data, y_dim, -1)
+        y_dim = src_data.ndim - 1
+    elif x_dim < y_dim:
+        src_data = np.moveaxis(src_data, x_dim, -1)
+        src_data = np.moveaxis(src_data, y_dim - 1, -2)
```
ehogan: Shouldn't `y_dim - 1` be just `y_dim`? In fact, can the last two loops be combined? Regardless of whether `x_dim` is greater or less than `y_dim`, the result is the same?

lib/iris/experimental/regrid.py
```diff
@@ -608,6 +631,20 @@ def _regrid_area_weighted_array(
...
+    elif y_dim_orig is not None and x_dim_orig is None:
+        new_data = np.moveaxis(new_data, -1, y_dim_orig)
+    elif x_dim_orig < y_dim_orig:
+        new_data = np.moveaxis(new_data, -1, x_dim_orig)
+        new_data = np.moveaxis(new_data, -1, y_dim_orig)
+    else:
+        new_data = np.moveaxis(new_data, -2, y_dim_orig)
+        new_data = np.moveaxis(new_data, -1, x_dim_orig)
+
     return new_data
```
ehogan: Would it be safer to use `x_dim` and `y_dim` rather than `-1` and `-2` here?

lib/iris/experimental/regrid.py
abooton: Yes, as y gets shifted along one position as x moves to the end.

lib/iris/experimental/regrid.py
```diff
@@ -491,6 +491,31 @@ def _regrid_area_weighted_array(
...
+        src_data = np.moveaxis(src_data, y_dim - 1, -2)
+        x_dim = src_data.ndim - 1
+        y_dim = src_data.ndim - 2
+    else:
+        src_data = np.moveaxis(src_data, x_dim, -1)
+        src_data = np.moveaxis(src_data, y_dim, -2)
+        x_dim = src_data.ndim - 1
+        y_dim = src_data.ndim - 2
+
     # Create empty data array to match the new grid.
```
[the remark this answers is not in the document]
abooton: Possibly.  If we add a test for data ordered:  [x, z, y], and like [y, x, z] I think this captures most of it.  In the next PR we ensure we have x and y dims to might want to re-evaluate this next.

lib/iris/experimental/regrid.py
```diff
@@ -577,8 +602,6 @@ def _regrid_area_weighted_array(
                 # Transpose weights to match dim ordering in data.
```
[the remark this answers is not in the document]
abooton: Yes - we'll make sure its out in 2 PR's time.

lib/iris/experimental/regrid.py
abooton: Could do.  They do the same thing so if it is working shall we leave until the next PRs go up?

trexfeathers:
Closing as superseded by #3594

PR #3589, opened 2019-12-09
_regrid_area_weighted_array: Move averaging out of loop

abooton:
Optimise the averaging in the _regrid_area_weighted_array.

This is achieved by refactoring the weights calculation, collating the data and the weights in a single array, then finally computing the weighted mean as a whole-array-manipulation.  

Calculating the weights first is necessary because the size (i.e. number of elements) of the weights and source data region varies per target grid point.  In order to record all the values in a single array, the maximum size of the weights array needs to be established, so the single array can be setup.  Therefore we first computing the weights and record the maximum weight-array size, hence we have refactored the weights calculation too.


abooton:
Closed superseded by #3596

PR #3590, opened 2019-12-09
PI-3473 Rename "engine.provides"

pp-mo:
During work on #3556, it became clear to me that the use of `engine.provides` within the pyke rules, and in a few places within `iris.fileformats.netcdf` is confusing :  Because this term collides with the special reserved meaning of "provides" within Pyke rules, as seen in the repeated usage of `facts_cf.provides(..)` in most of our Pyke rules.

In fact, although `engine` is a Pyke object,  `engine.provides` is just a monkey-patched additional property attached by our netcdf code, to keep track of cube component objects created during rules evaluation (coordinates, cell-measures and ancillaries).

So, this PR just renames it to something less confusing. 
This should produce **no functional change.**


pp-mo:
> pp-mo changed the base branch from master to launch_ancils 

Note: this makes sense due to already-mentioned dependence on #3556, which is on the feature branch

pp-mo:
Thanks @stephenworsley 
Potentially confusing name re-worded.  Hope this is now ok.

PR #3591, opened 2019-12-09
Update CF standard name table to v70

lbdreyer:
I noticed that our standard_name table is a little out of date.

In Iris we are currently on version v59 which was released on 25 September 2018.
~~The latest version is v69 which was released on 15 October 2019.~~
The latest version is v70 which was released on 10 December 2019.

I recommend we update to the latest version, ideally before the Iris 3 release??

Mostly this change includes updates to descriptions the standard names but there are few edits to the actual standard_names themselves, for example:
* `covariance_over_longitude_of_northward_wind_and_air_temperature` was added
* `charnock_coefficient_for_surface_roughness_length_for_momentum_in_air` was added
*`carbon_mass_flux_into_soil_and_litter_due_to_anthropogenic_land_use_or_land_cover_change` to `carbon_mass_flux_into_litter_and_soil_due_to_anthropogenic_land_use_or_land_cover_change`
* `ambient_aerosol_particle_diameter` has been removed
and a few other changes


lbdreyer:
Alternatively! We could target this at Iris 2.3?

pp-mo:
> Alternatively! We could target this at Iris 2.3?

Seems reasonable, especially if we consider Iris 2.3 the "last of the old line", like Python 2.7.
I think few if anyone will be upset by this.

lbdreyer:
This is now pointing at the v2.3.x branch

lbdreyer:
Thanks @pp-mo !

PR #3594, opened 2019-12-12
PI-2472: Tweak area weighting regrid move xdim and ydim axes

abooton:
Re-order the axis to be [..., y_dim, x_dim]. The weights.T then becomes redundant.
Changes are in preparation for optimising the averaging.

See original PR #3587 (targeting master) for initial comments


lib/iris/experimental/regrid.py
```diff
@@ -608,6 +631,20 @@ def _regrid_area_weighted_array(
...
+        new_data = np.moveaxis(new_data, -1, x_dim_orig)
+    elif y_dim_orig is not None and x_dim_orig is None:
+        new_data = np.moveaxis(new_data, -1, y_dim_orig)
+    elif x_dim_orig < y_dim_orig:
+        new_data = np.moveaxis(new_data, -1, x_dim_orig)
+        new_data = np.moveaxis(new_data, -1, y_dim_orig)
+    else:
+        new_data = np.moveaxis(new_data, -2, y_dim_orig)
+        new_data = np.moveaxis(new_data, -1, x_dim_orig)
+
```
ehogan: These last two loops can be consolidated, since `new_data` will end in `[..., y, x]` in either case.

lib/iris/tests/experimental/regrid/test_regrid_area_weighted_rectilinear_src_and_grid.py
```diff
@@ -322,19 +322,36 @@ def test_regrid_latlon_reduced_res(self):
...
+        # the mean and std statistics should be the same
+        data = np.moveaxis(src.data.copy(), 2, 0)
+        src = iris.cube.Cube(data)
+        src.add_dim_coord(z, 2)
+        src.add_dim_coord(lon, 1)
+        src.add_dim_coord(lat, 0)
+        dest = _resampled_grid(self.realistic_cube[0, 0, :3, :2], 3, 3)
+        res = regrid_area_weighted(src, dest)
+        self.assertArrayShapeStats(src, (3, 2, 4), 288.08868, 0.008262919)
+        self.assertArrayShapeStats(res, (9, 6, 4), 288.08865, 0.00826281)
```
ehogan: Should these different tests be in separate test functions?

lib/iris/experimental/regrid.py
```diff
@@ -491,6 +491,31 @@ def _regrid_area_weighted_array(
         cached_x_bounds.append(x_bounds)
         cached_x_indices.append(x_indices)
 
+    # Move y_dim and x_dim to last dimensions
+    x_dim_orig = copy.copy(x_dim)
+    y_dim_orig = copy.copy(y_dim)
```
ehogan: Stephen queried the use of `copy.copy` and I agree it isn't required here. I'll add this comment to #3596.

lib/iris/tests/experimental/regrid/test_regrid_area_weighted_rectilinear_src_and_grid.py
abooton: Usually yes, but in this case they align with the transpose function that this supersedes and also share the same results info.

lib/iris/experimental/regrid.py
abooton: Ok.  We'll revisit in PR  #3596 after everything has been rejigged anyway

PR #3599, opened 2019-12-15
Update iris colour logo

abooton:


pp-mo:
I note that the new pics still don't have the natural "centre" of the design (middle of the petals) at the exact centre of the image.

Adding halfway lines to the 100-size gets you this ...
Iris7_1_trim_100_withquarters


But it is now *close* + I think looks probably ok. 
Was this as intended ?

abooton:
> I note that the new pics still don't have the natural "centre" of the design (middle of the petals) at the exact centre of the image.
> 
> Adding halfway lines to the 100-size gets you this ...
> Iris7_1_trim_100_withquarters
> 
> But it is now _close_ + I think looks probably ok.
> Was this as intended ?

I cropped the canvas to the edge of the picture.  Maybe I should have cropped it so that it is to the edge of the maximum radius of the petals ?

pp-mo:
> Maybe I should have cropped it so that it is to the edge of the maximum radius of the petals ?

That sounds right.
Is it easy to do ??

abooton:
The current logo files are all cropped at the edge of the image, and the backgrounds are all transparent.
The size of the current logos has been replicated wrt the height of the images, but the new width of the logo is slightly different, hence the image sizes will differ.
I've added a "iris_colour_logo_centred.png" image.  This has the centre of the iris in the centre of the image, so can be used for twitter.

PR #3600, opened 2019-12-16
(WIP) Extend new_axis() to help concatenate cubes with ancils

stephenworsley:
Depends on either #3566 or #3583 changing the metadata structure of CellMeasure and AncillaryVariable.

The purpose of this pull request is to provide a simple workaround for cases where cubes can't merge properly because there are AuxCoords, CellMeasures or AncillaryVariables which ought to be merged in the final cube. In such a case, the function `iris.util.new_axis` is applied to all cubes and they are then concatenated. All AuxCoords, CellMeasures or AncillaryVariables which are to be merged in the final will be promoted by `iris.util.new_axis` by specifying them with the keywords `promoted_aux`, `promoted_cm` or `promoted_av`. This allows a greater degree of control than is possible with merge, since it will be possible to decide which AuxCoords, CellMeasures or AncillaryVariables should be merged (increasing their dimension) and which should be conserved (preserving their dimension).

I think this extra functionality could help simplify a lot of potentially fiddly or unstable user code (`iris.util.new_axis` accounts for a lot of nasty edge cases). The workaround for merge would only have to be a couple extra lines, eg.

```python
mcube = cubes.merge_cube()
```
vs
```python
cubes = CubeList(
    [new_axis(cube, scalar_coord="y", promoted_av=["ancil"]) for cube in cubes]
)
ccube = cubes.concatenate_cube()
```

In addition, Ancillary Variables and Cell Measures are no longer thrown away by `new_axis`.

### DIFF
--- a/lib/iris/cube.py
+++ b/lib/iris/cube.py
@@ -1399,6 +1399,8 @@ bound=(1994-12-01 00:00:00, 1998-12-01 00:00:00)
             The CellMeasure to look for.
 
         """
+        cell_measure = self.cell_measure(cell_measure)
+
         # Search for existing cell measure (object) on the cube, faster lookup
         # than equality - makes no functional difference.
         matches = [
@@ -1421,6 +1423,8 @@ bound=(1994-12-01 00:00:00, 1998-12-01 00:00:00)
             The AncillaryVariable to look for.
 
         """
+        ancillary_variable = self.ancillary_variable(ancillary_variable)
+
         # Search for existing ancillary variable (object) on the cube, faster
         # lookup than equality - makes no functional difference.
         matches = [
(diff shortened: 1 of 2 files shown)


pp-mo:
:+1: for the principle.

I really want to bank the elements of #3566 before considering this, so no review yet.

How will this play out it we wait for #3566 or #3583 
 - will it go easier before or after those ?

stephenworsley:
@pp-mo It will definitely go easier after #3566 or #3583. Currently this code won't work unless the metadata in iris.coords is edited, and it's probably not a great idea do do that in three parallel pull requests.
I would also like to put in an integration test with new_axis and concatenation working together when I get around to writing tests. I agree that this makes sense to be completed and reviewed after #3566.

stephenworsley:
Replaced by #4896.

PR #3604, opened 2019-12-17
Test for AreaWeightedRegridder

ehogan:
This pull requests adds some tests for [AreaWeightedRegridder](https://github.com/ehogan/iris/blob/area_weighted_regridder_test/lib/iris/analysis/_area_weighted.py#L14).

The `test_multiple_src_on_same_grid` test creates two source cubes on the same grid with different values. One of the sources cubes is used to instantiate `AreaWeightedRegridder`, then both source cubes are regridded. This test will ensure the regridding still works as expected once the refactoring (including caching the weights) is complete.

### DIFF
--- a/lib/iris/tests/unit/analysis/area_weighted/test_AreaWeightedRegridder.py
+++ b/lib/iris/tests/unit/analysis/area_weighted/test_AreaWeightedRegridder.py
@@ -128,7 +128,7 @@ class Test(tests.IrisTest):
             src1.coord(name).guess_bounds()
             src2.coord(name).guess_bounds()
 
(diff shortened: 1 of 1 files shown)


stephenworsley:
These tollerances seem quite large. Does the current behaviour of AreaWeightedRegridder require similar tolerances for this data?

ehogan:
The [tests for regrid_area_weighted_rectilinear_src_and_grid](https://github.com/ehogan/iris/blob/area_weighted_regridder_test/lib/iris/tests/experimental/regrid/test_regrid_area_weighted_rectilinear_src_and_grid.py#L246) removed the coord system and units so it is no longer spherical. When I did that in this branch the arrays were equal (see latest commit) :)

PR #3606, opened 2019-12-17
PI-2472: Refactor AreaWeightedRegridder

ehogan:
This pull request refactors `AreaWeightedRegridder` to call the relevant `__prepare` and `__perform` functions.

I've not come across `mock.sentinel` before, so I've had to comment out one assert that I can't get to work.

### DIFF
--- a/lib/iris/analysis/_area_weighted.py
+++ b/lib/iris/analysis/_area_weighted.py
@@ -51,8 +51,6 @@ class AreaWeightedRegridder:
         # impervious to external changes to the original source cubes.
         self._src_grid = snapshot_grid(src_grid_cube)
         self._target_grid = snapshot_grid(target_grid_cube)
-        self._src_grid_cube = src_grid_cube.copy()
-        self._target_grid_cube = target_grid_cube.copy()
         # Missing data tolerance.
         if not (0 <= mdtol <= 1):
             msg = "Value for mdtol must be in range 0 - 1, got {}."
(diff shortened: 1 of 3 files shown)


ehogan:
I managed to reproduce the doctest failure by updating my local conda install to match the environment that Travis uses. I'll continue debugging tomorrow (unless someone else gets to it first!) ;)

ehogan:
The doctest was failing due to the fact that the source cube is now being cached (a new cube is created) in [AreaWeightedRegridder](https://github.com/ehogan/iris/tree/refactor_area_weighted_regridder/lib/iris/analysis/_area_weighted.py#L68). This means that when [RectilinearRegridder._create_cube](https://github.com/ehogan/iris/tree/refactor_area_weighted_regridder/lib/iris/experimental/regrid.py#L970) is called, [copy_coords](https://github.com/ehogan/iris/tree/refactor_area_weighted_regridder/lib/iris/analysis/_regrid.py#L881), didn't correctly copy the `latitude` and `longitude` coordinates (using `is` checks whether both variables refer to the same object, which is no longer true), which resulted in         `VOLCANIC_ASH_AIR_CONCENTRATION / (g/m3) (-- : 73; -- : 96)` rather than `VOLCANIC_ASH_AIR_CONCENTRATION / (g/m3) (latitude: 73; longitude: 96)` (thanks to @stephenworsley for finding the root cause!).

ehogan:
`__prepare` is now called when the `AreaWeightedRegridder` is instantiated. This allows the use of `src_grid_cube` directly, without caching it, which enables the change I made that was detailed [here](https://github.com/SciTools/iris/pull/3606#issuecomment-567450435) to be removed. It's possible that `target_grid_cube` also doesn't need to be cached, since it is only now used in `__prepare` (which is only called once, when `AreaWeightedRegridder` is instantiated). However, I'm not sure if there would be any significant performance issues related to the fact that the cached `target_grid_cube` has no data associated with it.

lib/iris/analysis/_area_weighted.py
```diff
@@ -97,6 +101,6 @@ def __call__(self, cube):
                 "The given cube is not defined on the same "
                 "source grid as this regridder."
             )
-        return eregrid.regrid_area_weighted_rectilinear_src_and_grid(
-            cube, self._target_grid_cube, mdtol=self._mdtol
+        return eregrid._regrid_area_weighted_rectilinear_src_and_grid__perform(
+            cube, self._regrid_info, mdtol=self._mdtol
```
abooton: We need to have checks in here to ensure that the incoming src_cube coordinate info, matches that of the original src_cube used to generate the weights.  This latter information will now be in the _regrid_info variables.

ehogan: Done: [27a4add](https://github.com/ehogan/iris/commit/27a4addcdde174fa74285bccb957843847cb5982), with test: [ec9045f](https://github.com/SciTools/iris/pull/3606/commits/ec9045f2c131a4f2248c2c8f0dbd5610faa86976)

abooton:
Looks good.  :)
Now, most of the checks on the src and target cube are done at the instantiation of the regridder.  We then check that the src cube x&y dims match those that the checks were done on.  At a later date we might be able to revisit the restrictions and reduce them a little.  However this is not easy to determine until the weights refactor is in.

Issue #3474, opened 2019-10-21
Ensure flags are correctly loaded/saved to/from NetCDF


lbdreyer:
Part of #3358

According to CF, flags contain `flag_values`, `flag_masks` and `flag_meanings` attributes.

This ticket is to check whether flags require any special handling to ensure that they are correctly loaded/saved.

---

Acceptance Criteria
- [ ] A new Iris test that demonstrates flags are loaded/saved correctly

alastair-gemmell:
This ticket is best tackled after #3473

stephenworsley:
The convention with status_flags is to be unitless. An example of a status flag being described within CF conventions is given here: https://github.com/cf-convention/cf-conventions/issues/216 
```
variables:

    float sea_water_practical_salinity(time, z);
        sea_water_practical_salinity:units = "1";
        sea_water_practical_salinity:long_name = "Salinity";
        sea_water_practical_salinity:standard_name = "sea_water_practical_salinity";
        sea_water_practical_salinity:ancillary_variables = "sea_water_practical_salinity_qc_agg sea_water_practical_salinity_qc_flat_line_test";

    int sea_water_practical_salinity_qc_agg(time, z);
        sea_water_practical_salinity_qc_agg:long_name = "Salinity QARTOD Aggregate Flag";
        sea_water_practical_salinity_qc_agg:standard_name = "status_flag qartod_aggregate_quality_flag";
        sea_water_practical_salinity_qc_agg:missing_value = 2;
        sea_water_practical_salinity_qc_agg:flag_meanings = "PASS NOT_EVALUATED SUSPECT FAIL MISSING";
        sea_water_practical_salinity_qc_agg:flag_values = 1, 2, 3, 4, 9;

    int sea_water_practical_salinity_qc_flat_line_test(time, z);
        sea_water_practical_salinity_qc_flat_line_test:long_name = "Salinity QARTOD Flat Line Test Flag";
        sea_water_practical_salinity_qc_flat_line_test:standard_name = "status_flag qartod_flat_line_test_quality_flag";
        sea_water_practical_salinity_qc_flat_line_test:missing_value = 2;
        sea_water_practical_salinity_qc_flat_line_test:flag_meanings = "PASS NOT_EVALUATED SUSPECT FAIL MISSING";
        sea_water_practical_salinity_qc_flat_line_test:flag_values = 1, 2, 3, 4, 9;
```

[...]

Issue #3394, opened 2019-09-09
Aligning save behaviour of cube with units "unknown" and "no-unit"


bjlittle:
Consider the following treatment of the cube `units` for the value `unknown` and `no-unit`.

First, create come appropriate cubes, and save them (using the latest upstream/master pre 2.3.0):
```python
>>> import iris
>>> nounit = iris.cube.Cube(1, 'air_temp', units='no unit')
>>> nounit = iris.cube.Cube(1, 'air_temperature', units='no unit')
>>> iris.save(nounit, 'nounit.nc')
>>> unknown = iris.cube.Cube(1, 'air_temperature', units='unknown')
>>> iris.save(unknown, 'unknown.nc')
```
Now, let's look at the resultant NetCDF files:
```shell
$ ncdump -h nounit.nc
netcdf nounit {
variables:
	int64 air_temperature ;
		air_temperature:standard_name = "air_temperature" ;
		air_temperature:units = "no_unit" ;

// global attributes:
		:Conventions = "CF-1.5" ;
}

$ ncdump -h unknown.nc 
netcdf unknown {
variables:
	int64 air_temperature ;
		air_temperature:standard_name = "air_temperature" ;

// global attributes:
		:Conventions = "CF-1.5" ;
```
We created the `no_unit` unit (the aliases in `cf-units` for `no_unit` are `-`, `no unit`, `no-unit`, and `nounit`) for convenience, and doesn't actually map to an valid UDUNITS unit. This is also true for the `unknown` unit. 

The difference here is that `iris` doesn't explicitly save to NetCDF any units of `unknown`, and I argue this should also be that case for the `no_unit` unit e.g.,
```shell
$ cfchecker nounit.nc
...
------------------
Checking variable: air_temperature
------------------
ERROR: (3.1): Invalid units: no_unit

ERRORS detected: 1
WARNINGS given: 1
INFORMATION messages: 0
```
Both the `unknown` and `no_unit` units should have the *same* behaviour and should not be saved to NetCDF.

Plus, NetCDF with `units` containing `no_unit` may break upstream tools, as it's clearly not CF compliant.

bjlittle:
Ping @lbdreyer

stephenworsley:
For reference, it's probably just this line which needs changing in order to make this change: https://github.com/SciTools/iris/blob/26a32a3a76a9ae259d56626b59f93e3fa9ecf840/lib/iris/fileformats/netcdf.py#L1763

[...]

PR #3613, opened 2019-12-19
PI:3358 (WIP) Ensure flags load/save without units

stephenworsley:
Addresses #3474. Anything loaded from netCDF with a flag attribute (some attribute wich is either `flag_values`, `flag_masks` or `flag_meanings`) is explicitly given a unit of `"no_unit"` in iris. Anything in iris with a unit of `"no_unit"` is now saved to netCDF without units. This also addresses #3394.

### DIFF
--- a/lib/iris/fileformats/_pyke_rules/fc_rules_cf.krb
+++ b/lib/iris/fileformats/_pyke_rules/fc_rules_cf.krb
@@ -1689,7 +1689,7 @@ fc_extras
         if np.issubdtype(cf_var.dtype, np.str_):
             attr_units = cf_units._NO_UNIT_STRING
 
-        if any(name in cf_var.cf_data for name in ("flag_values", "flag_masks", "flag_meanings")):
+        if any(hasattr(cf_var.cf_data, name) for name in ("flag_values", "flag_masks", "flag_meanings")):
             attr_units = cf_units._NO_UNIT_STRING
 
         # Get any assoicated calendar for a time reference coordinate.
--- a/lib/iris/fileformats/_pyke_rules/fc_rules_cf.krb
+++ b/lib/iris/fileformats/_pyke_rules/fc_rules_cf.krb
@@ -1689,8 +1689,7 @@ fc_extras
         if np.issubdtype(cf_var.dtype, np.str_):
             attr_units = cf_units._NO_UNIT_STRING
 
-        if "flag_values" in cf_var._nc_attrs or "flag_masks" in cf_var._nc_attrs \
-                or "flag_meanings" in cf_var._nc_attrs:
+        if any(hasattr(cf_var.cf_data, name) for name in ("flag_values", "flag_masks", "flag_meanings")):
             attr_units = cf_units._NO_UNIT_STRING
 
         # Get any assoicated calendar for a time reference coordinate.
--- a/lib/iris/fileformats/_pyke_rules/fc_rules_cf.krb
+++ b/lib/iris/fileformats/_pyke_rules/fc_rules_cf.krb
@@ -1689,6 +1689,10 @@ fc_extras
         if np.issubdtype(cf_var.dtype, np.str_):
             attr_units = cf_units._NO_UNIT_STRING
 
+        if "flag_values" in cf_var._nc_attrs or "flag_masks" in cf_var._nc_attrs \
+                or "flag_meanings" in cf_var._nc_attrs:
+            attr_units = cf_units._NO_UNIT_STRING
+
         # Get any assoicated calendar for a time reference coordinate.
         if cf_units.as_unit(attr_units).is_time_reference():
             attr_calendar = getattr(cf_var, CF_ATTR_CALENDAR, None)
--- a/lib/iris/fileformats/_pyke_rules/fc_rules_cf.krb
+++ b/lib/iris/fileformats/_pyke_rules/fc_rules_cf.krb
@@ -1689,7 +1689,7 @@ fc_extras
         if np.issubdtype(cf_var.dtype, np.str_):
             attr_units = cf_units._NO_UNIT_STRING
 
-        if any(hasattr(cf_var.cf_data, name) for name in ("flag_values", "flag_masks", "flag_meanings")):
+        if any(name in cf_var.cf_data for name in ("flag_values", "flag_masks", "flag_meanings")):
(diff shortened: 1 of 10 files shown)


lib/iris/fileformats/_pyke_rules/fc_rules_cf.krb
```diff
@@ -1689,6 +1689,10 @@ fc_extras
         if np.issubdtype(cf_var.dtype, np.str_):
             attr_units = cf_units._NO_UNIT_STRING
 
+        if "flag_values" in cf_var._nc_attrs or "flag_masks" in cf_var._nc_attrs \
```
pp-mo, line 1692: I don't think we should be using a private property of CFVariable here if we can avoid it.
I've just spent much too long [trying to understand exactly what CFVariable does + how](https://github.com/SciTools/iris/pull/3725), 
and I would say, please let's not reply on that implementation here !

I think I would use `any(hasattr(cf_var, name) for name in (...))`.

It's true, this will be less efficient because [hasattr depends entirely on getattr](https://stackoverflow.com/a/24971134)
On the other hand, it will only fetch them once per variable, as the CFVariable implementation does ensure that.
Given that netcdf files don't have the proliferation of small things that field-based formats do, I think it will be ok for performance -- although we could test that ?

pp-mo:
After long, long thought comparing this to #3712, I think **I do prefer this approach**.

The real cause of the problems here lies in the confused division-of-labour between `netcdf.py`, `cf.py` and `_pyke_rules/fc_rules_cf.krb`.  
All of those carry _some_ of the encoding of CF concepts, unfortunately (though netcdf.py much less so).

I don't think we can seek to fix that here + now, but must live with it.
So, I think this approach is the more natural one, within the existing code context.

I also think that the whole business of [adjusting attributes during the cube-load in netcdf.py](https://github.com/SciTools/iris/blob/912f500c5a5683c74a7b57de5d8010c216cba89c/lib/iris/fileformats/netcdf.py#L601)  is probably the nastiest bit of netcdf.py, and the worst of the division-of-labour failure.
So I fear that #3712 would only perpetuate that awkwardness !

lib/iris/fileformats/_pyke_rules/fc_rules_cf.krb
stephenworsley: So it looks like there is some unusual behaviour associated with getting atrributes from `cf_var` where accessing an attribute will cause it to not be picked up as an attribute in the final object. I think this is due to this piece of code https://github.com/SciTools/iris/blob/715d01921f01903132538ea2574653155fa4412d/lib/iris/fileformats/cf.py#L175-L183

It seems like in the current implementation, "attributes" contains everything which has not already been accessed in this way. Trying to access these attributes prematurely causes them to be ignored later on.
I believe this is why I chose to take attributes from `cf_var._nc_attrs` rather than directly from `cf_var`. I tested your suggestion and it looks like this would cause attributes to load incorrectly.

This does mean that we are bypassing a mechanism which is specifically designed to improve performance. I have a feeling that the ideal solution may have to involve separating the mechanism for accessing `cf_var` attributes from the mechanism for chosing which `cf_var` attributes to ignore when setting iris attributes.

pp-mo: > @stephenworsley I have a feeling that the ideal solution may have to involve separating the mechanism for accessing cf_var attributes from the mechanism for chosing which cf_var attributes to ignore when setting iris attributes.

You're dead right ! 
I do think that the overlapping schemes in CFVariable generate a whole heap of extra confusion.
( I spent quite some time trying to work out why the 'ignored' and 'unused' ideas work in just the the way that they do.  )

stephenworsley: Having said this, checking the attributes of `cf_var.cf_data` seems to work while relying on more public API.

pp-mo: So, I'm really sorry for making a duff suggestion !
A simple "clean" solution would be to test `hasattr(cf_var.nc_variable)` instead of `hasattr(cf_var)`.  But, like you, I'm still a bit concerned about bypassing the speedup mechanism (even though I suspect that at least some of it is probably unnecessary).

Another really simply fix for my original objection would be to retain your existing code solution, but just make `cf_var._nc_attrs` *public* -- by renaming, removing the underscore.  
**How would you feel about that ?**

( Note: there aren't many existing uses of this, and they are all within cf.py )

pp-mo:
@stephenworsley WARNING: I see we are losing Travis testing again.
I wish I understood why this happens !
I know it can occur when there is a 'syntax error' (or equivalent) in `.travis.yml`, but I doubt that is the case this time ...

pp-mo:
> the errors you mention (which presumably are breaking some tests ?)

Yes, I find that 3 are failing in `iris.tests.test_netcdf`
Notably:
```
FAIL: test_cell_measures (__main__.TestNetCDFLoad)
  File "lib/iris/tests/test_netcdf.py", line 398, in test_cell_measures
    self.assertEqual(cms[0], expected)
AssertionError: CellM[203 chars] var_name='my_areas') != CellM[203 chars] var_name='my_areas', attributes={'custom': 'extra-attribute'})
```

lib/iris/fileformats/_pyke_rules/fc_rules_cf.krb
stephenworsley: I feel like either of these could work. To be honest, I don't have a strong enough picture of how `cf_var` is structured to have a clear idea of which feels more correct. In either case, I suspect both approaches would require changing the a lot of the tests currently using a mocked `cf_var`.

pp-mo:
Ok, this is looking good !
I think all your tweaks to the Mock tests look neat + sensible (though I know it can take ages  to work out _where_ to make these adjustments !)

So, I think that now happily "proves" the correctness of the essential change.
The problems/questions remaining here are :
  #1. includes tweaks to requirements, that will need removing when we merge in the relevant fixes on master
  #2. includes hacks for saving, that will need removing when we merge with `default_units` feature branch

I think the correct approach is just to merge this, leaving those changes in, and take care of them in the merge process.
At least, leaving in the requirements changes means that we can do further work with launch_ancils branch if we need to (though I suspect we don't need to)

PR #3614, opened 2019-12-19
Add colour logo - take2!

abooton:


abooton:
Should be included into master after 2.3.x mergeback.

PR #3615, opened 2019-12-19
Mergeback 2v3

pp-mo:
Merge-back from 2.3.x following 2.3.0 release.

Since we already merged back 2.2, and not a lot happened from 2.2.0 to 2.3.0, I think this is not so awful as sometimes occurs...


pp-mo:
> Some of the commits this is adding already exist on master

This seemed to relate to commits that introduced 2.3 whatsnew entries + then replaced them with 2.3.rst.  I've now skipped those, plus the version-number changes, and I think this should now suit ok ...

pp-mo:
***WARNING*** I had to add an extra commit to fix a problem with the changes to `iris.fileformats.um_cf_map`.
This is due to a sequence of changes on 2.3.x whereby a new-style licence header was introduced, then the python2 removed, then the old-style header was instated :  We skipped the ['reversion to oldstyle header'] commit but this has confused the merge process, and the result was an update um_cf_map with new translations, new metarelate ref-commit, and new-style header, but the py2 future and six imports are still in place.

Issue #3478, opened 2019-10-22
[PI] Implement new cube arithmetic proposal


lbdreyer:
Following the decision from #2599, we need to _implement_ the approach we decide on

Reference:
* [FB] PR #3583 
* [FB] PR #3739
* [FB] PR #3774
* PR #3785 (`cube-arithmetic` feature branch merge back to `master`)
* [FB] PR #3805
* [FB] PR #3806
* [FB] PR #3842
* [FB] PR #3869

----
#### Core deliverables
- [x] Add lenient infra-structure:
    - [x] client decorator
    - [x] service decorator
    - [x] context manager 
    - [x] runtime options state manager (thread-safe)
    - [x] client/service active protocol
- [x] Support lenient metadata functionality:
    - [x] comparison
    - [x] combination
    - [x] difference
- [x] Support lenient coordinate (macroscopic) compliance checking
- [x] ~Add cube arithmetic switch to opt-in to previous/strict behaviour~
- [x] Remove `AncillaryVariable`s and `CellMeasure`s from resultant cube
- [ ] Test coverage
- [ ] Discuss `metadata` rename to `signature` ? @pp-mo 
- [x] Add associated ASV metrics
- [x] Resolve dilemma of whether `circular` is part of `DimCoord` metadata
    - what are the implications of this?
        - require `DimCoordMetadata`, `AuxCoordMetadata` and `CoordMetadata` (common parent) support
        - utility of supporting cross-metadata setting for `DimCoord` and `AuxCoord`
- [ ] Close connective tissue PRs/Issues
- [x] Add supporting documentation and examples etc
    - Clarify metadata `name` vs `identity` 
    - Define lenient/strict operations for `__eq__`, `combine` and `difference`
    - `<container>.metadata` setter/getter patterns
- [ ] Add `whatsnew` entry

----
#### Optional deliverables
Not sure whether some of these may get pulled into the _core deliverables_...
- [ ] Support lenient coodinate metadata in `iris.cube.Cube.coords`
    - `iris.cube.Cube.replace_coord`
    - `iris.cube.Cube.remove_coord`
- [x] Unify metadata into a common API returning `namedtuple` derived classes:
    - [x] ~Use `lru_cache` for optimised LUT metadata generation~
        - Issues with non-hashable metadata :cry: 
            - `attributes`
            - `coord_system`
    - [x] Create a common `BaseMetadata` class hierarchy
        - [x] Use a `_BaseMeta` metaclass
[...]

PR #3618, opened 2019-12-20
PI-3483 (if not merging 3478) Safe arithmetic

stephenworsley:
Align arithmetic with previous behaviour wrt CellMeasure

In the absence of changes to arithmetic due with #3478, cube arithmetic is given the "safe" behaviour of always discarding cell measures and ancillary variables. This aligns with the previous behaviour of arithmetic discarding cell measures (which was previously due to copy discarding cell measures). This solves one of the remaining issues in #3483 and allows loading of ancillary variables to be merged in safely if the canges made to cube arithmetic don't make it into 3.0.

### DIFF
--- a/lib/iris/analysis/__init__.py
+++ b/lib/iris/analysis/__init__.py
@@ -2316,17 +2316,13 @@ class _Groupby:
 
 def clear_phenomenon_identity(cube):
     """
-    Helper function to clear the standard_name, attributes, cell_methods,
-    cell_measures and ancillary_variables of a cube.
+    Helper function to clear the standard_name, attributes and
+    cell_methods of a cube.
 
     """
     cube.rename(None)
     cube.attributes.clear()
     cube.cell_methods = tuple()
-    for cm in cube.cell_measures():
-        cube.remove_cell_measure(cm)
-    for av in cube.ancillary_variables():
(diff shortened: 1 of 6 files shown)


lib/iris/analysis/__init__.py
```diff
@@ -2316,13 +2316,17 @@ def __repr__(self):
 
 def clear_phenomenon_identity(cube):
     """
-    Helper function to clear the standard_name, attributes, and
```
pp-mo: This isn't a good extension of the function, because cell-measures and ancils are *not* part of phenomenon identity.
This is also a public API function, so it does need to do what it says on the tin.

I reckon the additional aspects should be separately handled, adjacent to where [this is function is called in `_math_op_common`](https://github.com/SciTools/iris/blob/b0027d4b72e7dd43e07a6d6571d93d23f4d8c641/lib/iris/analysis/maths.py#L946).

pp-mo:
:+1: LGTM
But not merging this yet.

As I understand it, we are holding this solution in reserve in case we want to "make arithmetic safe" for a possible future Iris 3.0 release that ***not including #3478***.
If we do complete #3478 and include that in Iris 3.0, this is then probably unneeded.

PR #3622, opened 2019-12-20
PI-3473 Rename "engine.provides" (#3590)

abooton:
During work on #3556, it became clear to me that the use of `engine.provides` within the pyke rules, and in a few places within `iris.fileformats.netcdf` is confusing :  Because this term collides with the special reserved meaning of "provides" within Pyke rules, as seen in the repeated usage of `facts_cf.provides(..)` in most of our Pyke rules.

In fact, although `engine` is a Pyke object,  `engine.provides` is just a monkey-patched additional property attached by our netcdf code, to keep track of cube component objects created during rules evaluation (coordinates, cell-measures and ancillaries).

So, this PR just renames it to something less confusing. 
This should produce **no functional change.**

Note - See #3590 for comments.  This PR is identical to it, but re-checks the tests as one of the travis tests had stalled for an unknown reason and couldn't be restarted.


abooton:
Closed. Duplicate oof #3590

PR #3623, opened 2019-12-23
PI-2472: Area weighted regridding

ehogan:
This pull requests is for the regridding feature branch :) :tada:


abooton:
Unfortunately the data-dims check on the regridder is too strict for global data, and I found some bugs in the unit tests that we should also fix.  Therefore #3625 is needed before this feature-branch can be merged.

abooton:
@stephenworsley has also reviewed this.  Looking good now.

PR #3625, opened 2019-12-29
PI-2472:  Update area weighted regridder to accept different dim data

abooton:
Alter the area-weighted regridder to accept multidimensional src cubes (i.e. data other than 2D lat-lon data.), identifying the x-y coordinates corresponding to src grid.
Update the regridder to only cache the vital regridding information (i.e. target grid info and the weights information.
Fix bug in the tests so that the prepare and perform functions are called sequentially.
Update "test_mismatched_data_dims" to accept multidimensional data, with dimensions in differing orders.

### DIFF
--- a/lib/iris/tests/unit/analysis/area_weighted/test_AreaWeightedRegridder.py
+++ b/lib/iris/tests/unit/analysis/area_weighted/test_AreaWeightedRegridder.py
@@ -232,6 +232,18 @@ class Test(tests.IrisTest):
         self.assertArrayShapeStats(
             result, (8, 5, 9), expected_mean, expected_std
         )
+        # Check data with dims in different order
+        # Reshape src so that the coords are ordered [y, x, z],
+        # the mean and std statistics should be the same
+        data = np.moveaxis(src.data.copy(), 2, 0)
+        src = Cube(data)
+        src.add_dim_coord(lat, 0)
+        src.add_dim_coord(lon, 1)
+        src.add_dim_coord(levels, 2)
+        result = regridder(src)
+        self.assertArrayShapeStats(
+            result, (9, 8, 5), expected_mean, expected_std
+        )
 
 
 if __name__ == "__main__":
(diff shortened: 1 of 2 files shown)


lib/iris/tests/unit/analysis/area_weighted/test_AreaWeightedRegridder.py
```diff
@@ -193,37 +199,39 @@ def test_multiple_src_on_same_grid(self):
...
+        src = Cube(data)
+        src.add_dim_coord(levels, 0)
+        src.add_dim_coord(lat, 1)
+        src.add_dim_coord(lon, 2)
+        result = regridder(src)
+        self.assertArrayShapeStats(
+            result, (5, 9, 8), expected_mean, expected_std
+        )
+        # Check data with dims in different order
+        # Reshape src so that the coords are ordered [x, z, y],
```
stephenworsley: There is no longer a case tested in which x is assigned to a higher dimension than y e.g.`[y, x, z]`. Given that these two cases are treated differently in the code (specifically, the code in `_regrid_area_weighted_array`), both of these cases should probably be covered in the tests.

lib/iris/tests/unit/analysis/area_weighted/test_AreaWeightedRegridder.py
```diff
@@ -30,42 +31,53 @@ def cube(self, x, y):
...
         cube.add_dim_coord(lon, 1)
+        cube.coord("latitude").guess_bounds()
+        cube.coord("longitude").guess_bounds()
         return cube
 
     def grids(self):
         src = self.cube(np.linspace(20, 30, 3), np.linspace(10, 25, 4))
-        target = self.cube(np.linspace(6, 18, 8), np.linspace(11, 22, 9))
+        target = self.cube(np.linspace(22, 28, 8), np.linspace(11, 22, 9))
         return src, target
```
abooton: Alter this so that we are passing data through the tests. It is ok to do this because all the tests already using it abort with an error and are independent of the data used.

PR #3628, opened 2020-01-06
Allow some 'key=None' args in Geostationary creation.

### LINKED ISSUE #3626: Inflexibility of Geostationary coordinate-systems args
#3406 added `iris.coord_systems.Geostationary` and netcdf read/write support for it.
However, this did not have much practical testing.

Since found a specific problem:

With a netcdf sourcefile looking like this :
```
netcdf geostationary_problem_case {
dimensions:
    y = 2 ;
    x = 3 ;
variables:
    short radiance(y, x) ;
        radiance:standard_name = "toa_outgoing_radiance_per_unit_wavelength" ;
        radiance:units = "W m-2 sr-1 um-1" ;
        radiance:coordinates = "y x" ;
        radiance:grid_mapping = "imager_grid_mapping" ;
    short y(y) ;
        y:units = "rad" ;
        y:axis = "Y" ;
        y:long_name = "fixed grid projection y-coordinate" ;
        y:standard_name = "projection_y_coordinate" ;
    short x(x) ;
        x:units = "rad" ;
        x:axis = "X" ;
        x:long_name = "fixed grid projection x-coordinate" ;
        x:standard_name = "projection_x_coordinate" ;
    int imager_grid_mapping ;
        imager_grid_mapping:long_name = "fixed grid projection" ;
        imager_grid_mapping:grid_mapping_name = "geostationary" ;
        imager_grid_mapping:perspective_point_height = 35786023. ;
        imager_grid_mapping:semi_major_axis = 6378137. ;
        imager_grid_mapping:semi_minor_axis = 6356752.31414 ;
        imager_grid_mapping:latitude_of_projection_origin = 0. ;
        imager_grid_mapping:longitude_of_projection_origin = -75. ;
        imager_grid_mapping:sweep_angle_axis = "x" ;
}
```

`iris.load` then fails like this :
```
>>> cubes = iris.load(filepath)
  File "/home/h05/itpp/git/iris/iris_main/lib/iris/__init__.py", line 304, in load
    return _load_collection(uris, constraints, callback).merged().cubes()
  File "/home/h05/itpp/git/iris/iris_main/lib/iris/__init__.py", line 271, in _load_collection
    result = iris.cube._CubeFilterCollection.from_cubes(cubes, constraints)
  File "/home/h05/itpp/git/iris/iris_main/lib/iris/cube.py", line 144, in from_cubes
    for cube in cubes:
  File "/home/h05/itpp/git/iris/iris_main/lib/iris/__init__.py", line 258, in _generate_cubes
    for cube in iris.io.load_files(part_names, callback, constraints):
  File "/home/h05/itpp/git/iris/iris_main/lib/iris/io/__init__.py", line 207, in load_files
    for cube in handling_format_spec.handler(fnames, callback):
  File "/home/h05/itpp/git/iris/iris_main/lib/iris/fileformats/netcdf.py", line 788, in load_cubes
    cube = _load_cube(engine, cf, cf_var, filename)
  File "/home/h05/itpp/git/iris/iris_main/lib/iris/fileformats/netcdf.py", line 598, in _load_cube
    engine.activate(_PYKE_RULE_BASE)
  File "/tmp/persistent/miniconda3/envs/iristest_py3/lib/python3.6/site-packages/pyke/knowledge_engine.py", line 297, in activate
    for rb_name in rb_names: self.get_rb(rb_name).activate()
  File "/tmp/persistent/miniconda3/envs/iristest_py3/lib/python3.6/site-packages/pyke/rule_base.py", line 159, in activate
    self.run_fc_rules(current_rb)
  File "/tmp/persistent/miniconda3/envs/iristest_py3/lib/python3.6/site-packages/pyke/rule_base.py", line 147, in run_fc_rules
    for fc_rule in rb.fc_rules: fc_rule.run()
  File "/tmp/persistent/miniconda3/envs/iristest_py3/lib/python3.6/site-packages/pyke/fc_rule.py", line 90, in run
    self.rule_fn(self)
  File "/home/h05/itpp/git/iris/iris_main/lib/iris/fileformats/_pyke_rules/compiled_krb/fc_rules_cf_fc.py", line 223, in fc_provides_grid_mapping_geostationary
    build_geostationary_coordinate_system(engine, cf_grid_var)
  File "/home/h05/itpp/git/iris/iris_main/lib/iris/fileformats/_pyke_rules/compiled_krb/fc_rules_cf_fc.py", line 2169, in build_geostationary_coordinate_system
    false_northing, ellipsoid)
  File "/home/h05/itpp/git/iris/iris_main/lib/iris/coord_systems.py", line 771, in __init__
    self.false_easting = float(false_easting)
TypeError: float() argument must be a string or a number, not 'NoneType'
>>> 
```

---

"The problem" is that Geostationary provides defaults for an unspecified `false_easting` or `false_northing`, but can't handle a passed `None` in these arguments.

Strictly speaking, the [cf conventions documentation](http://cfconventions.org/Data/cf-conventions/cf-conventions-1.7/cf-conventions.html#_geostationary_projection) doesn't suggest that any of the parameters can be omitted.
But that account is very incomplete, and it doesn't help if "standard"  data sources have this problem, like the original NOAA-derived test data this problem was found in.

pp-mo:
Closes #3626 

Targetting master for now, although this *may* be redirected to Iris 2.4 if we choose to go down that route.

### DIFF
--- a/lib/iris/coord_systems.py
+++ b/lib/iris/coord_systems.py
@@ -714,8 +714,8 @@ class Geostationary(CoordSystem):
         longitude_of_projection_origin,
         perspective_point_height,
         sweep_angle_axis,
-        false_easting=0,
-        false_northing=0,
+        false_easting=None,
+        false_northing=None,
         ellipsoid=None,
     ):
 
@@ -768,9 +768,13 @@ class Geostationary(CoordSystem):
         self.perspective_point_height = float(perspective_point_height)
 
         #: X offset from planar origin in metres.
+        if false_easting is None:
+            false_easting = 0
         self.false_easting = float(false_easting)
 
         #: Y offset from planar origin in metres.
+        if false_northing is None:
+            false_northing = 0
         self.false_northing = float(false_northing)
 
         #: The axis along which the satellite instrument sweeps - 'x' or 'y'.
--- a/lib/iris/tests/integration/test_netcdf.py
+++ b/lib/iris/tests/integration/test_netcdf.py
@@ -12,7 +12,9 @@ import iris.tests as tests
 from contextlib import contextmanager
 from itertools import repeat
 import os.path
+from os.path import join as path_join
 import shutil
+from subprocess import check_call
 import tempfile
 from unittest import mock
 import warnings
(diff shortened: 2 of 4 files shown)


lib/iris/coord_systems.py
```diff
@@ -714,8 +714,8 @@ def __init__(
         longitude_of_projection_origin,
         perspective_point_height,
         sweep_angle_axis,
-        false_easting=0,
-        false_northing=0,
```
pp-mo: I don't think this is an API change, in effect, because ...
 * you can either omit, or say =value, or =None
 * previously this resulted in 0; =value; fail
 * now results in 0; =value; 0.
 * so that is just extra additional behaviour : no usage is changed, except what used to fail

lib/iris/tests/unit/fileformats/pyke_rules/compiled_krb/fc_rules_cf_fc/test_build_geostationary_coordinate_system.py
```diff
@@ -22,47 +22,57 @@
...
         self._test(inverse_flattening=True)
+
+    def test_false_offsets_missing(self):
+        self._test(remove_props=['false_easting', 'false_northing'])
+
+    def test_false_offsets_none(self):
+        self._test(replace_props={'false_easting':None, 'false_northing':None})
+
+
+if __name__ == "__main__":
```
pp-mo: This enables script-usage, was previously omitted (a mistake, I think)

lib/iris/coord_systems.py
trexfeathers: Agreed

lib/iris/tests/unit/fileformats/pyke_rules/compiled_krb/fc_rules_cf_fc/test_build_geostationary_coordinate_system.py
trexfeathers: 😱 thanks

lib/iris/tests/integration/test_netcdf.py
```diff
@@ -592,5 +594,79 @@ def test_standard_name_roundtrip(self):
...
 
 
+class TestLoadMinimalGeostationary(tests.IrisTest):
+    """
+    Check we can load data with a geostationary grid-mapping, even when the
+    'false-easting' and 'false_northing' properties are missing.
+
+    """
+
+    _geostationary_problem_cdl = """
```
trexfeathers: If this is from real data, I suggest we should be storing the original in [iris-test-data](https://github.com/SciTools/iris-test-data) and using that. If this is a highly simplified test case then we should store it in a dedicated file within [/tests/results/integration/](lib/iris/tests/results/integration), as has been done with the other use of ncgen in testing - [/tests/integration/test_climatology.py](lib/iris/tests/integration/test_climatology.py).

pp-mo:
Rebased on top of (effectively) #3630.  See if that works...

lib/iris/tests/integration/test_netcdf.py
pp-mo: > If this is a highly simplified test case then we should store it in a dedicated file within /tests/results/integration/, as has been done with the other use of ncgen in testing - /tests/integration/test_climatology.py.

Well, it _is_ such a "highly simplified" testcase, I think.
But I _don't_ think it would be better to have the CDL in a file : I think it's simpler to have it here where you can read it, right in the only code that uses it.
In the 'test_climatology' case, I think we wanted an actual file under 'results', just so we could use `IrisTest.assertCDL` to check a round-tripped result against the original.

pp-mo:
> #3630. See if that works...

Seems to fix the key test problem.
So we should resolve that issue first..

pp-mo:
Rebased.  
Should be ok @trexfeathers ?

lib/iris/tests/integration/test_netcdf.py
trexfeathers: I agree with the principle. This isn't something that's been done before in Iris but hopefully it's a precedent we can establish while working on #3629 .

PR #3633, opened 2020-01-13
Bump version to 2.4.0rc0.

trexfeathers:


trexfeathers:
The latest failure is because I haven't yet included the commit that pins pillow. Rather than attempting to advance the version number in isolation I think it will be best to put everything together locally then submit as a working PR (or several).

PR #3634, opened 2020-01-14
Fix conflicts in netcdf var_name

stephenworsley:
There was a bug when two cubes were being saved where the first cube shared a var_name with a the var_name of a coordinate on the second cube. In such a case, Iris was failing to save. This can be solved by appropriately incrementing the var_name to avoid conflicts in such cases.


pp-mo:
?Maybe see also: #3643

stephenworsley:
I believe this is still worth adding.

trexfeathers:
@stephenworsley fancy updating the branch and we can restart the review process?

PR #3641, opened 2020-01-16
PI-3637: Allow derived coordinates to be removed

### LINKED ISSUE #3637: Cube.aggregated_by() fails when derived coordinates are present
Hi, I've encountered the following bug(?) while working with files that contain `derived coordinates`:

```python
import iris
import numpy as np
path = iris.sample_data_path('hybrid_height.nc')
cube = iris.load(path)[0]
aux_coord = iris.coords.AuxCoord(np.arange(cube.shape[1]) % 7,
                                 long_name='random aux coord')
cube.add_aux_coord(aux_coord, 1)
cube.aggregated_by('random aux coord', iris.analysis.MEAN)  # fails
```

fails with

```
  File "bug_derived_coord.py", line 9, in <module>
    cube.aggregated_by('random aux coord', iris.analysis.MEAN)
  File "miniconda3/envs/test/lib/python3.7/site-packages/iris/cube.py", line 3506, in aggregated_by
    self.coord_dims(coord))
  File "miniconda3/envs/test/lib/python3.7/site-packages/iris/cube.py", line 965, in add_aux_coord
    raise ValueError('Duplicate coordinates are not permitted.')
ValueError: Duplicate coordinates are not permitted.
```

If the derived coordinate and the coordinate used for aggregation do not share a common dimension, it works:

```python
import iris
import numpy as np
path = iris.sample_data_path('hybrid_height.nc')
cube = iris.load(path)[0]
cube = iris.util.new_axis(cube)
cube.add_aux_coord(iris.coords.AuxCoord(0, long_name='random aux coord'), 0)
cube = cube.aggregated_by('random aux coord', iris.analysis.MEAN)  # does not fail
```

If I remove the derived coordinate before the aggregation, it doesn't fail, too:

```python
import iris
import numpy as np
path = iris.sample_data_path('hybrid_height.nc')
cube = iris.load(path)[0]
aux_coord = iris.coords.AuxCoord(np.arange(cube.shape[1]) % 7,
                                 long_name='random aux coord')
cube.add_aux_coord(aux_coord, 1)
for aux_factory in cube.aux_factories:
    cube.remove_aux_factory(aux_factory)
cube = cube.aggregated_by('random aux coord', iris.analysis.MEAN)  # does not fail
print(cube)
```

correctly prints

```
air_potential_temperature / (K)     (model_level_number: 15; -- : 7; grid_longitude: 100)
     Dimension coordinates:
          model_level_number                           x        -                  -
          grid_longitude                               -        -                  x
     Auxiliary coordinates:
          atmosphere_hybrid_height_coordinate          x        -                  -
          sigma                                        x        -                  -
          grid_latitude                                -        x                  -
          random aux coord                             -        x                  -
          surface_altitude                             -        x                  x
     Scalar coordinates:
          forecast_period: 0.0 hours
          forecast_reference_time: 2009-09-09 17:10:00
          time: 2009-09-09 17:10:00
     Attributes:
          Conventions: CF-1.5
          STASH: m01s00i004
          source: Data from Met Office Unified Model 7.04
     Cell methods:
          mean: unknown
```

Thus, regular multidimensional coordinates that share dimensions with the aggregated coordinate (in this case `surface_altitude`) are also not a problem.

If this this is not a bug, a more precise error message would be nice. Thanks for your help!

stephenworsley:
A derived coordinate can now be removed by calling `Cube.remove_coord()`. Fixes #3637. `aggregated_by()` now handles derived coordinates. The behaviour of `aggregated_by` is now that derived coordinates will become regular coordinates if they span the dimension aggregated along.


pp-mo:
Thanks @stephenworsley .

Unfortunately I think merge was a bit premature : I just realised that it still needs a whatsnew.
Can you maybe address that ?

The effective change to 'aggregated_by' also needs a bit more attention, I think.
As a minimum, I'd suggest a separate whatsnew for that.
***However,*** I'm also still very doubtful whether the 'new' behaviour of 'aggregated_by' is a correct resolution :  *If we have aggregated all the dependency coordinates, can we not construct a proper derived coord for the result in this case?*

Issue #3186, opened 2018-10-04
Decide whether cubes are iterable


DPeterK:
At the moment, cubes report as being iterables, but raise an error if you try to iterate over a cube:

```python
>>> isinstance(cube, Iterable)
True

>>> for c in cube:
...     print c

TypeError: Cube is not iterable
```

This is confusing and potentially a bit awkward, as it looks like Iris can't make its mind up...

Of course this comes down to the presence of the `__iter__` method, whose presence means the Cube class reports as being iterable, while the method itself just raises an error. We could:

* remove the `__iter__` method, which is simple to do but unfortunately reverses the problem (Cube reports as not being iterable, while you actually _can_ then iterate over a cube on its leading dimension)
* explicitly support iterating by extending `__iter__`, which is more complicated and potentially undefined
* ... do something more clever than either of these suggestions
* just ignore it

pp-mo:
Long long ago... 
we explicitly stopped cubes being iterables because it was confusing a lot of users.
See : https://github.com/SciTools/iris/issues/86

PR #3656, opened 2020-02-13
Remove cube iter

jonseddon:
This change has been discussed with @lbdreyer. A TypeError with a useful message is still raised by Python if a Cube is iterated over but `isinstance(cube, collections.Iterable)` now behaves as would be expected. This behaviour is documented in https://docs.python.org/3/reference/datamodel.html#special-method-names.

### DIFF
--- a/docs/iris/src/userguide/subsetting_a_cube.rst
+++ b/docs/iris/src/userguide/subsetting_a_cube.rst
@@ -103,6 +103,9 @@ same way as loading with constraints:
 
 Cube iteration
 ^^^^^^^^^^^^^^^
+It is not possible to directly iterate over an Iris cube. That is, you cannot call `cube.next()` as
+happens in for loops and other forms of iteration. Instead you can iterate over cube slices, as this section details.
+
 A useful way of dealing with a Cube in its **entirety** is by iterating over its layers or slices.
 For example, to deal with a 3 dimensional cube (z,y,x) you could iterate over all 2 dimensional slices in y and x
 which make up the full 3d cube.::
--- a/docs/iris/src/userguide/subsetting_a_cube.rst
+++ b/docs/iris/src/userguide/subsetting_a_cube.rst
@@ -103,8 +103,8 @@ same way as loading with constraints:
 
 Cube iteration
 ^^^^^^^^^^^^^^^
-It is not possible to directly iterate over an Iris cube. That is, you cannot call `cube.next()` as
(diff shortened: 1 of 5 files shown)


DPeterK:
@jonseddon this looks good, thanks for putting it together! I wonder whether it would be worth a comment to this in the Iris user guide - probably in §5.2 - [cube iteration](https://scitools.org.uk/iris/docs/latest/userguide/subsetting_a_cube.html#cube-iteration)?

It could be reasonably simple - something like the following, for example:

```
It is not possible to directly iterate over an Iris cube. That is, you cannot call `cube.next()`. 
Instead you can iterate over cube slices, as this section details.
```

rcomer:
Closes #3186

jonseddon:
> @jonseddon this looks good, thanks for putting it together! I wonder whether it would be worth a comment to this in the Iris user guide - probably in §5.2 - [cube iteration](https://scitools.org.uk/iris/docs/latest/userguide/subsetting_a_cube.html#cube-iteration)?

@DPeterK  Thanks! Good suggestion. That's added now and the documentation's building successfully.

docs/iris/src/userguide/subsetting_a_cube.rst
```diff
@@ -103,6 +103,9 @@ same way as loading with constraints:
 
 Cube iteration
 ^^^^^^^^^^^^^^^
+It is not possible to directly iterate over an Iris cube. That is, you cannot call `cube.next()` as
```
pp-mo: I don't think it's very good (modern) Python usage to call a `next` method directly.  Cubes were once iterable by virtue of being indexable #273, but they never did have an actual 'next' function anyway. 
So, I think it would be better here to talk about `next(cube)`, or `list(cube)` or `for x in cube`.

jonseddon: @pp-mo Thanks for the review and suggestion. In 2a4f3d2 I've simplified the language slightly, but I think that this should be easily understandable by typical Iris users.

DPeterK:
Thanks @jonseddon! I think the update to the userguide sounds good 👍

pp-mo:
I'm :+1: , think this is good to go.

Any final comment @lbdreyer -- is this what you were expecting ?
( Just because you were namechecked ! )

lbdreyer:
> Any final comment @lbdreyer -- is this what you were expecting ?

Looks good to me :+1:

PR #3657, opened 2020-02-14
Add plugin support for loaders.

rhattersley:
This suggestion uses `iris.io.plugins` as a namespace package. Each plugin would need to provide one or more modules that install into the same namespace package. All the plugin modules get imported when `iris` is imported (strictly speaking, when `iris.fileformats` is imported), and the contents of any `FORMAT_SPECIFICATIONS` iterables get added to the FORMAT_AGENT.

NB. This is just an iniitial technical suggestion (e.g. it has no documentation!) and place to host the general discussion.


rhattersley:
The test failure is genuine and is caused by `setup.py` not installing the new namespace package - it actively filters out empty directories. 😒

bjlittle:
@rhattersley Welcome back :smile: ... where's this PR at? Good for review or still WIP?

rhattersley:
It needs a technical review really. e.g. Is the native-namespace mechanism something we're happy to run with? Is `iris.io.plugins` the right name? I'll pull out the ABF/L loader as a separate repo to provide an example of what it's like to supply a plugin.

Once the technical stuff is settled, then we can sort out documentation, what's new, ... etc.

rhattersley:
For an example plugin (with super-simple, modern packaging) see [SciTools/iris-abf](https://github.com/SciTools/iris-abf). :grin:

rhattersley:
NB. SciTools/iris-abf#1

rhattersley:
FWIW, I've updated the iris-abf package to include tests, use black formatting, and apply CI via GitHub actions.

The tests are brand new tests, with a small amount of "fruitiness" to work around an absence of iris in the test environment due to the inability to do "pip install iris". Hence, I couldn't just grab the tests from here. Installing iris would effectively require conda, which would complicate iris-abf and draw it away from the wholesome poetry, PEP 517, PEP 518, GitHub-actions goodness that it's currently using. This "I can't pip install iris" chicken-and-egg problem is one of the things this PR is seeking to address, but clearly more work is required elsewhere as well ... e.g. I'm looking at _you_, pyke & cf-units. :unamused:

pp-mo:
>  example plugin (with super-simple, modern packaging) see SciTools/iris-abf.

Interesting.
Can you explain what the install process for this would be : 
  * ? does it require to "inject itself" into an Iris installation
  * ? how would that work 
  * ? can it support a local (or "--user") install
  * ? can we automate installation / install from a channel ?
  * ? is it waiting on iris, to adopt the 'fileformats plugins directory', before we can make this work ?

pp-mo:
I still think we should implement this, but probably just a simple naming strategy as discussed above, so leaving this in place as an issue / reminder.
Also... there is an outstanding problem with loader selection : see https://github.com/SciTools/iris/issues/3918

trexfeathers:
Related: #4798

trexfeathers:
Possibly other example that would be ready to try: the NAME loader

trexfeathers:
Closed by #5144

PR #3659, opened 2020-02-14
Stop PPDataProxy accessing the file when no data is needed.

pp-mo:
We just found that PP and Fieldsfile loading has been slow and heavy on memory, apparently since Dask 2.0.

In these cases, the PPDataProxy does not actually use the file data, so **this fix allows it to return a suitable empty array object of the expected type and shape, without reading the file**.

---

The reason appears to be that `dask.array.from_array(arraylike)` now extracts an empty slice of the argument, like `arraylike[0:0, 0:0...]`.

Note, this is even worse than it sounds :
  1. the PP field will often need to be unpacked, which is time consuming
  1. since the 0-size slice is stored within the dask object, and it is a *view* referencing the full data array, hence **all field data is loaded, *and* remains in memory !**

See [this line in `meta_from_array`](https://github.com/dask/dask/blob/86d1011a8890f00acda73e5ccb091769f74df3b8/dask/array/utils.py#L88), called from [here in `Array.__new__`](https://github.com/dask/dask/blob/86d1011a8890f00acda73e5ccb091769f74df3b8/dask/array/core.py#L1035), as used in `from_array`.

Speculation : I think dask is doing this to capture the metadata of a slice from the argument, which it presumably believes can be different from that of the argument itself.  I'm guessing this is needed for generalised array types e.g. dask, CUDA


lib/iris/fileformats/pp.py
```diff
@@ -644,14 +644,30 @@ def ndim(self):
...
-                                               self.lbpack,
-                                               self.boundary_packing,
-                                               self.shape, self.src_dtype,
-                                               self.mdi)
+        def is_emptyslice(key):
+            return (isinstance(key, slice) and
+                    isinstance(key.start, int) and
+                    key.start == key.stop)
+
+        if any(is_emptyslice(key) for key in keys):
```
trexfeathers: Assume you're comfortable with the scenario where only _some_ of the slices are empty, but we still avoid accessing the file at all?

lib/iris/tests/unit/fileformats/pp/test_PPDataProxy.py
```diff
@@ -47,6 +49,27 @@ def test_lnpack_raw(self):
         self.assertEqual(proxy.lbpack.n3, lbpack // 100 % 10)
         self.assertEqual(proxy.lbpack.n4, lbpack // 1000 % 10)
 
+    def test__getitem__emptyslice(self):
```
trexfeathers: We should have two tests, which share as much as possible, but with opposite outcomes:
- One like this, showing no file access and an empty array
- Another showing file access and a non-empty array

lib/iris/fileformats/pp.py
```diff
@@ -644,14 +644,30 @@ def ndim(self):
...
-            pp_file.seek(self.offset, os.SEEK_SET)
-            data_bytes = pp_file.read(self.data_len)
-            data = _data_bytes_to_shaped_array(data_bytes,
-                                               self.lbpack,
-                                               self.boundary_packing,
-                                               self.shape, self.src_dtype,
-                                               self.mdi)
+        def is_emptyslice(key):
+            return (isinstance(key, slice) and
+                    isinstance(key.start, int) and
```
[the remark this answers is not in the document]
pp-mo: I see what you mean : essentially ":0" is the same as "0:0".  
On the other hand, we don't necessarily need to catch *all* possible '0-length' cases.  
There are anyway quite a few more, such as : for any number M greater than the dimension, "M:" and ":-M" are also empty.  So is any "i:(n-i)" or "(n-j):-j" i.e. there is an equivalent negative number for each positive one (in range).
The first-version working code here was simply `any(key == slice(0,0) for key in slices)`.
I thought that was seriously over-specific to what dask does right now, so I'm seeking generalise it "a bit".  Maybe I need to think a bit more about how general this is, or maybe there's a simpler way to make the test 
( to be continued ) ...

lib/iris/fileformats/pp.py
pp-mo: Absolutely, it seemed important to me to extend the capture beyond just exactly what dask does (at present).  In which case, it makes sense to me that it could sensibly skip some/any or all cases where no data is needed.  But exactly how to capture those I'm still thinking on (see [previous](https://github.com/SciTools/iris/pull/3659#discussion_r380237130))..

lib/iris/tests/unit/fileformats/pp/test_PPDataProxy.py
pp-mo: :+1:

pp-mo:
Hi @trexfeathers sorry for the delay on this -- I forgot to push it up !

Following offline discussion, I've now reduced the scope of the behaviour, in favour of simpler code that only supports the exact current need (e.g. it no longer catches "[1:1]" as an empty slice case)

However, I've also rather *increased* the complexity of the testing.
Again sorry, but I did spot some existing problems!
e.g. the previous code could not handle a *single* index key, because Python passes that as a bare object + not a 1-tuple.

As usual, it's tough to be comprehensive because numpy has so many indexing styles.
So, I'm taking it for granted that ***Dask will only use slices and integer indices***
This excludes (e.g.) :
  * keys which are arrays or lists (fancy indexing)
  * ellipsis
  * None / newaxis
  * boolean arrays

I'm just not considering those as possibilities.
Does this seem sound ??

pp-mo:
Note: when we have agreed on an implementation here, we should also adopt it in iris-grib.

docs/iris/src/whatsnew/contributions_2.4.1/bugfix_2020-Feb-14_pp_emptyslices.txt
```diff
@@ -0,0 +1,5 @@
+* Fixed a problem which was causing file loads to fetch *all* field data
```
lbdreyer: Why have you put this as a a 2.4.1 bug fix? Surely its going into Iris 2.4.0?

pp-mo: Ooops.
I thought we already *had* a whatsnew/2.4.rst, but now I see we don't.
So I now think this was just a mistake + will attempt a fix ..

lib/iris/fileformats/pp.py
```diff
@@ -644,14 +644,32 @@ def ndim(self):
...
+        if any(key == slice(0, 0) for key in keys_tuple):
+            # An 'empty' slice has no data : do not open + read the file !!
+            # In these cases, return a 'fake' data array instead.
+            # Needed because, for Dask >= 2.0, "dask.array.from_array" does a
+            # fetch like [0:0, 0:0, ...], to 'snapshot' the array metadata.
+            target_shape = list(self.shape)
+            for i_dim, key in enumerate(keys_tuple):
+                if key == slice(0, 0):
+                    target_shape[i_dim] = 0
+            data = np.zeros((1,), dtype=self.dtype)
```
trexfeathers: Can just pass `target_shape` to `np.zeros`, without needing the extra `np.broadcast_to` step

lib/iris/tests/unit/fileformats/pp/test_PPDataProxy.py
```diff
@@ -48,5 +52,122 @@ def test_lnpack_raw(self):
...
+
+    def test_slicing_1d_empty(self):
+        # A 1d testcase with an empty slicing.
+        self._check_slicing(
+            test_shape=(3,),
+            indices=Slices[0:0],
+            result_shape=(0,),
+            data_was_fetched=False)
+
+    def test_slicing_1d_unrecognised_empty(self):
```
trexfeathers: There are any number of cases that will return `result_shape=(0,)` and `data_was_fetched=True`, and I don't believe we get much value from just testing one of them. It would also be too much effort to try and cover all of them. This test should come out.

lib/iris/tests/unit/fileformats/pp/test_PPDataProxy.py
```diff
@@ -48,5 +52,122 @@ def test_lnpack_raw(self):
...
+    def test_slicing_2d_normal(self):
+        # A 2d testcase with no empty slices.
+        self._check_slicing(
+            test_shape=(3, 4),
+            indices=Slices[2, :3],
+            result_shape=(3,),
+            data_was_fetched=True)
+
+    def test_slicing_2d_allempty(self):
+        # A 2d testcase with an empty slice.
```
trexfeathers:
```suggestion
        # A 2d testcase with all empty slices.
```

lib/iris/fileformats/pp.py
pp-mo: :+1: Good thinking. I will check on this.

lib/iris/tests/unit/fileformats/pp/test_PPDataProxy.py
pp-mo: Ok.

pp-mo:
Thanks @trexfeathers 
I think I have it all done now.  Review away !

trexfeathers:
All looks great, great idea with the abstraction into util 👏

PR #3660, opened 2020-02-17
Fixed asv project name to 'scitools-iris'.

trexfeathers:
Without this, airspeed velocity is unable to pip uninstall iris after benchmarking a commit, resulting in the same commit being benchmarked repeatedly under the names of previous commits.

Appropriate for merging once we are satisfied that the performance issues related to Dask v2.0 are addressed (e.g. #3659).

### DIFF
--- a/asv.conf.json
+++ b/asv.conf.json
@@ -2,7 +2,7 @@
 // details on what can be included in this file.
 {
     "version": 1,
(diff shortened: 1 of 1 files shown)


trexfeathers:
Now that v2.4.0 is out with the performance fixes, this is OK to get merged

PR #3679, opened 2020-03-06
PI-3677: Remove regex from standard names check

### LINKED ISSUE #3677: Valid standard_name (present in STD_NAMES dict) is still rejected because of formatting maflormation
**Summary**
`atmosphere_optical_thickness_due_to_pm1_ambient_aerosol` is not recognized as a valid `standard_name` by iris although it really is, and present in the `STD_NAMES` dict too

**Description and pinpoint of bug**
Hey guys, apols if this is a duplicate but we have just seen this (these are code snippets from iris=2.3 but the codes are the same for 2.4, I just checked):

in `iris/_cube_coord_common.py` l/47:

```
        valid_name_pattern = re.compile(r'''^([a-zA-Z_]+)( *)([a-zA-Z_]*)$''')
        name_groups = valid_name_pattern.match(name)
```

`name_groups` is returned as `None` for `name_groups = valid_name_pattern.match("atmosphere_optical_thickness_due_to_pm1_ambient_aerosol")`
even if `atmosphere_optical_thickness_due_to_pm1_ambient_aerosol` is a valid standard name it's discarded right away. The name can be found in `STD_NAMES` in `iris/std_names.py` and is also a valid CF entry.

**Solution**
Reformatting the regexp call should do, I'll leave that to you :beer:

Point of contact: @bjlittle @lbdreyer (suggested)

jvegasbsc:
The regex in `get_valid_standard_name` was an overkill as the function was already checking the values against a list that contained all the possible values.

This pull request replaces it with a simple split and fixes the bug that was making iris not to accept `atmosphere_optical_thickness_due_to_pm1_ambient_aerosol` as a valid standard_name

Fixes #3677

### DIFF
--- a/lib/iris/_cube_coord_common.py
+++ b/lib/iris/_cube_coord_common.py
@@ -46,7 +46,6 @@ def get_valid_standard_name(name):
     # modifier, separated by one or more blank spaces
 
     if name is not None:
-        name_is_valid = False
         # Supported standard name modifiers. Ref: [CF] Appendix C.
         valid_std_name_modifiers = [
             "detection_minimum",
@@ -56,17 +55,22 @@ def get_valid_standard_name(name):
         ]
 
         name_groups = name.split(maxsplit=1)
-        std_name = name_groups[0]
-        name_is_valid = std_name in iris.std_names.STD_NAMES
-        try:
-            std_name_modifier = name_groups[1]
-        except IndexError:
-            pass  # No modifier
-        else:
-            name_is_valid &= (std_name_modifier in valid_std_name_modifiers)
+        if name_groups:
+            std_name = name_groups[0]
+            name_is_valid = std_name in iris.std_names.STD_NAMES
+            try:
+                std_name_modifier = name_groups[1]
+            except IndexError:
+                pass  # No modifier
+            else:
+                name_is_valid &= (
+                    std_name_modifier in valid_std_name_modifiers)
 
-        if not name_is_valid:
-            raise ValueError("{!r} is not a valid standard_name".format(name))
+            if not name_is_valid:
+                raise ValueError(
+                    "{!r} is not a valid standard_name".format(name))
+        else:
+            name_is_valid = False
 
     return name
 
--- a/lib/iris/_cube_coord_common.py
+++ b/lib/iris/_cube_coord_common.py
@@ -55,18 +55,17 @@ def get_valid_standard_name(name):
             "status_flag",
         ]
 
-        valid_name_pattern = re.compile(r"""^([a-zA-Z_]+)( *)([a-zA-Z_]*)$""")
-        name_groups = valid_name_pattern.match(name)
+        name_groups = name.split(maxsplit=1)
+        std_name = name_groups[0]
+        name_is_valid = std_name in iris.std_names.STD_NAMES
+        try:
+            std_name_modifier = name_groups[1]
+        except IndexError:
+            pass  # No modifier
+        else:
+            name_is_valid &= (std_name_modifier in valid_std_name_modifiers)
 
-        if name_groups:
-            std_name, whitespace, std_name_modifier = name_groups.groups()
-            if (std_name in iris.std_names.STD_NAMES) and (
-                bool(whitespace)
-                == (std_name_modifier in valid_std_name_modifiers)
-            ):
-                name_is_valid = True
-
-        if name_is_valid is False:
+        if not name_is_valid:
             raise ValueError("{!r} is not a valid standard_name".format(name))
 
     return name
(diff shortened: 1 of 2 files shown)


jvegasbsc:
I updated master, but changes must be propagated to 2.4.x to fix ESMValGroup/ESMValTool#1561

pp-mo:
:+1: 
I'd like to say this is good, but I'm not the best expert.

@bjlittle @lbdreyer do we have any further comment on this ?
I felt that this was over-complicated as was, but I could have missed something ..

lib/iris/_cube_coord_common.py
```diff
@@ -55,19 +54,23 @@ def get_valid_standard_name(name):
...
+                pass  # No modifier
+            else:
+                name_is_valid &= (
+                    std_name_modifier in valid_std_name_modifiers)
+
+            if not name_is_valid:
+                raise ValueError(
+                    "{!r} is not a valid standard_name".format(name))
+        else:
+            name_is_valid = False
```
bjlittle, line 73: @jvegasbsc There is no need for the `else` as `name_is_valid` is not used outside the scope of the `if`

lib/iris/tests/unit/cube_coord_common/test_get_valid_standard_name.py
```diff
@@ -19,10 +19,26 @@ class Test(tests.IrisTest):
     def setUp(self):
         self.emsg = "'{}' is not a valid standard_name"
 
+    def test_none_is_valid(self):
```
bjlittle, line 22: @jvegasbsc I'm going to be a tad fussy here, sorry... `None` is *valid* in the sense that no exception is raised by `get_valid_standard_name`, but obviously it's not an actual *valid* standard name.

Could we change the intent of this test name to be something like `test_none_passed_thru` instead?

lib/iris/tests/unit/cube_coord_common/test_get_valid_standard_name.py
```diff
@@ -19,10 +19,26 @@ class Test(tests.IrisTest):
...
+        name = None
+        self.assertEqual(get_valid_standard_name(name), name)
+
+    def test_empty_is_not_valid(self):
```
bjlittle, line 26: @jvegasbsc Same here, change test name to be `test_empty_passed_thru`...

lib/iris/tests/unit/cube_coord_common/test_get_valid_standard_name.py
```diff
@@ -19,10 +19,26 @@ class Test(tests.IrisTest):
...
+        name = ''
+        self.assertEqual(get_valid_standard_name(name), name)
+
+    def test_only_whitespace_is_not_valid(self):
```
bjlittle, line 30: @jvegasbsc Same here, change test name to be `test_only_whitespace_passed_thru`...

lib/iris/tests/unit/cube_coord_common/test_get_valid_standard_name.py
```diff
@@ -19,10 +19,26 @@ class Test(tests.IrisTest):
...
+        name = '       '
+        self.assertEqual(get_valid_standard_name(name), name)
+
     def test_valid_standard_name(self):
         name = "air_temperature"
         self.assertEqual(get_valid_standard_name(name), name)
 
+    def test_standard_name_alias(self):
+        name = "atmosphere_optical_thickness_due_to_pm1_ambient_aerosol"
+        self.assertEqual(get_valid_standard_name(name), name)
```
bjlittle, line 40: @jvegasbsc Nice, covers you test case, thanks! :+1:

bjlittle:
@jvegasbsc almost there... you just have a couple of `travis-ci` failures.

```shell
black, version 19.10b0
would reformat /home/travis/build/SciTools/iris/lib/iris/_cube_coord_common.py
would reformat /home/travis/build/SciTools/iris/lib/iris/tests/unit/cube_coord_common/test_get_valid_standard_name.py
Oh no! 💥 💔 💥
```

We've now adopted the black coding standard for `iris`. To fix your failing issues please install black into your environment e.g., `conda install -c conda-forge black` or `pip install black`, then go to the root directory of `iris` and perform the following command `black .`. This will reformat the non-compliant files `_cube_coord_common.py` or `test_get_valid_standard_name.py`.

Alternatively, you can could also install `pre-commit` into your environment from either `conda-forge` or `PyPI`, then go to the root directory of `iris` and perform the command `pre-commit install`... and this will install our git pre-commit hooks, which will perform a whole host of checks for each `git commit` on `iris` - thus catching such issues automatically

bjlittle:
@jvegasbsc Also see https://scitools-docs.github.io/iris/master/developers_guide/code_format.html?highlight=black

jvegasbsc:
Done!

bjlittle:
@jvegasbsc Awesome, thanks for your patience and fixing this issue, much appreciated :+1:

PR #3681, opened 2020-03-10
Populate `DimCoord.__init__` docstring

### LINKED ISSUE #3520: Add detail to DimCoord.__init__ docstring
Uncovered by work on #3514. The docstring for `Coord.__init__` contains a lot of detail that is lost in `DimCoord.__init__`.

rcomer:
Adds some detail about the args and kwargs to the `DimCoord.__init__` docstring.  I noticed this omission this morning while helping an inexperienced user who needs to create a `DimCoord` from scratch.

I've given some thought to the descriptions for `points`, `bounds` and `circular`, but everything else is a direct copy/paste from the equivalent `Coord` docstring.  I also tweaked the description for `points` in the `Coord` version, as this way scans better in my head.

A couple of things I'm not sure of:

* I've used "numpy array-like" in the new points description.  Is this a concept that we can assume users are familiar with, or is there a better way of saying it?  If this is the best way, should I also use it for bounds, and for points and bounds in the `Coord.__init__` docstring?
* Really not sure I've understood the `circular` attribute, so what I've written could be nonsense!

Closes #3520

### DIFF
--- a/lib/iris/coords.py
+++ b/lib/iris/coords.py
@@ -1453,8 +1453,8 @@ class Coord(_DimensionalMetadata):
         Args:
 
         * points:
-            The values (or value in the case of a scalar coordinate) of the
-            coordinate for each cell.
+            The values (or value in the case of a scalar coordinate) for each
+            cell of the coordinate.
 
         Kwargs:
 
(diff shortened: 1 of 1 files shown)


rcomer:
While I was getting to grips with the monotonicity rules, I realised it is possible to create a dimcoord whose points and bounds are monotonic in opposite directions.  Not sure that's quite right, but it's beyond the scope of this PR.

rcomer:
Travis failed on image comparisons.  I'm pretty sure I can't have caused this!

lib/iris/coords.py
```diff
@@ -2485,6 +2485,54 @@ def __init__(
         Create a 1D, numeric, and strictly monotonic :class:`Coord` with
         read-only points and bounds.
 
+        Args:
+
+        * points:
+            1D numpy array-like of values (or single value in the case of a
```
rcomer: Given that it’s specifically 1D, perhaps simply “Sequence of...” would be better?

pp-mo:
> Travis failed on image comparisons. I'm pretty sure I can't have caused this!

Yes, not your fault.  Working on it ... #3682

lib/iris/coords.py
pp-mo: I think it would be wiser to stick with "numpy array-like", because I think that has come to be recognised as a common usage meaning "I can do an as_array() on it". 
The idea may extend to other array types like CUDA or dask, as well as numpy.
In addition to iterables (of iterables), it can mean an object that provides `.shape` + `.dtype`, or `__array_interface__` or `__array__`.
There's a nice discussion here : https://stackoverflow.com/a/40380014

rcomer:
Rebased 😎

bjlittle:
@rcomer Thanks for this.... Don't quite understand why this is **still** unmerged, so I'm banking it 👍

PR #3684, opened 2020-03-13
Ugrid cf load

pp-mo:
~~**DO NOT MERGE**~~
~~Should re-target this at branch "ng-vat"~~

~~Did try that, but at present 'ng-vat' is stale, wanting a mergeback from master. See : #3685~~
~~**Do not merge** until that is resolved : this can then be rebased onto + re-targetted at ng-vat.~~

#3685 now merged, **this PR should be re-pointed at ng-vat**.


lib/iris/fileformats/_ugrid_cf_reader.py
```diff
@@ -0,0 +1,182 @@
...
+            i_dim, (mesh, mesh_location) = meshes_info[0]
+            cube.ugrid = CubeUgrid(
+                cube_dim=i_dim, grid=mesh, mesh_location=mesh_location
+            )
+        else:
+            # Add an empty 'cube.ugrid' to all cubes otherwise.
+            cube.ugrid = None
+        return
+
+    def __del__(self):
```
pp-mo: NOTE: 
This class has taken over the responsibility of doing this, from the `iris.fileformats.cf.CFReader` class.
So `CFReader.__init__` is now passed a open dataset object instead of a filename.
I think that's o.k. as CFReader is not used anywhere else.

The point is that the UGridCFReader holds the reference to the open nc.Dataset, so we can do extra anaylsis based on the file content withou re-opening it.  The same open Dataset is now used by this code, the gridded.pyugrid calls, and the CFReader instance.

lib/iris/fileformats/_ugrid_cf_reader.py
```diff
@@ -0,0 +1,187 @@
...
+_UGRID_LINK_PROPERTIES = [
+    "{}_coordinates".format(elem) for elem in _UGRID_ELEMENT_TYPE_NAMES
+]
+_UGRID_LINK_PROPERTIES += [
+    "{}_{}_connectivity".format(e1, e2)
+    for e1 in _UGRID_ELEMENT_TYPE_NAMES
+    for e2 in _UGRID_ELEMENT_TYPE_NAMES
+]
+
+# print('')
```
alastair-gemmell: Commented-out code

pp-mo: Ok, but we must nevertheless accept that this is all a bit POC and probably rather clunky + wonky.
Like the nasty `temp_xios_fix` kludge I just added below ...

pp-mo: Addressed by following commit https://github.com/SciTools/iris/pull/3684/commits/de0671ab1604597ea1f8ee8dd87ad8de24cff2c7 ?

alastair-gemmell: Sorry my original comment was intended to be a note to myself to come back to later during my review rather than one that I published, and yes, fair point on PoC. Thanks for the extra commit to tidy and clarify

lib/iris/tests/integration/ugrid_cf_reader/test_ugrid_load.py
```diff
@@ -0,0 +1,61 @@
...
+        loaded_cubes = CubeList(load_cubes(fp, temp_xios_fix=True))
+
+        # Here's a thing that at least works.  Just print some details.
+        # TODO: write some actual tests later...
+        print("")
+        print("All loaded cubes:")
+        print(loaded_cubes)
+
+        def select_cube_function(cube):
+            return cube.var_name == "theta"
```
lbdreyer: You never use this function, you use lambda functions instead. So you could just remove this?

pp-mo:
Thanks for some eyes on this @lbdreyer @alastair-gemmell 

I'm wondering if it is worth getting the test-data up somewhere online?
Then we could merge this + have the test functioning.

An alternative might be to have more focussed unit-tests, which could work from a synthetic file (somehow?).  But I think that is probably overkill + wasted time at this stage, No ?

lbdreyer:
> Thanks for some eyes on this @lbdreyer @alastair-gemmell
> 
> I'm wondering if it is worth getting the test-data up somewhere online?
> Then we could merge this + have the test functioning.
> 
> An alternative might be to have more focussed unit-tests, which could work from a synthetic file (somehow?). But I think that is probably overkill + wasted time at this stage, No ?

My preference would be towards using real data (not synthetic data) so we just need to get approval for a test file we can make public.
Are you happy to add the new file to iris-test-data? Or were you thinking of putting it somewhere else

pp-mo:
Well, here's a thing.
When I went to put a testfile in iris-test-data, there was already much older tiny XIOS file in there.
So I've just quickly rewritten as a real test that checks various aspects of the loaded result.
It seems to work to me, though clearly it is in a rather older format.

How do you feel about that, @lbdreyer @alastair-gemmell ?

alastair-gemmell:
> Well, here's a thing.
> When I went to put a testfile in iris-test-data, there was already much older tiny XIOS file in there.
> So I've just quickly rewritten as a real test that checks various aspects of the loaded result.
> It seems to work to me, though clearly it is in a rather older format.
> 
> How do you feel about that, @lbdreyer @alastair-gemmell ?

I suppose it depends on how different the older format is to the newer formats, balanced against the time and effort required to get a more up-to-date test file to the test repo?

pp-mo:
> how different the older format is
Well it is a valid UGRID format, and I think that maybe good enough for now.
My only slight disappointment is that this file has only one grid in it, so we are not testing all the functionality that we might be.  But as I've checked that aspect thoroughly, and it does work for now, I'm not that bothered.

In the meantime, there are a bunch of consequential errors in existing tests.
I'll fix those + see where it takes me.

pp-mo:
The remaining problems were caused by existing regridding tests that use the test file `theta_nodal_xios.nc`.
Since that is now interpreted as ugrid on loading, it messes these ones up.

https://github.com/SciTools/iris-test-data/pull/56 introduces a hacked version of that file, 
where I have renamed the defining "cf_role" property of the mesh node, so it won't be recognised as ugrid format.
The latest commit mods the tests to use that version, and they should pass once we have an updated iris-test-data to point it at ...

pp-mo:
Hang on a tick please.
Now rebasing onto ng-vat ...

pp-mo:
>  rebasing onto ng-vat ...

Done.
Enjoy ! :ice_cream: :balloon:

lib/iris/fileformats/netcdf.py
```diff
@@ -778,14 +779,17 @@ def load_cubes(filenames, callback=None):
 
     for filename in filenames:
         # Ingest the netCDF file.
-        cf = iris.fileformats.cf.CFReader(filename)
+        #         ncreader = iris.fileformats.cf.CFReader(filename)
```
alastair-gemmell: Did you mean to leave this commented-out line in? As it's proof of concept work it's fine, just checking in case you had intended to remove it

pp-mo: Not really.
Will remove :wink:

pp-mo:
Small changes.
I decided the 'private' nature of the module name was an unnecessary mistake.
Should now test out OK.

lib/iris/fileformats/ugrid_cf_reader.py
```diff
@@ -0,0 +1,195 @@
...
+                else:
+                    faces_dim_index = 0
+                faces_dim_name = faces_var.dimensions[faces_dim_index]
+                meshdims_map[faces_dim_name] = (mesh, "face")
+            if mesh.edges is not None:
+                # Work out name of edges dimension and record it.
+                edges_varname = mesh_var.edge_node_connectivity
+                edges_var = dataset.variables[edges_varname]
+                if "edge_dimension" in edges_var.ncattrs():
+                    edges_dim_index = faces_var.getncattr("face_dimension")
```
lbdreyer: Looks like this still needs fixing??

pp-mo: Sorry, I held off to avoid interrupting Travis, then forgot to push it..

lib/iris/fileformats/ugrid_cf_reader.py
```diff
@@ -0,0 +1,195 @@
...
+            if mesh.faces is not None:
+                # Work out name of faces dimension and record it.
+                faces_varname = mesh_var.face_node_connectivity
+                faces_var = dataset.variables[faces_varname]
+                if "face_dimension" in faces_var.ncattrs():
+                    faces_dim_index = faces_var.getncattr("face_dimension")
+                else:
+                    faces_dim_index = 0
+                faces_dim_name = faces_var.dimensions[faces_dim_index]
+                meshdims_map[faces_dim_name] = (mesh, "face")
```
lbdreyer: I'm struggling to make sense of this part of the code. The way it is written implies that there is optionally a `face_dimension` attribute on the `face_node_connectivity` variable (e.g.  what is called `Mesh2_face_nodes` in the example below), and the value of this is an integer.
However, from reading the ugrid documentation `face_dimension` is an (optional) attribute on the `mesh_topology` variable, with a value of a string, e.g. from the ugrid conventions:
image

lbdreyer:
```suggestion
            if mesh.faces is not None:
                # Work out name of faces dimension and record it.
                if "face_dimension" in mesh_var.ncattrs():
                    faces_dim_name = mesh_var.getncattr("face_dimension")
                else:
                    # Assume default ordering.
                    faces_varname = mesh_var.face_node_connectivity
                    faces_var = dataset.variables[faces_varname]
                    faces_dim_name = faces_var.dimensions[0]
                meshdims_map[faces_dim_name] = (mesh, "face")
```

lbdreyer: See https://github.com/SciTools/iris/pull/3684#discussion_r395281158 for a suggested change ???

pp-mo: Very sorry, you're entirely right !!
I must have skimmed the docs + then coded what I "thought" I remembered.
Thanks for suggestions, I'm just thinking about it.  Back shortly...

pp-mo: This suggestion fits neatly into the existing code, but I have some reservations as to (a) whether gridded actually handles the complications of non-standard dimensions and indexing, and (b) whether supporting it will over-complicate any subsequent code we might want to add here.

So, I'm leaning towards just ***outlawing*** any non-standard dimension order (and maybe 1-based indexing) for now.  That could make subsequent POC code much easier.
However, to *check* for this, I think it can only easily be done by not allowing the 'xxx_dimension' properties at all, because a "degenerate" setting that actually produces the default ordering is really hard to check for given the way it is coded (you would need to check the actual dimensions in all the mesh variables).

I'll just check if any of our test files use this, and if they don't I'll propose to outlaw it...

pp-mo: Oh rats, we do have it in files, so that won't work !
```
Problem in file "/project/avd/ng-vat/data/data_C4.nc": Mesh "topology:face_dimension = 'dim0'"
Problem in file "/project/avd/ng-vat/data/sample_mesh_files/data_C4.nc": Mesh "topology:face_dimension = 'dim0'"
Problem in file "/project/avd/ng-vat/data/sprint_0c_20200205/data_used_in_google_earth_demo/qrclim.ozone_L70_O70.ugrid.nc": Mesh "dynamics:face_dimension = 'dim2'"
Problem in file "/project/avd/ng-vat/data/sprint_0c_20200205/data_used_in_google_earth_demo/qrparm.orog.ugrid.nc": Mesh "dynamics:face_dimension = 'dim0'"
```

pp-mo: I'm accepting this + we'll just have to hope that supporting this won't add too much complexity.

pp-mo:
>  'boundary_node_connectivity' as a link property.

Yet another small tweak.
I reckon that this doesn't occur anywhere in the test files, though.

trexfeathers:
LGTM! 🎉

Issue #3686, opened 2020-03-16
iris.coord_categorisation.add_weekday() fails with cftime v1.1.1


TheElectricFlock:
iris.coord_categorisation.add_weekday() fails with cftime v1.1.1, the latest cftime release. Provided below is a traceback error.
``>>> import iris
``
``>>> import iris.coord_categorisation
``
``>>> from glob import glob
``
``>>> fnames = glob('/project/champ/data/cmip5/output1/ICHEC/EC-EARTH/historical/mon/atmos/Amon/r1i1p1/v20131231/tas/*.nc')
``
``>>> cube = iris.load_cube(fnames[0])``
``>>> iris.coord_categorisation.add_weekday(cube, 'time', name='weekday')``
``Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "/home/h01/jbedwell/miniconda/envs/test_environment/lib/python3.6/site-packages/iris/coord_categorisation.py", line 208, in add_weekday
    units='no_unit')
  File "/home/h01/jbedwell/miniconda/envs/test_environment/lib/python3.6/site-packages/iris/coord_categorisation.py", line 79, in add_categorised_coord
    result = category_function(from_coord, from_coord.points.ravel()[0])
  File "/home/h01/jbedwell/miniconda/envs/test_environment/lib/python3.6/site-packages/iris/coord_categorisation.py", line 207, in <lambda>
    lambda coord, x: calendar.day_abbr[_pt_date(coord, x).weekday()],
AttributeError: 'cftime._cftime.DatetimeGregorian' object has no attribute 'weekday'
``

trexfeathers:
I can confirm this is replicable when running the test suite in Travis CI. Will investigate a fix (pinning or otherwise)

PR #3682, opened 2020-03-10
New image hashes for mpl 3x2

pp-mo:
Fixes iris.test.test_plot.TestSymbols.test_cloud_cover 
failing since matplotlib 3.2.0.

See Travis error log [here](https://travis-ci.org/SciTools/iris/jobs/658323696)

Presumably due to [this change noted in mpl whatsnew](https://matplotlib.org/3.2.0/users/whats_new.html#kerning-adjustments-now-use-correct-values)

PR #3687, opened 2020-03-17
Update add_weekday for latest cftime

trexfeathers:
[cftime PR 135](https://github.com/Unidata/cftime/pull/135) has now been released ([see v1.1.0 in the changelog](https://github.com/Unidata/cftime/blob/master/Changelog)), meaning that datetimes are `cftime.datetime` by default, instead of `datetime.datetime`.

This means that the datetimes Iris uses no longer have the `weekday()` method, but instead have the `dayofwk` attribute.

Have therefore adjusted the 3 methods that were using `weekday()` to use `dayofwk`.

Note that this will not currently pass Travis CI while #3682 remains open (which is pending some other fixes to get Travis to pass).

### DIFF
--- a/lib/iris/coord_categorisation.py
+++ b/lib/iris/coord_categorisation.py
@@ -182,7 +182,7 @@ def add_day_of_year(cube, coord, name="day_of_year"):
 def add_weekday_number(cube, coord, name="weekday_number"):
     """Add a categorical weekday coordinate, values 0..6  [0=Monday]."""
     add_categorised_coord(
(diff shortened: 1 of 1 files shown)


trexfeathers:
Closes #3686

trexfeathers:
Since we now have two PR's (this and #3682), which both also need the fix from the other, we will be merging this PR while the image hash tests are still failing. #3682 can then be rebased onto the changes in this PR, giving us a fully passing suite of Travis tests :-)

PR #3689, opened 2020-03-19
Skip ngvat

pp-mo:
Define a skipper for tests broken in the NG-VAT POC.
I.E. stuff that we accept is broken, but won't fix for now.
This helps us to keep automated testing functional, and also marks the technical debt of things we might still need to fix.


lbdreyer:
Should this not point at the ng-vat branch?

pp-mo:
> Should this not point at the ng-vat branch?

Oops.  Thanks!

lbdreyer:
Thanks @pp-mo !

PR #3694, opened 2020-04-03
Update docs CubeList.extract method

lbdreyer:
This originated from a support query. The user was using cubelist.extract_strict() with a single constraint and was confused why this was returning a *cube* rather than a *cubelist*. The pertinent lines are these: 
https://github.com/SciTools/iris/blob/cbfbbb2ba9a8aff046b20e46d56810d34d6a1d11/lib/iris/cube.py#L324-L325

I think we need to highlight the fact that using strict=True and one constraint returns a single cube.


pp-mo:
I'm :+1: on what you've done here.  
But I do wish it was never written like this in the first place.  We don't tend to return different types of things in different cases + I think it is a bad practice generally.  
As it's really cheap, I'm going to [suggest we fix this for Iris 3.0]

lbdreyer:
Thanks @pp-mo !

> But I do wish it was never written like this in the first place

I do agree with this. I found the behaviour to be quite surprising

PR #3699, opened 2020-04-15
Unstructured multiple region extracts

trexfeathers:
Enhanced region-based extraction:
 - [x] Work with nodes, edges or faces
 - [x] Work with multiple regions
 - [x] "centre" slice type
 - [x] "enclose" slice type
 - [ ] "intersect" slice type
 - [ ] "trim" slice type
 - [ ] support variable node numbers in faces
 - [ ] comprehensive tests

#### "Intersect" slice type
Current intersect method checks that at least one of the face/edge's nodes lie within the region(s) specified. This doesn't pick up faces/edges that might fully enclose a region in one/both dimensions, even though the face and region would definitely be considered as intersecting each other.

First attempt did not meet requirements. A shame since it had some good features: used numpy boolean arrays, taking advantage of some assumptions about cubespheres to get good performance without resorting to a fully generalisable intersect method (e.g. shapely, geopandas). Was also able to run the centre and enclose slice types. Should try to retain these features in any other attempts.

#### "Trim" slice type
Work not yet started. "Trim" describes a mode where any faces/edges crossing region bounds are cut so that only the parts within the bounds are retained. This will require a weighting method to determine the data values that are retained for any trimmed faces/edges. Probably see area-weighted regridding for inspiration.

#### Support variable node numbers in faces
The current code assumes that faces and their nodes can be represented in a numpy array. If the number of nodes in faces varies this would require a ragged array - not possible in numpy. Need to write an alternative slower method using loops / list comprehension for this situation, while retaining the numpy method for situations when it can still be used.

#### Comprehensive tests
This is a proof of concept so questionable how thorough the tests should be, but recorded here for information:
 - [x] Different slice types
 - [x] Multiple regions
 - [x] Faces
 - [ ] Nodes (need different data)
 - [ ] Edges (need different data)
 - [ ] Errors
 - [ ] Tests for the `coords_within_regions` function

### DIFF
--- a/lib/iris/tests/integration/ugrid/test_ucube_operations.py
+++ b/lib/iris/tests/integration/ugrid/test_ucube_operations.py
@@ -138,7 +138,7 @@ class TestlatlonExtract(tests.IrisTest):
     def test_indexer(self):
         cube = load_unstructured_testcube()
         region = [-20, 60, 20, 65]
-        region_cube = latlon_extract_faces(cube, region)
+        region_cube = latlon_extract_faces(cube, region, "centre")
         self.assertIsNotNone(region_cube.ugrid)
         self.assertEqual(region_cube.ugrid.grid.mesh_name, "mesh")
         self.assertEqual(region_cube.shape, (7,))
--- a/lib/iris/tests/integration/ugrid/test_ucube_operations.py
+++ b/lib/iris/tests/integration/ugrid/test_ucube_operations.py
@@ -24,7 +24,7 @@ from iris.util.ucube_operations import (
     ugrid_subset,
     pseudo_cube,
     PseudoshapedCubeIndexer,
-    latlon_extract_faces,
+    xy_region_extract,
 )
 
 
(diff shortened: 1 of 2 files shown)


pp-mo:
> "Intersect" slice type .. 
Current intersect method checks that at least one of the face/edge's nodes lie within the region(s) specified. This doesn't pick up faces/edges that might fully enclose a region in one/both dimensions,

Just to note, if we are concerned about the correct intersection between true-latlon and great-circle bounded regions, there are other cases where a region can intersect ***without having any of its bounding points within the target region***.

For example, typically we will be determining overlap between sources cells defined in great-circle terms (like cubesphere cells), and a target lat-lon bounded region.
So, suppose my latitude range is +70N to +80N.  If I have a cubesphere-type cell with its upper corners just below the 70N line, the centre of its upper edge can easily extend over the +70N line, as in latlon space it curves to the north. 
This picture at least gives an idea of the relationship of the curvatures...
cubespere_edges_plot

trexfeathers:
> ... if we are concerned about the correct intersection between true-latlon and great-circle bounded regions ...

Point taken; right now this PoC is fully unaware of great circles, with the understanding that we _may_ have to revisit later and make things a lot cleverer, depending on other decisions.

trexfeathers:
Closing since unlikely to be a workable solution. Content is still here if we need it 🙂

PR #3702, opened 2020-05-06
check on maskedArray with core_data

BenMGeo:
Use of core_data instead of data to not realizing the data.

I had some issues with a custom aggregator that in this step got an extra dimension when computed in dask. 
I had no idea where the extra dimension came from, but the error vanished using core_data() instead. The results were not of any difference, so it should be save to use core_data().

### DIFF
--- a/lib/iris/cube.py
+++ b/lib/iris/cube.py
@@ -4144,7 +4144,7 @@ bound=(1994-12-01 00:00:00, 1998-12-01 00:00:00)
                 # Determine aggregation result data type for the aggregate-by
                 # cube data on first pass.
                 if i == 0:
-                    if ma.isMaskedArray(self.data):
+                    if ma.isMaskedArray(self.core_data()):
                         aggregateby_data = ma.zeros(
                             data_shape, dtype=result.dtype
                         )


rcomer:
Hi @BenMGeo,

is your custom aggregator implemented in `dask`?  If so, I think you need to explicitly set the [`lazy_func`](https://scitools.org.uk/iris/docs/latest/iris/iris/analysis.html?highlight=iris%20analysis#iris.analysis.Aggregator) when you instantiate it (though I admit I've never created a lazy aggregator myself!)

Then, when you pass it to `aggregated_by` on a lazy cube, you'd be in the first branch of this `if` loop, and the masked array check doesn't get called.
https://github.com/SciTools/iris/blob/f5feb2831582e6b333923c3135cce4cca1d865a3/lib/iris/cube.py#L4113

Using `ma.isMaskedArray` on a dask array doesn't look right to me, as it will return `False` even if the dask array is masked:

```python
import dask.array as da
import numpy.ma as ma

arr = da.ma.masked_array(range(5), mask=[0, 0, 1, 0, 1])
print(ma.isMaskedArray(arr))
```
```

BenMGeo:
You are perfectly right. I got lost in the aggregated_by code over hours of debugging and searching for the extra dimension and at one point it worked.

`lazy_func` keyword works like a charm. Sorry for bothering you.

Issue #3585, opened 2019-12-06
Default units


bjlittle:
Currently `iris` implements the following default `units` for its containers (ignoring `AuxCoordFactory` here, as the `units` of the resultant derived coordinate is determined from the `units` of `AuxCoordFactory` coordinate dependencies)

Class | Default Unit | Abstract
--- | :---: | :---:
`_DimensionalMetadata` | `no-unit` | :white_check_mark: 
`AncillaryVariable` | `no-unit`
`AuxCoord` | `1`
`CellMeasure` | `1`
`Coord` | `1` | :white_check_mark: 
`Cube` | `unknown`
`DimCoord` | `1`

**Are these sensible defaults?**

I think that we should answer this question as part of the MVP of `3.0.0`.

bjlittle:
Ping @SciTools/iris-devs

lbdreyer:
My initial reaction is that they should all be "unknown", like cube is, for similar reasons that the name is "unknown" unless otherwise specified.

Are there any cases where it would be reasonable for the user to expect a default unit?

bjlittle:
> Are there any cases where it would be reasonable for the user to expect a default unit?

@lbdreyer I'm not 100% confident, but I don't think so... To me it seems sensible to assume a default of `unknown`, which tells the user that there is **no metadata** available to specify what the `units` should be - and this would align with the default philosophy applied to the `Cube`.

I can't recall the history behind this decision, and I'm sure that it must have been debated at some point. It's probably goes way, way back to the origin of `iris`... perhaps @rhattersley might have some thoughts, recollections or an opinion to help clarify?

I can only assume that the default of `1` comes from the **canonical units** reference in [CF Metadata Section 3.3 Standard Names](http://cfconventions.org/Data/cf-conventions/cf-conventions-1.7/cf-conventions.html#standard-name)... dunno :confused: 

Otherwise, at the moment I'm erring on the blanket default of `unknown` for `units` everywhere...

rhattersley:
>  perhaps @rhattersley might have some thoughts, recollections or an opinion to help clarify?

IIRC it comes from section ["3.1. Units"](http://cfconventions.org/Data/cf-conventions/cf-conventions-1.7/cf-conventions.html#units) that states:
> A variable with no units attribute is assumed to be dimensionless.

[...]

Issue #3358, opened 2019-07-15
[PI] Quality Flag Handling


abooton:
Provision of support for "Quality Flags"
---
**Required work**
The relevant PRs to implement support for quality flags are below:

- [x] Include support for standard_name modifiers #3393
- [x] Decide on how to handle units and standard name modifiers #3408
- [x] Include support for Ancillary Variables #3422
- [x] Included loading/saving of Ancillary Variables (outstanding issue #3473)
- [x] Support for status flags (outstanding issue #3474)

---
**Description:**

Quality flags can be encoded in CF compliant data by using the "ancillary variable" and with "flag" values.    This convention allows for a more explicit descriptions of the "quality flags" that one would like to associate with the data.  (See snipped below for example)

According to the CF conventions:  http://cfconventions.org/Data/cf-conventions/cf-conventions-1.7/cf-conventions.html#ancillary-data

However, at present, there are a couple of small issues when creating a "flag" auxiliary coordinate, as the current implementation does allow the issues below to be correctly handled:
1) The status_flag standard name should be recognised as so if the cube name is also a standard_name.  (At present it come in as a long name.) See Appendix C, Standard Name Modifiers.
2) The ancillary_variable should be a local attribute, as it is directly associated the specific data. (i.e. like a latitude coord would be local to the data).
3) status_flags should not have units (they are described instead).  See Appendix C, Standard Name Modifiers.

""
![image](https://user-images.githubusercontent.com/19527482/61215240-2ab83800-a702-11e9-93c8-b6500c39479c.png)

The following code was used to generate the flag coordinate, the meta data for which is shown above:

![image](https://user-images.githubusercontent.com/19527482/61217403-876a2180-a707-11e9-9751-91c134516b7c.png)

lbdreyer:
Delaying this til the next Iris release (Iris 3).

Unfortunately we have run out of time to properly implement this (in the way we want to which would include an improvement to how Cube dimensional metadata is handled), so this will instead be included in the Iris 3 release. Hopefully that shouldn't be an issue as Iris 3 is only a month away!

lbdreyer:
The relevant PRs to implement support for quality flags are below:

* Include support for standard_name modifiers #3393 
* Decide on how to handle units and standard name modifiers #3408
* Include support for Ancillary Variables #3422
* Included loading/saving of Ancillary Variables (outstanding)
* Support for status flags (outstanding)

stephenworsley:
In the following line in the netCDF file:
```
[...]

PR #3705, opened 2020-05-13
Change default loading unit from "1" to "unknown"

stephenworsley:
Adresses the loading part of #3585 and the remaining issues with quality flags discussed in #3358 and #3474. A NetCDF variable without a unit will be asigned a unit of "unknown" when loaded. When this is then saved, the resulting NetCDF file will now have no unit. This is particularly important in the case of quality flags where the convention is to be represented as a unitless NetCDF variable.


abooton:
This PR is associated with improving the choice of default units, please target the default_units feature branch.

pp-mo:
@stephenworsley is this simply obsolete + should be closed ??

stephenworsley:
@pp-mo  Yeah, good call.

PR #3709, opened 2020-05-15
Change default loading unit from "1" to "unknown" (correct branch)

stephenworsley:
The same as #3705 except it is targeting a branch specifically for changes to default units. Because of this, the test is changed to no longer require the loading of ancillary variables.

### DIFF
--- a/docs/iris/src/whatsnew/contributions_3.0.0/incompatiblechange_2020-May-15_change_default_unit_loading.txt
+++ b/docs/iris/src/whatsnew/contributions_3.0.0/incompatiblechange_2020-May-15_change_default_unit_loading.txt
@@ -1,0 +1,1 @@
+* NetCDF files containing variables without units will be given a unit of "unknown" rather than "1"
--- a/docs/iris/src/whatsnew/contributions_3.0.0/incompatiblechange_2020-May-15_change_default_unit_loading.txt
+++ b/docs/iris/src/whatsnew/contributions_3.0.0/incompatiblechange_2020-May-15_change_default_unit_loading.txt
@@ -1,1 +1,1 @@
-* NetCDF files containing variables without units will be given a unit of "unknown" rather than "1"
+* When loading data from netcdf-CF files, where a variable has no "units" property, the corresponding Iris object will have "units='unknown'". Prior to Iris 3.0, these cases defaulted to "units='1'".
--- a/lib/iris/fileformats/_pyke_rules/fc_rules_cf.krb
+++ b/lib/iris/fileformats/_pyke_rules/fc_rules_cf.krb
@@ -1195,6 +1195,8 @@ fc_extras
     UD_UNITS_LON = ['degrees_east', 'degree_east', 'degree_e', 'degrees_e',
                     'degreee', 'degreese', 'degrees', 'degrees east',
                     'degree east', 'degree e', 'degrees e']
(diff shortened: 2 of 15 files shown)


lib/iris/fileformats/_pyke_rules/fc_rules_cf.krb
```diff
@@ -1651,9 +1651,9 @@ fc_extras
 
     ################################################################################
     def get_attr_units(cf_var, attributes):
-        attr_units = getattr(cf_var, CF_ATTR_UNITS, cf_units._UNIT_DIMENSIONLESS)
+        attr_units = getattr(cf_var, CF_ATTR_UNITS, cf_units._UNKNOWN_UNIT_STRING)
```
pp-mo, line 1654: I see no need here to reference private bits of cf_units !
We should set up our own global for this, with something like `UNKNOWN_UNIT_STRING = str(cf_units(None))`

docs/iris/src/whatsnew/contributions_3.0.0/incompatiblechange_2020-May-15_change_default_unit_loading.txt
```diff
@@ -0,0 +1 @@
+* NetCDF files containing variables without units will be given a unit of "unknown" rather than "1"
```
pp-mo, line 1: This needs fleshing out to make it clearer how it may affect the user.
Something like ... 
"When loading data from netcdf-CF files, if a variable has no "units" property the corresponding Iris object will now have a ".units='unknown'".  Previously, these cases defaulted to "units=1".  
This applies to cubes, coordinates, and ancilary data."

lib/iris/fileformats/_pyke_rules/fc_rules_cf.krb
stephenworsley: Would it make sense to change all the other references to private bits of cf_units to something similar? i.e. cf_units._NO_UNIT_STRING.

lib/iris/fileformats/_pyke_rules/fc_rules_cf.krb
```diff
@@ -1195,6 +1195,8 @@ fc_extras
     UD_UNITS_LON = ['degrees_east', 'degree_east', 'degree_e', 'degrees_e',
                     'degreee', 'degreese', 'degrees', 'degrees east',
                     'degree east', 'degree e', 'degrees e']
+    UNKNOWN_UNIT_STRING = str(cf_units.as_unit(None))
```
pp-mo, line 1198: A tiny point, but I think it would be better to be definite, and use `'unknown'` instead of `None`.

pp-mo: Additional:
I think, from the code here, we don't actually _need_ these to match some canonical form, or even be Unit instances, as they are only used as defaults and assigned values.
In that case, you can simply have strings, with your own choice of form, and don't need to wrap them in `as_unit` at all.

Also, looking again at cf_units, another simple alternative is to use the "unit symbol"s, 
which is perhaps clearer and more unambiguous.
With those, you can just have `"?"` for unknown, and `"-"` for no-unit,  
which at least avoids any uppercase / lowercase uncertainties.

pp-mo:
Thanks @stephenworsley !
I still wasn't totally sure of my logic there, but if the tests pass I think that shows this is OK

PR #3711, opened 2020-05-18
Unify saving behaviour of "unknown" and "no_unit"

stephenworsley:
Adresses #3394.

### DIFF
--- a/lib/iris/fileformats/netcdf.py
+++ b/lib/iris/fileformats/netcdf.py
@@ -22,6 +22,7 @@ import string
 import warnings
 
 import dask.array as da
+import cf_units
 import netCDF4
 import numpy as np
 import numpy.ma as ma
@@ -1760,7 +1761,7 @@ class Saver:
         # Add the data to the CF-netCDF variable.
         cf_var[:] = data
 
-        if not dimensional_metadata.units.is_udunits():
+        if dimensional_metadata.units.is_udunits():
             _setncattr(cf_var, "units", str(dimensional_metadata.units))
 
         if dimensional_metadata.standard_name is not None:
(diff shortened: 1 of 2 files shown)


pp-mo:
I think it might be a good idea to replace `unit not in ("no_unit", "unknown")` here with `not unit.is_udunit()`.
That does seem to have the same effect, but I'm not totally sure if it's a correct equivalent meaning, according to what we want and what cf-units means by it.
I'll report back after looking into the implementation within cf-units...

pp-mo:
> replace `unit not in ("no_unit", "unknown")` with `not unit.is_udunits()`

*After some reading up, I conclude that I think this _is_ a good idea.*

The cf_units docs for the 'category' property explain that it currently comprises types "udunints, no-units and unknown" : https://scitools.org.uk/cf-units/docs/latest/unit.html#cf_units.Unit.category

From reading the code, 
```
unit.isudunits() 
  === unit.category = _CATEGORY_UDUNIT
  === unit.ut_unit is not _ud.NULL_UNIT
```
It also has specific category methods  `is_unknown()` and `is_no_unit()`

So, I think that makes sense.
However I must admit that the case ins undermined by some confusion about what is really public API.  E.G. cf_units does *not* publish the category values, `_CATEGORY_UDUNITS` etc, despite the `Unit.category` being pubic property. etc.
I'm guessing that is just some kind of slip.  Likewise, why don't they publish '_UNKNOWN_UNIT_STRING' and the like ...

pp-mo:
Thanks for humouring me !
The more I think, the more this makes sense to me : The objects' ".units" are cf_unit objects, but not all cf_units are valid udunits : so, they only get saved if they are.

PR #3712, opened 2020-05-19
(WIP) Alternative method for assigning flags "no_unit"

stephenworsley:
An alternative approach to #3613. The reason there are multiple approaches is due to the fact that it is necessary to check the attributes of an ancillary variable to know if it is a flag. However, with the current method of loading NetCDF files, some steps are done in [netcdf.py](https://github.com/SciTools/iris/blob/master/lib/iris/fileformats/netcdf.py) (i.e. loading in attributes) and others are done in [fc_pyke_rules.krb](https://github.com/SciTools/iris/blob/master/lib/iris/fileformats/_pyke_rules/fc_rules_cf.krb) (i.e. determining units).

#3613 took the approach of having all the code determining units in the same place. This meant that attributes had to be loaded prematurely (and probably redundantly) in order to make this check. This pull request takes the approach that attributes should be loaded all at once and only when they are loaded should this check take place. The downside to this is that the unit has already been determined and will have to be overwritten. To summarize, #3613 has redundancy in the reading of NetCDF file and this pull request has redundancy in the writing iris objects.

It may well be possible to find a more elegant approach, in which case these two pull requests act as an illustration of the problem that must be overcome.

### DIFF
--- a/lib/iris/fileformats/_pyke_rules/fc_rules_cf.krb
+++ b/lib/iris/fileformats/_pyke_rules/fc_rules_cf.krb
@@ -1669,7 +1669,7 @@ fc_extras
     def get_attr_units(cf_var, attributes):
         attr_units = getattr(cf_var, CF_ATTR_UNITS, cf_units._UNIT_DIMENSIONLESS)
         if not attr_units:
-            attr_units = cf_units._UNKNOWN_UNIT_STRING
+            attr_units = '1'
 
         # Sanitise lat/lon units.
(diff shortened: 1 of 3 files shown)


pp-mo:
As explained [here](https://github.com/SciTools/iris/pull/3613#issuecomment-637712754), after long long thoughts I would opt for #3613 instead of this.
To be closed, after discuss + agree @stephenworsley ?

PR #3713, opened 2020-05-19
Change default units to "unknown" for all DimensionalMetadata

stephenworsley:
In line with suggestions from #3585, default units for all DimensionalMetadata are now "unknown".


stephenworsley:
Two things to consider based on the tests which are erroring:

1. PP loading relies on the default units in certain cases. I assume it would be appropriate to edit the PP loading to keep the same PP loading behaviour as before in most cases, however it will be worthwhile checking if this is indeed appropriate in all cases.
2. Certain aux_factories require coordinate units to be dimensionless (i.e. "1"). Coordinates made with default units will now break aux_factory creation.

stephenworsley:
Another change which will have to be considered in order to fix the tests:

What should be the default unit for coord_categorisation.add_categorised_coord()? The default is currently "1", changing the default to "unknown" is probably correct, though there might need to be some discussion as to which unit is most correct.

stephenworsley:
This line creates a coordinate in the process of collapsing, I have for the moment reverted the behaviour back to giving a unit of "1", though that may not necessarily be the most appropriate unit and this will have to be decided on:
https://github.com/SciTools/iris/blob/f5feb2831582e6b333923c3135cce4cca1d865a3/lib/iris/analysis/__init__.py#L805

pp-mo:
Thanks @stephenworsley all checks, I'm good with this now !

PR #3715, opened 2020-05-22
Cubelist extract strict

### LINKED ISSUE #3695: Bring cubelist "extract_strict" into line with other Iris methods
There's a tweak to CubeList.extract_strict which makes it return a single cube as it is, instead of return a "list of one cube".
We just [documented this more fully](https://github.com/SciTools/iris/pull/3694)
But, [as mentioned there](https://github.com/SciTools/iris/pull/3694#issuecomment-608376310) I don't personally like returning different types of object from different operations, which I think is a bit of an anti-pattern...

IIRC this is why we have provided other functions in multiple convenience versions, 
like "load or load_cube" and "cube.coords or cube.coord"  : these forms combine convenience + syntactic elegance with better clarity as to the exact function + interface.
Even more relevant, right here in CubeList, we have `merge()` / `merge_cube()` and `concatenate` / `concatenate_cube`.

In this case, there is a bit of an obstacle in the dual choice between "cube or cubes", and "strict or not".  I guess we are *not* going to have all of "extract, extract_strict, extract_cube, extract_cube_strict" ...

To be definite, I propose we should drop `extract_strict` *altogether*, 
and **replace it with `extract_cube` and `extract_cubes`**.
I think that works, and would be much more in line with the other methods.

So, for examples of usage change :
  * `extract_strict(some_constraints)`  -->  `extract_cubes(some_constraints)`.
  * `extract_strict(one_constraint)`  -->  `extract_cube(one_constraint)`.

pp-mo:
Replace "cubelist.extract(..., strict=True)" and the equivalent "cubelist.extract_strict(...)" with a newer-style API.

Closes #3695
As noted there, this is more in keeping with other facilities, and also addresses the slightly quirky return-type behaviour of the old style :  The old method would return a cubelist for 0 or 2..N constraints/results, but just a cube for exactly one constraint.

### DIFF
--- a/lib/iris/cube.py
+++ b/lib/iris/cube.py
@@ -355,7 +355,7 @@ class CubeList(list):
         if return_single_cube:
             if len(result) != 1:
                 # Practically this should never occur, as we now *only* request
-                # single cube results for 'extract_cube'.
+                # single cube result for 'extract_cube'.
                 msg = "Got {!s} cubes for constraints {!r}, expecting 1."
                 raise iris.exceptions.ConstraintMismatchError(
                     msg.format(len(result), constraints)
--- a/lib/iris/cube.py
+++ b/lib/iris/cube.py
@@ -278,9 +278,14 @@ class CubeList(list):
         **n** when filtered with **m** constraints can generate a maximum of
         **m * n** cubes.
 
+        Args:
+
+        * constraints (:class:`~iris.Constraint` or iterable of constraints):
+            A single constraint or an iterable.
+
         Keywords:
 
-        * strict - boolean
+        * strict (bool):
             If strict is True, then there must be exactly one cube which is
             filtered per constraint. Note: if a single constraint is given, a
             Cube is returned rather than a CubeList.
--- a/lib/iris/cube.py
+++ b/lib/iris/cube.py
@@ -268,7 +268,7 @@ class CubeList(list):
         # return our newly created XML string
         return doc.toprettyxml(indent="  ")
 
-    def extract(self, constraints, strict=False):
+    def extract(self, constraints):
         """
         Filter each of the cubes which can be filtered by the given
         constraints.
@@ -283,16 +283,9 @@ class CubeList(list):
         * constraints (:class:`~iris.Constraint` or iterable of constraints):
             A single constraint or an iterable.
 
-        Keywords:
-
-        * strict (bool):
-            If strict is True, then there must be exactly one cube which is
-            filtered per constraint. Note: if a single constraint is given, a
-            Cube is returned rather than a CubeList.
-
         """
         return self._extract_and_merge(
-            self, constraints, strict, merge_unique=None
+            self, constraints, strict=False, merge_unique=None
         )
 
     def extract_cube(self, constraint):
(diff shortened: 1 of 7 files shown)


pp-mo:
I don't understand why it's just stopped testing in Travis.
TBD...

lib/iris/cube.py
```diff
@@ -307,10 +341,6 @@ def _extract_and_merge(cubes, constraints, strict, merge_unique=False):
                 if sub_cube is not None:
                     cube_list.append(sub_cube)
 
-        if merge_unique is not None:
```
trexfeathers, line 310: I'm a little nervous that there may have been a future plan for this, but I suppose un-used code is always superfluous at the end of the day?

lib/iris/tests/test_merge.py
```diff
@@ -79,9 +79,7 @@ def custom_coord_callback(cube, field, filename):
         # Load slices, decorating a coord with custom attributes
         cubes = iris.load_raw(self._data_path, callback=custom_coord_callback)
         # Merge
-        merged = iris.cube.CubeList._extract_and_merge(
-            cubes, constraints=None, strict=False, merge_unique=False
-        )
+        merged = iris.cube.CubeList(cubes).merge()
```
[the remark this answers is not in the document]
pp-mo: Well.... I think that the purpose of this test function is to show that identical attributes are preserved in a merge.  
So I don't understand why the cube-private function was used here, because this isn't supposed to be testing that code, but the general merge code.  Apart from this call, `_extract_and_merge` is _only_ used by the extract methods in the Cube class.

I'm presuming this was a hack, put in when a plain merge couldn't do this  -- though it now does.
I can't prove that, because it dates to the original GitHub commit 8 years ago !

lib/iris/cube.py
pp-mo: Yes, this code has become redundant, i.e. never called, at some point.
I suspect, from the name, that `_extract_and_merge` used to be called by CubeList.merge also, but no longer is ?
I'm just happy to see the scope of this reduced, as it is all a bit confusing !!

Probably this whole implementation can be further simplified, but I just didn't want to overhaul the whole thing here+now as it would further complicate this PR.

lib/iris/tests/unit/cube/test_CubeList.py
```diff
@@ -294,6 +295,247 @@ def test_scalar_cube_data_constraint(self):
...
+            ):
+                method(constraints)
+        else:
+            result = method(constraints)
+            if expected is None:
+                self.assertIsNone(result)
+            elif isinstance(expected, Cube):
+                self.assertIsInstance(result, Cube)
+                self.assertEqual(result, expected)
+            else:
```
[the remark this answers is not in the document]
pp-mo: Ok, I think that is clearer!

lib/iris/cube.py
pp-mo: > _extract_and_merge used to be called by CubeList.merge also,

After some digging I find it goes back to just before Iris v1.0, when we replaced 'load' and 'load_strict' functions with load / load_cubes / load_raw.
https://github.com/SciTools/iris/pull/78
Prior to that, the load mechanism was exposing 'unique' and 'strict' options, and it all came through `CubeList._extract_and_merge` with the differing options.
I don't think the "merge_unique" option here has been used since then, so it is seriously ancient !

trexfeathers:
Everything has passed, LGTM, thanks @pp-mo

Issue #3333, opened 2019-06-18
More control over lazy data creation (chunking)


TomekTrzeciak:
Currently, it is not possible to control how the lazy data gets chunked. It is also not possible to change that afterwards (dask `rechunk` function does not change the original chunking, it only adds additional split/merge operations on top of it). While the default choice of chunking might be OK in some cases, in other it might be unsuitable and it would be useful to allow for user choice in this respect.

pp-mo:
Hi @TomekTrzeciak .  Thanks for suggesting.

In what operations do you want to control chunking ?

You can already pass you own pre-configured dask array to the cube constructor, or assign it into cube.data.
We also have chunking control when saving to netcdf :  [iris.fileformats.netcdf.save](https://scitools.org.uk/iris/docs/latest/iris/iris/fileformats/netcdf.html#iris.fileformats.netcdf.save) supports a 'chunksize' keyword, so that can also be passed in the kwargs of an [iris.save](https://scitools.org.uk/iris/docs/latest/iris/iris.html#iris.save) call.

TomekTrzeciak:
> In what operations do you want to control chunking ?

`iris.load` & co.

> You can already pass a pre-created dask array to the cube constructor, or assign it into cube.data.

Constructing the cube directly is not very convenient. I guess reassigning cube.data could be an option, but still rather awkward to write something like this:

```
cube = iris.load_cube(filepath)
cube.data = dask.array.from_array(netCDF4.Dataset(filepath)[cube.var_name], chunks=-1)
```

instead of just:

```
cube = iris.load_cube(filepath, chunks=-1)
```

pp-mo:
Regarding load, options will depend on the source format.

The "field-based" file formats (FF, PP, GRIB) deal only in 2D fields, and they don't have any efficient access to subregions of a 2d field (i.e. the format code can only load a whole field then extract from it).  Thus, the natural chunksize is the whole field, and I don't think there will ever be any practical use for chunking differently in those cases.

But I guess you are talking about netCDF ???
What we _currently_ have there is basically determined by [this code](https://github.com/SciTools/iris/blob/master/lib/iris/fileformats/netcdf.py#L504) and [this code](https://github.com/SciTools/iris/blob/master/lib/iris/_lazy_data.py#L108) .
In summary :
  * take a whole variable as a chunk, 
  * .. or use file chunking if set
  * .. *and* reduce the earlier dimensions if resulting chunks are impracticably large

[...]

PR #3720, opened 2020-05-27
Support loader-specific args in iris load functions.

pp-mo:
This has been [requested, to support #3647](https://github.com/SciTools/iris/pull/3647#discussion_r428610665).
Previously considered for other cases, e.g. #3333 (but we didn't proceed with that).

Note: decided *not* to implement arbtrary additional (unnamed) args here, as the already optional positional args (constraints, callback) would make that rather messy.
We could still implement `**kwargs` instead of a specific keyword, if preferred?  But I thought that might also lead to confusion..

Note: the changes reach further than I had imagined (into iris.fileformats and iris.io), but I believe this still does not introduce any breaking changes.  Nevertheless, I've marked it up for Iris 3.0

### DIFF
--- a/lib/iris/__init__.py
+++ b/lib/iris/__init__.py
@@ -240,7 +240,7 @@ else:
     _update(site_configuration)
 
 
-def _generate_cubes(uris, callback, constraints):
+def _generate_cubes(uris, callback, constraints, loader_kwargs=None):
     """Returns a generator of cubes given the URIs and a callback."""
     if isinstance(uris, (str, pathlib.PurePath)):
         uris = [uris]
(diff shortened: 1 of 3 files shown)


pp-mo:
I'm reviving this because I am again contemplating #3333 following some further experience with problem cases.
It will be needed to implement that.

Unfortunately we currently have a train wreck in the CI due to a problem with latest-version matplotlib docs.
So will come back here when that is fixed

pp-mo:
#4572 now proposes an alternative control mechanism.
I'm thinking this is actually _superior_ to load kwargs, because it groups control features in separate areas?

pp-mo:
Not currently needed / no driver
We can re-open if needed

PR #3725, opened 2020-06-02
Clarify behaviour of "wrapper" objects derived from CFVariable.

pp-mo:
I spent ages trying to fully understand this, and especially what was introduced in #213
(all that time ago!)
I think there is an excess of subtlety here, and I would have liked to remove some, but that is quite complicated !

So instead, here is an attempt to make the mechanisms clearer :  Mainly, just renaming a few things, and adding a few choice comments which I wish I'd had to go on.

Frankly, I'd like to go further and ditch the "wrapper" implementation, which allows us to use a "cf_var" (of a CFVariable-derived type) as a proxy for the underlying netcdf variable object.
( "cf_var.cf_data" -- or as I'm now calling it, "cf_var.nc_variable" ).
Unfortunately there is a lot of other code using that, in both Pyke rules ([example](https://github.com/SciTools/iris/blob/v2.4.0/lib/iris/fileformats/_pyke_rules/fc_rules_cf.krb#L1852)) and netcdf.py ([example](https://github.com/SciTools/iris/blob/v2.4.0/lib/iris/fileformats/netcdf.py#L497)), so I think that change is just too big to include with this.

### DIFF
--- a/lib/iris/fileformats/cf.py
+++ b/lib/iris/fileformats/cf.py
@@ -14,7 +14,7 @@ References:
 
 """
 
-from abc import ABCMeta, abstractmethod
+from abc import ABCMeta, abstractclassmethod
 
 from collections.abc import Iterable, MutableMapping
 import os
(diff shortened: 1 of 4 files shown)


pp-mo:
Note: lots of Mock-ist tests to be adjusted.

That's pretty mechanical, but I'll wait to see whether anyone wants these changes first.

pp-mo:
Peloton : code proposals now out-of-date
Create an issue to cover instead

PR #3734, opened 2020-06-10
Ideas to clarify documentation of common metadata

pp-mo:
**WIP**
Just to share some ideas.

I have now also pushed this as master-repo branch, so we can see [the built docs](https://scitools-docs.github.io/iris/common_metadata_docs/index.html).
***build coming shortly, I hope ? ...***

However, a lot of these are dev documentation, which is either private anyway, or possibly ought to be !

### DIFF
--- a/lib/iris/common/mixin.py
+++ b/lib/iris/common/mixin.py
@@ -229,13 +229,14 @@ class CFVariableMixin:
         as its Coords, CellMeasures or Ancillary Variables.
 
         For example:
-            >>> print(cube.signature)
+
+            >>> print(cube.signature)  #doctest:+SKIP
             CubeMetadata(standard_name=None, long_name='q', var_name=None, units=Unit('unknown'), attributes={}, cell_methods=())
-            >>> print(cube.signature._fields)
+            >>> print(cube.signature._fields)  #doctest:+SKIP
             ('standard_name', 'long_name', 'var_name', 'units', 'attributes', 'cell_methods')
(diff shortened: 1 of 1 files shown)


pp-mo:
? not clear why this should be failing tests ?
Presumably, existing problems with the cube-arithmetic FB

lib/iris/common/metadata.py
```diff
@@ -75,10 +82,33 @@ class BaseMetadata(metaclass=_NamedTupleMeta):
     """
     Container for common metadata.
 
+    This is the 'signature' of some cube component type (e.g. a cube or coord).
+    Being a 'namedtuple', it is an immutable, hashable value.
+
+    It's `_fields` property is a list of the 'signature' properties used by an
```
trexfeathers, line 88:
```suggestion
    Its `_fields` property is a list of the 'signature' properties used by an
```

lib/iris/common/metadata.py
```diff
@@ -289,7 +323,16 @@ def MetadataManagerFactory(cls, **kwargs):
         fields will default to a value of 'None'.
 
     """
+    # Restrict factory to appropriate metadata classes only.
```
trexfeathers, line 326: Why is this repeating the same section already in `__init__`?

pp-mo: Good spot.
I removed the one in the __init__ : it is only ever called from the factory function itself, so let's just check once there.

PR #3731, opened 2020-06-05
Merge "default_units" feature branch (to master)

pp-mo:
**WIP: DO NOT MERGE**
We will eventually do this...

It seems we have gained a test failure somehow 
Otherwise, nearly ready to go.

PR #3736, opened 2020-06-11
Explicitly set coords in nimrod load to have units "1"

stephenworsley:
This should fix one of the failing tests in #3731 which would change the behaviour of nimrod loading due to a change in default units upon coord creation.

### DIFF
--- a/lib/iris/fileformats/nimrod_load_rules.py
+++ b/lib/iris/fileformats/nimrod_load_rules.py
@@ -301,7 +301,9 @@ def experiment(cube, field):
     """Add an 'experiment number' to the cube, if present in the field."""
     if not is_missing(field, field.experiment_num):
         cube.add_aux_coord(
-            DimCoord(field.experiment_num, long_name="experiment_number")
+            DimCoord(
+                field.experiment_num, long_name="experiment_number", units="1"
+            )
         )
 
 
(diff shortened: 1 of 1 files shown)


stephenworsley:
It may still be worth checking if units are always explicitly set on lines:
https://github.com/SciTools/iris/blob/a07800c8d95e861072bf7ae26ceeca5c117311b7/lib/iris/fileformats/nimrod_load_rules.py#L571-L573
and
https://github.com/SciTools/iris/blob/a07800c8d95e861072bf7ae26ceeca5c117311b7/lib/iris/fileformats/nimrod_load_rules.py#L821-L823

pp-mo:
> @stephenworsley may still be worth checking

Undestood.
Is this otherwise ready to go, do you think ?

stephenworsley:
@pp-mo Yeah, I think this should be

stephenworsley:
After looking through the code a bit more, I believe the two cases I mentioned before should always set explicit units. In the first case because the dictionary set here contains units for every value:
https://github.com/SciTools/iris/blob/a07800c8d95e861072bf7ae26ceeca5c117311b7/lib/iris/fileformats/nimrod_load_rules.py#L488-L493

and in the seond case because this logic ensures that a unit is always present in the arguments:
https://github.com/SciTools/iris/blob/a07800c8d95e861072bf7ae26ceeca5c117311b7/lib/iris/fileformats/nimrod_load_rules.py#L765-L768

pp-mo:
> @stephenworsley the two cases I mentioned before should always set explicit units.

Dead right, I had just come to the conclusion myself.

I also checked all the ".add_xxx_coord(" throughout, and I believe they all check out.
So I reckon this is all good + we are done here ! :+1:

PR #3737, opened 2020-06-12
Nimrod loader: Fix bug in forecast_reference_time

MoseleyS:
Adds rounding to nearest-second to forecast_reference_time coord generation.


pp-mo:
Ok, this looks fine to me, as TIME_UNIT is always in seconds.

PR #3742, opened 2020-06-17
Remove iris-grib dependency from Iris tests.

pp-mo:
Removes iris-grib from the iris tests.
We no longer need iris-grib for *any* Iris testing - it has now all moved to iris-grib.

This includes the removal of the 'fileformat interoperability' tests `tests/integration/format_interop` 
Really, these were all testing grib saving, and have been moved to iris-grib : see SciTools/iris-grib#213 

Unfortunately GRIB support *is* still used  by one of the gallery examples, so I wasn't able to completely remove it from `.travis.yml`.
See : https://github.com/SciTools/iris/issues/3745  (to complete the job)


pp-mo:
> one of the gallery examples

That is, `docs/iris/example_code/General/polar_stereo.py`.
Frustratingly, the GRIB data this uses has a stereographic coordinate system with a 'true scale latitude', and *we don't support nonstandard true-scale-lat in cf load/save*.
So I can't replace with a netcdf version of the data, which would have been a nice solution.

pp-mo:
Hi @trexfeathers thanks for help.
Regarding your points -- I think some of that removing or fixing refs to grib/iris-grib goes a bit beyond what I was trying to achieve here, but they are all good spots so I will try + fix them.
Watch this space ...

In the meantime, if we are also close to getting #3746 in, then we can follow this up by removing all use of iris-grib.
I'll either do that here, or afterwards+separately depending how the timings fall out.

lib/iris/tests/runner/_runner.py
```diff
@@ -143,12 +143,6 @@ def run(self):
             regexp_pat,
             "--process-timeout=180",
         ]
-        try:
-            import gribapi  # noqa
-        except ImportError:
-            args.append("--exclude=^grib$")
-        if self.stop:
```
trexfeathers, line 150: You need to bring back the final 2 lines of the 6 you removed!

pp-mo: Ooops
Good spot!

pp-mo:
Sorry for slow response @trexfeathers 
Distracted elsewhere.  Hopefully that resolves it

Issue #3616, opened 2019-12-19
Constraint equality?


rcomer:
This may be a fairly niche requirement but...

Sometimes we write functions that return `iris.Constraint` instances.  It is not obvious how to write unit tests for these functions, as you can't check whether two constraints are the same.  How easy would it be to implement constraint equality?

nhsavage:
I would also like to be able to do this.

pp-mo:
At a basic level, I think this ought to be fairly simple.
I think there are only a few classes for which a basic `__eq__`  needs to be written, all in `iris._constraints`.

However I note that a local discussion on this (Yammer) has gone a bit off-piste on ways to identify equivalent lambda functions, which is pretty hard-core !!
It does point up a real problem though:  constraints on coords, attributes and names can all be provided with callable functions instead of values to match, in which case it won't be possible to do equality matching (an "identical" definition won't give something "equal").
However, I **_think_** that we could make this work for constraints based on value matching.

rcomer:
Thanks @pp-mo for giving this some thought.  For the callable-function case, I wonder if it's possible to get hold of the function from the Constraint instance.  Then our testing approach could be something like:

* Assert constraint works on time coordinate with a function.
* Given this list of datetimes, assert said function produces this expected list of booleans.

rcomer:
> list of datetimes

Or rather, a list of Cells, I guess....

pp-mo:
Hi @rcomer @nhsavage 
Does #3749 cover your needs, that prompted this ?

----
however...
> @rcomer the callable-function case,

I really don't think we can or should do this.  
A function is a black box, and we can't check whether it does the same as another function, only (at best) whether it is the same for specific test values.  But those totally depend on the cubes, and I don't think we have any sensible way of generating testpoints in the general case anyway.  And for coords, it could be using the bounds too ...
**Is this a deal-breaker ??**

rcomer:
Thanks @pp-mo

Here are the cases I know of in our group's projects.  I've had a play with the non-callable-function ones and your branch and, as far as I can see, the equality behaviour works nicely.

1. [stash_constraint](https://code.metoffice.gov.uk/trac/glosea/browser/gspp/trunk/PSChecks/check_parallel_suite.py#L30) ✔️ 

2. [fields.py](https://code.metoffice.gov.uk/trac/glosea/browser/gspp/trunk/GSDS_S/bin/shared/fields.py#L56) has
[...]

PR #3749, opened 2020-07-02
Implemented constraint equality.

pp-mo:
Aiming to address #3616

### DIFF
--- a/lib/iris/_constraints.py
+++ b/lib/iris/_constraints.py
@@ -299,20 +299,9 @@ class _CoordConstraint:
         )
 
     def __eq__(self, other):
-        # Note: the so-called 'coord_name' can in fact also be a coord or defn.
-        # However it is only *used* to select a coord, as "cube(coord_name)"
-        # (see in 'extract', below).
-        # Hence, in comparing actual coords, we must compare only the defns.
-        def _comparer(co):
-            if isinstance(co, iris.coords.Coord):
-                co = co._as_defn()
-            # NOTE: a CoordDefn can anyway only == another one, as it cannot
-            # behave identically to a simple string.
-            return co
-
         eq = (
             type(other) == _CoordConstraint
-            and _comparer(self.coord_name) == _comparer(other.coord_name)
+            and self.coord_name == other.coord_name
             and self._coord_thing == other._coord_thing
         )
         return eq
--- a/lib/iris/_constraints.py
+++ b/lib/iris/_constraints.py
@@ -131,6 +131,30 @@ class Constraint:
                 _CoordConstraint(coord_name, coord_thing)
             )
 
+    def __eq__(self, other):
+        # Equivalence is defined, but is naturally limited for any Constraints
+        # based on callables, i.e. "cube_func", or value functions for
+        # attributes/names/coords :  These can only be == if they contain the
+        # *same* callable object (i.e. same object identity).
+        eq = (
+            type(other) == Constraint
+            and self._name == other._name
+            and self._cube_func == other._cube_func
+            and self._coord_constraints == other._coord_constraints
+        )
+        # NOTE: theoretically, you could compare coord constraints as a *set*,
+        # as order should not affect matching.
+        # Not totally sure, so for now let's not.
+        return eq
+
+    def __hash__(self):
+        # We want constraints to have hashes, so they can act as e.g.
+        # dictionary keys or tuple elements.
+        # So, we *must* provide this, as overloading '__eq__' automatically
+        # disables it.
+        # Just use basic object identity.
+        return id(self)
+
     def __repr__(self):
         args = []
         if self._name:
@@ -218,6 +242,19 @@ class ConstraintCombination(Constraint):
         self.rhs = rhs_constraint
         self.operator = operator
 
+    def __eq__(self, other):
+        eq = (
+            type(other) == ConstraintCombination
+            and self.lhs == other.lhs
+            and self.rhs == other.rhs
+            and self.operator == other.operator
+        )
+        return eq
+
+    def __hash__(self):
+        # Must re-define if you overload __eq__ : Use object identity.
+        return id(self)
+
     def _coordless_match(self, cube):
         return self.operator(
             self.lhs._coordless_match(cube), self.rhs._coordless_match(cube)
@@ -260,6 +297,29 @@ class _CoordConstraint:
             self.coord_name,
             self._coord_thing,
         )
+
+    def __eq__(self, other):
+        # Note: the so-called 'coord_name' can in fact also be a coord or defn.
+        # However it is only *used* to select a coord, as "cube(coord_name)"
+        # (see in 'extract', below).
+        # Hence, in comparing actual coords, we must compare only the defns.
+        def _comparer(co):
+            if isinstance(co, iris.coords.Coord):
+                co = co._as_defn()
+            # NOTE: a CoordDefn can anyway only == another one, as it cannot
+            # behave identically to a simple string.
+            return co
+
+        eq = (
+            type(other) == _CoordConstraint
+            and _comparer(self.coord_name) == _comparer(other.coord_name)
+            and self._coord_thing == other._coord_thing
+        )
+        return eq
+
+    def __hash__(self):
+        # Must re-define if you overload __eq__ : Use object identity.
+        return id(self)
 
     def extract(self, cube):
         """
@@ -493,6 +553,17 @@ class AttributeConstraint(Constraint):
         self._attributes = attributes
         super().__init__(cube_func=self._cube_func)
 
+    def __eq__(self, other):
+        eq = (
+            type(other) == AttributeConstraint
+            and self._attributes == other._attributes
+        )
+        return eq
+
+    def __hash__(self):
+        # Must re-define if you overload __eq__ : Use object identity.
+        return id(self)
+
     def _cube_func(self, cube):
         match = True
         for name, value in self._attributes.items():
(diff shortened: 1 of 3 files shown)


lbdreyer:
Although this is technically in progress, it isn't necessarily going to go into Iris 3.2. So to avoid it cluttering the "in progress" column, I will move it to to do. We can then pick it up if we time.

wjbenfold:
If we do get to it then we'll need a rebase onto main before it can be reviewed I think

lib/iris/_constraints.py
```diff
@@ -241,6 +278,29 @@ def __repr__(self):
...
         )
 
+    def __eq__(self, other):
+        # Note: the so-called 'coord_name' can in fact also be a coord or defn.
+        # However it is only *used* to select a coord, as "cube(coord_name)"
+        # (see in 'extract', below).
+        # Hence, in comparing actual coords, we must compare only the defns.
+        def _comparer(co):
+            if isinstance(co, iris.coords.Coord):
+                co = co._as_defn()
```
stephenworsley, line 288: This will have to become `co.metadata` after rebasing.

stephenworsley (changes requested):
I'm happy with this approach. I think the important thing here is that no two constraints with different behaviour can show up as equal.

I think the only changes needed here are changing references from `defn` to `metadata`.

trexfeathers:
@pp-mo are you able to finish this one off before getting back into #5016? Sounds like this is pretty close.

pp-mo:
N.B. put in draft temporarily

It is *not* good to go, since the auto-merged result is still using Coord._as_defn()
And we now want to use common metadata instead.

pp-mo:
**Code now simplified ...**
On closer inspection, I conclude that the 'coord_name' of a _CoordConstraint is only ever intended to be a string.
The only way it can be otherwise is if you construct `Constraint(coord_values={something:value})`.  But whereas my earlier code suggested that 'something' could be a CoordDefn (or, now, a metadata), it actually can't since that is not a hashable type, so cannot act as a dict key.
It is still just about "possible" that you could pass a Coord instance as the identifier in this case, but we have no tests for it.  
I simply don't believe it is useful.

pp-mo:
@trexfeathers I think this is good to go.   Stephen's comment with change request is outstanding, but I think you can dismiss that if you think it is now OK.

trexfeathers:
Thanks @pp-mo! I'll let @stephenworsley take it over the line.

lib/iris/tests/unit/constraints/test_Constraint_equality.py
```diff
@@ -0,0 +1,266 @@
...
+        c2 = NameConstraint(STASH="m01s02i003", long_name=lambda n: True)
+        self.assertNotEqual(c1, c2)
+
+
+class Test___hash__(tests.IrisTest):
+    def test_empty(self):
+        c1 = Constraint() & Constraint()
+        c2 = Constraint() & Constraint()
+        self.assertEqual(hash(c1), hash(c1))
+        self.assertNotEqual(hash(c1), hash(c2))
```
stephenworsley, line 244: I think it would also be worth testing the behaviour of two different combinations of an identical pair of constraints. i.e.
```python
c1 = Constraint()
c2 = Constraint()
cc1 = c1 & c2
cc2 = c1 & c2
```

lib/iris/tests/unit/constraints/test_Constraint_equality.py
```diff
@@ -0,0 +1,266 @@
...
+        self.assertEqual(c1, c2)
+
+    def test_funcs_differ(self):
+        # Identical but different funcs do not match.
+        c1 = Constraint(cube_func=lambda c: False)
+        c2 = Constraint(cube_func=lambda c: False)
+        self.assertNotEqual(c1, c2)
+
+    def test_coord_names_same(self):
+        c1 = Constraint(coord=3)
```
stephenworsley, line 70: I think it would be clearer to replace instances the coord name "coord" with something like "latitude". This makes it clear that the keyword represents the name of the coord and is not some special keyword like `coord_values`

lib/iris/tests/unit/constraints/test_Constraint_equality.py
```diff
@@ -0,0 +1,266 @@
+# Copyright Iris contributors
+#
+# This file is part of Iris and is released under the LGPL license.
+# See COPYING and COPYING.LESSER in the root of the repository for full
+# licensing details.
+"""Unit tests for equality testing of different constraint types."""
+
+from __future__ import absolute_import, division, print_function
+
+from six.moves import filter, input, map, range, zip  # noqa
```
stephenworsley, line 10: I believe these imports are no longer necessary.

stephenworsley (changes requested):
Looking good, just a few minor changes and a whatsnew entry and I think this should be good to go

lib/iris/tests/unit/constraints/test_Constraint_equality.py
```diff
@@ -0,0 +1,266 @@
...
+
+    def test_multi_components_same(self):
+        c1 = Constraint("a") & Constraint(b=1)
+        c2 = Constraint("a") & Constraint(b=1)
+        self.assertEqual(c1, c2)
+
+    def test_multi_components_diff(self):
+        c1 = Constraint("a") & Constraint(b=1, c=2)
+        c2 = Constraint("a") & Constraint(b=1)
+        self.assertNotEqual(c1, c2)
```
stephenworsley, line 262: I think it's also worth testing the behaviour of constraint combination for the same arguments combined in a different order, i.e.
```python
c1 = Constraint("a")
c2 = Constraint("b")
cc1 = c1 & c2
cc2 = c2 & c1
```

lib/iris/tests/unit/constraints/test_Constraint_equality.py
pp-mo: :+1:

lib/iris/tests/unit/constraints/test_Constraint_equality.py
pp-mo: :+1:

pp-mo: :+1:

lib/iris/tests/unit/constraints/test_Constraint_equality.py
pp-mo: :+1:

pp-mo:
Thanks @stephenworsley 
I hope this latest addresses all the points raised.
Sorry for the delay !

lib/iris/tests/unit/constraints/test_Constraint_equality.py
```diff
@@ -67,38 +63,38 @@ def test_funcs_differ(self):
...
-        c2 = Constraint(coord2=3)
+        c1 = Constraint(coordname_A=3)
+        c2 = Constraint(coordname_B=3)
         self.assertNotEqual(c1, c2)
 
     def test_coord_values_differ(self):
-        c1 = Constraint(coord=3)
-        c2 = Constraint(coord=4)
+        c1 = Constraint(coordname=3)
+        c2 = Constraint(coordname=4)
```
stephenworsley, line 77: This is a bit of a nitpick, but I think this would be better as `some_coordname` like above.

lib/iris/tests/unit/constraints/test_Constraint_equality.py
```diff
@@ -67,38 +63,38 @@ def test_funcs_differ(self):
...
         self.assertNotEqual(c1, c2)
 
     def test_coord_values_functions_same(self):
         def func(coord):
             return False
 
-        c1 = Constraint(coord=func)
-        c2 = Constraint(coord=func)
+        c1 = Constraint(coordname=func)
+        c2 = Constraint(coordname=func)
```
stephenworsley, line 91: As above, replace with `some_coordname`.

lib/iris/tests/unit/constraints/test_Constraint_equality.py
```diff
@@ -67,38 +63,38 @@ def test_funcs_differ(self):
...
         self.assertEqual(c1, c2)
 
     def test_coord_values_functions_differ(self):
         # Identical functions are not the same.
-        c1 = Constraint(coord=lambda c: True)
-        c2 = Constraint(coord=lambda c: True)
+        c1 = Constraint(coordname=lambda c: True)
+        c2 = Constraint(coordname=lambda c: True)
```
stephenworsley, line 97: As above, replace with `some_coordname`.

stephenworsley (changes requested):
Looking good, just some very minor changes and I think we're good to go.

pp-mo:
Thanks @stephenworsley
I think I've now fixed them all to be the same.

stephenworsley (approved):
Looks good!

stephenworsley (changes requested):
I forgot to mention before, but this still needs a whatsnew entry.

pp-mo:
> I forgot to mention before, but this still needs a whatsnew entry.

Whoops sorry forgot.
Doing it now ...

pp-mo:
Ok, see if that suits @stephenworsley  ?

pp-mo:
Hang on now I need to rebase/resolve...

pp-mo:
Ok, rebased with added whatsnew
 - please see if that suits, @stephenworsley

stephenworsley (approved):
Looks good!

pp-mo:
Thanks @stephenworsley for going the distance !

rcomer:
Thanks guys! 😀

PR #3753, opened 2020-07-15
fix travis-ci push-built-docs

bjlittle:
This PR makes a minor change to fix `travis-ci` push-built-docs job.


bjlittle:
@tkknight Awesome, thanks! :smile:

PR #3757, opened 2020-07-20
Unpin cftime

### LINKED ISSUE #3751: Unpin cftime
The `cftime` package has been pinned to `1.1.3` as the lastest offering `1.2.0` is causing epoch unit test failures, see https://github.com/SciTools/iris/pull/3750

We require to investigate and unpin asap. Either:

- Fix `iris`, or
- Raise an issue with the `cftime` developers

stephenworsley:
Closes #3751.

### DIFF
--- a/requirements/core.txt
+++ b/requirements/core.txt
@@ -6,7 +6,7 @@
 cartopy>=0.12
 #conda: proj4<6
 cf-units>=2
-cftime==1.1.3
+cftime
 dask[array]>=2  #conda: dask>=2
 matplotlib<3.3
 netcdf4


stephenworsley:
There still seem to be a few tests failing. I'm not sure if these are due to changes in cftime. The one i've looked into is `system_test_supported_filetypes` in `iris.tests.system_test.SystemInitialTest` where it seems that time point is now being loaded from pp as an integer rather than a float. While I haven't found the ultimate cause for this change, I have found a part of the Iris code which seems like it could be responsible for such a discrepancy.
https://github.com/SciTools/iris/blob/9bc92f78896f216fe4afd5d59c9584cf51c73681/lib/iris/fileformats/pp_load_rules.py#L627-L649
In the above code, `round()` will give an int while `np.vectorize(round)()` will preserve type. There will therefore be a discrepancy on load depending on dimensionality.

bjlittle:
@stephenworsley Awesome, thanks for following through on this, and raising this issue with the `cftime` community (https://github.com/Unidata/cftime/issues/182), who responded with at `1.2.1` patch fix, see their [changlog](https://github.com/Unidata/cftime/blob/v1.2.1rel/Changelog). 

:+1: :tada:

PR #3758, opened 2020-07-22
Numpy rounding fix

stephenworsley:
As mentioned in #3757, there is a test currently failing due to a number being rounded to an `int` rather than a `float`. It turns out this is not due to a change in cftime as previously thought, but due to a [change in numpy](https://numpy.org/doc/stable/release/1.19.0-notes.html#change-output-of-round-on-scalars-to-be-consistent-with-python).
Matplotlib is also being pinned here to get a separate test to pass.

### DIFF
--- a/requirements/core.txt
+++ b/requirements/core.txt
@@ -8,7 +8,7 @@ cartopy>=0.12
 cf-units>=2
 cftime==1.1.3
 dask[array]>=2  #conda: dask>=2
-matplotlib==1.2.2
+matplotlib<1.3
 netcdf4
 numpy>=1.14
 scipy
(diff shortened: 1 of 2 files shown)


bjlittle:
Awesome thanks @stephenworsley

requirements/core.txt
```diff
@@ -8,7 +8,7 @@ cartopy>=0.12
 cf-units>=2
 cftime==1.1.3
 dask[array]>=2  #conda: dask>=2
-matplotlib
+matplotlib<3.3
```
bjlittle, line 11: @stephenworsley Happy with the temporary pin here, and we can address these test failures seperately

Issue #3744, opened 2020-06-18
In gallery consider moving ORCA plotting example from general to oceanography


jonseddon:
In #3710 it was suggested that @abooton is reviewing the documentation and galley examples. As part of this, is it worth considering moving the ORCA plotting example (https://scitools.org.uk/iris/docs/latest/examples/General/orca_projection.html) from the General section to the Oceanography section of https://scitools.org.uk/iris/docs/latest/gallery.html so that it's easier for people working with ocean data to find?

tkknight:
The new Read The Docs style documentation has now been merged to master (not released), see https://github.com/SciTools/iris/pull/3752

You can see the gallery via:  https://scitools-iris.readthedocs.io/en/latest/generated/gallery/index.html

I am happy to get this moved, I will create a PR.

jonseddon:
Amazing, thanks to everyone for doing that!

PR #3761, opened 2020-07-23
moved to oceanography section of gallery

tkknight:
Move an orca plot to the more appropriate section of the gallery.

Ref #3744


bjlittle:
@tkknight Awesome, thanks 👍

PR #3764, opened 2020-07-28
Scripted Logo Generation

trexfeathers:
Something I've been working on in my spare time for fun (don't judge)... This PR proposes generating Iris' logos via a script, rather than holding fixed image files in the repository. **I appreciate integrating this will be a lower priority given our current projects.**

For an idea of what the script can generate: [see the comment below](https://github.com/SciTools/iris/pull/3764#issuecomment-674581736) or run the script yourself 🙂

### Advantages over existing method

1. No reliance on fixed files that are created using un-traceable methods (typically via a GUI graphics program)
2. Script is fully visible to version control
3. Image configuration easily adjustable (colours, transparency, gradients etc.)
Iris' logo is made from relatively few elements with mostly simple configurations
4. Title image includes parameterised text - particularly useful if we continue to modularise Iris and need an matching title logo for each package

5. SVG format renders crisply at any resolution
Also doesn't blur in PDF's, unlike rasters (even high def ones)
6. SVG format means very small generated file sizes (<30kB)
7. Pythonic, in line with the rest of Iris' code
8. Don't need to hold multiple versions of the same image - all generated by a single script

### Rotation

The script also generates a zip archive of rotated-earth images, and a README, that can be used to generate a rotating GIF logo - useful for publicity ([e.g. see this tweet](https://twitter.com/trexfeathers/status/1306200332691738625)). (Experiments were tried with animated SVG but those are not really designed for frame-based animation and caused high CPU demand.)

### To do

I've got as far as I can solo. Should we proceed with this, here is what I would need help with:

- [x] Adjust image configuration to match existing logo.
I did my best to reverse engineer from [Iris7_1_trim_full.png](https://github.com/SciTools/iris/blob/master/docs/iris/src/_static/Iris7_1_trim_full.png), but a few things are a bit off since I didn't have access to the original Inkscape file (see advantage number 1). I believe @abooton has the original file for the current logo?
- [x] Integrate [generate_logo.py](https://github.com/trexfeathers/iris/blob/program_logo2/docs/iris/src/_static/generate_logo.py) into the docs build as a Sphinx extension. I would need @tkknight's help to understand and achieve this.
- [x] If above integration is successful:
  - Adjust all existing logo references to point at generated logos in the built docs.
([Tephi's README is a good example of this](https://raw.githubusercontent.com/SciTools/tephi/master/README.md))
  - Remove all logo image files from the [docs src _static directory](https://github.com/SciTools/iris/tree/master/docs/iris/src/_static).


trexfeathers:
# Examples
[](https://raw.githubusercontent.com/trexfeathers/iris/program_logo_demo/docs/iris/src/_static/iris-logo.svg)
[](https://raw.githubusercontent.com/trexfeathers/iris/program_logo_demo/docs/iris/src/_static/iris-logo-title.svg)
[](https://raw.githubusercontent.com/trexfeathers/iris/program_logo_demo/docs/iris/src/_static/iris_grib-logo-title.svg)

trexfeathers:
# Rotating
[](https://raw.githubusercontent.com/trexfeathers/iris/program_logo_demo/docs/iris/src/_static/iris-logo-title-rotate.svg)

trexfeathers:
This is now ready for review, whenever there is time. It should ideally have @abooton 's review for content, and @tkknight 's review for sphinx integration.

### Proposal (left) compared to current logo (right)
* Coastlines simplified ([generate_logo.py#L103](https://github.com/SciTools/iris/blob/6f299c7c7b527416119fafd26e22514ce9144b76/docs/iris/src/sphinxext/generate_logo.py#L103-L107)), but no longer smoothed.
  This is due to the difficulty of smoothing/rounding using the script, but personally I like the blocky coastlines.
* Globe is now fully round.
  I'm told the current distortion is a mistake (rather than an attempt to simulate Earth's oblate-spheroidal shape!).
* Globe is no longer tilted anti-clockwise.
  I assumed this was a mistake, but happy to be corrected.

tkknight:
I suggest updating your branch as there have been a few documentation related changes merged to master since youre branch was created, including **conf.py**

docs/iris/src/sphinxext/generate_logo.py
```diff
@@ -0,0 +1,441 @@
...
+from os import environ
+from pathlib import Path
+from re import sub as re_sub
+from xml.etree import ElementTree as ET
+from xml.dom import minidom
+
+from cartopy import crs as ccrs, __version__ as cartopy_version
+from cartopy.feature import LAND
+from matplotlib import pyplot as plt, rcParams
+from numpy import linspace
```
QuLogic, line 24: Use standard import form.
```suggestion
import numpy as np
```

docs/iris/src/sphinxext/generate_logo.py
```diff
@@ -0,0 +1,441 @@
...
+# XML ElementTree setup.
+namespaces = {"svg": "http://www.w3.org/2000/svg"}
+ET.register_namespace("", namespaces["svg"])
+
+################################################################################
+# Create new SVG elements for logo.
+
+# The elements that are displayed (not just referenced).
+# Order is important for layering.
+artwork_dict = OrderedDict.fromkeys(["background", "glow", "sea", "land",])
```
QuLogic, line 76: Python 2 support was dropped, wasn't it? The normal `dict` would be ordered anyway.

docs/iris/src/sphinxext/generate_logo.py
```diff
@@ -0,0 +1,441 @@
...
+    projection_rotated = ccrs.Orthographic(central_longitude=lon,
+                                           central_latitude=central_latitude)
+
+    fig = plt.figure(0, figsize=(figure_inches,) * 2)
+    ax = plt.axes(projection=projection_rotated)
+    # TODO: remove version check Iris is pinned appropriately.
+    if cartopy_version < "0.18":
+        fig.canvas.draw()
+    # Use constants set earlier to achieve desired dimensions.
+    plt.tight_layout(pad=pad_inches * inch_font_ratio)
```
QuLogic, line 135: It would be safer if the figure size were correct from the beginning, if possible.

docs/iris/src/sphinxext/generate_logo.py
```diff
@@ -0,0 +1,441 @@
...
+    ax.add_feature(LAND)
+
+    # Save as SVG and extract the resultant code.
+    svg_bytes = BytesIO()
+    plt.savefig(svg_bytes, format="svg")
+    svg_mpl = ET.fromstring(svg_bytes.getvalue())
+
+    # Find land paths and convert to clip paths.
```
QuLogic, line 143: Can you not call `.set_clip_path` on the Features instead of postprocessing the SVG?

docs/iris/src/sphinxext/generate_logo.py
trexfeathers: Thanks, good to learn that's an important distinction to the community.

docs/iris/src/sphinxext/generate_logo.py
trexfeathers: It's frankly embarrassing that I didn't already know that `dict`s as now ordered!

docs/iris/src/sphinxext/generate_logo.py
trexfeathers: Cool, I've switched to using `plt.subplots_adjust` and that's removed a few hacky lines. Hopefully the new code is a more correct use.

docs/iris/src/sphinxext/generate_logo.py
trexfeathers: If this were a one line thing I'd be in favour. But from what I can see, I would need to iterate through the `Polygon`s within `LAND.geometries()` and use each one as a clip path, as well as adding something artificial to be clipped? Given I would still need to extract the `Element`(s) I wanted for the eventual SVG, I don't think this change would reduce the complexity. Happy to be corrected 🙂

trexfeathers:
> I suggest updating your branch as there have been a few documentation related changes merged to master since youre branch was created, including **conf.py**

@abooton what are your feelings on me rebasing? There's a lot of useful content in this PR so I'd be reluctant to create a new one just to get the latest changes, especially since the new PR might also stay un-merged for a week or two (during which time master will change some more).

docs/iris/src/sphinxext/generate_logo.py
QuLogic: I did not look too closely, and perhaps the display works in the reverse of what I expected. Originally, I was thinking that everything went through Matplotlib, but I guess those gradient backgrounds are pure SVG.

I think my suggestion still makes partial sense though, in that you could clip the continents by the petals, as those would seem to be much simpler shapes for clipping. The main aim to get rid of the artifacts at the transition from continent to whitespace between the petals, which partially shows the ocean colour.

trexfeathers: Yes, I had indeed hoped to do most/all the logo using Matplotlib, but the existing logo (which I am attempting to closely replicate) relies on bezier curves (the petals) and several gradients, neither of which Matplotlib writes to SVG. So in the end I've only used Cartopy/Matplotlib to provide the continents, and used relatively simple raw SVG objects for everything else.

What you've observed with the petal clipping is an artifact created by some SVG renderers - I've seen it in Edge/Chrome, but not Firefox/Inkscape/PowerPoint, and does not get bigger when zooming - the actual SVG code clips _everything_ with the petals and doesn't treat the ocean any differently to the continent.

trexfeathers:
> I suggest updating your branch as there have been a few documentation related changes merged to master since youre branch was created, including conf.py

@tkknight done

trexfeathers:
A "What's New" entry will be provided in a separate PR once this one is approved.

trexfeathers:
Closing, since this is non-essential and there is limited interest

PR #3765, opened 2020-07-30
enabled pdf creation

tkknight:
**Enabled the pdf creation on** the Read The Docs (RTD) service.

Note that we cannot create a pdf locally as RTD does some magic to do this.  We should not need to as the PDF is based upon the same source as the built html, which we can build locally to check.

Every time RTD creates a new version of the html it should also create a PDF one too.

This PR also covers:

- Fixing indention on a Makefile for pdf creation (not used at this time) - could be removed
- Purging all latex/man options in the **conf.py**

When his PR is merged if you view the documentation via https://scitools-iris.readthedocs.io/en/latest/, if you click on the **current version at the bottom of the sidebar** you will have an option to view the PDF for the active version.

### DIFF
--- a/docs/iris/Makefile
+++ b/docs/iris/Makefile
@@ -5,19 +5,6 @@ html:
         echo "make html in $$i..."; \
         (cd $$i; $(MAKE) $(MFLAGS) $(MYMAKEFLAGS) html); done
 
-pdf:
-	@for i in $(SUBDIRS); do\
-		echo "make latex in $$i.."; \
-		(cd $$i; $(MAKE) $(MFLAGS) $(MYMAKEFLAGS) latex); done
-		echo "\def\sphinxdocclass{MO_report}" > build/latex/docs.tex
-		echo "\documentclass[letterpaper,10pt,english]{MO_report}" >> build/latex/docs.tex
-		tail -n +4 build/latex/Iris.tex >> build/latex/docs.tex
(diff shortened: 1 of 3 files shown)


docs/iris/Makefile
```diff
@@ -7,16 +7,16 @@ html:
...
-	echo "\def\sphinxdocclass{MO_report}" > build/latex/docs.tex
-	echo "\documentclass[letterpaper,10pt,english]{MO_report}" >> build/latex/docs.tex
-	tail -n +4 build/latex/Iris.tex >> build/latex/docs.tex
-	sed 's/\\tableofcontents/\\tableofcontents\n\\pagenumbering\{arabic\}/' build/latex/docs.tex > build/latex/docs2.tex
-	sed 's/subsection{/section{/' build/latex/docs2.tex > build/latex/documentation.tex
-	(cd build/latex; pdflatex -interaction=scrollmode documentation.tex)
-	# call latex again to get page numbers right...
-	(cd build/latex; pdflatex -interaction=scrollmode documentation.tex);
+		echo "make latex in $$i.."; \
+		(cd $$i; $(MAKE) $(MFLAGS) $(MYMAKEFLAGS) latex); done
```
QuLogic, line 11: The `done` on the look is here; the rest shouldn't be indented.

tkknight: Good spot.  The "make pdf" has not been used in a very long time.  As the Read The Docs service can now do this I will remove the "pdf" target.

bjlittle:

PR #3767, opened 2020-07-30
Allow CFReader to exclude given variables

trexfeathers:
This PR adds flexibility to the CFReader so that variable names can be specified for exclusion.

Also refactors so that the dataset must be extracted from a file _before_ passing to CFReader, which will allow more efficient code in situations where the excluded var names are being read by opening the file ahead of instantiating CFReader.

(initial purpose is to enable an adapted CF reader in [SciTools-incubator/iris-ugrid](https://github.com/SciTools-incubator/iris-ugrid))


trexfeathers:
Replaced by #3768 
(incorrect source branch in this one)

PR #3778, opened 2020-08-06
Netcdf loader controls to support iris-ugrid.

pp-mo:
**WIP**
still needs some tests for the new functionality.

Some extra refactoring to make `iris.fileformats.netcdf.load_cubes` and `iris.fileformats.cf.CFReader` work with the new separated code in https://github.com/SciTools-incubator/iris-ugrid

### DIFF
--- a/lib/iris/fileformats/netcdf.py
+++ b/lib/iris/fileformats/netcdf.py
@@ -752,7 +752,7 @@ def _load_aux_factory(engine, cube):
         cube.add_aux_factory(factory)
 
 
-def load_cubes(filenames, callback=None):
+def load_cubes(filenames, callback=None, create_reader=None):
     """
     Loads cubes from a list of NetCDF filenames/URLs.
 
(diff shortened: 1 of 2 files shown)


lib/iris/fileformats/netcdf.py
```diff
@@ -789,6 +802,10 @@ def load_cubes(filenames, callback=None):
         for cf_var in data_variables:
             cube = _load_cube(engine, cf, cf_var, filename)
 
+            # Call a post-processing completion step for each cube, if provided.
```
pp-mo, line 805: This approach is based on a "smallest necessary changes" strategy,
which was appropriate the original "ugrid proof-of-concept" work phase, but is perhaps a bit untidy now.

The original code views the 'cf' object (actually, a CFReader) as a box of information to draw on, and the _load_cube call is in charge of the cube.
The _load_cube call "knows" about the pyke engin;  the cf data structures in the reader; and how to construct a cube from that information.
This new callback hook approach is now giving the reader a degree of "final control" over each cube, which was not there before.

alastair-gemmell: As discussed offline - happy with this pragmatic approach for now given what we're working with as a starting point. Could be refactored further at a future date if/when Pyke rules are stripped out.

PR #3779, opened 2020-08-10
Spell checker for the sphinx documentation (not enabled for now)

tkknight:
### Adopted the use of sphinxcontrib-spelling extension to check for spelling mistakes.  

A dependency named **sphinxcontrib-spelling** needs to be added to the requirements but cannot be resolved right at this moment.  

To benefit from this as is, this PR has many **spelling mistakes corrected** but the actual spelling extension is **disabled**.  Once the dependency in the requirement is in place it may enabled.

When enabled, it will allow:

- To be run using '**make spelling**' in the documentation directory 
- To be run as part of the travis-ci job (new sub task).  Will need to check it will fail the travis job is a spelling issue is identified
- A file named **spelling_allow.txt** contains a list of all allowed words that are not in the dictionary.  There are many as there are many technical terms and acronyms that are used in the documentation.


docs/iris/gallery_code/meteorology/plot_COP_1d.py
```diff
@@ -18,7 +18,7 @@
    doi:10.1007/s00382-011-1005-5.
 
    Lowe J.A., C.D. Hewitt, D.P. Van Vuuren, T.C. Johns, E. Stehfest, J-F.
-   Royer, and P. van der Linden, 2009. New Study For Climate Modeling,
+   Royer, and P. van der Linden, 2009. New Study For Climate Modelling,
```
ajdawson, line 21: Here the original spelling is correct, because this is a title of a document where the publisher uses the US spelling. Will it be possible to ignore particular instances of spelling errors? I don't know if this kind of thing would come up more generally or not. In this case the damage from spelling the title differently to the publisher is probably minimal. I don't think this is anything to worry about, but just wondered how flexible the machinery is behind the checks?

tkknight: New commit has reverted this change now.  Thanks for spotting this @ajdawson.

When the spell check is enabled there are two ways to say a word is **allowed**.  The first is to add the word to **docs/iris/src/spelling_allow.txt** (one word per line).  The second way is to have a local sphinx directive, for example::

```
.. spelling::

   thiswordisok
```

For the word **Modeling**, it is actually is already in the **spelling_allow.txt** so it should not be raised as a spelling mistake once the spelling extensions is enabled in the future.

bjlittle:
@tkknight Awesome, once we solve the dependence for `pyenchant` then e can turn this on 🤟👍

PR #3787, opened 2020-08-17
set third party default logging level

bjlittle:
This PR explicitly sets the `logging` level of third party dependencies that are causing `DEBUG` noise in `travis-ci`.

There may be another way of doing this, but this seems like a reasonable initial approach for now...

### DIFF
--- a/lib/iris/etc/logging.yaml
+++ b/lib/iris/etc/logging.yaml
@@ -33,6 +33,12 @@ loggers:
     level: INFO
     handlers: [console-func]
     propagate: no
+  matplotlib:
+    level: INFO
+  PIL:
+    level: INFO
+  urllib3:
+    level: INFO
 
 root:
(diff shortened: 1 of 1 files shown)


tkknight:
The changes in this PR has successfully reduced the increased verbosity of the tests in travis.  

The only remaining logging that appeared is for **pyugird**, see https://travis-ci.org/github/SciTools/iris/jobs/718253885#L1290

For reference, a travis build from 5 days ago shows now verbose logging: https://travis-ci.org/github/SciTools/iris/jobs/717339839#L1279

Happy to merge this as is @bjlittle and add the **pyugrid** logging change in another PR.

bjlittle:
@tkknight Go for it 👍

tkknight:
**pyugrid** is **INFO** level logging so it ok - no further changes needed.

PR #3791, opened 2020-08-19
Modify parsing of ff headers to avoid numpy.fromfile.

owena11:
Recently noticed that fieldsfile loading was causing a bottleneck in our iris usage with iris running under python3, profiling seemed to show that half of our runtime was being spent in parsing fieldsfile headers. With `numpy.fromfile` being the main sink of time within the loading path (Seems to be a more widely noted issue for python3 versions of numpy, https://github.com/numpy/numpy/issues/13319),

Created a rough timing script to show for some basic timing loading a specific field from a file 1000 times:
```python
# An quick timing script to show the effect of removing fromfile.
import iris
import time
import os

f_size = os.path.getsize('big_fields_file')
print(f"File 'big_fields_file size': {f_size/1024**3:.2f}GB")

total_time = 0
count = 1000

for i in range(count):
    if i % 100 == 0:
        print(f'Run: {i}')
    s_time = time.time()
    # Load soil temp, just happened to be the data to hand!
    cube = iris.load('big_fields_file', iris.AttributeConstraint(STASH='m01s08i223'))
    r_time = time.time() - s_time
    total_time += r_time

print(f'==========================================================')
print(f'Total time: {total_time}, Average time: {total_time/count}')
print(f'==========================================================')
```

With output when running on master vs this branch

Branch
```
(iris-dev) [aowen@vld588:/data/users/aowen/iris_timings]$ python -m cProfile -o branch.pstat  timing_script.py 
File 'big_fields_file size': 5.04GB
Run: 0
Run: 100
Run: 200
Run: 300
Run: 400
Run: 500
Run: 600
Run: 700
Run: 800
Run: 900
==========================================================
Total time: 443.8917579650879, Average time: 0.4438917579650879
==========================================================
``` 

Master
```
(iris-dev) [aowen@vld588:/data/users/aowen/iris_timings]$ python -m cProfile -o master.pstat  timing_script.py 
File 'big_fields_file size': 5.04GB
Run: 0
Run: 100
Run: 200
Run: 300
Run: 400
Run: 500
Run: 600
Run: 700
Run: 800
Run: 900
==========================================================
Total time: 520.458238363266, Average time: 0.520458238363266
==========================================================
```

In this scenario we're seeing roughly 15% improvement with this change to the load pipeline, but it seems highly variable dependant on the file system being read from. I'd expect the improvement to be much higher on lustre based file systems where the initial issue was spotted, where we saw an overall reduction of around 45% in total  runtime.

### DIFF
--- a/lib/iris/fileformats/_ff.py
+++ b/lib/iris/fileformats/_ff.py
@@ -329,7 +329,7 @@ class FFHeader:
         # Read the FF header data
         with open(filename, "rb") as ff_file:
             # typically 64-bit words (aka. int64 or ">i8")
-            header_data = np.fromfile(
+            header_data = _parse_binary_stream(
                 ff_file,
                 dtype=">i{0}".format(word_depth),
                 count=FF_HEADER_DEPTH,
@@ -351,19 +351,19 @@ class FFHeader:
                         ff_file.seek((addr[0] - 1) * word_depth, os.SEEK_SET)
                         if len(addr) == 2:
                             if elem == "integer_constants":
-                                res = np.fromfile(
+                                res = _parse_binary_stream(
                                     ff_file,
                                     dtype=">i{0}".format(word_depth),
                                     count=addr[1],
                                 )
                             else:
-                                res = np.fromfile(
+                                res = _parse_binary_stream(
                                     ff_file,
                                     dtype=">f{0}".format(word_depth),
                                     count=addr[1],
                                 )
                         elif len(addr) == 3:
-                            res = np.fromfile(
+                            res = _parse_binary_stream(
                                 ff_file,
                                 dtype=">f{0}".format(word_depth),
                                 count=addr[1] * addr[2],
@@ -695,7 +695,7 @@ class FF2PP:
                 ff_file_seek(table_offset, os.SEEK_SET)
 
                 # Read the current PP header entry from the FF LOOKUP table.
-                header_longs = np.fromfile(
+                header_longs = _parse_binary_stream(
                     ff_file,
                     dtype=">i{0}".format(self._word_depth),
                     count=pp.NUM_LONG_HEADERS,
@@ -704,7 +704,7 @@ class FF2PP:
                 if header_longs[0] == _FF_LOOKUP_TABLE_TERMINATE:
                     # There are no more FF LOOKUP table entries to read.
                     break
-                header_floats = np.fromfile(
+                header_floats = _parse_binary_stream(
                     ff_file,
                     dtype=">f{0}".format(self._word_depth),
                     count=pp.NUM_FLOAT_HEADERS,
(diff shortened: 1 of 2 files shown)


lib/iris/fileformats/_ff.py
```diff
@@ -816,6 +816,35 @@ def __iter__(self):
...
+    # There are a wide range of types supported, we just need to know the byte
+    # size of the object, so we just make sure we've go an instance of a
+    # np.dtype
+    if not isinstance(dtype, np.dtype):
+        dtype = np.dtype(dtype)
+
+    _buffer = file_like.read(count * dtype.itemsize)
+
+    # Let numpy do the heavy lifting once we've sorted the file reading.
+    array = np.frombuffer(_buffer, dtype=dtype, count=-1)
```
bjlittle, line 844: @owena11 You're effectively reading the data twice here, right? Once to read the bytes from file, and once to parse the bytes into a `numpy` array using `frombuffer`.

I don't have access to a laptop at the moment, but I'm assuming that a file pointer to a stream doesn't support the buffer protocol? I guess I'm just wondering whether there is a way to do this by only streaming through the data once? 

Perhaps that's not possible...I don't immediately know 🤔

Also, don't we need to care about byte order here?

bjlittle (changes requested):
@owena11 Awesome, thanks for submitting this PR 🤟

Just a couple of minor review suggestions, and outstanding questions to address, thanks.

lib/iris/fileformats/_ff.py
```diff
@@ -816,6 +816,35 @@ def __iter__(self):
         return pp._interpret_fields(self._extract_field())
 
 
+def _parse_binary_stream(file_like, dtype=np.float, count=-1):
+    """
+    Replacement `numpy.fromfile` due to python3 performance issues.
```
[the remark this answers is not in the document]
owena11: Thanks Bill, not overly confident with sphinx's document style so this passed me by! Changed as requested.

lib/iris/fileformats/_ff.py
owena11: Not sure if I fully understand the first question here, but as an attempt at an answer anyway.....

We _probably_ don't do two passes over the data with this implementation, It's a bit of an guess that I can go and look up the implementation details of numpy if needed, but the `np.frombuffer` has all of the information to create the _header/object_ section of a `np.array`, so the implementation of `frombuffer` probably can be as simple as creating the array object and pointing the internal reference to the data to the `_buffer` object we supply as it supports the buffer protocol, so can share the data.  

This is guess is supported by the associated [flags](https://numpy.org/doc/stable/reference/generated/numpy.ndarray.flags.html#numpy.ndarray.flags) on the array created:
```
  C_CONTIGUOUS : True
  F_CONTIGUOUS : True
  OWNDATA : False
  WRITEABLE : False
  ALIGNED : True
  WRITEBACKIFCOPY : False
  UPDATEIFCOPY : False
```

The array we create isn't writable because the `bytes` we passed are immutable. Numpy can then handle the copying if we ever need to modify the data (I don't think we ever do modify the header data in this module).

Happily I can answer more confidently for byte order! Byte order will be encoded within the `np.dtype` object, defaulting to the machine default. Throughout the calls in this module the dtype description passed specifies endianness which will be preserved when we convert to the `np.dtype object`.

lib/iris/fileformats/_ff.py
```diff
@@ -816,6 +816,35 @@ def __iter__(self):
...
+
+    Args:
+
+    * file_like - Standard python file_like object.
+
+    Kwargs:
+
+    * dtype  - Data type to be parsed out, used to work out bytes read in.
+
+    * count - The number of values expected to be generated from the parsing.
```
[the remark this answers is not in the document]
owena11: Added!

lib/iris/fileformats/_ff.py
owena11: Ah after thinking about this more the arrays generated definitely aren't being modified prior to a copy, otherwise numpy would throw a `ValueError`. However there is no performance hit in my timings for using the mutable `bytearray` rather than simple read so I've push up that as a change.

Would probably prevent annoying gremlins for anyone toughing this code in the future, with more expected behavior.

lib/iris/fileformats/_ff.py
```diff
@@ -816,6 +816,40 @@ def __iter__(self):
...
+        object and generate as many values as possible.
+
+    """
+
+    # There are a wide range of types supported, we just need to know the byte
+    # size of the object, so we just make sure we've go an instance of a
+    # np.dtype
+    if not isinstance(dtype, np.dtype):
+        dtype = np.dtype(dtype)
+
```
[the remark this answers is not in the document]
owena11: OK definitely fixed my text editor to strip white space now...

lib/iris/fileformats/_ff.py
bjlittle: @owena11 Awesome. Okay, let's roll with this... Looks good to me 👍

bjlittle:
@owena11 Nice one, thanks for taking the time to fix this, much appreciated 👍

owena11:
@bjlittle Thanks for the review 👍

Issue #3473, opened 2019-10-21
Add Support for Loading/Saving Ancillary Variable to/from NetCDF


lbdreyer:
Part of #3358

Modify the NetCDF loader/saver such that it supports AncillaryVariables.

---
### Acceptance Criteria
- [x] Added tests for loading a NetCDF file with an Ancillary Variable (that are passing) : see #3556 
- [x] Added tests for saving a cube with an AncillaryVariable to NetCDF file (that are passing)  : see #3550 
 - [x] Additional related : #3557

pp-mo:
All the criteria are met, so this is all done 
( all bar the feature-branch mergeback : https://github.com/SciTools/iris/pull/3704 )
Closing.

PR #3797, opened 2020-08-20
[FB] launch_ancils:  Update merge-base to commit 42a1964

stephenworsley:
Updating the `launch_ancils` feature branch to latest master (42a1964)

(Merged using the "Commit merge" to preserve the original)

Contributes to #3473


lib/iris/tests/unit/fileformats/netcdf/test__load_aux_factory.py
```diff
@@ -26,8 +26,10 @@ def setUp(self):
         self.ap = mock.MagicMock(units="units")
```
abooton, line 26: It is noted that this file conflicted when merged.  The resolution is essentially to retain the cube_parts variables in place of provides.
(Is not as a severe conflict than it initially appeared 👍  )

abooton:
It is noted that when the feature branch was update, there were conflicts in the following files:
requirements/core.txt   - resolved by using requirements in master (the pinning in previous commits is no longer required)
lib/iris/tests/unit/fileformats/netcdf/test__load_aux_factory.py  - these conflicts are resolved by keeping the "cube_parts".

Issue #3708, opened 2020-05-15
[PI] Improve load/create to default to "unknown"


abooton:
### Overview

As in #3585 the units should default to "unknown" when loading/creating:

### Acceptance Criteria

 - All cubes and DimensionalMetadata whose units are not otherwise specified are loaded with units of "unknown". Specifically, defaults are set for:
   - [x] Cubes
   - [x] _DimensionalMetadata
     - [x] `Coord`
     - [x] `DimCoord`
     - [x] `AuxCoord`
     - [x] `CellMeasure`
     - [x] `AncillaryVariable`

 - All cubes and DimensionalMetadata created in iris without specified units are created with units of "unknown". Specifically, defaults are set for:
   - [x] Cubes (this is already the default)
   - [x] _DimensionalMetadata
     - [x] `Coord`
     - [x] `DimCoord`
     - [x] `AuxCoord`
     - [x] `CellMeasure`
     - [x] `AncillaryVariable`

### Other useful "unit associated" items:
Correcting saving behaviour is addressed in #3394
Ancillary variables will be addressed in #3473
Improvements to flags will be addressed in #3474

---
### Context

Units can be classified into different types (Known units vs unknown vs no-units. With dimensionless units as a subtype of known units). We interpret these types as follows:

Known dimensional units:  e.g. mm 
If the data's units are known and are recognized by cf, the units are loaded accordingly, used and saved to files.  Known units are comprised of a prefix and unit, e.g mm. (See http://cfconventions.org/Data/cf-conventions/cf-conventions-1.7/cf-conventions.html#units section 3.1 for supported prefixes)

Known dimensionless units:  e.g. 1 (number of "parts")
Is applicable to data where dimensional analysis gives a "pure ratio".  The value represents the number of "parts" so is typically "1", but there are other units which are considered dimensionless such as "degrees" or "percent". Dimensionless units can also be a n arbitrary value, for example "1e-6" would indicate data is parts per million (this is similar to the concept of a prefix as mentioned above). The value will be saved to file.

"no-unit":  no-unit or no_unit
This implies that units are not appropriate for the data.  e.g. if data is a string. Data with "no-unit" is disallowed from arithmetic, such operations are considered inappropriate.
This concept is not described by the CF conventions (it is borrowed from cf_units) and therefore a unit of "no-unit" will not be saved to file. Unitless variables are acceptable in CF conventions.

"unknown":
[...]

PR #3803, opened 2020-08-25
Clarify unit handling

stephenworsley:
Given that we have clarified our position on units, as detailed in #3708, our interpretation of units should be made clear somewhere in the documentation. I have done this as part of a discussion on the functional effect units have on arithmetic within the user guide, which was also missing and is probably also worth documenting.


alastair-gemmell (approved):
Looks good @stephenworsley !  This clearly worded description will really help users understand the change.

PR #3805, opened 2020-08-26
[FB] [PI-3478] Fix resolve free flip of src/tgt mapping

bjlittle:
This PR fixes the case where free dimension flipping of the src to tgt mapping was not constrained to the case of src and tgt cubes of equal dimensionality.

Testing will follow as a separate task, as per all new code introduced for the cube arithmetic over-haul.


bjlittle:
@rcomer Awesome, thanks 🤟

PR #3806, opened 2020-08-27
[FB] [PI-3478] cube arithmetic whatsnew entries

bjlittle:
This PR adds a few `whatsnew` entries relevant to cube arithmetic.


abooton:
Looks good.  When we write some examples of cube arithmetic it might be nice to include an example of how to use the logging with it.

PR #3813, opened 2020-09-03
Legacy doc notice

tkknight:
Added a note at the top of the main index page to highlight the legacy documentation.

Minor layout tweaks to ensure readability.


bjlittle (approved):
@tkknight This will make it easier for users to round trip between the legacy documentation on `scitools.org.uk` and the `readthedocs` documentation.

PR #3815, opened 2020-09-04
Parse with packaging version

bjlittle:
This PR uses `packaging.version` to safely parse version strings.

See here for a [fuller](https://stackoverflow.com/questions/11887762/how-do-i-compare-version-numbers-in-python) (opinionated) discussion.

### DIFF
--- a/lib/iris/quickplot.py
+++ b/lib/iris/quickplot.py
@@ -13,8 +13,10 @@ See also: :ref:`matplotlib <matplotlib:users-guide-index>`.
 
 """
 
+from packaging import version
+
 import cf_units
-from matplotlib import __version__ as mpl_version
+from matplotlib import __version__ as _mpl_version
 import matplotlib.pyplot as plt
 
 import iris.config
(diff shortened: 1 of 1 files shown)


lib/iris/quickplot.py
```diff
@@ -13,8 +13,10 @@
 
 """
 
+import packaging
```
trexfeathers, line 16: Currently failing with an `AttributeError` (e.g. https://travis-ci.org/github/SciTools/iris/jobs/724152197#L1475)
Suggest switching to `from packaging import version`, and all the follow-ons from that. I prefer this syntax anyway since it's safe against such `AttributeError`s.

trexfeathers (changes requested):
Problem!

lib/iris/quickplot.py
bjlittle: @trexfeathers That'll teach me to push without running if first :laughing:

trexfeathers:
@bjlittle are you happy with us merging a PR that includes a rebase, despite @abooton 's misgivings about this?

bjlittle:
I don't see the problem myself... all I've done is used `fixup` to join two of my own commits together, and everything is atop the last commit on `master`.

I personally have no problem with rebasing.

trexfeathers:
>I don't see the problem myself... all I've done is used fixup to join two of my own commits together, and everything is atop the last commit on master.
>
>I personally have no problem with rebasing.

Understood

trexfeathers (approved):

bjlittle:

BTW For an interesting read, see [here](https://medium.com/@fredrikmorken/why-you-should-stop-using-git-rebase-5552bee4fed1)

PR #3833, opened 2020-09-09
Iris readme minimal

tkknight:
The Iris **README.md** that is displayed on the GitHub landing page https://github.com/SciTools/iris has been pruned down to the minimal needed.  That being the logo, some badges, the headline description of Iris and a link to the documentation.

This minimal approach removes some duplication to make it easier to manage.

Also updated the headline for Iris to be **Python package** instead of **Python library**


tkknight:
A separate PR will be created to update the **Iris headline** on https://scitools.org.uk/

bjlittle:
@tkknight Lovely, akin to the [dask README.md](https://github.com/dask/dask#dask). Minimalist works for me.

Particularly as I'm keen that we don't have the same information in multiple places, otherwise it's a maintenance overhead... with the high potential to go stale or out of sync. I like the *one stop shop* approach to directing everyone to the [documentation](https://scitools-iris.readthedocs.io/en/latest/) instead. Makes total sense to me 👍 

Also, the License is highlighted by GitHub in the `iris` repo **About** section, in the RHS gutter... so that's fine by me too.

PR #3842, opened 2020-09-11
[FB] [PI-3478] Add resolve doc-strings

bjlittle:
This PR adds doc-strings and doc-tests primarily for `iris.common.resolve.Resolve`

Note that, full test coverage for `Resolve` is to follow as part of the roadmap for cube arithmetic - as agreed offline with other core devs

### DIFF
--- a/lib/iris/common/resolve.py
+++ b/lib/iris/common/resolve.py
@@ -3,6 +3,13 @@
 # This file is part of Iris and is released under the LGPL license.
 # See COPYING and COPYING.LESSER in the root of the repository for full
 # licensing details.
+"""
+Provides the infrastructure to support the analysis, identification and
+combination of metadata common between two :class:`~iris.cube.Cube`
+operands into a single resultant :class:`~iris.cube.Cube`, which will be
+auto-transposed, and with the appropriate broadcast shape.
+
+"""
 
 from collections import namedtuple
 from collections.abc import Iterable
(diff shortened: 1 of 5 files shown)


lib/iris/common/resolve.py
```diff
@@ -57,12 +64,260 @@
...
+        self.lhs_cube = lhs
+        self.rhs_cube = rhs
+
+        # Determine the initial direction to map operands.
+        # This may flip for operands with equal rank, particularly after
+        # later analysis informs the decision.
+        if self.lhs_cube.ndim >= self.rhs_cube.ndim:
+            self.map_rhs_to_lhs = True
+        else:
+            self.map_rhs_to_lhs = False
```
bjlittle, line 320: Decided to make a slight infrastructure change to support the documentation of `Resolve` attribute members.

This involves pretty much taking the contents of `_init` and relocating it to `__init__` so that `sphinx` captures the doc-strings (`#:`) of `Resolve` attribute members (this **must** happen in `__init__`, otherwise `sphinx` won't document them), and ensuring any other necessary logic is placed here in `__call__`.

lib/iris/common/resolve.py
```diff
@@ -76,6 +331,8 @@ def __call__(self, lhs, rhs):
         self._metadata_mapping()
         self._metadata_prepare()
 
+        return self
+
```
bjlittle, line 335: Returning `self` supports method chaining after making a call to a `Resolve` instance, which I wanted to do in the doc-tests e.g., now supports such patterns like
```python
resolve = Resolve()
resolve(c1, c2).map_rhs_to_lhs
```
which is pretty handy for certain workflows

lib/iris/common/resolve.py
```diff
@@ -675,13 +853,16 @@ def _metadata_mapping(self):
...
         src_free = set(src_dim_coverage.dims_free) & set(
             src_aux_coverage.dims_free
         )
         tgt_free = set(tgt_dim_coverage.dims_free) & set(
             tgt_aux_coverage.dims_free
         )
-        free_flip = len(tgt_free) > len(src_free)
+        free_flip = src_cube.ndim == tgt_cube.ndim and len(tgt_free) > len(
+            src_free
+        )
```
bjlittle, line 865: This was merged to the feature-branch `cube-arithmetic-post` as part of #3805

lib/iris/common/resolve.py
```diff
@@ -775,6 +956,20 @@ def _metadata_resolve(self):
...
+        # Categorised dim, aux and scalar coordinate items local to RHS cube only.
+        self.rhs_cube_category_local = _CategoryItems(
+            items_dim=[], items_aux=[], items_scalar=[]
+        )
+        # Categorised dim, aux and scalar coordinate items common to both
+        # LHS cube and RHS cube.
+        self.category_common = _CategoryItems(
+            items_dim=[], items_aux=[], items_scalar=[]
+        )
+
```
bjlittle, line 972: Do this here, rather in the now defunct `_init`.

And we don't want to do this within `__init__`, as we want it to happen per `__call__`

lib/iris/common/resolve.py
```diff
@@ -1351,6 +1546,7 @@ def _prepare_points_and_bounds(
 
     @property
     def _src_cube(self):
+        assert self.map_rhs_to_lhs is not None
```
bjlittle, line 1549: Defensive code to ensure the `_init` restructuring to `__inti__` et al, hasn't had any side-effects that we don't know about.

But in general, it a healthy `assert` to adopt

lib/iris/common/resolve.py
```diff
@@ -1531,13 +1808,137 @@ def cube(self, data, in_place=False):
...
+            True
+            >>> resolver.map_rhs_to_lhs
+            False
 
         """
-        return self._src_cube.ndim == len(self.mapping)
+        result = None
+        if self.mapping is not None:
+            result = self._src_cube.ndim == len(self.mapping)
+        return result
```
bjlittle, line 1885: Adds a bit more rigour to deal with the case where the user attempt to determine `mapped`, but no operands have been provisioned to analyse

lib/iris/common/resolve.py
```diff
@@ -57,12 +64,260 @@
...
+        import numpy as np
+        from iris.common import Resolve
+        cube1 = iris.load_cube(iris.sample_data_path("A1B_north_america.nc"))
+        cube2 = iris.load_cube(iris.sample_data_path("E1_north_america.nc"))[0]
+        cube2.transpose()
+        cube3, cube4 = cube1, cube2
+        data = np.zeros(cube1.shape)
+        data1 = data * 10
+        data2 = data * 20
+        data3 = data * 30
```
bjlittle, line 96: Note that, the `testsetup` instances created here for this doc-string is available for all other doc-strings, thanks to the magic of `sphinx`. 

Hence, there is no need to replicate the `testsetup` directive per doc-string test.

lib/iris/common/resolve.py
```diff
@@ -57,12 +64,260 @@
...
+                  forecast_period                x              -              -
+             Scalar coordinates:
+                  forecast_reference_time: 1859-09-01 06:00:00
+                  height: 1.5 m
+             Attributes:
+                  Conventions: CF-1.5
+                  STASH: m01s03i236
+                  source: Data from Met Office Unified Model 6.05
+             Cell methods:
+                  mean: time (6 hour)
```
bjlittle, line 155: The doc-test equivalent of a picture to 1000 words... there is no need to describe what's happening to a detailed level... just show it. Works for me as a philosophy :wink:

tkknight:
thanks for the hangout @bjlittle to walk thru this.  LGTM.

bjlittle:

PR #3855, opened 2020-09-16
"What's New" entry for #3846.

trexfeathers:
### Description

"What's New" entry for #3846.

---
[Consult Iris pull request check list]( https://scitools-iris.readthedocs.io/en/latest/developers_guide/pulls.html#the-iris-check-list)

### DIFF
--- a/docs/iris/src/whatsnew/latest.rst
+++ b/docs/iris/src/whatsnew/latest.rst
@@ -268,6 +268,10 @@ This document explains the changes made to Iris for this release
   for very minor changes in `Matplotlib`_ version 3.3 (colormaps, fonts and
   axes borders). (:pull:`3762`)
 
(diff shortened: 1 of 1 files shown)


rcomer:
Oh OK, so when you said every change should have a whatsnew, you really did mean _every_ change!

trexfeathers:
> Oh OK, so when you said every change should have a whatsnew, you really did mean every change!

Feel free to disagree! I'm only comfortable with this because of the `Internal` heading, and the fact that it's at the bottom. It keeps developers informed more easily than browsing the repo, and helps demonstrate how much work does go in to keep Iris working even if a lot doesn't manifest as a tangible change.

rcomer:
No disagreement in principle, though in this case writing the whatsnew was probably more effort than moving the offending line!

PR #3856, opened 2020-09-16
Fix whatsnew api links.

pp-mo:
### Description
A really tiny improvement : I was checking out the new whatsnew format + I noticed that these API refs were subtly wrong + don't function as links.

### DIFF
--- a/docs/iris/src/whatsnew/latest.rst
+++ b/docs/iris/src/whatsnew/latest.rst
@@ -22,13 +22,14 @@ This document explains the changes made to Iris for this release
   module to provide richer meta-data translation when loading ``Nimrod`` data
   into cubes. This covers most known operational use-cases. (:pull:`3647`)
 
-* `@stephenworsley`_ improved the handling of :class:`iris.coord.CellMeasure` in
-  the statistical operations :meth:`iris.cube.Cube.collapsed`,
-  :meth:`iris.cube.Cube.aggregated_by` and
-  :meth:`iris.cube.Cube.rolling_window`. These previously removed every
-  :class:`iris.coord.CellMeasure` attached to the cube.  Now, a
-  :class:`iris.coord.CellMeasure` will only be removed if it is associated with
-  an axis over which the statistic is being run. (:pull:`3549`)
+* `@stephenworsley`_ improved the handling of
+  :class:`iris.coords.CellMeasure`\ s in the :class:`~iris.cube.Cube`
+  statistical operations :meth:`~iris.cube.Cube.collapsed`,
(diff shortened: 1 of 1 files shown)


trexfeathers:
@pp-mo I think there might be some more too. I'm just going to build some docs from this PR, go through and make a full list...

trexfeathers (changes requested):
While here, please could you also address:
* `~iris.Cube.cube.remove_coord` on L50
* L64, where trailing an `s` after `~iris.coords.CellMeasure` appears to have prevented several words rendering

trexfeathers (approved):

pp-mo:
:+1: for diligence.

I followed your suggestions.
I have also improved/corrected a few other formatting problems :  
  #. changed some `'single backticks'` for "code element" formatting (i.e. not external links) to `''double backticks''`
    * (NB backticks confuse GitHub markdown, it  don't provide a generic escape ?!?  boo, hiss :angry: )
  #. added some `~` to a description with a lot of long module paths

pp-mo:
Some ideas also occurred to me, to improve the Ancillary Data features entry...
I think this should highlight the new classes and cube access features, as well as describing the netcdf format handling.
(though it already diverts a bit to explain Flags).

But maybe I should just create another PR for that?
I am also wondering whether it would be worth making this into a 'featured change' or 'showcase feature' box,
something like file:///net/home/h05/itpp/git/iris/iris_main/docs/iris/src/_build/html/whatsnew/1.12.html#features

? What do you think @trexfeathers @bjlittle

trexfeathers:
> I think this should highlight the new classes and cube access features, as well as describing the netcdf format handling.

@pp-mo I'm afraid, having not worked on this area of the code at all, I don't have an educated view on the correct amount of detail. This entry was the one merged from the feature branch, so I had assumed that all 3 of you working on it had seen it at the time.

---

> But maybe I should just create another PR for that?

@pp-mo that would be good Git practice I think

---

> I am also wondering whether it would be worth making this into a 'featured change' or 'showcase feature' box,

@pp-mo I was imagining that something of this sort should be written once we're sure nothing else is getting merged. I'm not sure how we decide what goes into it, since none of us have touched **all** the new features.

pp-mo:
Sorry, a bunch of commits targettting the whatsnew were just merged, and it was simpler to rebase, so I did.
I think we've more-or-less done with the prior discussion, and it's pretty easy to just check the new for of the changes.
If you object, I can try a merge-back approach instead.

pp-mo:
> I should just create another PR for that?
See #3865

Issue #3848, opened 2020-09-15
Replace deprecated IndexFormatter


stephenworsley:
## 📰 Custom Issue

Matplotlib has deprecated the IndexFormatter class https://github.com/matplotlib/matplotlib/pull/16170.This is used in only one place in the Iris code:
https://github.com/SciTools/iris/blob/a2f49c3eebfa26180c6299f9cc59687ca8303a40/lib/iris/plot.py#L251-L256
A suggestion is given here for how to replace IndexFormatter with a given example https://github.com/matplotlib/matplotlib/issues/16631#issuecomment-594008777.

stephenworsley:
It seems as though Iris is setting its own underlying values in the case of string coordinates on this line:
https://github.com/SciTools/iris/blob/a2f49c3eebfa26180c6299f9cc59687ca8303a40/lib/iris/plot.py#L441
These values seem to be anticipating the behaviour of IndexFormatter and this line would have to change if we adopted the formatter described in the above link.

rcomer:
Another [comment](https://github.com/matplotlib/matplotlib/issues/16631#issuecomment-594108101) further down that issue suggests using `FixedLocator` and `FixedFormatter`.  I think that may be simpler than the `FuncFormatter` option.

bjlittle:
@stephenworsley This is the same as #3822, right? :thinking:

stephenworsley:
@bjlittle yes you're right, I hadn't spotted that one.

stephenworsley:
Closed by #3857.

PR #3857, opened 2020-09-16
Replace deprecated IndexFormatter

stephenworsley:
### Description

Addresses #3848 (and the duplicate issue #3821).

The replacement formatter is not designed to be an exact copy of IndexFormatter since it seems that IndexFormatter can be considered bugged. Specifically, IndexFormatter would add a duplicate of the label at the 0 tick to the -1 tick. This behaviour was being accounted for by our tests, which have been changed to reflect the new behaviour. The new behaviour also relies on ticks being located precisely on integers (the IndexFormatter would perform some rudimentary rounding operations). This should always be the case since the locator is `MaxNLocator(integer=True)`.

### DIFF
--- a/lib/iris/plot.py
+++ b/lib/iris/plot.py
@@ -260,6 +260,7 @@ def _string_coord_axis_tick_labels(string_axes, axes=None):
             label_dict = dict(zip(tick_locations, labels))
             label = label_dict.get(x, "")
             return label
+
         formatter = mpl_ticker.FuncFormatter(ticker_func)
         locator = mpl_ticker.MaxNLocator(integer=True)
         this_axis = getattr(ax, axis)
--- a/lib/iris/plot.py
+++ b/lib/iris/plot.py
@@ -249,16 +249,14 @@ def _string_coord_axis_tick_labels(string_axes, axes=None):
 
     ax = axes if axes else plt.gca()
     for axis, ticks in string_axes.items():
-        # Define a tick formatter, this replaces the now deprecated
-        # mpl_ticker.IndexFormatter. Note that this formatter behaves
-        # differently to IndexFormatter and will only assign labels to
-        # ticks which are located precisely on integers in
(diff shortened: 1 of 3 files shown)


lib/iris/plot.py
```diff
@@ -249,7 +249,19 @@ def _string_coord_axis_tick_labels(string_axes, axes=None):
 
     ax = axes if axes else plt.gca()
     for axis, ticks in string_axes.items():
-        formatter = mpl_ticker.IndexFormatter(ticks)
+        # Define a tick formatter, this replaces the now deprecated
+        # mpl_ticker.IndexFormatter. Note that this formatter behaves
+        # differently to IndexFormatter and will only assign labels to
```
trexfeathers, line 254: I don't think it's right to talk about a change in behaviour in a comment - that's only appropriate for the PR and the "What's New". Do you have some simple context on why `FuncFormatter` needs an input function like this?

trexfeathers (changes requested):
Thanks @stephenworsley ❤️ 

I'm satisfied that:
* We already have enough test coverage to allow for the new behaviour
  (since you had to change an expected result as part of this change, for a valid reason)
* Nothing has broken
  (since all the tests pass)

Please could you address my one comment, and I think we'll be done 🙂

lib/iris/plot.py
stephenworsley: I've tried to clarify what the function is doing by making some of the variables more descriptive. I'm hoping that this + the new comment should make it clear how FuncFormatter is using this function. (i.e. that this function takes in the location of a tick and returns a label.) I felt that making it too explicit what was going on would have made the comment a little too verbose.

trexfeathers: That's super helpful. I had been seeing `FuncFormatter` as a closed magic box but I appreciate much better what's going on now.

trexfeathers (approved):

PR #3864, opened 2020-09-17
Update black

bjlittle:
### Description
This PR bumps the version of the `black` pre-commit hook, and adds format changes to files for the new version standard.

---
[Consult Iris pull request check list]( https://scitools-iris.readthedocs.io/en/latest/developers_guide/pulls.html#the-iris-check-list)

PR #3866, opened 2020-09-17
update black version

bjlittle:
### Description
:clapper: Take two...

Encountered unexpected minor carnage in #3864, when attempting to upgrade the `black` version from `19.10b0` to `20.8b1`, so it was actually easier to close #3864 and start again :wink: 

Okay, so this was interesting... you could just leave `black` to do it's thing, and mindlessly accept what it generates. However, a wee bit of manual intervention is required in the case where `black` had previously (`19.10b0`) injected trailing `,`'s at the end of sequences et al... which is fine in itself, but when injested by `20.8b1` it reformats it in a rather hideous way e.g., this
```python
np.logical_and(mask_invert[:-1,], diffs_along_y)
```
goes to
```python
np.logical_and(
    mask_invert[
        :-1,
    ],
    diffs_along_y,
)
```
:vomiting_face: 

In response, the manual intervention involves deciding whether the previously injected `,` was necessary or not, particularly if it was causing icky reformatting to occur.

Also, I ended up making _acceptable_ but slightly _subtle_ tuples more explicit, also in order to avoid *over-blackification*:tm: (that's now an official term btw :wink:) e.g., this
```python
(2, 0, 1),
```
becomes this
```python
((2, 0, 1),)
```
Meh... I'm not overly offended by it, particularly when it stops this
```python
         self.check((3,), Index[(2, 0, 1),], [(np.array([2, 0, 1]), Ellipsis)])
```
transmuting into this little rose bud
```python
        self.check(
            (3,),
            Index[
                (2, 0, 1),
            ],
            [(np.array([2, 0, 1]), Ellipsis)],
        )
```

---
[Consult Iris pull request check list]( https://scitools-iris.readthedocs.io/en/latest/developers_guide/pulls.html#the-iris-check-list)


.flake8
```diff
@@ -4,6 +4,8 @@
 #   https://flake8.readthedocs.io/en/latest/user/error-codes.html
 #   https://pycodestyle.readthedocs.io/en/latest/intro.html#error-codes
 
+max-line-length = 80
+select = C,E,F,W,B,B950
```
bjlittle, line 8: See https://black.readthedocs.io/en/stable/the_black_code_style.html#line-length

lib/iris/common/mixin.py
```diff
@@ -7,7 +7,6 @@
 
 from collections.abc import Mapping
 from functools import wraps
-import re
```
bjlittle, line 10: `flake8` caught this... don't quite understand how it slipped though :thinking: i.e., the `re` module isn't used in `iris.common.mixin` any more

trexfeathers:
> e.g., this
> `(2, 0, 1),`
> 
> becomes this
> `((2, 0, 1),)`

That change is generally quite helpful to a Python yearling like me ❤

bjlittle:
@trexfeathers I suppose this requires a `whatsnew` entry, right?

If so, would you prefer me to tack it along with this PR, or raise it separately?

trexfeathers:
> If so, would you prefer me to tack it along with this PR, or raise it separately?

I _hope_ this will be a quick one, so just include it here. Thanks!

trexfeathers (approved):
Great, thanks for persevering @bjlittle 🖌️ ⚫

bjlittle:
@trexfeathers Awesome, thank! :+1:

rcomer:
Wow, just wow. 🤯

PR #3867, opened 2020-09-17
Update CF standard name table to v75

lbdreyer:
As part of any release, we should ideally also update the CF standard name table to the latest version.

This PR updates the table currently available in iris (v70, released 10 December 2019) to the latest version (v75, released 15 September 2020).

The latest version was only release a couple of days ago, so I suspect that it's unlikely they will do another release before we cut the rc.

### DIFF
--- a/etc/cf-standard-name-table.xml
+++ b/etc/cf-standard-name-table.xml
@@ -1,7 +1,7 @@
 <?xml version="1.0"?>
 <standard_name_table xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xsi:noNamespaceSchemaLocation="cf-standard-name-table-1.1.xsd">
-   <version_number>70</version_number>
-   <last_modified>2019-12-10T14:47:41Z</last_modified>
+   <version_number>75</version_number>
+   <last_modified>2020-09-15T14:45:31Z</last_modified>
    <institution>Centre for Environmental Data Analysis</institution>
    <contact>support@ceda.ac.uk</contact>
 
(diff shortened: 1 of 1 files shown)


etc/cf-standard-name-table.xml
```diff
@@ -9954,13 +10149,6 @@
       <description>The quantity with standard name mass_fraction_of_rainfall_falling_onto_surface_snow is the mass of rainfall falling onto snow as a fraction of the mass of rainfall falling within the area of interest. The phrase &quot;surface_snow&quot; means snow lying on the surface. Unless indicated in the cell_methods attribute, a quantity is assumed to apply to the whole area of each horizontal grid box.</description>
    </entry>
   
-   <entry id="mass_fraction_of_rain_in_air">
```
trexfeathers, line 9957: I assume this change to being an `alias` rather than an `entry` won't negatively affect [the entry in um_cf_map.py](https://github.com/SciTools/iris/blob/f1eb79d41eb232a3eeb6776ab25183a82a3f4db6/lib/iris/fileformats/um_cf_map.py#L187)?

trexfeathers (changes requested):
Thanks @lbdreyer ! Aside from my one comment, I am satisfied that any name removals/changes will have no effect on our source code, as there are no assumptions based on their existence in examples/tests.

That said, we can't make the same assumption for users - please include a "What's New" entry to make sure everyone is informed of the change 🙂

etc/cf-standard-name-table.xml
trexfeathers: Have discussed with @bjlittle , am happy that this is fine

trexfeathers (approved):

trexfeathers:
I will add the "What's New".

PR #3785, opened 2020-08-14
[PI-3478] Merge cube arithmetic feature branch

bjlittle:
This PR is the cube arithmetic feature branch merge back to master.

It represents the culmination of PR #3583, PR #3739, and PR #3774, which introduce a common metadata API, a lenient metadata framework, and a richer, fuller, more powerful cube arithmetic capability.

A known list of outstanding tasks still require to be completed once this PR lands, see PR #3774 and parent Issue #3478.

Holy cappuccino!

PR #3869, opened 2020-09-18
[FB] [PI-3478] Metadata and cube maths documentation

bjlittle:
### Description
This PR adds user guide documentation support for the common metadata API from #3785.

This PR is in draft, with the following sections still to-do:

- [x] [Metadata combination](https://my-iris.readthedocs.io/en/latest/further_topics/metadata.html#metadata-combination) section
- [x] [Metadata conversion](https://my-iris.readthedocs.io/en/latest/further_topics/metadata.html#metadata-conversion) section
- [x] [Metadata assignment](https://my-iris.readthedocs.io/en/latest/further_topics/metadata.html#metadata-assignment) section
- [x] [Lenient metadata](https://my-iris.readthedocs.io/en/latest/further_topics/lenient_metadata.html) section
- [x]  [Lenient cube arithmetic](https://my-iris.readthedocs.io/en/latest/further_topics/lenient_maths.html) section

For reviewers, the rendered changes in this PR are available on [readthedocs](https://my-iris.readthedocs.io/en/latest/further_topics/metadata.html#).

There is also a new [Futher Topics](https://my-iris.readthedocs.io/en/latest/further_topics/index.html) chapter.

---


docs/iris/src/userguide/saving_iris_cubes.rst
```diff
@@ -54,6 +56,16 @@ The :py:func:`iris.save` function passes all other keywords through to the saver
...
+.. testcleanup::
+
+    import pathlib
+    p = pathlib.Path("myfile.pp")
+    if p.exists():
+        p.unlink()
+    p = pathlib.Path("myfile.nc")
+    if p.exists():
+        p.unlink()
+
```
bjlittle, line 68: I've been meaning to fix this for years... it simply tidies up the `myfile.pp` and `myfile.nc` created by the `doctest`.

Hence there will be less `unstaged` files in your branch after running the `doctests`. Win.

rcomer (commented):
All this is nice and clear (and I love the smiley face attributes 😁), but I fear it's too much depth for the basic user.  In fact, my hope is that lenient arithmetic (and other lenient operations hopefully added in future) will make metadata a thing that most users don't need to care about because it's handled for them under the hood.

The [cylc tutorial](http://metomi.github.io/rose/2019.01.2/html/tutorial/cylc/index.html) has a Further Topics section that you can just go look at when you need those specific things.  Maybe we could do similar with an "Advanced Topics" section - _Metadata_ could go in there, and maybe the _Real and Lazy Data_ section too.

trexfeathers:
> Maybe we could do similar with an "Advanced Topics" section

@rcomer we have the [Iris Technical Papers](https://scitools-iris.readthedocs.io/en/latest/techpapers/index.html). If we decide things are too detailed for the User Guide (I haven't looked at this PR yet) then the Technical Papers would be an ideal home.

rcomer:
Thanks @trexfeathers that's a whole part of the docs I never knew existed!

docs/iris/src/userguide/metadata.rst
```diff
@@ -0,0 +1,646 @@
...
+   ``coord_system``                                            ✔                              ✔                                                                                                     ✔
+   ``climatological``                                          ✔                              ✔                                                                                                     ✔
+   ``measure``                                                                                                                           ✔
+   ``cell_methods``                                                                                                                                                        ✔
+   ``circular``                                                                                                                                                                                     ✔
+   =================== ======================================= ============================== ========================================== ================================= ======================== ==============================
+
```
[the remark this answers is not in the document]
bjlittle: Yeah, not much I can do about that... But it's a pretty valuable table, says it all really

bjlittle:
@trexfeathers and @rcomer Yup, hear what you're both saying... and to be honest it feels like I've artificially squeezed this whole chapter in not quite the right place 👍

However, I don't think it really belongs in the tech papers section either. The tech papers for me, are more for iris devs, and the detail in this chapter is too high level for that audience - although a deeper treatment of suitable implementation detail would be appropriate.

I'll have a play, but I like suggestion from @rcomer to have a Further Topics section... Which is almost like a middle ground between user guide and tech papers

bjlittle:
@rcomer We now have a [Further Topics](https://my-iris.readthedocs.io/en/latest/further_topics/index.html) section, which I think works quite well... what do you think :thinking:

rcomer:
“more advanced or curious user” - I like that 👍

lbdreyer (commented):
Good improvements for consistency sake :+1: 

As for the docs I do want to preface my comments with the acknowledgement of just how subjective "good documentation" can be. Reviewing docs is hard ( and I did find reviewing this particularly hard) since you continually butt up against the problem of "that's not how I would do it". I think we tend to write the docs we would want to read, but we all have such different opinions of how we would want the docs to be. Personally, I like things to be as concise as possible, so I don't have to invest too much time in reading docs, before I can get back to writing code and feeling productive. So it's worth bearing that in mind, since I'm going to have a tendency to want to cut things out unless they are absolutely necessary. 
I mention this so that you can take my comments with a large pinch of salt, as they are very much just my personal opinions and, no doubt, many would disagree with me. So I'm happy if you choose to ignore most of my comments included here.

That all being said, one thing I would be quite keen on addressing is the introduction to the metadata page. I find that that page reads a bit like a story in the sense of you don't know what will happen. I would prefer to include a bit of a summary of what will be covered in the introduction section. That way user's know what to expect and are more inclined to invest the time in reading it, or at least skip to the part that is relevant to them. Currently, the only way it is possible to read it is to just buckle down and read the whole thing.

bjlittle:
@lbdreyer Cool. 

Docs are hard, and can be interpreted quite differently by different people. I think that's normal. So, pinch of salt accepted, and understood :wink: 

That said, happy to add a preface that highlights what's to come, totally reasonable, and super helpful to the reader.

Personally, the difference here for me is the difference between a user guide and a reference. This isn't a reference, and I didn't intend it to be read like a reference. Akin to the rest of the docs in the user guide it's a narrative, but it highlights the different areas in the API that you should know about and may be useful to you - it's not about the best way to do things, as that really depends on what your trying to do.

Personally, I tend to skim through a user guide and then constantly hit the reference for detail, but you need somewhere to set the context and scene, and for me, that's the user guide... otherwise you don't know what to look for in the reference.

Anyways, thanks for sticking with it and wading through... all comments welcome, and I'll service them as best I can :+1:

docs/iris/src/further_topics/metadata.rst
```diff
@@ -0,0 +1,984 @@
...
+collective of individual metadata members used to define it. Almost all
+of these members reference specific `CF Conventions`_ terms. However, some
+of these members, such as :attr:`~iris.coords.DimCoord.var_name` and
+:attr:`~iris.coords.DimCoord.circular` are Iris specific terms.
+
+For example, the collective metadata used to define an
+:class:`~iris.coords.AncillaryVariable` are the ``standard_name``, ``long_name``,
+``var_name``, ``units``, and ``attributes`` members. Note that, these are the
+actual `data attribute`_ names of the actual metadata members on the actual
+Iris class.
```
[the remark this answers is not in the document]
bjlittle: It was intended to stress the fact, but if it reads clunkily, then happy to drop it

docs/iris/src/further_topics/metadata.rst
```diff
@@ -0,0 +1,984 @@
...
+What each of these **different** Iris `CF Conventions`_ classes all have in
+**common** is that **metadata** is used to define them and give them meaning.
+
+.. _metadata members:
+.. table:: - Iris classes that model `CF Conventions`_ metadata
+   :widths: auto
+   :align: center
+
+   =================== ======================================= ============================== ========================================== ================================= ======================== ============================== ===================
+   Metadata members    :class:`~iris.coords.AncillaryVariable` :class:`~iris.coords.AuxCoord` :class:`~iris.aux_factory.AuxCoordFactory` :class:`~iris.coords.CellMeasure` :class:`~iris.cube.Cube` :class:`~iris.coords.DimCoord` Metadata members
```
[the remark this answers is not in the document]
bjlittle: The table is quite wide, and I can't do much about that... this is particularly a problem on a mobile device, so scrolling to the right it is easy to lose track of what rows are for what metadata members.

Repeating the column at the end of the table help alleviate this problem IMHO

docs/iris/src/further_topics/metadata.rst
```diff
@@ -0,0 +1,984 @@
...
+Common Metadata API
+===================
+
+.. testsetup::
+
+    import iris
+    cube = iris.load_cube(iris.sample_data_path("A1B_north_america.nc"))
+
+As of Iris ``3.0.0``, a unified treatment of metadata has been applied
+across each Iris class in :numref:`metadata members` to allow users to easily
```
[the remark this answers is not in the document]
bjlittle: Took a slightly difference route to clarifying this, see what you think...

docs/iris/src/further_topics/metadata.rst
```diff
@@ -0,0 +1,984 @@
...
+
+    >>> tuple([getattr(metadata, member) for member in metadata._fields if member.endswith("name")])
+    ('longitude', None, 'longitude')
+
+Note that, `namedtuple._fields`_ is also a class method, so you don't need
+an instance to determine the members of a metadata class, e.g.,
+
+    >>> from iris.common import CubeMetadata
+    >>> CubeMetadata._fields
+    ('standard_name', 'long_name', 'var_name', 'units', 'attributes', 'cell_methods')
```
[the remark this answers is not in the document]
bjlittle: Fair point, but I think a lot of users haven't been exposed to namedtuples.

Also, if they have, then this highlights that you can easily rip apart the metadata, and iterate over it easily, rather than create your own enumerated list of members.

So, I think it has value... but that's based on speculation - that's why docs are hard as you don't know the experience level of the readership. One persons dross is another persons gold.

docs/iris/src/further_topics/metadata.rst
```diff
@@ -0,0 +1,984 @@
...
+
+    >>> longitude.metadata == longitude.metadata
+    True
+
+Or alternatively, using the ``equal`` method instead,
+
+.. doctest:: richer-metadata
+
+    >>> longitude.metadata.equal(longitude.metadata)
+    True
```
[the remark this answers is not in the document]
bjlittle: I wanted to add a hook after highlighting `equal` as an alternative to `==`, as the `equal` methods supports the `lenient` kwarg, where as that's impossible for `__eq__`... I've not added that, as I'm still writing the section on lenient metadata.

As a reminder I'll add a `todo`, which we can merge into the feature branch, but I'll ensure that all `todo`s are removed before merging to `master`.

docs/iris/src/userguide/real_and_lazy_data.rst
```diff
@@ -10,7 +10,7 @@
 
 
 ==================
-Real and Lazy Data
+Real and lazy data
```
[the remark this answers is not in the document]
bjlittle: If you care so passionately about it, then please push up a PR

I'd happily merge it :wink:

docs/iris/src/further_topics/metadata.rst
```diff
@@ -0,0 +1,985 @@
...
+For example, the collective metadata used to define an
+:class:`~iris.coords.AncillaryVariable` are the ``standard_name``, ``long_name``,
+``var_name``, ``units``, and ``attributes`` members. Note that, these are the
+actual `data attribute`_ names of the actual metadata members on the actual
+Iris class.
+
+As :numref:`metadata members table` highlights, **specific** metadata is used to
+define and represent each **specific** Iris `CF Conventions`_ class. This means
+that this **specific** metadata can then be used to easily **identify**,
+**compare** and **differentiate** between individual class instances.
```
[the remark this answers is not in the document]
bjlittle: Yeah, that is a bit `specific` tastic :rofl:

docs/iris/src/further_topics/metadata.rst
```diff
@@ -0,0 +1,985 @@
...
+    >>> longitude.attributes = {"grinning face": "🙂"}
+    >>> other = longitude.metadata._replace(attributes={"grinning face":  "🙃"})
+    >>> other
+    DimCoordMetadata(standard_name='longitude', long_name=None, var_name='longitude', units=Unit('degrees'), attributes={'grinning face': '🙃'}, coord_system=GeogCS(6371229.0), climatological=False, circular=False)
+    >>> longitude.metadata == other
+    False
+
+Note that, although the ``==`` operator and the ``equal`` method are
+functionally equivalent, the ``equal`` method also provides a means
+to support **lenient** equality, as discussed in :ref:`lenient metadata`.
```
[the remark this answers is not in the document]
bjlittle: Yup, I'll move this back to the previous section, which will kill two birds with one stone, thanks.

docs/iris/src/further_topics/metadata.rst
```diff
@@ -0,0 +1,985 @@
...
+:numref:`metadata members table`. However, there is an exception to the rule...
+
+
+.. _exception rule:
+
+Exception to the rule
+~~~~~~~~~~~~~~~~~~~~~
+
+In general, **different** metadata classes cannot be compared, however support
+is provided for comparing :class:`~iris.common.CoordMetadata` and
```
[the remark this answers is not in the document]
bjlittle: Table 1 and 2 explain what `CoordMetadata` is and where exactly it appears in Iris.

The user will end up using `CoordMetadata` at some point if they ever use the common metadata API.

docs/iris/src/further_topics/metadata.rst
```diff
@@ -0,0 +1,985 @@
...
+
+The ``difference`` method returns a :class:`~iris.common.DimCoordMetadata` instance, when
+there is **at least** one ``metadata`` member with a different value, where,
+
+- ``None`` means that there was **no** difference for the member,
+- a `tuple`_ containing the two different associated values for the member.
+
+Given our example, only the ``long_name``, ``var_name`` and ``units`` members
+have different values, as expected. Note that, the ``difference`` method **is
+not** commutative. The the order of the tuple member values is the same order
```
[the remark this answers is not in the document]
bjlittle: Great spot :eyes:

docs/iris/src/further_topics/metadata.rst
```diff
@@ -0,0 +1,985 @@
...
+between each of the associated metadata member values,
+
+- if the values are **different**, then the combined result is ``None``
+- otherwise, the combined result is the **common value**
+
+Let's reinforce this behaviour, but this time by combining metadata where the
+``attributes`` `dict`_ member is different, where,
+
+- the ``STASH`` and ``source`` keys are **missing**,
+- the ``Model scenario`` key has the same **same value**,
```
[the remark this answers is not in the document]
bjlittle: Nice one :smile:

docs/iris/src/userguide/real_and_lazy_data.rst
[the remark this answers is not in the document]
bjlittle: Phew! Although, we should do it sometime... fancy making an issue? That way we won't forget about it.

Then we can fix this for `iris` 3.1 :smile:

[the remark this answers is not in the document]
bjlittle: @lbdreyer BTW [dask](https://docs.dask.org/en/latest/) pretty much complies to the capitalise headings philosophy (which doesn't surprise me), whereas [xarray](https://xarray.pydata.org/en/latest/) is a bit all over the shop, [numpy](https://numpy.org/doc/1.19/) mostly complies, [scipy](https://docs.scipy.org/doc/scipy/reference/) kinda almost complies, and [matplotlib](https://matplotlib.org/tutorials/index.html) is a mixed bag... interesting.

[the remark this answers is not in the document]
bjlittle: @lbdreyer I'm actually going to make this change, but not in this PR... I'll do it as a follow-up action :+1:

docs/iris/src/further_topics/lenient_metadata.rst
```diff
@@ -0,0 +1,390 @@
...
+- the ``grinning face`` key is **missing**,
+- the ``neutral face`` key has the **same value**, and
+- the ``upside-down face`` key is **new**
+
+.. doctest:: lenient-behaviour
+
+    >>> attributes = {"neutral face": "😐", "upside-down face": "🙃"}
+    >>> metadata = latitude.metadata._replace(attributes=attributes)
+    >>> metadata  # doctest: +SKIP
+    DimCoordMetadata(standard_name='latitude', long_name=None, var_name='latitude', units=Unit('degrees'), attributes={'neutral face': '😜', 'upside-down face': '🙃'}, coord_system=GeogCS(6371229.0), climatological=False, circular=False)
```
stephenworsley, line 367: I don't understand where 😜 is coming from in this example? Should this be 😐 since that is what's in both `attributes` and `latitude.attributes`?

bjlittle: @stephenworsley Don't worry, my fault and awesome spot, thanks!

I was flipping on that `doctest` and forgot to update the output, so that's why it's confusing, simply because it's wrong. Since I'm using dictionaries I need to turn `doctest` off to account for dictionaries not being ordered, otherwise there is the risk of the doc-tests failing at some random point.

Pushed an update now, so that should have resolved this :+1:

docs/iris/src/further_topics/metadata.rst
```diff
@@ -0,0 +1,987 @@
...
+within Iris. In particular, it discusses what metadata is, where it fits
+into Iris, and more importantly how you can create, access, manipulate,
+and analyse your metadata.
+
+All the finer details covered here may not be entirely relevant to your use
+case, but it's here if you ever need it. In fact, you may want to skip
+straight ahead to :ref:`richer metadata`, and take it from there.
+
+
+Introduction
```
abooton, line 17: As I mentioned in my main comment:

It took me a while to follow the introduction and first section as I think the paragraphs hop around slightly.  All the vital information in there, so I doubt anyone else will notice, but I would:
* Introduction:  Move the first paragraph down, so all the class information is together, and give a more explicit reference to where the Cube, DimCoord, AuxCoord info is introduced wrt CF by re-referencing the "Data structures doc"

```suggestion
Introduction
============

As discussed in :ref:`iris_data_structures`, Iris draws heavily from the
`NetCDF CF Metadata Conventions`_ as a source for its data model, thus building
on the widely recognised and understood terminology defined within those
`CF Conventions`_ by the scientific community.

In :ref:`iris_data_structures` we also introduced several classes in Iris that care about your ``data``,
and also your ``metadata``: the :class:`~iris.cube.Cube`, the :class:`~iris.coords.AuxCoord`, and the
:class:`~iris.coords.DimCoord`.  In addition to these Iris models several other classes of `CF Conventions`_ metadata: the :class:`~iris.coords.AncillaryVariable`
(see `Ancillary Data`_ and `Flags`_), the :class:`~iris.coords.CellMeasure`
(see `Cell Measures`_), and also the :class:`~iris.aux_factory.AuxCoordFactory`
(see `Parametric Vertical Coordinate`_).

```

docs/iris/src/further_topics/metadata.rst
```diff
@@ -0,0 +1,987 @@
...
+`CF Conventions` classes.
+
+.. hint::
+
+    If there are any `CF Conventions`_ metadata missing from Iris that you
+    care about, then please let us know by raising a `GitHub Issue`_ on
+    `SciTools/iris`_
+
+
+Common Metadata
```
abooton, line 46: As I mentioned in my main comment:

All the info is here, but I found it easier to understand the table once I'd understood what is meant by metadata.  So I'd introduce what a metadata is first, before one tries to read it.  So, I’d move the table to the end of the section, and the example describing it immediately before.
Also:
* its ambiguous as to whether "var_name" and "circular" are the only Iris specific terms.
* In the table we have "standard name" is this the same as the "standard_name" in the pros ? 
* Why do some of the metadata members listed in the table have underscores in them, but others don't.  Are they the same things?  Is one a method?
* And - what does the red represent - I've been wondering this throughout our documentation.


```suggestion
Common Metadata
===============

Each of the Iris `CF Conventions`_ classes use  **metadata** to define them and give them meaning.

A small collective of individual **metadata members** can be used to define the
Iris `CF Conventions`_ classes.  All the members reference specific `CF Conventions`_ terms, 
except for  :attr:`~iris.coords.DimCoord.var_name` and :attr:`~iris.coords.DimCoord.circular` which are Iris specific terms.  The individual metadata members used to define the various Iris `CF Conventions`_ classes are shown in :numref:`metadata members table` . 

As :numref:`metadata members table` highlights, **specific** metadata is used to
define and represent each of the Iris `CF Conventions`_ classes. Thus the
metadata can be used to easily **identify**, **compare** and **differentiate**
between individual class instances.

For example, the collective metadata used to define an
:class:`~iris.coords.AncillaryVariable` are the ``standard_name``, ``long_name``,
``var_name``, ``units``, and ``attributes`` members. Note that, these are the
actual `data attribute`_ names of the metadata members on the Iris class.

```
.. and now have Table 1.  Should the member names all have "_"'s?

docs/iris/src/further_topics/metadata.rst
```diff
@@ -0,0 +1,987 @@
...
+just as you might expect. For example, given the following ``metadata``,
+
+    >>> metadata
+    DimCoordMetadata(standard_name='longitude', long_name=None, var_name='longitude', units=Unit('degrees'), attributes={'grinning face': '🙃'}, coord_system=GeogCS(6371229.0), climatological=False, circular=False)
+
+We can use the `namedtuple._make`_ method to create a **new**
+:class:`~iris.common.metadata.DimCoordMetadata` instance from an existing sequence
+or iterable:
+
+    >>> values = (1, 2, 3, 4, 5, 6, 7, 8)
```
abooton, line 241: Q: How do I know which metadata member the values are assigned to?  I'm assuming you have a secret hidden order?

docs/iris/src/further_topics/metadata.rst
```diff
@@ -0,0 +1,987 @@
...
+    >>> longitude.attributes = {"grinning face": "🙂"}
+    >>> other = longitude.metadata._replace(attributes={"grinning face":  "🙃"})
+    >>> other
+    DimCoordMetadata(standard_name='longitude', long_name=None, var_name='longitude', units=Unit('degrees'), attributes={'grinning face': '🙃'}, coord_system=GeogCS(6371229.0), climatological=False, circular=False)
+    >>> longitude.metadata == other
+    False
+
+One further point worth highlighting is it is possible for `NumPy`_ scalars
+and arrays to appear in the ``attributes`` `dict`_ of some Iris metadata class
+instances. Normally, this would cause issues,
```
abooton, line 374:
```suggestion
instances. Normally, this would cause issues. For example,
```

docs/iris/src/further_topics/metadata.rst
```diff
@@ -0,0 +1,987 @@
...
+to the :ref:`difference <metadata difference>` and :ref:`combine <metadata combine>`
+methods of metadata classes.
+
+
+.. _metadata difference:
+
+Metadata difference
+^^^^^^^^^^^^^^^^^^^
+
+Being able to compare metadata is valuable. Particularly when we have the
```
abooton, line 492:
```suggestion
Being able to compare metadata is valuable, especially when we have the
```

docs/iris/src/further_topics/metadata.rst
```diff
@@ -0,0 +1,987 @@
...
+convenience of being able to do this easily with metadata classes. However,
+when the result of comparing two metadata instances is ``False``, it begs
+the next obvious question, "**what's the difference?**"
```
abooton, line 495:
```suggestion
the question, "**what's the difference?**"
```

docs/iris/src/further_topics/metadata.rst
```diff
@@ -0,0 +1,987 @@
...
+.. doctest:: richer-metadata
+
+    >>> longitude.metadata.difference(metadata)
+    DimCoordMetadata(standard_name=None, long_name=(None, 'lon'), var_name=('longitude', 'lon'), units=(Unit('degrees'), Unit('radians')), attributes=None, coord_system=None, climatological=None, circular=None)
+
+The ``difference`` method returns a :class:`~iris.common.metadata.DimCoordMetadata` instance, when
+there is **at least** one ``metadata`` member with a different value, where,
+
+- ``None`` means that there was **no** difference for the member,
+- a `tuple`_ containing the two different associated values for the member
```
abooton, line 535:
```suggestion
- a `tuple`_ contains the two different associated values for the member
```

docs/iris/src/further_topics/metadata.rst
```diff
@@ -0,0 +1,987 @@
...
+- the ``upside-down face`` key is **new**
+
+.. doctest:: richer-metadata
+
+    >>> attributes = {"grinning face": "😀", "neutral face": "😜", "upside-down face": "🙃"}
+    >>> metadata = longitude.metadata._replace(attributes=attributes)
+    >>> metadata
+    DimCoordMetadata(standard_name='longitude', long_name=None, var_name='longitude', units=Unit('degrees'), attributes={'grinning face': '😀', 'neutral face': '😜', 'upside-down face': '🙃'}, coord_system=GeogCS(6371229.0), climatological=False, circular=False)
+
+Now, let's compare the two above instances for differences, and see what we get,
```
abooton, line 582: I was being slow...

```suggestion
Now, let's compare the two above instances for differences, and we see the attribute member contains a tuple of the differing information as expected, but each value is a dictionary containing the differing key/values,
```

docs/iris/src/further_topics/metadata.rst
```diff
@@ -0,0 +1,987 @@
...
+
+We can perform the **identity function** by comparing the metadata with itself,
+
+.. doctest:: metadata-combine
+
+    >>> metadata = cube.metadata.combine(cube.metadata)
+    >>> cube.metadata == metadata
+    True
+
+As you might suspect, combining identical metadata returns metadata that is
```
abooton, line 671: Q:  Do you mean "As you might expect" rather that "as expected" because, one might assume the metadata gets smashed together?  (As in leinient data? - I don't know as I've not read that section yet).

docs/iris/src/further_topics/metadata.rst
```diff
@@ -0,0 +1,987 @@
...
+Before combining their metadata together,
+
+.. doctest:: metadata-combine
+
+    >>> forecast_period.metadata.combine(longitude.metadata)
+    CoordMetadata(standard_name=None, long_name=None, var_name=None, units=None, attributes={}, coord_system=None, climatological=False)
+    >>> longitude.metadata.combine(forecast_period.metadata)
+    DimCoordMetadata(standard_name=None, long_name=None, var_name=None, units=None, attributes={}, coord_system=None, climatological=False, circular=None)
+
+However, note that commutativity in this case cannot be honoured, for obvious reasons.
```
abooton, line 770: Haha - the obvious reason is because the time can't have circular?  How have you managed to make it intelligently understand this!!  That's pretty cool given it isn't to do with the type of coord, but the scientific nature of the data being processed.

docs/iris/src/further_topics/metadata.rst
```diff
@@ -0,0 +1,987 @@
...
+   import iris
+   from iris.common import DimCoordMetadata
+   cube = iris.load_cube(iris.sample_data_path("A1B_north_america.nc"))
+   longitude = cube.coord("longitude")
+
+In general, the :ref:`equal <metadata equality>`, :ref:`difference <metadata difference>`,
+and :ref:`combine <metadata combine>` methods only support operations on instances
+of the same metadata class (see :ref:`exception to the rule <exception rule>`).
+
+However, metadata may be converted from one metadata class to another by using
```
abooton, line 789:
```suggestion
However, metadata may be converted from one metadata class to another using
```

docs/iris/src/further_topics/metadata.rst
```diff
@@ -0,0 +1,987 @@
...
+.. testsetup:: metadata-assign
+
+   import iris
+   cube = iris.load_cube(iris.sample_data_path("A1B_north_america.nc"))
+   longitude = cube.coord("longitude")
+   original = longitude.copy()
+   latitude = cube.coord("latitude")
+
+The ``metadata`` property available on each Iris `CF Conventions`_ container
+class (see :numref:`metadata classes table`) can be use not only **to get**
```
abooton, line 850:
```suggestion
class (see :numref:`metadata classes table`) can not only be used **to get**
```

docs/iris/src/further_topics/metadata.rst
```diff
@@ -0,0 +1,987 @@
...
+the metadata of an instance, but also **to set** the metadata on an instance.
+
```
abooton, line 852: For some reason I did not pickup on the difference between longitude and latitude.  Maybe just one line and one box would be ok:

```
For example, we can assign the :class:`~iris.common.metadata.DimCoordMetadata` metadata from the
``longitude`` coordinate directly to the ``latitude`` coordinate,

.. doctest:: metadata-assign

    >>> longitude.metadata
    DimCoordMetadata(standard_name='longitude', long_name=None, var_name='longitude', units=Unit('degrees'), attributes={}, coord_system=GeogCS(6371229.0), climatological=False, circular=False)
    >>> longitude.metadata = latitude.metadata
    >>> longitude.metadata
    DimCoordMetadata(standard_name='latitude', long_name=None, var_name='latitude', units=Unit('degrees'), attributes={}, coord_system=GeogCS(6371229.0), climatological=False, circular=False)
```

docs/iris/src/further_topics/metadata.rst
```diff
@@ -0,0 +1,987 @@
...
+    DimCoordMetadata(standard_name='longitude', long_name=None, var_name='longitude', units=Unit('degrees'), attributes={}, coord_system=GeogCS(6371229.0), climatological=False, circular=False)
+
+We can assign to it directly using the :class:`~iris.common.metadata.DimCoordMetadata` of the ``latitude``
+coordinate,
+
+.. doctest:: metadata-assign
+
+    >>> longitude.metadata = latitude.metadata
+    >>> longitude.metadata
+    DimCoordMetadata(standard_name='latitude', long_name=None, var_name='latitude', units=Unit('degrees'), attributes={}, coord_system=GeogCS(6371229.0), climatological=False, circular=False)
```
abooton, line 868: Q:  What happens to the metadata-data values, are we trying to put lat values on longs?
Lats -90 to 90.  Lons go -180 to 180.  Do you end up with a 90 by 90 deg region?

docs/iris/src/further_topics/metadata.rst
```diff
@@ -0,0 +1,987 @@
...
+on the widely recognised and understood terminology defined within those
+`CF Conventions`_ by the scientific community.
+
+Iris models several other classes of `CF Conventions`_ metadata apart from
+those mentioned above. Such as the :class:`~iris.coords.AncillaryVariable`
+(see `Ancillary Data`_ and `Flags`_), the :class:`~iris.coords.CellMeasure`
+(see `Cell Measures`_), and also the :class:`~iris.aux_factory.AuxCoordFactory`
+(see `Parametric Vertical Coordinate`_).
+
+Collectively, the asforementioned classes will be known here as the Iris
```
abooton, line 36: typo
```suggestion
Collectively, the aforementioned classes will be known here as the Iris
```

docs/iris/src/userguide/cube_maths.rst
```diff
@@ -1,6 +1,6 @@
-======================
-Basic cube mathematics
-======================
+================
+Basic cube maths
+================
 
 
 The section :doc:`navigating_a_cube` highlighted that
```
abooton, line 6: I started reading this section and hadn't realised I was in a specific example, and I started reading the previous section for the start of the example, and then found myself second guessing the example
Given that this epitomises the value of Iris for tieing metadata to data, and having the ability to do science without having to concern oneself with it, I think it is worth improving this introduction.

```suggestion
 The section Navigating a cube highlighted that every cube has a data attribute; this attribute can then be manipulated directly.  However, the problem with manipulating the data directly is that other metadata may become inconsistent.  For example, assume we have the following temperature data and we wish to convert it to units of Centigrade::
  
   >>> print(cube.data)
   293.15
   >>> print(cube.units)
   “kelvin”

We could manipulate the data directly as follows, but the units would remain in “kelvin” unless we rectify the units attribute too::

   cube.data -= 273.15
   cube.units = “celsius”

In order to reduce the likely hood of metadata becoming inconsistent, we can use the :meth:`iris.cube.Cube.convert_units` method to automatically convert a cube’s data and update its units attribute. So, the two steps above can be achieved by::

   cube.convert_units('celsius')

This issue is quickly exacerbated when carrying out more complex, fundamental arithmetic operations such as addition, subtraction, division and multiplication.  However, applying the functions directly to the cube, using “cube maths”, alleviates the problem just as applying the iris.cube.Cube.convert_units()  does in the example above.

This issue is quickly exacerbated when carrying out more complex, fundamental arithmetic operations such as addition, subtraction, division and multiplication.  However, applying the functions directly to the cube, using “cube maths”, alleviates the problem just as applying the iris.cube.Cube.convert_units()  does in the example above.
```

abooton (commented):
Bill,  I've had a look through all but the Lenient metadata documentation and I think it is really good; I think it is some of the clearest technical documentation I've read for a long time.  I hope your pleased with it.
I have some really minor comments, and I've been uber picky on something that is subjective to begin with - so take them or leave them as you wish.  You can stop reading now if you want!


In order to review the Metadata section I went back and re-read "Navigating a Cube" and "Basic cube maths".  I'll add a few comments on those too as I noticed you are editing the files anyway.  Also, I'm afraid I've not really read the previous review comments so I apologise if I mention something again.

1) MetaData page
* Overall I thought it was extremely clear.  Just a few typos, I had minor suggestions for slicking up the introduction.  It seemed like you really hit your stride when you got to the "Common Metadata API"!
* We recommend that readers jump to the "Richer metadata section" - but its rather buried in the side bar.
* Typo:  "asforementioned" => "aforementioned"
* It took me a while to follow the introduction and first section until I realised, I just needed to read the paragraphs in a different order!  (You've already got all the vital information in there, so ignore my comments as you see fit. I doubt anyone else will notice, but I read very slowly, line by line)
Personally I would:
* Introduction:  Move the first paragraph down, so all the class information is together, and make it more explicit by re-referencing the "Data structures doc" that definitively introduces Cube, AuxCoord and Coord classes.
(Also it is a bit weird that we tell the user in the advanced section that "i.e. data about data" yet we don't explain it this simply in the "Basic User Guide"!)
Introduction
As discussed in Iris data structures, Iris draws heavily from the NetCDF CF Metadata Conventions as a source for its data model, thus building on the widely recognised and understood terminology defined within those CF Conventions by the scientific community.  
In Iris data structures we also introduced several classes in Iris that care about your data, and also your metadata: the Cube, the AuxCoord, and the DimCoord.  In addition to these, Iris models several other classes of CF Conventions metadata: the AncillaryVariable (see Ancillary Data and Flags), the CellMeasure (see Cell Measures), and also the AuxCoordFactory (see Parametric Vertical Coordinate).

* Common Metadata: I'd introduce what a metadata is first, and I'd introduce whats in the table before one tries to read it.  So, I’d move the table to the end of the section, and the example describing it immediately before. 
More like:
Common Metadata
Each of the Iris CF Conventions classes use metadata to define them and give them meaning.    
A small collective of individual metadata members can be used to define the Iris CF Conventions classes.  All the members reference specific CF Conventions terms, except from var_name and circular which are Iris specific terms.  The individual metadata members used to define the various Iris CF Conventions classes are shown in Table 1.  
As Table 1 highlights, a specific subset of the metadata members are used to define and represent each of the Iris CF Conventions classes. Thus the metadata can be used to easily identify, compare and differentiate between individual class instances.  
For example, the collective metadata used to define an AncillaryVariable are the standard_name, long_name, var_name, units, and attributes members. Note that, these are the actual data attribute names of the metadata members on the Iris class.
* All my other (tiny/minor) comments are on the files


2) Document ordering
* I really like having the advanced/detailed "Metadata" section in the "Further Reading".  I did find that (having read lots) I’d overlooked “Further Topics” as I think I associate it as “Miscellaneous” topics, maybe because it has its own Introduction so I skimmed past it as separate to the user guide. 
* Maybe combine the introduction information into the main User Guide section?
I wondered if the following is worth thinking about (not here) in the future:
* Having the User Guide in a couple of parts, so we have one introduction in which we recommend users read Part1 which provides an essential orientation to Iris.  For the advanced users, further topics are explained in Part 2?  E.g. 
* User Guide - Part 1, User Guide - Part 2, User Guide - Technical Papers (in light of Ruth's comment about not having found them before) 
Or   * User Guide: An Introduction, User Guide: Advanced/Further Topics, User Guide

3) "Basic cube maths"
* Title of "Basic cube maths" => "Cube maths"
  * Then the title is aligned with the other titles "Cube interpolation", "Cube statistics". (And I don't think 'basic' does it justice anymore! I think it will make it easier to spot in th euser guide)
* When I started this section I felt like I was second guessing what was going on.  I'll add some comments about the opening paragraph, to make the example being discussed more explicit. (Particularly as I think this epitomises what Iris helps the user best)

4) LH side bar
* I was having rather a lot of issues with the scrolling of the LHS bar - I don't think it is quite right.
* I can't often see the top of the LHS bar, and it scrolls with the main documentation page.  (I can show if you want)
* It seems to work fine when looking at search results but not on documentation pages?

5) Miscellaneous
* I think I spotted inconsistent use of "meta-data" vs "metadata" in the User Guide
* I wondered if an extra link in the "Basic cube maths" section to the "Metadata" section (can't remember if there is one) would be good for advertising all your fab work

docs/iris/src/userguide/cube_maths.rst
rcomer: "likelihood" should be one word, and the last paragraph is duplicated here.

docs/iris/src/further_topics/lenient_metadata.rst
```diff
@@ -0,0 +1,477 @@
...
+.. doctest:: strict-behaviour
+
+    >>> metadata.name()
+    'latitude'
+    >>> latitude.name()
+    'latitude'
+
+Regardlessly, at the end of the day we may not actually care that the
+``var_name`` is different. However, Iris relentlessly forces us to deal
+with such a difference; sometimes this can be challenging to overcome.
```
lbdreyer, line 132: I think you might mean "Regardless"? `Regardlessly` is an adverb so it doesn't make sense to be used in this way. Also, `At end of the day` is very informal language. I suppose it is in keeping with the tone of the rest of the metadata documentation, but I personally find that phrase, in particular, is a bit too informal. It may be fine in a email or in conversation, but not in official documentation. 

`Iris relentlessly forces us to deal with such a difference; sometimes this can be challenging to overcome.`
I find this way too critical of Iris, not too mention unfair. If I was a user and I read this, I would think "Why should I use a package that even the developer's are happily criticising in their own documentation!"
I do get the point you're trying to make, but it makes Iris sounds like too much of a bother to use, so I think it needs to be rephrased in a different way.

Maybe something like:
Regardless of whether the difference in var_name matters, Iris will treat the metadata as different. Thus the user is required to manually modify the metadata to avoid this.

docs/iris/src/further_topics/lenient_metadata.rst
```diff
@@ -0,0 +1,477 @@
...
+
+.. _lenient members:
+
+Lenient members
+---------------
+
+:ref:`lenient behaviour` is not applied regardlessly across all metadata members
+participating in a lenient ``equal``, ``difference`` or ``combine`` operation.
+Rather, a more pragmatic application is employed based on the `CF Conventions`_
+meaning of the member, and whether being lenient would result in erroneous
```
lbdreyer, line 391: It may just be me, but I found that I had to reread this sentence a few times.
How about changing `meaning` -> `definition`

docs/iris/src/further_topics/lenient_metadata.rst
```diff
@@ -0,0 +1,477 @@
...
+
+.. _lenient members table:
+.. table:: - Lenient member participation
+   :widths: auto
+   :align: center
+
+   ============================================================================================= ================== ============
+   Metadata Class                                                                                Member             Behaviour
+   ============================================================================================= ================== ============
+   †                                                                                             ``standard_name``  ``lenient``‡
```
lbdreyer, line 402: Seems a bit strange to just have the dagger here. Why not have something like
`All metadata objects†`

And then keep the key as it was before

docs/iris/src/further_topics/metadata.rst
bjlittle: @abooton It's no secret, it's implied that it matches the `namedtuple._fields`. I've added clarification.

docs/iris/src/further_topics/metadata.rst
bjlittle: @abooton I wasn't as verbose, as I don't quite think was quite necessary, but I've updated to highlight that the difference is in the `attributes` dictionary. Hope that's a happy compromise...

docs/iris/src/further_topics/metadata.rst
bjlittle: @abooton I've printed the `latitude.metadata` just to clarigy, what's what...

docs/iris/src/further_topics/metadata.rst
bjlittle: @abooton This is just an example of assigning metadata... it's really a bit of a nonsense, but conveniently serves the purpose of highlighting the utility of being easily able to assign **all** the metadata from one instance to another... regardless of whether it makes semantic sense or not i.e., yup, indeed, you now have a `longitude` points/bounds for a `DimCoord` that according to the metadata is a `latitude`

docs/iris/src/userguide/cube_maths.rst
bjlittle: @abooton I really didn't want to start straying into other parts of the docs and rewriting them... mainly because this PR is getting rather overwhelming. Perhaps raise an issue to tackle this separately? Or even better a PR of your own? :stuck_out_tongue_winking_eye:

docs/iris/src/further_topics/lenient_maths.rst
```diff
@@ -0,0 +1,281 @@
...
+**scalar coordinates**, or the ``experiment-id`` in the ``attributes`` dictionary.
+
+This is because strict cube maths, in general, will only return common metadata
+and common coordinates that are :ref:`strictly equivalent <strict equality>`.
+
+
+Finer detail
+============
+
+In general, if you want preserve as much metadata and coordinate information as
```
rcomer, line 235: if you want **to** preserve...

docs/iris/src/further_topics/lenient_maths.rst
```diff
@@ -0,0 +1,281 @@
...
+  the metadata of coordinates on the resultant :class:`~iris.cube.Cube`,
+- lenient behaviour will attempt to cover each dimension with a :class:`~iris.coords.DimCoord`
+  in the resultant :class:`~iris.cube.Cube`, even though only one :class:`~iris.cube.Cube`
+  operand may describe that dimension,
+- lenient behaviour will attempt to include **auxiliary coordinates** in the
+  resultant :class:`~iris.cube.Cube` that are local to only one :class:`~iris.cube.Cube`
+  operand,
+- lenient behaviour will attempt to include **scalar coordinates** in the
+  resultant :class:`~iris.cube.Cube` that are local to only one :class:`~iris.cube.Cube`
+  operand,
```
rcomer, line 255: not sure about the phrase "are local to", maybe "exist on"???

rcomer (commented):
Hi @bjlittle, I've just binge-read the lot!  In general it all looks really clear to me.  My comments are mostly just thoughts that occurred while reading and don't necessarily need any particular action.

docs/iris/src/further_topics/metadata.rst
```diff
@@ -0,0 +1,988 @@
...
+:class:`~iris.common.metadata.DimCoordMetadata` instance.
+
+Note that, the ``from_metadata`` method is also available on a metadata
+class instance,
+
+.. doctest:: metadata-convert
+
+    >>> longitude.metadata.from_metadata(cube.metadata)
+    DimCoordMetadata(standard_name='air_temperature', long_name=None, var_name='air_temperature', units=Unit('K'), attributes={'Conventions': 'CF-1.5', 'STASH': STASH(model=1, section=3, item=236), 'Model scenario': 'A1B', 'source': 'Data from Met Office Unified Model 6.05'}, coord_system=None, climatological=None, circular=None)
+
```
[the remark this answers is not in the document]
bjlittle: @rcomer Yup, it is, but only once its assigned to the class container... subtle difference, and I didn't want to get into that in the docs. Interesting that you picked up on it :100: 

Just for your own weirded out sanity... that's the job of the `iris.common.mixin.CFVariableMixin` which will **validate** any content assigned to the class container (there's a separate metadata manager under the hood, and that's way out of scope of the docs). At the runtime point of assignment, then the sanity check of type etc takes place, and not before.

So essentially you can assign what you like to a metadata instance, it doesn't care and it doesn't have to... it's only class container that **requires** to care. That's why I'm able to assign tuples to a metadata instance resulting from the `difference` operation :wink: 

Does that help de-weirded-out you?

docs/iris/src/further_topics/metadata.rst
```diff
@@ -0,0 +1,988 @@
...
+It is also possible to assign to the ``metadata`` property of an Iris
+`CF Conventions`_ container with an iterable containing the **correct
+number** of associated member values, e.g.,
+
+.. doctest:: metadata-assign
+
+    >>> values = [getattr(latitude, member) for member in latitude.metadata._fields]
+    >>> longitude.metadata = values
+    >>> longitude.metadata
+    DimCoordMetadata(standard_name='latitude', long_name=None, var_name='latitude', units=Unit('degrees'), attributes={}, coord_system=GeogCS(6371229.0), climatological=False, circular=False)
```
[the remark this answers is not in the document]
bjlittle: To be honest I think that's always (maybe) been there... anyways, there it is in all it's magnificent glory

docs/iris/src/further_topics/metadata.rst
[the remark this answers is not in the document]
rcomer: Thanks @bjlittle I will sleep better now! 😆

docs/iris/src/further_topics/lenient_metadata.rst
```diff
@@ -0,0 +1,476 @@
...
+
+.. doctest:: lenient-behaviour
+
+    >>> metadata.equal(latitude.metadata)
+    False
+    >>> metadata.equal(latitude.metadata, lenient=True)
+    True
+
+Again, lenient equality (:numref:`lenient equality table`) offers a more
+forgiving and practical alternative to strict behaviour.
```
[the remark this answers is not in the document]
bjlittle: @rcomer It kinda is, and I'm glad you're thinking about that... clearly my job is done here :wink: 

But in essence, they are different, will result in `False` if compared (leniently or strictly) and can't be combined... as they're different.

The tables in the `Lenient metadata` section might help clarify that for strict and lenient behaviour.

HTH

docs/iris/src/further_topics/lenient_maths.rst
bjlittle: @rcomer Awesome spot :eyes: ... that section is hot off the press, so thanks for taking a look at it, much appreciated :+1:

rcomer:
Looks good to me 😎

docs/iris/src/further_topics/lenient_metadata.rst
```diff
@@ -0,0 +1,476 @@
...
+Rather, a more pragmatic application is employed based on the `CF Conventions`_
+definition of the member, and whether being lenient would result in erroneous
+behaviour or interpretation.
+
+.. _lenient members table:
+.. table:: - Lenient member participation
+   :widths: auto
+   :align: center
+
+   ============================================================================================= ================== ============
```
abooton, line 397: I read this table a bit too quickly to being I think...
At a glance, it looked like only the "standa_name", "long_name", "var_name" and the "attributes" that have the lenient property.  However, re-reading it, I see that eg "measure" is only strict on CellMeasureMetadata.

I wonder whether it would be clearer if we had the equivalent of table 1,  but indicating if strict/lenient ?

abooton (approved):
This is looking really good.  
I read the metadata section really thoroughly, and it reads really well.
I've given the Lenient metadata and the lenient maths a quick read.  I think all the relevant information needed is here.  So I'm going to merge it in 👍  
As noted in my previous comment there were a couple of misc. items that we may wish to raise as follow on issues, so we can give the latter sections a thorough read at the same time.

bjlittle:
Awesome, thanks!

PR #3870, opened 2020-09-18
Whatsnew for effects on aux factories of units defaulting to 'unknown'.

pp-mo:
## 🚀 Pull Request
Documenting possible changed behaviour in netcdf loads.
But it really only applies to files that were "wrong".

### DIFF
--- a/docs/iris/src/whatsnew/latest.rst
+++ b/docs/iris/src/whatsnew/latest.rst
@@ -140,12 +140,12 @@ This document explains the changes made to Iris for this release
 
 * `@stephenworsley`_ changed Iris objects loaded from NetCDF-CF files to have
   ``units='unknown'`` where the corresponding NetCDF variable has no ``units``
-  property. Previously these cases defaulted to ``units='1'``. (:pull:`3795`).
+  property. Previously these cases defaulted to ``units='1'``.
   This affects loading of coordinates whose file variable has no "units"
   attribute (not valid, under `CF units rules`_):  These will now have units
   of `"unknown"`, rather than `"1"`, which **may prevent the creation of
   a hybrid vertical coordinate**.  While these cases used to "work", this was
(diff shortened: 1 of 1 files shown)


docs/iris/src/whatsnew/latest.rst
```diff
@@ -140,7 +140,12 @@ This document explains the changes made to Iris for this release
 
 * `@stephenworsley`_ changed Iris objects loaded from NetCDF-CF files to have
   ``units='unknown'`` where the corresponding NetCDF variable has no ``units``
-  property. Previously these cases defaulted to ``units='1'``. (:pull:`3795`)
+  property. Previously these cases defaulted to ``units='1'``. (:pull:`3795`).
+  This affects loading of coordinates whose file variable has no "units"
+  attribute (not valid, under `CF units rules`_):  These will now have units
+  of `"unknown"`, rather than `"1"`, which **may prevent the creation of
+  a hybrid vertical coordinate**.  While these cases used to "work", this was
+  never really correct behaviour.
```
trexfeathers, line 148:
```suggestion
  property. Previously these cases defaulted to ``units='1'``.
  This affects loading of coordinates whose file variable has no "units"
  attribute (not valid, under `CF units rules`_):  These will now have units
  of `"unknown"`, rather than `"1"`, which **may prevent the creation of
  a hybrid vertical coordinate**.  While these cases used to "work", this was
  never really correct behaviour. (:pull:`3795`)
```

In accordance with the order in [Contributing a “What’s New” entry](https://scitools-iris.readthedocs.io/en/latest/developers_guide/documenting/whats_new_contributions.html#writing-a-contribution)

trexfeathers (changes requested):

stephenworsley (approved):
Looks good to me.

trexfeathers (approved):

PR #3873, opened 2020-09-21
linkcheck ignore for iris pull and issue url

tkknight:
### Description

Added 2 urls to be **ignored** when running the **linkcheck** to avoid spamming github when running a **linkcheck**.

Since we have started to use the pull and issue directives:

```
:issue:`9999`
:pull:`9999`
```

... in the **what's new** that is later substituted for a full url, when running the **linkcheck** from **travis** the host (github) may **refuse the connection** as it is essentially being hit frequently from many different travis jobs.  This will show as a **warning/error** and fail the linkcheck travis task.

The ignore urls only covers Iris **pull** and **issue**, it will still check links to **blobs** (source code).

---


bjlittle (approved):
@tkknight Nice one. We've had such issues before, do this change is most welcome 👍

tkknight:
thanks @bjlittle

PR #3752, opened 2020-07-12
Read The Docs documentation

tkknight:
Iris documentation overhaul to use a modern responsive theme and to be served via https://readthedocs.org/.  This is the new baseline for future improvements to be made.

- Added page for contributing to the documentation
- Adopted **sphinx_rtd_theme** (responsive to device, standard formatting)
- Adopted **sphinx-copybutton**
- Adopted **sphinx-gallery** and removed the custom one.  Updated references to the new gallery
- Added "make html-noplot" option to skip the gallery creation (fast when developing)
- Renamed the examples_code and example_tests to be gallery_code and gallery_tests
- Prefixed all gallery files with "plot_"
- Ensured a **clean build with no warnings** (there were many)
- Setup to now build by readthedocs (see the config file in the root **.readthedocs.yml that uses conda**)
- Added travis test to include "**make linkcheck**" that checks that all urls are valid - fixed many that were not
- Updated image on iris github homepage to be better resolution and standard font (Georgia)
- Build directory now in src directory, named "_build"
- Remove legacy files from "_static"
- Sphinx conf.py file has better structure and comments
- Copyright date range end is now dynamic
- Various typos and formatting in source
- API rst builder now show no warnings when run
- Added link to the "**legacy**" documentation
- Put in place a **block list for troublesome modules** when generating the API documentation
- Changed "Whitepapers" to "Technical Papers"

PR #3877, opened 2020-09-22
Docs warning to error

tkknight:
### Description

Configured **sphinx-build** to promote **warnings to errors** when building the documentation via ``make html``.

Updated the **Makefile** to use two new arguments:

- ``-W``. Turn warnings into errors. This means that the build stops at the first warning and sphinx-build exits with exit status 1.
- ``--keep-going``. With -W option, keep going processing when getting warnings to the end of build, and sphinx-build exits with exit status 1.

Also updated the ``contributing to the documentation`` in the Iris docs to state that warnings are promoted to errors.

For more info see https://www.sphinx-doc.org/en/master/man/sphinx-build.html?highlight=--keep-going#cmdoption-sphinx-build-W

### Why do this?

Before the overhaul of the Iris docs (https://github.com/SciTools/iris/pull/3752), there were many warnings in the build but were not actively fixed as no error was raised and the travis task completed ok.

### Why not promote all warnings to errors?

I believe the Travis jobs (of a task) will currently fail correctly for the following:

- ``make gallerytest``
- ``make doctest``
- ``make linkcheck``

For ``make html-noplot`` (skip the gallery creation) it could be applied but this is usually used as a quick way to build the docs and the ``make html`` that is part of the travis jobs (and can be run manually) will capture any issues anyway.

---

### DIFF
--- a/docs/iris/src/developers_guide/contributing_documentation.rst
+++ b/docs/iris/src/developers_guide/contributing_documentation.rst
@@ -46,9 +46,6 @@ If you wish to run a clean build you can run::
 
 This is useful for a final test before committing your changes.
 
-.. todo:: warning to errors just for ``make html``?? what about ``doctest``,
-          ``gallerytest``, ``linkcheck``, ``html-noplot``?
-
 .. note:: In order to preserve a clean build for the html, all **warnings**
           have been promoted to be **errors** to ensure they are addressed.
           This check is only applied when ``make html`` is run.
--- a/docs/iris/src/developers_guide/contributing_documentation.rst
+++ b/docs/iris/src/developers_guide/contributing_documentation.rst
@@ -48,7 +48,7 @@ This is useful for a final test before committing your changes.
 
 .. note:: In order to preserve a clean build for the html, all **warnings**
           have been promoted to be **errors** to ensure they are addressed.
-          This is **only** applied when ``make html`` is run.
+          This **only** applies when ``make html`` is run.
 
 .. _travis-ci: https://travis-ci.org/github/SciTools/iris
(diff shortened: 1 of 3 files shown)


docs/iris/src/developers_guide/contributing_documentation.rst
```diff
@@ -46,11 +46,9 @@ If you wish to run a clean build you can run::
...
 This is useful for a final test before committing your changes.
 
-.. note:: In addition to the automated `travis-ci`_  build of the documentation,
-          the https://readthedocs.org/ service is also used.  The configuration
-          of this held in a file in the root of the
-          `github Iris project <https://github.com/SciTools/iris>`_  named
-          ``.readthedocs.yml``.
+.. note:: In order to preserve a clean build for the html, all **warnings**
+          have been promoted to be **errors** to ensure they are addressed.
+          This **only** applies when ``make html`` is run.
```
bjlittle, line 51: BTW is it now best to in general do
```
.. note::

   <content>
```
rather than
```
.. note:: <content>
```
Just curious...

tkknight: Either is ok in my books.  I prefer the latter as it helps the note be more visual - but this is my preference.  If you have a particularly long note then the first option may be better as there is more text per line (due to no indentation)

bjlittle (approved):
@tkknight Lovely.

This elevates the docs to be a first class citizen and ensures that we don't bank broken docs... you've worked incredibly hard to get us to this point.

Awesome sauce :100: :beers: :partying_face: :tada:

PR #3892, opened 2020-09-30
Deprecate iris.util.as_compatible_shape

bjlittle:
### Description
The PR marks the `iris.util.as_compatible_shape` function for deprecation from `iris` in a future release, in preference for using an instance of the `iris.common.resolve.Resolve` class.

---

### DIFF
--- a/lib/iris/util.py
+++ b/lib/iris/util.py
@@ -24,6 +24,7 @@ import numpy as np
 import numpy.ma as ma
 
 import iris
+from iris._deprecation import warn_deprecated
 import iris.coords
 import iris.exceptions
 import iris.cube
@@ -1222,6 +1223,8 @@ def as_compatible_shape(src_cube, target_cube):
 
     .. note:: This function will load and copy the data payload of `src_cube`.
 
+    .. deprecated:: 3.0.0
+
     Args:
 
     * src_cube:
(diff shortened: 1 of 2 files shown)


lib/iris/util.py
```diff
@@ -1222,6 +1223,8 @@ def as_compatible_shape(src_cube, target_cube):
 
     .. note:: This function will load and copy the data payload of `src_cube`.
 
+    .. deprecated:: 3.0.0
+
```
trexfeathers, line 1227: According to [contributing_deprecations](https://github.com/SciTools/iris/blob/80a887877fed618b20755ec420d8fc46390f1a59/docs/iris/src/developers_guide/contributing_deprecations.rst), you should also include some prose in the docstring:

>As with the deprecation warning, you should include advice on how to
   avoid using the deprecated API within the content of this directive.
   Feel free to include more detail in the updated docstring than in the
   deprecation warning.

trexfeathers (changes requested):
Thanks @bjlittle 

As well as my comment below, there are a number of references within the [test_util](https://github.com/SciTools/iris/blob/80a887877fed618b20755ec420d8fc46390f1a59/lib/iris/tests/test_util.py) module. These should either be changed in this PR, or an issue should be raised for future work. Thanks!

lib/iris/util.py
bjlittle: Good call... I'm just being lazy :sleepy:

trexfeathers (approved):

trexfeathers:
> there are a number of references within the test_util module. These should either be changed in this PR, or an issue should be raised for future work.

So it's apparently not appropriate to remove such references until the feature is actually removed, once it has been deprecated for long enough. There will be no need to raise an issue since final removal will cause test failures so they can't be missed.

PR #3701, opened 2020-04-28
Lazy regridding with Linear, Nearest, and AreaWeighted

bouweandela:
This pull request implements lazy regridding for the Linear, Nearest, and AreaWeighted regridders.

Closes #3700

Issue #3294, opened 2019-03-07
Rationalise percentile methods


rcomer:
The [_percentile](https://github.com/SciTools/iris/blob/343d3297b547b0837b794d8b1f60e9eb46aee34c/lib/iris/analysis/__init__.py#L1155) function that `iris.analysis.PERCENTILE` is based on allows the keyword `fast_percentile_method` so that `np.percentile` may be used in place of `scipy.mstats.mquantiles`.  However, this option does not appear in the [documentation](https://scitools-iris.readthedocs.io/en/latest/generated/api/iris/analysis.html#iris.analysis.PERCENTILE) for `iris.analysis.PERCENTILE`.  [Addresssed in #3901]

Also, is there any reason why the `_percentile` function and `PERCENTILE` aggregator shouldn't allow passing of `**kwargs` to `np.percentile` if this option is chosen?  If implemented, this would allow users to specify an interpolation method in `np.percentile`. [Addressed in #4791]

mo-g:
Stylistically, would it be cleaner to allow passing all kwargs, to filter passed kwargs, or to manually expose the desired functionality by manually supporting and passing the specific argument?

It's been so long since I coded with others, that I honestly don't remember at all.

DPeterK:
It's probably just due to ignorance / oversight at the time of putting `PERCENTILE` together... Have you tried just passing the kwarg and seeing what happens? If it all just works then a quick docs update would be great, and if it doesn't and you feel like tackling the error (for example, the return type / shape / other from the NumPy functionality may not match what's expected by Iris) then that's also most welcome, of course 👍

rcomer:
I should probably add, this was prompted by a question on our internal Yammer.  The user specifically wanted to use "nearest" interpolation method, which does not appear to be available with `scipy.mstats.mquantiles`, but is available in `np.percentile`.

> Have you tried just passing the kwarg and seeing what happens?

I guess you mean the `fast_percentile_method` kwarg?  No I haven't, but I'd be willing to bet that you're right and it just works.  Any thoughts on additional passing of the `interpolation` kwarg?  As @mo-g points out, the right way to do that is not trivially obvious.  It also occurs to me that we have 3 methods that do very similar things and we don't want to confuse users more than necessary:

- `PERCENTILE` with `fast_percentile_method` kwarg: fastest but does not handle weights or masked data.
- straight `PERCENTILE`: slower but handles masked data.
[...]

PR #3901, opened 2020-10-02
Lazy Percentile Aggregator

rcomer:
### Description
Implements lazy aggregation for the `iris.analysis.PERCENTILE` aggregator.

My experience with `dask` is minimal, so I had a look at #3701 and basically used that approach as a template for the `_percentile` function implementation.  I believe I've understood the intent of the `map_complete_blocks` function but really not how it works.  So it's possible I've missed something important there.

I note that [dask.array.percentile](https://docs.dask.org/en/latest/array-api.html#dask.array.percentile) exists for 1D arrays, but I'm not clear how we'd incorporate the check for masked data if using that.

leaving as draft for now as I have an outstanding question about code structure and associated tests (see inline comments).

### Priority
I realise you are all busy so I'm not expecting this to get attention till after the v3 final release.

---

### DIFF
--- a/lib/iris/_lazy_data.py
+++ b/lib/iris/_lazy_data.py
@@ -359,7 +359,7 @@ def map_complete_blocks(src, func, dims, out_sizes):
 
     Args:
 
-    * src (:class:`~iris.cube.Cube`):
+    * src (:class:`~iris.cube.Cube` or lazy array):
         Source cube that function is applied to.
     * func:
         Function to apply.
@@ -369,10 +369,12 @@ def map_complete_blocks(src, func, dims, out_sizes):
         Output size of dimensions that cannot be chunked.
 
     """
-    if not src.has_lazy_data():
+    if is_lazy_data(src):
+        data = src
+    elif not src.has_lazy_data():
         return func(src.data)
-
-    data = src.lazy_data()
+    else:
+        data = src.lazy_data()
 
     # Ensure dims are not chunked
     in_chunks = list(data.chunks)
--- a/lib/iris/_lazy_data.py
+++ b/lib/iris/_lazy_data.py
@@ -359,7 +359,7 @@ def map_complete_blocks(src, func, dims, out_sizes):
 
     Args:
 
-    * src (:class:`~iris.cube.Cube` or lazy array):
+    * src (:class:`~iris.cube.Cube` or array-like):
         Source cube that function is applied to.
     * func:
         Function to apply.
@@ -371,6 +371,9 @@ def map_complete_blocks(src, func, dims, out_sizes):
     """
     if is_lazy_data(src):
         data = src
+    elif not hasattr(src, "has_lazy_data"):
+        # Not a lazy array and not a cube.  So treat as ordinary numpy array.
+        return func(src)
     elif not src.has_lazy_data():
         return func(src.data)
     else:
--- a/lib/iris/analysis/__init__.py
+++ b/lib/iris/analysis/__init__.py
@@ -1278,7 +1278,6 @@ def _percentile(
 
     """
     if not isinstance(percent, Iterable):
-        scalar_percent = True
         percent = [percent]
     percent = np.array(percent)
 
@@ -1304,7 +1303,7 @@ def _percentile(
 
     # Check whether to reduce to a scalar result, as per the behaviour
     # of other aggregators.
-    if result.shape == (1,) and scalar_percent:
+    if result.shape == (1,):
         result = np.squeeze(result)
 
     return result
--- a/lib/iris/analysis/__init__.py
+++ b/lib/iris/analysis/__init__.py
@@ -1267,12 +1267,24 @@ def _percentile(data, percent, fast_percentile_method=False, **kwargs):
     If a new additive dimension is formed, then it will always be the last
     dimension of the resulting percentile data payload.
 
+    Args:
+
+    * data (array-like)
+        array from which percentiles are to be calculated
+    * axis (int or tuple of int)
+        Axis or axes along which the percentiles are computed
+        (handled by decorator)
+
     Kwargs:
 
-    * fast_percentile_method (boolean) :
+    * fast_percentile_method (boolean)
         When set to True, uses the numpy.percentiles method as a faster
         alternative to the scipy.mstats.mquantiles method. Does not handle
         masked arrays.
+
+    **kwargs
+        passed to scipy.stats.mstats.mquantiles if fast_percentile_method is
+        False
 
     """
     if not isinstance(percent, Iterable):
--- a/lib/iris/analysis/__init__.py
+++ b/lib/iris/analysis/__init__.py
@@ -1241,7 +1241,7 @@ def _percentile(
     # Check whether to reduce to a scalar result, as per the behaviour
     # of other aggregators.
     if result.shape == (1,) and scalar_percent:
-        result = result[0]
+        result = np.squeeze(result)
 
     return result
 
--- a/lib/iris/analysis/__init__.py
+++ b/lib/iris/analysis/__init__.py
@@ -1245,21 +1245,17 @@ def _percentile(
         masked arrays.
 
     """
-    # Express axis as list of non-negative integers.
-    if not isinstance(axis, Iterable):
-        axis = [axis % data.ndim]
-    else:
-        axis = sorted([dim % data.ndim for dim in axis])
-
     # Get data as a 1D or 2D view with the target axis as the trailing one.
-    end_size = np.prod([data.shape[dim] for dim in axis])
-    untouched_dims = [dim for dim in range(data.ndim) if dim not in axis]
-    shape = [data.shape[dim] for dim in untouched_dims]
+    if not isinstance(axis, Iterable):
+        axis = (axis,)
+    end = range(-len(axis), 0)
 
-    data = data.transpose(untouched_dims + axis)
-
-    if shape or len(axis) > 1:
-        data = data.reshape([np.prod(shape), end_size])
+    data = np.moveaxis(data, axis, end)
+    shape = data.shape[: -len(axis)]
+    if shape:
+        data = data.reshape(np.prod(shape), -1)
+    else:
+        data = data.flatten()
 
     if not isinstance(percent, Iterable):
         scalar_percent = True
--- a/lib/iris/analysis/__init__.py
+++ b/lib/iris/analysis/__init__.py
@@ -1204,6 +1204,42 @@ def _build_dask_mdtol_function(dask_stats_function):
     return inner_stat
 
 
+def _axis_to_single_trailing(stats_function):
+    """
+    Given a statistical function that acts on the trailing axis of a 1D or 2D
+    array, wrap it so that higher dimension arrays can be passed, as well as any
+    axis as int or tuple.
+
+    """
+
+    @wraps(stats_function)
+    def inner_stat(data, axis, *args, **kwargs):
+        # Get data as a 1D or 2D view with the target axis as the trailing one.
+        if not isinstance(axis, Iterable):
+            axis = (axis,)
+        end = range(-len(axis), 0)
+
+        data = np.moveaxis(data, axis, end)
+        shape = data.shape[: -len(axis)]  # Shape of dims we won't collapse.
+        if shape:
+            data = data.reshape(np.prod(shape), -1)
+        else:
+            data = data.flatten()
+
+        result = stats_function(data, *args, **kwargs)
+
+        # Ensure to unflatten any leading dimensions.
+        if shape:
+            # Account for the additive dimension if necessary.
+            if result.size > np.prod(shape):
+                shape += (-1,)
+            result = result.reshape(shape)
+
+        return result
+
+    return inner_stat
+
+
 def _calc_percentile(data, percent, fast_percentile_method=False, **kwargs):
     """
     Calculate percentiles along the trailing axis of a 1D or 2D array.
@@ -1226,8 +1262,9 @@ def _calc_percentile(data, percent, fast_percentile_method=False, **kwargs):
         return ma.MaskedArray(result)
 
 
+@_axis_to_single_trailing
 def _percentile(
-    data, axis, percent, fast_percentile_method=False, lazy=False, **kwargs
+    data, percent, fast_percentile_method=False, lazy=False, **kwargs
 ):
     """
     The percentile aggregator is an additive operation. This means that
@@ -1245,18 +1282,6 @@ def _percentile(
         masked arrays.
 
     """
-    # Get data as a 1D or 2D view with the target axis as the trailing one.
-    if not isinstance(axis, Iterable):
-        axis = (axis,)
-    end = range(-len(axis), 0)
-
-    data = np.moveaxis(data, axis, end)
-    shape = data.shape[: -len(axis)]
-    if shape:
-        data = data.reshape(np.prod(shape), -1)
-    else:
-        data = data.flatten()
-
     if not isinstance(percent, Iterable):
         scalar_percent = True
         percent = [percent]
@@ -1282,12 +1307,6 @@ def _percentile(
             **kwargs,
         )
 
-    # Ensure to unflatten any leading dimensions.
-    if shape:
-        # Account for the additive dimension.
-        if percent.shape > (1,):
-            shape += percent.shape
-        result = result.reshape(shape)
     # Check whether to reduce to a scalar result, as per the behaviour
     # of other aggregators.
     if result.shape == (1,) and scalar_percent:
--- a/lib/iris/analysis/__init__.py
+++ b/lib/iris/analysis/__init__.py
@@ -1198,14 +1198,22 @@ def _percentile(
         masked arrays.
 
     """
-    # Ensure that the target axis is the last dimension.  Note that np.rollaxis
-    # returns a dask array if data is dask array.
-    data = np.rollaxis(data, axis, start=data.ndim)
+    # Express axis as list of non-negative integers.
+    if not isinstance(axis, Iterable):
+        axis = [axis % data.ndim]
+    else:
+        axis = sorted([dim % data.ndim for dim in axis])
 
-    shape = data.shape[:-1]
+    # Get data as a 1D or 2D view with the target axis as the trailing one.
+    end_size = np.prod([data.shape[dim] for dim in axis])
+    untouched_dims = [dim for dim in range(data.ndim) if dim not in axis]
+    shape = [data.shape[dim] for dim in untouched_dims]
+
+    data = data.transpose(untouched_dims + axis)
+
     # Flatten any leading dimensions.
     if shape:
-        data = data.reshape([np.prod(shape), data.shape[-1]])
+        data = data.reshape([np.prod(shape), end_size])
 
     if not isinstance(percent, Iterable):
         scalar_percent = True
--- a/lib/iris/analysis/__init__.py
+++ b/lib/iris/analysis/__init__.py
@@ -719,7 +719,7 @@ class PercentileAggregator(_Aggregator):
             None,
             functools.partial(_percentile, lazy=False),
             units_func=units_func,
-            lazy_func=functools.partial(_percentile, lazy=True),
+            lazy_func=_lazy_percentile,
             **kwargs,
         )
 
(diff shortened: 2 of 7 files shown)


lib/iris/tests/unit/analysis/test_PercentileAggregator.py
```diff
@@ -15,25 +15,22 @@
...
         units_func = mock.sentinel.units_func
-        lazy_func = mock.sentinel.lazy_func
-        aggregator = PercentileAggregator(
-            units_func=units_func, lazy_func=lazy_func
-        )
+        aggregator = PercentileAggregator(units_func=units_func)
         self.assertEqual(aggregator.name(), name)
-        self.assertIs(aggregator.call_func, call_func)
         self.assertIs(aggregator.units_func, units_func)
-        self.assertIs(aggregator.lazy_func, lazy_func)
```
rcomer, line 36: Removed these assertions as they failed for obvious reasons, and I'm not clear why it matters exactly what the `call_func` and `lazy_func` are (as long as the output of the function is correct).  Happy to re-jig and reinstate if someone explains their purpose.

lib/iris/analysis/__init__.py
```diff
@@ -1169,46 +1241,66 @@ def _percentile(data, axis, percent, fast_percentile_method=False, **kwargs):
...
         # Account for the additive dimension.
         if percent.shape > (1,):
             shape += percent.shape
         result = result.reshape(shape)
     # Check whether to reduce to a scalar result, as per the behaviour
     # of other aggregators.
-    if result.shape == (1,) and quantiles.ndim == 0:
-        result = result[0]
+    if result.shape == (1,) and scalar_percent:
+        result = np.squeeze(result)
```
rcomer, line 1294: Changed this to `np.squeeze` because, having forced `percent` to be a list before the call to `_calc_percentile`, this loop was now getting triggered for masked array types with the fast_percentile_method.  If you index a size (1,) masked array, you get a plain numpy array, which was causing failures e.g. [here](https://github.com/SciTools/iris/blob/2dba026c6c37cfd27f5014dbce864965c0e13642/lib/iris/tests/test_analysis.py#L604).

pp-mo:
>  dask.array.percentile exists for 1D arrays

It's an approximate solution, maybe something like an adaptive histogram ?
It seems a bit weird that it does not expand to the other dims, and I can't see how to add that with any efficiency (i.e. without a Python loop).
So for both these reasons, I think not relevant.

Peter9192:
I just tried to use this branch with the new multimodel statistics implementation in ESMValCore (https://github.com/ESMValGroup/ESMValCore/pull/968). It works, and manages small tasks pretty well! A very simple regression tests suggest that this implementation produces the same output as the master branch.

One small note is that the function doesn't seem to scale very well. Performance deteriorates with the number of chunks, and also with chunk size. I think this is not new though; it seems to come from the masked quantiles implementation. Maybe you could add a note in the documentation about how the performance depends on the mask property? 

It would be helpful for the ESMValCore development if this PR is merged. Can you provide an estimate of what is still needed to get it across the finish line?

lib/iris/analysis/__init__.py
```diff
@@ -1152,7 +1200,31 @@ def inner_stat(array, axis=-1, mdtol=None, **kwargs):
...
+    """
+    if fast_percentile_method:
+        msg = "Cannot use fast np.percentile method with masked array."
+        if ma.is_masked(data):
+            raise TypeError(msg)
+        result = np.percentile(data, percent, axis=-1)
+        result = result.T
+    else:
+        quantiles = percent / 100.0
+        result = scipy.stats.mstats.mquantiles(
```
Peter9192, line 1216: I saw your example in https://github.com/SciTools/iris/issues/4039 where you fill the masked data with NaN, so you can use a nanmedian function. I wonder if the same approach would work here, as numpy does have a nanpercentile function.

Peter9192: Maybe that's beyond the scope of this PR, though. I just tried to understand what's going on.

rcomer: The idea of using `np.nanpercentile` for the masked arrays is attractive, since it has the same calling sequence as `np.percentile` (used in our "fast percentile method") so we could simplify our API.  However the scipy `mquantiles` function does seem to be a lot more flexible and we would be losing functionality for users who care about particular interpolation methods.  So I think we'd need to retain it somehow.

As you say, I think it's out of scope for this PR, but useful to add to a wider discussion about how to organise the various percentile options (#3294).

rcomer:
Hi @Peter9192, thanks for testing this!  Thanks also for the tips on chunking - I've got some user code I'm testing this with, so will try your approach of rechunking to a larger size in there.  From the discussion at https://github.com/ESMValGroup/ESMValCore/pull/968, it looks like the issue is not specific to percentiles.  Are you saying it's worse with percentiles?  Just wondering how/where would be the best place to document this.

I can't really answer how close this is to finished: this one is something of a learning project for me so potentially it might take a few iterations!  My guess is it would go into Iris 3.1, which is scheduled for September https://github.com/SciTools/iris/milestones.  I'll see if I can find someone who knows more...

rcomer:
@bjlittle has confirmed that Iris 3.1 in September is the most likely release for this one.

Peter9192:
Hi @rcomer indeed we're seeing that performance is quite sensitive to size of the input data and chunk configuration for all aggregators. However this effect seems to much worse for percentiles. That's why I thought the masked quantiles calculation might be the culprit, especially after seeing the inline comment about "Cannot use fast np.percentile method with masked array."

I now noticed you already highlighted the fact that this "fast method" is not documented in #3294, so I guess that's exactly what I was looking for. I'd also be interested to find out exactly why scipy is apparently so much slower, because I think this might quickly become a bottleneck for the cases I was testing.

pp-mo:
I may be out of order here, but have you checked how the chunking affects the calculation?
Rank statistic measures like percentile (and by implication median) have special problems with parallelism, beyond what you get with linear measures like mean/min/max/stdev/variance/stdev, because the basic underlying calculation **cannot be chunked**.

This is fundamental, because rank statistics are based on a complete sort of the input data, which means you can't perform the calculation in 'sections'.
For example, consider an array. length 9, with values [1, 2, 2, 1, 1, 2, 1, 2]
the mean is sum(a) / 9 = 13/9, but this is equal to (sum(a[0:3]) + sum(a[3:6] + sum(a[6:10]) / 9
The **median** value is 1, because there are 5 1's and only 4 2's.  
But you have to put all the original numbers in order to work this out, i.e. you can't divide a into sections, calculate a scalar statistic over each section, and combine the results for the total result.

You can see this is not possible, because changing any one input value can change the result by an arbitrary amount.
E.G. median([0, 0, 0, 0, 1,10, 10, 10, 10] = 1, 
but median([0, 0, 0, 0, 9, 10, 10, 10, 10] = 9

So, in Dask terms, chunking is not possible along the axis of calculation of percentile, or any such rank statistic

Peter9192:
@pp-mo you're absolutely right. Let me elaborate. What I'm after is the following:

I have a big cube with many dimensions, e.g. "multimodel, time, lat, lon". I want to compute the 90th percentile of the multimodel dimension. So I chunked my data as follows: no chunking on the multimodel dim (like you say, this is not possible), reasonable block sizes on the time dimension, no chunking on the remaining dims. 

What I want/expect is that iris takes one chunk at a time, collapses it (computing the 90th percentile for that chunk), and finally combines the results. As far as I understand, this is exactly what the PR tries to achieve, right?

rcomer:
> What I want/expect is that iris takes one chunk at a time, collapses it (computing the 90th percentile for that chunk), and finally combines the results. As far as I understand, this is exactly what the PR tries to achieve, right?

That's what I'm aiming for, yes!

rcomer:
Rebased and updated the `PERCENTILE` docstring.

lib/iris/analysis/__init__.py
```diff
@@ -1169,46 +1241,66 @@ def _percentile(data, axis, percent, fast_percentile_method=False, **kwargs):
...
+
+    # Get data as a 1D or 2D view with the target axis as the trailing one.
+    end_size = np.prod([data.shape[dim] for dim in axis])
+    untouched_dims = [dim for dim in range(data.ndim) if dim not in axis]
+    shape = [data.shape[dim] for dim in untouched_dims]
+
+    data = data.transpose(untouched_dims + axis)
+
+    if shape or len(axis) > 1:
+        data = data.reshape([np.prod(shape), end_size])
```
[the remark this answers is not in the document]
rcomer: I've gone ahead and factored out the array reshaping stuff as a decorator.  Even if it's never used for other functions, I think the separation of concerns is good to have.

rcomer:
> I'd also be interested to find out exactly why scipy is apparently so much slower

Looking at the [scipy code](https://github.com/scipy/scipy/blob/v1.7.1/scipy/stats/mstats_basic.py#L2659-L2784), it implements a function for 1D arrays and then just uses `apply_along_axis`.  I think `apply_along_axis` is a python loop, so I guess this accounts for at least some of the slowness.

rcomer:
I think this is now ready for review.  The failing doctests and link checks appear to be because the Natural Earth server is down.

lib/iris/tests/unit/analysis/test__axis_to_single_trailing.py
```diff
@@ -0,0 +1,150 @@
...
+        result = wrapped_stat_func(data, axis=axis)
+        self.assertTrue(is_lazy_data(result))
+        self.assertArrayEqual(as_concrete_data(result), expected)
+
+    def test_3d_input_middle_multiple_stat(self):
+        # result shape should match non-aggregated input dims, plus trailing dim
+        # with size determined by the stat function.
+        data = np.empty((2, 3, 4))
+        axis = 1
+        self.stat_func.return_value = np.arange(8 * 5)
```
wjbenfold, line 142: Shouldn't this then be reshaped to (8, 5)? That was my understanding of the output of a stats function that adds a trailing dim

lib/iris/tests/unit/analysis/test_PERCENTILE.py
```diff
@@ -97,6 +98,170 @@ def test_masked_2d_multi(self):
         self.assertArrayAlmostEqual(actual, expected)
 
 
+class LazyMixin:
```
wjbenfold, line 101: Before I read this file, I was wondering if we could leverage the existing tests and just run them once with real data and once with lazy data to show that the effect is identical - is that possible? We could also cover off real fast percentile aggregation, which I think we support but isn't currently tested?

lib/iris/analysis/__init__.py
rcomer: I tried replacing `np.squeeze` with `array.item()` and got the following failures:

```
FAILED lib/iris/tests/test_analysis.py::TestAggregators::test_fast_percentile_1d_75_percent_masked_type_no_mask - AssertionError: <class 'numpy.ndarray'> !...
FAILED lib/iris/tests/test_analysis.py::TestAggregators::test_fast_percentile_2d_two_coords_masked_type_no_mask - AssertionError: <class 'numpy.ndarray'> !...
FAILED lib/iris/tests/unit/analysis/test_PERCENTILE.py::Test_aggregate::test_1d_single - AttributeError: 'float' object has no attribute 'shape'
FAILED lib/iris/tests/unit/analysis/test_PERCENTILE.py::Test_aggregate::test_masked_1d_single - AttributeError: 'float' object has no attribute 'shape'
FAILED lib/iris/tests/unit/analysis/test_PERCENTILE.py::Test_lazy_fast_aggregate::test_1d_single - AttributeError: 'Array' object has no attribute 'item'
FAILED lib/iris/tests/unit/analysis/test_PERCENTILE.py::Test_lazy_aggregate::test_1d_single - AttributeError: 'Array' object has no attribute 'item'
FAILED lib/iris/tests/unit/analysis/test_PERCENTILE.py::Test_lazy_aggregate::test_masked_1d_single - AttributeError: 'Array' object has no attribute 'item'
```

rcomer: > It seems to me that if the returned cube is non-scalar, then the "percentile_over_foo" coordinate should be on its dimension. So we should either fix the coordinate for this case (and add a test) or take out the scalar_percent. Do you have a preference?

I just checked what happens if you do similarly with a 2D coordinate

```python
cube = iris.cube.Cube(np.arange(10).reshape(2, 5))
coord1 = iris.coords.DimCoord([2, 3], long_name="foo")
coord2 = iris.coords.DimCoord(range(5), long_name="bar")
cube.add_dim_coord(coord1, 0)
cube.add_dim_coord(coord2, 1)

print(cube.collapsed("foo", iris.analysis.PERCENTILE, percent=[50]))
```

```
unknown / (unknown)                 (bar: 5)
    Dimension coordinates:
        bar                             x
    Scalar coordinates:
        foo                         2, bound=(2, 3)
        percentile_over_foo         50 percent

```

So I think for consistency it makes sense to take out the `scalar_percent` check, as you originally suggested.

lib/iris/analysis/__init__.py
```diff
@@ -1751,14 +1859,26 @@ def interp_order(length):
...
 **For example**:
 
 To compute the 10th and 90th percentile over *time*::
 
     result = cube.collapsed('time', iris.analysis.PERCENTILE, percent=[10, 90])
 
-This aggregator handles masked data.
+This aggregator handles masked data and lazy data.
+
+.. note::
```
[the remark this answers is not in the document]
rcomer: It renders OK for me locally but what you suggest is consistent with other "notes" in this module, so done.

lib/iris/tests/unit/analysis/test_PercentileAggregator.py
```diff
@@ -126,6 +122,22 @@ def test_multi_multiple_points(self):
...
+        kwargs = dict(percent=percent)
+        shape = self.cube_multi.shape + percent.shape
+        data = da.arange(np.prod(shape)).reshape(shape)
+        coords = [self.coord_multi_0]
+        actual = aggregator.post_process(
+            self.cube_multi, data, coords, **kwargs
+        )
+        self.assertEqual(actual.shape, percent.shape + self.cube_multi.shape)
+        self.assertTrue(actual.has_lazy_data())
+        expected = np.rollaxis(as_concrete_data(data), -1)
```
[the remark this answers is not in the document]
rcomer: A lot of these tests I copied from the non-lazy ones, as you have probably guessed.

lib/iris/_lazy_data.py
```diff
@@ -369,10 +369,15 @@ def map_complete_blocks(src, func, dims, out_sizes):
         Output size of dimensions that cannot be chunked.
 
     """
-    if not src.has_lazy_data():
+    if is_lazy_data(src):
+        data = src
+    elif not hasattr(src, "has_lazy_data"):
```
rcomer, line 374: Not sure whether this is the best way to check for a plain array.  I know that previously explicitly checking types has been discouraged (hence just checking for `compute` attribute in `_is_lazy_data`).

rcomer:
Thanks @wjbenfold for the thorough review.  There are still a few things to address in the tests, but I thought it might be worth you checking what I've done in the main code first as obviously that can affect what happens with the tests.  Still to think about:

- [ ] refactor tests to re-use methods for lazy and non-lazy cases
- [ ] add tests to check **kwargs are passed down to scipy (maybe mock the scipy function?)
- [ ] maybe replace `rollaxis` with `moveaxis`

lib/iris/analysis/__init__.py
wjbenfold: Yep, I think I agree (though I was swayed by your earlier argument). I was tempted to say that when a percent list of length 1 was provided we should make that a coordinate of length 1, but I'm pretty sure that's what a scalar coordinate is.

So I'm happy to lose `scalar_percent`

lib/iris/analysis/__init__.py
```diff
@@ -1152,7 +1200,68 @@ def inner_stat(array, axis=-1, mdtol=None, **kwargs):
...
+            data, quantiles, axis=-1, **kwargs
+        )
+    if not ma.isMaskedArray(data) and not ma.is_masked(result):
+        return np.asarray(result)
+    else:
+        return ma.MaskedArray(result)
+
+
+@_axis_to_single_trailing
+def _percentile(
```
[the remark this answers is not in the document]
wjbenfold: > Presumably having `map_complete_blocks` handle a numpy array may also prove useful for other aggregators.

Yep, my run-length aggregator will use it :)

lib/iris/_lazy_data.py
wjbenfold: I think I like it as a test because you're testing for the exact thing that will cause an attribute error if you let it through (because the next check is `src.has_lazy_data()`)?

lib/iris/analysis/__init__.py
```diff
@@ -760,6 +756,45 @@ def aggregate(self, data, axis, **kwargs):
...
+
+        * data (array):
+            A lazy array (:class:`dask.array.Array`).
+
+        * axis (int or list of int):
+            The dimensions to aggregate over -- note that this is defined
+            differently to the 'aggregate' method 'axis' argument, which only
+            accepts a single dimension index.
+
+        Kwargs:
```
wjbenfold, line 779: Should `mdtol` be in this list like it's in the `aggregate` one?

lib/iris/analysis/__init__.py
```diff
@@ -1263,13 +1263,25 @@ def _percentile(data, percent, fast_percentile_method=False, **kwargs):
     If a new additive dimension is formed, then it will always be the last
     dimension of the resulting percentile data payload.
 
+    Args:
+
+    * data (array-like)
+        array from which percentiles are to be calculated
+    * axis (int or tuple of int)
+        Axis or axes along which the percentiles are computed
+        (handled by decorator)
```
wjbenfold, line 1272: It feels weird to describe an argument in the docstring that doesn't feature in the signature. Is this the normal way of doing that?

wjbenfold (commented):
I think I'm caught up on review changes now @rcomer

lib/iris/analysis/__init__.py
rcomer: Mmmm not sure, and I dithered over it before I did it!  Happy to take it out.

lib/iris/tests/unit/analysis/test_PercentileAggregator.py
```diff
@@ -85,7 +81,7 @@ def test_simple_multiple_points(self):
             self.cube_simple, data, coords, **kwargs
         )
         self.assertEqual(actual.shape, percent.shape + self.cube_simple.shape)
-        expected = np.rollaxis(data, -1)
+        expected = data.T
```
rcomer, line 84: This array is only 2D so seems simpler to just use transpose.

rcomer (commented):
Hi @wjbenfold I think I've now covered everything.  `test_PERCENTILE.py` now looks a lot cleaner to me with the refactor.  Where methods are now defined on different objects I moved and modified them in separate commits, so hopefully it's relatively easy to see what actually changed.

lib/iris/tests/unit/analysis/test_PERCENTILE.py
```diff
@@ -9,92 +9,276 @@
...
+        )
+
+    @mock.patch("scipy.stats.mstats.mquantiles")
+    def test_default_kwargs_passed(self, mocked_mquantiles):
+        data = np.arange(5)
+        percent = 50
+        axis = 0
+        self.agg_method(data, axis=axis, percent=percent)
+        for key in ["alphap", "betap"]:
+            assert mocked_mquantiles.call_args.kwargs[key] == 1
```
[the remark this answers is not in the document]
rcomer: Whooops! Done.

> symptomatic of having to write unittest-style tests when you're used to pytest.

I'll take that excuse 😅

rcomer:
[Rebased as the tests didn't run after my final commit]

lib/iris/tests/unit/analysis/test_PERCENTILE.py
wjbenfold: This happened, it was pretty sweet

wjbenfold (approved):

rcomer:
Thanks for your epic reviewing efforts @wjbenfold, made it much better 🍻

trexfeathers:
@EastaughK after the next Iris release you and your team will be able to get percentiles without the memory concerns you brought to the Surgery 🙂

PR #3914, opened 2020-10-26
Cube print respin

pp-mo:
Now with no specific reference to ugrid.

This is now an extensive refactor of the Iris Cube `summary` and `_repr_html_` methods.
Including changes needed to enable the forthcoming iris-ugrid `UCube`s to produce "extended" cube summmaries.

I have added testing for the new factored-out helper methods.  But I haven't provided any test of how modifying these can produce extended cube summaries, as its not really an "API" + the requirement is hard to generalise.
That usage will eventually get tested by the specifics in iris-ugrid.

Meanwhile, you should see that all the existing tests simply pass. :bow_and_arrow: 

Proposals for the specific use of this in iris-ugrid are here : https://github.com/SciTools-incubator/iris-ugrid/pull/26

### DIFF
--- a/lib/iris/cube.py
+++ b/lib/iris/cube.py
@@ -2318,14 +2318,13 @@ bound=(1994-12-01 00:00:00, 1998-12-01 00:00:00)
         dim_coords = self.dim_coords
         aux_coords = self.aux_coords
         all_coords = dim_coords + aux_coords + derived_coords
-        scalar_coords = [
-            coord
+        scalar_coord_ids = [
+            id(coord)
             for coord in all_coords
             if not self.coord_dims(coord) and coord.shape == (1,)
         ]
         # Determine the cube coordinates that are not scalar BUT
         # dimensioned.
-        scalar_coord_ids = set(map(id, scalar_coords))
         vector_dim_coords = [
             coord for coord in dim_coords if id(coord) not in scalar_coord_ids
         ]
@@ -2344,20 +2343,14 @@ bound=(1994-12-01 00:00:00, 1998-12-01 00:00:00)
         ]
 
         # Ancillary Variables
-        vector_ancillary_variables = [av for av in self.ancillary_variables()]
+        vector_ancillary_variables = [
+            av for av in self.ancillary_variables() if av.shape != (1,)
+        ]
 
-        # Sort scalar coordinates by name.
-        scalar_coords.sort(key=lambda coord: coord.name())
-        # Sort vector coordinates by data dimension and name.
-        vector_dim_coords.sort(
-            key=lambda coord: (self.coord_dims(coord), coord.name())
-        )
-        vector_aux_coords.sort(
-            key=lambda coord: (self.coord_dims(coord), coord.name())
-        )
-        vector_derived_coords.sort(
-            key=lambda coord: (self.coord_dims(coord), coord.name())
-        )
+        # Sort vector coordinates by data dimension.
+        vector_dim_coords.sort(key=self.coord_dims)
+        vector_aux_coords.sort(key=self.coord_dims)
+        vector_derived_coords.sort(key=self.coord_dims)
         Spec = self._VectorSectionSpec
         section_specs = [
             Spec("Dimension coordinates", vector_dim_coords, True),
--- a/lib/iris/cube.py
+++ b/lib/iris/cube.py
@@ -2469,58 +2469,26 @@ bound=(1994-12-01 00:00:00, 1998-12-01 00:00:00)
                     ),
                 )
 
-            if vector_dim_coords:
-                dim_coord_summary, cube_header = vector_summary(
-                    vector_dim_coords,
-                    cube_header,
-                    max_line_offset,
-                    add_extra_lines=True,
-                )
-                summary += "\n     Dimension coordinates:\n" + "\n".join(
-                    dim_coord_summary
-                )
-
-            if vector_aux_coords:
-                aux_coord_summary, cube_header = vector_summary(
-                    vector_aux_coords,
-                    cube_header,
-                    max_line_offset,
-                    add_extra_lines=True,
-                )
-                summary += "\n     Auxiliary coordinates:\n" + "\n".join(
-                    aux_coord_summary
-                )
-
-            if vector_derived_coords:
-                derived_coord_summary, cube_header = vector_summary(
-                    vector_derived_coords,
-                    cube_header,
-                    max_line_offset,
-                    add_extra_lines=True,
-                )
-                summary += "\n     Derived coordinates:\n" + "\n".join(
-                    derived_coord_summary
-                )
-
             #
-            # Generate summary of cube cell measures attribute
+            # Generate textual summary for each type of vector component.
             #
-            if vector_cell_measures:
-                cell_measure_summary, cube_header = vector_summary(
-                    vector_cell_measures, cube_header, max_line_offset,
-                )
-                summary += "\n     Cell measures:\n"
-                summary += "\n".join(cell_measure_summary)
-
-            #
-            # Generate summary of cube ancillary variables attribute
-            #
-            if vector_ancillary_variables:
-                ancillary_variable_summary, cube_header = vector_summary(
-                    vector_ancillary_variables, cube_header, max_line_offset,
-                )
-                summary += "\n     Ancillary variables:\n"
-                summary += "\n".join(ancillary_variable_summary)
+            vector_summary_specs = [
+                ("Dimension coordinates", vector_dim_coords, True),
+                ("Auxiliary coordinates", vector_aux_coords, True),
+                ("Derived coordinates", vector_derived_coords, True),
+                ("Cell measures", vector_cell_measures, False),
+                ("Ancillary variables", vector_ancillary_variables, False),
+            ]
+            for title, elements, add_extra_lines in vector_summary_specs:
+                if elements:
+                    summary_lines, cube_header = vector_summary(
+                        vector_items=elements,
+                        cube_header=cube_header,
+                        max_line_offset=max_line_offset,
+                        add_extra_lines=add_extra_lines,
+                    )
+                    summary += f"\n     {title}:\n"
+                    summary += "\n".join(summary_lines)
 
             #
             # Generate textual summary of cube scalar coordinates.
--- a/lib/iris/cube.py
+++ b/lib/iris/cube.py
@@ -2289,6 +2289,85 @@ bound=(1994-12-01 00:00:00, 1998-12-01 00:00:00)
 
         return name
 
+    _VectorSectionSpec = namedtuple(
+        "Spec", ("title", "elements", "add_extra_lines")
+    )
+
+    def _summary_vector_sections_info(self):
+        """
+        Return a table describing the order and content of the 'vector'
+        sections of the cube summary.
+
+        Returns:
+
+            * summary_info (list of cube._VectorSectionSpec)
+                A list of summary sections info, in summary order.
+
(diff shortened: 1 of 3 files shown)


pp-mo:
Hi @trexfeathers 
Hoping you can do this ?  It's a bit of a big lump to swallow in one, but I've tried to make the individual commit changes intelligible.

trexfeathers:
>Hi @trexfeathers
Hoping you can do this ?

@pp-mo I'm on it now 👀 sorry for the delay.

trexfeathers:
Note to self: have so far reviewed up to and including: 988710ca9a50cb161bf40b5ac99931c42ee6632b

lib/iris/cube.py
```diff
@@ -2263,35 +2263,125 @@ def _summary_extra(self, coords, summary, indent):
...
+        ]
+
+        # Ancillary Variables
+        vector_ancillary_variables = [av for av in self.ancillary_variables()]
+
+        # Sort scalar coordinates by name.
+        scalar_coords.sort(key=lambda coord: coord.name())
+        # Sort vector coordinates by data dimension and name.
+        vector_dim_coords.sort(
+            key=lambda coord: (self.coord_dims(coord), coord.name())
```
trexfeathers, line 2353:
```suggestion
            key=lambda coord: (self.coord_dims(coord))
```

Isn't this redundant, since it's not possible to have multiple `DimCoord`s against a single dimension?

lib/iris/tests/unit/cube/test_Cube.py
```diff
@@ -2434,5 +2434,124 @@ def test_cell_method_correct_order(self):
...
+
+    def test_dim_aux_coords(self):
+        cube = Cube(np.zeros((3, 5, 4)))
+        cube.add_dim_coord(
+            DimCoord(np.arange(3), long_name="lats", units="degrees"), 0
+        )
+        cube.add_aux_coord(
+            DimCoord(np.arange(5), standard_name="longitude", units="degrees"),
+            1,
+        )
```
trexfeathers, line 2504: After this: add a second, almost identical instance with an alphabetically earlier name to test the [sorting behaviour](https://github.com/SciTools/iris/pull/3914/files#diff-38eaa350bf479c486e9f6659b8077be2d52448f17739872e50285b98b5423415R2355-R2357).

lib/iris/tests/unit/cube/test_Cube.py
```diff
@@ -2434,5 +2434,124 @@ def test_cell_method_correct_order(self):
...
+        cube = Cube(np.zeros((5, 3)))
+        cube.add_aux_coord(
+            DimCoord(np.arange(5), standard_name="longitude", units="degrees"),
+            0,
+        )
+        result = cube._summary_dim_name(0)
+        self.assertEqual(result, "-- ")
+
+
+class Test__summary_vector_sections_info(tests.IrisTest):
```
trexfeathers, line 2472: Why nothing for [`derived_coords`](https://github.com/SciTools/iris/pull/3914/files#diff-38eaa350bf479c486e9f6659b8077be2d52448f17739872e50285b98b5423415R2315)?

trexfeathers, line 2472: Need something to confirm the [specialised handling of `scalar_coords`](https://github.com/SciTools/iris/pull/3914/files#diff-38eaa350bf479c486e9f6659b8077be2d52448f17739872e50285b98b5423415R2321-R2339).

lib/iris/tests/unit/cube/test_Cube.py
```diff
@@ -2434,5 +2434,124 @@ def test_cell_method_correct_order(self):
         self.assertTrue(cube1 == cube2)
 
 
+class Test__summary_dim_name(tests.IrisTest):
```
trexfeathers, line 2437: Need a test for the [multiple dimension coords mapping `ValueError`](https://github.com/SciTools/iris/pull/3914/files#diff-38eaa350bf479c486e9f6659b8077be2d52448f17739872e50285b98b5423415R2280-R2288).

trexfeathers (changes requested):
@pp-mo this looks like a **huge** improvement! 💐

A few things for you to address/answer, mainly on test coverage. Thanks!

lib/iris/cube.py
```diff
@@ -2263,35 +2263,125 @@ def _summary_extra(self, coords, summary, indent):
...
+            if id(coord) not in scalar_coord_ids
+        ]
+
+        # cell measures
+        vector_cell_measures = [
+            cm for cm in self.cell_measures() if cm.shape != (1,)
+        ]
+
+        # Ancillary Variables
+        vector_ancillary_variables = [av for av in self.ancillary_variables()]
```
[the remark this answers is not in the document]
pp-mo: Actually I didn't do that just because [the existing code doesn't](https://github.com/SciTools/iris/blob/5eaa58a1993c63bff8c4fe55535b6f8d365fd2e5/lib/iris/cube.py#L2348).
Reading that, I had thought that maybe its because we don't allow scalar ancils.
But now you raise it, and looking at the code, I fear that may be wrong.  So, I will add it here too.

lib/iris/cube.py
```diff
@@ -2263,35 +2263,125 @@ def _summary_extra(self, coords, summary, indent):
...
+
+        # Sort scalar coordinates by name.
+        scalar_coords.sort(key=lambda coord: coord.name())
```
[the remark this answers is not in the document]
pp-mo: Good spot !

Actually it's worse than that .. this variable is never used again (it's local to the routine).
To avoid a nastier API to the routine, I chose to re-calculate this separately in the main routine.
So in _this_ routine we actually only use the 'ids'. I will tidy that up.

lib/iris/cube.py
pp-mo: Another good one!

I think, like the [former code to select dim-names from multiple dim-coords](https://github.com/SciTools/iris/blob/5eaa58a1993c63bff8c4fe55535b6f8d365fd2e5/lib/iris/cube.py#L2272-L2285), which I already removed, this is really a ancient legacy of a time when we hadn't yet sorted out what was and was not allowed for dim-coords.

lib/iris/tests/unit/cube/test_Cube.py
pp-mo: I don't think so.  That is just defensive programming : logically it can never happen unless the Cube is "broken".

lib/iris/tests/unit/cube/test_Cube.py
pp-mo: Ok!

pp-mo: Ok!

pp-mo:
Many thanks @trexfeathers .  You seem to have done a very thorough job here ! :rocket: 
Hoping the latest addresses your comments.

lib/iris/cube.py
```diff
@@ -2344,20 +2343,14 @@ def _summary_vector_sections_info(self):
...
-        vector_aux_coords.sort(
-            key=lambda coord: (self.coord_dims(coord), coord.name())
-        )
-        vector_derived_coords.sort(
-            key=lambda coord: (self.coord_dims(coord), coord.name())
-        )
+        # Sort vector coordinates by data dimension.
+        vector_dim_coords.sort(key=self.coord_dims)
+        vector_aux_coords.sort(key=self.coord_dims)
+        vector_derived_coords.sort(key=self.coord_dims)
```
trexfeathers, line 2353: I don't actually know about `derived_coords`, but I was fairly certain that you _are_ allowed to have multiple `aux_coords` against a single dimension, so it's still appropriate to keep the second sort by `coord.name()`. Am I wrong?

pp-mo:
@trexfeathers 
Just to note, some of the points you raised I was originally side-stepping, because I was rather focussed on "pure refactoring".
I.E. I was sticking to "no functional change, even to fix things" , so as not to possibly break backwards-compatibility.

But in retrospect, that may be a useful reference idea, but it is over-strict : 
We don't really expect anyone's code to be undermined by a slightly different cube-summary output, especially if it is "better".
We can take the opportunity to improve some things, just as we do to add some missing testing.

lib/iris/tests/unit/cube/test_Cube.py
```diff
@@ -2517,6 +2517,30 @@ def test_dim_aux_coords(self):
...
+        cube.add_aux_coord(DimCoord(np.arange(3), long_name="a1", units=1), 0)
+        cube.add_aux_coord(DimCoord(np.arange(3), long_name="b1", units=1), 0)
+        cube.add_aux_coord(DimCoord(np.arange(4), long_name="b2", units=1), 1)
+        cube.add_aux_coord(DimCoord(np.arange(4), long_name="a2", units=1), 1)
+        result = cube._summary_vector_sections_info()
+        self.assertEqual(
+            [[co.name() for co in sect[1]] for sect in result],
+            [
+                [],
+                ["a1", "b1", "a12", "b12", "a2", "b2"],
```
trexfeathers, line 2537: This would need changing depending on [my previous comment regarding `vector_aux_coords.sort()`](https://github.com/SciTools/iris/pull/3914/commits/5c9ace49f3080a53bf1eafe691a81c8b96990818#diff-38eaa350bf479c486e9f6659b8077be2d52448f17739872e50285b98b5423415R2352).

lib/iris/tests/unit/cube/test_Cube.py
```diff
@@ -2517,6 +2517,30 @@ def test_dim_aux_coords(self):
             ],
         )
 
+    def test_aux_coords_ordering(self):
```
trexfeathers, line 2520: Would you need the same for `derived_coords`?

lib/iris/cube.py
trexfeathers: @pp-mo has modified but not certain we're there yet. See: https://github.com/SciTools/iris/pull/3914#discussion_r514334619

lib/iris/tests/unit/cube/test_Cube.py
trexfeathers: Turns out this is co-incidentally correct even if we included alphabetical sorting!

lib/iris/cube.py
trexfeathers: We have now discovered that **all 5** of the components in `section_specs` are intrinsically sorted since that's the way they come out of the `Cube`!

lib/iris/tests/unit/cube/test_Cube.py
trexfeathers: Having discussed with @pp-mo, we've agreed that all we need to do is convert this test into a _general_ confirmation of correct sorting, by:
* Changing the test name
* Adding a few `dim_coord`s

pp-mo:
Since @trexfeathers prompting above, I have realised that *none* of the element sorting is actually needed, as all the relevant cube access methods provide results already sorted, by (cube-dims, name), which is exactly what we wanted.  
E.G. [cube.aux_coords](https://scitools-iris.readthedocs.io/en/latest/generated/api/iris/cube.html#iris.cube.Cube.aux_coords).

So, I have now removed all that sorting code.

Also I have now rationalised all the scalar-element testing : "Scalar" elements are now consistently those with an empty "`.cube_dims(cube)`"  -- any testing for a shape of (1,) I think was legacy which is now doing simply nothing. 

ALSO, I've added a possible "scalar ancillary variables" section, which was previously missing.

trexfeathers:
So long as the tests pass, I will be happy with this..!

trexfeathers (approved):

PR #3922, opened 2020-11-13
Cube._summary_coord_extra: efficiency and bugfix

### LINKED ISSUE #3921: Printing similar coordinates when one has numpy array attributes
## 🐛 Bug Report
If you print a cube which has two similar coordinates, there is some logic which (as far as I can tell) creates a string to distinguish the two coordinates.  This logic fails if one of the coordinates happens to have attributes whose values are numpy arrays.

## How To Reproduce
Minimal example:

```python
import iris
import numpy as np

cube = iris.cube.Cube(range(5))
coord1 = iris.coords.DimCoord(
    range(5), long_name='foo', attributes=dict(bar=np.array([7, 42])))
coord2 = coord1.copy()
coord2.units = 'K'

cube.add_dim_coord(coord1, 0)
cube.add_aux_coord(coord2, 0)

print(cube)
```
Result:
```
  File "iris3_cube_print_fail.py", line 13, in <module>
    print(cube)
  File "[site-packages-path]/lib/python3.6/site-packages/iris/cube.py", line 2599, in __str__
    return self.summary()
  File "[site-packages-path]/lib/python3.6/site-packages/iris/cube.py", line 2445, in summary
    vector_dim_coords, cube_header, max_line_offset
  File "[site-packages-path]/lib/python3.6/site-packages/iris/cube.py", line 2423, in vector_summary
    vectors, vector_summary, extra_indent
  File "[site-packages-path]/lib/python3.6/site-packages/iris/cube.py", line 2215, in _summary_extra
    extra = self._summary_coord_extra(coord, indent)
  File "[site-packages-path]/lib/python3.6/site-packages/iris/cube.py", line 2198, in _summary_coord_extra
    if attributes.setdefault(key, value) != value:
ValueError: The truth value of an array with more than one element is ambiguous. Use a.any() or a.all()
```

## Expected behaviour
I should be able to print this cube.

## Environment 
 - OS & Version: RHEL7.9
 - Iris Version: v3 release candidate

rcomer:
### Description
I started looking at this method with a view to fixing #3921, but once I looked couldn't ignore the fact that this method seemed a bit inefficient: we ultimately only want the elements of `vary` that exist as keys in `coord`'s attributes dictionary, so I think we should only loop through those keys to begin with.

I've added one simple test for the #3921 case.  Note there is also one existing test for this method in the [legacy tests](https://github.com/SciTools/iris/blob/v3.0.x/lib/iris/tests/test_cdm.py#L329).

I'd like to get this into Iris v3 if possible, as I believe lenient cube arithmetic will increase the prevalence of Frankencubes*.  So this method will get called more often.  And the change is only small :innocent: 

Closes #3921.

---
***Frankencube** _noun_: a cube with multiple coordinates that have the same name.  Coined at https://github.com/ESMValGroup/ESMValCore/issues/373

### DIFF
--- a/docs/iris/src/whatsnew/3.0.rst
+++ b/docs/iris/src/whatsnew/3.0.rst
@@ -168,7 +168,11 @@ This document explains the changes made to Iris for this release
   Previously, the first tick label would occasionally be duplicated. This also
   removes the use of Matplotlib's deprecated ``IndexFormatter``. (:pull:`3857`)
 
-* `@znicholls`_ fixed :meth:`~iris.quickplot._title` to only check ``units.is_time_reference`` if the ``units`` symbol is not used. (:pull:`3902`)
+* `@znicholls`_ fixed :meth:`~iris.quickplot._title` to only check 
+  ``units.is_time_reference`` if the ``units`` symbol is not used. (:pull:`3902`)
+
+* `@rcomer`_ fixed a bug whereby numpy array type attributes on a cube's 
+  coordinates could prevent printing it.  See :issue:`3921`.  (:pull:`3922`)
 
 .. _whatsnew 3.0 changes:
 
--- a/lib/iris/cube.py
+++ b/lib/iris/cube.py
@@ -2189,11 +2189,14 @@ bound=(1994-12-01 00:00:00, 1998-12-01 00:00:00)
                     if key not in similar_coord.attributes:
                         vary.add(key)
                         break
-                    if similar_coord.attributes[key] != value:
+                    if not np.array_equal(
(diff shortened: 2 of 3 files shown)


rcomer:
Added a further efficiency: we don't need to compare the given coordinate with itself.  Think I'm done now.

pp-mo:
Thanks @rcomer , looks useful.

Aside : could we define a Frankencube as one that has DoppleGangCoords ?

rcomer:
Thanks @pp-mo 

I see a whole new Glossary section of the docs coming on! :laughing:

PR #3932, opened 2020-11-28
Ditch `if __name__ == "__main__"` in gallery examples

rcomer:
### Description
All our gallery examples are currently set out as scripts, with a `main` function and `if __name__ == "__main__"` loop.  We now have the option to [download the examples as Jupyter notebooks](https://scitools-iris.readthedocs.io/en/latest/_downloads/76212d3ad65134e03284d4129976c811/plot_polar_stereo.ipynb) and I think the `if __name__ == "__main__"` layout looks somewhat odd in that context.  I clicked on a few examples in the Matplotlib gallery, and those do not have this layout (Cartopy gallery examples do though).

As a proof of concept, I've modified the `plot_lineplot_with_legend.py` example.  This also required modifying the gallery test, as it relied on running the `main` function.  I've never been near those tests before so happy to take advice on a better way to do them!  I've established that I still get the image comparison failure if I e.g. change the legend labels.

This particular gallery example is also included in the User Guide and is, I think, the only User Guide example that has the `if __name__ == "__main__"` layout.  I speculate that the note here may have been a response to users asking "why is this example different".  However, I think advice about general python script structure is out of scope for the Iris User Guide, so removing that note is an added bonus of modifying the example.  This note is the only place our docs link to `effbot.org`, so we no longer need to explicitly skip that in the linkchecks.

If I get a thumbs up, I will go ahead and modify the other gallery examples.

### Question

How do you do that thing of making temporary builds of the docs for review?  I think it might be useful here...

---

### DIFF
--- a/docs/iris/gallery_code/general/plot_lineplot_with_legend.py
+++ b/docs/iris/gallery_code/general/plot_lineplot_with_legend.py
@@ -11,39 +11,34 @@ import iris.plot as iplt
 import iris.quickplot as qplt
 
 
-def main():
-    fname = iris.sample_data_path("air_temp.pp")
+fname = iris.sample_data_path("air_temp.pp")
 
-    # Load exactly one cube from the given file.
-    temperature = iris.load_cube(fname)
+# Load exactly one cube from the given file.
+temperature = iris.load_cube(fname)
 
-    # We only want a small number of latitudes, so filter some out
-    # using "extract".
-    temperature = temperature.extract(
-        iris.Constraint(latitude=lambda cell: 68 <= cell < 78)
-    )
+# We only want a small number of latitudes, so filter some out
+# using "extract".
+temperature = temperature.extract(
+    iris.Constraint(latitude=lambda cell: 68 <= cell < 78)
+)
 
-    for cube in temperature.slices("longitude"):
+for cube in temperature.slices("longitude"):
 
-        # Create a string label to identify this cube (i.e. latitude: value).
-        cube_label = "latitude: %s" % cube.coord("latitude").points[0]
+    # Create a string label to identify this cube (i.e. latitude: value).
+    cube_label = "latitude: %s" % cube.coord("latitude").points[0]
 
-        # Plot the cube, and associate it with a label.
-        qplt.plot(cube, label=cube_label)
+    # Plot the cube, and associate it with a label.
+    qplt.plot(cube, label=cube_label)
 
-    # Add the legend with 2 columns.
-    plt.legend(ncol=2)
+# Add the legend with 2 columns.
+plt.legend(ncol=2)
 
-    # Put a grid on the plot.
-    plt.grid(True)
+# Put a grid on the plot.
+plt.grid(True)
 
-    # Tell matplotlib not to extend the plot axes range to nicely
-    # rounded numbers.
-    plt.axis("tight")
+# Tell matplotlib not to extend the plot axes range to nicely
+# rounded numbers.
+plt.axis("tight")
 
-    # Finally, show it.
-    iplt.show()
-
-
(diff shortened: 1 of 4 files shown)


trexfeathers:
I agree with what @rcomer has said, but unfortunately this statement also applies to me:

>I've never been near those tests before

So I can't really be the seal of approval.

jamesp:
I'm slightly torn on this one so would appreciate user thoughts, @rcomer - you're better placed than I am to say what is most useful!

On the one hand, as a script it's good practice to have ifmain. Although I'll be honest I'm not the best at sticking to that when putting together an analysis code, I'd be comfortable to see that go.

On the other, even without the ifmain, this code still looks clunky and non-idiomatic in a notebook.  All the code goes into a single cell, whereas better practice when working in a notebook would be to put e.g. imports in separate cells to logic.  Also the `.show()` is not necessary in a properly configured notebook.  

There is an option to denote what should be in a separate cell in the `.py` script  using a line of `#######` (documented [here](https://sphinx-gallery.github.io/stable/tutorials/plot_parse.html#sphx-glr-tutorials-plot-parse-py)).  But this also has the effect of rendering the html of the script in the gallery documentation with code interspersed with text, making it harder to copy-paste!

All that is to say it's probably quite hard to get a script that's going to look right in both script form and notebook form, we should go with the form that is going to be most useful to new users.

rcomer:
Hi @jamesp, thanks for your thoughts on this.  I completely agree that the everything-in-a-single-cell isn't a good way to make a notebook, and I didn't know about the lines of `######`.  Also agree though that using this would make copying from the webpage harder!

In terms of what's useful: my main use of gallery examples is to think "I need to do this thing in my code" and then go look for a plot that has something similar.  So probably I'll only copy/paste out the relevant couple of lines straight from the webpage.  That said, I'm only one user!

For a new user, it's possible that running the code a few lines at a time in a notebook could be useful, as they would with the [courses](https://github.com/SciTools/courses), but now I'm basically just speculating....

If we did decide to split these down into separate cells for the notebooks, I'm not sure I'd want to commit to tackling them all in one PR: probably some would be straightforward, but I wouldn't be surprised if some have issues (e.g. huge loops) that require a refactor before they'll work nicely separated out (obviously huge loops should be avoided anyway, but we are where we are!)

For the ifmain specifically, the only time I know it's a problem to not have it is when you want to import something into another script/module.  I can't imagine ever wanting to do that from a gallery example script though - wouldn't you just copy/paste the relevant object definition into your working script?

rcomer:
Incidentally, https://effbot.org/ appears to be having some sort of existential reevaluation.  So whatever else we do, we should definitely remove that link!

rcomer:
> If we did decide to split these down into separate cells for the notebooks, I'm not sure I'd want to commit to tackling them all in one PR

In fact, I don't want to commit to it at all: I'm pretty sure I can write a script to remove the ifmain's, but I think figuring out all the cells would be a not-very-fun manual slog through all the scripts.  I started poking this when I was at a loose end one weekend, but nobody's ever that bored...! 😆

jamesp:
Thansk @rcomer. Perhaps this is one to tackle on new gallery examples, rather than going back and editing the old ones?  I can see value in having both forms of document available, they serve different purposes and give new users an idea of how they would use in both contexts.  I'm happy to put together a notebook friendly example to add to the gallery, it's a good opportunity to play with the example data.

pp-mo:
There is another possible approach that allows you to use python code as your definite source,  and have a notebook as a derivative of that : https://jupytext.readthedocs.io/en/latest/index.html 

The key benefit is that change is managed in a human-readable format (Python), which makes controlling change and PRs much less tortuous. 

I really like the idea of this, but I haven't investigated further.

PR #3933, opened 2020-11-29
Gallery: update seasonal ensemble example

rcomer:
### Description
Iris has moved on _a lot_ since many of the gallery examples were added.  This PR simplifies the seasonal forecast example as well as dropping the old school "%s" string formatting.

---

### DIFF
--- a/docs/iris/gallery_code/meteorology/plot_lagged_ensemble.py
+++ b/docs/iris/gallery_code/meteorology/plot_lagged_ensemble.py
@@ -33,10 +33,10 @@ def realization_metadata(cube, field, fname):
     in the cube.
 
     """
-    # add an ensemble member coordinate if one doesn't already exist
+    # Add an ensemble member coordinate if one doesn't already exist.
     if not cube.coords("realization"):
-        # the ensemble member is encoded in the filename as *_???.pp where ???
-        # is the ensemble member
+        # The ensemble member is encoded in the filename as *_???.pp where ???
+        # is the ensemble member.
         realization_number = fname[-6:-3]
         realization_coord = iris.coords.AuxCoord(
             np.int32(realization_number), "realization", units="1"
@@ -45,11 +45,16 @@ def realization_metadata(cube, field, fname):
 
 
 def main():
-    # extract surface temperature cubes which have an ensemble member
-    # coordinate, adding appropriate lagged ensemble metadata
+    # Create a constraint to extract surface temperature cubes which have a
+    # "realization" coordinate.
+    constraint = iris.Constraint(
+        "surface_temperature", realization=lambda value: True
+    )
+    # Use this to load our ensemble.  The callback ensures all our members
+    # have the "realization" coordinate and therefore they will all be loaded.
     surface_temp = iris.load_cube(
         iris.sample_data_path("GloSea4", "ensemble_???.pp"),
-        iris.Constraint("surface_temperature"),
+        constraint,
         callback=realization_metadata,
     )
 
@@ -57,17 +62,19 @@ def main():
     # Plot #1: Ensemble postage stamps
     # -------------------------------------------------------------------------
 
-    # for the purposes of this example, take the last time element of the cube
-    last_timestep = surface_temp[:, -1, :, :]
+    # For the purposes of this example, take the last time element of the cube.
+    # First get hold of the last time by slicing the coordinate.
+    last_time_coord = surface_temp.coord("time")[-1]
+    last_timestep = surface_temp.subset(last_time_coord)
 
     # Find the maximum and minimum across the dataset.
     data_min = np.min(last_timestep.data)
     data_max = np.max(last_timestep.data)
 
-    # Create a wider than normal figure to support our many plots
+    # Create a wider than normal figure to support our many plots.
     plt.figure(figsize=(12, 6), dpi=100)
 
-    # Also manually adjust the spacings which are used when creating subplots
+    # Also manually adjust the spacings which are used when creating subplots.
     plt.gcf().subplots_adjust(
         hspace=0.05,
         wspace=0.05,
@@ -77,13 +84,13 @@ def main():
         right=0.925,
     )
 
-    # iterate over all possible latitude longitude slices
+    # Iterate over all possible latitude longitude slices.
     for cube in last_timestep.slices(["latitude", "longitude"]):
 
-        # get the ensemble member number from the ensemble coordinate
+        # Get the ensemble member number from the ensemble coordinate.
         ens_member = cube.coord("realization").points[0]
 
-        # plot the data in a 4x4 grid, with each plot's position in the grid
+        # Plot the data in a 4x4 grid, with each plot's position in the grid
         # being determined by ensemble member number.  The special case for the
         # 13th ensemble member is to have the plot at the bottom right.
         if ens_member == 13:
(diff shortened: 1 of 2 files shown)


docs/iris/gallery_code/meteorology/plot_lagged_ensemble.py
```diff
@@ -51,7 +52,7 @@ def main():
     # coordinate, adding appropriate lagged ensemble metadata
     surface_temp = iris.load_cube(
         iris.sample_data_path("GloSea4", "ensemble_???.pp"),
-        iris.Constraint("surface_temperature", realization=lambda value: True),
```
rcomer, line 54: I don't actually understand the realization part of this constraint but removing it didn't make any difference to the plot.

docs/iris/gallery_code/meteorology/plot_lagged_ensemble.py
```diff
@@ -116,42 +115,29 @@ def main():
...
-    # Nino 3.4 lies between: 170W and 120W, 5N and 5S, so define a constraint
-    # which matches this
-    nino_3_4_constraint = iris.Constraint(
-        longitude=lambda v: -170 + 360 <= v <= -120 + 360,
-        latitude=lambda v: -5 <= v <= 5,
+    # Nino 3.4 lies between: 170W and 120W, 5N and 5S, so use the intersection
+    # method to restrict to this region.
+    nino_cube = surface_temp.intersection(
+        latitude=[-5, 5], longitude=[-170, -120]
     )
```
rcomer, line 131: ref #1027

docs/iris/gallery_code/meteorology/plot_lagged_ensemble.py
```diff
@@ -116,42 +115,29 @@ def main():
...
-
-    # Calculate the horizontal mean for the nino region
+    # Calculate the horizontal mean for the nino region.
     mean = nino_cube.collapsed(["latitude", "longitude"], iris.analysis.MEAN)
 
-    # Calculate the ensemble mean of the horizontal mean. To do this, remove
-    # the "forecast_period" and "forecast_reference_time" coordinates which
-    # span both "relalization" and "time".
-    mean.remove_coord("forecast_reference_time")
-    mean.remove_coord("forecast_period")
```
rcomer, line 150: ref #3028

docs/iris/gallery_code/meteorology/plot_lagged_ensemble.py
```diff
@@ -116,42 +115,29 @@ def main():
...
+    # method to restrict to this region.
+    nino_cube = surface_temp.intersection(
+        latitude=[-5, 5], longitude=[-170, -120]
     )
 
-    nino_cube = surface_temp.extract(nino_3_4_constraint)
-
-    # Subsetting a circular longitude coordinate always results in a circular
-    # coordinate, so set the coordinate to be non-circular
-    nino_cube.coord("longitude").circular = False
```
rcomer, line 141: Not really sure why this mattered...?

docs/iris/gallery_code/meteorology/plot_lagged_ensemble.py
```diff
@@ -62,10 +60,9 @@ def main():
     # for the purposes of this example, take the last time element of the cube
     last_timestep = surface_temp[:, -1, :, :]
```
rcomer, line 61: This might be better using `surface_temp.subset(surface_temp.coord(“time”)[-1])`, as it removes the requirement to know the dimension order.  Bit verbose though.

BryanLaura (approved):

docs/iris/gallery_code/meteorology/plot_lagged_ensemble.py
```diff
@@ -116,42 +112,29 @@ def main():
 
     # set a global title for the postage stamps with the date formated by
     # "monthname year"
-    plt.suptitle(
-        "Surface temperature ensemble forecasts for %s"
-        % (time.strftime("%B %Y"),)
-    )
+    title = "Surface temperature ensemble forecasts for {}"
+    plt.suptitle(title.format(time.strftime("%B %Y")))
```
bjlittle, line 116: @rcomer Even better, use an f-string here? i.e.,
```python
plt.suptitle(f"Surface temperature ensemble forecasts for {time.strftime('%B %Y')}")
```

rcomer: Ooooo I've never used an f-string before 😁

docs/iris/gallery_code/meteorology/plot_lagged_ensemble.py
bjlittle: Wow. Okay, what the heck... hmmm without digging this seems like an ancient workaround for something.

The only thing I can think of is that it's ensuring that the `realization` coordinate is always present? Dunno... anyways if removing it makes no difference, then let's remove it. Great spot :eyes:

docs/iris/gallery_code/meteorology/plot_lagged_ensemble.py
bjlittle: Personally, I've never used `subset`, remarkable, huh...

But the gallery is an ideal place to advertise functionality and patterns for `iris`... so I vote for adding it to the shop window. Go for it :+1:

docs/iris/gallery_code/meteorology/plot_lagged_ensemble.py
bjlittle: Hmm I strongly suspect it is to do with how the plot was rendered around/over the dateline,

rcomer: But then we're taking an area mean before we plot it... 🤷

docs/iris/gallery_code/meteorology/plot_lagged_ensemble.py
rcomer: OK, because I hate not knowing, I had a play:

* If I don't use the callback or the realization constraint, I get a `MergeError` because of mismatched "realization" coordinates, as we might expect.
* If I don't use the callback but reinstate the realization constraint, the plot works fine but more maps  and lines are missing.  So yeah, it seems this is a "coordinate exists" constraint, which actually seems quite useful, though I guess you could get the same effect using a `cube_func`.

docs/iris/gallery_code/meteorology/plot_lagged_ensemble.py
bjlittle: Just as `format` is a step-up from `%`, f-strings are a step-up from `format` IMHO

e.g., generally once you've tried them, you don't look back. I kinda love'em now.

docs/iris/gallery_code/meteorology/plot_lagged_ensemble.py
bjlittle: @rcomer Well at the very least it deserves a comment to explain what the heck it's doing. 

It's a neat pattern to employ without cracking open the `cube_func`. So let's honour the original authors intent but with some clarification for the reader... what say ye? :thinking:

rcomer: Definitely: if we're going to include it, it should be explained!  But... if we advertise this behaviour, should there also be a test for this behaviour?  _Is_ there a test for this behaviour?  If not, should I add one, and if so where?

And I thought a gallery update would be straightforward 🙄 🤣

bjlittle: No worries. Let's conquer and divide here :muscle: 

How's about you add a comment, if you're happy doing that and I'll add some appropriate test coverage, and target a pull-request on your `rcomer:gallery-seasonal-ensemble` branch.

Deal? 

(Ode to a simple life :pray:)

rcomer: Yeah sounds good to me.

rcomer: I figured it's easier to explain the constraint if it's defined on its own line.

docs/iris/gallery_code/meteorology/plot_lagged_ensemble.py
```diff
@@ -80,83 +84,68 @@ def main():
...
-    # set a global title for the postage stamps with the date formated by
-    # "monthname year"
-    plt.suptitle(
-        "Surface temperature ensemble forecasts for %s"
-        % (time.strftime("%B %Y"),)
-    )
+    # Set a global title for the postage stamps with the date formated by
+    # "monthname year".
+    time_string = time.strftime("%B %Y")
+    plt.suptitle(f"Surface temperature ensemble forecasts for {time_string}")
```
rcomer, line 122: I thought it might make it more obvious what's going on if the `time_string` is calculated before creating the f-string.

rcomer (commented):
Hi @bjlittle, I think I've covered those points now.  I also, somewhat belatedly, realised that the comments were a bit inconsistent.  So have been through and enforced capital letters and full stops.

bjlittle (approved):
@rcomer LGTM thanks :+1: 

Thanks for sticking with it :smile:

rcomer:
Thanks @bjlittle 🎉

PR #3934, opened 2020-12-07
Gallery: update COP maps example 

rcomer:
### Description
Some bits of this example weren't doing what its comments said it was doing.  Other bits just didn't make sense to me:

* The callback function was supposed to get the "experiment" from the directory name, but the files are all in the top level `sample_data` directory, so this added coordinate was the same regardless of file.  Part of the reason this hasn't been noticed before was probably because the "experiment" coordinate never got used after it was added.  I've updated the callback to get the experiment out of the filename.  Also loaded the two files together, then extracted the cubes in a loop.  This is a little contrived, but gives us an excuse to use the coordinate we went to the trouble of adding with the callback.  It's also the first time I've used `CubeList.extract_cube` (new at Iris v3!)

* Looping through slices of the cubes doesn't make sense to me as there is only one latitude-longitude slice.  I have instead looped through the experiments as this reduces repeated code.

* The `suptitle` comment claimed it was using the date in the format "MonthName Year", when it was actually just "Year" 🤷‍♀️

* For the colour array, I've used `np.stack` instead of `np.array` and `.T` because, well, I just like `np.stack`.

---

### DIFF
--- a/docs/gallery_code/meteorology/plot_COP_maps.py
+++ b/docs/gallery_code/meteorology/plot_COP_maps.py
@@ -38,7 +38,7 @@ def cop_metadata_callback(cube, field, filename):
     filename.
     """
 
-    # Extract the experiment name (such as a1b or e1) from the filename (in
+    # Extract the experiment name (such as A1B or E1) from the filename (in
     # this case it is just the start of the file name, before the first ".").
     fname = os.path.basename(filename)  # filename without path.
     experiment_label = fname.split(".")[0]
--- a/docs/gallery_code/meteorology/plot_COP_maps.py
+++ b/docs/gallery_code/meteorology/plot_COP_maps.py
@@ -39,30 +39,28 @@ def cop_metadata_callback(cube, field, filename):
     """
 
     # Extract the experiment name (such as a1b or e1) from the filename (in
-    # this case it is just the parent folder's name)
-    containing_folder = os.path.dirname(filename)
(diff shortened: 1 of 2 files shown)


rcomer:
Alternatively, the callback function could just be removed.  The plot will obviously work fine without it.  For a user who is actively looking for advice on writing callback functions, I'm not sure that the gallery would be the obvious place to look, as it's really about plotting.

rcomer:
Rebased 💃

jamesp:
Hi @rcomer, I'm going to take a look at this PR for you.  The docs have had a bit of a re-org over the last week, would you mind rebasing this PR on the latest main?  thanks!

rcomer:
Thanks @jamesp, I have now rebased and think everything is in order.

jamesp (approved):
Looks good to me @rcomer, thanks for the contribution! 👍

jamesp:
> Alternatively, the callback function could just be removed. The plot will obviously work fine without it. For a user who is actively looking for advice on writing callback functions, I'm not sure that the gallery would be the obvious place to look, as it's really about plotting.

As a new user getting to know iris, I found the callback use helpful.  Up until then I didn't realise it was possible so I wouldn't have even known what to search for!  It's a good addition to the examples.

rcomer:
Thanks @jamesp, great to have a fresh pair of eyes on this 👀

PR #3935, opened 2020-12-08
Scripted Logo Generation, take two

trexfeathers:
### Description

This PR proposes generating Iris' logos via a script, rather than holding fixed image files in the repository. I have updated the script since the original PR (#3764) to bring in line with @abooton's latest designs, as I believe there is an upcoming need for a finalised logo.

[Rendered docs here](https://trexfeathers-iris.readthedocs.io/en/program_logo/index.html)

For an idea of what the script can generate: [see the comment below](https://github.com/SciTools/iris/pull/3935#issuecomment-740276365) or run the script yourself 🙂

#### Advantages over existing method

1. No reliance on fixed files that are created using un-traceable methods (typically via a GUI graphics program)

2. Script is fully visible to version control

3. Image configuration easily adjustable (colours, transparency, gradients etc.)
   Iris' logo is made from relatively few elements with mostly simple configurations

4. All content is perfectly centred thanks to the programmatic generation
   I know this has been a problem with past logos

5. Title image includes parameterised text - particularly useful if we continue to modularise Iris and need a matching title logo for each package

6. SVG format renders crisply at any resolution
   Also doesn't blur in PDF's, unlike rasters (even high def ones)

7. SVG format means very small generated file sizes (<30KB)

8. Pythonic, in line with the rest of Iris' code

9. Don't need to hold multiple versions of the same image - all generated by a single script

#### Rotation
The script can also optionally generate rotating-earth versions of the logo - useful for publicity. This takes 1-2mins on most machines, and it is worth testing CPU demand before choosing to display a rotating image anywhere (GIF marginally better than SVG in this regard).

---


trexfeathers:
# Examples
[](https://raw.githubusercontent.com/trexfeathers/iris/program_logo_demo/docs/iris/src/_static/iris-logo.svg)
[](https://raw.githubusercontent.com/trexfeathers/iris/program_logo_demo/docs/iris/src/_static/iris-logo-title.svg)
[](https://raw.githubusercontent.com/trexfeathers/iris/program_logo_demo/docs/iris/src/_static/iris_grib-logo-title.svg)

pp-mo:
BTW the "hidden" errors in the log here are the same as I have on #3943 ...
There are "WARNING"s in the build : https://cirrus-ci.com/task/4660899346644992?command=tests#L795
... and it treats those as errors 
The actual problems are then visible in output further up, e.g. https://cirrus-ci.com/task/4660899346644992?command=tests#L168

HTH!

trexfeathers:
>BTW the "hidden" errors in the log here are the same as I have on #3943 ...

Yes! In 'take one' we hadn't yet elevated warnings, and I haven't accounted for the change yet. As it happens, I have an idea for an alternative implementation that might also fix this problem. Will change to draft in the meantime.

Thanks

docs/src/conf.py
```diff
@@ -216,11 +216,37 @@ def autolog(message):
 
 # -- Options for HTML output --------------------------------------------------
 
+# Generate logos.
+from importlib.util import spec_from_file_location, module_from_spec
+from pathlib import Path
+import requests
+from tempfile import TemporaryDirectory
+
+logo_gen_url = "https://raw.githubusercontent.com/trexfeathers/marketing/program_logo/iris/logo/generate_logo.py"
```
trexfeathers, line 225: Actual URL obviously pending the merge of SciTools/marketing#1.

trexfeathers: Fixed. Thanks for merging @bjlittle

tkknight:
Some initial thoughts after a quick play:

1. I was concerned that it may take a while to create the logo but in fact it < 2s, win!
1. Is there a built version on readthedocs yet?  If not please create one to see if it works ok.  I had issues with an svg logo appearing on the docs when viewed via a mobile device.  Ping me if you need some help on getting rtd setup.
2. Personally, if we have build output we need to be informative, particularly from where n print is coming from.  For example building the docs shows (snippets, not raw output):

```
[conf.py] Iris Version = 3.1.dev0
[generate_package_rst.py] Excluding module in package: 'experimental/raster'
```

I think the prefix of the python file would be useful when stating ``LOGO GENERATION START``. and ``LOGO GENERATION COMPLETE``. (and perhaps not in capitals)

3. A call to requests is used to get https://github.com/SciTools/marketing/blob/master/iris/logo/generate_logo.py.  This is probably fine but does add an internet dependency when building locally.  I believe github does have a symbolic link feature that allows you to point to another repo (maybe called submodule - not used it before so not sure)
4. The version text is small.  I realise this may depend on screen resolution but in comparison to the version in the sidebar in the pic below it is hard to read:
image
The font size in the svg could be increased and maybe the css for the logo could be overridden to make it bigger.  The css in this post may help: https://stackoverflow.com/a/67503668.  Perhaps a combination of a bigger font and a slightly bigger logo is the answer.

trexfeathers:
#4661 means we no longer need automatic logo regeneration with the relevant version string included. So in 68cb2f3f70a3f165077df48941a14f2c242b63bd I've now included the logos in the repo, together with guidance on regeneration/editing, and done away with the Sphinx automation.

tkknight (approved):
Nice job.

PR #3939, opened 2020-12-21
CI requirements drop pip packages

bjlittle:
### Description
This PR drops the use of `pip` packages from the `conda` CI requirements and uses the `conda` equivalent packages instead.

This PR also extends `nox` to offer `python` sessions for both `3.6` and `3.7`, and opting to use `pip install --editable .` for `iris` instead of `python setup.py develop`, but with `--no-deps` (as all the package dependencies are satisfied by `conda` and therefore not required to be checked by `pip`)

---

### DIFF
--- a/noxfile.py
+++ b/noxfile.py
@@ -164,7 +164,7 @@ def tests(session):
         cache_venv(session)
 
     cache_cartopy(session)
-    session.run("python", "setup.py", "develop")
+    session.install("--no-deps", "--editable", ".")
     session.run(
         "python",
         "-m",
@@ -207,7 +207,7 @@ def gallery(session):
         cache_venv(session)
 
     cache_cartopy(session)
-    session.run("python", "setup.py", "develop")
+    session.install("--no-deps", "--editable", ".")
     session.run(
         "python",
         "-m",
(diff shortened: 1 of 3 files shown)


noxfile.py
```diff
@@ -41,7 +41,7 @@ def venv_cached(session):
 
     """
     result = False
-    yml = Path(f"requirements/ci/py{PY_VER.replace('.', '')}.yml")
+    yml = Path(f"requirements/ci/py{session.python.replace('.', '')}.yml")
```
bjlittle, line 44: See [nox documentation](https://nox.thea.codes/en/stable/config.html#nox.sessions.Session.python) for further details.

tkknight (approved):
Code changes look ok.  

Would the testing instructions on the docs also need changing https://scitools-iris.readthedocs.io/en/latest/installing.html#running-the-tests and https://scitools-iris.readthedocs.io/en/latest/developers_guide/contributing_running_tests.html#developer-running-tests?

bjlittle:
@tkknight I'm going to target the docs update in a separate PR, if that's okay... Where I'll deal with the travis-ci to cirrus-ci, using Nox, when to use Nox and when not etc

In fact there are a few minor docs changes I want to make, and I'd rather have several small connected prs rather than one large pr.

tkknight:
Ok. Happy to merge as is then

PR #3941, opened 2020-12-28
Gallery: don't support mpl v1.2

rcomer:
### Description
A simple one:  I think the days of worrying about Matplotlib v1.2.x should be over.

Given how minor this change is, I'm unconvinced it warrants a whatsnew.

---

### DIFF
--- a/docs/gallery_code/meteorology/plot_deriving_phenomena.py
+++ b/docs/gallery_code/meteorology/plot_deriving_phenomena.py
@@ -26,14 +26,7 @@ def limit_colorbar_ticks(contour_object):
     number of ticks on the colorbar to 4.
 
     """
-    # Under Matplotlib v1.2.x the colorbar attribute of a contour object is
-    # a tuple containing the colorbar and an axes object, whereas under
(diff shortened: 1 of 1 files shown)


rcomer:
Link check failure seems to be because external docs have moved.  So I have opened #3942.

trexfeathers:
>Link check failure seems to be because external docs have moved. So I have opened #3942.

@rcomer since #3942 targeted `v3.0.x`, does this mean we can't currently merge anything to `master`?

rcomer:
> since #3942 targeted v3.0.x, does this mean we can't currently merge anything to master?

I think so.  I just assumed that `v3.0.x` was the priority at the moment and would be merged back into `master` before long.

rcomer:
Rebased 😇

rcomer:
Re-rebased 😎

jamesp (approved):
Seems good to me.  matplotlib 1.2.x is only supported up to Python 3.3, Iris now tested and supported only on >=3.6 so there's no overlap in supported versions.

rcomer:
Thanks @jamesp!

PR #3951, opened 2021-01-20
Remove warning for convert_units on lazy data

rcomer:
### Description
Removed the warning that `Cube.convert_units` will load lazy data.  Looks like this has not been true since #2964.

```python
In [1]: import iris

In [2]: import dask.array as da

In [3]: import numpy as np

In [4]: cube = iris.cube.Cube(da.from_array(np.arange(5)), units='m')

In [5]: cube.has_lazy_data()
Out[5]: True

In [6]: cube.convert_units('feet')

In [7]: cube.has_lazy_data()
Out[7]: True
```
Since the change is tiny, I'm targeting the v3 branch.  Happy to switch to `master` though.

---
[Consult Iris pull request check list]( https://scitools-iris.readthedocs.io/en/latest/developers_guide/contributing_pull_request_checklist.html)

PR #3947, opened 2021-01-12
Resolve test coverage

bjlittle:
### Description
This PR adds the missing unit test coverage for the `iris.common.resolve` module.

**To-do**

Unit test coverage is required for the following `iris.common.resolve.Resolve` methods:

- [x] `__init__`
- [x] `__call__`
- [x] `_metadata_resolve`
  - [x] `_categorise_items` 
- [x] `_metadata_coverage`
  - [x] `_dim_coverage`
  - [x] `_aux_coverage` 
- [x] `_metadata_mapping`
  - [x] `_dim_mapping`
  - [x] `_aux_mapping`
  - [x] `_free_mapping` 
  - [x] `_as_compatible_cubes`
- [x] `_metadata_prepare` 
  - [x] `_prepare_common_dim_payload`
  - [x] `_prepare_points_and_bounds`
  - [x] `_prepare_common_aux_payload` 
  - [x] `_prepare_local_payload` 
    - [x] `_create_prepared_item`
    - [x] `_prepare_local_payload_dim` 
    - [x] `_prepare_local_payload_aux` 
    - [x] `_prepare_local_payload_scalar`
  - [x] `_prepare_factory_payload` 
    - [x] `_get_prepared_item` 
- [x] `_src_cube`
- [x] `_src_cube_position`
- [x] `_src_cube_resolved`
- [x] `_tgt_cube`
- [x] `_tgt_cube_position`
- [x] `_tgt_cube_resolved`
- [x] `_tgt_cube_prepare`
- [x] `cube`
- [x] `mapped`
- [x] `shape`

---
[Consult Iris pull request check list]( https://scitools-iris.readthedocs.io/en/latest/developers_guide/contributing_pull_request_checklist.html)

PR #3952, opened 2021-01-21
Refactor cube summary

stephenworsley:
### Description

Refactors cube summary to make tabular representation more straightforward.

### DIFF
--- a/lib/iris/_representation.py
+++ b/lib/iris/_representation.py
@@ -1,0 +1,283 @@
+# Copyright Iris contributors
+#
+# This file is part of Iris and is released under the LGPL license.
+# See COPYING and COPYING.LESSER in the root of the repository for full
+# licensing details.
+"""
+Provides objects describing cube summaries.
+"""
+
+import iris.util
+
+
+def sorted_axes(axes):
+    """
+    Returns the axis names sorted alphabetically, with the exception that
+    't', 'z', 'y', and, 'x' are sorted to the end.
+    """
+    return sorted(
+        axes,
+        key=lambda name: ({"x": 4, "y": 3, "z": 2, "t": 1}.get(name, 0), name),
+    )
+
+
+class DimensionHeader:
+    def __init__(self, cube):
+        if cube.shape == ():
+            self.scalar = True
+            self.dim_names = []
+            self.shape = []
+            self.contents = ["scalar cube"]
+        else:
+            self.scalar = False
+            self.dim_names = []
+            for dim in range(len(cube.shape)):
+                dim_coords = cube.coords(
+                    contains_dimension=dim, dim_coords=True
+                )
+                if dim_coords:
+                    self.dim_names.append(dim_coords[0].name())
+                else:
+                    self.dim_names.append("-- ")
+            self.shape = list(cube.shape)
+            self.contents = [
+                name + ": %d" % dim_len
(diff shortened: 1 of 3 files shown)


trexfeathers:
This is a **big** thing to throw into `v3.0.x`, which I believe needs to be cut in the next couple of days. Are you sure this is the right place to point the PR?

pp-mo:
> big thing to throw into v3.0.x
It was my suggestion to target v3, thinking we might otherwise fall foul of forthcoming code changes there, as it's moved on so much.  Master will hopefully soon move forward to encompass v3.
Until then, this is only a draft, so not intending to merge it directly / immediately.

We can re-consider this, should not be a big problem as this is only one commit + introduces all-new content.

pp-mo:
The link check is general problem, not introduced here of course : @rcomer latest #3951 is also failing on this.

rcomer:
> link check is a general problem

I notice that @bjlittle is methodically turning them off (e.g. [here](https://github.com/SciTools/iris/pull/3947/files#diff-5cf76ad9090289a76548961e66bfc4b4fcc95ce6e1d686fb274357238f7f4d3a)).

pp-mo:
I just put up a PR with some ideas that I've been adopting to get the basic operations working.
That's just to ensure we have a common agenda really : I guess you're already working on some similar changes, which may not look the same (in which case, let's have yours + I'll review)

pp-mo:
Before making more concrete suggestions/requests, I want to wait until the code here settles a bit + you have some added tests.

In the meantime though, I do have some general points you might like to consider ...

(1) the instance properties (either controls or maybe just "constants") intended to control indenting : `item_indent`, `section_indent`, `extra_indent` ... 
 * are all a bit broken at this point, because they are not "passed down" when creating the contained Section-s (and Summary-s within them)
 * I think *as far as possible* we should be considering those things as 'styling' and so maybe should be leaving all that to the representation routines, not encoding them as properties of the basic summary object.  Tellingly (I think!), such choices may not be the same for different representations...
 * ... **However**, as discussed yesterday, I can see that there is an apparent need to control the 'clip length' in some areas :  I think that can best be managed as a simple 'max length' value, maybe doesn't even really *need* to be different between 'item' and 'extra'.  Actually, maybe ***such necessary clipping can also be delegated to the representation methods?***

(2) Where we have general context and control-info to "pass down" through Section and Summary creation, I think it would be easier + better to simply pass the parent reference to the sub-components.  
  * For example, instead of passing 'cube' in `VectorSection(title, cube, contents, iscoord)` and then `VectorSummary(cube, vector, iscoord)`, it would be simpler to pass the "summary" object itself as a constructor arg, and then use `self.summary.cube` in the lower-level code.  This then works for _any_ context or options settings which are set at the top level, e.g. could be used for the 'indent' controls at present (though I'm now arguing for removing those anyway!)

bjlittle:
@stephenworsley and @pp-mo I think this PR shouldn't be targeting `v3.0.x`, unless I misunderstand the intention here... but things have moved on rapidly since this PR was originally raised.

Is it okay to close this PR?

I think that we should be targeting `master` now, which will eventually end up in `v3.1.x`.

stephenworsley:
@bjlittle this can be closed since it has been superceded by #3987.

Issue #3961, opened 2021-01-26
Derived coordinates are not loaded anymore?


valeriupredoi:
## 📰 Custom Issue

Hey guys, one more for me today, I am using this [file](https://github.com/ESMValGroup/ESMValCore/blob/master/tests/integration/cmor/_fixes/test_data/common_cl_ap.nc) and loading it simple:
```python
import iris

cs = iris.load("/home/valeriu/ESMValCore/tests/integration/cmor/_fixes/test_data/common_cl_ap.nc")
for c in cs:
    print(c)
```
with iris=3 I am not seeing the derived coordinates no more:

- with `iris=2.4`:
```
b_bnds / (1)                        (atmosphere_hybrid_sigma_pressure_coordinate: 2; -- : 2)
     Dimension coordinates:
          atmosphere_hybrid_sigma_pressure_coordinate                           x       -
     Auxiliary coordinates:
          ap                                                                    x       -
          b                                                                     x       -
     Derived coordinates:
          air_pressure                                                          x       -
ap_bnds / (1)                       (atmosphere_hybrid_sigma_pressure_coordinate: 2; -- : 2)
     Dimension coordinates:
          atmosphere_hybrid_sigma_pressure_coordinate                           x       -
     Auxiliary coordinates:
          ap                                                                    x       -
          b                                                                     x       -
     Derived coordinates:
          air_pressure                                                          x       -
cloud_area_fraction_in_atmosphere_layer    (time: 1; atmosphere_hybrid_sigma_pressure_coordinate: 2; latitude: 3; longitude: 4)
     Dimension coordinates:
          time                                            x                                               -            -             -
          atmosphere_hybrid_sigma_pressure_coordinate     -                                               x            -             -
          latitude                                        -                                               -            x             -
          longitude                                       -                                               -            -             x
     Auxiliary coordinates:
          surface_air_pressure                            x                                               -            x             x
          ap                                              -                                               x            -             -
          b                                               -                                               x            -             -
     Derived coordinates:
          air_pressure                                    x                                               x            x             x
surface_air_pressure / (Pa)         (time: 1; latitude: 3; longitude: 4)
     Dimension coordinates:
          time                           x            -             -
          latitude                       -            x             -
          longitude                      -            -             x
     Attributes:
          additional_attribute: xyz
```
- with `iris=3.0`:
```
b_bnds / (unknown)                  (atmosphere_hybrid_sigma_pressure_coordinate: 2; -- : 2)
     Dimension coordinates:
[...]

PR #3795, opened 2020-08-20
Merge "default_units" feature branch

abooton:
Merge updated feature branch (see #3794) to master.
(Supersedes #3731)

These changes address the features in (parent issue) #3708

PR #3965, opened 2021-01-27
Promote unknown units to dimensionless in aux factories

bjlittle:
### Description
This PR adds graceful behaviour to the creation of auxiliary factories, such that formula terms that have `units` of `unknown` but **must** be dimensionless (`1`), have their `units` promoted from `unknown` to `1`.

This concession in behaviour is the result of real-world examples where NetCDF files missing an appropriate `units` attribute on formula terms variables, cause the failure of the expected auxiliary factory (and derived coordinate) to appear on the cube.

Previously, `iris` 2.4.0 loaded such files as expected, but as a result of #3795, `iris` 3.0.0 (advertised this as breaking behaviour) fails to behave as `iris` 2.4.0.

This PR alleviates this breaking change somewhat for the community, as it would cause major issues that are not easily avoidable. 

---

### DIFF
--- a/lib/iris/tests/unit/fileformats/netcdf/test__load_aux_factory.py
+++ b/lib/iris/tests/unit/fileformats/netcdf/test__load_aux_factory.py
@@ -53,8 +53,8 @@ class TestAtmosphereHybridSigmaPressureCoordinate(tests.IrisTest):
         self.assertEqual(factory.surface_air_pressure, self.ps)
 
     def test_formula_terms_a_p0(self):
-        coord_a = DimCoord(np.arange(5), units="Pa")
-        coord_p0 = DimCoord(10, units="1")
+        coord_a = DimCoord(np.arange(5), units="1")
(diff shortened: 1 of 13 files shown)


bjlittle:
Closed #3961

lib/iris/tests/unit/fileformats/netcdf/test__load_aux_factory.py
```diff
@@ -53,8 +53,44 @@ def test_formula_terms_ap(self):
         self.assertEqual(factory.surface_air_pressure, self.ps)
 
     def test_formula_terms_a_p0(self):
-        coord_a = DimCoord(np.arange(5), units="Pa")
```
bjlittle, line 56: Note that, the units were originally wrong here, so I swapped them around to be as expected.

trexfeathers:
LGTM

trexfeathers (approved):

PR #3968, opened 2021-01-28
Connectivity Class (AVD-1595)

trexfeathers:
### Description

Adds the `Connectivity` and accompanying `ConnectivityMetadata` classes - the first of many steps to allow Iris' core data model to represent the [CF-UGRID unstructured mesh format](https://ugrid-conventions.github.io/ugrid-conventions/) in addition to the existing CF structured grid format. Note this is targeting the new [`mesh-data-model`](https://github.com/SciTools/iris/tree/mesh-data-model) feature branch.

'Connectivity' name chosen since the most faithful interpretation of the [UGRID conventions](https://ugrid-conventions.github.io/ugrid-conventions/) is that a topology is made up of one or more self-contained connectivities.

Implemented as another member of [`iris.coords`](https://github.com/SciTools/iris/blob/master/lib/iris/coords.py), since `Connectivity`  benefits from inheriting the majority of [`_DimensionalMetadata`](https://github.com/SciTools/iris/blob/master/lib/iris/coords.py#L42-L680)'s functionality. `Connectivity` deviates slightly from existing `_DimensionalMetadata` children:

* Will never be attached to a traditional cube dimension, so [`cube_dims`](https://github.com/SciTools/iris/blob/master/lib/iris/coords.py#L186-L200) is just overridden with `NotImplementedError`.
* All the additional attributes all inter-relate - changing the value of just one will often render the others meaningless.
  As such none of them have a `.setter` method, and either need a special method to modify or the user is expected to create a new `Connectivity` with their new attribute values.
* The values of the `indices` array attribute are only meaningful subject to several requirements regarding `shape` and `dtype`, so `indices` is heavily validated whenever it is set.

---

### DIFF
--- a/lib/iris/common/metadata.py
+++ b/lib/iris/common/metadata.py
@@ -182,9 +182,10 @@ class BaseMetadata(metaclass=_NamedTupleMeta):
                     return result
 
                 # Note that, for strict we use "_fields" not "_members".
-                # The "circular" member does not participate in strict equivalence.
+                # The "circular" and "src_dim" members do not participate in strict equivalence.
                 fields = filter(
-                    lambda field: field != "circular", self._fields
+                    lambda field: field not in ("circular", "src_dim"),
+                    self._fields,
                 )
                 result = all([func(field) for field in fields])
 
@@ -882,6 +883,8 @@ class ConnectivityMetadata(BaseMetadata):
 
     """
 
+    # The "src_dim" member is stateful only, and does not participate in
+    # lenient/strict equivalence.
     _members = ("cf_role", "start_index", "src_dim")
 
     __slots__ = ()
@@ -933,11 +936,15 @@ class ConnectivityMetadata(BaseMetadata):
             Boolean.
 
         """
-        # Perform "strict" comparison for "cf_role", "start_index", "src_dim".
+        # Perform "strict" comparison for "cf_role", "start_index".
+        # The "src_dim" member is not part of lenient equivalence.
+        members = filter(
+            lambda member: member != "src_dim", ConnectivityMetadata._members
+        )
         result = all(
             [
                 getattr(self, field) == getattr(other, field)
-                for field in ConnectivityMetadata._members
+                for field in members
             ]
         )
         if result:
--- a/lib/iris/common/metadata.py
+++ b/lib/iris/common/metadata.py
@@ -34,10 +34,10 @@ __all__ = [
     "AncillaryVariableMetadata",
     "BaseMetadata",
     "CellMeasureMetadata",
+    "ConnectivityMetadata",
     "CoordMetadata",
     "CubeMetadata",
     "DimCoordMetadata",
-    "ConnectivityMetadata",
     "metadata_manager_factory",
 ]
 
@@ -876,6 +876,120 @@ class CellMeasureMetadata(BaseMetadata):
         return super().equal(other, lenient=lenient)
 
 
+class ConnectivityMetadata(BaseMetadata):
+    """
+    Metadata container for a :class:`~iris.coords.Connectivity`.
+
+    """
+
+    _members = ("cf_role", "start_index", "element_dim")
+
+    __slots__ = ()
+
+    @wraps(BaseMetadata.__eq__, assigned=("__doc__",), updated=())
+    @lenient_service
+    def __eq__(self, other):
+        return super().__eq__(other)
+
+    def _combine_lenient(self, other):
+        """
+        Perform lenient combination of metadata members for connectivities.
+
+        Args:
+
+        * other (ConnectivityMetadata):
+            The other connectivity metadata participating in the lenient
+            combination.
+
+        Returns:
+            A list of combined metadata member values.
+
+        """
+        # Perform "strict" combination for "cf_role", "start_index", "element_dim".
+        def func(field):
+            left = getattr(self, field)
+            right = getattr(other, field)
+            return left if left == right else None
+
+        # Note that, we use "_members" not "_fields".
+        values = [func(field) for field in ConnectivityMetadata._members]
+        # Perform lenient combination of the other parent members.
+        result = super()._combine_lenient(other)
+        result.extend(values)
+
+        return result
+
+    def _compare_lenient(self, other):
+        """
+        Perform lenient equality of metadata members for connectivities.
+
+        Args:
+
+        * other (ConnectivityMetadata):
+            The other connectivity metadata participating in the lenient
+            comparison.
+
+        Returns:
+            Boolean.
+
+        """
+        # Perform "strict" comparison for "cf_role", "start_index", "element_dim".
+        result = all(
+            [
+                getattr(self, field) == getattr(other, field)
+                for field in ConnectivityMetadata._members
+            ]
+        )
+        if result:
+            # Perform lenient comparison of the other parent members.
+            result = super()._compare_lenient(other)
+
+        return result
+
+    def _difference_lenient(self, other):
+        """
+        Perform lenient difference of metadata members for connectivities.
+
+        Args:
+
+        * other (ConnectivityMetadata):
+            The other connectivity metadata participating in the lenient
+            difference.
+
+        Returns:
+            A list of difference metadata member values.
+
+        """
+        # Perform "strict" difference for "cf_role", "start_index", "element_dim".
+        def func(field):
+            left = getattr(self, field)
+            right = getattr(other, field)
+            return None if left == right else (left, right)
+
+        # Note that, we use "_members" not "_fields".
+        values = [func(field) for field in ConnectivityMetadata._members]
+        # Perform lenient difference of the other parent members.
+        result = super()._difference_lenient(other)
+        result.extend(values)
+
+        return result
+
+    @wraps(BaseMetadata.combine, assigned=("__doc__",), updated=())
+    @lenient_service
+    def combine(self, other, lenient=None):
+        return super().combine(other, lenient=lenient)
+
+    @wraps(BaseMetadata.difference, assigned=("__doc__",), updated=())
+    @lenient_service
+    def difference(self, other, lenient=None):
+        return super().difference(other, lenient=lenient)
+
+    @wraps(BaseMetadata.equal, assigned=("__doc__",), updated=())
+    @lenient_service
+    def equal(self, other, lenient=None):
+        return super().equal(other, lenient=lenient)
+
+
 class CoordMetadata(BaseMetadata):
     """
     Metadata container for a :class:`~iris.coords.Coord`.
@@ -1317,120 +1431,6 @@ class DimCoordMetadata(CoordMetadata):
         return super().equal(other, lenient=lenient)
 
 
-class ConnectivityMetadata(BaseMetadata):
-    """
-    Metadata container for a :class:`~iris.coords.Connectivity`.
-
-    """
-
-    _members = ("cf_role", "start_index", "element_dim")
-
-    __slots__ = ()
-
-    @wraps(BaseMetadata.__eq__, assigned=("__doc__",), updated=())
-    @lenient_service
-    def __eq__(self, other):
-        return super().__eq__(other)
-
-    def _combine_lenient(self, other):
-        """
-        Perform lenient combination of metadata members for connectivities.
-
-        Args:
-
-        * other (ConnectivityMetadata):
-            The other connectivity metadata participating in the lenient
-            combination.
-
-        Returns:
-            A list of combined metadata member values.
-
-        """
-        # Perform "strict" combination for "cf_role", "start_index", "element_dim".
-        def func(field):
-            left = getattr(self, field)
-            right = getattr(other, field)
-            return left if left == right else None
-
-        # Note that, we use "_members" not "_fields".
-        values = [func(field) for field in ConnectivityMetadata._members]
-        # Perform lenient combination of the other parent members.
-        result = super()._combine_lenient(other)
-        result.extend(values)
-
-        return result
-
-    def _compare_lenient(self, other):
-        """
-        Perform lenient equality of metadata members for connectivities.
-
-        Args:
-
-        * other (ConnectivityMetadata):
-            The other connectivity metadata participating in the lenient
-            comparison.
-
-        Returns:
-            Boolean.
-
-        """
-        # Perform "strict" comparison for "cf_role", "start_index", "element_dim".
-        result = all(
-            [
-                getattr(self, field) == getattr(other, field)
-                for field in ConnectivityMetadata._members
-            ]
-        )
-        if result:
-            # Perform lenient comparison of the other parent members.
-            result = super()._compare_lenient(other)
-
-        return result
-
-    def _difference_lenient(self, other):
-        """
-        Perform lenient difference of metadata members for connectivities.
-
-        Args:
-
-        * other (ConnectivityMetadata):
-            The other connectivity metadata participating in the lenient
-            difference.
-
-        Returns:
-            A list of difference metadata member values.
-
-        """
-        # Perform "strict" difference for "cf_role", "start_index", "element_dim".
-        def func(field):
-            left = getattr(self, field)
-            right = getattr(other, field)
-            return None if left == right else (left, right)
-
-        # Note that, we use "_members" not "_fields".
-        values = [func(field) for field in ConnectivityMetadata._members]
-        # Perform lenient difference of the other parent members.
-        result = super()._difference_lenient(other)
-        result.extend(values)
-
-        return result
-
-    @wraps(BaseMetadata.combine, assigned=("__doc__",), updated=())
-    @lenient_service
-    def combine(self, other, lenient=None):
-        return super().combine(other, lenient=lenient)
-
-    @wraps(BaseMetadata.difference, assigned=("__doc__",), updated=())
-    @lenient_service
-    def difference(self, other, lenient=None):
-        return super().difference(other, lenient=lenient)
-
-    @wraps(BaseMetadata.equal, assigned=("__doc__",), updated=())
-    @lenient_service
-    def equal(self, other, lenient=None):
-        return super().equal(other, lenient=lenient)
-
-
 def metadata_manager_factory(cls, **kwargs):
     """
     A class instance factory function responsible for manufacturing
@@ -1574,10 +1574,10 @@ SERVICES_COMBINE = (
     AncillaryVariableMetadata.combine,
     BaseMetadata.combine,
     CellMeasureMetadata.combine,
+    ConnectivityMetadata.combine,
     CoordMetadata.combine,
     CubeMetadata.combine,
     DimCoordMetadata.combine,
-    ConnectivityMetadata.combine,
 )
 
 
@@ -1586,10 +1586,10 @@ SERVICES_DIFFERENCE = (
     AncillaryVariableMetadata.difference,
     BaseMetadata.difference,
     CellMeasureMetadata.difference,
+    ConnectivityMetadata.difference,
     CoordMetadata.difference,
     CubeMetadata.difference,
     DimCoordMetadata.difference,
-    ConnectivityMetadata.difference,
 )
 
 
@@ -1601,14 +1601,14 @@ SERVICES_EQUAL = (
     BaseMetadata.equal,
     CellMeasureMetadata.__eq__,
     CellMeasureMetadata.equal,
+    ConnectivityMetadata.__eq__,
+    ConnectivityMetadata.equal,
     CoordMetadata.__eq__,
     CoordMetadata.equal,
     CubeMetadata.__eq__,
     CubeMetadata.equal,
     DimCoordMetadata.__eq__,
     DimCoordMetadata.equal,
-    ConnectivityMetadata.__eq__,
-    ConnectivityMetadata.equal,
 )
 
 
--- a/lib/iris/common/metadata.py
+++ b/lib/iris/common/metadata.py
@@ -882,7 +882,7 @@ class ConnectivityMetadata(BaseMetadata):
 
     """
 
-    _members = ("cf_role", "start_index", "element_dim")
+    _members = ("cf_role", "start_index", "src_dim")
 
     __slots__ = ()
 
@@ -905,7 +905,7 @@ class ConnectivityMetadata(BaseMetadata):
             A list of combined metadata member values.
 
         """
-        # Perform "strict" combination for "cf_role", "start_index", "element_dim".
+        # Perform "strict" combination for "cf_role", "start_index", "src_dim".
         def func(field):
             left = getattr(self, field)
             right = getattr(other, field)
@@ -933,7 +933,7 @@ class ConnectivityMetadata(BaseMetadata):
             Boolean.
 
         """
-        # Perform "strict" comparison for "cf_role", "start_index", "element_dim".
+        # Perform "strict" comparison for "cf_role", "start_index", "src_dim".
         result = all(
             [
                 getattr(self, field) == getattr(other, field)
@@ -960,7 +960,7 @@ class ConnectivityMetadata(BaseMetadata):
             A list of difference metadata member values.
 
         """
-        # Perform "strict" difference for "cf_role", "start_index", "element_dim".
+        # Perform "strict" difference for "cf_role", "start_index", "src_dim".
         def func(field):
             left = getattr(self, field)
             right = getattr(other, field)
--- a/lib/iris/common/metadata.py
+++ b/lib/iris/common/metadata.py
@@ -37,6 +37,7 @@ __all__ = [
     "CoordMetadata",
     "CubeMetadata",
     "DimCoordMetadata",
+    "ConnectivityMetadata",
     "metadata_manager_factory",
 ]
 
@@ -1316,6 +1317,120 @@ class DimCoordMetadata(CoordMetadata):
         return super().equal(other, lenient=lenient)
 
 
+class ConnectivityMetadata(BaseMetadata):
+    """
+    Metadata container for a :class:`~iris.coords.Connectivity`.
+
+    """
+
+    _members = ("cf_role", "start_index", "element_dim")
+
+    __slots__ = ()
+
+    @wraps(BaseMetadata.__eq__, assigned=("__doc__",), updated=())
+    @lenient_service
+    def __eq__(self, other):
+        return super().__eq__(other)
+
+    def _combine_lenient(self, other):
+        """
+        Perform lenient combination of metadata members for connectivities.
+
+        Args:
+
+        * other (ConnectivityMetadata):
+            The other connectivity metadata participating in the lenient
+            combination.
+
+        Returns:
+            A list of combined metadata member values.
+
+        """
+        # Perform "strict" combination for "cf_role", "start_index", "element_dim".
+        def func(field):
+            left = getattr(self, field)
+            right = getattr(other, field)
+            return left if left == right else None
+
+        # Note that, we use "_members" not "_fields".
+        values = [func(field) for field in ConnectivityMetadata._members]
+        # Perform lenient combination of the other parent members.
+        result = super()._combine_lenient(other)
+        result.extend(values)
+
+        return result
+
+    def _compare_lenient(self, other):
+        """
+        Perform lenient equality of metadata members for connectivities.
+
+        Args:
+
+        * other (ConnectivityMetadata):
+            The other connectivity metadata participating in the lenient
+            comparison.
+
+        Returns:
+            Boolean.
+
+        """
+        # Perform "strict" comparison for "cf_role", "start_index", "element_dim".
+        result = all(
+            [
+                getattr(self, field) == getattr(other, field)
+                for field in ConnectivityMetadata._members
+            ]
+        )
+        if result:
+            # Perform lenient comparison of the other parent members.
+            result = super()._compare_lenient(other)
+
+        return result
+
+    def _difference_lenient(self, other):
+        """
+        Perform lenient difference of metadata members for connectivities.
+
+        Args:
+
+        * other (ConnectivityMetadata):
+            The other connectivity metadata participating in the lenient
+            difference.
+
+        Returns:
+            A list of difference metadata member values.
+
+        """
+        # Perform "strict" difference for "cf_role", "start_index", "element_dim".
+        def func(field):
+            left = getattr(self, field)
+            right = getattr(other, field)
+            return None if left == right else (left, right)
+
+        # Note that, we use "_members" not "_fields".
+        values = [func(field) for field in ConnectivityMetadata._members]
+        # Perform lenient difference of the other parent members.
+        result = super()._difference_lenient(other)
+        result.extend(values)
+
+        return result
+
+    @wraps(BaseMetadata.combine, assigned=("__doc__",), updated=())
+    @lenient_service
+    def combine(self, other, lenient=None):
+        return super().combine(other, lenient=lenient)
+
+    @wraps(BaseMetadata.difference, assigned=("__doc__",), updated=())
+    @lenient_service
+    def difference(self, other, lenient=None):
+        return super().difference(other, lenient=lenient)
+
+    @wraps(BaseMetadata.equal, assigned=("__doc__",), updated=())
+    @lenient_service
+    def equal(self, other, lenient=None):
+        return super().equal(other, lenient=lenient)
+
+
 def metadata_manager_factory(cls, **kwargs):
     """
     A class instance factory function responsible for manufacturing
@@ -1462,6 +1577,7 @@ SERVICES_COMBINE = (
     CoordMetadata.combine,
     CubeMetadata.combine,
     DimCoordMetadata.combine,
+    ConnectivityMetadata.combine,
 )
 
 
@@ -1473,6 +1589,7 @@ SERVICES_DIFFERENCE = (
     CoordMetadata.difference,
     CubeMetadata.difference,
     DimCoordMetadata.difference,
+    ConnectivityMetadata.difference,
 )
 
 
@@ -1490,6 +1607,8 @@ SERVICES_EQUAL = (
     CubeMetadata.equal,
     DimCoordMetadata.__eq__,
     DimCoordMetadata.equal,
+    ConnectivityMetadata.__eq__,
+    ConnectivityMetadata.equal,
 )
 
 
--- a/lib/iris/coords.py
+++ b/lib/iris/coords.py
@@ -31,9 +31,9 @@ from iris.common import (
     BaseMetadata,
     CFVariableMixin,
     CellMeasureMetadata,
+    ConnectivityMetadata,
     CoordMetadata,
     DimCoordMetadata,
-    ConnectivityMetadata,
     metadata_manager_factory,
 )
 import iris.exceptions
--- a/lib/iris/coords.py
+++ b/lib/iris/coords.py
@@ -2850,14 +2850,14 @@ class Connectivity(_DimensionalMetadata):
         * attributes (dict):
             A dictionary containing other cf and user-defined attributes.
         * start_index (int):
-            Either ``0`` or ``1``. Denotes whether :attr:`indices` uses 0-based
-            or 1-based indexing (allows support for Fortran and legacy NetCDF
-            files).
+            Either ``0`` or ``1``. Default is ``0``. Denotes whether
+            :attr:`indices` uses 0-based or 1-based indexing (allows support
+            for Fortran and legacy NetCDF files).
         * src_dim (int):
-            Either ``0`` or ``1``. Denotes which dimension of :attr:`indices`
-            varies over the :attr:`src_location`s (the alternate dimension
-            therefore varying within individual :attr:`src_location`s).
-            (This parameter allows support for fastest varying index being
+            Either ``0`` or ``1``. Default is ``0``. Denotes which dimension
+            of :attr:`indices` varies over the :attr:`src_location`s (the
+            alternate dimension therefore varying within individual
+            :attr:`src_location`s). (This parameter allows support for fastest varying index being
             either first or last).
             E.g. for ``face_node_connectivity``, for 10 faces:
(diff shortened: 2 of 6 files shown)


lib/iris/coords.py
```diff
@@ -2775,6 +2779,366 @@ def xml_element(self, doc):
...
+        * start_index (int):
+            Either ``0`` or ``1``. Denotes whether :attr:`indices` uses 0- or
+            1-based indexing (allows support for Fortran and legacy NetCDF
+            files).
+        * element_dim (int):
+            Either ``0`` or ``1``. Denotes which dimension of :attr:`indices`
+            refers to the contents of individual elements (allows support
+            for fastest varying index being either first or last). E.g. for
+            ``face_node_connectivity``, where the faces are 3-sided:
+            ``indices.shape[element_dim] = 3``.
```
trexfeathers, line 2837: In the [UGRID Conventions](https://ugrid-conventions.github.io/ugrid-conventions/), this isn't an expected attribute of a `_connectivity` variable (unlike `Connectivity`'s other parameters), but is important to include based on UGRID Conventions notes regarding `face_dimension` and `edge_dimension`.

lib/iris/common/metadata.py
```diff
@@ -37,6 +37,7 @@
     "CoordMetadata",
     "CubeMetadata",
     "DimCoordMetadata",
+    "ConnectivityMetadata",
```
bjlittle, line 40: Alphabetical? i.e., slot in after `CellMeasureMetadata`

lib/iris/common/metadata.py
```diff
@@ -1462,6 +1577,7 @@ def values(self):
     CoordMetadata.combine,
     CubeMetadata.combine,
     DimCoordMetadata.combine,
+    ConnectivityMetadata.combine,
```
bjlittle, line 1580: Same here... you know what I mean :stuck_out_tongue_winking_eye:

lib/iris/coords.py
```diff
@@ -2775,6 +2779,366 @@ def xml_element(self, doc):
...
+
+    def __init__(
+        self,
+        indices,
+        cf_role,
+        long_name=None,
+        var_name=None,
+        attributes=None,
+        start_index=0,
+        element_dim=1,
```
bjlittle, line 2800: Let me catchup on understanding `element_dim` :+1:

bjlittle: Agreed that we'll adopt the common metadata api `members` i.e., include `standard_name` and `units`.

lib/iris/coords.py
```diff
@@ -2775,6 +2779,366 @@ def xml_element(self, doc):
...
+        * long_name (str):
+            Descriptive name of the connectivity.
+        * var_name (str):
+            The netCDF variable name for the connectivity.
+        * attributes (dict):
+            A dictionary containing other cf and user-defined attributes.
+        * start_index (int):
+            Either ``0`` or ``1``. Denotes whether :attr:`indices` uses 0- or
+            1-based indexing (allows support for Fortran and legacy NetCDF
+            files).
```
bjlittle, line 2831: Mention the default value here and for `element_dim`

lib/iris/coords.py
```diff
@@ -32,6 +33,7 @@
     CellMeasureMetadata,
     CoordMetadata,
     DimCoordMetadata,
+    ConnectivityMetadata,
```
bjlittle, line 36: @trexfeathers Could we make this alphabetical, thanks :+1:

lib/iris/coords.py
```diff
@@ -2775,6 +2779,366 @@ def xml_element(self, doc):
         return cellMethod_xml_element
 
 
+class Connectivity(_DimensionalMetadata):
+    """
+    A CF-UGRID topology connectivity, describing the topological relationship
+    between two lists of dimensional elements. One or more connectivities
+    make up a CF-UGRID topology - a constituent of a CF-UGRID mesh.
```
bjlittle, line 2786: Not quite sure I agree on the description here...

The `Connectivity` is a definition (in terms of indices) of the interconnection between geometrical elements of a mesh, see [here](https://ugrid-conventions.github.io/ugrid-conventions/#topology).

Could we make this doc-string a wee bit more specific to the role and purpose of `Connectivity`?

lib/iris/coords.py
```diff
@@ -2775,6 +2779,366 @@ def xml_element(self, doc):
...
+            "face_node_connectivity",
+            "face_edge_connectivity",
+            "face_face_connectivity",
+            "edge_face_connectivity",
+            "boundary_node_connectivity",
+            "volume_node_connectivity",
+            "volume_edge_connectivity",
+            "volume_face_connectivity",
+            "volume_volume_connectivity",
+        ]
```
bjlittle, line 2858: This could be a class attribute, which within the class namespace would give it context e.g., `Connectivity.cf_roles` and would be accessible with or without a `Connectivity` instance.

I suggest `cf_roles` here, rather than `valid_cf_roles` i.e., no need to qualify it, as only those entries in the `Connectivity.cf_roles` are indeed valid.

Just a thought...

lib/iris/coords.py
```diff
@@ -2775,6 +2779,366 @@ def xml_element(self, doc):
...
+        self._metadata_manager.start_index = start_index
+        self._metadata_manager.element_dim = element_dim
+        self._metadata_manager.cf_role = cf_role
+
+        super().__init__(
+            values=indices,
+            standard_name=None,
+            long_name=long_name,
+            var_name=var_name,
+            units="1",
```
bjlittle, line 2870: `standard_name` and `units` default to `None` as kwargs in the `__init__`, so then just pass them through here and the inherited `CFVariableMixin` business logic will dutifully verify it for free.

lib/iris/coords.py
```diff
@@ -2775,6 +2779,366 @@ def xml_element(self, doc):
...
+            attributes=attributes,
+        )
+
+    @staticmethod
+    def _validate_arg_vs_list(arg_name, arg, list):
```
bjlittle, line 2875: :bomb: `list` is a `<class 'list'>` ... best avoided are a variable name

lib/iris/coords.py
```diff
@@ -2775,6 +2779,366 @@ def xml_element(self, doc):
...
+        if arg not in list:
+            error_msg = (
+                f"Invalid {arg_name} . Got: {arg} . Must be one of: "
+                f"{list} ."
```
bjlittle, line 2879: Curious, but why the spacing before the period (` .`)?

lib/iris/coords.py
```diff
@@ -2775,6 +2779,366 @@ def xml_element(self, doc):
...
+            standard_name=None,
+            long_name=long_name,
+            var_name=var_name,
+            units="1",
+            attributes=attributes,
+        )
+
+    @staticmethod
+    def _validate_arg_vs_list(arg_name, arg, list):
+        if arg not in list:
```
bjlittle, line 2876: I guess this is internal, so we don't need to check that `list` is actually an `Iterable` and not a `str`... your call

lib/iris/coords.py
trexfeathers: I originally did this, but it resulted in instances of `Connectivity` having both `my_conn.cf_role` and `my_conn.cf_roles`, which I considered confusing. What do you think?

lib/iris/coords.py
trexfeathers: I really struggled to come up with a decent description, any help is appreciated.

lib/iris/coords.py
trexfeathers: 😱

lib/iris/coords.py
trexfeathers: I found my original outputs difficult to read; having space either side of the value(s) in focus made them easier to pick out.

lib/iris/coords.py
trexfeathers: I admit I hadn't considered this, but yes the method is purely there due to its repeat use in validating `__init__` arguments and not intended for any 'non-planned' use.

lib/iris/coords.py
```diff
@@ -2775,6 +2779,366 @@ def xml_element(self, doc):
...
+    def cf_role(self):
+        """
+        The category of topological relationship that this connectivity
+        describes.
+        **Read-only** - validity of :attr:`indices` is dependent on
+        :attr:`cf_role`. A new :class:`Connectivity` must therefore be defined
+        if a different :attr:`cf_role` is needed.
+
+        """
+        return self._metadata_manager.cf_role
```
bjlittle, line 2912: :+1: for this being a `read-only` attribute

lib/iris/coords.py
```diff
@@ -2775,6 +2779,366 @@ def xml_element(self, doc):
...
+
+    @property
+    def cf_role_element(self):
```
bjlittle, line 2915: The term `element` is kinda generic and overloaded... could be borrow from UGRID terminology itself?

It's almost akin to the `location` of a UGRID NetCDF data variable, but not for the focus of where data payload is on the mesh, but rather the focus is for what fundamentally is this connectivity relationship about e.g., connecting `𝛼` to `𝛽`, or `face` to `node`, where the focus/`location` is `face`.

Would you object to this being called `location`?

So you could say:
```python
connectivity.location
```
which seems more meaningful than:
```python
connectivity.cf_role_element
```
which is harder (for me) to determine what it's telling me.

Also, does this need to be public? ... not that it greatly matters, as all `Connectivity` will be managed by the `Mesh`... I think I've answered my own question... public is fine.

lib/iris/coords.py
```diff
@@ -2775,6 +2779,366 @@ def xml_element(self, doc):
...
+        """
+        Derived from the connectivity's :attr:`cf_role` - the first part, e.g.
+        ``face`` in ``face_node_connectivity``. Refers to the elements
+        described by the structure of the connectivity's :attr:`indices` array.
+
+        """
+        return self._metadata_manager.cf_role.split("_")[0]
```
bjlittle, line 2922: If the `cf_role` metadata is `read-only`, then this only needs to be done once in the `__init__` and cached.

So this could be reduced to something like:
```python
return self._location
```
... if you buy into the previous comment on `cf_role_element`

lib/iris/coords.py
trexfeathers: Great idea since, as you say, we will be using `location` elsewhere to refer to these same things. "Located on this location" seems nice and self-reinforcing vs "Located on this element".

lib/iris/coords.py
trexfeathers: Clever. I also like the preservation of this as a property by using a private variable to store the info.

lib/iris/coords.py
bjlittle: @trexfeathers The definition of `element_dim` is the complete **opposite** to my expectation, given the meaning of UGRID `face_dimension` and `edge_dimension` e.g.,
```shell
dimensions:
nMesh2_node = 4 ; // nNodes
nMesh2_edge = 5 ; // nEdges
nMesh2_face = 2 ; // nFaces

Two = 2 ;
Three = 3 ;

variables:
// Mesh topology
integer Mesh2 ;
Mesh2:cf_role = "mesh_topology" ;
Mesh2:long_name = "Topology data of 2D unstructured mesh" ;
Mesh2:topology_dimension = 2 ;

Mesh2:face_node_connectivity = "Mesh2_face_nodes" ;
Mesh2:face_dimension = "nMesh2_face" ;

integer Mesh2_face_nodes(nMesh2_face, Three) ;
Mesh2_face_nodes:cf_role = "face_node_connectivity" ;
Mesh2_face_nodes:long_name = "Maps every triangular face to its three corner nodes." ;
Mesh2_face_nodes:start_index = 1 ;
```
So for the above example, `face_dimension` refers to the named dimension `nMesh2_face`, which is dimension `0` of the connectivity variable `Mesh2_face_nodes`. This is the **opposite** to this doc-string explanation. Here you're saying that `element_dim` would be `1`.

... and yet later for `Connectivity.cf_role_element` you use `element` to mean the `face_dimension` dimension, not the `element_dim` dimension... I'm finding the nomenclature not quite consistent, which is confusing me.

lib/iris/coords.py
bjlittle: In fact, the `cf_role_element` is fixed given the `cf_role`, so it's safe to calculate this and cache it once on the instance within the `__init__` rather than calculating it every call.

lib/iris/coords.py
```diff
@@ -2775,6 +2779,366 @@ def xml_element(self, doc):
...
+
+    @property
+    def cf_role_indexed_element(self):
+        """
+        Derived from the connectivity's :attr:`cf_role` - the second part, e.g.
+        ``node`` in ``face_node_connectivity``. Refers to the elements indexed
+        by the values in the connectivity's :attr:`indices` array.
+
+        """
+        return self._metadata_manager.cf_role.split("_")[1]
```
bjlittle, line 2932: Same again applies regarding naming, caching and whether this should be a public method.

lib/iris/coords.py
```diff
@@ -2775,6 +2779,366 @@ def xml_element(self, doc):
...
+
+    @property
+    def start_index(self):
+        """
+        The base value of the connectivity's :attr:`indices` array; either
+        ``0`` or ``1``.
+        **Read-only** - to change value: use :meth:`switch_start_index`.
+
+        """
+        return self._metadata_manager.start_index
```
bjlittle, line 2942: :+1: for this being a `read-only` member

lib/iris/coords.py
```diff
@@ -2775,6 +2779,366 @@ def xml_element(self, doc):
...
+
+    @property
+    def element_dim(self):
+        """
+        The dimension of the connectivity's :attr:`indices` array that details
+        individual elements.
+        **Read-only** - to change value: use :meth:`switch_element_dim`.
+
+        """
+        return self._metadata_manager.element_dim
```
bjlittle, line 2952: :+1: for this being a `read-only` member

lib/iris/coords.py
```diff
@@ -2775,6 +2779,366 @@ def xml_element(self, doc):
...
+                raise NotImplementedError
+            if (element_lengths < min_size).any().compute():
+                len_req_fail = f"len>={min_size}"
+        if len_req_fail:
+            indices_error(
+                f"Not all elements meet requirement: {len_req_fail} - needed "
+                f"to describe '{self.cf_role}' ."
+            )
+
+    def switch_start_index(self):
```
bjlittle, line 3042: Why do we need this?

So the `start_index` is actually not `read-only` then

lib/iris/coords.py
```diff
@@ -2775,6 +2779,366 @@ def xml_element(self, doc):
...
+        Change :attr:`start_index` to its alternative value and change the
+        :attr:`indices` values to match. (:attr:`start_index` can be either
+        ``0`` or ``1``).
+
+        """
+        new_index = 1 - self.start_index
+        index_change = new_index - self.start_index
+        self._metadata_manager.start_index = new_index
+        # Modify the indices values to correspond with the new start_index.
+        self._values += index_change
```
bjlittle, line 3055: This is a complete deal breaker for me.

The `Connectivity` doesn't live in isolation, it's used to index into the associated mesh coordinates. Changing the `start_index` and then using it will result in something very broken.

lib/iris/coords.py
```diff
@@ -2775,6 +2779,366 @@ def xml_element(self, doc):
...
+
+        """
+        self._metadata_manager.element_dim = 1 - self.element_dim
+        # Modify the indices structure to match the new element_dim.
+        # We know that indices is 2-dimensional so can just swap the two axes.
+        new_indices = self._values.swapaxes(0, 1)
+        # Replace the data manager with one for the new shaped indices.
+        # This is safe to override since Connectivity is not attached to a cube
+        # dimension unlike other _DimensionalMetadata-using classes.
+        self._values_dm = DataManager(new_indices)
```
bjlittle, line 3072: Again, I don't see why we need this behaviour.

Plus, supporting this is really saying that `element_dim` isn't `read-only`... this is a setter in disguise.

Don't have it part of the contract. The alternative is that the users just creates a different `Connectivity` instance instead.

lib/iris/coords.py
```diff
@@ -2775,6 +2779,366 @@ def xml_element(self, doc):
...
+        return super()._has_lazy_values()
+
+    def __str__(self):
+        result = repr(self)
+        return result
+
+    def __repr__(self):
+        cls = type(self).__name__
+        result = (
+            f"{cls}({self.indices}), cf_role="
```
bjlittle, line 3120: There is an extra `)` here that requires to be removed i.e.,
```python
f"{cls}({self.indices}, cf_role="
```

lib/iris/common/metadata.py
```diff
@@ -1316,6 +1317,120 @@ def equal(self, other, lenient=None):
...
+    """
+
+    _members = ("cf_role", "start_index", "element_dim")
+
+    __slots__ = ()
+
+    @wraps(BaseMetadata.__eq__, assigned=("__doc__",), updated=())
+    @lenient_service
+    def __eq__(self, other):
+        return super().__eq__(other)
```
bjlittle, line 1333: I don't question that `element_dim` is a member of the `ConnectivityMetadata`, but I do question whether it should participate in equality.

For this we need to consider this in the context of the `Connectivity` class, with two instances, each of which have the transpose of the others `indices`, and each have the flip of the others `element_dim` - and let's say for arguments sake this is for the `face_node_connectivity` for a regular quad mesh with 3 faces.

So we have

```python
a_indices = array([[ 0,  1,  2,  3],
                   [ 4,  5,  6,  7],
                   [ 8,  9, 10, 11]])
a_element_dim = 1
```
and
```python
b_indices = array([[ 0,  4,  8],
                   [ 1,  5,  9],
                   [ 2,  6, 10],
                   [ 3,  7, 11]])
b_element_dim = 0
```
You could argue, given how we understand connectivity, these are actually equivalent e.g.,
```
>>> a_indices[0, :]
array([0, 1, 2, 3])

>>> b_indices[:, 0]
array([0, 1, 2, 3])
```
The `indices` have a different `shape`, but given our interpretation of how `element_dim` is used, the results of indexing is the same... and yet `element_dim` is different and `indices` are different.

If you agree to this interpretation (up for debate), then we need to support a different implementation of `__eq__` for `ConnectivityMetadata` and `Connectivity` :thinking:

lib/iris/coords.py
bjlittle: The `__str__` and `__repr__` don't marry up with the metadata, which ultimately I think is confusing e.g.,
```python
>>> c = Connectivity(np.arange(12).reshape(3, 4), "face_node_connectivity", "long-name", "var-name", dict(one="world"), 0, 1)
>>> print(c)
Connectivity([[ 0  1  2  3]
 [ 4  5  6  7]
 [ 8  9 10 11]]), cf_role='face_node_connectivity', start_index=0, element_dim=1, long_name='long-name', var_name='var-name', attributes={'one': 'world'})
>>> print(c.metadata)
ConnectivityMetadata(standard_name=None, long_name='long-name', var_name='var-name', units=Unit('1'), attributes={'one': 'world'}, cf_role='face_node_connectivity', start_index=0, element_dim=1)
```

lib/iris/coords.py
```diff
@@ -2775,6 +2779,366 @@ def xml_element(self, doc):
...
+        """
+        A boolean stating whether all values from
+        :attr:`element_lengths` are equal. Provided
+        for lazy calculation without needing to realise :attr:`indices` or
+        :attr:`element_lengths`.
+
+        """
+        lengths = self._lazy_element_lengths(self.indices, self.element_dim)
+        has_equal = lengths.min() == lengths.max()
+        return has_equal.compute()
```
bjlittle, line 2991: Why do we need this behaviour? What's it's purpose?

This will be a massively expensive compute to perform.

lib/iris/coords.py
```diff
@@ -2775,6 +2779,366 @@ def xml_element(self, doc):
...
+
+    def _validate_indices(self, indices):
+        def indices_error(message):
+            raise ValueError("Invalid indeces provided. " + message)
```
bjlittle, line 2995: `indeces` -> `indices`

lib/iris/coords.py
```diff
@@ -2775,6 +2779,366 @@ def xml_element(self, doc):
...
+                f"dtype must be numpy integer subtype, got: {indices_dtype} ."
+            )
+
+        indices_min = indices.min()
+        if _lazy.is_lazy_data(indices_min):
+            indices_min = indices_min.compute()
+        if indices_min < self.start_index:
+            indices_error(
+                f"Lowest index: {indices_min} < start_index: {self.start_index} ."
+            )
```
bjlittle, line 3011: I understand this validation step, but there is a cost here.

Firstly, this validation step is locked in, and there is absolutely no way (at the moment) for it to be avoided on the creation of a `Connectivity` instance.

Secondly, consider a `C1668` cubed-sphere quad mesh. That has 1668*1668*6 = 16,693,344 faces, each face with 4 nodes... so that's 66,773,376 node indexes to search through for the `min`. In terms of memory, you'd need a 4-byte integer to represent the maximum index value, so that's loading and searching through ~254MB.

This won't come close to even causing an issue, but this is all avoidable, along with all the lovely validation that follows this in this method.

Validation is good, but make it optional. The users may totally trust their data, given its provenance, and so validation is not necessary. Other users may definitely want to exercise this validation, but I'd argue that should be an explicit choice.

bjlittle (changes requested):
@trexfeathers This is good, but there are quite a few things to mull over and address. So let's discuss them together.

I'd also question whether we want to inherit all the behaviour of the `_DimensionalMetadata` class i.e., do we need a slimmer base class of `_DimensionalMetadata` and inherit from that instead.

I'd also question inheriting the `_DimensionMetadata.__getitem__` behaviour, which copies the payload... perhaps we may want to consider avoiding such standard behaviour for all mesh like objects.

lib/iris/coords.py
bjlittle: Or, given a `cf_role = "face_node_connectivity"`, we could go with `connectivity.src` and `connectivity.tgt`, where:

- `connectivity.src` returns `"face"`
- `connectivity.tgt` returns `"node"`

This is building on the `"face"` being the source (`src`) of the connectivity, and `"node"` being the target (`tgt`) of the connectivity relationship.

There are few options here... but his might be even clearer.

lib/iris/coords.py
trexfeathers: This is probably related to my difficulty in talking about the structure of the `indices` array and what each dimension is for. I should have just taken my lead from the UGRID conventions rather than converting to my own [difficult to explain] understanding.

lib/iris/coords.py
trexfeathers: I really like this 👍

lib/iris/coords.py
trexfeathers: Re: 'read-only': the user cannot assign values to `start_index` in the usual way, and can only modify via a specific API. If not 'read-only' there must be some other term that makes this distinction.

Re: why: it was easy to implement and banks the logic rather than someone needing to think through it if they wanted to do this in future.

Anyway I think my explanations are probably irrelevant due to wider concerns.

lib/iris/coords.py
trexfeathers: Yup, that's not something I thought about at all 😳

lib/iris/coords.py
trexfeathers: Since we aware of a specific use-case for UGRID that is strictly uniformly shaped faces (the LFRic cubesphere), I figured it would be useful to distinguish such configurations from the standard [2D flexible mesh](https://ugrid-conventions.github.io/ugrid-conventions/#2d-flexible-mesh-mixed-triangles-quadrilaterals-etc-topology), and I wanted to avoid any chance of realising the size of every face so it seemed simple to write this convenience.

>This will be a massively expensive compute to perform.

I had imagined that lazy computation would take care of this problem, but given you raised a concern I wrote a test and it used `1GB` for a couple of seconds. With that in mind I agree this definitely isn't worth offering.

lib/iris/coords.py
trexfeathers: I had imagined that lazy computation would take care of this problem, but given you raised a concern I wrote a test and it did indeed use around that much memory each time 😔. Given this I question whether it's worth going into the depth of checking variable face sizes during validation or whether we just validate the raw `shape` of the array.

Having disproved my assumptions about computation I'd actually question whether [`_lazy_element_lengths`](https://github.com/SciTools/iris/pull/3968/files/ff08a8c0fcd14955be960c91341fb6e55b492ff5#diff-8fd786735e6d5fbb391db7380cb986b8836869167159414ca5a83141860b835dR2884-R2889) is worth the resource in any case - i.e. is it worth giving `Connectivity` the intelligence to inspect any masks in `indices` to calculate individual element lengths? I'd consider removing all uses of it.

lib/iris/coords.py
bjlittle: Or even, e.g.,

- `connectivity.src_location` returns `"face"`
- `connectivity.src_dim` returns, say `0`
- `connectivity.tgt_location` returns `"node"`
- `connectivity.tgt_dim` returns `1`

trexfeathers: In this situation we would only have `src_dim` as a parameter of `Connectivity`. The class itself can maintain an awareness/understanding of the alternate values of `src_dim` and `tgt_dim`, which could be returned / used internally whenever needed.

lib/iris/coords.py
trexfeathers: Easy to write or not, this method is risky to offer unless asked for as it has too much potential to create v confusing situations.

lib/iris/coords.py
trexfeathers: We have agreed that:

* This is effectively a `transpose` method, and should be named as such
* To preserve read-only, it should return a new `Connectivity`, rather than being in-place.
   [This is in line with Numpy](https://numpy.org/doc/stable/reference/generated/numpy.transpose.html), and is also the way that `Cube.transpose()` _should_ work.

lib/iris/common/metadata.py
trexfeathers: I would agree with this **if** the standard way of interacting with a `Connectivity` instance baked-in the intelligence around `src_dim` - i.e. a iterator that looks something like this:

```Python
for face_nodes in my_connectivity.src_iter():
   print(face_nodes)

> [0, 1, 2, 3]
> [4, 5, 6, 7]
```

Basically, we would need a recommended/safe/common way to iterate over two `Connectivity`s that are the transpose of each other. Without this, there would be dangers of it appearing safe to interact with / index two `Connectivity`s in the same way, even if that would have differing results.

lib/iris/coords.py
trexfeathers: See https://github.com/SciTools/iris/pull/3968#discussion_r570120871 for an agreed solution.

lib/iris/common/metadata.py
trexfeathers: >If you agree to this interpretation (up for debate), then we need to support a different implementation of `__eq__` for `ConnectivityMetadata` and `Connectivity`

To be clear: `ConnectivityMetadata` equality does not care about `src_dim`, while `Connectivity` equality accounts for transposition by intelligently checking `indices` and `src_dim`. (if a user/developer really cares about true equality they can check `indices` and `src_dim` manually).

lib/iris/coords.py
trexfeathers: @bjlittle please could you provide a suggestion for us to talk over? Thanks 🍺

lib/iris/coords.py
trexfeathers: Class attribute called `UGRID_CF_ROLES` should avoid confusion and be a better implementation

lib/iris/coords.py
trexfeathers: Also, define within `__init__`

lib/iris/coords.py
trexfeathers: See https://github.com/SciTools/iris/pull/3968#discussion_r570120871 for the agreed solution.

lib/iris/coords.py
trexfeathers: Validate method should have a `shapes_only (bool)` parameter. If `shapes_only == True`, just validate the shapes, otherwise do the thorough validation that supports masked index 'slots'.

lib/iris/coords.py
```diff
@@ -2775,6 +2779,366 @@ def xml_element(self, doc):
...
+        **Read-only** - index values are only meaningful when combined with
+        an appropriate :attr:`cf_role`, :attr:`start_index` and
+        :attr:`element_dim`. A new :class:`Connectivity` must therefore be
+        defined if different indices are needed.
+
+        """
+        return self._values
+
+    @property
+    def element_lengths(self):
```
trexfeathers, line 2970: Rename this

lib/iris/coords.py
trexfeathers: @bjlittle don't know if this info is useful to you but this problem started when I had the blinkers on and was looking at [`CellMeasure.__repr__()`](https://github.com/SciTools/iris/blob/3ff15a636f59d4cc3858199dfaaf83990e935154/lib/iris/coords.py#L861-L872), without noticing no other class overrides [`_DimensionalMetadata.__repr__()`](https://github.com/SciTools/iris/blob/3ff15a636f59d4cc3858199dfaaf83990e935154/lib/iris/coords.py#L324-L342); perhaps `CellMeasure` ought to also be fixed in future, since it creates a similar divergence from `CellMeasureMetadata.__repr__()`?

lib/iris/coords.py
trexfeathers: Will keep the private method, which will be used during `__init__` validation using `shapes_only=True`. Will also offer a public method, which calls the private method always using `shapes_only=False`, since we can guarantee that the `Connectivity` instance has already been through the `shapes_only=True` validation and that `indices` have not been modified. Will be sure to warn of possible resource demand in the docstring.

lib/iris/coords.py
trexfeathers: 6139e563acdd0805e72a4a3e2ffa4391c3e07337

lib/iris/coords.py
trexfeathers: 6139e563acdd0805e72a4a3e2ffa4391c3e07337

lib/iris/common/metadata.py
trexfeathers: c2ac0fc05949f07cd3483e802994708aa427da22

lib/iris/common/metadata.py
trexfeathers: c2ac0fc05949f07cd3483e802994708aa427da22

lib/iris/coords.py
trexfeathers: 85b98c992188ee3e41d6b2a98139024090ad2231

lib/iris/coords.py
trexfeathers: 52aa5c2dc32d5e8579163a2e8a9774195d462940
16c5e759bd4d1b7f2668507a88ce0472a8063786

lib/iris/coords.py
trexfeathers: c2ac0fc05949f07cd3483e802994708aa427da22

lib/iris/coords.py
trexfeathers: b9f470a79683d5d0f204ad7183f6749206acfc06

lib/iris/coords.py
trexfeathers: 85b98c992188ee3e41d6b2a98139024090ad2231

lib/iris/coords.py
trexfeathers: d512fa2c9dd8a378954592be090b2c7bcff1170a

lib/iris/coords.py
trexfeathers: d512fa2c9dd8a378954592be090b2c7bcff1170a

lib/iris/coords.py
trexfeathers: 6139e563acdd0805e72a4a3e2ffa4391c3e07337

lib/iris/coords.py
trexfeathers: 6139e563acdd0805e72a4a3e2ffa4391c3e07337

lib/iris/coords.py
trexfeathers: 97310e758a5b23a92c28910d885bfaa2aa2c79dd

lib/iris/coords.py
trexfeathers: 97310e758a5b23a92c28910d885bfaa2aa2c79dd

lib/iris/coords.py
trexfeathers: 97310e758a5b23a92c28910d885bfaa2aa2c79dd

lib/iris/coords.py
trexfeathers: 2808afe2ea513b3549cdf4173bb87a6f74fd5c4b

lib/iris/common/metadata.py
trexfeathers: 116a683cd9eb310b4ea00cded8397c3e26d60d1c

Decided not to implement `__iter__`. Having thought at length, it seems too early to be make assumptions / be opinionated about how developers and advanced users will want to interact with this class. Instead I've explicitly created the [`indices_by_src()`](https://github.com/SciTools/iris/pull/3968/commits/116a683cd9eb310b4ea00cded8397c3e26d60d1c#diff-8fd786735e6d5fbb391db7380cb986b8836869167159414ca5a83141860b835dR3003-R3037) method, to make it maximally clear what's going on. If someone wants to iterate by `src_location` they can always iterate through the `indices_by_src()` output.

lib/iris/coords.py
trexfeathers: c04f2308cfab2a96ebb146a3034e468d397af06c

lib/iris/coords.py
trexfeathers: 8d14157d99e7921caa80635cf9c1f89c6a8ef599

lib/iris/coords.py
trexfeathers: 7354e89a998bd64bf8fff3166131628db385b2a4

lib/iris/coords.py
trexfeathers: 6139e563acdd0805e72a4a3e2ffa4391c3e07337

trexfeathers:
Back to you @bjlittle 🙂 Thanks for all your helpful comments, I'm pleased with how things have come out.

I've addressed everything except https://github.com/SciTools/iris/pull/3968#discussion_r568624569, since we were planning to talk about this. I also haven't taken any actions on your general notes but there seems to be a general awareness of those anyway.

lib/iris/common/metadata.py
trexfeathers: @bjlittle what makes our discussion here regarding `src_dim` different from `start_index`?

As far as I can see, a `Connectivity` with 'shifted' `indices` and the alternate `start_index` is equivalent to a `Connectivity` without these, in exactly the same way as we've discussed a `Connectivity` with transposed `indices` and the alternate `src_dim`.

And it would be safe to treat these as equivalent if there was an agreed 'safe', 'neutralised' object that represents `indices` after correcting `start_index` differences, in the same way we discussed above for `src_dim` differences.

What have I missed? Is the problem that one couldn't offer up a **view** with shifted indices in the same way that one could with a transposed view, thus making such an operation more expensive?

lib/iris/coords.py
```diff
@@ -2775,6 +2778,469 @@ def xml_element(self, doc):
...
+            the indices of the 'target locations' it relates to.
+            Use a :class:`numpy.ma.core.MaskedArray` if :attr:`src_location`
+            lengths vary - mask unused index 'slots' within each
+            :attr:`src_location`. Use a :class:`dask.array.Array` to keep
+            indices 'lazy'.
+        * cf_role (str):
+            Denotes the topological relationship that this connectivity
+            describes. Made up of this array's locations, and the indexed
+            'target location' within each location.
+            Examples: ``face_node_connectivity``, ``edge_face_connectivity``.
```
bjlittle, line 2833: @trexfeathers You could change this line to simply say, 
```
See `Connectivity.UGRID_CF_ROLES`.
```

lib/iris/coords.py
```diff
@@ -2775,6 +2778,469 @@ def xml_element(self, doc):
...
+            # corrects for this difference. (To enable this, src_dim does
+            # not participate in ConnectivityMetadata to ConnectivityMetadata
+            # equivalence).
+            if hasattr(other, "metadata"):
+                # metadata comparison
+                eq = self.metadata == other.metadata
+                eq = (
+                    eq
+                    and (self.indices_by_src() == other.indices_by_src()).all()
+                )
```
bjlittle, line 3130: @trexfeathers This could be simplified to the following pattern:
```python
if eq:
    eq = (self.indices_by_src() == other.indices_by_src()).all()
```

trexfeathers: 6c9c76510b80fe3f9fae7a81ac1a449b0a577855

lib/iris/coords.py
trexfeathers: 6c9c76510b80fe3f9fae7a81ac1a449b0a577855

bjlittle (approved):

bjlittle:
@trexfeathers Awesome work, thanks for sticking with it :+1: :100: :tada: 

Let's see the utility of `Connectivity.indices_by_src` vs `Connectivity.__iter__` when we come to use it in anger... let's not second guess that right now. It will answer itself eventually (which is very soon :wink:)

:beers:

trexfeathers:
Thanks @bjlittle, much appreciated!

>Let's see the utility of `Connectivity.indices_by_src` vs `Connectivity.__iter__` when we come to use it in anger... let's not second guess that right now.

I suppose that was my point - before we know the answer I'd rather make it **ex**plicit rather than implicit.

Issue #3967, opened 2021-01-28
iris.coords.Coord() gives me a TypeError


valeriupredoi:
## 📰 Custom Issue

Hey guys, me again :grin: Silly question - I have this snippet:
```python
import iris
print(iris.__version__)
import numpy as np

lat_coord = iris.coords.Coord(
    np.linspace(-90.0, 90.0, 100), var_name='lat',
    standard_name='latitude', long_name='latitude', units='degrees')
```
and in `iris=3.0.1` gives me this malarkey:
```
/home/valeriu/miniconda3/envs/iris31/lib/python3.9/site-packages/iris/config.py:139: UserWarning: Ignoring config item 'Resources':'test_data_dir' (section:option) as '/home/h05/itwl/projects/git/iris-test-data/test_data' is not a valid directory path.
  warnings.warn(msg.format(section, option, c_path))
3.0.1
Traceback (most recent call last):
  File "/home/valeriu/ESMValTool/testcoord.py", line 5, in <module>
    lat_coord = iris.coords.Coord(
TypeError: Can't instantiate abstract class Coord with abstract method __init__
```
and in `iris=2.4.0` it runs a-ok, with no warnings. First off, I don't understand the warning, it seems to be pointing to a location that's got nothing to do with me :grin: Second, why is it not constructing my coordinate? Looking at the [latest documentation](https://scitools-iris.readthedocs.io/en/latest/generated/api/iris/coords.html?highlight=coord#iris.coords.Coord) it seems support should be as previous version? Cheers muchly in advance, as usual, for great support :beer:

bjlittle:
@valeriupredoi This is now the correct behavour. 

The `iris.coords.Coord` is a base class and has now been made **abstract** i.e., you cannot create an instance of it, because you shouldn't. 

Prior to `iris` 3.0.0 that wasn't the case. Essentially up to `iris` 2.4.0 we were allowing you to do the wrong thing. We've disallowed that not, for very good reasons.

You can only create a `iris.coords.DimCoord` or a `iris.coords.AuxCoord`. That's is.

bjlittle:
@valeriupredoi Does this help?

So either do:
```python
lat_coord = iris.coords.DimCoord(
    np.linspace(-90.0, 90.0, 100), var_name='lat',
    standard_name='latitude', long_name='latitude', units='degrees')
```
or:
```python
lat_coord = iris.coords.AuxCoord(
    np.linspace(-90.0, 90.0, 100), var_name='lat',
[...]

PR #3971, opened 2021-01-31
document that iris.coords.Coord is an ABC

bjlittle:
### Description
This PR clarifies in the documentation that `iris.coords.Coord` is an abstract base class (ABC) as of `iris` 3.0.0, and therefore users cannot create an instance of it.

It also organises the end section of the `whatsnew` into two categories, one for entry author names, and the other for entry resources, using the `rst` `only` directive, see [here](https://sublime-and-sphinx-guide.readthedocs.io/en/latest/conditions.html).

---


bjlittle:
Closes #3967

tkknight (approved):

PR #3976, opened 2021-02-02
Iris py38

bjlittle:
### Description
This PR extends the `iris` testing to cover `python` 3.8.x.

In order to provision this testing infra-structure, work was required on `cml` generation, as [xml.dom.minidom](https://docs.python.org/3/library/xml.dom.minidom.html#module-xml.dom.minidom) has changed from 3.8 to preserve xml element attribute assignment order - most likely due to `python` dictionaries now preserving order.

In addition to this a minor fix was required for the doc-tests, as a `namedtuple._todict()` now returns a `dict` rather than an `OrderedDict`.

Note that, all the CI documentation is performed soley for `python` 3.8 i.e.,
* the documentation gallary
* the doc-tests
* the documentation link-check

---

### DIFF
--- a/lib/iris/coords.py
+++ b/lib/iris/coords.py
@@ -2275,9 +2275,7 @@ class Coord(_DimensionalMetadata):
         # class name
         element = super().xml_element(doc=doc)
 
-        element.setAttribute("points", self._xml_array_repr(self.points))
-
-        # Add bounds handling
+        # Add bounds, points are handled by the parent class.
         if self.has_bounds():
             element.setAttribute("bounds", self._xml_array_repr(self.bounds))
 
--- a/lib/iris/coords.py
+++ b/lib/iris/coords.py
@@ -578,27 +578,25 @@ class _DimensionalMetadata(CFVariableMixin, metaclass=ABCMeta):
         return self._values_dm.shape
 
     def xml_element(self, doc):
-        """Return a DOM element describing this metadata."""
-        # Create the XML element as the camelCaseEquivalent of the
-        # class name.
+        """Return a DOM element describing this metadata and a dictionary of its attributes."""
+        # Create the XML element as the camelCaseEquivalent of the class name.
         element_name = type(self).__name__
         element_name = element_name[0].lower() + element_name[1:]
         element = doc.createElement(element_name)
+        attributes = {}
 
-        element.setAttribute("id", self._xml_id())
+        attributes["id"] = self._xml_id()
 
         if self.standard_name:
-            element.setAttribute("standard_name", str(self.standard_name))
+            attributes["standard_name"] = str(self.standard_name)
         if self.long_name:
-            element.setAttribute("long_name", str(self.long_name))
+            attributes["long_name"] = str(self.long_name)
         if self.var_name:
-            element.setAttribute("var_name", str(self.var_name))
-        element.setAttribute("units", repr(self.units))
+            attributes["var_name"] = str(self.var_name)
+        attributes["units"] = repr(self.units)
         if isinstance(self, Coord):
             if self.climatological:
-                element.setAttribute(
-                    "climatological", str(self.climatological)
-                )
+                attributes["climatological"] = str(self.climatological)
 
         if self.attributes:
             attributes_element = doc.createElement("attributes")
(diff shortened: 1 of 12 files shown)


lib/iris/coords.py
```diff
@@ -777,6 +775,24 @@ def cube_dims(self, cube):
         """
         return cube.ancillary_variable_dims(self)
 
+    def xml_element(self, doc, split=None):
```
jamesp, line 778: is there a reason this method returns a different signature (element, attributes) vs. (element,) depending on the param value `split`?  different use cases?

jamesp: The default signature `xml_element(doc) => (element,)` is different to the default signature of e.g. `Coord#xml_element => (element, attributes)`.  The api seems inconsistent throughout this file, is this expected?

jamesp (commented):
Started a review, this is a great way to get stuck into understanding the codebase!  Will give it some more time later, a couple of comments/questions inline below.

lib/iris/coords.py
bjlittle: @jamesp Yes, this is intentional, and I couldn't see a neater way around it.

The issue is being forced by the inheritance hierarchy and the fact that the `xml_element` method in the **final** inheritance call chain **must** return an `xml.dom.minidom.Element`.

Just be to explicit we have the following:
```
CellMethod
`-- xml_element()

_DimensionalMetadata
|-- AncillaryVariable
|   |-- CellMeasure
|   |   `-- xml_element()
|   `-- xml_element()
|-- Coord
|   |-- AuxCoord
|   |   `-- xml_element()
|   |-- DimCoord
|   |   `-- xml_element()
|   `-- xml_element()
`-- xml_element()
```

So `AncillaryVariable.xml_element` by default requires to return an `xml.dom.minidom.Element` if it is **directly called**, otherwise, if `AncillaryVariable.xml_element` is called **indirecty** as part of the `super` in `CellMeasure.xml_element`, `CellMeasure` requires `AncillaryVariable.xml_element` to return `(xml.dom.minidom.Element, dict)` instead as it still has has to mix-in the `measure` attribute before finally calling `xml.dom.minidom.Element.setAttribute`.

Why are we doing this now?

Well, basically, for Python `<3.8` the `__str__`/`__repr__` of a `xml.dom.minidom.Element` would show its elements "for free" in alphabetical order... for Python `>=3.8` it now shows its elements in assigned order - which breaks all our `cml` tests i.e., I'm guessing this is a direct side-effect of `dict`s behaving more like `OrderedDict`s.

So the deal is that we need **all** the attributes (`key`-`values` pairs) that we're ever going to set on the `xml.dom.minidom.Element` **all** at once before doing so, but sort them by `key` first, to get the same behaviour pre-3.8.

The fact that `CellMeasure` is a sub-class of `AncillaryVariable` is forcing this dual return type behaviour from `AncillaryVariable`.

Does this help?

bjlittle: BTW I'm not wedded to the `split` kwarg as a name. If you agree on my strategy of doing all this, and the purpose of the `split` kwarg... then please suggest a rename for it, if you can think of something better.

bjlittle: I guess I could be a good citizen here and add all the missing doc-strings for the `xml_element` methods documenting the args, kwargs and return types...

bjlittle: In fact, I'm just going to go ahead an just do that... why not :wink:

tkknight:
Overall looks ok to me @bjlittle .  I would only add to ensure the docs are updated to include that **Python 3.8** is now supported and tested.  See the existing not on the installing Iris page: https://scitools-iris.readthedocs.io/en/latest/installing.html

Over to you @jamesp to finish your review with @bjlittle. 😃

lib/iris/coords.py
bjlittle: Done

bjlittle:
@jamesp I think this is good to go 😀

jamesp:
@bjlittle ok! 👍

lib/iris/coords.py
```diff
@@ -2228,14 +2256,26 @@ def nearest_neighbour_index(self, point):
...
+            The :class:`xml.dom.minidom.Element` that will describe this
+            :class:`DimCoord`, and the dictionary of attributes that require
+            to be added to this element.
+
+        """
         # Create the XML element as the camelCaseEquivalent of the
         # class name
         element = super().xml_element(doc=doc)
 
-        element.setAttribute("points", self._xml_array_repr(self.points))
```
jamesp, line 2236: @bjlittle can you just confirm this line supposed to be deleted? comment below suggests that it is

jamesp (approved):

lib/iris/coords.py
bjlittle: @jamesp Yup, that's fine, and it's a good question...

It's actually redundant, and I only noticed it by working through the inheritance behaviour. 

`Coord.xml_element` calls `_DimensionalMetadata.xml_element`, and in the super class it performs the following:
```python
        # The values are referred to "points" of a coordinate and "data"
        # otherwise.
        if isinstance(self, Coord):
            values_term = "points"
        else:
            values_term = "data"
        element.setAttribute(values_term, self._xml_array_repr(self._values))
```
So as it stood, the `_DimensionalMetadata` was setting the `points` attribute, then the `Coord` came along an stomped over it with the same result... seemed like worthwhile trimming that fat whilst we're here.

If it didn't work... we'd have test failures all over the place. So I believe this is fine.

Lovely spot though :eagle: :eyes:

bjlittle:
@jamesp I've got come conflicts that I need to resolve...

PR #3981, opened 2021-02-04
automate docs discovery of iris and python versions

bjlittle:
### Description
This PR automates the discovery of the `iris` version for the `whatsnew`, and also the `python` version used to build the documentation via the `sphinx` [rst_epilog](https://www.sphinx-doc.org/en/master/usage/configuration.html#confval-rst_epilog).

These are available in all `rst` files via `|iris_version|` and `|python_version|` directives.

The rendered version of the `whatsnew` generated by this PR is available [here](https://bjlittle-iris.readthedocs.io/en/latest/whatsnew/latest.html).

The rendered version of the `release` documentation changes from this PR are available [here](https://bjlittle-iris.readthedocs.io/en/latest/developers_guide/release.html#release-steps).

P.S. the `|build_date|` directive is now also available to automate the `whatsnew` release/build date.

---

### DIFF
--- a/docs/src/conf.py
+++ b/docs/src/conf.py
@@ -96,7 +96,7 @@ autolog("Iris Release = {}".format(release))
 # Create a variable that can be inserted in the rst "|copyright_years|".
 # You can add more variables here if needed.
 rst_epilog = f"""
-.. |copyright_years| replace:: {2010 - upper_copy_year}
+.. |copyright_years| replace:: 2010 - {upper_copy_year}
 .. |python_version| replace:: {'.'.join([str(i) for i in sys.version_info[:3]])}
 .. |iris_version| replace:: v{version}
 .. |build_date| replace:: ({datetime.datetime.now().strftime('%d %b %Y')})
--- a/docs/src/conf.py
+++ b/docs/src/conf.py
@@ -99,10 +99,12 @@ rst_epilog = """
 .. |copyright_years| replace:: {year_range}
 .. |python_version| replace:: {python_version}
 .. |iris_version| replace:: {iris_version}
+.. |build_date| replace:: {build_date}
 """.format(
     year_range=f"2010 - {upper_copy_year}",
     python_version=".".join([str(i) for i in sys.version_info[:3]]),
     iris_version=f"v{version}",
+    build_date=f"({datetime.datetime.now().strftime('%d %b %Y')})",
 )
 
 # Add any Sphinx extension module names here, as strings. They can be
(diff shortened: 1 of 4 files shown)


trexfeathers:
Did you intend some use of `|python_version|` or is this just flexibility for the future?

docs/src/conf.py
```diff
@@ -92,12 +93,16 @@ def autolog(message):
 
 # -- General configuration ---------------------------------------------------
 
-# Create a variable that can be insterted in the rst "|copyright_years|".
-# You can add more vairables here if needed
+# Create a variable that can be inserted in the rst "|copyright_years|".
+# You can add more variables here if needed.
 rst_epilog = """
```
jamesp, line 98: Is there a reason for using format strings in the `.format` method, rather than having a single format string 
```
f"""
|copyright_years| replace:: 2010 - {upper_copy_year}
[...etc]
"""
```

jamesp (commented):
Looks ok to me, question inline

bjlittle:
@trexfeathers The `|python_version|` is a hook for @tkknight later on...

As part of #3976, I was forced to make the decision to only build the documentation for the latest supported version of python for iris i.e., 3.8, simply due to changes in some doc-tests output which differs based on the python version. 

This poses a problem, as we don't want to be in the situation of having different doc-tests based on the underlying python version. To keep this simple we should only build and publish the latest supported version of python for Iris. In doing so, we should make it clear to users what version of python the documentation is built against, hence the need for `|python_version|`.

@tkknight will push a follow-on PR to propose how we highlight this information.

docs/src/conf.py
bjlittle: Yup, could do, thanks.

I was lazily following what was there before... I'll tidy it up :+1:

bjlittle: Done. Nice, that's way tidier, thanks.

docs/src/conf.py
```diff
@@ -95,17 +95,12 @@ def autolog(message):
...
-.. |iris_version| replace:: {iris_version}
-.. |build_date| replace:: {build_date}
-""".format(
-    year_range=f"2010 - {upper_copy_year}",
-    python_version=".".join([str(i) for i in sys.version_info[:3]]),
-    iris_version=f"v{version}",
-    build_date=f"({datetime.datetime.now().strftime('%d %b %Y')})",
-)
+rst_epilog = f"""
+.. |copyright_years| replace:: {2010 - upper_copy_year}
```
jamesp, line 99: I think this should probably be: `2010 - {upper_copy_year}`?

bjlittle: Oopsie... good spot :eyes:

tkknight:
can we get this merged now @jamesp & @bjlittle ?

jamesp (approved):
Looks good to me 👍

PR #3984, opened 2021-02-07
Update from master

pp-mo:
### Description
Merge to update the "mesh-data-model" feature branch to = current master.
Nothing new has happened on mesh-data-model since it was cut, whereas master has some confusing reshuffling of the docs.
So, it will be good to equalise this f-b with master before starting to build on it.

---


pp-mo:
Tests are broken :-(

It seems that the introduction of nox has a nasty hidden cost : 
The testing env is no longer directly determined by requirements/ci/pyXX.yml, and you can no longer snapshot the env with a simple "conda list", because nox uses a combination of conda + pip to build the env.  That's a backward step for debugging Ci testing problems.
E.G. in this case _why the tests work in Python 3.7 but fail in 3.6_ :nauseated_face: . 
In the old world I would start by comparing the "conda list"s.

bjlittle:
@pp-mo You could always simply just delete the `mesh-data-model` feature branch, and fork it afresh from `master`... might be easier...

bjlittle:
> The testing env is no longer directly determined by requirements/ci/pyXX.yml, and you can no longer snapshot the env with a simple "conda list", because nox uses a combination of conda + pip to build the env. That's a backward step for debugging Ci testing problems.

`pip` is only used for the `flake8` and `black` sessions, otherwise `conda` is used to create the environment from the requirements `*.yml` file.

I don't understand your issue here :confused:

bjlittle (changes requested):
@pp-mo I'd suggest that you opt to re-cut the feature branch afresh from `master`

PR #3985, opened 2021-02-07
Show installed packages when running tests.

pp-mo:
### Description
To enable easier debugging of test failures.

It's complicated because, since adoption of nox, the testing environment is built with a combination of pip and conda.

---

### DIFF
--- a/noxfile.py
+++ b/noxfile.py
@@ -165,6 +165,10 @@ def tests(session):
 
     cache_cartopy(session)
     session.install("--no-deps", "--editable", ".")
+
+    # Always echo the env content for CI test failure debugging.
+    session.run("python", "tools/env_listing.py")
+
     session.run(
         "python",
         "-m",
--- a/noxfile.py
+++ b/noxfile.py
@@ -167,6 +167,7 @@ def tests(session):
     session.install("--no-deps", "--editable", ".")
 
     # Always echo the env content for CI test failure debugging.
+    session.install("beautifultable")
     session.run("python", "tools/env_listing.py")
 
     session.run(
--- a/tools/env_listing.py
+++ b/tools/env_listing.py
@@ -1,0 +1,124 @@
+#!/usr/bin/env python
+# Copyright Iris contributors
+#
+# This file is part of Iris and is released under the LGPL license.
+# See COPYING and COPYING.LESSER in the root of the repository for full
+# licensing details.
+"""
+Produce a listing output of the calling Python environment.
+
+Combine "conda list" with a "pip list" to describe all installed packages.
+
+"""
+from subprocess import check_output
+import beautifultable as bt
+
+
+def sanitise_lines(lines, header_line_hint=None):
+    """
+    Split string by spaces, removing leading, trailing and repeated whitespace.
+    Remove header lines up to the last one containing 'header_line_hint'.
+
+    """
+    pkg_info = []
+    for line in lines:
+        line = line.strip()
+        while '\t' in line:
+            line = line.replace('\t', ' ')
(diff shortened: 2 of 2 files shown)


noxfile.py
```diff
@@ -165,6 +165,11 @@ def tests(session):
 
     cache_cartopy(session)
     session.install("--no-deps", "--editable", ".")
+
+    # Always echo the env content for CI test failure debugging.
+    session.install("beautifultable")
+    session.run("python", "tools/env_listing.py")
```
bjlittle, line 171: @pp-mo Personally, I'd opt to add the following 3 line `conda_script` in `.cirrus-ci` instead:
```python
#
# YAML alias for common linux test infra-structure.
#
linux_task_template: &LINUX_TASK_TEMPLATE
  auto_cancellation: true
  ...
  nox_cache:
     ...
  conda_script:
    - conda list --name base
    - conda list --name base --explicit
    - conda info
```

bjlittle (changes requested):
@pp-mo Lovely work... but I don't think this is necessary, unless I've misunderstood.

Adding my suggested `conda_script` to the `.cirrus-ci` would align this with the output from the retired `.travis-ci`

pp-mo:
Well I did try adding "conda list" first, but I found a lot of the packages were missing, apparently because nox has installed them via pip instead ?
So, I put that just in the 'tests' function, where the 'env_listing' now is.

If you look at the outputs in the test logs, those with the source "pip" (rather than "conda" or "both") are not present in the "conda list" output.
Am I missing something here -- is it different if you add it in the common template, for some reason I'm not getting ?

pp-mo:
P.S. the original problem for which I introduced this was a failure to check out on python 3.7 in the latest merge to master :

The message was like this...
```
...
FAIL: test_self_referencing_load_issue_3367 (iris.tests.integration.test_netcdf.TestSelfReferencingVarLoad)
  File "/tmp/cirrus-ci-build/lib/iris/tests/integration/test_netcdf.py", line 405, in test_self_referencing_load_issue_3367
    warn.assert_called_with(expected_msg)
  File "/tmp/cirrus-ci-build/.nox/tests/lib/python3.7/unittest/mock.py", line 878, in assert_called_with
    raise AssertionError(_error_message()) from cause
AssertionError: Expected call: warn("Missing CF-netCDF measure variable 'areacello', referenced by netCDF variable 'volcello'")
Actual call: warn('`np.bool` is a deprecated alias for the builtin `bool`. To silence this warning, use `bool` by itself. Doing this will not modify any behavior and is safe. If you specifically wanted the numpy scalar type, use `np.bool_` here.\nDeprecated in NumPy 1.20; for more details and guidance: https://numpy.org/devdocs/release/1.20.0-notes.html#deprecations', <class 'DeprecationWarning'>, stacklevel=2)
Ran 5040 tests in 127.803s
FAILED (SKIP=22, failures=1)
nox > Command python -m iris.tests.runner --default-tests --system-tests failed with exit code 1
nox > Session tests failed.

```

So, I really thought that looked like a solid logical problem, possibly to do with a changed numpy version, and not a glitch.
But since I re-spun the Cirrus tests, the problem vanished. :grimacing:

pp-mo:
> @bjlittle  I'd opt to add the following 3 line conda_script in .cirrus-ci instead ...

Just trying that #3986 ..

bjlittle:
@pp-mo Could we close this in preference for #3990 ?

What do you think?

pp-mo:
> Could we close this in preference for #3990 ?
Absolutely.  You fixed it !

PR #3987, opened 2021-02-08
Add abstract cube summary

pp-mo:
### Description
A simple replacement for #3952
 * now based on current "master"
 * commits squashed

With this, I'm intending to bank what is now on #3952.
I still think there are some small improvements outstanding, but we can fix those later, along with adding my new cube summary code -- all in subsequent PRs.
So I'm now intending to **put this onto master** (as originally suggested by @bjlittle), 

Future plans... 
Subsequently when adding my new cube-summary code, I will find a quick way not to break the existing `_repr_html_` implementation.
I **think** I can make the string summary close enough in form that I can also make the existing `_repr_html_` code work with minimal changes.  If not, I will leave the existing cube summary code in place, under a different name, until we find time to fix it properly  -- which would be messy, but means we can get on.
So, **given that approach, I think we can still target master here**

---

### DIFF
--- a/lib/iris/_representation.py
+++ b/lib/iris/_representation.py
@@ -1,0 +1,273 @@
+# Copyright Iris contributors
+#
+# This file is part of Iris and is released under the LGPL license.
+# See COPYING and COPYING.LESSER in the root of the repository for full
+# licensing details.
+"""
+Provides objects describing cube summaries.
+"""
+
+import iris.util
+
+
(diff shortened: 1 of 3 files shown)


stephenworsley (approved):
Looks like this should be okay to merge. There's no breaking changes and it's all private so it should be safe even if it doesn't introduce the completed refactor.

PR #3982, opened 2021-02-05
Cubesummary nobt

pp-mo:
### Description
New code for first eyes.  Replaces #3962

Replaced beautifultable with minimalist home-brew solution for printing text entries in columns.

## Still **WIP** 
TODO : better code comments + proper testing
FUTURE :  This should convert into a follow-on to #3952, when that is done.  It will need resolving against changes in that.
IN PROGRESS : I'm working on replacing actual cube summaries with this, and fixing existing tests for that.

---
[Consult Iris pull request check list]( https://scitools-iris.readthedocs.io/en/latest/developers_guide/contributing_pull_request_checklist.html)

PR #3988, opened 2021-02-08
Cubesummary tidy

pp-mo:
### Description
Some small improvements I thought worthwhile.
As a preliminary to adding the new "table based summary" code, as  previewed in #3982

---

### DIFF
--- a/lib/iris/_representation.py
+++ b/lib/iris/_representation.py
@@ -6,8 +6,10 @@
 """
 Provides objects describing cube summaries.
 """
-import numpy as np
+import re
+
 import iris.util
+from iris.common.metadata import _hexdigest as quickhash
 
 
 class DimensionHeader:
@@ -46,6 +48,35 @@ class FullHeader:
         self.dimension_header = DimensionHeader(cube)
 
 
+def string_repr(text, quote_strings=False):
+    """Produce a one-line printable form of a text string."""
+    if re.findall("[\n\t]", text) or quote_strings:
+        # Replace the string with its repr (including quotes).
+        text = repr(text)
+    return text
+
+
+def array_repr(arr):
+    """Produce a single-line printable repr of an array."""
+    # First take whatever numpy produces..
+    text = repr(arr)
+    # ..then reduce any multiple spaces and newlines.
+    text = re.sub("[ \t\n]+", " ", text)
+    return text
+
+
+def value_repr(value, quote_strings=False):
+    """
+    Produce a single-line printable version of an attribute or scalar value.
+    """
+    if hasattr(value, "dtype"):
+        value = array_repr(value)
+    elif isinstance(value, str):
+        value = string_repr(value, quote_strings=quote_strings)
+    value = str(value)
+    return value
+
+
 class CoordSummary:
     def _summary_coord_extra(self, cube, coord):
         # Returns the text needed to ensure this coordinate can be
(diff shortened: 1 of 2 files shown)


pp-mo:
@stephenworsley effectively, this is the remaining suggestions left over from #3987

lib/iris/tests/unit/representation/test_representation.py
```diff
@@ -182,6 +182,56 @@ def test_cell_methods(self):
...
+        co1 = cube.coord("latitude")
+        co1.attributes.update(dict(a=1, array=np.array([1.2, 3])))
+        co2 = co1.copy()
+        co2.attributes.update(dict(b=2, array=np.array([3.2, 1])))
+        cube.add_aux_coord(co2, cube.coord_dims(co1))
+        rep = iris._representation.CubeSummary(cube)
+        co1_summ = rep.vector_sections["Dimension coordinates:"].contents[0]
+        co2_summ = rep.vector_sections["Auxiliary coordinates:"].contents[0]
+        self.assertEqual(co1_summ.extra, "array=array([1.2, 3. ])")
+        self.assertEqual(co2_summ.extra, "array=array([3.2, 1. ]), b=2")
```
stephenworsley, line 233: For the sake of testing numpy array equality,  it might also be worth testing the case where two arrays are equal and the case where arrays have different shapes.

lib/iris/_representation.py
```diff
@@ -66,7 +66,11 @@ def _summary_coord_extra(self, cube, coord):
                         vary.add(key)
                         break
                     value = similar_coord.attributes[key]
-                    if attributes.setdefault(key, value) != value:
+                    # Like "if attributes.setdefault(key, value) != value:"
+                    # ..except setdefault fails if values are numpy arrays.
+                    if key not in attributes:
+                        attributes[key] = value
+                    elif not np.all(attributes[key] == value):
```
stephenworsley, line 73: It looks like there exists logic in the metadata to handle this kind of comparison: https://github.com/SciTools/iris/blob/a136915a8421864a4bb18dcc06bade3b324ff1f3/lib/iris/common/metadata.py#L51
This would handle edge cases differently (dtype, masked arrays etc.)

pp-mo: Yes, that really does look like the right way.

pp-mo:
I got into some deep water there :  I found that the handling of array representations was highly undesirable due to embedded newlines.  So I have fixed that, and also for strings with newlines, in attribute values and coord-extras.

I have not added tests to check that it distinguishes arrays masked/unmasked or with different masks.  

Again, the existing code does not handle any of these things very well, and there is no existing testing.
So it feels like this might be overkill -- there seems to be no end to this stuff ?

pp-mo:
I am also realising that the presentation of scalar string coords is not escaping the text.  
Basically I think we should be probably be escaping  awkward strings here, and only presenting simple printable strings to representation handlers..
I think I will try and deal with this later, via tests in the printout code.

pp-mo:
Hang on with this...
I think if we're going to  do this at all we should have a more comprehensive approach to escaping the string results.
I'll think about that + maybe come up with a tidier solution...

pp-mo:
Ok, I think I'm happy that this now works reasonably. 
Text is now 'escaped' from effects of rogue linefeeds,and clipped to length for "oneline output"
in all relevant places  :
  * scalar test coord point
     * string attributes are only quoted when needed (embedded tabs_newlines)
  * cube attributes
     * as for the scalar string coord values
  * coord 'extra' strings (both scalar + vector)
     * string attributes are *always* quoted in this case
 
Not everything is totally tested (e.g. longer line clipping), but I think it will do.
If problems arise, we can add further regression tests, but I think we've already spent much too long on this :disappointed_relieved:

stephenworsley (approved):
Looks good, should be significantly more robust than the existing code.

PR #3995, opened 2021-02-10
move connectivity to experimental

bjlittle:
### Description
This PR migrates the `Connectivity` and `ConnectivityMetadata` classes (ref: #3968) to `iris.experimental.ugrid`.

Note that, this is targeting the `mesh-data-model` feature branch.

---


lib/iris/coords.py
```diff
@@ -625,7 +624,7 @@ def xml_element(self, doc):
         # otherwise.
         if isinstance(self, Coord):
             values_term = "points"
-        elif isinstance(self, Connectivity):
+        elif hasattr(self, "indices"):
```
bjlittle, line 627: Ideally, we'd want to change this back, but only once `iris.experimental.ugrid` is pulled in to the core of `iris`.

lib/iris/common/metadata.py
```diff
@@ -1577,46 +1456,42 @@ def values(self):
...
-    ConnectivityMetadata.__eq__,
-    ConnectivityMetadata.equal,
     CoordMetadata.__eq__,
     CoordMetadata.equal,
     CubeMetadata.__eq__,
     CubeMetadata.equal,
     DimCoordMetadata.__eq__,
     DimCoordMetadata.equal,
-)
+]
```
bjlittle, line 1494: Ideally, we'd want to change these back to `tuple`s once `iris.experimental.ugrid` is merged into the core of `iris`.

lib/iris/common/metadata.py
```diff
@@ -1577,46 +1456,42 @@ def values(self):
 
 
 #: Convenience collection of lenient metadata combine services.
-SERVICES_COMBINE = (
+SERVICES_COMBINE = [
```
trexfeathers, line 1459:
```suggestion
# TODO: change lists back to tuples once CellMeasureMetadata is re-integrated
# here (currently in experimental.ugrid).
SERVICES_COMBINE = [
```

lib/iris/experimental/ugrid.py
```diff
@@ -5,42 +5,632 @@
...
+
+
+__all__ = [
+    "Connectivity",
+    "ConnectivityMetadata",
+]
+
+
+# Configure the logger.
+logger = get_logger(__name__, fmt="[%(cls)s.%(funcName)s]")
```
trexfeathers, line 39: I'm not clear why this is necessary? `logger` isn't used here. Perhaps it's used by the parent classes, but in that case I assume we wouldn't need to import it here too?

trexfeathers (changes requested):
Thanks @bjlittle! A few minor changes/clarifications please 🙂

lib/iris/experimental/ugrid.py
bjlittle: It's interesting that `flake8` doesn't complain about this... I though that it might.

Happy to remove it :+1:

trexfeathers:
I really like this concept, since it makes it possible to release some of these features before too long. There may be some Iris issues that have gone on the back burner for a while, but we can show the world that we're still working hard!

trexfeathers (approved):

bjlittle:

PR #3998, opened 2021-02-10
Cube printout

pp-mo:
### Description
Adds cube printout code, based on `iris._representation.CubeSummary`.
This supercedes #3982

N.B. we already merged the [`iris._representation.SubeCummary` PR](https://github.com/SciTools/iris/pull/3987).
As it happens that is all private, and nothing else yet uses it, so we can still evolve that if needed.

Eventually we will replace the existing `Cube.summary` code with this (but not just yet)

---


pp-mo:
At this point, nearly there.
I still need to add tests for cell methods, but I need to mug up on what they can contain.

pp-mo:
Ok, finally complete !

Please review.

lib/iris/_representation/cube_printout.py
```diff
@@ -0,0 +1,348 @@
...
+"""
+from copy import deepcopy
+
+
+class Table:
+    """
+    A container of text strings in rows + columns, that can format its content
+    into a string per row, with contents in columns of fixed width.
+
+    Supports left- or right- aligned columns, alignment being set "per row".
```
pp-mo, line 18: Mistake here 
```suggestion
    Supports left- or right- aligned columns, alignment being set "per column".
```

pp-mo:
Following offline discussion with @stephenworsley , we think we are now fairly confident in the "old/new equivalence testing" here, and the scope of the new tests.
So the next stage is to actually remove Pyke and simplify everything, including the "equivalence" testing
Closing in favour of #4206

PR #4002, opened 2021-02-12
MeshMetadata class.

trexfeathers:
### Description

Implemented the `MeshMetadata` class and tests, in readiness to support the `Mesh` class.

---

### DIFF
--- a/lib/iris/common/metadata.py
+++ b/lib/iris/common/metadata.py
@@ -181,9 +181,17 @@ class BaseMetadata(metaclass=_NamedTupleMeta):
                     return result
 
                 # Note that, for strict we use "_fields" not "_members".
-                # The "circular" and "src_dim" members do not participate in strict equivalence.
+                # Certain members never participate in strict equivalence, so
+                # are filtered out.
                 fields = filter(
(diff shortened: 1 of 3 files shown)


trexfeathers:
@bjlittle can you clarify: do we definitely not want to include `node_dimension`, `edge_dimension` and `face_dimension` (and in future `volume_dimension`) in `MeshMetadata._members`? I don't have a good sense of what would (dis)qualify them, and I can't remember if we discussed. Thanks

bjlittle:
@trexfeathers From our discussion, what I recall is that `node_dimension` et al **are** a part of the `MeshMetadata`. However they **do not** participate in `__eq__`.

trexfeathers:
>I _think_ this is true for both strict and lenient

It should be the same behaviour for either, since we do not plan to allow for `None` values of `node_dimension` et al anyway (we fill with a default if no alternative is supplied).

trexfeathers:
Ready for your review again @bjlittle. I've added the extra members for the dimension names.

bjlittle (approved):
@trexfeathers Awesome LGTM :+1: 

Thanks!

PR #4014, opened 2021-02-17
MeshCoordMetadata [AVD-1629]

pp-mo:
### Description

Adds MeshCoordMetadata class (with tests)
Preliminary to creating a MeshCoord

Note: also fixed some problems I stumbled over in existing metadata testing.
Some of which  had already been copied into other tests! ...
There **_is_** a lot of boilerplate here : 
much of the  code in `lib/iris/tests/unit/experimental/ugrid/test_MeshCoordMetadata.py` is a total copy 
of that in `lib/iris/tests/unit/experimental/ugrid/test_ConnectivityMetadata.py` (i.e. ultimately from #3968)

As we are in a hurry, we are resolved to fix that "later"... 

---

### DIFF
--- a/lib/iris/experimental/ugrid.py
+++ b/lib/iris/experimental/ugrid.py
@@ -46,6 +46,7 @@ __all__ = [
     "MeshFaceCoords",
     "MeshNodeCoords",
     "MeshMetadata",
+    "MeshCoordMetadata",
 ]
 
 
@@ -1989,32 +1990,11 @@ class _Mesh2DConnectivityManager(_MeshConnectivityManagerBase):
         return self._members["face_node_connectivity"]
 
 
-#: Convenience collection of lenient metadata combine services.
-_services = [ConnectivityMetadata.combine, MeshMetadata.combine]
-SERVICES_COMBINE.extend(_services)
-SERVICES.extend(_services)
-#: Convenience collection of lenient metadata difference services.
-_services = [ConnectivityMetadata.difference, MeshMetadata.difference]
-SERVICES_DIFFERENCE.extend(_services)
-SERVICES.extend(_services)
-
-#: Convenience collection of lenient metadata equality services.
-_services = [
-    ConnectivityMetadata.__eq__,
-    ConnectivityMetadata.equal,
-    MeshMetadata.__eq__,
-    MeshMetadata.equal,
-]
-SERVICES_EQUAL.extend(_services)
-SERVICES.extend(_services)
-
-del _services
-
-
 class MeshCoordMetadata(BaseMetadata):
     """
     Metadata container for a :class:`~iris.coords.MeshCoord`.
     """
+
     _members = ("location", "axis")
     # NOTE: in future, we may add 'mesh' as part of this metadata,
     # as the Mesh seems part of the 'identity' of a MeshCoord.
@@ -2137,8 +2117,18 @@ _op_names_and_service_collections = [
     ("__eq__", SERVICES_EQUAL),
     ("equal", SERVICES_EQUAL),
 ]
-for _cls in (ConnectivityMetadata, MeshCoordMetadata):
+_metadata_classes = [ConnectivityMetadata, MeshMetadata, MeshCoordMetadata]
+for _cls in _metadata_classes:
     for _name, _service_collection in _op_names_and_service_collections:
         _method = getattr(_cls, _name)
         _service_collection.append(_method)
         SERVICES.append(_method)
+
+del (
+    _op_names_and_service_collections,
+    _metadata_classes,
+    _cls,
+    _name,
+    _service_collection,
+    _method,
+)
--- a/lib/iris/experimental/ugrid.py
+++ b/lib/iris/experimental/ugrid.py
@@ -1993,8 +1993,6 @@ class _Mesh2DConnectivityManager(_MeshConnectivityManagerBase):
 _services = [ConnectivityMetadata.combine, MeshMetadata.combine]
 SERVICES_COMBINE.extend(_services)
 SERVICES.extend(_services)
-
-
 #: Convenience collection of lenient metadata difference services.
 _services = [ConnectivityMetadata.difference, MeshMetadata.difference]
 SERVICES_DIFFERENCE.extend(_services)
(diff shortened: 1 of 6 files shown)


stephenworsley (approved):
Looks good.

trexfeathers:
>Note: also fixed some problems I stumbled over in existing metadata testing.

Thanks for the catch @pp-mo!

pp-mo:
Sorry all -- I failed to include all the changes, at first.
Now fixed

lib/iris/experimental/ugrid.py
```diff
@@ -615,17 +615,130 @@ def equal(self, other, lenient=None):
...
+    @lenient_service
+    def equal(self, other, lenient=None):
+        return super().equal(other, lenient=lenient)
+
+
+# Add our new optional metadata operations into the 'convenience collections'
+# of lenient metadata services.
+# TODO: when included in 'iris.common.metadata', install each one directly ?
+for cls in (ConnectivityMetadata, MeshCoordMetadata):
+    op_names_and_collections = {
```
stephenworsley, line 735: It might make more sense to define `op_names_and_collections` outside the loop.

lib/iris/experimental/ugrid.py
```diff
@@ -615,17 +615,130 @@ def equal(self, other, lenient=None):
...
+
+        * other (MeshCoordMetadata):
+            The other MeshCoord metadata participating in the lenient
+            difference.
+
+        Returns:
+            A list of different metadata member values.
+
+        """
+        # Perform "strict" difference for "cf_role", "start_index", "src_dim".
```
bjlittle, line 701: @pp-mo You just need to update this comment to align it with its members 👍

lib/iris/experimental/ugrid.py
bjlittle: @pp-mo If you do make convenience variables at the module level, could you also please purge them (`del`) from the namespace please 👍

bjlittle:
@pp-mo Awesome stuff! A few suggested changes, but LGTM. Great work 😀

We can start using the `MeshCoordMetadata` in anger to see how it feels, and revisit `axis` at a later date (soonish) to see what we think about it... plus we may have some concrete use case data that helps sway up in an informed direction. Regardlessly, changing the makeup of the `MeshCoordMetadata` is a cheap and easy step to take now 👍

bjlittle (changes requested):
@pp-mo Awesome stuff! Just a few suggested changes, but LGTM. Great work 😀

We can start using the `MeshCoordMetadata` in anger to see how it feels, and revisit `axis` at a later date (soonish) to see what we think about it... plus we may have some actual concrete use case data that helps us weigh up and make an informed decision. 

Regardlessly, changing the makeup of the `MeshCoordMetadata` is a cheap and easy step to take now 👍

🎉 🥳🍻

lib/iris/experimental/ugrid.py
pp-mo: 👍 😊

lib/iris/experimental/ugrid.py
pp-mo: Awseome spot !
👍

pp-mo: Just as an afterthought : I'm a bit weak on this...
Would you consider it equivalent to define an `__all__`, or use only underscore-prefixed things in the package namespace, or is it really that much better to totally tidy it ??

( I always felt that trying to fix this was somewhat swimming against the tide, given the way that imports mess it all up anyway )

stephenworsley (changes requested):
Just a couple more comments about the testing along with the existing comments. Otherwise, this looks good!

pp-mo:
Hoping that latest addresses @bjlittle comments. Ping!

Still thinking about @stephenworsley tests changes.
Will respond on that shortly..

pp-mo:
Rebased to fix problem with tests.

lib/iris/experimental/ugrid.py
pp-mo: Done!

lib/iris/experimental/ugrid.py
pp-mo: Done, I think.

lib/iris/tests/unit/experimental/ugrid/test_MeshCoordMetadata.py
```diff
@@ -0,0 +1,740 @@
...
+        rmetadata = self.cls(**self.values)
+
+        with mock.patch("iris.common.metadata._LENIENT", return_value=False):
+            self.assertTrue(lmetadata.__eq__(rmetadata))
+            self.assertTrue(rmetadata.__eq__(lmetadata))
+
+    def test_op_strict_different(self):
+        lmetadata = self.cls(**self.values)
+        right = self.values.copy()
+        right["long_name"] = self.dummy
```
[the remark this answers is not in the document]
pp-mo: Effectively I "inherited" all this from the test_ConnectivityMetadata.
I have now "rationalised" the use of the non-member metadata in these tests : uses of "units" and "var_name" have been replaced with "long_name".  I thought that made some sense, as long_name can be just anything whereas units=None would read back as "Unit('unknown')" if using a real object.

stephenworsley (approved):

lib/iris/experimental/ugrid.py
```diff
@@ -615,17 +615,130 @@ def equal(self, other, lenient=None):
...
+class MeshCoordMetadata(BaseMetadata):
+    """
+    Metadata container for a :class:`~iris.coords.MeshCoord`.
 
+    """
 
-#: Convenience collection of lenient metadata difference services.
-SERVICES_DIFFERENCE.append(ConnectivityMetadata.difference)
-SERVICES.append(ConnectivityMetadata.difference)
+    _members = ("mesh", "location", "axis")
```
bjlittle, line 624: @pp-mo Whoa... are you really suggesting that `mesh` is metadata here?

This is the payload of the `MeshCoord`, right? If so, then I don't agree with this... it's akin to making `points` and `bounds` members of the `CoordMetadata` or `DimCoordMetadata` :-1: 

Or do you mean `mesh` to be something different?

trexfeathers: Obviously having the actual `Mesh` as part of the metadata wouldn't work, but given the idea of shared `Mesh`es I'd say some sort of mesh identifier is an important component of the `MeshCoord` metadata.

bjlittle: If `mesh` is a cheap comparison, then yes this could be part of the metadata, but until we know better I think we should err on the side of this not being part the metadata

bjlittle: Also, `__eq__` hasn't been defined within `iris.experimental.ugrid.Mesh` for this very reason... we're kinda sitting on the fence until we know better

pp-mo:
Ok, now removed 'mesh' from the metadata, and rebased to accomodate the latest on the mesh-data-model target.
Also tidied up the 'service collections' adjustments, now all handled at the end of 'ugrid.py'

lib/iris/experimental/ugrid.py
pp-mo: Ok removed it

pp-mo:
@stephenworsley @bjlittle think we have addressed all previous issues now.
Sorry about the rebase.
Please re-review !

bjlittle (approved):

bjlittle:
@pp-mo and @stephenworsley LGTM :+1:

PR #4016, opened 2021-02-18
Meshcoord dataview daskidea

pp-mo:
### Description
A demonstration of how the MeshCoord points/bounds calculation could be done **with Dask**.

A new test module shows how we can construct an AuxCoord mimicking what we want from a MeshCoord,
and adds various tests that demonstrate desired behaviour.

One desirable thing that is _missing_, is that the dask operations used don't allow an efficient partial access : i.e. if you compute some portion of a result (even the points access, which is simpler), it _always_ fetches the whole data, before sub-indexing that.

We have found before that this kind of problem **_can_** often be fixed, with a different construction of the Dask calculation.
In any case, given the expected semi-random ordering of information relative to 'useful' data regions, we expect that effectively enabling this may require a more involved solution.
Having simply confirmed what it now does, I am leaving any further investigation to the future.

---

### DIFF
--- a/lib/iris/tests/unit/experimental/ugrid/test_meshcoord__dataview.py
+++ b/lib/iris/tests/unit/experimental/ugrid/test_meshcoord__dataview.py
@@ -1,0 +1,440 @@
+# Copyright Iris contributors
+#
+# This file is part of Iris and is released under the LGPL license.
+# See COPYING and COPYING.LESSER in the root of the repository for full
+# licensing details.
+"""Unit tests for the :class:`iris.experimental.ugrid.Connectivity` class."""
+
+# Import iris.tests first so that some things can be initialised before
+# importing anything else.
+import iris.tests as tests
+
+import dask.array as da
+import numpy as np
+
+from iris.cube import Cube
+from iris.coords import AuxCoord, DimCoord
+
+
+def array_index_with_missing(array, index_array):
+    # Index the first array with the second,
+    # EXCEPT putting NaNs where any index is < 0.
+    # Works on *either* real or lazy arrays.
+    result = da.where(index_array < 0, np.nan, array[index_array])
+    return result
+
+
+def create_meshcoordlike(
+    cube, face_x_coordname, node_x_coordname, face_node_inds_coordname
+):
+    """
+    Proof of Concept to create an AuxCoord which mimics intended MeshCoord
+    behaviour, for fetching points + bounds.
+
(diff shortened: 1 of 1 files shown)


lib/iris/tests/unit/experimental/ugrid/test_meshcoord__dataview.py
```diff
@@ -0,0 +1,440 @@
...
+    # ...so we build that calculation by iterating over the final (bounds)
+    # index, and stacking those to reconstruct the output.
+    # Within that operation, we also apply our 'array_index_with_missing'
+    # function, which ensures that any -1 indices yield NaN points.
+    lazy_face_bounds_x = da.stack(
+        [
+            array_index_with_missing(
+                lazy_node_x, lazy_facenodeinds[:, i_bound]
+            )
+            for i_bound in range(4)
```
stephenworsley, line 153: Range should vary with face index shape.

pp-mo: Good spot.
Will do this

pp-mo:
It was noticed in discussion with @stephenworsley that 
  1. in `test_partial_access_points` I had as my "partial access" `mesh_coord.points[1:2]`.  Which of course fetches the whole data, and should really use `mesh_coord[1:2].points` or maybe `mesh_coord.lazy_points()[1:2].compute`.
 2. the lazy-bounds calculation based on 'da.stack' could be done more neatly by using flattened indices and then reshaping
 
I have since corrected (1.) and tried (2.) as an alternative calculation method.

This enabled me to prove that partial fetches of mesh_coord points and bounds values **_can_** result in  fetching only _part_ of the source data arrays. 
I also introduced some non-trivial chunking, and investigated the effects of that too.
The results are now documented in the code, with some alternatives and their effects noted.

So I've fixed the tests to specific results, and all pass.
But I have left in the code for some unused alternatives,  which we would not wish to merge as-is.
So I think this Draft PR now documents the investigation + some findings, but is obviously not suitable for merging.

If we want to merge some of this code as an actual solution, we should cut this down + present in a new PR.

pp-mo:
More ideas here : https://github.com/pp-mo/iris/pull/65
This is very much a performance-driven implementation detail.  But I think it show the nasty problem that Dask has here, and one  way around it.

pp-mo:
Closing, as this PR has served its purpose as a basis for discussion.
We will include some of these ideas in forthcoming development to add a MeshCoord class.

PR #4020, opened 2021-02-22
make iris.common.metadata._hexdigest public

bjlittle:
### Description
This PR makes the `iris.common.metadata._hexdigest` function public i.e., `iris.common.metadata.hexdigest`.

Originally this function was written to solely support the needs of the common metadata API. However, it is already being used elsewhere within `iris`, and there are plans to leverage its benefit in other parts of `iris`.

Additionally, it may be of use to the `iris` community.

---

### DIFF
--- a/docs/src/whatsnew/latest.rst
+++ b/docs/src/whatsnew/latest.rst
@@ -42,6 +42,12 @@ This document explains the changes made to Iris for this release
    ``iris.plot.plot(z_cube)`` will produce a z-vs-phenomenon plot, where before
    it would have produced a phenomenon-vs-z plot. (:pull:`3906`)
 
+#. `@bjlittle`_ introduced :func:`iris.common.metadata.hexdigest` to the
+   public API. Previously it was a private function introduced in ``v3.0.0``.
+   Given any object, :func:`~iris.common.metadata.hexdigest` returns a string
+   representation of the 64-bit non-cryptographic hash of the object using the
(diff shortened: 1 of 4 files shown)


trexfeathers (approved):
Yep, totally up for this 👍

I can confirm that all references have been changed from `_hexdigest` to `hexdigest`.

jamesp:
Sounds good! What is the interaction of calling this method with lazy data?

bjlittle:
It's simply going to checksum the `item` that it's given, so it wouldn't realise lazy data.

If we want that to happen then that's an extension to the current behaviour, and also it will have to be advertised as realising anything lazy to users.

We may want to think about this i.e., give the caller the option to realise and checksum, or checksum as is. I guess my point is that `hexdigest` should not make assumptions, and it's on the caller to be explicit about what should happen to lazy data and what should be checksummed.

At present it's not been used by `iris` to checksum lazy data, but users may and will probably do this.

@jamesp Fancy making an issue so that we don't forget about this?

jamesp:
Happy to make an issue.  I'm not sure what the right behaviour is, but it might be worth making it explicit in the docstring what the interaction is if hexdigest is now going to be a public method.

As a user, I would expect a hexdigest to be "immutable" in the sense that it represents the underlying object, and only changes if the object changes.  I think it would be undesirable to have it return something different depending on whether the data is realised or not - that is an implementation detail that should probably be hidden from the user, and may not even be known by the user / could change depending on order of operations, or even introspection.

PR #4033, opened 2021-02-24
add mesh __str__ and __repr__ support

bjlittle:
### Description
This PR adds `__str__` and `__repr__` support for the `Mesh` et al.

Note that, for the managers and `Connectivity` I've opted to make the `__repr__` unambiguous, whereas the `__str__` is a slicker summary of content.

For the `Mesh`, which is metadata and data rich, I've not made this distinction i.e., only implemented `Mesh.__repr__`, hence `str(mesh)` will return the same result as `repr(mesh)`.

---

### DIFF
--- a/lib/iris/experimental/ugrid.py
+++ b/lib/iris/experimental/ugrid.py
@@ -13,6 +13,7 @@ CF UGRID Conventions (v1.0), https://ugrid-conventions.github.io/ugrid-conventio
 from abc import ABC, abstractmethod
 from collections import Iterable, namedtuple
 from functools import wraps
+import re
 
 import dask.array as da
 import numpy as np
@@ -47,6 +48,12 @@ __all__ = [
     "MeshNodeCoords",
     "MeshMetadata",
 ]
+
+
+#: Numpy "threshold" printoptions default argument.
+NP_PRINTOPTIONS_THRESHOLD = 10
+#: Numpy "edgeitems" printoptions default argument.
+NP_PRINTOPTIONS_EDGEITEMS = 2
 
 
 # Configure the logger.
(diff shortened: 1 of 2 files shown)


lib/iris/experimental/ugrid.py
```diff
@@ -204,6 +211,58 @@ def validate_arg_vs_list(arg_name, arg, valid_list):
...
+                        and not v
+                    ):
+                        result = False
+                    elif k == "units" and v == "unknown":
+                        result = False
+            return result
+
+        def array2repr(array):
+            if self.has_lazy_indices():
+                result = repr(array)
```
bjlittle, line 232: Note that, for the first time this will keep the payload lazy after performing a `repr(...)`.

As a follow-on pull-request, I'm going to apply this to all `_DimensionalMetadata` for `__repr__` and also make the `__str__` a slicker summary that also doesn't realise any lazy data... an obvious and small win :+1:

lib/iris/experimental/ugrid.py
```diff
@@ -932,8 +991,75 @@ def __ne__(self, other):
...
+                    if v is not None:
+                        result = True
+                        if (
+                            not isinstance(v, str)
+                            and isinstance(v, Iterable)
+                            and not v
+                        ):
+                            result = False
+                        elif k == "units" and v == "unknown":
+                            result = False
```
bjlittle, line 1051: If the `units` are `unknown`, then don't print them... this is the `units` equivalent to being `None`

lib/iris/experimental/ugrid.py
```diff
@@ -204,6 +211,58 @@ def validate_arg_vs_list(arg_name, arg, valid_list):
...
+                if v is not None:
+                    result = True
+                    if (
+                        not isinstance(v, str)
+                        and isinstance(v, Iterable)
+                        and not v
+                    ):
+                        result = False
+                    elif k == "units" and v == "unknown":
+                        result = False
```
bjlittle, line 227: If the `units` are `unknown`, then don't print them... this is the `units` equivalent to being `None`

lib/iris/experimental/ugrid.py
```diff
@@ -832,13 +891,13 @@ def __init__(
         topology_dimension,
         node_coords_and_axes,
         connectivities,
+        edge_coords_and_axes=None,
+        face_coords_and_axes=None,
```
bjlittle, line 895: @trexfeathers Hmmm I'm thinking it's better to group the `data` likes together, before the `metadata` likes...

What do you think? :thinking:

trexfeathers: Yes this is a little clearer, I'm game

trexfeathers (approved):
Nicely thought out @bjlittle ; I agree with your decisions, and everything behaves as expected under testing 💐

PR #4040, opened 2021-02-26
metadata string reduction

rcomer:
### Description
Suppose I make, for example, a coordinate, and print it

```python
coord = iris.coords.DimCoord(range(3), long_name='foo', units='m')
print(coord)
```
I get info about what it contains:
```
DimCoord(array([0, 1, 2]), standard_name=None, units=Unit('m'), long_name='foo')
```

However, if I just print the metadata,
```python
print(coord.metadata)
```
I see all the unset stuff too, which I'm not really interested in:
```
DimCoordMetadata(standard_name=None, long_name='foo', var_name=None, units=Unit('m'), attributes={}, coord_system=None, climatological=False, circular=False)
```
For differences, it's hard to see at a glance what is actually different:
```python
coord2 = iris.coords.DimCoord(range(3), long_name='foo', units='m', attributes=dict(bar=1), var_name='wibble')
print(coord2.metadata.difference(coord.metadata))
```
```
DimCoordMetadata(standard_name=None, long_name=None, var_name=('wibble', None), units=None, attributes=({'bar': 1}, {}), coord_system=None, climatological=None, circular=None)
```
With my branch, these outputs become:
```
DimCoordMetadata(long_name=foo, units=m, attributes={}, climatological=False, circular=False)
DimCoordMetadata(var_name=('wibble', None), attributes=({'bar': 1}, {}))
```
Note that, if you do want to see all the unset metadata, you can still use the `repr`.  We could potentially also skip empty attributes dictionaries for the `str`.

Does something like this need tests?  I had a peek at the [coord string tests](https://github.com/SciTools/iris/blob/cbaa95524649378779f148cf5897611f208c76fb/lib/iris/tests/unit/coords/test_Coord.py#L882), but they seem to be focussed on getting date formats right.

---

### DIFF
--- a/docs/src/whatsnew/latest.rst
+++ b/docs/src/whatsnew/latest.rst
@@ -48,6 +48,9 @@ This document explains the changes made to Iris for this release
    representation of the 64-bit non-cryptographic hash of the object using the
    extremely fast `xxhash`_ hashing algorithm. (:pull:`4020`)
 
+#. `@rcomer`_ implemented a string method for metadata classes, so printing
+   these objects skips metadata elements that are set to None. (:pull:`4040`)
+
 
 🐛 Bugs Fixed
 =============
--- a/docs/src/whatsnew/latest.rst
+++ b/docs/src/whatsnew/latest.rst
@@ -48,8 +48,9 @@ This document explains the changes made to Iris for this release
    representation of the 64-bit non-cryptographic hash of the object using the
    extremely fast `xxhash`_ hashing algorithm. (:pull:`4020`)
 
-#. `@rcomer`_ implemented a string method for metadata classes, so printing
-   these objects skips metadata elements that are set to None. (:pull:`4040`)
+#. `@rcomer`_ implemented a ``__str__`` method for metadata classes, so
+   printing these objects skips metadata elements that are set to None or an
+   empty string or dictionary. (:pull:`4040`)
 
 
 🐛 Bugs Fixed
--- a/lib/iris/common/metadata.py
+++ b/lib/iris/common/metadata.py
@@ -231,8 +231,9 @@ class BaseMetadata(metaclass=_NamedTupleMeta):
         field_strings = []
         for field in self._fields:
             value = getattr(self, field)
-            if value is not None:
-                field_strings.append(f"{field}={value}")
+            if value is None or isinstance(value, (str, dict)) and not value:
+                continue
+            field_strings.append(f"{field}={value}")
 
         return f"{type(self).__name__}({', '.join(field_strings)})"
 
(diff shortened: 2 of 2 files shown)


bjlittle:
@rcomer Hey, thanks for pushing this :+1: 

I'm about to make a similar change to `_DimensionalMetadata` for having a slimmer `__str__` as opposed to an unambiguous `__repr__` ... btw it will also allow you to keep data lazy rather realizing it just to print it.

docs/src/whatsnew/latest.rst
```diff
@@ -48,6 +48,9 @@ This document explains the changes made to Iris for this release
    representation of the 64-bit non-cryptographic hash of the object using the
    extremely fast `xxhash`_ hashing algorithm. (:pull:`4020`)
 
+#. `@rcomer`_ implemented a string method for metadata classes, so printing
```
bjlittle, line 51: Perhaps explicitly clarify, with `string` -> `__str__` ?

lib/iris/common/metadata.py
```diff
@@ -227,6 +227,15 @@ def __ne__(self, other):
 
         return result
 
+    def __str__(self):
+        field_strings = []
+        for field in self._fields:
+            value = getattr(self, field)
+            if value is not None:
```
bjlittle, line 234: @rcomer I think that we could be even more liberal here and ignore empty strings perhaps, but definitely and empty `dict` for the `attributes` member.

bjlittle (changes requested):
@rcomer Awesome, thanks! :smile: 

Just a couple of suggestions... also is you wanted to add some specific tests to exercise this, then that would be great :+1:

lib/iris/common/metadata.py
rcomer: It took me longer than I care to admit to figure out the logic for this 😆

rcomer:
Hi @bjlittle, I've addressed those specific comments.  Do you have anything particular in mind for tests?  I haven't come up with any prime candidate for "this might go wrong", so was just thinking of checking the method works without crashing for one example out of each subclass.  It would be good to know that's the Right Way Forward before embarking on that though.

lib/iris/common/metadata.py
rcomer: In fact it's still not right - I appear to have lost my boolean (False) elements somehow...

rcomer:
I can confirm that this method gets called indirectly by 60 other tests because, well, I broke them!

bjlittle:
@rcomer I think this is totally fine TBH.

I may add two `iris.common.metadata.BaseMetadata` tests just to be explicit.

I'll push that separately, if you fancy reviewing and merged?

bjlittle (approved):

bjlittle:
@rcomer See #4043 :+1:

rcomer:
Sweet.  Thanks @bjlittle!

PR #4046, opened 2021-03-02
fix intersection out of bounds point

### LINKED ISSUE #3391: cube.intersection out of bounds point
Example: when the upper bound of a point_x is at 0, the following code returns point_x, instead of wrapping the point to 360-point_x:
`target_grid.intersection(iris.coords.CoordExtent('longitude', 0, 360))`

rcomer:
### Description
Fixes #3391 I think.  This addresses the very specific case where you request the entire 360 degrees, and one of the bounds is at the maximum/minimum threshold.  Taking @valeriupredoi's example from #3391, you would still get that negative longitude point if you did something like `cube.intersection(longitude=(0, 50))`.  I'm not sure that that makes sense but, looking at the tests, it appears to have been intentional.

I haven't added any extra comment to explain my logic, but I think it's covered by the existing:
https://github.com/SciTools/iris/blob/a9ae72fa3c420da408fb4f69bba4035c2c43f769/lib/iris/cube.py#L3178-L3180

---


bjlittle:
@rcomer Hmmm you may have to re-open a new commit based on `v3.0.x` and cherry-pick out your commits onto it... or whatever is easiest.

As it stands, this PR is pulling unwanted `master` commits onto the `v3.0.x` release feature branch. Soz.

rcomer:
😱

rcomer:
@bjlittle see #4059

PR #4052, opened 2021-03-05
Added AtmosphereSigmaFactory

### LINKED ISSUE #4051: Support of AtmosphereSigmaFactory
## ✨ Feature Request

`iris.aux_factory` does not provide an aux factory for [`atmosphere_sigma_coordinate`](https://cfconventions.org/Data/cf-conventions/cf-conventions-1.7/cf-conventions.html#_atmosphere_sigma_coordinate) yet. It would be nice to have a class `AtmosphereSigmaFactory` similar to [`iris.aux_factory.HybridPressureFactory`](https://scitools-iris.readthedocs.io/en/stable/generated/api/iris/aux_factory.html#iris.aux_factory.HybridPressureFactory).

## Motivation

Right now, we have a custom [`AtmosphereSigmaFactory`](https://github.com/ESMValGroup/ESMValCore/blob/0c798617b885777a5ffa1697c39bb46974a73bab/esmvalcore/cmor/_fixes/shared.py#L18) implemented into ESMValTool. This works nicely, but needs to be added manually to the cube after loading the file. By including this class to iris an [automatic loading](https://github.com/SciTools/iris/blob/4047aa1833966f5838f64207c26679990c2d1116/lib/iris/fileformats/netcdf.py#L660) of the aux factory would be possible.

Since we already have this custom version of `AtmosphereSigmaFactory` in the ESMValTool repository I will try to implement this feature into iris.

schlunma:
### Description

`iris.aux_factory` does not provide an aux factory for [`atmosphere_sigma_coordinate`](https://cfconventions.org/Data/cf-conventions/cf-conventions-1.7/cf-conventions.html#_atmosphere_sigma_coordinate) yet. This PR adds the class `AtmosphereSigmaFactory` similar to [`iris.aux_factory.OceanSigmaFactory`](https://scitools-iris.readthedocs.io/en/stable/generated/api/iris/aux_factory.html#iris.aux_factory.OceanSigmaFactory). Most of the code for the new class is taken from this. Right now, we have a custom [`AtmosphereSigmaFactory`](https://github.com/ESMValGroup/ESMValCore/blob/0c798617b885777a5ffa1697c39bb46974a73bab/esmvalcore/cmor/_fixes/shared.py#L18) implemented into ESMValTool. This works nicely, but needs to be added manually to the cube after loading the file.

Closes #4051.

Note: The "What's new" is still missing, I will add that later!

Thanks!

---


schlunma:
Just signed the CLA :+1:

pp-mo:
Hi @schlunma 
Sorry for such a long delay on this, we have been rather short of time + pulled in all directions !
I see it now needs at least a(nother) rebase -- sorry about that.
Can we see about getting this finished now, within next 2-3 weeks ?
If so, we can get it into [Iris 3.1.0](https://github.com/SciTools/iris/discussions/4072), which would be great.

schlunma:
Hi @pp-mo, thanks! I will try to get the `rebase` done as quickly as possible, just had a quick look and saw that this probably requires lots of changes with `pyke` gone.

schlunma:
This is ready for review now! Required less changes than expected :+1:

lib/iris/tests/unit/aux_factory/test_AtmosphereSigmaFactory.py
```diff
@@ -0,0 +1,285 @@
...
+                pressure_at_top=self.pressure_at_top,
+                sigma=None,
+                surface_air_pressure=self.surface_air_pressure,
+            )
+        with self.assertRaises(ValueError):
+            AtmosphereSigmaFactory(
+                pressure_at_top=self.pressure_at_top,
+                sigma=self.sigma,
+                surface_air_pressure=None,
+            )
```
lbdreyer, line 56: Could you break down the tests here (and in the rest of the file) so that there is one call `AtmosphereSigmaFactory` (or whichever method it is testing) per test? For example, `test_insufficient_coordinates` would look like:

```
    def test_insufficient_coordinates_no_args(self):
        with self.assertRaises(ValueError):
            AtmosphereSigmaFactory()

    def test_insufficient_coordinates_no_ptop(self):
        with self.assertRaises(ValueError):
            AtmosphereSigmaFactory(
                pressure_at_top=None,
                sigma=self.sigma,
                surface_air_pressure=self.surface_air_pressure,
            )

     def test_insufficient_coordinates_no_sigma(self):
        with self.assertRaises(ValueError):
            AtmosphereSigmaFactory(
                pressure_at_top=self.pressure_at_top,
                sigma=None,
                surface_air_pressure=self.surface_air_pressure,
            )

    def test_insufficient_coordinates_no_ps(self):            
        with self.assertRaises(ValueError):
            AtmosphereSigmaFactory(
                pressure_at_top=self.pressure_at_top,
                sigma=self.sigma,
                surface_air_pressure=None,
            )
```

I see that in the unit tests for other aux coord factories we have a similar pattern to what you have done here, but that code was written a long time ago! In more recent times we have tried to aim for one call to the function per test. It does results in more boilerplate, and you have to come up for a name for each test, but it is more in-keeping with the idea of what a unit test should be. If a unit test includes multiple individual tests (e.g. having multiple calls to the function) in a single test then, if the first test fails, it won't run the other tests as it has encountered the error so you don't know whether they are working until you fix the first test.
Having a single test in each `def test...` ensures that they all run, and you get a complete of what is working in a single run of the tests.

So could you split out these test in this file so that each test function only has a single call to the function it is testing?

lbdreyer (changes requested):
Thanks for working on this @schlunma, and thank you for updating it to work with the replacement of pyke! This is a great addition!

I have a few comments. Once those are addressed we can hopefully get this in in time for Iris 3.1. We are hoping to cut the release candidate at the start of next week. If you think that may be too little time, please let me know and we can try to sort something out!

schlunma:
Thanks @lbdreyer for your review. I tried to address all your comments; let me know if you need anything else!

lbdreyer (approved):
Brilliant! Thanks @schlunma for responding so promptly!

PR #4053, opened 2021-03-05
Better calculations for MeshCoord lazy/real points and bounds. [AVD-1590]

pp-mo:
### Description
Functional lazy calculations for lazy points+bounds of MeshCoord

These calcs contain snapshots of the data content of the relevant Mesh components when the MeshCoord is created, but are not able to dynamically track Mesh contents, as originally envisaged.
However they should work with either real or lazy data content.

---

### DIFF
--- a/lib/iris/experimental/ugrid.py
+++ b/lib/iris/experimental/ugrid.py
@@ -2933,6 +2933,8 @@ class MeshCoord(AuxCoord):
             # Bounds are calculated from a connectivity and the node points.
             # Data can be real or lazy, so operations must work in Dask, too.
             indices = bounds_connectivity.core_indices()
+            # Normalise indices dimension order to [faces/edges, bounds]
+            indices = bounds_connectivity.indices_by_src(indices)
             node_points = node_coord.core_points()
             lazy = hasattr(indices, "compute") or hasattr(
                 node_points, "compute"
--- a/lib/iris/experimental/ugrid.py
+++ b/lib/iris/experimental/ugrid.py
@@ -2935,6 +2935,9 @@ class MeshCoord(AuxCoord):
             indices = bounds_connectivity.core_indices()
             # Normalise indices dimension order to [faces/edges, bounds]
             indices = bounds_connectivity.indices_by_src(indices)
+            # Normalise the start index
+            indices = indices - bounds_connectivity.start_index
+
             node_points = node_coord.core_points()
             lazy = hasattr(indices, "compute") or hasattr(
                 node_points, "compute"
@@ -2945,14 +2948,17 @@ class MeshCoord(AuxCoord):
             # must flatten it and restore the shape later.
             flat_inds = indices.flatten()
             # NOTE: the connectivity array can have masked points, but we can't
-            # effectively index with those.  So we must post-mask the results.
-            # We also mask any places where the index is out of range.
-            missing_inds = array.logical_or(
-                flat_inds < 0,
-                flat_inds >= n_nodes,
-                array.ma.getmaskarray(flat_inds),
+            # effectively index with those.  So use a non-masked index array
+            # with "safe" index values, and post-mask the results.
+            flat_inds_mask = array.ma.getmaskarray(flat_inds)
+            flat_inds_nomask = array.ma.filled(flat_inds, 0)
+            # Note: *also* mask any places where the index is out of range.
+            bad_inds = array.logical_or(
+                flat_inds_nomask < 0, flat_inds_nomask >= n_nodes
             )
-            flat_inds_safe = array.where(missing_inds, 0, flat_inds)
+            # NB "OR" of 3 arrays in 2 steps, as Dask does not provide 'reduce'.
+            missing_inds = array.logical_or(flat_inds_mask, bad_inds)
+            flat_inds_safe = array.where(missing_inds, 0, flat_inds_nomask)
             # Here's the core indexing operation.
             # The comma applies all inds-array values to the *first* dimension.
             bounds = node_points[
--- a/lib/iris/experimental/ugrid.py
+++ b/lib/iris/experimental/ugrid.py
@@ -2950,14 +2950,11 @@ class MeshCoord(AuxCoord):
             # NOTE: the connectivity array can have masked points, but we can't
             # effectively index with those.  So use a non-masked index array
             # with "safe" index values, and post-mask the results.
-            flat_inds_mask = al.ma.getmaskarray(flat_inds)
-            flat_inds_nomask = al.ma.filled(flat_inds, 0)
+            flat_inds_nomask = al.ma.filled(flat_inds, -1)
             # Note: *also* mask any places where the index is out of range.
-            bad_inds = al.logical_or(
-                flat_inds_nomask < 0, flat_inds_nomask >= n_nodes
+            missing_inds = (flat_inds_nomask < 0) | (
+                flat_inds_nomask >= n_nodes
             )
-            # NB "OR" of 3 arrays in 2 steps, as Dask does not provide 'reduce'.
-            missing_inds = al.logical_or(flat_inds_mask, bad_inds)
             flat_inds_safe = al.where(missing_inds, 0, flat_inds_nomask)
             # Here's the core indexing operation.
             # The comma applies all inds-array values to the *first* dimension.
--- a/lib/iris/experimental/ugrid.py
+++ b/lib/iris/experimental/ugrid.py
@@ -2903,7 +2903,7 @@ class MeshCoord(AuxCoord):
         Mesh, according to the location and axis.
 
         Returns:
-        * points, bounds (array or None)
+        * points, bounds (array or None):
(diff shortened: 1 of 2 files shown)


lib/iris/experimental/ugrid.py
```diff
@@ -2900,37 +2905,66 @@ def _construct_access_arrays(self):
...
+            bounds_connectivity = mesh.face_node_connectivity
+
+        # The points output is the points of the relevant element-type coord.
+        points = points_coord.core_points()
+        if bounds_connectivity is None:
+            bounds = None
+        else:
+            # Bounds are calculated from a connectivity and the node points.
+            # Data can be real or lazy, so operations must work in Dask, too.
+            indices = bounds_connectivity.core_indices()
```
trexfeathers, line 2941:
```suggestion
            indices = bounds_connectivity.indices_by_src(bounds_connectivity.core_indices())
```

This functionality in `Connectivity` will allow us to handle transposed indices if present (`src_dim=1`), and without needing to modify the `Connectivity` instance itself, so it won't jeopardise future round-tripping.

pp-mo: :+1: Dead right!
And it proves that `indices_by_src` really was a useful thing to provide.

pp-mo: Actually, no I don't think we can use this because it is explicitly realising.
So, I **_think_** `connectivity.indices_by_src(self)` [--> self._indices](https://github.com/SciTools/iris/blob/be62cdfccd54e992ee8240a37ddea79c37772d6e/lib/iris/experimental/ugrid.py#L401), 
which then [--> self._values](https://github.com/SciTools/iris/blob/be62cdfccd54e992ee8240a37ddea79c37772d6e/lib/iris/experimental/ugrid.py#L380), 
[--> self._dm_data](https://github.com/SciTools/iris/blob/be62cdfccd54e992ee8240a37ddea79c37772d6e/lib/iris/coords.py#L228)
which [explicitly realises](https://github.com/SciTools/iris/blob/be62cdfccd54e992ee8240a37ddea79c37772d6e/lib/iris/_data_manager.py#L193-L204).

So, we really needed a `core_indices_by_src` or `lazy_indices_by_src` ?

pp-mo: I've now added this, but put in a separate line of comment with a comment.

pp-mo: P.S. now adding a test for this ...

pp-mo:
***Note:*** @bjlittle  this has prompted me to add testing for start-index = 1 and src_dim = 1 also.
Doing that now..

pp-mo:
Ok, I've now ensured it can handle non-standard connectivity options (dim-order and start-index), plus testing with edge location.
There was actually a bug in the existing masked-array processing :  `np.logical_or(a, b, c)` does **_not_** do what you might expect !

pp-mo:
OK, that look oks on the testing, 

I will now rebase + modify to adopt #4057
(i.e. get rid of `one_mesh_coord`)

Back shortly @bjlittle ...

pp-mo:
Ok here it is, rebased + I hope working.
Please re-check @bjlittle

pp-mo:
Latest change, responding to https://github.com/SciTools/iris/pull/4054#discussion_r589412705

Done now, promise !

lib/iris/experimental/ugrid.py
```diff
@@ -2903,31 +2903,70 @@ def _construct_access_arrays(self):
...
+            )
+            al = da if lazy else np
+            # NOTE: Dask cannot index with a multidimensional array, so we
+            # must flatten it and restore the shape later.
+            flat_inds = indices.flatten()
+            # NOTE: the connectivity array can have masked points, but we can't
+            # effectively index with those.  So use a non-masked index array
+            # with "safe" index values, and post-mask the results.
+            flat_inds_mask = al.ma.getmaskarray(flat_inds)
+            flat_inds_nomask = al.ma.filled(flat_inds, 0)
```
pp-mo, line 2954: hi @bjlittle
Don't want to interrupt again if you're already looking into this,
but TBH I think this can still be improved.

I previously was tripped up + made it more complicated because I found "True OR masked --> masked".

So I **_think_** we can have here :
```
flat_inds_nomask = al.ma.filled(flat_inds, -1)
missing_inds = (flat_inds_nomask < 0) | (flat_inds_nomask >= n_nodes)
flat_inds_safe = al.where(missing_inds, 0, flat_inds_nomask)
```

So, 
 (1) use "filled" to combine mask with 'invalid' points, and
 (2) "|" operator instead of 'logical_or'.

There is also scope here for a bit more testing : I haven't checked the handling of non-masked out-of-range index values.
Let me know if you're up for this further change.

pp-mo: >  this further change.

If you have the appetite, I have some additional including testing ready to add on

pp-mo:
Thanks @bjlittle 
Finished fiddling : all yours ! :crossed_fingers: :four_leaf_clover:

lib/iris/tests/unit/experimental/ugrid/test_MeshCoord.py
```diff
@@ -431,5 +425,274 @@ def test_basic(self):
...
+            n_missing_expected += facenodes_changes["n_extra_bad_points"]
+        self.assertEqual(np.count_nonzero(expected.mask), n_missing_expected)
+        # Check results match, *including* location of masked points.
+        self.assertMaskedArrayAlmostEqual(result, expected)
+
+    def test_points_values(self):
+        """Basic points content check, on real data."""
+        meshcoord = self.meshcoord
+        self.assertFalse(meshcoord.has_lazy_points())
+        self.assertFalse(meshcoord.has_lazy_points())
```
bjlittle, line 594: @pp-mo Shouldn't this be `has_lazy_bounds()`?

pp-mo: :+1:

pp-mo: Ok, will fix.

lib/iris/tests/unit/experimental/ugrid/test_MeshCoord.py
```diff
@@ -431,5 +425,274 @@ def test_basic(self):
...
+        edge_nodes_array = np.array([[0, 2], [1, 3], [1, 4], [3, 7]])
+        # Connectivity uses *masked* for missing points.
+        edge_nodes_array = np.ma.masked_less(edge_nodes_array, 0)
+
+        n_faces = face_nodes_array.shape[0]
+        n_edges = edge_nodes_array.shape[0]
+        n_nodes = int(face_nodes_array.max() + 1)
+        face_xs = 500.0 + np.arange(n_faces)
+        edge_xs = 600.0 + np.arange(n_edges)
+        node_xs = 100.0 + np.arange(n_nodes)
```
bjlittle, line 482: @pp-mo You could make these magic numbers class attributes so that they can be referenced later in the tests e.g., 
```python
Test_MeshCoord__dataviews(...):
    FACE_OFFSET = 500.0
    EDGE_OFFSET = 600.0
    NODE_OFFSET = 100.0
```
Then use them as needed e.g., 
```python
node_xs = self.NODE_OFFSET + np.arange(n_nodes)
```
But more importantly, later is the actual tests i.e., see `_check_expected_bounds_values`, `test_edge_bounds` etc

lib/iris/tests/unit/experimental/ugrid/test_MeshCoord.py
```diff
@@ -431,5 +425,274 @@ def test_basic(self):
...
+        self.assertTrue(meshcoord.has_lazy_points())
+        self.assertTrue(meshcoord.has_lazy_bounds())
+        meshcoord.points
+        meshcoord.bounds
+        self.assertFalse(meshcoord.has_lazy_points())
+        self.assertFalse(meshcoord.has_lazy_bounds())
+
+        # Check all the source coords are still lazy.
+        for coord in fetch_sources_from_mesh():
+            self.assertTrue(hasattr(coord._core_values(), "compute"))
```
bjlittle, line 670: @pp-mo Use `is_lazy_data(coord._core_values())` from `iris._lazy_data`?

Just a suggestion, happy either way :wink:

trexfeathers: +1

lib/iris/tests/unit/experimental/ugrid/test_MeshCoord.py
pp-mo: Done this.
I note that the face/edge values are not actually used.
Don't think that matters.

pp-mo:
OK, thanks @bjlittle .
I think that deals with the above -- how does it look ?

bjlittle:
@pp-mo Once `cirrus-ci` is happy, then so am I :+1:

PR #4054, opened 2021-03-05
Meshcoord printout [AVD-1592]

pp-mo:
### Description
Added string representations for MeshCoord

Waiting until #4053 is in :  Which this is based on.
(so it won't merge into the target branch, as yet)

I have made the str/repr versions slightly different, following discussion with @bjlittle 
I'm still a little uncertain that the results are the most useful, though:
  * if you enter a MeshCoord at the command line, you get "repr", which shows a mesh as `<Mesh object at 0x{address}>`, 
     * whereas the "str" form attempts a more human-readable account of the Mesh, as `Mesh('name')`
  * This  corresponds with the "standard" Python usage ...
  * ... **_but_**,  for a cube, the "repr" version is the *short form*, whereas print(cube) gives full details.
    * .. so making repr the "more precise and less human" is **_against_** what we do for cubes.  

---

### DIFF
--- a/lib/iris/experimental/ugrid.py
+++ b/lib/iris/experimental/ugrid.py
@@ -2772,7 +2772,6 @@ class MeshCoord(AuxCoord):
             raise ValueError(msg)
 
         # Get the 'coord identity' metadata from the relevant node-coordinate.
-        # N.B. mesh.coord returns a dict
         node_coord = self.mesh.coord(include_nodes=True, axis=self.axis)
         # Call parent constructor to handle the common constructor args.
         super().__init__(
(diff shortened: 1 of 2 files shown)


bjlittle:
>     * .. so making repr the "more precise and less human" is **_against_** what we do for cubes.

@pp-mo Point taken. 

To be honest we should rationalise and align all our `__repr__` and `__str__` usage throughout `iris`... it's well overdue and the state of affairs that we find ouselves in is a fair reflection on the unopinionated, organic way in which `iris` has developed over the years

lib/iris/experimental/ugrid.py
```diff
@@ -2894,43 +2899,122 @@ def __eq__(self, other):
...
+        if bounds_connectivity is None:
+            bounds = None
+        else:
+            # Bounds are calculated from a connectivity and the node points.
+            # Data can be real or lazy, so operations must work in Dask, too.
+            indices = bounds_connectivity.core_indices()
+            node_points = node_coord.core_points()
+            lazy = hasattr(indices, "compute") or hasattr(
+                node_points, "compute"
+            )
```
bjlittle, line 2995: @pp-mo `iris._lazy_data.is_lazy_data` is a handy convenience function... if you were to buy into that, then this would give:
```python
lazy = is_lazy_data(indices) or is_lazy_data(node_points)
```
which reads quite nicely

lib/iris/experimental/ugrid.py
```diff
@@ -2894,43 +2899,122 @@ def __eq__(self, other):
...
+            array = da if lazy else np
```
bjlittle, line 2996: @pp-mo It's easy for the readers eye to sweep over `array`... could I be really painful and suggest a rename?

Say something like `handler`?

lib/iris/experimental/ugrid.py
```diff
@@ -2894,43 +2899,122 @@ def __eq__(self, other):
...
+            mesh_string = f"<Mesh object at {hex(mesh_id)}>"
+        result = (
+            f"mesh={mesh_string}"
+            f", location={self.location!r}"
+            f", axis={self.axis!r}"
+        )
+        # Add 'other' metadata that is drawn from the underlying node-coord.
+        # But put these *afterward*, unlike other similar classes.
+        for item in ("standard_name", "units", "long_name", "attributes"):
+            # NOTE: order of these matches Coord.summary, but omit vor_name.
```
bjlittle, line 2934: @pp-mo Type `vor_name` -> `var_name`

bjlittle (changes requested):
@pp-mo Some suggested changes... nothing major. See what you think.

Otherwise LGTM :+1:

lib/iris/experimental/ugrid.py
rcomer: [Elsewhere](https://github.com/SciTools/iris/blob/a9ae72fa3c420da408fb4f69bba4035c2c43f769/lib/iris/coords.py#L2000) we have `al` (array library).

lib/iris/experimental/ugrid.py
```diff
@@ -2683,6 +2683,11 @@ def face_node(self):
         return self._members["face_node_connectivity"]
 
 
+def _one_meshcoord(meshcoords_dict):
+    (one,) = meshcoords_dict.values()
+    return one
+
+
```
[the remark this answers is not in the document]
pp-mo: :+1: !

lib/iris/experimental/ugrid.py
pp-mo: :+1:

lib/iris/experimental/ugrid.py
pp-mo: ok

lib/iris/experimental/ugrid.py
pp-mo: Ok, but could you move this comment across to #4053 where it probably belongs ?

pp-mo:
Rebased onto #4053 (effectively).
But, this may need doing again.

In the meantime, attending to the existing comments above (though now outdated).

lib/iris/experimental/ugrid.py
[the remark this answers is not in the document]
pp-mo: done!

lib/iris/experimental/ugrid.py
pp-mo: done in #4053 (now here)

lib/iris/experimental/ugrid.py
pp-mo: done in #4053

pp-mo:
I think that's it for the outstanding review issues.
I'm taking this out of Draft, but it still won't look nice here (or merge into the target) until #4053 is in.

pp-mo:
Rebased again ( :sweat: )
Ready to go @bjlittle !

bjlittle (approved):

PR #4057, opened 2021-03-08
align mesh api with cube

bjlittle:
### Description
A minor refactor of the `Mesh.coord`, `Mesh.coords`, `Mesh.connectivity`, and `Mesh.connectivities` API to align with `Cube.coord` and `Cube.coords`, in that a single instance or a `list` containing zero or more instances is returned, rather than a `dict`. 

---

### DIFF
--- a/lib/iris/experimental/ugrid.py
+++ b/lib/iris/experimental/ugrid.py
@@ -1410,7 +1410,7 @@ class Mesh(CFVariableMixin):
             that matched the given criteria.
 
         """
-        return self._connectivity_manager.filters(
+        result = self._connectivity_manager.filters(
             item=item,
             standard_name=standard_name,
             long_name=long_name,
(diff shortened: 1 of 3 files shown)


lib/iris/experimental/ugrid.py
```diff
@@ -2769,8 +2773,7 @@ def __init__(
 
         # Get the 'coord identity' metadata from the relevant node-coordinate.
         # N.B. mesh.coord returns a dict
-        node_coords = self.mesh.coord(include_nodes=True, axis=self.axis)
-        (node_coord,) = list(node_coords.values())
+        node_coord = self.mesh.coord(include_nodes=True, axis=self.axis)
```
bjlittle, line 2776: @pp-mo Oh blessed simplicity... that is way, way, way better IMHO

pp-mo: nice ! :+1:

PR #4059, opened 2021-03-10
Intersection bounds fix (replacement PR)

### LINKED ISSUE #3391: cube.intersection out of bounds point

rcomer:
### Description
Replacement for #4046, targetting v3.0.x

Fixes #3391 I think.  This addresses the very specific case where you request the entire 360 degrees, and one of the bounds is at the maximum/minimum threshold.  Taking @valeriupredoi's example from #3391, you would still get that negative longitude point if you did something like `cube.intersection(longitude=(0, 50))`.  I'm not sure that that makes sense but, looking at the tests, it appears to have been intentional.

I haven't added any extra comment to explain my logic, but I think it's covered by the existing:
https://github.com/SciTools/iris/blob/a9ae72fa3c420da408fb4f69bba4035c2c43f769/lib/iris/cube.py#L3178-L3180

---

### DIFF
--- a/lib/iris/cube.py
+++ b/lib/iris/cube.py
@@ -3188,7 +3188,10 @@ bound=(1994-12-01 00:00:00, 1998-12-01 00:00:00)
                 indices = inside_indices[split_cell_indices]
                 cells = bounds[indices]
                 if maximum % modulus not in cells:
-                    # Recalculate the extended minimum.
+                    # Recalculate the extended minimum only if the output bounds
+                    # do not span the requested (minumum, maximum) range.  If
+                    # they do span that range, this adjustment would give unexpected
+                    # results (see #3391).
                     cells_delta = np.diff(coord.bounds[indices])
 
                     # Watch out for ascending/descending bounds.
--- a/lib/iris/cube.py
+++ b/lib/iris/cube.py
@@ -3185,18 +3185,19 @@ bound=(1994-12-01 00:00:00, 1998-12-01 00:00:00)
                     pre_wrap_delta != post_wrap_delta
                 )
 
-                # Recalculate the extended minimum.
                 indices = inside_indices[split_cell_indices]
                 cells = bounds[indices]
-                cells_delta = np.diff(coord.bounds[indices])
+                if maximum % modulus not in cells:
+                    # Recalculate the extended minimum.
+                    cells_delta = np.diff(coord.bounds[indices])
 
-                # Watch out for ascending/descending bounds
-                if cells_delta[0, 0] > 0:
-                    cells[:, 0] = cells[:, 1] - cells_delta[:, 0]
-                    minimum = np.min(cells[:, 0])
-                else:
-                    cells[:, 1] = cells[:, 0] + cells_delta[:, 0]
-                    minimum = np.min(cells[:, 1])
+                    # Watch out for ascending/descending bounds.
+                    if cells_delta[0, 0] > 0:
+                        cells[:, 0] = cells[:, 1] - cells_delta[:, 0]
+                        minimum = np.min(cells[:, 0])
+                    else:
+                        cells[:, 1] = cells[:, 0] + cells_delta[:, 0]
+                        minimum = np.min(cells[:, 1])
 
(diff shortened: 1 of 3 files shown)


lib/iris/cube.py
```diff
@@ -3130,18 +3130,19 @@ def _intersect_modulus(
...
-                    minimum = np.min(cells[:, 0])
-                else:
-                    cells[:, 1] = cells[:, 0] + cells_delta[:, 0]
-                    minimum = np.min(cells[:, 1])
+                if maximum % modulus not in cells:
+                    # Recalculate the extended minimum.
+                    cells_delta = np.diff(coord.bounds[indices])
+
+                    # Watch out for ascending/descending bounds.
+                    if cells_delta[0, 0] > 0:
```
jamesp, line 3140: Previously, `minimum` would have been set by either branch of this `if` statement on every call.  Now that it is contained within `if maximum % modulus not in cells` it may not be.  Should there be an `else` to the `if` on L3135?

lib/iris/cube.py
```diff
@@ -3130,18 +3130,19 @@ def _intersect_modulus(
...
-                cells_delta = np.diff(coord.bounds[indices])
-
-                # Watch out for ascending/descending bounds
-                if cells_delta[0, 0] > 0:
-                    cells[:, 0] = cells[:, 1] - cells_delta[:, 0]
-                    minimum = np.min(cells[:, 0])
-                else:
-                    cells[:, 1] = cells[:, 0] + cells_delta[:, 0]
-                    minimum = np.min(cells[:, 1])
+                if maximum % modulus not in cells:
```
jamesp, line 3135: Perhaps a comment here, even referencing the exact problem that this code is aiming to solve.

jamesp (commented):
Looks great, the new test passes with the new `lib/iris/cube.py`, and fail without it 👍 

One question inline, due to my unfamiliarity with the whole codebase.  If it's all good, LGTM

lib/iris/cube.py
rcomer: It's only been two months but I had completely forgotten the logic here!  I think this is fine though: we don't explicitly set `minimum` here, so it is just left to whatever `minimum` was originally passed into the method.  Note that because of the previous logic

`maximum - modulus` <= `minimum` (line 3081)
`cells` >= `minimum` (line 3098)

So the only way that `maximum - modulus in cells` will be `True` is if `maximum - modulus` == `minimum` and `minimum` is in `cells`.   I.e. the very specific case described in #3391.  In that case, it is expected that the bounds on the output coordinate will precisely span the requested (minimum, maximum) range.

Definitely needs a comment though!

rcomer:
Now my tests are failing because this is targetting the `v3.0.x` branch, so doesn't have any of the CI goodness that's just been added to `master`. 😔

jamesp:
Oh no! well you've only added comments so functionally the same as before.  

We should probably decide if we want to cherry pick some of the CI changes into the v3.0.x branch @bjlittle @trexfeathers

valeriupredoi:
> Now my tests are failing because this is targetting the `v3.0.x` branch, so doesn't have any of the CI goodness that's just been added to `master`. pensive

merge `master` in here?

rcomer:
Rebased.  Main tests pass locally under py3.6 and py3.7  py3.8 tests look to be running OK in Cirrus.

py3.6
```
Ran 5036 tests in 307.194s

OK (SKIP=22)
```

py3.7
```
Ran 5017 tests in 513.501s

OK (SKIP=22)
```

py3.7 really seems to be misplacing tests....

rcomer:
Doc link check and gallery test are as per #4130.  The gallery test still passes locally for me under py3.8.

rcomer:
@jamesp I think this is now the best I can do with this one.

jamesp:
Happy with this, LGTM

PR #4063, opened 2021-03-12
Cube.coords mesh_coords key [AVD-1662]

pp-mo:
### Description
Adding a 'mesh_coords' keyword to Cube.coords.
Behaves mostly like the existing `dim_coords` keyword.

(This is a simple preliminary to adding the more complex cube api to come :
  * `cube.mesh` and `.location`
  * `cube.__eq__` should compare mesh and location
  * various checks that any MeshCoord-s in a cube are always compatible.
)

---

### DIFF
--- a/lib/iris/experimental/ugrid.py
+++ b/lib/iris/experimental/ugrid.py
@@ -2896,6 +2896,10 @@ class MeshCoord(AuxCoord):
 
         return eq
 
+    # Exactly as for Coord.__hash__ :  See there for why.
+    def __hash__(self):
+        return hash(id(self))
+
     def _string_summary(self, repr_style):
         # Note: bypass the immediate parent here, which is Coord, because we
         # have no interest in reporting coord_system or climatological, or in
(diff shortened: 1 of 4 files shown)


lib/iris/experimental/ugrid.py
```diff
@@ -2927,7 +2931,13 @@ def _string_summary(self, repr_style):
         )
         # Add 'other' metadata that is drawn from the underlying node-coord.
         # But put these *afterward*, unlike other similar classes.
-        for item in ("standard_name", "units", "long_name", "attributes"):
+        for item in (
+            "shape",
```
bjlittle, line 2935: @pp-mo Snuck in `shape` here... no objections, just curious to the reasoning

bjlittle, line 2935: @pp-mo Snuck in `shape` here... no objections, just curious as to the reasoning

bjlittle (approved):
@pp-mo Lovely how easily this integrates into `iris` :smile: 

Awesome, thanks! :+1: 

Just a super minor question to answer (which isn't a block to this PR merging)

lib/iris/experimental/ugrid.py
pp-mo: Well it seems clear that when printing these mesh information components, we don't want to realise or print out contents.
But when debugging I saw the printouts, it seemed odd you couldn't even see the shape., unlike dask arrays which show the shape + dtype.  
So I added it in (but not dtype).  

Since then I realised that you + @trexfeathers had already agreed a printout format for Connectivity, which does not include shape.  
So I now think that we should probably at least make those 2 the same :  
**What do you think + which way do you reckon we should we jump ?**

bjlittle: @pp-mo Awesome, thanks.

I was going to re-haul how coordinates in general do `__str__` and `__repr__` in order to keep everything lazy.

I think there is real tangible benefit there, but outside the scope of this PR

PR #4064, opened 2021-03-12
Support for cube mesh api [AVD-1663]

pp-mo:
### Description
Adding additonal mesh-specific Cube support:
 * add `cube.mesh()` and `cube.location()` access methods
 * make `cube.__eq__` compare mesh + location
 * adjust all cube creation + coord management to ensure any/all MeshCoords are always inter-compatible

Reference #4063.

---

### DIFF
--- a/lib/iris/cube.py
+++ b/lib/iris/cube.py
@@ -1919,6 +1919,32 @@ bound=(1994-12-01 00:00:00, 1998-12-01 00:00:00)
 
         return result
 
+    def mesh(self):
+        """
+        Return the unstructured :class:`~iris.experimental.ugrid.Mesh`
+        associated with the cube, or None if there is none.
+
+        """
+        result = self.coords(mesh_coords=True)
+        if len(result) == 0:
+            result = None
+        else:
+            result = result[0].mesh
+        return result
+
+    def location(self):
+        """
+        Return the mesh location of the cube, if the cube has an unstructured
+        :class:`~iris.experimental.ugrid.Mesh`, or None if there is none.
+
+        """
+        result = self.coords(mesh_coords=True)
+        if len(result) == 0:
+            result = None
+        else:
+            result = result[0].location
+        return result
+
     def cell_measures(self, name_or_cell_measure=None):
         """
         Return a list of cell measures in this cube fitting the given criteria.
--- a/lib/iris/cube.py
+++ b/lib/iris/cube.py
@@ -1134,7 +1134,7 @@ bound=(1994-12-01 00:00:00, 1998-12-01 00:00:00)
     def _add_unique_aux_coord(self, coord, data_dims):
         data_dims = self._check_multi_dim_metadata(coord, data_dims)
         if hasattr(coord, "mesh"):
-            mesh = self.mesh()
+            mesh = self.mesh
             if mesh:
                 msg = (
                     "{item} of Meshcoord {coord!r} is "
@@ -1150,7 +1150,7 @@ bound=(1994-12-01 00:00:00, 1998-12-01 00:00:00)
                             ownval=mesh,
                         )
                     )
-                location = self.location()
+                location = self.location
                 if coord.location != location:
                     raise ValueError(
                         msg.format(
@@ -1957,7 +1957,8 @@ bound=(1994-12-01 00:00:00, 1998-12-01 00:00:00)
 
         return result
 
-    def _a_meshcoord(self):
+    def _any_meshcoord(self):
+        """Return a MeshCoord if there are any, else None."""
         mesh_coords = self.coords(mesh_coords=True)
         if mesh_coords:
             result = mesh_coords[0]
@@ -1965,6 +1966,7 @@ bound=(1994-12-01 00:00:00, 1998-12-01 00:00:00)
             result = None
         return result
 
+    @property
     def mesh(self):
         """
         Return the unstructured :class:`~iris.experimental.ugrid.Mesh`
@@ -1980,11 +1982,12 @@ bound=(1994-12-01 00:00:00, 1998-12-01 00:00:00)
             or ``None``.
 
         """
-        result = self._a_meshcoord()
+        result = self._any_meshcoord()
         if result is not None:
             result = result.mesh
         return result
 
+    @property
     def location(self):
         """
         Return the mesh "location" of the cube data, if the cube has any
@@ -2000,7 +2003,7 @@ bound=(1994-12-01 00:00:00, 1998-12-01 00:00:00)
             or ``None``.
 
         """
-        result = self._a_meshcoord()
+        result = self._any_meshcoord()
         if result is not None:
             result = result.location
         return result
@@ -2019,7 +2022,7 @@ bound=(1994-12-01 00:00:00, 1998-12-01 00:00:00)
             or ``None``.
 
         """
-        result = self._a_meshcoord()
+        result = self._any_meshcoord()
         if result is not None:
             (result,) = self.coord_dims(result)  # result is a 1-tuple
         return result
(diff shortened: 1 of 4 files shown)


lib/iris/cube.py
```diff
@@ -1919,6 +1957,73 @@ def coord_system(self, spec=None):
 
         return result
 
+    def _a_meshcoord(self):
```
bjlittle, line 1960: @pp-mo Could we rename this just to be just `_meshcoord`? I think that's fine. Or even `_any_meshcord`?

Also add a slim doc-string for developers i.e., `returns any MeshCoord that's available`

lib/iris/cube.py
```diff
@@ -1919,6 +1957,73 @@ def coord_system(self, spec=None):
...
+        mesh_coords = self.coords(mesh_coords=True)
+        if mesh_coords:
+            result = mesh_coords[0]
+        else:
+            result = None
+        return result
+
+    def mesh(self):
```
bjlittle, line 1968: @pp-mo We're never (never say never) going to pass parameters to this, so could we err on the side of starting this methods life as a read-only property i.e.,
```python
@property
def mesh(self):
    ...
```
This makes the usage a lot cleaner IMHO i.e.,
```python
cube.mesh
```
rather than
```python
cube.mesh()
```
I realise that `Cube.coord_system` is indeed a method, but that's a slightly different beast... plus it does accept `kwargs`.

lib/iris/cube.py
```diff
@@ -1919,6 +1957,73 @@ def coord_system(self, spec=None):
...
+            (i.e. one of 'face' / 'edge' / 'node'),
+            or ``None``.
+
+        """
+        result = self._a_meshcoord()
+        if result is not None:
+            result = result.location
+        return result
+
+    def mesh_dim(self):
```
bjlittle, line 2008: @pp-mo I think this call signature is in keeping with `Cube.coord_dims` ... and ultimately it may be confusing to make this a `@property`. So this is fine by me :+1:

lib/iris/tests/unit/cube/test_Cube.py
```diff
@@ -1960,6 +1960,52 @@ def test__lazy(self):
...
+    self.dimco_mesh = mesh_dimco
+    if not nomesh:
+        self.meshco_x = meshx
+        self.meshco_y = meshy
+    self.auxco_x = auxco_x
+    self.allcoords = mesh_coords + [zco, mesh_dimco, auxco_x]
+    self.cube = cube
+    self.mesh = mesh
+
+
```
bjlittle, line 2008: @pp-mo If we're serious about supporting unstructured meshes and cube-spheres... and I think we are... would you consider us acknowledge this by curating some `iris.tests.stock` cubes dedicated to testing `iris` with synthetic based cubes such as this?

I seems to me that this kind of cube generation has value beyond this specific pull-request. At least we could start that commitment here.

So if you buy into that point-of-view, fancy migrating this into a separate stock cube namespace, say `iris.tests.stock._mesh`, and then update `iris.tests.stock` to `from ._mesh import <whatever>` ?

lib/iris/tests/unit/cube/test_Cube.py
```diff
@@ -2049,6 +2069,301 @@ def test_coords__nodimcoords__meshcoords(self):
...
+    """
+    Test that creation with mesh-coords functions, and prevents a cube having
+    incompatible mesh-coords.
+
+    """
+
+    def setUp(self):
+        # Create a standard test mesh and other useful components.
+        mesh = create_test_mesh()
+        meshco = create_test_meshcoord(mesh=mesh)
```
bjlittle, line 2148: @pp-mo Are these also ingredients that can be mixed into the stock?

lib/iris/tests/unit/cube/test_Cube.py
```diff
@@ -2049,6 +2069,301 @@ def test_coords__nodimcoords__meshcoords(self):
...
+        _add_test_meshcube(self, mesh=self.mesh)
+        cube2 = self.cube
+        self.assertEqual(cube1, cube2)
+
+    def test_new_mesh_different(self):
+        cube1 = self.cube
+        # re-create an identical cube, using the same mesh.
+        _add_test_meshcube(self)
+        cube2 = self.cube
+        self.assertNotEqual(cube1, cube2)
```
bjlittle, line 2364: @pp-mo Do we need a test here where the the cubes are the same apart from the meshcoords which reference different meshes. So effectively, cube equality will fail due to the respective meshcoords referencing different meshes?

bjlittle (changes requested):
@pp-mo It's almost there... nice one! :+1: 

Just a few outstanding review comments to address. Hope you don't mind :smile:

lib/iris/tests/unit/cube/test_Cube.py
trexfeathers: We are already committed to synthetic data generation as part of MPerf testing 👍

pp-mo (commented):
?? I seem to be blocked here by something + it thinks I am writing a review.
Submit/completing that to see if it will sort itself out...

lib/iris/tests/unit/cube/test_Cube.py
pp-mo: > synthetic data generation
We are, but I think that all goes through *files* at this point ?
So we could maybe still use a means of generating stock *cubes*.

I will look + see if we can maybe integrate the two, though : Perhaps the same code, with the same options, can be used to generate either files or cubes.
)Obviously, save capability is still some way off)

lib/iris/cube.py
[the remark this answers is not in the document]
pp-mo: Ok :+1:

lib/iris/cube.py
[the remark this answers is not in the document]
pp-mo: Well, my view on this was that it isn't a simple pragmatic choice whether to use `cube.mesh` or `cube.mesh()` as they "should" indicate different kinds of things.
I.E. in my head, an instance property is a simple piece of the object state, which you can expect to be a simple get/set., whereas methods are used for derived stuff and calculations.

So in this case, a MeshCoord has mesh + location + axis properties, because they are the core parts of its state/identity, whereas a Cube should have mesh() location() etc because they are calculated values, not fundamental to the object itself.

Obviously, all that isn't a 'real' distinction in Python, unlike other O-O languages.
Do you have any sympathy with this idea ?

lib/iris/tests/unit/cube/test_Cube.py
pp-mo: Well, in effect that *is* exactly what this testcase is doing. 
It's a bit hidden because we make one simple `_add_test_meshcube(self)` call in setUp -- so that is what makes `cube1`. Then we do the exact same thing again here, producing a `cube2` which is basically the same, except that it cannot be "==" because it is based on a different (but identically constructed) mesh object.

lib/iris/cube.py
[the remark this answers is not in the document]
bjlittle: @pp-mo Regardless, what your effectively saying is that you're happy for force users to do this?
```python
cube.location() == coord.location and cube.mesh() == coord.mesh
``` 
Which makes me fundamentally sad to my core; sorry but it's true :disappointed: :-1: 

Whereas, what I'm suggesting is this...
```python
cube.location == coord.location and cube.mesh == coord.mesh
```
Yes, please :smile: :+1:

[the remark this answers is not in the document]
pp-mo: Well I guess so, if these things are not the same _kind_ of thing for the cube and the coord.
Which is why we have cube.coords() and not cube,coords.   Like dict.values().
But I can see I just won't win this argument.

bjlittle (approved):

PR #4106, opened 2021-04-22
Logging fixes

trexfeathers:
### Description

Improvements to all uses of logging in `experimental.ugrid`, following recent team discoveries.

---

### DIFF
--- a/lib/iris/experimental/ugrid/__init__.py
+++ b/lib/iris/experimental/ugrid/__init__.py
@@ -1229,7 +1229,7 @@ class Mesh(CFVariableMixin):
                     "Not setting face_dimension (inappropriate for "
                     f"topology_dimension={self.topology_dimension} ."
                 )
-                logger.debug(message)
+                logger.debug(message, extra=dict(cls=self.__class__.__name__))
         elif not name or not isinstance(name, str):
             face_dimension = f"Mesh{self.topology_dimension}d_face"
         else:
@@ -3332,7 +3332,11 @@ class CFUGridConnectivityVariable(cf.CFVariable):
                                     f"{name}, referenced by netCDF variable "
                                     f"{nc_var_name}"
                                 )
-                                logger.debug(message)
+                                # TODO: reconsider logging level when we
+                                #  have consistent practice.
+                                logger.warning(
+                                    message, extra=dict(cls=cls.__name__)
+                                )
                         else:
                             # Restrict to non-string type i.e. not a
                             # CFLabelVariable.
@@ -3346,7 +3350,9 @@ class CFUGridConnectivityVariable(cf.CFVariable):
                                     f"as a CF-UGRID connectivity - is a "
                                     f"CF-netCDF label variable."
                                 )
-                                logger.debug(message)
+                                logger.debug(
+                                    message, extra=dict(cls=cls.__name__)
+                                )
 
         return result
 
@@ -3403,7 +3409,11 @@ class CFUGridAuxiliaryCoordinateVariable(cf.CFVariable):
                                         f"referenced by netCDF variable "
                                         f"{nc_var_name}"
                                     )
-                                    logger.debug(message)
+                                    # TODO: reconsider logging level when we
+                                    #  have consistent practice.
+                                    logger.warning(
+                                        message, extra=dict(cls=cls.__name__)
+                                    )
                             else:
                                 # Restrict to non-string type i.e. not a
                                 # CFLabelVariable.
@@ -3420,7 +3430,9 @@ class CFUGridAuxiliaryCoordinateVariable(cf.CFVariable):
                                         f"coordinate - is a CF-netCDF label "
                                         f"variable."
                                     )
-                                    logger.debug(message)
+                                    logger.debug(
+                                        message, extra=dict(cls=cls.__name__)
+                                    )
 
         return result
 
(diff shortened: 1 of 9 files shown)


lib/iris/experimental/ugrid/__init__.py
```diff
@@ -3673,7 +3689,6 @@ def _build_mesh(cf, mesh_var, file_path):
             ", consistent with the attached connectivities."
         )
         # TODO: reconsider logging level when we have consistent practice.
-        # TODO: logger always requires extras['cls'] : can we fix this?
```
trexfeathers, line 3676: `extra` is useful for almost all examples in this module, excepting those in `_build_mesh()`. AFAICT the only way to not require this kwarg, while still offering the benefits when wanted, would be something like this at the top of the module:

```Python
logger._log = functools.partial(logger._log, extra=dict(cls=None))
```

... and that feels pretty hacky, right?

lib/iris/experimental/ugrid/__init__.py
```diff
@@ -3687,7 +3702,6 @@ def _build_mesh(cf, mesh_var, file_path):
                 " -- ignoring this as it is inconsistent."
             )
             # TODO: reconsider logging level when we have consistent practice.
-            # TODO: logger always requires extras['cls'] : can we fix this?
```
trexfeathers, line 3690: https://github.com/SciTools/iris/pull/4106/files#r618543039

pp-mo:
Useful improvements you've made here.
But on the testing side, I couldn't avoid thinking about how to improve this so I went digging and experimenting.
I came up with a possible way here : https://github.com/pp-mo/iris/pull/68/files

Don't want to undermine the work here, but I think the test changes are looking messy.
Would you consider incorporating into this PR the above extension to the `assertLogs` method ??
I do still have some concerns about the completeness of that aproach, though : see https://github.com/pp-mo/iris/pull/68

trexfeathers:
>Don't want to undermine the work here, but I think the test changes are looking messy.

3cf66627856db2214eeeae05ff4a61a349172449  Does this help at all?

>Would you consider incorporating into this PR the above extension to the assertLogs method ??

bfbed13737fe3099aef67864a26cd8dd38b30e96  I can confirm from local testing that this addresses the issue you originally identified.

>I do still have some concerns about the completeness of that aproach, though

I get what you're saying, however this area is new to us and we have addressed all issues we've encountered so far. No problem IMO.

lib/iris/tests/__init__.py
```diff
@@ -607,6 +607,31 @@ def assertWarnsRegexp(self, expected_regexp=""):
         msg = msg.format(expected_regexp)
         self.assertTrue(matches, msg)
 
+    @contextlib.contextmanager
+    def assertLogs(self, logger=None, level=None):
```
pp-mo, line 611:
```suggestion
    def assertLogs(self, logger=None, level=None, msg_regex=None):
```

lib/iris/tests/__init__.py
```diff
@@ -607,6 +607,31 @@ def assertWarnsRegexp(self, expected_regexp=""):
...
+        """
+        An extended version of the usual :meth:`unittest.TestCase.assertLogs`,
+        which also exercises the logger's message formatting.
+
```
pp-mo, line 615:
```suggestion
        Also adds the 'msg_regex' kwarg :  
        If used, check that the result is a single message of the expected level, 
        and that it matches this regex.
```

lib/iris/tests/__init__.py
```diff
@@ -607,6 +607,31 @@ def assertWarnsRegexp(self, expected_regexp=""):
...
+        """
+        # Invoke the standard assertLogs behaviour.
+        assertlogging_context = super().assertLogs(logger, level)
+        with assertlogging_context as watcher:
+            # Run the caller context, as per original method.
+            yield watcher
+        # Check for any formatting errors by running all the formatters.
+        for record in watcher.records:
+            for handler in assertlogging_context.logger.handlers:
+                handler.format(record)
```
pp-mo, line 633:
```suggestion
                handler.format(record)

        # Check message, if requested
        if msg_regex:
            self.asssertEqual(len(watcher.records), 1)
            rec = watcher.records[0]
            self.assertEqual(level, rec.levelname)
            self.assertRegex(rec.msg, msg_regex)
```

pp-mo (commented):
> > Don't want to undermine the work here, but I think the test changes are looking messy.

> 3cf6662 Does this help at all?

I really meant the way we formerly had extra code to 'exercise' the logging before capturing + checking it.
So that is all squared away now 👍 


> > Would you consider incorporating into this PR the above extension to the assertLogs method ??

> bfbed13 I can confirm from local testing that this addresses the issue you originally identified.

Brilliant, thanks ! 💐 


> > I do still have some concerns about the completeness of that aproach, though

> I get what you're saying, however this area is new to us and we have addressed all issues 
we've encountered so far. No problem IMO.

👍 I like your approach 😁 


One thing that does still jar here, for me, is the repetition of the 'check_log` method : we now have 4 copies (!)
I was very close to adding that (i.e. the regex check) into the new-version `assertLogs`, but I felt it was perhaps too much in one.
But what the hell, why not add an optional keyword like 'message_regex' + integrate the  two?
The only snag, IMO, is that it only applies when you're expecting a single logged message.  But naming/documentation can just allow for that.
See my suggestions (only partial changes : you will need to replace all the `check_log`s with `assertLogs` changes)

**How does that way look to you @trexfeathers ?**

trexfeathers:
>How does that way look to you @trexfeathers ?

Very nice idea @pp-mo! 5a997d333175ded17c2f9202f849a6cb2317e6c1 should take care of this - involved a fair amount of refactoring but the result is simpler, a little more akin to `AssertRaisesRegex()`, which is what I was originally after anyway.

pp-mo:
This all seems good to me now.
Thanks for sticking with it @trexfeathers

PR #4104, opened 2021-04-20
Linkcheck update

tkknight:
### Description

Fix linkcheck.  

This should close #4103 

---
[Consult Iris pull request check list]( https://scitools-iris.readthedocs.io/en/latest/developers_guide/contributing_pull_request_checklist.html)

PR #4108, opened 2021-04-27
CI lockfiles

### LINKED ISSUE #4105: Replace conda update / conda cache in cirrus with conda-lock
## 📰 Custom Issue

Current cirrus ci resolves a new environment every run.  Due to computational constraints, sometimes this can take more than an hour and force the job to timeout when running the command `conda env update --prefix=/tmp/cirrus-ci-build/.nox/tests --file=requirements/ci/py36.yml --prune`.  Proposed alternative:

1. Create environment:
    a. if `environment-py36.lock` doesn't exist, resolve an environment with `conda env create`. use `conda-lock` to create an explicit lock file `environment-py36.yml`.
   b. if `environment-py36.lock` does exist, use `conda create -f environment-py36.yml` to avoid resolving a new env.
3. Run tests.
4. Only cache `environment-py36.yml` with our current cache invalidation logic (weekly I think?) rather than the whole conda directory.

This would mean reproducible environments, but not committing an explicit environment to our repo and avoid long resolve and update calls in the CI.

Alternatively, we could resolve and check-in lock files to our repository.  This is standard practice in the javascript world, see https://stackoverflow.com/questions/44206782/do-i-commit-the-package-lock-json-file-created-by-npm-5 for some discussions of the pros and cons of doing so.

jamesp:
### Description

Here's a proof of concept for adding locked dependencies and using them in the nox testing framework.

* Lock files are found in `requirements/ci/nox.lock`.  These are explicit conda environment files which means they do not require resolving, they are the resolved manifestation of the equivalent `requirements/ci/py3X.yml`.  This means that they can be installed using `conda create -n my_new_env --file requirements/ci/nox.lock/py3X-linux-64.lock`.
* `noxfile.py` updated to use these resolved environments and avoid rebuilding / updating until the lockfile changes.  This should dramatically reduce the *second* time the nox test suite runs.  It should probably also greatly reduce the time the first run takes too, as there is no need for conda to resolve or update the environment.
* A github action workflow to generate new lockfiles every Saturday morning, and create a PR with the changed files.  This can then be included/ignored/deleted at our leisure.

There's quite a lot in this, so whoever feels like reviewing, lets go through it together 👍 

This should resolve #4105 

---


jamesp:
You can see an example of the GH Actions run here: https://github.com/jamesp/iris/actions/runs/790009248

And the PR it produces here: https://github.com/jamesp/iris/pull/3

bjlittle:
@jamesp Did you try using `conda-lock` with `mamba` in `miniforge` to revolve in realtime?

Or would that also struggle in `cirrus-ci`?

bjlittle:
We could also rationalise the requirements themselves, as `conda-lock` can resolve multiple `yml` files at once, which is super handy. i.e., the testing jobs don't need the documentation requirements, and I suspect that the minimal testing jobs don't need the optional requirements.

There are options to make this less bloaty, package wise... but maybe that can come later.

jamesp:
No I didn't but I'm sure that would work too and would be a viable alternative to this setup.  But then you are still doing a resolve on every run of the CI.

We first need to decide if lockfiles belong in version control.  If yes, this scheme provides a mechanism for keeping them up to date.  If no, this scheme should be discarded.  we could instead keep the lockfiles cached in CI to reduce the number of resolves

trexfeathers:
>We first need to decide if lockfiles belong in version control. If yes, this scheme provides a mechanism for keeping them up to date. If no, this scheme should be discarded. we could instead keep the lockfiles cached in CI to reduce the number of resolves

I'm 100% in favour of the former and against the latter. As I said before, this can solve issues beyond CI - benchmarking being the example I'm aware of currently.

jamesp:
The pro for this approach is that we get a known, reproducible, test environment for every CI run until we explicitly update the lockfiles committed to git.  

As a workflow, I was thinking somethign along the lines of:
1. GH Action updates the lockfiles once a week and creates a PR.  If all tests pass, we merge the PR and now tests run against latest dependencies.
2. If tests fail on this PR (most likely being matplotlib image hash discrepancies) we CO the PR, fix the code & tests, commit and merge.  Changes required due to dependency change are therefore isolated from other development changes.

bjlittle:
@jamesp How will this work for feature branches, that pull in additional or different requirements. Any thoughts?

bjlittle:
@jamesp Interesting that there are graphical `phash` failures... what's changed?

bjlittle:
@jamesp You need to accompany this PR with developer doc changes... to clarify the workflow.

That might be helpful for those that follow...

trexfeathers:
OMG I'd forgotten the second benchmarking benefit 😀 I was originally just thinking about maintaining the correct dependencies for running the tests, which can be slow. But even more importantly, having dependency versions under 'commit control' will for the first time allow us to **distinguish performance regressions specifically caused by dependency changes**, as opposed to changes we have made ourselves.

This was actually raised as an original NG-VAT requirement (AVD-1382), but it was going to be an intimidating piece of work which I hadn't made a start on. But maybe now we won't have to!

jamesp:
> @jamesp How will this work for feature branches, that pull in additional or different requirements. Any thoughts?

Good question that I don't have an answer to!  The lockfiles would need to be regenerated in the feature branch, we could provide documentation/ script for doing that, or automate it (such as with precommit-ci if `requirements/ci/py3X.yml` files have changed.

rcomer:
> The lockfiles would need to be regenerated in the feature branch, we could provide documentation/ script for doing that

A script that one could run locally would also be useful for general PRs if they involve adding/pinning/unpinning packages.

bjlittle:
> Good question that I don't have an answer to! The lockfiles would need to be regenerated in the feature branch, we could provide documentation/ script for doing that, or automate it (such as with precommit-ci if `requirements/ci/py3X.yml` files have changed.

Yeah, we can see how we roll... it's not a blocker for me, just a passing thought on workflow.

Otherwise, this is awesome :beers:

pp-mo:
>  there are graphical phash failures... 

Might be.. ?
At present we have pinned versions : imagehash >= 4 **_and_** pillow(aka PIL) < 7

There is an outstanding problem here, where we **_assumed_** that imagehash results (hashes) should be stable over time (i.e. package versions), but that isn't actually guaranteed by the packages in question.
So we pinned 
See : https://github.com/SciTools/test-iris-imagehash/issues/28
and related change in Iris : https://github.com/SciTools/iris/pull/3630

pp-mo:
> phash

I see the latest is installing imagehash 4.2.0 [see here](https://cirrus-ci.com/task/5125449773744128)

Unfortunately, AFAICT when it finds a cached env, it no longer emits a conda list, so I can't easily work out what version was used in the latest commit with successful tests, i.e. https://github.com/SciTools/iris/commit/a7a8fcef7a0e452f2c102a05b997bbb611ed6ece

It would be good to fix that ?

pp-mo:
> phash

[Latest commit with successful tests on master, full-py36](https://github.com/SciTools/iris/commit/83eead65363b85eb7c5adba312c97ab97bca8546) does have [a conda-list](https://cirrus-ci.com/task/5742722557411328?logs=tests#L99), and this has installed ..
 ```
imagehash                 4.2.0              pyhd8ed1ab_0    conda-forge
pillow                    6.2.2            py36h8328e55_0    conda-forge
```

Meanwhile, the failing full-py36 test here has :
```
imagehash                 4.2.0              pyhd8ed1ab_0    conda-forge
pillow                    6.2.2            py36h8328e55_0    conda-forge
```

This looks identical..
So I guess that means a real change in the image outputs.
For same pair of runs, we have
```
old-ok: 
matplotlib                3.3.4            py36h5fab9bb_0    conda-forge
matplotlib-base           3.3.4            py36h62a2d02_0  

new-failing:
matplotlib                3.3.4            py36h5fab9bb_0    conda-forge
matplotlib-base           3.3.4            py36hd391965_0    conda-forge
```

There is a tiny change there with a newer build.
**Can that really be significant ??**
I am getting a bit lost here ...

**Comment by rcomer** [comment]
@pp-mo Cartopy 0.19 was released last week.  Could it be that?

**Comment by pp-mo** [comment]
> @rcomer Cartopy 0.19 was released last week. Could it be that?

Ah yes, that is **_far_** more likely !!

I  guess *all * the tests images are generated with Cartopy.
Sounds like we will need to generate some new test images + if ok add them to test-iris-imagehash

**Comment by jamesp** [comment]
Approx 50s of every cirrus run is loading the conda cache (https://github.com/SciTools/iris/pull/4108/files#diff-62587956f943bb2503db7bc6dd27d0d888074a1c0ecaab3f570ad611aff0f7bbR59).

I think with this change, we potentially only need to cache the nox folder.  We would have to download the packages again directly from conda-forge if the manifest changes, but that should be much less frequent than every CI run

File: .github/workflows/refresh-lockfiles.yml [hunk 1]
```diff
@@ -0,0 +1,73 @@
+# This workflow periodically creates new environment lock files based on the newest
+# available dependency pacakges on conda-forge.
```
rcomer, line 2: typo “pacakges”

noxfile.py
```diff
@@ -112,41 +120,47 @@ def prepare_venv(session):
...
-            f"--prefix={session.virtualenv.location}",
-            f"--file={fname}",
-            "--prune",
-        )
-        session._run(*command, silent=True, external="error")
+    lockfile = session_lockfile(session)
+    venv_dir = session.virtualenv.location_name
+
+    if not venv_populated(session):
+        # environment has been created but pacakages not yet installed
```
rcomer, line 127: typo “pacakges”

.github/workflows/refresh-lockfiles.yml
jamesp: 🤦

docs/src/developers_guide/contributing_pull_request_checklist.rst
```diff
@@ -29,6 +29,10 @@ is merged.  Before submitting a pull request please consider this list.
 #. **Check all modified and new source files conform to the required**
    :ref:`code_formatting`.
 
+#. **Check all new dependencies added to the requirements/ci/*.yml files.**  If
+   dependencies have been added new nox testing lockfieles should be generated too,
```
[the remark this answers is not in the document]
jamesp: 🤦

jamesp:
Thanks @rcomer! And good spot on Cartopy 0.18 too.  I've put that back to 0.18 in the lockfiles and tests pass 🎉

tools/update_lockfiles.py
```diff
@@ -0,0 +1,55 @@
...
+parser.add_argument('files', nargs='+', 
+    help="List of environment.yml files to lock")
+parser.add_argument('--output-dir', '-o', default='.', 
+    help="Directory to save output lock files")
+
+args = parser.parse_args()
+
+for infile in args.files:
+    print(f"generating lockfile for {infile}", file=sys.stderr)
+    fname = os.path.basename(infile)
```
trexfeathers, line 41: Switch to `Pathlib` if possible.

noxfile.py
```diff
@@ -184,7 +198,7 @@ def black(session):
 
 
 @nox.session(python=PY_VER, venv_backend="conda")
-def tests(session):
+def tests(session: nox.sessions.Session):
```
trexfeathers, line 201: Why is `session: nox.sessions.Session` not necessary with the subsequent sessions in this file?

tools/update_lockfiles.py
jamesp: Interesting! What are the benefits of using `pathlib` over `os.path`?  I'll be honest, I'm not super keen on the overloading of the `__div__` operator

trexfeathers: @pp-mo presented it to me as the modern way to go, and TBH I have found it nice and intuitive through lots of small conveniences. [No doubt there are myriad reasons](https://treyhunner.com/2018/12/why-you-should-be-using-pathlib/), but two that spring to mind from personal experience:

* The availability of `Path.glob()`
* Being able to write `mypath / subdir / subsubdir`

noxfile.py
jamesp: They are necessary. I hadn't completed the type annotations.  Done now

noxfile.py
```diff
@@ -25,57 +26,64 @@
...
-
-    Parameters
-    ----------
-    session: object
-        A `nox.sessions.Session` object.
+def session_lockfile(session: nox.sessions.Session) -> Path:
+    """Return the path of the session lockfile."""
+    return Path(
+        f"requirements/ci/nox.lock/py{session.python.replace('.', '')}-linux-64.lock"
+    )
```
trexfeathers, line 33:
```suggestion
    file_name = f"py{session.python.replace('.', '')}-linux-64.lock"
    file_path = Path("requirements") / "ci" / "nox.lock" / file_name
    return file_path
```

.github/workflows/refresh-lockfiles.yml
```diff
@@ -0,0 +1,73 @@
+# This workflow periodically creates new environment lock files based on the newest
+# available dependency packages on conda-forge.
```
trexfeathers, line 2: Whether or not `conda-forge` is the source depends on the files within `requirements/ci`. Could we come up with a more agnostic statement here?

docs/src/developers_guide/contributing_ci_tests.rst
```diff
@@ -38,6 +38,36 @@ The above `cirrus-ci`_ tasks are run automatically against all `Iris`_ branches
...
 
+
+.. _cirrus_test_env:
+
+Cirrus CI Test environment
+--------------------------
+
+The test environment on the Cirrus-CI service is determined from the requirement files
+in `requirements/ci/py**.yml`.  These are conda environment files that list the entire
+set of build, test and run requirements for iris.
```
trexfeathers, line 49:
```suggestion
set of build, test and run requirements for Iris.
```

docs/src/developers_guide/contributing_ci_tests.rst
```diff
@@ -38,6 +38,36 @@ The above `cirrus-ci`_ tasks are run automatically against all `Iris`_ branches
...
+
+   python tools/update_lockfiles.py -o requirements/ci/nox.lock requirements/ci/py*.yml
+
+or simply::
+
+   make lockfiles
+
+and add the changed lockfiles to your pull request.
+
+New lockfiles are generated automatically each week to ensure that iris continues to be
```
trexfeathers, line 67:
```suggestion
New lockfiles are generated automatically each week to ensure that Iris continues to be
```

docs/src/developers_guide/contributing_ci_tests.rst
```diff
@@ -38,6 +38,36 @@ The above `cirrus-ci`_ tasks are run automatically against all `Iris`_ branches
...
 `Cirrus-CI Dashboard`_ for details of recent past and active Iris jobs.
 
+
+.. _cirrus_test_env:
+
+Cirrus CI Test environment
+--------------------------
+
+The test environment on the Cirrus-CI service is determined from the requirement files
+in `requirements/ci/py**.yml`.  These are conda environment files that list the entire
```
trexfeathers, line 48:
```suggestion
in ``requirements/ci/py**.yml``.  These are conda environment files that list the entire
```

docs/src/developers_guide/contributing_ci_tests.rst
```diff
@@ -38,6 +38,36 @@ The above `cirrus-ci`_ tasks are run automatically against all `Iris`_ branches
...
+set of build, test and run requirements for iris.
+
+For reproducible test results, these environments are resolved for all their dependencies
+and stored as lock files in `requirements/ci/nox.lock`.  The test environments will not
```
trexfeathers, line 52:
```suggestion
and stored as lock files in ``requirements/ci/nox.lock``.  The test environments will not
```

docs/src/developers_guide/contributing_pull_request_checklist.rst
```diff
@@ -29,6 +29,10 @@ is merged.  Before submitting a pull request please consider this list.
 #. **Check all modified and new source files conform to the required**
    :ref:`code_formatting`.
 
+#. **Check all new dependencies added to the requirements/ci/*.yml files.**  If
```
trexfeathers, line 32:
```suggestion
#. **Check all new dependencies added to the ``requirements/ci/*.yml`` files.**  If
```

trexfeathers:
@jamesp this looks great! 💐

All my comments are minor things - it already works AFAICT. [Although there still seems to be an imagehash failure](https://github.com/SciTools/iris/pull/4108/checks?check_run_id=2465477040) so there must still be a dependency issue somewhere.

What do you want to do about the link=check failures, given we know there is a fix in #4104?

noxfile.py
jamesp: I get the concept, but this is three lines instead of one, and is it more readable?  As far as I can tell, `pathlib.Path` understands / within a string as an OS agnostic path separator

.github/workflows/refresh-lockfiles.yml
jamesp: Good point, will change

noxfile.py
trexfeathers: >As far as I can tell, `pathlib.Path` understands / within a string as an OS agnostic path separator

I didn't know that - this was my main reason for suggesting. Regarding readability: FWIW I do find the 3 lines more readable but I'm not concerned either way.

rcomer:
> there still seems to be an imagehash failure

That particular gallery example is missing bits (see #4110).  I have no idea if the failure is related to that though.

jamesp:
> What do you want to do about the link-check failures, given we know there is a fix in #4104?

Given the output of the test above suggests the code is working, it's the same test that's failing,I would suggest, if everyone is comfortable, we ignore that here, and rebase #4104.

I'm not sure about the gallery failure, I'll take a look at the imagehash on 3.8 and see what the issue is. the hamming distance score suggests it's not far off.

rcomer:
@jamesp I think a change at matplotlib v3.4 has _reinstated correct behaviour_ for that gallery example.  See #4110

jamesp:
We should address that image hash in a separate PR, I think this one is about complete now.

jamesp:
What more do we need to do to get comfortable with merging this with the failing doc tests?   We seem to be approaching a circular dependency across a number of PRs!

trexfeathers:
>What more do we need to do to get comfortable with merging this with the failing doc tests? We seem to be approaching a circular dependency across a number of PRs!

* https://github.com/SciTools/iris/pull/4112#issuecomment-830277530
* https://github.com/SciTools/iris/pull/4108#discussion_r622943260

trexfeathers:
@jamesp don't get me wrong, I'm also **very** keen we get this over the line before we get an even bigger spider web of blocking problems.

jamesp:
Thanks @trexfeathers.  Pushed commits to address those points above.

I agree this is a change to how things are done that should be discussed.  Could @trexfeathers or @pp-mo setup a quick call for all those interested?

rcomer:
That [anomaly log gallery failure](https://cirrus-ci.com/task/4674708315570176?logs=tests#L550) also came up in my local testing but I ignored it because it wasn't previously failing here.  The idiff output looked similar to the change shown at https://github.com/SciTools/test-iris-imagehash/pull/40, except the colour bar was also affected.

For the square root failure, see #3993.

😔

jamesp:
I've reverted the changes to `imagerepo.json` as they should be in a separate PR.  Let's keep this one entirely focussed on the CI changes so that it can be easily reverted if needed.

Once this is merged as a priority we should fix:
1. square root failure #3993 
2. image hashes for galleries and test
3. linkcheck failures.

trexfeathers (approved):
Can confirm this only includes the _expected_ CI failures. Merging as agreed.

PR #4112, opened 2021-04-30
Update hashes for gallery wind plot

### LINKED ISSUE #4110: Gallery: wind plot missing lake features
## 📚 Documentation

The [Plotting Wind Direction Using Quiver](https://scitools-iris.readthedocs.io/en/stable/generated/gallery/meteorology/plot_wind_speed.html) example uses NaturalEarth to add lake outlines.  But those outlines don't actually show up in the displayed plots.  It looks like they disappeared at Iris 2.1:

* [Iris 2.0 version](https://scitools.org.uk/iris/docs/v2.0/examples/Meteorology/wind_speed.html)
* [Iris 2.1 version](https://scitools.org.uk/iris/docs/v2.1/examples/Meteorology/wind_speed.html)

rcomer:
### Description

Closes #4110 

Image update at https://github.com/SciTools/test-iris-imagehash/pull/42

If I've understood, we don't usually remove the old hashes.  However in this case the old ones didn't include the lake outline that should have been there so were giving us false passes.

---

### DIFF
--- a/lib/iris/tests/results/imagerepo.json
+++ b/lib/iris/tests/results/imagerepo.json
@@ -146,11 +146,11 @@
     ],
     "gallery_tests.test_plot_wind_speed.TestWindSpeed.test_plot_wind_speed.0": [
         "https://scitools.github.io/test-iris-imagehash/images/v4/bcf924fb9306930ce12ccf97c73236b28ecec4cd3e29847b18e639e6c14f1a09.png",
-        "https://scitools.github.io/test-iris-imagehash/images/v4/e9e960e996169306c1ee9e96c29e36739e13c07d3d61c07f39a139a1c07f3f01.png"
+        "https://scitools.github.io/test-iris-imagehash/images/v4/e9e960e996169306c1fe9e96c29e36739e13c06c3d61c07f39a139e1c07f3f01.png"
     ],
     "gallery_tests.test_plot_wind_speed.TestWindSpeed.test_plot_wind_speed.1": [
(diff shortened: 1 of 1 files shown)


trexfeathers:
>If I've understood, we don't usually remove the old hashes.

It think it's a backwards compatibility thing. When we were putting a minimum pin on a dependency last year (#3762) I took the extra step of removing the old references (**but not the files themselves from test-iris-imagehash**), [although it was clear I wasn't expected to](https://github.com/SciTools/iris/pull/3762#issuecomment-685902750).

trexfeathers (changes requested):
~~Following from https://github.com/SciTools/iris/pull/4112#issuecomment-830248442, please could you keep the old references too?~~

~~As I understand it, the changes are caused by new dependency versions (correct me if wrong), but we have no contract in place stating that Iris won't work with the previous dependencies, so the test harness shouldn't make this assumption either.~~

Apologies, I'd missed your point about false passing - this should **never** have been OK.

trexfeathers:
>So this is only going to pass when combined with #4108.

I'd therefore vote for 566f187fcfd40495f200020b3c7d8f1905fce67f being pushed to [jamesp:ci-lock](https://github.com/jamesp/iris/tree/ci-lock), but I expect Git decorum says we should asked @jamesp's permission first!

rcomer:
Change now incorporated into #4108, so closing this as duplicate.

rcomer:
No longer included in #4108.

trexfeathers:
@rcomer are you OK to rebase?

trexfeathers (approved):
Does as intended, one step closer to eliminating the suite of failures recently arisen on `master`. Thanks @rcomer!

PR #4087, opened 2021-04-08
Fix cirrus ci and mpl

bjlittle:
### Description
Follow on PR from #4084

---
[Consult Iris pull request check list]( https://scitools-iris.readthedocs.io/en/latest/developers_guide/contributing_pull_request_checklist.html)

PR #4115, opened 2021-05-04
Corrected plot_anomaly_log_colouring for new Matplotlib linscale rules.

trexfeathers:
### Description

Restores output of the anomaly log colouring gallery to pre #4087 by using a new `linscale` argument (can't go back to `linscale=0` as no longer allowed by Matplotlib).

---

### DIFF
--- a/docs/gallery_code/general/plot_anomaly_log_colouring.py
+++ b/docs/gallery_code/general/plot_anomaly_log_colouring.py
@@ -68,13 +68,13 @@ def main():
     # Create a 'logarithmic' data normalization.
     anom_norm = mcols.SymLogNorm(
         linthresh=minimum_log_level,
-        linscale=1,
+        linscale=0.01,
         vmin=-maximum_scale_level,
         vmax=maximum_scale_level,
     )
     # Setting "linthresh=minimum_log_level" makes its non-logarithmic
     # data range equal to our 'zero band'.
-    # Setting "linscale=1" maps the whole zero band to the middle colour value
+    # Setting "linscale=0.01" maps the whole zero band to the middle colour value
     # (i.e., 0.5), which is the neutral point of a "diverging" style colormap.
 
     # Create an Axes, specifying the map projection.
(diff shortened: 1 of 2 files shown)


docs/gallery_code/general/plot_anomaly_log_colouring.py
```diff
@@ -68,13 +68,13 @@ def main():
...
-        linscale=1,
+        linscale=0.01,
         vmin=-maximum_scale_level,
         vmax=maximum_scale_level,
     )
     # Setting "linthresh=minimum_log_level" makes its non-logarithmic
     # data range equal to our 'zero band'.
-    # Setting "linscale=1" maps the whole zero band to the middle colour value
+    # Setting "linscale=0.01" maps the whole zero band to the middle colour value
     # (i.e., 0.5), which is the neutral point of a "diverging" style colormap.
```
rcomer, line 78: I'll be honest, I'm really struggling to match this comment up with what the [mpl docs](https://matplotlib.org/stable/api/scale_api.html?highlight=symmetricallogscale#matplotlib.scale.SymmetricalLogScalel) say about `linscale`.  I'm getting the feeling that either the original comment was wrong, or the behaviour has changed.  Or I'm just missing something... 😕

rcomer: It's the first bit about "linthresh=minimum_log_level" that I'm particularly struggling with, rather than the changed bit.

trexfeathers: The language of these comment lines makes no sense to me either, but I assumed that to be due to the writer's greater knowledge in this area. Seems important to understand and correct before merging.

rcomer: Looks like @pp-mo was the originator, but a Whole Seven Years Ago!!! #1040

rcomer: Given how much matplotlib would have moved on in seven years, it's entirely possible that there are better ways of achieving this outcome now anyway.  So possibly we should just merge this now for the sake of fixing master, and open a new issue to revisit later?

trexfeathers:
The good news is that this finally gives us a passing [py3.8 doc tests (gallery)](https://github.com/SciTools/iris/pull/4115/checks?check_run_id=2501524464).

jamesp (approved):
Looks good to me, the only failing tests now are the square root failures 👍

docs/gallery_code/general/plot_anomaly_log_colouring.py
trexfeathers: Having read in more detail and talked it over with @stephenworsley, I'm reasonably happy with how it's written already.

pp-mo: Just found this.
FWIW there was a perfectly good documentation for this up to  Matplotlib v3.3,
https://matplotlib.org/3.3.4/api/_as_gen/matplotlib.colors.SymLogNorm.html

But they seem to have lost  it somewhere since ??

PR #4120, opened 2021-05-06
Gallery: simplify quiver example

rcomer:
### Description

* This example predates `iris.plot.quiver` but, now we have `iris.plot.quiver` we really should advertise it here!
* I'm not sure why the 360 degree adjustment to the longitudes was required, but the plots work fine without it now.
* Removed some minor redundancies:
  * Unused `coord_categorisation` import
  * First explicit call to `plt.figure()` (I initially also removed the second call to `plt.figure()` which was fine for interactive plotting and the gallery test, but Sphinx only produced one image containing the second contour and two colorbars).
  * Triple calculation of windspeed.

---


bjlittle (approved):
@rcomer Awesome, the example is a lot clearer now, thanks :+1:

trexfeathers:
I admit I haven't looked, but do any of the changes here also apply to [the recently added barbs plot example](https://github.com/SciTools/iris/blob/121df0564d09c2c7d9dcc50a9bccd87b834ffea0/docs/gallery_code/meteorology/plot_wind_barbs.py) too?

rcomer:
@trexfeathers if I was going to be really picky, I would say that
* `magnitude` and `windspeed` appear to be the same thing
* I _think_ both of the above should automatically be in knots because `uwind` and `vwind` are, so some of the `convert_units` calls are unnecessary
* The call to `plt.figure` would happen "behind the scenes" anyway if not done explicitly

rcomer:
Thanks @bjlittle!

PR #4127, opened 2021-05-10
Contourf workaround for Cartopy 0.19

rcomer:
### Description
A change at Cartopy 0.19 causes axes extent to be updated when contours are added (see https://github.com/SciTools/cartopy/issues/1780).  This is a problem for our special handling for antialiasing in `iris.plot.contourf` because maps are cropped around the extra contours (see discussion at #4125).  @greglucas has written a fix at https://github.com/SciTools/cartopy/pull/1784 but advised that it is unlikely to make it into a release soon.  So if we want Iris v3.0.2 to play nicely with Cartopy 0.19, we will probably need a workaround.  I propose this one, which just reinstates whatever the axes extent was before the call to `contour`.

---

### DIFF
--- a/lib/iris/plot.py
+++ b/lib/iris/plot.py
@@ -1073,6 +1073,15 @@ def contourf(cube, *args, **kwargs):
             # any boundary shift.
             zorder = result.collections[0].zorder - 0.1
             axes = kwargs.get("axes", None)
+
+            # Workaround for cartopy#1780.  We do not want contour to modify
+            # extent, so get hold of it and re-apply it after contour call.
+            if axes is None:
+                _axes = plt.gca()
+            else:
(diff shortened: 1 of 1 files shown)


greglucas:
Another idea (untested) would be to call
`ax.dataLim.update_from_data_xy(extent, ignore=False)`
before the contour and then call it with True and the current extents right after the contour call reset it (just swap what `None` is defaulting too within Cartopy). I think this would get you the expanding part too in case you want that.

rcomer:
Thanks @greglucas.  My version has definitely broken at least one example, so I will give your suggestion a try!

rcomer:
Now also in v.3.0.x branch via #4130 👍

PR #4138, opened 2021-05-15
 Preserve dtype for wrap_lons 

### LINKED ISSUE #4119: Preserve dtype for wrap_lons
## 📰 Custom Issue

Currently, `iris.analysis.cartography.wrap_lons` promotes dtype to `np.float64`. This is causing some users to have to manually change the dtype back to its original version.
https://github.com/SciTools/iris/blob/98baee21721826a2a3c25d9169ed7e97e16e5e12/lib/iris/analysis/cartography.py#L70-L73
I believe this is done to avoid rounding errors within the calculation. If this is the case, It may be worth converting back to the original dtype after the calculation with something like the following code:

```
orig_dtype = lons.dtype
lons = lons.astype(np.float64)
return (((lons – base + period * 2) % period) + base).astype(orig_dtype)
```

Rohansjamadagni:
### Description
Fixes #4119  

NOTE: Adding this change resulted in a failure of the cube unit test,  so I have added the changes necessary to account the changes in the dtype

### DIFF
--- a/lib/iris/analysis/cartography.py
+++ b/lib/iris/analysis/cartography.py
@@ -69,8 +69,9 @@ def wrap_lons(lons, base, period):
     """
     # It is important to use 64bit floating precision when changing a floats
     # numbers range.
+    orig_dtype = lons.dtype
     lons = lons.astype(np.float64)
-    return ((lons - base + period * 2) % period) + base
+    return (((lons - base + period * 2) % period) + base).astype(orig_dtype)
 
 
 def unrotate_pole(rotated_lons, rotated_lats, pole_lon, pole_lat):
--- a/lib/iris/plot.py
+++ b/lib/iris/plot.py
@@ -1087,6 +1087,18 @@ def contourf(cube, *args, **kwargs):
             # any boundary shift.
             zorder = result.collections[0].zorder - 0.1
             axes = kwargs.get("axes", None)
+
+            # Workaround for cartopy#1780.  We do not want contour to shrink
+            # extent.
+            if axes is None:
+                _axes = plt.gca()
+            else:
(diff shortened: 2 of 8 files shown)


Rohansjamadagni:
I have now signed the CLA

rcomer:
Hi @Rohansjamadagni and @jeffreypaul15 , thanks for contributing!  Sorry that nobody responded to this sooner - Team Iris have been very busy.  There are just a few things to address to get our checks passing:

* Because this change also includes a commit from @jeffreypaul15, he will also need to sign the CLA.
* The `main` branch has moved on since you created your branch, so you will need to rebase to pick up those changes.  If you've not done that before, this should work:
```
git checkout bug-fix-issue-4119
git fetch upstream
git rebase upstream/main
```
* Thanks for fixing the unit test to reflect the new behaviour.  There is also a doctest that needs updating [here](https://github.com/SciTools/iris/blob/6109cfb64ddfe9bf50a8a34e21eb9f9c62a077c5/lib/iris/analysis/cartography.py#L67).  I think it should just be a case of removing the decimal points.

rcomer:
Hi @jeffreypaul15, are you able to sign our [CLA](https://scitools.org.uk/cla/v4/form)?  We appreciate that you both might not have time to work on this PR anymore, and we can make the required branch changes ourselves if that is the case.  However, we are not allowed to accept any commit without a CLA from the author, so we can't take this any further without that.

rcomer:
This one has now stalled for some time so I will close it.  We can re-open if there is interest but, given that there is now a conflict in `cartography.py`, it may be just as easy to start again with a new branch and new PR.

PR #4132, opened 2021-05-13
Fix broken links

rcomer:
### Description

Fixed a couple of broken links.

---
[Consult Iris pull request check list]( https://scitools-iris.readthedocs.io/en/latest/developers_guide/contributing_pull_request_checklist.html)

PR #4139, opened 2021-05-15
Fix coord_categorisation.add_weekday

rcomer:
### Description

See #4137 

Explicitly make the `coord_categorisation` functions always use a `cftime.datetime` instance, so that tests pass with `cf_units` v2.1.5.  I also removed the comment about `Unit.num2date` being deprecated.  That comment dated back to the days when `Unit` was part of Iris, and I can't see anything in `cf_units.__init__.py` to suggest `Unit.num2date` is deprecated.

I also fixed a link to the `black` docs.  This duplicates part of #4132, but it looks like the main tests in that PR can't pass without pulling in this change anyway.

I am on leave for the next week but happy if someone else applies any changes here if they are needed in the meantime.

---

### DIFF
--- a/docs/src/developers_guide/contributing_code_formatting.rst
+++ b/docs/src/developers_guide/contributing_code_formatting.rst
@@ -39,7 +39,7 @@ linted according to our ``.flake8`` configuration file. Note that,
 for each ``.pre-commit-config.yaml`` git hook.
 
 Additionally, you may wish to enable ``black`` for your preferred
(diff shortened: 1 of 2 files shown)


trexfeathers (approved):
Thanks @rcomer!

While there are many other uses of `num2date()`, they all appear to be happily working with the new default return of `cftime.datetime` already, so no need to add the same argument to all those calls too.

PR #4135, opened 2021-05-14
Tweak to speed up dask wrapping of netcdf variables

pp-mo:
### Description
Hopefully, addresses #4134
This significantly improves a problem with slow loading, which applies to netcdf files with lots of  variables
The change is tiny, but significant !

I think this problem has existed ever since we moved to Dask 2.0.
At that point we identified much slower loading of large PP/FF files, and fixed it [like this](https://github.com/SciTools/iris/pull/3659).
[That code is still in place](https://github.com/SciTools/iris/blob/v3.0.x/lib/iris/fileformats/pp.py#L597-L600) : in view of this new approach, that can probably now be removed -- and should be !
But as this is intended as a "quick win" for v3.0.2, I'll defer that until later.

---
[Consult Iris pull request check list]( https://scitools-iris.readthedocs.io/en/latest/developers_guide/contributing_pull_request_checklist.html)

PR #4141, opened 2021-05-18
Pp daskfix

pp-mo:
### Description
This is a follow-up to #4135 and ***should not be merged before that one*** (hence draft WIP)
( and will doubtless need re-basing anyway, due to various current testing problems )

I realised that when #4135 is accepted, it also stops the `dask.array.from_array` within `iris._lazy_data.as_lazy_data()` from realising PP/FF data.
So we no longer need the previous um-data-specfic workaround from #3659.
This PR removes that code.

---

### DIFF
--- a/lib/iris/tests/unit/fileformats/pp/test_PPDataProxy.py
+++ b/lib/iris/tests/unit/fileformats/pp/test_PPDataProxy.py
@@ -10,7 +10,6 @@
 import iris.tests as tests
 
 from unittest import mock
-import numpy as np
 
 from iris.fileformats.pp import PPDataProxy, SplittableInt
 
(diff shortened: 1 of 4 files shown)


lbdreyer:
I have rebased this PR onto the v3.0.x branch now that #4135 is merged in

lib/iris/tests/unit/fileformats/pp/test_PPDataProxy.py
```diff
@@ -34,128 +33,5 @@ def test_lbpack_raw(self):
...
-        )
-
-        # Test the requested indexing operation.
-        result = proxy.__getitem__(indices)
-
-        # Check the behaviour and results were as expected.
-        self.assertEqual(mock_fileopen.called, data_was_fetched)
-        self.assertIsInstance(result, np.ndarray)
-        self.assertEqual(result.dtype, test_dtype)
-        self.assertEqual(result.shape, result_shape)
```
lbdreyer, line 94: These tests are no longer necessary as we now *always* open the file when calling `__getitem__ ` making 
```self.assertEqual(mock_fileopen.called, data_was_fetched)```
redundant.

The other checks are testing functionality we have tests for elsewhere.

lbdreyer:
@bjlittle I've made a functional change to this PR so I couldn't merge it. Do you mind having a look over it?

bjlittle (approved):
@pp-mo and @lbdreyer This looks good to me.

I don't think that we need extra test coverage here, as we're pretty much back where we started pre #3695, which I'm fine with :+1:

bjlittle:
@lbdreyer Perhaps add an entry on the whatsnew?

pp-mo:
@lbdreyer @bjlittle many thanks for progressing this in my absence.
respect! 🙏

PR #4158, opened 2021-05-26
Pp mo nc multivat speedup

lbdreyer:
### Description

---


lbdreyer:
It had looked like the tests in  #4135 were failing, but now the cirrus-ci is updated, it is showing that they all pass. So this (which was going to be my backup) is no longer needed and can be closed.

PR #4162, opened 2021-06-01
Bump black version

bjlittle:
### Description
This PR updates the `black` version to the latest offering by unpinning.

I did consider removing `black` from the list of dependencies, but we do require to include `black` for those that want to test `iris` with `nox` on their local host.

Rather than continually update the specific `black` version in several places, I suggest we simply let the `black` version float to the latest version at environment package dependency resolve time, and allow the `pre-commit-bot` to notify us when a later version of `black` is available to the community.

Seems like a reasonable compromise to take here.

---

### DIFF
--- a/noxfile.py
+++ b/noxfile.py
@@ -190,7 +190,7 @@ def black(session: nox.sessions.Session):
 
     """
     # Pip install the session requirements.
-    session.install("black==21.5b2")
+    session.install("black")
     # Execute the black format checker on the package.
     session.run("black", "--check", PACKAGE)
     # Execute the black format checker on this file.
--- a/noxfile.py
+++ b/noxfile.py
@@ -190,7 +190,7 @@ def black(session: nox.sessions.Session):
 
     """
     # Pip install the session requirements.
-    session.install("black==21.5b1")
+    session.install("black==21.5b2")
     # Execute the black format checker on the package.
     session.run("black", "--check", PACKAGE)
     # Execute the black format checker on this file.
--- a/noxfile.py
+++ b/noxfile.py
@@ -190,7 +190,7 @@ def black(session: nox.sessions.Session):
 
     """
     # Pip install the session requirements.
-    session.install("black")
+    session.install("black==21.5b2")
     # Execute the black format checker on the package.
     session.run("black", "--check", PACKAGE)
     # Execute the black format checker on this file.
--- a/requirements/ci/py36.yml
+++ b/requirements/ci/py36.yml
@@ -34,7 +34,7 @@ dependencies:
 
 # Test dependencies.
   - asv
-  - black=21.5b2
+  - black
   - filelock
   - flake8
   - imagehash>=4.0
--- a/requirements/ci/py36.yml
+++ b/requirements/ci/py36.yml
@@ -34,7 +34,7 @@ dependencies:
 
 # Test dependencies.
   - asv
-  - black
+  - black=21.5b2
   - filelock
   - flake8
   - imagehash>=4.0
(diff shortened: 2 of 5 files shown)


noxfile.py
```diff
@@ -190,7 +190,7 @@ def black(session: nox.sessions.Session):
 
     """
     # Pip install the session requirements.
-    session.install("black==21.5b1")
+    session.install("black")
```
trexfeathers, line 193:
```suggestion
    with Path(".pre-commit-config.yaml").open() as pc_file:
        pc_yaml = yaml.load(pc_file, Loader=yaml.Loader)
    black_repo = list(filter(lambda r: r["repo"] == "https://github.com/psf/black", pc_yaml["repos"]))[0]
    black_rev = black_repo["rev"]
    session.install(f"black={black_rev}")
```

requirements/ci/py36.yml
```diff
@@ -34,7 +34,7 @@ dependencies:
 
 # Test dependencies.
   - asv
-  - black=21.5b1
+  - black
```
trexfeathers, line 37: Remove this line?

requirements/ci/py37.yml
```diff
@@ -34,7 +34,7 @@ dependencies:
 
 # Test dependencies.
   - asv
-  - black=21.5b1
+  - black
```
trexfeathers, line 37: Remove this line?

requirements/ci/py38.yml
```diff
@@ -34,7 +34,7 @@ dependencies:
 
 # Test dependencies.
   - asv
-  - black=21.5b1
+  - black
```
trexfeathers, line 37: Remove this line?

requirements/test.txt
```diff
@@ -1,7 +1,7 @@
 # Test dependencies.
 
 asv
-black==21.5b1
+black
```
trexfeathers, line 4: Remove this line?

trexfeathers (changes requested):
>we do require to include `black` for those that want to test `iris` with `nox` on their local host.

Since Nox takes does its own `pip install` via `Session.install()`, I don't think this is true.

I therefore think we can get away with only requiring Black in the pre-commit config and in the noxfile. And we can have the noxfile read the pre-commit file to get a version. See my suggestions.

bjlittle:
It's not necessarily all about `black` here... as a developer I might want all the dependencies, but not run the tests. This could also apply to other packages like `flake8`.

As a minimal change dropping the pin seems appropriate.

If we want to start to rationalise the dependencies that we explicitly include, then I think that's a different PR entirely - just so as to not muddy the waters here.

bjlittle:
I've got thoughts on how we can go about that, and I don't want to make that the focus of this PR. Also, it might be worth a separate dev discussion before committing changes to code.

trexfeathers:
OK granted, but there is still a blocker to merging this one.

Pre-commit is the gatekeeper for commits, while Nox is the gatekeeper for merging PR's. If anyone pushes a PR **before** we get round to merging a pre-commit auto-update, we will get a very confusing PR.

So for me the bare minimum is getting Nox and pre-commit working with the same version. If there's something wrong with my suggestion I think we'll need to keep those two manually aligned as they are at the moment.

bjlittle:
Your suggested change only fits the use case for developer using `pre-commit`, which isn't a given. So they could commit changes with a later version of `black` that's ahead of `pre-commit` and `nox`.

I'm just going to backtrack and pin again. Safest option all round.

trexfeathers (approved):

bjlittle:
@trexfeathers Nice one thanks. 

Slightly more interesting than I was bargaining for :wink: ... although worth thinking about :thinking:

PR #4163, opened 2021-06-01
drop py36 support

bjlittle:
### Description
This PR drops `iris` support for `python` 3.6.x, which is now well overdue, see https://endoflife.date/python.

---


docs/src/conf.py
```diff
@@ -101,9 +102,34 @@ def autolog(message):
 
 build_python_version = ".".join([str(i) for i in sys.version_info[:3]])
 
+
+def _dotv(version):
+    result = version
+    match = re.match("^py(\d+)$", version)  # noqa: W605
```
jamesp, line 108: Could be a raw string instead of overriding the linter?

jamesp: https://www.flake8rules.com/rules/W605.html

noxfile.py
```diff
@@ -20,7 +20,7 @@
 PACKAGE = str("lib" / Path("iris"))
 
 #: Cirrus-CI environment variable hook.
-PY_VER = os.environ.get("PY_VER", ["3.6", "3.7", "3.8"])
+PY_VER = os.environ.get("PY_VER", ["3.7", "3.8"])
```
jamesp, line 23: Could /should this use the same discovery code that you've written for the docs?

jamesp: What we're actually testing in nox is not a specific version of python, but a specific environment that is defined by a yaml file in `requirements/ci`.

jamesp (commented):
Looking forward to seeing this one go in! 🚀

docs/src/conf.py
bjlittle: Done :+1:

noxfile.py
bjlittle: Hmmm I'll give it a try and see what it looks like... the hardest part is deciding on where such common code will live.

But yeah, that would be a nice win :+1:

jamesp: @bjlittle I'm happy for this to go in as-is with your latest updates.  Let me know if you plan to rework otherwise I'm merging today

bjlittle: @jamesp Okay, so I gave this a reasonably punt... but you swiftly run into a bit of a pickle as everything under the `requirements` directory isn't installed as a package, so you can't really safely reference it from `nox` or the docs without jumping through some hoops.

It is possible, and I suspect we'd either want to re-structure our resources or use `setuptools` sub-packages to do that properly... but I'm not really inclined to pursue this ATM, unless you want me to follow through on this?

trexfeathers: I'm surprised this is an issue. Sure, we run our tests on a installed instance of Iris, so there are concerns on inclusion/exclusion there. But as far as I was aware Nox was expected to operate on a checkout of a repo, rather than within an install. Indeed I've based a few things on this assumption.

lib/iris/tests/integration/test_pp.py
```diff
@@ -414,7 +414,7 @@ def test_hybrid_height_round_trip_no_reference(self):
             "Unable to create instance of HybridHeightFactory. "
             "The source data contains no field(s) for 'orography'."
         )
-        warn.assert_called_once_with(msg)
+        warn.assert_called_with(msg)
```
trexfeathers, line 417: I'm concerned that we got this error before you made the change:

```
FAIL: test_hybrid_height_round_trip_no_reference (iris.tests.integration.test_pp.TestVertical)
  File "/tmp/cirrus-ci-build/lib/iris/tests/integration/test_pp.py", line 417, in test_hybrid_height_round_trip_no_reference
    warn.assert_called_once_with(msg)
  File "/tmp/cirrus-ci-build/.nox/tests/lib/python3.7/unittest/mock.py", line 888, in assert_called_once_with
    raise AssertionError(msg)
AssertionError: Expected 'warn' to be called once. Called 2 times.
```

Doesn't this mean behaviour has changed? Why?

bjlittle: I'll see if I can replicate this locally...

noxfile.py
jamesp: I mean, you must be referencing `requirements` _somewhere_ from `nox` as those environments yamls are being read to create the test envs. But I think I understand your point to be that the introspection bit of nox where it works out what sessions it has might not be able to safely operate.  either way though, nox isn't going to work if the `requirements/ci` folder isn't there.

This isn't a big issue, happy to leave it hard coded.

lib/iris/tests/integration/test_pp.py
bjlittle: Okay, so frustratingly I can't recreate this failure locally. Either with a test environment that I resolve myself from the `py37` dependencies, nor with the `py37` lock...

I'm not overly concerned about this... although, I'm not loving this inconsistency in behavour. I could be down to a `nose` race condition... dunno.

That said, I'd be more worried if the warning wasn't being generated.

So, personally, I'm not seeing this as a blocker to merging unless you guys want to pursue it further.

noxfile.py
bjlittle: Just to be clear... the issues isn't about whether the directory exists, per se... rather that I was attempting to bank common code under `requirements/__init__.py` for both the docs and `nox` to use - that's the problem here.

The simple alternative is to put common code within `iris` (which I really don't want to do) or I duplicate the code within the `noxfile` and within the docs.

So it's do-able, just whatever you prefer... if you want to follow-thru with this

bjlittle: Either way, I don't think this should be a blocker for this PR. We can happily bank this PR "as is", and iterate on this later.

bjlittle: Also, for me this is all becoming a bit of a moot point, as a developer still requires to manually update the `matrix` for the minimal and full tests.

jamesp:
I'm happy with this, are you all done with changes @bjlittle?

bjlittle:
@jamesp Go for it :rocket:

jamesp (approved):

bjlittle:
@jamesp and @trexfeathers Awesome, thanks :beers:

lib/iris/tests/integration/test_pp.py
rcomer: I think I’ve seen this elsewhere, and it went away when tests were re-spun.

PR #4170, opened 2021-06-02
Replace pyke

pp-mo:
### Description
Eventually, will address #3415

#### Technical details of approach
  * I started by writing a drop-in replacement approach to replace Pyke rules with so-called 'action' routines.
    * at present it purposely mimics the structure of existing Pyke rules, plus the three-stage translation process built into 
       `iris.fileformats.netcdf._load_cube` , i.e. broadly [assert facts](https://github.com/SciTools/iris/blob/master/lib/iris/fileformats/netcdf.py#L592-L604); [run rules](https://github.com/SciTools/iris/blob/master/lib/iris/fileformats/netcdf.py#L592-L604); [complete cube](https://github.com/SciTools/iris/blob/master/lib/iris/fileformats/netcdf.py#L609-L652)
    * it does that by providing a new [thin wrapper object](https://github.com/SciTools/iris/blob/b118666e62a194eb131513ce77aeef869beba87b/lib/iris/fileformats/_nc_load_rules/engine.py#L57-L64) which can [mimic the Pyke "engine"](https://github.com/SciTools/iris/blob/b118666e62a194eb131513ce77aeef869beba87b/lib/iris/fileformats/netcdf.py#L418-L430).
    * for now, I have retained the current approach to storing Iris info on the 'engine' ([example](https://github.com/SciTools/iris/blob/22ff3ebacf04b741994578489706a833a3f05fb6/lib/iris/fileformats/_nc_load_rules/actions.py#L181)), and also [mimicked the storage of intermediate 'facts' there too](https://github.com/SciTools/iris/blob/22ff3ebacf04b741994578489706a833a3f05fb6/lib/iris/fileformats/_nc_load_rules/engine.py#L91-L98).
    * some of that could well simplify in future, but I'm aiming to ***first ensure equivalence with the exisiting behaviour***
    * N.B. currently incomplete, have only so far done the 'major' rules (grid-mapping and dim-coords).
  * I am also creating some targetted testing for this ...
    * the choice of tests is heavily informed by the structure of the existing rules
    * the area tested, essentially what is currently "the Pyke rules bit", is not maybe _that_ obvious a separate distinct category ..
      * .. but it should be an appropriate starting point for testing the 'behavioural equivalence'
      * I intend eventually to test for each code-branch of the rules behaviour
    * ... so, again, in future we could simplify/rework this 

#### Current progress + status summary
Done
  * actions + tests for
    * grid-mappings + dim-coords
    * time-coords
    * hybrid vertical coords

Todo:
  * some areas are still unaddressed - lacking _both_ actions implementation + tests
  * known problems
    1. already observed that actions do **_not_**  match original behaviour for some of the grid-mapping/dimcoords areas
         * ( intend to complete the test coverage before addressing this )
    2. missing parts of grid-mapping/dimcoords tests : 
        * still not exercising the following rules at all:
          * fc_build_auxiliary_coordinate_latitude
          * fc_build_auxiliary_coordinate_latitude_rotated
          * fc_build_auxiliary_coordinate_longitude
          * fc_build_auxiliary_coordinate_longitude_rotated
    3. odd extras : Ukmo-attributes ; label-coords ; cell-measures; ancillary-variables
  * Later (intended) stages : 
     1. get all functionality tested + working with existing (Pyke) code
     2. add actions missing for TODO features
     3. get all tests pasing with actions implementation
     4. check that _all_ Iris tests run with temporary shim : do all netcdf loads _both_ ways + check results are the same
     5. remove the Pyke implementation (+dependency), and tidy away all the scaffolding

**NOTE**: I'm assuming that we will have enough confidence to "simply replace", which means
  1. everything goes via "new route"
  2. completely removed the old Pyke-based rules (and dependency)
  3. don't need, or have, any 'Future'-style switch for user to revert to 'old' mechanics

---


pp-mo:
@trexfeathers if you can find a moment, I'd be interested to know if you can work through what's going on here ?

Especially, I'd like to think that once we're confident that behaviour is preserved, we can subsequently remove some of the scaffolding which is the legacy of "how it used to be done".

So, I'm hoping at some point to run all exisiting Iris tests, with changes to run any netcdf loading via **_both_** routes, and check for identical results in all cases.
With that, and given ~full codebranch testing (with the tests I'm still writing), I hope we can feel confident simply to replacing the "old way" with the "new" (and drop the Pyke dependency).

The 'engine mimic' approach I'm using should enable this extensive equivalence testing, and also to focus on **_just_** replacing the 'rules' with 'actions'.
But obviously that is going to leave a good deal of unnecessary complication in place.
I'm hoping I can complete the main effort by next week, and hopefully there will also be some time to introduce some subsequent simplification (i.e. remove the worst vestiges of the 'old way').

pp-mo:
Note: at this point [i.e. commit https://github.com/SciTools/iris/pull/4170/commits/5f18f2369b8ccc9e6844825dbf2af356abcfbf83], removing the skip on the non-Pyke testing shows quite a few anomalies.
I'm going to ignore that for now, while aiming to complete test coverage.

TBH, writing this test suite is taking rather longer than I had thought.

DPeterK:
@pp-mo this code might be of interest to the conversation here: https://github.com/informatics-lab/tiledb_netcdf/blob/master/nctotdb/data_model.py#L11. I had a go at writing a non-Pyke NetCDF Dataset "understander" recently, too... It functions, but isn't full-featured, as I thnk there are some cf-netcdf elements it still ignores. Might provide some food for thought

lib/iris/fileformats/_nc_load_rules/actions.py
```diff
@@ -0,0 +1,337 @@
...
+    rulename_prefix = "fc_"  # To match existing behaviours
+    rule_name = func_name
+    if rule_name.startswith(funcname_prefix):
+        rule_name = rule_name[len(funcname_prefix) :]
+    if not rule_name.startswith(rulename_prefix):
+        rule_name = rulename_prefix + rule_name
+    return rule_name
+
+
+def _default_rulenamesfunc(func_name):
```
stephenworsley, line 63: This seems to be functionally a copy of `convert_action_to_rulename`. Do there need to be two different functions here?

lib/iris/fileformats/_nc_load_rules/helpers.py
```diff
@@ -0,0 +1,1308 @@
+# Copyright Iris contributors
+#
+# This file is part of Iris and is released under the LGPL license.
+# See COPYING and COPYING.LESSER in the root of the repository for full
+# licensing details.
+"""
+All the pure-Python 'helper' functions which previously included in the Pyke
```
stephenworsley, line 7: "which were"

lib/iris/fileformats/_nc_load_rules/actions.py
pp-mo: You're right :+1: 
This has been through several phases of rework + got simplified, and probably isn't done yet (still developing).
I **_might_** still decide that all actions should simply return a rule-name-like-string, and then no wrapper is needed.
At present, that is looking simpler.

I'd like to leave this conversation unresolved + come back to it when I've done more dev on the actions.
(my  present focus is on getting tests for everything)

pp-mo:
@trexfeathers @stephenpascoe @bjlittle :eyes: 
Just fixed the issue header :  updated account of "progress, status + plans".

pp-mo:
(Updated status in header, again)

lib/iris/fileformats/_nc_load_rules/helpers.py
pp-mo: :+1:

lib/iris/fileformats/_nc_load_rules/actions.py
pp-mo: Ok, I have now simplified, but have not removed the wrapper / default-name mechanisms
(in https://github.com/SciTools/iris/pull/4170/commits/b26fbcd79e3e34d10f122cbadab46123348830c0)

pp-mo:
So, I believe the tests now at least exercise all the Pyke rules.
Hopefully they also cover all the rule activation patterns -- well I hope the important ones, at least.
Awaiting your view on that @stephenworsley  !

Next up: I will progressively extend + debug the 'actions', so that they match rules behaviour in all the tests.

pp-mo:
N.B. the test failure is yet another link-test failure :disappointed: 
But we can just ignore that.

rcomer:
@pp-mo I am pretty sure that link has been fixed in `master`, so should pass if you rebase 👍

lib/iris/tests/unit/fileformats/netcdf/load_cube/load_cube__activate/test__grid_mappings.py
```diff
@@ -0,0 +1,866 @@
...
+        if xco_is_dim:
+            # xdim has same name as xco, making xco a dim-coord
+            xdim_name = "xco"
+        else:
+            # use alternate dim-name, and put xco on the 'coords' list
+            xdim_name = "xdim_altname"
+            phenom_auxcoord_names.append("xco")
+        if yco_is_dim:
+            # ydim has same name as yco, making yco a dim-coord
+            ydim_name = "yco"  # This makes the X coord a dim-coord
```
stephenworsley, line 107: I don't understand this comment, was this an artefact of copying the line or is something happening that  don't understand?

lib/iris/tests/unit/fileformats/netcdf/load_cube/load_cube__activate/test__grid_mappings.py
```diff
@@ -0,0 +1,866 @@
...
+        else:
+            # use alternate dim-name, and put yco on the 'coords' list
+            ydim_name = "ydim_altname"  # This makes the X coord a dim-coord
```
stephenworsley, line 110: Same as above, I don't see how this would affect the X coord.

lib/iris/tests/unit/fileformats/netcdf/load_cube/load_cube__activate/test__grid_mappings.py
pp-mo: Yes, it's a copying error.
Thanks !

lib/iris/tests/unit/fileformats/netcdf/load_cube/load_cube__activate/test__grid_mappings.py
pp-mo: :+1:

pp-mo:
Rebased onto master, to pick up link-check fix :crossed_fingers:

pp-mo:
Ok, I broke some of the other tests there by turning on the "compare all results pyke + non-pyke" option.
But I'll fix those too shortly.

pp-mo:
That's it !
I've now added support for **all** the rules behaviour, and it all seems to all work.
At present, the rules tests are running everything through both routes + checking that results compare.
 :grin:

pp-mo:
I have also managed now to run _**all**_ Iris tests with the new mechanism replacing pyke + no further problems emerged,
i.e. no existing tests failed when using the non-pyke mechanism.
Looking forward, I think this looks very hopeful for us to entirely remove Pyke within this PR :rocket: !

We should get the new code + tests reviewed as-is, 
Once we're happy that nothing is missing, I should be able to simplify this branch + remove Pyke altogether.

Regarding the future of the new tests, I _was_ intending to provide two TestCase classes for every set of tests, one to run against the old mechanism and one against the new (by setting the class `use_pyke` switch).
However, I realise now that we just don't need that :  Anyone can easily run tests, manually, with either mechanism.  
So, once we have prior confidence in all these tests, we can just switch over + be happy as long as they still pass.  
So the tests can be usefully simplified from what we now have, and I will aim to do that going forward, along with whatever other changes may be needed.

lib/iris/tests/unit/fileformats/netcdf/load_cube/load_cube__activate/__init__.py
```diff
@@ -0,0 +1,244 @@
...
+        do_nonpyke = not self.use_pyke or self.compare_pyke_nonpyke
+        if do_pyke:
+            pyke_engine = iris.fileformats.netcdf._pyke_kb_engine_real()
+        if do_nonpyke:
+            nonpyke_engine = iris.fileformats._nc_load_rules.engine.Engine()
+
+        iris.fileformats.netcdf.DEBUG = self.debug
+
+        # Call the main translation function to load a single cube.
+        def load_single_cube(engine):
```
stephenworsley, line 134: Would this make more sense to be named `load_specific_cube`? It seems like `_load_cube` is already loading a 'single' cube (i.e. the result is just one cube),  but this function if different in that it has already specified which cube it will load.

pp-mo:
> Added actions for formulae (aka hybrid coords, factories). https://github.com/SciTools/iris/pull/4170/commits/3e1575087884cb9e83c85758f30d0de4a735d938

I think this is what was wrong!
Not quite sure how I missed doing that.  
I still need to confirm that this works for existing tests, so this is **maybe just a preview**

pp-mo:
Ok, this time it's real : 
I **_have_** now tested running all Iris tests (with the test-data), and a small hack to send all netcdf loads through the new mechanism.
FYI those temporary hacks are shown in [another draft PR](https://github.com/SciTools/iris/pull/4188)
  -- that shows **_nearly_** everything passing : it just upsets some tests that check the warnings output

lib/iris/fileformats/_nc_load_rules/actions.py
```diff
@@ -0,0 +1,562 @@
...
+    hh.build_auxiliary_coordinate(engine, var)
+
+
+@action_function
+def action_formula_type(engine, formula_root_fact):
+    """Register a CFVariable as a formula root."""
+    rule_name = "fc_formula_type"
+    (var_name,) = formula_root_fact
+    cf_var = engine.cf_var.cf_group[var_name]
+    # var.standard_name is a formula type (or we should never get here).
```
stephenworsley, line 458: I believe this comment should refer to `cf_var.standard_name`.

lib/iris/fileformats/_nc_load_rules/actions.py
```diff
@@ -0,0 +1,562 @@
...
+    return rule_name
+
+
+@action_function
+def action_formula_term(engine, formula_term_fact):
+    """Register a CFVariable as a formula term."""
+    # Must run AFTER formula root identification.
+    (termvar_name, rootvar_name, term_name) = formula_term_fact
+    # The rootname is implicit :  have only one per cube
+    # TODO: change when we adopt cf-1.7 advanced grid-mping syntax
```
stephenworsley, line 487: I'm guessing "grid-mping" should be "grid-mapping"

lib/iris/tests/unit/fileformats/netcdf/load_cube/load_cube__activate/__init__.py
pp-mo: I thin I may just call this 'translate_cube'.  
It is only needed to encapsulate the "call _load_cube and record formula info" operation, as we need to do that twice.

lib/iris/tests/unit/fileformats/netcdf/load_cube/load_cube__activate/__init__.py
```diff
@@ -0,0 +1,244 @@
...
+            nonpyke_cube = load_single_cube(nonpyke_engine)
+
+        # If requested, directly compare the pyke and non-pyke outputs.
+        if self.compare_pyke_nonpyke:
+            # Compare the loaded cubes from both engines.
+            # print("\nPYKE-NONPYKE COMPARE")
+
+            # First zap cube-data, as masked data does not compare well.
+            def unmask_cube(cube):
+                # preserve the original, we're going to realise..
```
stephenworsley, line 164: This comment could be clearer, I can't make out what "we're going to realise" is refering to.
Also, capitalise the first letter for consistency.

lib/iris/fileformats/_nc_load_rules/actions.py
```diff
@@ -0,0 +1,562 @@
...
+    engine.requires.setdefault("formula_terms", {})[term_name] = termvar_name
+    rule_name = f"fc_formula_term({term_name})"
```
stephenworsley, line 489: Should this not be "fc_formula_type" ?

pp-mo:
I've now created a temporary PR to remove pyke entirely + fix all the legacy tests.
#4198 show this running tests against scitools/master, but it includes all the commits already seen here
https://github.com/pp-mo/iris/pull/69 shows what has changed since this PR (until we add more here)

lib/iris/tests/unit/fileformats/netcdf/load_cube/load_cube__activate/test__hybrid_formulae.py
```diff
@@ -0,0 +1,225 @@
...
+        )
+        # Check that it picks up the terms, but *not* the factory root coord,
+        # which is simply discarded.
+        self.check_result(result, factory_type=None, formula_terms=["a", "b"])
+
+
+# Add in tests methods to exercise each (supported) vertical coordinate type
+# individually.
+# NOTE: hh.CF_COORD_VERTICAL lists all the valid types, but we don't yet
+# support all of them.
```
stephenworsley, line 176: Are these types not supported in the code or just in these tests?

pp-mo:
@stephenworsley pointed out ..
> So it seems as though the failure cases (where the rule_name has "FAILED" appended to it) aren't being tested for in your cases. Is the idea that we are more interested in testing the type of files which we expect to actually encounter or is there some other rational here?

+ I replied
> This is a really good point. 
I guess I was overly focussed on showing the equivalence with the "old world", where of course a rule that fails to fire is not logged.  But I think because of that, I have failed to ensure that the 'new world' is properly tested in its' own right.
... I will investigate this + add testcases as needed to cover the possible code-branches of the new mechanism.

pp-mo:
Likewise ... I realised that I have no unit testing of the 'Engine' class :  I originally thought of that as a temporary fix, but I guess I now think it's simpler if that stays in for now. 
So, I have now added some direct tests for this class (pretty basic).

pp-mo:
> it seems as though the failure cases (where the rule_name has "FAILED" appended to it) aren't being tested for in your cases.

I have now used a coverage analysis to check this out.
There are indeed a few missing code branches associated with error cases.
Actually, some of them may be logically unreachable -- problems of defensive programming !

I am working on targetting these lines, though I may instead remove some code, if it can never be invoked for logical reasons.

pp-mo:
Ok, following @stephenworsley suggestions above, I used coverage to identify + fix missed code branches in the rules.
(see most recent commits)

I have now used coverage to confirm that ...
  * the new tests cover all the old Pyke rules routines (i.e. all fc_xxx functions in the built `fc_rules_cf_fc.py`).
  * the same tests, running with the non-pyke mechanism, provide complete coverage for `actions.py`
  * that the new `test_engine` covers all code in `engine.py`

As this branch now has a CI problem, I would like next to rebase it, and also pull in the subsequent improvements + simplifications waiting on https://github.com/pp-mo/iris/pull/69 (or https://github.com/SciTools/iris/pull/4198 in full).

As that means dropping Pyke itself, I will just wait until we've discussed + are happy with the scope coverage here @stephenworsley, since we can't really go back to that discussion afterwards.

It should all get much simpler after that !

lib/iris/tests/unit/fileformats/netcdf/load_cube/load_cube__activate/test__grid_mappings.py
```diff
@@ -846,9 +846,9 @@ def test_aux_lat_rotated(self):
         # 	005 : fc_build_coordinate_latitude_rotated
```
stephenworsley, line 846: Shouldn't this change also affect the rules triggered?

pp-mo: Yes, you're right -- I haven't updated these comments.
But they are only a reference for debug comparison.
I don't think it's worth fixing, as we are going to remove them all when we are done.

stephenworsley: Fair enough.

pp-mo:
Rebased, now checking we are okay with all CI checks.

Next I will be pulling in code from #4198, which removes Pyke and simplifies all the testing, but has not yet got more recent test additions from here.

pp-mo:
Basically, #4198 now replaces this.
But this may be a useful reference for the scope, while it was still testing against both mechanisms

Issue #4134, opened 2021-05-14
Load is VERY slow for a NetCDF multi-variable file


senesis:
## 📰 Custom Issue

When loading a single variable from a quite small NetCDF file which includes 300 variables, the load time is very large : around 100 seconds (while it is less than 0.1s for a similar single variable file). 

This is a bottleneck for trying to use Iris (trough [ESMValTool](https://github.com/ESMValGroup/ESMValTool)) for handling some climate model native data format.

The attached notebook [load_time_histmth.pdf](https://github.com/SciTools/iris/files/6477355/load_time_histmth.pdf) demonstrates the issue and includes a profiling, which shows that the most time consuming function is (by large) [NetCDFDataProxy.__getitem__](https://github.com/SciTools/iris/blob/83309c32a6c9cfbd4603732a4075bae0f6f07d45/lib/iris/fileformats/netcdf.py#L434)

The data file is available [here](https://drive.google.com/file/d/1_M7Hno-FFnkfU88jVMNCHvzyPTZWsbQA/view?usp=sharing)

System info is :
> uname -a 
> Linux ciclad-ng.private.ipsl.fr 2.6.32-754.35.1.el6.x86_64 #1 SMP Wed Oct 7 03:47:54 CDT 2020 x86_64 x86_64 x86_64 GNU/Linux
>

bjlittle:
@senesis Thanks for taking the time to report this, much appreciated.

Could you just confirm the version of `iris` that you're using? I'm assuming `v3.0.1`?

senesis:
Yes it is v3.0.1

bjlittle:
@senesis Great thanks. And which version of Python? v3.8?

bjlittle:
@senesis We have a patch in the pipeline that will go towards significantly alleviating this issue.

We're going to target this for the forthcoming `iris` v3.0.2 release :+1: 

See the [v3.0.x Release Discussion](https://github.com/SciTools/iris/discussions/4071) for further details.

pp-mo:
@senesis great issue -- your account is excellent + should make the problem reproducible !

I was already looking into issues with slow netcdf loads, specifically with lots of variables (i.e. we also found something similar).
I think I may already have a solution that will at least "alleviate" it ... coming shortly I hope.

In the meantime though, I wonder if you could put up your 'load_time_histmth' notebook so I can test out with that, maybe in a Gist ??

pp-mo:
Stop press: see #4135 
[...]

PR #4175, opened 2021-06-03
Simple support for selective netcdf loading.

pp-mo:
### Description
An initial heads-up on a really simple way of speeding up netcdf loads ...
... with files of many variables, as in #4134 

This is actually much simpler than I imagined, but ..
**Blockers to completion**
  * probably needs discussion
  * not sure how to test this

### Background...

Following #4135, ESMValTool devs are reporting that iris loading is [still too slow, where you want only one a whole lot of diagnostics](https://github.com/ESMValGroup/ESMValCore/pull/1153#issuecomment-853223610).

This example follows what we did for PP files in similar circumstances.
It's notable that, in creating a `iris.fileformats.cf.CFReader`, we are still doing a whole-file analysis, that includes the unwanted data-variables.  
I actually don't think you can avoid that, as only context will distinguish a CF data-variable from an aux-coord.
However, the cost of this is not huge.  I am finding <1sec for the testfile mentioned in #4134 (~250mB, 300 variables of content float[1,100]).

#### Some sample timings:
Using testfiles with many identical (small) variables:
**n-vars** : timings without // with fix
    1 :   0.04  //  0.01   [loading 1 of N named variables]
  10 :   0.14  //  0.02
  30 :   0.45  //  0.03
100 :   1.70  //  0.07
300 :   8.19  //  0.40
(**314**) :  44.17  //  0.61

case (**314**) is based on the testfile mentioned in #4314 
( I suspect it may be slower than the '300' because the variables _data_ is larger??  WIP )
with the code like
```
cube = iris.load_cube(
    'Iris_multivar_data_file.nc',
    NameConstraint(long_name='Air Surface Temperature'))
```

---

### DIFF
--- a/lib/iris/fileformats/__init__.py
+++ b/lib/iris/fileformats/__init__.py
@@ -95,7 +95,12 @@ FORMAT_AGENT.add_spec(
 #
 FORMAT_AGENT.add_spec(
     FormatSpecification(
-        "NetCDF", MagicNumber(4), 0x43444601, netcdf.load_cubes, priority=5
+        "NetCDF",
+        MagicNumber(4),
+        0x43444601,
+        netcdf.load_cubes,
+        priority=5,
+        constraint_aware_handler=True,
     )
 )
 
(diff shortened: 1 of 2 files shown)


pp-mo:
Wrong branch..

PR #4176, opened 2021-06-03
Simple support for selective netcdf loading.

pp-mo:
### Description
An initial heads-up on a really simple way of speeding up netcdf loads ...
... with files of many variables, as in #4134 

This is actually much simpler than I imagined, but ..
**Blockers to completion**
  * probably needs discussion
  * intended usage might require to select multiple data-vars
  * not sure how to test this

### Background...

Following #4135, ESMValTool devs are reporting that iris loading is [still too slow, where you want only one from a whole lot of diagnostics](https://github.com/ESMValGroup/ESMValCore/pull/1153#issuecomment-853223610).

This example follows [what we did for UM files in similar circumstances](https://github.com/SciTools/iris/pull/1240).
It's notable that, in creating a `iris.fileformats.cf.CFReader`, we are still doing a whole-file analysis, that includes the unwanted data-variables.  
I actually don't think you can avoid that, as only context will distinguish a CF data-variable from an aux-coord.
However, the cost of this is not huge.  I am finding <1sec for the testfile mentioned in #4134 
( which is : ~250mB, ~300 variables of content float[1 * 79 * 143 * 144] )

#### Some sample timings:
Using testfiles with many identical (small) variables:
**n-vars**  :  **timings without // with fix**
    1 :   0.04  //  0.01   [loading 1 of N named variables]
  10 :   0.14  //  0.02
  30 :   0.45  //  0.03
100 :   1.70  //  0.07
300 :   8.19  //  0.40
(**314**) :  45.36  //  0.59

case (**314**) is based on the testfile mentioned in #4134 
( I suspect it may be slower than the '300' because the variables _data_ is larger??  WIP )
with the code like
```
cube = iris.load_cube(
    'Iris_multivar_data_file.nc',
    NameConstraint(long_name='Air Surface Temperature'))
```

---

### DIFF
--- a/docs/src/whatsnew/latest.rst
+++ b/docs/src/whatsnew/latest.rst
@@ -62,6 +62,10 @@ This document explains the changes made to Iris for this release
    :attr:`~iris.cube.Cube.attributes`, and improving spacing throughout.
    (:pull:`4206`)
 
+#. `@pp-mo`_ and `@lbdreyer`_ optimised loading netcdf files, resulting in a
+   speed up when loading with a single :func:`~iris.NameConstraint`.
+   (:pull:`4176`)
+
 
 🐛 Bugs Fixed
 =============
--- a/docs/src/whatsnew/latest.rst
+++ b/docs/src/whatsnew/latest.rst
@@ -63,7 +63,9 @@ This document explains the changes made to Iris for this release
    (:pull:`4206`)
 
 #. `@pp-mo`_ and `@lbdreyer`_ optimised loading netcdf files, resulting in a
-   speed up when loading with a single :func:`~iris.NameConstraint`.
+   speed up when loading with a single :func:`~iris.NameConstraint`. Note, this
+   optimisation only applies when matching on standard name, long name or
+   NetCDF variable name, not when matching on STASH.
    (:pull:`4176`)
 
 
--- a/lib/iris/fileformats/__init__.py
+++ b/lib/iris/fileformats/__init__.py
@@ -90,7 +90,12 @@ FORMAT_AGENT.add_spec(
 #
 FORMAT_AGENT.add_spec(
     FormatSpecification(
-        "NetCDF", MagicNumber(4), 0x43444601, netcdf.load_cubes, priority=5
+        "NetCDF",
+        MagicNumber(4),
+        0x43444601,
+        netcdf.load_cubes,
+        priority=5,
+        constraint_aware_handler=True,
     )
 )
 
@@ -102,6 +107,7 @@ FORMAT_AGENT.add_spec(
         0x43444602,
         netcdf.load_cubes,
         priority=5,
+        constraint_aware_handler=True,
     )
 )
 
@@ -114,6 +120,7 @@ FORMAT_AGENT.add_spec(
         0x894844460D0A1A0A,
         netcdf.load_cubes,
         priority=5,
+        constraint_aware_handler=True,
     )
 )
 
@@ -124,6 +131,7 @@ _nc_dap = FormatSpecification(
     lambda protocol: protocol in ["http", "https"],
     netcdf.load_cubes,
     priority=6,
+    constraint_aware_handler=True,
 )
 FORMAT_AGENT.add_spec(_nc_dap)
 del _nc_dap
--- a/lib/iris/fileformats/netcdf.py
+++ b/lib/iris/fileformats/netcdf.py
@@ -37,6 +37,7 @@ from iris.aux_factory import (
     OceanSigmaZFactory,
 )
 import iris.config
+import iris._constraints
 import iris.coord_systems
 import iris.coords
 import iris.exceptions
@@ -759,7 +760,46 @@ def _load_aux_factory(engine, cube):
         cube.add_aux_factory(factory)
 
 
-def load_cubes(filenames, callback=None):
+def translate_constraints_to_var_callback(constraints):
+    """
+    Translate load constraints into a simple data-var filter function, if possible.
+
+    Returns:
+         * function(cf_var:CFDataVariable): --> bool,
+            or None.
+
+    For now, ONLY handles a single NameConstraint with no 'STASH' component.
+
+    """
+    constraints = iris._constraints.list_of_constraints(constraints)
+    result = None
+    if len(constraints) == 1:
+        (constraint,) = constraints
+        if (
+            isinstance(constraint, iris._constraints.NameConstraint)
+            and constraint.STASH == "none"
+        ):
+            # As long as it doesn't use a STASH match, then we can treat it as
+            # a testing against name properties of cf_var.
+            # That's just like testing against name properties of a cube, except that they may not all exist.
+            def inner(cf_datavar):
+                match = True
+                for name in constraint._names:
+                    expected = getattr(constraint, name)
+                    if name != "STASH" and expected != "none":
+                        # Fetch property : N.B. CFVariable caches the property values
+                        # The use of a default here is the only difference from the code in NameConstraint.
+                        actual = getattr(cf_datavar, name, "")
+                        if actual != expected:
+                            match = False
+                            break
+                return match
+
+            result = inner
+    return result
+
+
+def load_cubes(filenames, callback=None, constraints=None):
     """
     Loads cubes from a list of NetCDF filenames/URLs.
 
@@ -779,6 +819,9 @@ def load_cubes(filenames, callback=None):
     """
     from iris.io import run_callback
 
+    # Create a low-level data-var filter from the original load constraints, if they are suitable.
+    var_callback = translate_constraints_to_var_callback(constraints)
+
     # Create an actions engine.
     engine = _actions_engine()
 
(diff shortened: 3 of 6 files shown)


lib/iris/fileformats/__init__.py
```diff
@@ -95,7 +95,12 @@ def _load_grib(*args, **kwargs):
 #
 FORMAT_AGENT.add_spec(
     FormatSpecification(
-        "NetCDF", MagicNumber(4), 0x43444601, netcdf.load_cubes, priority=5
+        "NetCDF",
+        MagicNumber(4),
+        0x43444601,
+        netcdf.load_cubes,
+        priority=5,
+        constraint_aware_handler=True,
```
bjlittle, line 103: @pp-mo Nice. I'd totally forgotten about this ancient part of `iris` :tada:

lib/iris/fileformats/netcdf.py
```diff
@@ -38,6 +38,7 @@
     OceanSg2Factory,
 )
 import iris.config
+import iris._constraints
```
bjlittle, line 41: After https://github.com/SciTools/iris/pull/4174 opened my eyes to the wiley world of imports, would you mind moving this module level import to `translate_constraints_to_var_callback` instead please? i.e.,
```python
def translate_constraints_to_var_callback(constraints):
    """
    ...
    """
    import iris._constraints
```

trexfeathers:
I've stared at this for a while now, and it makes sense.

My main question is: what does this look like beyond MVP? There are obviously a lot of limitations and hard-coding in this implementation; how would you make it more generalisable? I.e. multiple constraints, non-`NameConstraint`s, less code duplication? Since we don't know when we might have time to expand the functionality, it would be good to make sure the merged code is something that can be iterated on, rather than something that would need replacing.

You can ignore my question if this is far as we're expecting to take this functionality in the foreseeable future.

pp-mo:
> what does this look like beyond MVP? 

Well, first we need to check that this is addressing the immediate needs of the ESMValTool devs, otherwise it's not worth it.

A simple list of constraints acts with an "OR"-like behaviour, so that is mostly a non-starter for a speedup solution.  
The simplest exception to that is provided for in [the PP implementation](https://github.com/SciTools/iris/blob/master/lib/iris/fileformats/pp.py#L1942-L2008), where we handle a list of constraints *all* of which are of the form `AttributeConstraint(STASH=<selected>)`.  
You _could_ in principle extend that to cases where all the top-level constraints in the list are formed as '&' -combinations (i.e. [ConstraintCombination](https://github.com/SciTools/iris/blob/master/lib/iris/_constraints.py#L180-L216)), and one of those is a type we can "translate".
But in the PP case we never even got that far.

Apart from that, I think in this case we **_can_** work out how to do the equivalent of a Constraint(name=string), allowing that we decode a possible STASH attribute on the variable.  
So I think that's about all that is technically possible + we need anyway to first decide what is actually useful before committing more time.

trexfeathers:
>Well, first we need to check that this is addressing the immediate needs of the ESMValTool devs, otherwise it's not worth it.

Yup, makes sense not to invest time thinking about the future yet. In that case do you think it would be worth including a `note` or `todo` in the docstring to indicate that it's a bit of an experiment? If it's valuable: we'll need to re-write it more generally, if it's not: we'll need to back it out.

pp-mo:
>  move translate_constraints_to_var_callback to the experimental module.

Ok I don't think there's any point doing that directly, as it isn't a public thing (unless I got it wrong + it needs 'hiding'?).

My feeling is that this is (should be) a implementation shortcut guaranteed to have **_no_** effect on correctness I.E. the final result is the same in all cases.  
Equally importantly, we can then provide it with no visible user control.
So if you can keep it like that, then it doesn't really matter which cases it covers and which not,
( except whether it addresses a performance need ).

pp-mo:
Thanks for inputs @trexfeathers @bjlittle 
Will address those points in due course.
But, I'm going to hold fire on this unitl it's clear whether it adresses the ESMValTool devs' need.

pp-mo:
Some further practical discussion at https://github.com/ESMValGroup/ESMValCore/pull/1153#issuecomment-859703186

In [my reply](https://github.com/ESMValGroup/ESMValCore/pull/1153#issuecomment-860562278) , I tried to explain how the existing solution here is only a quick hack.

As-is, it **_only_** supports a constraint expression which is a single NameConstraint.
But as noted above, we already know how to support a list of several of those (done for PP).

We _could_ also support, for example, any &-combinations involving a NameConstraint.  
We can also do AttributeConstraints if wanted.
But **_the approach does not generalise fully_** -- i.e. some types of expression which "could" be translated (+ therefore speeded up) will remain 'opaque' to this type of analysis and it won't be possible to do everything.
Most notably,  anything involving a cube_function is obviously "opaque".
Likewise, in any case where we are given a list of constraints, we need to be able to 'translate' _everything_ on the list, or we can't select on data-variable at all.

lib/iris/fileformats/netcdf.py
```diff
@@ -771,7 +772,46 @@ def coord_from_term(term):
...
+            # That's just like testing against name properties of a cube, except that they may not all exist.
+            def inner(cf_datavar):
+                match = True
+                for name in constraint._names:
+                    expected = getattr(constraint, name)
+                    if name != "STASH" and expected != "none":
+                        # Fetch property : N.B. CFVariable caches the property values
+                        # The use of a default here is the only difference from the code in NameConstraint.
+                        actual = getattr(cf_datavar, name, "")
+                        if actual != expected:
```
pp-mo, line 805: N.B. this still works if 'expected' happened to be a function or a list of values, instead of a simple constant.
Which must be considered as [those are also valid possibilties](https://scitools-iris.readthedocs.io/en/latest/generated/api/iris.html#iris.NameConstraint)

zklaus:
For us, the main use case is to load one variable from a multi-variable file. I think that is really all we need.

lib/iris/fileformats/netcdf.py
pp-mo: :+1: will do

lib/iris/fileformats/netcdf.py
pp-mo: Looking again, no longer convinced this is 'real' problem.  A function will not compare to string, so no problem ??

pp-mo:
Reviewing of outstanding work, as we want to get this into [Iris 3.1.0](https://github.com/SciTools/iris/discussions/4072)

 * respond to outstanding (small) review comment, and any further
 * fix for multiple constraints, as in PP code (_possibly_ combine in common code ??)
 * add whatsnew
 
 I (@pp-mo) can tackle this in ~2 weeks when I return from holiday, 
 or _someone else can do it ?_

lib/iris/fileformats/netcdf.py
```diff
@@ -759,7 +759,49 @@ def coord_from_term(term):
...
+        ):
+            # As long as it doesn't use a STASH match, then we can treat it as
+            # a testing against name properties of cf_var.
+            # That's just like testing against name properties of a cube, except that they may not all exist.
+            def inner(cf_datavar):
+                match = True
+                for name in constraint._names:
+                    expected = getattr(constraint, name)
+                    if name != "STASH" and expected != "none":
+                        attr_name = "cf_name" if name == "var_name" else name
```
lbdreyer, line 791: `var_name` is not available at this point, but `cf_name` is, so we can just use that

lbdreyer:
I have taken this on whilst @pp-mo is on holiday.

Changes I have made:
* Added a whatsnew
* Added unit and integration tests
* Added [support for NameConstraints with a var_name](https://github.com/SciTools/iris/pull/4176#discussion_r685914675)

lbdreyer:
Rebased to resolve merge conflict now that https://github.com/SciTools/iris/pull/4206 has gone in

docs/src/whatsnew/latest.rst
```diff
@@ -62,6 +62,10 @@ This document explains the changes made to Iris for this release
    :attr:`~iris.cube.Cube.attributes`, and improving spacing throughout.
    (:pull:`4206`)
 
+#. `@pp-mo`_ and `@lbdreyer`_ optimised loading netcdf files, resulting in a
+   speed up when loading with a single :func:`~iris.NameConstraint`.
```
bjlittle, line 66: @lbdreyer Should we explicitly advertise that it doesn't handle `STASH`?

lib/iris/fileformats/netcdf.py
```diff
@@ -759,7 +759,49 @@ def coord_from_term(term):
...
+                        # Fetch property : N.B. CFVariable caches the property values
+                        # The use of a default here is the only difference from the code in NameConstraint.
+                        actual = getattr(cf_datavar, attr_name, "")
```
bjlittle, line 794: @lbdreyer Is there an edge case here, where we've set the default `actual` value to be `""` for an `attr_name` that doesn't exist on the `cf_datavar`, but that could match an `expected` of `""`, leaving `match = True` but instead it should actually be `False`.

Is it safer to skip the `actual` and `expected` value comparison if the `cf_datavar` doen't have an `attr_name`? i.e.,

```python
if not hasattr(cf_datavar, attr_name):
    continue
actual = getattr(cf_datavar, attr_name)
...
```
Not sure whether this is relevant here... just a thought :thinking:

bjlittle (changes requested):
@lbdreyer Awesome! Thanks for helping to nudge this PR across the line :beers: 

Just a couple of comments to service, then we're good to go :rocket:

lbdreyer:
@bjlittle Thanks for your review! I believe I have now resolved the review actions you raised, so could you have a final check?

bjlittle (approved):

PR #4181, opened 2021-06-04
drop black and flake8 dependencies

bjlittle:
### Description
Akin to PR #4174 and `isort`, this PR drops the `black` and `flake` package dependencies from both our `conda` YAML and PyPI `setup.cfg` requirements.

We're now advocating the use of `pre-commit` for developers, which will take care of your linting needs. Additionally, developers (and users alike) can easily install these packages explicitly, if they wish.

The reason for dropping these packages is to reduce the `iris` dependency bloat, but also due to the regular cadence in version bumping that is required - hardly a major maintenance overhead, but this has cause recent difficulties, particularly with `black`.

---

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -2,7 +2,7 @@
 # See https://pre-commit.com/hooks.html for more hooks
 repos:
 -   repo: https://github.com/pre-commit/pre-commit-hooks
-    rev: 'v4.0.1'
+    rev: "v4.0.1"
     hooks:
         # Prevent giant files from being committed.
     -   id: check-added-large-files
@@ -17,24 +17,23 @@ repos:
         # Don't commit to main branch.
     -   id: no-commit-to-branch
 -   repo: https://github.com/psf/black
-    rev: '21.6b0'
+    rev: "21.6b0"
     hooks:
     -   id: black
         # Force black to run on whole repo, using settings from pyproject.toml
         pass_filenames: false
-        args: [--config=./pyproject.toml, .]
+        args: ["--config=./pyproject.toml", "--check", "docs", "lib", "noxfile.py", "setup.py"]
 -   repo: https://github.com/PyCQA/flake8
-    rev: '3.9.2'
+    rev: "3.9.2"
     hooks:
-        # Run flake8.
     -   id: flake8
-        args: [--config=./setup.cfg]
+        args: ["--config=./setup.cfg"]
 -   repo: https://github.com/pycqa/isort
-    rev: 5.9.1
+    rev: "5.9.1"
     hooks:
     -   id: isort
         name: isort
-        args: ["--filter-files"]
+        args: ["--filter-files", "--check"]
 -   repo: https://github.com/asottile/blacken-docs
     rev: v1.8.0
     hooks:
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -20,18 +20,18 @@ repos:
     rev: "21.6b0"
     hooks:
     -   id: black
-        # Force black to run on whole repo, using settings from pyproject.toml
         pass_filenames: false
         args: ["--config=./pyproject.toml", "--check", "docs", "lib", "noxfile.py", "setup.py"]
 -   repo: https://github.com/PyCQA/flake8
     rev: "3.9.2"
     hooks:
     -   id: flake8
-        args: ["--config=./setup.cfg"]
+        args: ["--config=./setup.cfg", "docs", "lib", "noxfile.py", "setup.py"]
 -   repo: https://github.com/pycqa/isort
     rev: "5.9.1"
     hooks:
     -   id: isort
+        # Use settings from pyproject.toml
         name: isort
         args: ["--filter-files", "--check"]
 -   repo: https://github.com/asottile/blacken-docs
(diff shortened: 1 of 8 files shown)


bjlittle:
@trexfeathers Cool. I get `nox` to scrape and align with the `pre-commit` versions of `black` and `flake8` (I guess this should also apply to `isort` too)

And I'll add something to the docs :+1:

.pre-commit-config.yaml
```diff
@@ -17,21 +17,20 @@ repos:
...
     -   id: no-commit-to-branch
 -   repo: https://github.com/psf/black
-    rev: '21.6b0'
+    rev: "21.6b0"
     hooks:
     -   id: black
         # Force black to run on whole repo, using settings from pyproject.toml
         pass_filenames: false
-        args: [--config=./pyproject.toml, .]
+        args: ["--config=./pyproject.toml", "--check", "docs", "lib", "noxfile.py", "setup.py"]
```
trexfeathers, line 25: `tools` dir as well?

pyproject.toml
```diff
@@ -53,4 +55,10 @@ extend_skip = [
   "tools",
 ]
 skip_gitignore = "True"
+src_paths = [
+  "docs",
+  "lib",
+  "noxfile.py",
+  "setup.py",
+]
```
trexfeathers, line 63: `tools` dir as well?

.pre-commit-config.yaml
trexfeathers: For consistency this should be transferred to `pyproject.toml`, [I believe using using the `include` key](https://black.readthedocs.io/en/stable/usage_and_configuration/the_basics.html?highlight=include#configuration-via-a-file)?

.pre-commit-config.yaml
```diff
@@ -17,21 +17,20 @@ repos:
...
 -   repo: https://github.com/PyCQA/flake8
-    rev: '3.9.2'
+    rev: "3.9.2"
     hooks:
-        # Run flake8.
     -   id: flake8
-        args: [--config=./setup.cfg]
+        args: ["--config=./setup.cfg"]
```
trexfeathers, line 30: For consistency with `black` and `isort`: we should include a [`filename`](https://flake8.pycqa.org/en/latest/user/options.html#cmdoption-flake8-filename) key in `setup.cfg` to limit what is being tested, as we were doing previously using commands in `noxfile.py`.

trexfeathers (changes requested):
Thanks @bjlittle, I'm glad you think pre-commit is a good pattern. I have some config suggestions.

pyproject.toml
bjlittle: Some of the tools are auto generated code, so we shouldn't be touching those.

.pre-commit-config.yaml
bjlittle: Cool, I'll take a peek

.pre-commit-config.yaml
bjlittle: I'll check that, thanks

noxfile.py
```diff
@@ -213,14 +167,9 @@ def isort(session: nox.sessions.Session):
...
-    # Execute the isort import checker on the package.
-    session.run("isort", "--check", str(PACKAGE))
-    # Execute the isort import checker on the documentation.
-    session.run("isort", "--check", str(ROOT_DIR / "docs"))
-    # Execute the isort import checker on the root Python files.
-    for fname in ROOT_DIR.glob("*.py"):
-        session.run("isort", "--check", str(fname))
+    session.install("pre-commit")
+    # Execute the pre-commit linting tasks.
+    session.run("pre-commit", "run", "--all-files")
```
trexfeathers, line 172:
```suggestion
    args = ["pre-commit", "run", "--all-files"]
    hooks = ["black", "flake8", "isort"]
    for hook in hooks:
        session.run(*args, hook)
    
```

I've just discovered with cf-units that this doesn't work well since it's also running the various non-linting hooks, which leads to a failure post-merge ("don't commit to main branch").

bjlittle: Nice. It's a shame you can't do something like `pre-commit list`... I'll dig :+1:

.pre-commit-config.yaml
bjlittle: hmmm this is slightly frustrating, in that `include` doesn't quite behave the same as `src_paths` from `isort`... so we need to be explicit. `include` is like a filter of files to include from what's found to the call to `black`, rather than specifically only include these files and directories.

.pre-commit-config.yaml
bjlittle: same again, this is as frustrating as `black`, in that `filename` doesn't quite behave the same as `src_paths` from `isort`... so again we need to be explicit.

noxfile.py
bjlittle: Shame. There isn't anything obvious... which I'm surprised by.

.pre-commit-config.yaml
trexfeathers: That is frustrating, yeah.

Given we are forced to use CLI for `black` and `flake8`, I'd prefer that we use the CLI for `isort` too. Makes for less confusion for the user.

noxfile.py
bjlittle: Happy to go with this... and I've updated it to include the `blacken-docs` hook

trexfeathers (approved):

bjlittle:
@trexfeathers Great, thanks! :beers:

PR #4192, opened 2021-06-11
Test Wrangling

rcomer:
### Description

I tried pointing `pytest` at Iris's tests and got 3 errors and 3 failures.  The errors were easy to fix (see comment on `test_merge.py`).  The failures are genuine, and it seems that `test_PartialDateTime` has not been running under `setuptools` or `nox`.  I've added an `__init__.py` file to get those tests running.  Now we need to decide whether `PartialDateTime` needs fixing or its tests need changing.  The failures are:

* `PartialDateTime() == 1` now returns `False` but a `TypeError` is expected.
*  `PartialDateTime() != 1` now returns `True` but a `TypeError` is expected.

As far as I can tell, `PartialDateTime.__eq__` is returning `NotImplemented` in the above.

* `PartialDateTime(month=3, microsecond=2) == cftime.datetime(year=2013, month=3, day=20, second=2)` now returns `False` but `True` is expected.

The microsecond is ignored if `other` doesn't have that attribute.  Did `cftime.datetime` gain a microsecond attribute at some point?

Anyone have an opinion on what the desired behaviour should be for the above cases?

---

### DIFF
--- a/lib/iris/tests/test_merge.py
+++ b/lib/iris/tests/test_merge.py
@@ -25,7 +25,7 @@ import iris.exceptions
 import iris.tests.stock
 
 
-class TestMixin:
+class MergeMixin:
     """
     Mix-in class for attributes & utilities common to these test cases.
 
@@ -55,7 +55,7 @@ class TestMixin:
 
 
 @tests.skip_data
-class TestSingleCube(tests.IrisTest, TestMixin):
+class TestSingleCube(tests.IrisTest, MergeMixin):
     def setUp(self):
         self._data_path = tests.get_data_path(("PP", "globClim1", "theta.pp"))
         self._num_cubes = 1
@@ -63,7 +63,7 @@ class TestSingleCube(tests.IrisTest, TestMixin):
 
 
 @tests.skip_data
-class TestMultiCube(tests.IrisTest, TestMixin):
+class TestMultiCube(tests.IrisTest, MergeMixin):
     def setUp(self):
         self._data_path = tests.get_data_path(
             ("PP", "globClim1", "dec_subset.pp")
--- a/lib/iris/tests/unit/time/test_PartialDateTime.py
+++ b/lib/iris/tests/unit/time/test_PartialDateTime.py
@@ -224,12 +224,12 @@ class Test___eq__(tests.IrisTest, _Test_operator):
         self.expected_value = EQ_EXPECTATIONS
 
     def test_cftime_equal(self):
-        pdt = PartialDateTime(month=3, microsecond=2)
+        pdt = PartialDateTime(month=3, second=2)
         other = cftime.datetime(year=2013, month=3, day=20, second=2)
         self.assertTrue(pdt == other)
 
     def test_cftime_not_equal(self):
-        pdt = PartialDateTime(month=3, microsecond=2)
+        pdt = PartialDateTime(month=3, second=2)
         other = cftime.datetime(year=2013, month=4, day=20, second=2)
         self.assertFalse(pdt == other)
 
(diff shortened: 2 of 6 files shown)


lib/iris/tests/test_merge.py
```diff
@@ -25,7 +25,7 @@
 import iris.tests.stock
 
 
-class TestMixin:
+class MergeMixin:
```
rcomer, line 28: `pytest` tried to run this as a test class, so errored all three tests when it couldn't find `_data_path`.  Removing "Test" from the class name stops `pytest` trying to do that.

rcomer:
Found and added another missing `__init__.py`.

rcomer:
Now `setuptools` and `nox` are running **two more** tests than `pytest`.

rcomer:
Looks like it was the [system tests](https://github.com/SciTools/iris/blob/master/lib/iris/tests/system_test.py) that `pytest` wasn't discovering.  Because the test case names started with `system` instead of `test`.

rcomer:
`pytest`, `setuptools` and `nox` are now all producing the same numbers of passes, fails and skips.  I did some pattern matching on the `pytest` and `setuptools` output which verified that all but 13 of the passing tests are the same.  Where test cases have docstrings, `setuptools` prints the docstring rather then the path, so it's hard to automatically match up.  It seems a pretty good bet to me that these 13 unmatched cases are the same though.  Same can be said for the 22 skips.

Back to the failures described in the OP: I think `PartialDateTime() == 1` probably should produce an error, as it doesn't make sense to constrain a coordinate with a `PartialDateTime` if the coordinate points are not datetime-like.  I don't know how to implement that though - if I've understood, the comparison is defaulting back to [checking whether the object identities are the same](https://docs.python.org/3/reference/expressions.html#value-comparisons).  How would you override that?

rcomer:
>  the comparison is defaulting back to checking whether the object identities are the same. How would you override that?

We previously used the [cmp method](https://github.com/SciTools/iris/blob/bc3b7487951ea00c7bca4e544df5db489fd10d52/lib/iris/time.py#L168) to raise the error.  But that is [not a thing in python3](https://docs.python.org/3/whatsnew/3.0.html#ordering-comparisons).

lib/iris/time.py
```diff
@@ -155,23 +155,21 @@ def __eq__(self, other):
...
             except AttributeError:
-                result = NotImplemented
+                result = other.__eq__(self)
+                if result is NotImplemented:
+                    # Equaliy is undefined between these objects.  We don't
+                    # want Python to fall back to the default `object`
+                    # behaviour (which compares using object IDs), so we raise
+                    # an exception here instead.
+                    fmt = "unable to compare PartialDateTime with {}"
+                    raise TypeError(fmt.format(type(other)))
```
rcomer, line 167: This works but feels like a bit of a hack.  Alternative solutions welcome!

lib/iris/tests/unit/time/test_PartialDateTime.py
```diff
@@ -224,12 +224,12 @@ def setUp(self):
         self.expected_value = EQ_EXPECTATIONS
 
     def test_cftime_equal(self):
-        pdt = PartialDateTime(month=3, microsecond=2)
+        pdt = PartialDateTime(month=3, second=2)
         other = cftime.datetime(year=2013, month=3, day=20, second=2)
         self.assertTrue(pdt == other)
```
rcomer, line 229: This was failing because the equality operator returned `False`.  Possibly `cftime.datetime`s (or their predecessors) did not used to have a microseconds, but they do now.

bjlittle:
@rcomer I'm trying to curate the `v3.0.x` branch into shape for a `v3.0.4` release.

There are a few PRs that need to land in the `v3.0.x` release feature branch, most notably #4222. Once that happens I'd be keen to see this PR target `v3.0.x` instead of `main` and rebased, as I'd be keen to see this PR as part of the `v3.0.4` release.

rcomer:
@bjlittle please see #4249, which targets the v3.0.x branch.

PR #4196, opened 2021-06-15
Clarify interpolate and regrid scheme use

lbdreyer:
Recently we had a user having trouble with regridding as they had not realised that they need to provide an _instance_ of the scheme, rather than just the class.

I.e. they did
`cube.regrid(target_grid, iris.analysis.Linear)`
rather than 
`cube.regrid(target_grid, iris.analysis.Linear())`

From the docs, we don't make it clear that we want an instance rather the class itself. And it's perfectly reasonable for a function to accept a class (and then under the hood, that function instantiates the class when/how it needs to)

I wondered if we should clarify this in the docs, specifically in the API docs (we already have examples of regrid/interpolate in the user guide).

In this PR I have proposed a simple addition for both the regrid and interpolate methods. However I note that the interpolate method includes an example and maybe it would be preferable to, instead, just include an example for the regrid?

### DIFF
--- a/lib/iris/cube.py
+++ b/lib/iris/cube.py
@@ -4450,7 +4450,7 @@ bound=(1994-12-01 00:00:00, 1998-12-01 00:00:00)
             dates or times may optionally be supplied as datetime.datetime or
             cftime.datetime instances.
         * scheme:
-            The type of interpolation to use to interpolate from this
+            An instance of the type of interpolation to use to interpolate from this
             :class:`~iris.cube.Cube` to the given sample points. The
             interpolation schemes currently available in Iris are:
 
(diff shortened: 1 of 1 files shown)


bjlittle:
@lbdreyer Let's align the doc-string for `regrid` to be the same as `interpolate`, that might certainly help.

Also, if you wanted, see how feasible it is to catch the case for both where the user passes in a class instead of an instance, and produces a helpful `TypeValue` traceback. What do you think?

pp-mo:
>  catch the case for both where the user passes in a class instead of an instance, 

Just for my little say on this...  I've long wished we did this, but it **_is_** still potentially confusing.
You can convert a class to an instance call, but you want to pass in the extra args/kwargs.  Not all aggregator classes can be created with no args, so there can be failures.  And **_when_** it fails, it's hard to ensure a sensible message.
So the result is, some cases now "just work", but the ones that don't may be more confusing than they were.
Unless you can see your way round these problems ??

bjlittle (approved):

PR #4198, opened 2021-06-15
Replace pyke nopyke

pp-mo:
## 🚀 Pull Request
Encompasses all of the work in #4170, but further changes now complete the **_removal_** of pyke, switch all loading to 'new' mechanisms, and simplify all the testing that used to perform new/old comparison testing.

I probably should have squashed the commits from #4170 , but instead I squashed what was formerly here to aid rebase/merge.

---


pp-mo:
Sorry, for clarity should be targetting the #4170 branch ... will replace ...

pp-mo:
Reopened because https://github.com/pp-mo/iris/pull/69 won't run tests.

lib/iris/tests/unit/fileformats/nc_load_rules/actions/test__grid_mappings.py
```diff
@@ -830,5 +830,32 @@ def test_aux_lat_rotated(self):
         self.check_result(result, xco_is_aux=True, xco_no_cs=True)
 
 
+class Test__nondimcoords(Mixin__grid_mapping, tests.IrisTest):
```
pp-mo, line 833: Note: this code was moved here unchanged from the former file 
`lib/iris/tests/unit/fileformats/nc_load_rules/actions/test__additional.py`, now removed

pp-mo:
Rebased, to see how it goes

pp-mo:
re-merged with auto commit from pre-commit.ci, and added small change to run local pre-commit process

? This seems to have been necessary to satisfy CI pre-commit  process after rebasing, as it won't apply automatic Black changes.
Or maybe it's because the black version updated ??

trexfeathers:
@pp-mo from what I can see it was isort changes that needed to be applied.

pp-mo:
> @trexfeathers from what I can see it was isort changes that needed to be applied.

Actually, the failing 'pre-commit' **_initially_** showed a merge-check failure.
But now it seems to be showing something different -- as you can see by navigating from the red cross.
Maybe it was due to change and I looked at it too soon (it had not finished all other tests).

I don't really understand why we seem to be doing this twice : once in the "pre-commit.ci" test, and once in the "black, flake8 and isort".  I thought one test had made the other obsolete ?
**Can you explain what this actually means @trexfeathers ?**

trexfeathers:
>I don't really understand why we seem to be doing this twice : once in the "pre-commit.ci" test, and once in the "black, flake8 and isort". I thought one test had made the other obsolete ?
>**Can you explain what this actually means @trexfeathers ?**

In short: no 😄. I actually raised the same question with @bjlittle earlier today.

bjlittle:
I'll take a look at this... the only advantage of doing it locally is that the round-trip time to detecting a failure and fixing it is shorter i.e., on your `git commit` you find a problem, rather than committing, pushing and waiting for the `pre-commit.ci` bot to barf

pp-mo:
Rebased.
next intending to merge these changes back into #4170 
See [comment there](https://github.com/SciTools/iris/pull/4170#issuecomment-866818189)

pp-mo:
> making developer git commits painful

nothing particularly to do with this PR.
But a key problem now is that nothing happens automatically when you rebase - e.g. [latest on #4170](https://results.pre-commit.ci/run/github/5312648/1624453411.1JvygVFiT9a0sYO5E9UnCw)
Which is one time I really **_did_** prefer to let the CI do the work.

pp-mo:
I have now merged this with #4170 
I had to squash the previous commits here, but it was not that complicated.
A certain amount of manual resolution was needed, in view of subsequent changes and additions on #4170 .
This should now work, but we'll see if tests pass ...

pp-mo:
Ooh. That's nasty.
The recently-added testcase for what happens if you have a variable with two hybrid coords is failing under py37 and ok under py38.
But it turns out there is an indeterminacy in the order in which it processes the coords, presumably down to the use of sets in `cf.py`.
So for me, it literally **_succeeds on some runs and fails on others_** ...
```
$ python -m unittest discover ./lib/iris/tests/unit/fileformats/nc_load_rules
..........................................F............................................................................................................./home/h05/itpp/git/iris/iris_main/lib/iris/fileformats/_nc_load_rules/helpers.py:643: UserWarning: Ignoring netCDF variable 'sound_frequency' invalid units '♫'
  warnings.warn(msg)
.....................................
FAIL: test_two_formulae (actions.test__hybrid_formulae.Test__formulae_tests)
  File "/net/home/h05/itpp/git/iris/iris_main/lib/iris/tests/unit/fileformats/nc_load_rules/actions/test__hybrid_formulae.py", line 239, in test_two_formulae
    formula_terms=["a", "b", "depth", "eta", "orog", "sigma"],
  File "/net/home/h05/itpp/git/iris/iris_main/lib/iris/tests/unit/fileformats/nc_load_rules/actions/test__hybrid_formulae.py", line 117, in check_result
    self.assertEqual(cube._formula_type_name, factory_type)
AssertionError: 'atmosphere_hybrid_height_coordinate' != 'ocean_sigma_coordinate'
- atmosphere_hybrid_height_coordinate
+ ocean_sigma_coordinate


Ran 189 tests in 2.954s

FAILED (failures=1)

$ python -m unittest discover ./lib/iris/tests/unit/fileformats/nc_load_rules
......................................................................................................................................................../home/h05/itpp/git/iris/iris_main/lib/iris/fileformats/_nc_load_rules/helpers.py:643: UserWarning: Ignoring netCDF variable 'sound_frequency' invalid units '♫'
  warnings.warn(msg)
.....................................
Ran 189 tests in 2.481s

OK
$ 
```

Ouch :scream:

pp-mo:
Nearly there. I'm just scanning what is introduced, and correcting some comments.

lib/iris/fileformats/netcdf.py
```diff
@@ -455,85 +429,78 @@ def __setstate__(self, state):
...
     for cf_name in cf_group.labels.keys():
-        engine.add_case_specific_fact(_PYKE_FACT_BASE, "label", (cf_name,))
+        engine.add_case_specific_fact("label", (cf_name,))
 
     # Assert facts for CF formula terms associated with the cf_group
     # of the CF data variable.
-    formula_root = set()
+
+    # Collect varnames of formula-root variables as we go.
+    # NOTE: use dictionary keys as an 'OrderedDict'
```
pp-mo, line 467:
```suggestion
    # NOTE: use dictionary keys as an 'OrderedSet'
```

The 'OrderedDict' is a real thing.
I **_meant_** to say 'OrderedSet' here, which is a missing concept.

lib/iris/tests/unit/fileformats/nc_load_rules/actions/__init__.py
```diff
@@ -0,0 +1,159 @@
...
+    checking.
+    Both '_make_testcase_cdl' and 'check_result' are not defined here :  They
+    are to be variously implemented by the inheritors.
+
+    """
+
+    # "global" test setting : whether to output various debug info
+    debug = False
+
+    # whether to perform action in both ways and compare results.
```
pp-mo, line 65:
```suggestion
```

Leftover : mistake !

lib/iris/tests/unit/fileformats/nc_load_rules/actions/__init__.py
```diff
@@ -0,0 +1,158 @@
...
+    are to be variously implemented by the inheritors.
+
+    """
+
+    # "global" test setting : whether to output various debug info
+    debug = False
+
+    @classmethod
+    def setUpClass(cls):
+        # # Control which testing method we are applying.
```
pp-mo, line 67:
```suggestion
```

Another obsolete comment line!

stephenworsley:
I suspect that the test currently failing is due to a failure elsewhere. The warning message causing a failure seems to be raised due to garbage collection of an Animation object (probably created during this test https://github.com/SciTools/iris/blob/9268ca9c70bd160c35c5099025d632ae33faf858/lib/iris/tests/experimental/test_animate.py#L54). I would imagine that the error can be considered unrelated and would go away after respinning the test, though it may be worth raising another issue to investigate the source.

lib/iris/tests/unit/fileformats/nc_load_rules/actions/__init__.py
```diff
@@ -0,0 +1,157 @@
...
+        # Simulate the inner part of the file reading process.
+        cf = CFReader(nc_path)
+        # Grab a data variable : FOR NOW always grab the 'phenom' variable.
+        cf_var = cf.cf_group.data_variables["phenom"]
+
+        engine = iris.fileformats.netcdf._actions_engine()
+
+        # If debug enabled, switch on the activation summary debug output.
+        # Use 'patch' so it is restored after the test.
+        self.patch("iris.fileformats.netcdf.DEBUG", self.debug)
```
stephenworsley, line 99: Where do we get `self.patch` from? Shouldn't the class have to derive from `tests.IrisTest` for this to work?

pp-mo: Yes : the 'mixin' is only finally actually called from a 'test' class, which **_does_** inherit 'IrisTest'.
So, it's the same inheritance which provides all the 'assertXXX' methods.
(e.g. [here](https://github.com/SciTools/iris/blob/97ca13a532340a540e55cae701b985d8f758241a/lib/iris/tests/unit/fileformats/nc_load_rules/actions/test__hybrid_formulae.py#L21).
You don't see that in this file, because it is all in the inheritors.

lib/iris/fileformats/netcdf.py
```diff
@@ -455,85 +429,78 @@ def __setstate__(self, state):
...
+    #   - see: https://stackoverflow.com/a/53657523/2615050
```
stephenworsley, line 468: Note: this trick only works in python 3.7 and up. We can use this because we are no longer supporting 3.6.

pp-mo:
Rebased again, but I had to force it to re-include the `requirements/ci/nox.lock` directory for some reason.  So the history may not  check out at all points now, but I think the result will be OK.

lib/iris/fileformats/netcdf.py
pp-mo: :+1:

pp-mo:
> Rebased again

I foolishly failed to include the small comment changes I made here in GitHub.
I will re-include those in a new commit, coming shortly ...

lib/iris/fileformats/_nc_load_rules/__init__.py
```diff
@@ -0,0 +1,16 @@
...
+# licensing details.
+"""
+Support for cube-specific CF-to-Iris translation operations.
+
+Interprets CF concepts identified by :mod:`iris.fileformats.cf` to add
+components into loaded cubes.
+
+For now : the API mimics :class:`pyke.knowledge_engine.engine`.
+As this is aiming to replace the old Pyke-based logic rules.
+TODO: simplify once the parallel operation with Pyke is no  longer required.
```
stephenworsley, line 14: Can we remove this line now that we've removed pyke?

pp-mo: Yes, will do 👍

lib/iris/fileformats/_nc_load_rules/actions.py
```diff
@@ -0,0 +1,557 @@
...
+    grid_mapping_facts = engine.fact_list("grid_mapping")
+    # For now, there should be at most *one* of these.
+    assert len(grid_mapping_facts) in (0, 1)
+    for grid_mapping_fact in grid_mapping_facts:
+        action_provides_grid_mapping(engine, grid_mapping_fact)
+
+    # identify + record aka "PROVIDE" specific named coordinates
+    # N.B. cf.py has identified that these are dim-coords, NOT aux-coords
+    # (which are recorded separately).
+    # TODO: can probably remove this step ??
```
stephenworsley, line 512: I'm assuming that the appropriate time to remove this step would be when simplifying the new behaviour as opposed to now when we are simply copying the behaviour.

pp-mo: Yes, I think so.
It is certainly now _possible_ to further simplify the operations, 
and especially to combine the 'provides-coordinate' and 'build-dimension-coordinate' routines.
But for the present I think it remains simpler+safer  to leave this all as it is, in case it may help us debug any unexpected user-reported cases where the new loading does not match the old.
While the overhead of understanding this code might still be a bit high, the practical inefficiency of it should be negligible.

lib/iris/fileformats/netcdf.py
```diff
@@ -581,29 +548,50 @@ def _get_cf_var_data(cf_var, filename):
...
     engine.reset()
 
-    # Initialise pyke engine rule processing hooks.
+    # Initialise engine rule processing hooks.
     engine.cf_var = cf_var
     engine.cube = cube
     engine.cube_parts = {}
     engine.requires = {}
-    engine.rule_triggered = set()
+    engine.rule_triggered = OrderedAddableList()  # set()
```
stephenworsley, line 583: This comment could be clearer, maybe "formerly set()".

lib/iris/tests/unit/fileformats/nc_load_rules/actions/__init__.py
```diff
@@ -0,0 +1,157 @@
...
+from iris.fileformats.netcdf import _load_cube
+
+"""
+Notes on testing method.
+
+IN cf : "def _load_cube(engine, cf, cf_var, filename)"
+WHERE:
+  - engine is a :class:`iris.fileformats._nc_load_rules.engine.Engine`
+  - cf is a CFReader
+  - cf_var is a CFDAtaVariable
```
stephenworsley, line 30: I think CFDAtaVariable should be CFDataVariable here.

lib/iris/fileformats/netcdf.py
pp-mo: :+1: 
I think it is explained in the definition of OrderedAddableList, 
so can I just remove the comment ?  I think it is probably just an editing leftover (reminder to self).

lib/iris/tests/unit/fileformats/nc_load_rules/actions/__init__.py
pp-mo: :+1:

lib/iris/tests/unit/fileformats/nc_load_rules/actions/test__grid_mappings.py
```diff
@@ -0,0 +1,884 @@
...
+        # Check with bad latitude units : 'degrees' in place of 'degrees_north'.
+        #
+        # Rules Triggered:
+        #     001 : fc_default
+        #     002 : fc_provides_grid_mapping_(latitude_longitude)
+        #     003 : fc_default_coordinate_(provide-phase)
+        #     004 : fc_provides_coordinate_(longitude)
+        #     005 : fc_build_coordinate_(miscellaneous)
+        #     006 : fc_build_coordinate_(longitude)
+        # Notes:
```
stephenworsley, line 408: Is there some logic to the way these notes are formatted? There seems to be some inconsistency with the indenting. Perhaps this is a bit nitpicky, but I think with the use of "=" as what I read as some kind of shorthand, keeping the formatting more consistent may improve readability. Also, what exactly does "=" mean here?

lib/iris/tests/unit/fileformats/nc_load_rules/actions/test__grid_mappings.py
```diff
@@ -0,0 +1,884 @@
...
+        # Rules Triggered:
+        #     001 : fc_default
+        #     002 : fc_provides_grid_mapping_(stereographic) --(FAILED check has_supported_stereographic_parameters)
+        #     003 : fc_provides_coordinate_(projection_y)
+        #     004 : fc_provides_coordinate_(projection_x)
+        #     005 : fc_build_coordinate_(projection_y)(FAILED projected coord with non-projected cs)
+        #     006 : fc_build_coordinate_(projection_x)(FAILED projected coord with non-projected cs)
+        # Notes:
+        #     as for 'mercator__fail_unsupported', above
+        #     = NO dim-coords built (cube has no coords)
```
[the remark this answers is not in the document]
pp-mo: In this case, it just meant "i.e.", or "by which I mean.."
I will try to make these clearer

stephenworsley:
Shouldn't this line be removed from setup.cfg? https://github.com/SciTools/iris/blob/af13e14e39a9851e50d54fa6835054357f00f42d/setup.cfg#L126

stephenworsley:
Also, shouldn't https://github.com/SciTools/iris/blob/9a064854b8b65aaa3a9f61944c9d80bf0d2c99ca/pyproject.toml#L35-L36 also be removed?

pp-mo:
> Shouldn't this line [[iris/setup.cfg:: 26]] be removed from setup.cfg?
> Also, shouldn't .. [[iris/pyproject.toml:: 35-36]] also be removed?

:+1:  Whoops.

I think missed these, because they don't mention Pyke directly.
I'll do another search + fix those things...

lib/iris/tests/unit/fileformats/nc_load_rules/actions/test__grid_mappings.py
pp-mo: Ok I will review all of these and try to make them consistent + clearer.

pp-mo:
Thanks @stephenworsley .
Hope I have now addressed all the outstanding comments.
Please re-review.

lib/iris/tests/unit/fileformats/nc_load_rules/actions/test__time_coords.py
```diff
@@ -359,31 +364,17 @@ def test_aux_not_in_phenom_coords(self):
...
-        # specific coord type, it makes no practical difference.
-        result = self.run_testcase(
-            coord_dim_name="dim_renamed",
-            dimname="dim_renamed",
-            in_phenomvar_coords=True,
-            units="1",
-        )
-        self.check_result(result, "aux")
-
-    def test_aux_no_coordsref(self):
```
pp-mo: NOTE: removed this testcase, as it seems to be a mistaken leftover -- identical to the previous one. 
From the name, the intended testcase was probably the same as the existing `test_aux_not_in_phenom_coords`.

lib/iris/tests/unit/fileformats/nc_load_rules/actions/test__time_coords.py
```diff
@@ -359,31 +364,17 @@ def test_aux_not_in_phenom_coords(self):
...
-        # The coord variable is identified as a CFAuxiliaryCoordinate by cf.py,
-        # but having the wrong units causes it to fail the 'is_time' or
-        # 'is_period' test.
+        # We provide a non-dimension coord variable, identified as a
+        # CFAuxiliaryCoordinate by cf.py, but we also give it the wrong units,
+        # which causes it to fail both 'is_time' and 'is_period' tests, so the
+        # 'provides_coord' rule fails to trigger.
```
stephenworsley, line 370: Is it still true that the 'provides_cord' rule fails to trigger? My understanging is that this used to be the case, but we now handle everything with `action_provides_coordinate` which now "triggers" in a similar way to the other cases.

Also, I'm not sure how clear/accurate it is to describe the units as "wrong", it seems to me that they are simply generic and do not trigger a special rule, unless they contradict some other part of the file that I'm not aware of

pp-mo: :+1: yes. good spot.  My logic is wrong, or at least the explanation of how it translates is incorrect.
Will fix !

lib/iris/tests/unit/fileformats/nc_load_rules/actions/test__grid_mappings.py
```diff
@@ -0,0 +1,905 @@
...
+        # Rules Triggered:
+        #     001 : fc_default
+        #     002 : fc_provides_grid_mapping_(albers_conical_equal_area)
+        #     003 : fc_provides_coordinate_(rotated_latitude)
+        #     004 : fc_provides_coordinate_(rotated_longitude)
+        #     005 : fc_build_coordinate_(rotated_latitude)(rotated no-cs : discarded projected cs)
+        #     006 : fc_build_coordinate_(rotated_longitude)(rotated no-cs : discarded projected cs)
+        # Notes:
+        #     * coords built : rotated-lat + lon, with no coord-system (see above)
+        # TODO: should this change ??
```
stephenworsley, line 697: This TODO: could be made clearer.

lib/iris/tests/unit/fileformats/nc_load_rules/actions/test__time_coords.py
```diff
@@ -0,0 +1,459 @@
...
+        #     003 : fc_build_coordinate_(time[[_period]])
+        msg = "Failed to create.* dimension coordinate"
+        result = self.run_testcase(values_all_zero=True, warning=msg)
+        self.check_result(result, "aux")
+
+    def test_dim_fails_typeident(self):
+        # The coord variable is identified as a CFDimensionCoordinate by cf.py,
+        # but having the wrong units causes it to fail the 'is_time' or
+        # 'is_period' test, so the 'provides_coord' rule fails to trigger.
+        # So it is built as a 'miscellaneous' dim-coord.
```
stephenworsley, line 323: This comment should probably be updated also to match the changes in `test_aux_fails_typeident`.

pp-mo:
Some further small changes, as requested.
Thanks for your conscientious application + patience  @stephenworsley !

stephenworsley (approved):
I think this looks good now, all my comments have been addressed and I'd be happy with merging this now.

PR #4207, opened 2021-06-23
Update cf_units pin

lbdreyer:
Related to [this change](https://github.com/conda-forge/iris-feedstock/pull/67) to the iris conda-forge recipe.

I have updated the pin for cf_units. Note there is no change to setup.cfg as that only exists on the main branch.

I had considered waiting for the mergeback then doing it on main, but as we are now planning a 3.0.3 release it makes sense to just put it here

### DIFF
--- a/requirements/ci/py36.yml
+++ b/requirements/ci/py36.yml
@@ -12,7 +12,7 @@ dependencies:
 
 # Core dependencies.
   - cartopy>=0.18
-  - cf-units>=2
+  - cf-units>=2.1.5
   - cftime<1.3.0
   - dask>=2
   - matplotlib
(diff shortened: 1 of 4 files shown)


bjlittle:
@lbdreyer We might hang fire on this PR, and wait for the latest release of `cf-units`, which means that the minimum pin is on `cftime>=1.5`

See https://github.com/SciTools/nc-time-axis/issues/71

bjlittle (changes requested):
@lbdreyer We need to wait to make the minimum pin `cftime>=1.5`

bjlittle:
Replaced by #4222

lbdreyer:
Closed, superceded by #4222

PR #4208, opened 2021-06-23
optimise pre-commit

bjlittle:
### Description
This PR plays to the strengths of `pre-commit` and filters files at the right level before running the hooks.

This configuration works for local developer commits on only the files that have changed (optimisation), and for `cirrus-ci` when `pre-commit` hooks are run against all files. In addition, the behaviour of the tools which the hooks call still honour their specific configuration.

This undoes a major regression introduced in #4181. Oops :blush: 

---

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -1,8 +1,18 @@
 # See https://pre-commit.com for more information
 # See https://pre-commit.com/hooks.html for more hooks
+
+files: |
+    (?x)(
+        noxfile\.py|
+        setup\.py|
+        docs\/.+\.py|
+        lib\/.+\.py
+    )
+minimum_pre_commit_version: 1.21.0
+
 repos:
 -   repo: https://github.com/pre-commit/pre-commit-hooks
-    rev: "v4.0.1"
+    rev: v4.0.1
     hooks:
         # Prevent giant files from being committed.
     -   id: check-added-large-files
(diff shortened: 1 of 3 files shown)


pyproject.toml
```diff
@@ -13,34 +13,18 @@ build-backend = "setuptools.build_meta"
 line-length = 79
 target-version = ['py37', 'py38']
 include = '\.pyi?$'
-exclude = '''
```
pp-mo, line 16: Unfortunately, whereas the pre-commit part makes sense to me,
I just don't feel anyway competent to judge on this bit.

Could we have avoided changes to pyproject.toml + done those separately (cos then I might merge) ?

.pre-commit-config.yaml
```diff
@@ -1,8 +1,18 @@
 # See https://pre-commit.com for more information
 # See https://pre-commit.com/hooks.html for more hooks
+
+files: |
```
pp-mo, line 4: I guess this is the key part, but I'm shaky what this actually means ...
Does it somehow get "passed down" to configure the scope of the individual checking tasks,
or does it just configure what triggers the checks ?

If the latter, then I think we need more context-limitation.
I am finding that the presence of a .nox directory makes black take forever, so some excludes are needed.

pyproject.toml
bjlittle: This part is actually redundant.

Note the change in directive from `exclude` to `extend-exclude`. We're now getting a lot for free, whereas before we were being unnecessarily explicit.

.pre-commit-config.yaml
bjlittle: This took me a while to figure out from reading the `pre-commit` docs and sifting through their banked issues.

But essentially, yes, now `pre-commit` can filter files globally prior passing them to hooks. This is exactly the level of control that we need i.e., filter the changed files that we care about, rather than checking everything.

`black` is slightly different, but I've fixed the `.nox` issue by using `extend-exclude` rather than `exclude` in the `pyproject.toml` file... and `black` by default ignores `.nox` directories. Win.

pp-mo (approved):
Ok, I think  I get this now.
And I can confirm that the speedup seems to be working for me, including ignoring the nox directories, which before it was also not doing.

trexfeathers:
Nice work @bjlittle !

PR #4210, opened 2021-06-24
Only run docs-building sessions with Python 3.8.

pp-mo:
### Make sessions that build docs (doctest, gallery, linkcheck) **_only_** support Python 3.8.

Hitherto we offered 'docstest-3.7' and 'docstest-3.8'.  We could still do that, but I'm choosing not to.
(since Py3.7 gives [different results in some cases](https://github.com/SciTools/iris/pull/4206#issuecomment-867465560))
Likewise, if nox is called specifying a specific _unsuitable_ python, via the PY_VER environment variable, we could still run the session, with Python3.8, but instead I chose not to allow it all.

So..
  1. `nox -s doctest` now runs with Python 3.8
     * whereas previously it ran "doctest-3.7", giving different results
  1. `PY_VER=3.8 nox -s doctest` still works
    * this is how cirrus invokes it
  2. `PY_VER=3.7 nox -s doctest` now gives an error
     * previously available

New behaviours demonstrated :
```
$ nox -l
Sessions defined in /net/home/h05/itpp/git/iris/iris_main/noxfile.py:

* lint -> Perform pre-commit linting of iris codebase.
* tests-3.7 -> Perform iris system, integration and unit tests.
* tests-3.8 -> Perform iris system, integration and unit tests.
* gallery -> Perform iris gallery doc-tests.
* doctest -> Perform iris doc-tests.
* linkcheck -> Perform iris doc link check.

sessions marked with * are selected, sessions marked with - are skipped.
```
```
$ PY_VER=3.8 nox -l
Sessions defined in /net/home/h05/itpp/git/iris/iris_main/noxfile.py:

* lint -> Perform pre-commit linting of iris codebase.
* tests -> Perform iris system, integration and unit tests.
* gallery -> Perform iris gallery doc-tests.
* doctest -> Perform iris doc-tests.
* linkcheck -> Perform iris doc link check.

sessions marked with * are selected, sessions marked with - are skipped.
```
```
$ PY_VER=3.7 nox -l
Sessions defined in /net/home/h05/itpp/git/iris/iris_main/noxfile.py:

* lint -> Perform pre-commit linting of iris codebase.
* tests -> Perform iris system, integration and unit tests.

sessions marked with * are selected, sessions marked with - are skipped.
```
```
$ nox -s doctest
nox > Running session doctest
nox > Creating conda env in .nox/doctest with python=3.8
 . . .
```
```
$ PY_VER=3.7 nox -s doctest
nox > Error while collecting sessions.
nox > Sessions not found: doctest
$ 
```

### DIFF
--- a/noxfile.py
+++ b/noxfile.py
@@ -15,8 +15,15 @@ from nox.logger import logger
 #: Default to reusing any pre-existing nox environments.
 nox.options.reuse_existing_virtualenvs = True
 
+#: Python versions we can run sessions under
+_PY_VERSIONS_ALL = ["3.7", "3.8"]
+_PY_VERSION_LATEST = _PY_VERSIONS_ALL[-1]
+
+#: One specific python version for docs builds
+_PY_VERSION_DOCSBUILD = _PY_VERSION_LATEST
+
 #: Cirrus-CI environment variable hook.
-PY_VER = os.environ.get("PY_VER", ["3.7", "3.8"])
+PY_VER = os.environ.get("PY_VER", _PY_VERSIONS_ALL)
 
 #: Default cartopy cache directory.
 CARTOPY_CACHE_DIR = os.environ.get("HOME") / Path(".local/share/cartopy")
@@ -197,7 +204,7 @@ def tests(session: nox.sessions.Session):
     )
 
 
-@nox.session(python=PY_VER, venv_backend="conda")
+@nox.session(python=_PY_VERSION_DOCSBUILD, venv_backend="conda")
 def doctest(session: nox.sessions.Session):
     """
     Perform iris doctests and gallery.
@@ -231,7 +238,7 @@ def doctest(session: nox.sessions.Session):
     )
 
 
-@nox.session(python=PY_VER, venv_backend="conda")
+@nox.session(python=_PY_VERSION_DOCSBUILD, venv_backend="conda")
 def linkcheck(session: nox.sessions.Session):
     """
     Perform iris doc link check.


noxfile.py
```diff
@@ -197,76 +204,78 @@ def tests(session: nox.sessions.Session):
     )
 
 
-@nox.session(python=PY_VER, venv_backend="conda")
-def gallery(session: nox.sessions.Session):
-    """
-    Perform iris gallery doc-tests.
+if PY_VER == _PY_VERSION_DOCSBUILD or _PY_VERSION_DOCSBUILD in PY_VER:
```
trexfeathers, line 207: I can see the difference that this line makes, but I can't see that it actually _gives_ us anything. We have enough security in the `@nox.session(python=` kwarg that Nox will only run the given sessions with the correct python version. 

Indeed your changes mean we can instead drop the `PY_VER` setting from the relevant Cirrus runs.

trexfeathers (changes requested):
Good idea @pp-mo, but I think we can reduce the complexity slightly.

noxfile.py
pp-mo: It's only for local runs :  because **_if_** you say "PY_VER=3.7 nox -s doctests", it will run under 3.8, which seems unexpected and presumably not what was specifically intended.

trexfeathers: >because if you set "PY_VER=3.7" and run "nox -s doctests", it will run under 3.8 regardless

This is no longer true thanks to you using `python=_PY_VERSION_DOCSBUILD` in the relevant decorators.

pp-mo: > thanks to you using python=_PY_VERSION_DOCSBUILD in the relevant decorators.

? but [as currently configured](https://github.com/SciTools/iris/blob/28d1472417ea05bd5015a6f66e3e35085c1ad9ad/noxfile.py#L19-L23), _PY_VERSION_DOCSBUILD  is _always_ == 3.8

So surely it _does_ run under 3.8, if it runs at all.
( The aim here was to stop it running at all, if asked for 3.7 )

trexfeathers: > So surely it does run under 3.8, if it runs at all.

Yes, precisely. But that is already achieved _without_ generating the relevant sessions within a conditional block. IMO the extra convenience of the conditional block isn't worth the extra complexity / potential confusion.

pp-mo: Ok, I'll simplify.

pp-mo:
Simplified code as review requested.
But now need to rebase.

pp-mo:
Rebased.
 @trexfeathers can you please re-review ?

trexfeathers (approved):

trexfeathers:
Thanks @pp-mo, LGTM!

PR #4218, opened 2021-07-01
drop cirrus-ci minimal tests

bjlittle:
### Description
This PR consolidates the `minimal` and `full` testing for `cirrus-ci` into one task that runs the `full` tests. The `full` tests are fully inclusive of the `minimal` tests.

Running both `minimal` and `full` test tasks (for all Python distributions) is a waste of time, resource and compute credits. 

---
[Consult Iris pull request check list]( https://scitools-iris.readthedocs.io/en/latest/developers_guide/contributing_pull_request_checklist.html)

PR #4219, opened 2021-07-01
consolidate cirrus-ci documentation tasks

bjlittle:
### Description
This PR consolidates the `cirrus-ci` documentation gallery task and documentation testing task into one task, with one associated `nox` session.

---

### DIFF
--- a/noxfile.py
+++ b/noxfile.py
@@ -200,7 +200,7 @@ def tests(session: nox.sessions.Session):
 @nox.session(python=PY_VER, venv_backend="conda")
 def doctest(session: nox.sessions.Session):
     """
-    Perform iris gallery and doc-tests.
+    Perform iris doctests and gallery.
 
     Parameters
     ----------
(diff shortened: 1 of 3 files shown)


bjlittle:
@tkknight I'll add a `whatsnew` to cover this PR and #4218

noxfile.py
```diff
@@ -198,9 +198,9 @@ def tests(session: nox.sessions.Session):
 
 
 @nox.session(python=PY_VER, venv_backend="conda")
-def gallery(session: nox.sessions.Session):
+def doctest(session: nox.sessions.Session):
     """
-    Perform iris gallery doc-tests.
+    Perform iris gallery and doc-tests.
```
pp-mo, line 203: Would it not make more sense to run the gallery tests *after* the docs-build ?
(or does that leave a mess behind?)

pp-mo (changes requested):
Otherwise seems a useful simplification.

noxfile.py
bjlittle: @pp-mo The doctests require to change directory but the gallery doesn't, so I erred on the sided of doing the gallery first, then change directory... I'll play and see, but yeah perhaps the doctests are more likely to fail, so it might be more economical to do them first :+1:

tkknight (approved):

pp-mo (approved):

bjlittle:
Awesome, thanks @tkknight and @pp-mo :rocket:

PR #1641, opened 2015-04-22
numerical tolerance

marqh:
There is a bug in the cube intersection code where 
```
np.where(pre_wrap_delta != post_wrap_delta)
```
can return a result due to numerical tolerances in the floating point coordinate value

I have trapped this and added a test.  This is a reported bug against 1.8.0, so please may it be targeted at the 1.8.1 milestone?

thank you

PR #4220, opened 2021-07-02
In cube.intersection, find split cells using a tolerant equality check

lbdreyer:
When performing a cube.intersection, Iris sometimes has to wrap the coord bounds. As an example when the coord on the original cube is between (0,360), but someone requests an intersection of (-60, 60). This can result in split cells that straddle the wrap point. For example, your original bounds may be:
```
array([[299.7, 299.8],
       [299.8, 299.9],
       [299.9, 300. ],
       [300. , 300.1],
       [300.1, 300.2]])
```
But your wrapped bounds, with wrapping point of -60, will look like:
```
array([[299.7, 299.8],
       [299.8, 299.9],
       [299.9, -60. ], # straddle point!
       [-60. , -59.9],
       [-59.9, -59.8]])
```

Iris calculates the diff between the upper and lower bounds, for bounds *inside* the requested intersection, resulting in the variables `pre_wrap_delta` and `post_wrap_delta`, and then it compares these to find any straddle points.
If there are any differences, it will then use these to determine where the new wrapping point (which avoids the straddle points) should be.

Notably, Iris first checks whether there are *any* diffs between `pre_wrap_delta` and `post_wrap_delta` with `np.allclose()`. If there are differences, it then determines where they are. This `np.allclose` check was introduced in https://github.com/SciTools/iris/pull/1641 as we had found previous issues with numerical tolerances (`post_wrap_delta` was almost like `pre_wrap_delta`). 
However, if `np.allclose` returns False as there is a definite straddle point, in order to find where the straddle points are Iris uses  `np.where(pre_wrap_delta!= post_wrap_delta)`. But this again causes problems with numerical tolerances, and so when Iris tries to work out where the new wrapping point should be, it gets this wrong.

I've outlined an example below :

Using the bounds and wrapped bounds we have above, the `pre_wrap_delta` looks like:
```
array([[0.1],
        [0.1],
        [0.1],
        [0.1]]),
```
(note the intersections doesn't include the first cell (299.8,299.9) as that is outside our intersection area of (-60, 60)
and the `post_wrap_delta` looks like
```
array([[ 1.000e-01],
       [-3.599e+02],
       [ 1.000e-01],
       [ 1.000e-01]])
```
Those _almost_ match but not quite:
```
>>> pre_wrap_delta != post_wrap_delta
array([[ True],
       [ True],
       [ True],
       [ True]])
>>> pre_wrap_delta[-1][0]
0.0999999999999659
>>> post_wrap_delta[-1][0]
0.10000000000000142
```
So Iris thinks they are *all* straddle points, and so ends up choosing a wrong new wrap point.

This PR changes the search for straddle points to instead use a tolerant equality check:
```
split_cell_indices, _ = np.where(
                    ~np.isclose(pre_wrap_delta, post_wrap_delta)
                )
```
Which allows for these floating point problems.

### DIFF
--- a/docs/iris/src/whatsnew/3.0.3.rst
+++ b/docs/iris/src/whatsnew/3.0.3.rst
@@ -1,0 +1,587 @@
+.. include:: ../common_links.inc
+
+v3.0.3 (XX July 2021)
+*********************
+
+This document explains the changes made to Iris for this release
+(:doc:`View all changes <index>`.)
+
+
+.. dropdown:: :opticon:`alert` v3.0.3 Patches
+   :container: + shadow
+   :title: text-primary text-center font-weight-bold
+   :body: bg-light
+   :animate: fade-in
+   :open:
+
+   The patches included in this release include:
+
+   🐛 **Bugs Fixed**
+
+   #. `@lbdreyer`_ modified :meth:`~iris.cube.Cube.intersection` to use a tolerant 
+      equality check, when looking for cells that straddle the wrapping point.
+      (:pull:`4220`)
+
+
+.. dropdown:: :opticon:`alert` v3.0.2 Patches
+   :container: + shadow
+   :title: text-primary text-center font-weight-bold
+   :body: bg-light
+   :animate: fade-in
+   :open:
+
+   The patches included in this release include:
+
+   🐛 **Bugs Fixed**
+
+   #. `@jonseddon`_ handled a malformed ``um_stash_source`` CF variable attribute in
+      a netCDF file rather than raising a ``ValueError``. (:pull:`4035`)
+
+   #. `@rcomer`_ fixed :meth:`~iris.cube.Cube.intersection` for special cases
+      where one cell's bounds align with the requested maximum and minimum, as
+      reported in :issue:`3391`. (:pull:`4059`)
+
+   #. `@bjlittle`_ resolved a regression in arithmetic behaviour between a coordinate
+      and a cube which resulted in a ``NotYetImplementedError`` being raised, as reported
+      in :issue:`4000`. This fix supports ``+``, ``-``, ``*``, and ``/`` operations
+      between a coordinate and a cube, and for convenience additionally includes
+      :meth:`iris.cube.Cube.__neg__` support. (:pull:`4159`)
+
+   📚 **Documentation**
+
+   #. `@bjlittle`_ updated the ``intersphinx_mapping`` and fixed documentation
+      to use ``stable`` URLs for `matplotlib`_. (:pull:`4003`) [``pre-v3.1.0``]
+
+   💼 **Internal**
+
+   #. `@jamesp`_ updated a test to the latest numpy version (:pull:`3977`) [``pre-v3.1.0``]
+
+   #. `@bjlittle`_ enabled `cirrus-ci`_ compute credits for non-draft pull-requests
+      from collaborators targeting the Iris ``master`` branch. (:pull:`4007`)
+      [``pre-v3.1.0``]
+
+   #. `@bjlittle`_ added conditional task execution to `.cirrus.yml`_ to allow
+      developers to easily disable `cirrus-ci`_ tasks. See
+      :ref:`skipping Cirrus-CI tasks`. (:pull:`4019`) [``pre-v3.1.0``]
+
+   #. `@pp-mo`_ adjusted the use of :func:`dask.array.from_array` in :func:`iris._lazy_data.as_lazy_data`,
+      to avoid the dask 'test access'. This makes loading of netcdf files with a
+      large number of variables significantly faster. (:pull:`4135`)
+
+   #. `@pp-mo`_ reverted a change made previously in (:pull:`3659`) to
+      :meth:`iris.fileformats.pp.PPDataProxy.__getitem__`. The check for empty slicings
+      is no longer needed since (:pull:`4135`) was added. (:pull:`4141`)
+
+   Note that, the above contributions labelled with ``pre-v3.1.0`` are part of the forthcoming
+   Iris v3.1.0 release, but require to be included in this patch release.
+
+
+.. dropdown:: :opticon:`alert` v3.0.1 Patches
+   :container: + shadow
+   :title: text-primary text-center font-weight-bold
+   :body: bg-light
+   :animate: fade-in
+
+   The patches included in this release include:
+
+   💼 **Internal**
+
+   #. `@bjlittle`_ gracefully promote formula terms within :mod:`~iris.aux_factory` that have ``units`` of ``unknown``
+      to ``units`` of ``1`` (dimensionless), where the formula term **must** have dimensionless ``units``. Without this
+      graceful treatment of ``units`` the resulting :class:`~iris.cube.Cube` will **not** contain the expected auxiliary
+      factory, and the associated derived coordinate will be missing. (:pull:`3965`)
+
+
+.. dropdown:: :opticon:`report` v3.0.0 Release Highlights
+   :container: + shadow
+   :title: text-primary text-center font-weight-bold
+   :body: bg-light
+   :animate: fade-in
+
+   The highlights for this major release of Iris include:
+
+   * We've finally dropped support for ``Python 2``, so welcome to ``Iris 3``
+     and ``Python 3``!
+   * We've extended our coverage of the `CF Conventions and Metadata`_ by
+     introducing support for `CF Ancillary Data`_ and `Quality Flags`_,
+   * Lazy regridding is now available for several regridding schemes,
+   * Managing and manipulating metadata within Iris is now easier and more
+     consistent thanks to the introduction of a new common metadata API,
+   * :ref:`Cube arithmetic <cube maths>` has been significantly improved with
+     regards to extended broadcasting, auto-transposition and a more lenient
+     behaviour towards handling metadata and coordinates,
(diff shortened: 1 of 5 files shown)


lib/iris/tests/unit/cube/test_Cube.py
```diff
@@ -1817,6 +1817,17 @@ def test_numerical_tolerance(self):
...
+    def test_numerical_tolerance_wrapped(self):
+        # test the tolerance on the coordinate value causes modulus wrapping
+        # where appropriate
+        cube = create_cube(0.5, 3600.5, bounds=True)
+        lons = cube.coord("longitude")
+        lons.points = lons.points / 10
+        lons.bounds = lons.bounds / 10
+        result = cube.intersection(longitude=(-60, 60))
+        self.assertAlmostEqual(result.coord("longitude").points[0], -60.05)
+        self.assertAlmostEqual(result.coord("longitude").points[-1], 60.05)
```
lbdreyer, line 1829: With iris 3.0.2, this would return a lon coord that looks like:
```
>>> result.coord('longitude')
DimCoord(array([-59.95, -59.85, -59.75, ..., 299.75, 299.85, 299.95]), bounds=array([[-60. , -59.9],
       [-59.9, -59.8],
       [-59.8, -59.7],
       ...,
       [299.7, 299.8],
       [299.8, 299.9],
       [299.9, 300. ]]), standard_name='longitude', units=Unit('degrees'))
```

rcomer:
So the logic introduced at #4059 assumed there would be only one split cell, and that split cell would be around the requested minimum.  This PR ensures that that assumption is valid, so looks right to me.

Looking at this again makes me realise #4059 should have been slightly different though.  If the check had used `maximum - modulus` instead of `maximum % modulus`, the [recalculation](https://github.com/SciTools/iris/blob/274057bcf875c16fb4e5b4517471aa143e0e372e/lib/iris/cube.py#L3190-L3203) would have happened, since -300 is not in `cells`.  I've opened #4221 to follow up on that at some point (I don't think it's urgent).

lib/iris/tests/unit/cube/test_Cube.py
bjlittle, line 1829: @lbdreyer Could we explicitly test the bounds here within this test?

Previously, this might have helped us detect that all was not well e.g., something long the lines of:
```python
        lon = result.coord("longitude")
        self.assertAlmostEqual(lon.points[0], -60.05)
        self.assertAlmostEqual(lon.points[-1], 60.05)
        dtype = lon.bounds.dtype
        np.testing.assert_array_almost_equal(lon.bounds[0], np.array([-60.1, -60.], dtype=dtype))
        np.testing.assert_array_almost_equal(lon.bounds[-1], np.array([60., 60.1], dtype=dtype))
```
And perhaps also do the same for the previous test `test_numerical_tolerance` :+1:

bjlittle (changes requested):
I'd forgotten just how much `cube.intersection` hurts my head :exploding_head: 

Awesome fix :rocket: :100: 

Just a couple of minor changes to bolster the tests and I think we're there :+1:

lbdreyer:
Docs are broken :weary:
I'll try fixing tomorrow morning!

docs/iris/src/whatsnew/3.0.3.rst
```diff
@@ -0,0 +1,587 @@
...
+      equality check, when looking for cells that straddle the wrapping point.
+      (:pull:`4220`)
+
+
+.. dropdown:: :opticon:`alert` v3.0.2 Patches
+   :container: + shadow
+   :title: text-primary text-center font-weight-bold
+   :body: bg-light
+   :animate: fade-in
+   :open:
```
bjlittle, line 31: @lbdreyer Could you remove this `:open:`.

For patches I've been only keeping the latest patch dropdown open by default, so that it is highlighted to the reader - and they can open the other dropdowns by choice.

firefox

docs/iris/src/whatsnew/3.0.3.rst
```diff
@@ -0,0 +1,587 @@
...
+     and ``Python 3``!
+   * We've extended our coverage of the `CF Conventions and Metadata`_ by
+     introducing support for `CF Ancillary Data`_ and `Quality Flags`_,
+   * Lazy regridding is now available for several regridding schemes,
+   * Managing and manipulating metadata within Iris is now easier and more
+     consistent thanks to the introduction of a new common metadata API,
+   * :ref:`Cube arithmetic <cube maths>` has been significantly improved with
+     regards to extended broadcasting, auto-transposition and a more lenient
+     behaviour towards handling metadata and coordinates,
+   * Our :ref:`documentation <whatsnew 3.0.2 docs>` has been refreshed,
```
bjlittle, line 113: Change this to `<whatsnew 3.0.3 docs>`

docs/iris/src/whatsnew/3.0.3.rst
```diff
@@ -0,0 +1,587 @@
...
+     restructured, revitalised and rehosted on `readthedocs`_,
+   * It's now easier than ever to :ref:`install Iris <installing_iris>`
+     as a user or a developer, and the newly revamped developers guide walks
+     you though how you can :ref:`get involved <development_where_to_start>`
+     and contribute to Iris,
+   * Also, this is a major release of Iris, so please be aware of the
+     :ref:`incompatible changes <whatsnew 3.0.2 changes>` and
+     :ref:`deprecations <whatsnew 3.0.2 deprecations>`.
```
bjlittle, line 121: As above, change to `<whatsnew 3.0.3 changes>` and `<whtasnew 3.0.3 deprecations>` :+1:

docs/iris/src/whatsnew/3.0.3.rst
```diff
@@ -0,0 +1,587 @@
...
+   Previously, the first tick label would occasionally be duplicated. This also
+   removes the use of the deprecated `matplotlib`_ ``IndexFormatter``. (:pull:`3857`)
+
+#. `@znicholls`_ fixed :meth:`~iris.quickplot._title` to only check
+   ``units.is_time_reference`` if the ``units`` symbol is not used. (:pull:`3902`)
+
+#. `@rcomer`_ fixed a bug whereby numpy array type attributes on a cube's
+   coordinates could prevent printing it.  See :issue:`3921`.  (:pull:`3922`)
+
+.. _whatsnew 3.0.2 changes:
```
bjlittle, line 267: Now also change the associated anchors i.e., `.. _whatsnew 3.0.3 changes:`

docs/iris/src/whatsnew/3.0.3.rst
```diff
@@ -0,0 +1,587 @@
...
+
+#. `@bjlittle`_ changed the :meth:`iris.analysis.maths.IFunc.__call__` to ignore
+   any surplus ``other`` keyword argument for a ``data_func`` that requires
+   **only one** argument. This aligns the behaviour of
+   :meth:`iris.analysis.maths.IFunc.__call__` with
+   :func:`~iris.analysis.maths.apply_ufunc`. Previously a ``ValueError``
+   exception was raised. (:pull:`3785`)
+
+
+.. _whatsnew 3.0.2 deprecations:
```
bjlittle, line 330: Same here, thanks :+1:

docs/iris/src/whatsnew/3.0.3.rst
```diff
@@ -0,0 +1,587 @@
...
+   necessary now that ``Python2`` support has been dropped. (:pull:`3468`)
+
+#. `@stephenworsley`_ and `@trexfeathers`_ unpinned Iris to use the latest version
+   of `Proj`_. (:pull:`3762`)
+
+#. `@stephenworsley`_ and `@trexfeathers`_ removed GDAL from the extensions
+   dependency group. We no longer consider it to be an extension. (:pull:`3762`)
+
+
+.. _whatsnew 3.0.2 docs:
```
bjlittle, line 379: And this is the last one to change

bjlittle (changes requested):
@lbdreyer Awesome. Just some minor changes to the `whatsnew`. I think my suggestions should fix the docs building.

Also, could you:

- add a reference to the `3.0.3.rst` at the top of the `docs/iris/src/whatsnew/index.rst`
- bump the `iris.__version__` to be `3.0.3` (might as well do this now :+1:)

docs/iris/src/whatsnew/3.0.3.rst
```diff
@@ -0,0 +1,587 @@
+.. include:: ../common_links.inc
+
+v3.0.3 (XX July 2021)
```
bjlittle, line 3: We can update the date just before we release.

We need to do this manually in the `v3.0.x` release feature branch. Thanksfully this is now automated in `main`

docs/iris/src/whatsnew/3.0.3.rst
```diff
@@ -0,0 +1,587 @@
...
+   :animate: fade-in
+   :open:
+
+   The patches included in this release include:
+
+   🐛 **Bugs Fixed**
+
+   #. `@lbdreyer`_ modified :meth:`~iris.cube.Cube.intersection` to use a tolerant 
+      equality check, when looking for cells that straddle the wrapping point.
+      (:pull:`4220`)
```
bjlittle, line 23: Lovely :+1:

bjlittle (approved):
@lbdreyer Fantastic, thanks :beers:

PR #4222, opened 2021-07-02
unpin cftime

bjlittle:
### Description
This PR is targeting the `v3.0.x` release feature branch, and is part of the `iris` 3.0.4 patch release.

This patch release is dedicated to removing technical debt by unpinning `cftime`, which will then allow the community to use `iris` with the latest `cftime`, `cf-units`, and `nc-time-axis` packages.

Note that, this patch release has forced us to drop support of `python` 3.6 prematurely due to third party package dependencies. `iris` was scheduled to drop 3.6 support in `iris` 3.1

The rendered `whatsnew` is available [here](https://bjlittle-iris.readthedocs.io/en/latest/whatsnew/3.0.4.html)

This PR welcomes reviewer comments, but will stay in draft until the `iris` 3.0.3 release is performed. 

---


lib/iris/fileformats/pp.py
```diff
@@ -1513,9 +1545,34 @@ def t2(self, dt):
...
+        dates.
+
+        """
+        self.lbyrd = 0
+        self.lbmond = 0
+        self.lbdatd = 0
+        self.lbhrd = 0
+        self.lbmind = 0
+        self.lbdayd = 0
+
```
rcomer, line 1575: I don't know if it helps here, but calendar agnostic datetimes are still possible at `cftime` v1.5 e.g.

```python
cftime.datetime(0, 0, 0, calendar=None)
```

lib/iris/fileformats/pp.py
```diff
@@ -1476,8 +1480,18 @@ def t1(self):
 
         """
         if not hasattr(self, "_t1"):
+            has_year_zero = self.lbyr == 0
+            lbmon, lbdat = self.lbmon, self.lbdat
+            if lbmon == 0 and lbdat == 0:
+                lbmon = lbdat = 1
```
bjlittle, line 1486: It's invalid to have a zero `month` or `day` in any calendar.

However, it is possible to have a PP field with zero values already configured from file loading (not set via property setter methods). 

The approach here is to only ever return a valid `cftime.datetime` regardless. Raising an exception at this point is really not acceptable to users.

lib/iris/fileformats/pp.py
```diff
@@ -1489,6 +1503,14 @@ def t1(self, dt):
         self.lbhr = dt.hour
         self.lbmin = dt.minute
         self.lbday = int(dt.strftime("%j"))
+        if self.lbyr == 0:
+            if self.lbmon == 0 and self.lbdat == 0:
+                self.lbmon = self.lbdat = 1
```
bjlittle, line 1508: Account for the historical special case of zero `year`, `month`, `day`.

lib/iris/fileformats/pp.py
```diff
@@ -1513,9 +1545,34 @@ def t2(self, dt):
...
+        dates are not valid :func:`datetime.datetime` or :func`cftime.datetime`
+        dates.
+
+        """
+        self.lbyrd = 0
+        self.lbmond = 0
+        self.lbdatd = 0
+        self.lbhrd = 0
+        self.lbmind = 0
+        self.lbdayd = 0
```
bjlittle, line 1574: We only have the use case of wanting to zero `t2` date field, not the `t1` date field - so I've not included that behaviour.

docs/iris/src/whatsnew/3.0.4.rst
```diff
@@ -0,0 +1,592 @@
+.. include:: ../common_links.inc
+
+v3.0.4 (XX July 2021)
```
bjlittle, line 3: Still require to set the date

lib/iris/fileformats/pp.py
bjlittle: Don't know either, but I'd imaging holding on to the `calendar` is more important rather than being able to represent an invalid date.

Most of the fixes throughout this PR involve baking the `calendar` back into the `cftime.datetime`, because now it's possible.

rcomer: > I can perhaps give it a try and see what happens...

Maybe wait for a proper reviewer who understands this module!

bjlittle: @rcomer To be honest, that was a lovely nugget of a simplification that I've now applied, thanks :star_struck:

trexfeathers (changes requested):
Nice work @bjlittle!

I'd also like to see the same checks in [nimrod_load_rules](https://github.com/bjlittle/iris/blob/630be29bd00f67521b7650eb6781fdcdc140f042/lib/iris/fileformats/nimrod_load_rules.py), does that sound sensible?

lib/iris/tests/unit/fileformats/pp_load_rules/test__convert_time_coords.py
```diff
@@ -47,7 +47,9 @@ class TestLBTIMx0x_SingleTimepoint(TestField):
     def _check_timepoint(self, lbcode, expect_match=True):
         lbtim = _lbtim(ib=0, ic=1)
         t1 = nc_datetime(1970, 1, 1, hour=6, minute=0, second=0)
-        t2 = nc_datetime(0, 0, 0)  # not used in result
+        t2 = nc_datetime(
+            0, 0, 0, calendar=None, has_year_zero=True
```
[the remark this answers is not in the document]
bjlittle: Confusingly, you can create a `cftime.datetime(0, 0, 0, calendar=None)` with `has_year_zero` set to either `True` or `False`.

This makes no difference until you come to compare e.g.,
```python
from cftime import datetime

>>> a = datetime(0, 0, 0, calendar=None)
>>> b = datetime(0, 0, 0, calendar=None, has_year_zero=True)
>>> a == a
True
>>> b == b
True
>>> a == b
TypeError: cannot compare cftime.datetime(0, 0, 0, 0, 0, 0, 0, calendar='', has_year_zero=False) and cftime.datetime(0, 0, 0, 0, 0, 0, 0, calendar='', has_year_zero=True)
```
Just for consistency, I'm going to adopt explicitly setting `has_year_zero=True` in this case, although it would be nice for `cftime` to just handle this... in fact, I might raise an issue to see if the devs complain for a good reason or adopt it as a use case :+1:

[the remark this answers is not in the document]
bjlittle: @trexfeathers For the `nimrod_load_rules`, in both instances where a `cftime.datetime` object is created there is surrounding logic to ensure that the `year` used is non-zero. So I think this is okay as it stands, right? 

Just check in case I've missed something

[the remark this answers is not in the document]
trexfeathers: The month and day could still be `0` though.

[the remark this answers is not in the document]
bjlittle: Also, see https://github.com/Unidata/cftime/issues/248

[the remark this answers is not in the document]
bjlittle: > The month and day could still be `0` though.

Ahhh sorry, I see what you mean...

My take on this was that the `nimrod` use case is slightly different from the `pp` use case, in that for `nimrod` the `calendar` is explicitly nailed down to being `gregorian`, see https://github.com/SciTools/iris/blob/main/lib/iris/fileformats/nimrod_load_rules.py#L27

They are not ever expecting to see a zero `year`, nor a zero `month` or `day`, as it's indeed an invalid date for that `calendar`. So I think it's more likely that the `nimrod` datetime data is genuinely invalid in this case and an exception should be raised; and it will by `cftime`. 

I think that if we instead returned a valid `cftime.datetime` with `calendar=None`, then that might be more dangerous/unexpected for `nimrod` users.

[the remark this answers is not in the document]
trexfeathers: Nice catch, I hadn't spotted that. Totally agree with you.

trexfeathers:
After 2cef34830f780448b944ca65b9175037e0947f34 there are still 3 more cases to sort out in `test__epoch_date_hours`.

trexfeathers (approved):

pp-mo (approved):
After some to + fro I found everything I thought I was checking was already covered by @trexfeathers + in most cases discussed at length.

So, LGTM !

bjlittle:
@trexfeathers and @pp-mo Awesome, thanks :beers:

PR #4224, opened 2021-07-05
Default connectivity start_index

pp-mo:
### Description
Allow 'start_index' to default to 0 when creating Connectivity from a file-var, if var has no 'start_index' attribute.

---

### DIFF
--- a/lib/iris/experimental/ugrid/__init__.py
+++ b/lib/iris/experimental/ugrid/__init__.py
@@ -3643,7 +3643,7 @@ def _build_connectivity(connectivity_var, file_path, location_dims):
     indices_data = netcdf._get_cf_var_data(connectivity_var, file_path)
 
     cf_role = connectivity_var.cf_role
-    start_index = connectivity_var.start_index
+    start_index = getattr(connectivity_var, 'start_index', None)
 
     dim_names = connectivity_var.dimensions
     # Connectivity arrays must have two dimensions.
--- a/lib/iris/experimental/ugrid/__init__.py
+++ b/lib/iris/experimental/ugrid/__init__.py
@@ -3643,7 +3643,7 @@ def _build_connectivity(connectivity_var, file_path, location_dims):
     indices_data = netcdf._get_cf_var_data(connectivity_var, file_path)
 
     cf_role = connectivity_var.cf_role
-    start_index = getattr(connectivity_var, "start_index", None)
+    start_index = getattr(connectivity_var, "start_index", 0)
 
     dim_names = connectivity_var.dimensions
     # Connectivity arrays must have two dimensions.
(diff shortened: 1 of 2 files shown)


lib/iris/experimental/ugrid/__init__.py
```diff
@@ -3643,7 +3643,7 @@ def _build_connectivity(connectivity_var, file_path, location_dims):
     indices_data = netcdf._get_cf_var_data(connectivity_var, file_path)
 
     cf_role = connectivity_var.cf_role
-    start_index = connectivity_var.start_index
+    start_index = getattr(connectivity_var, "start_index", None)
```
trexfeathers, line 3646: Either:

* The default here should be `0`.
* Creation of a `Connectivity` instance should only pass a `start_index` kwarg if `start_index != 1`.
* The `Connectivity` class should be refactored to make `start_index` an optional parameter, applying the default `0` value **within** `__init__()` rather than making it a parameter default.
  (I'm guessing this is the preferred option?)

pp-mo: But, what is wrong with passing `None` to indicate 'use default' ?
The use of 'None' as a default is a public in the signature of the Connectivity constructor -- so it is _equivalent_  to not specifying start_index, and you can trust that.
We then just need to trust that Connectivity implements its default in line with the UGRID rules.  Which seem perfectly reasonable to me.

trexfeathers: Yes, I thought that's exactly what I said in the third bullet?

pp-mo: > Yes, I thought that's exactly what I said in the third bullet?
> I thought that's exactly what I said in the third bullet?

Sorry, you're right.  I had mis-read it somehow.
We already have a 'start_index=0', and can't pass a None.
So I would _instead_ fix as your first bullet "default here should be 0"

pp-mo:
```suggestion
    start_index = getattr(connectivity_var, "start_index", 0)
```

trexfeathers:
Your proposal will certainly be OK, but I would be open to use modifying `Connectivity` if you'd prefer?

pp-mo:
> Your proposal will certainly be OK, but I would be open to use modifying Connectivity if you'd prefer?

No I have no real preference, so happiest to take the least-change way

But tests are failing now, looks like the merge is not delivering a good merge on imagerepo.json.
Will try a rebase ...

pp-mo:
Rebase doesn't do anything, as it is down to a change in test-iris-imagehash, and the matching Iris change (to remove a ref) is on the main branch :disappointed: 

The latest change here mirrors that.
I guess it will cause a merge conflict when we merge back mesh-data-model, but it will be a drop in the ocean !

trexfeathers:
@pp-mo you also need to bring in the correct use of `linscale` from #4115.

pp-mo:
Closed in favour of #4231

PR #4227, opened 2021-07-06
lru metadata manager factory

bjlittle:
### Description
This PR cycles back to the Common Metadata API and includes some small but significant optimisations to the `iris.common.metadata.metadata_manager_factory` function.

The `metadata_manager_factory` function leverages the benefit of creating metadata manager classes dynamically at runtime, but this is a relatively expensive operation compared to creating equivalent but static classes.

Employing a simple LRU caching strategy for the metadata manager classes created at runtime provides significant savings, along with careful avoidance of property lookups and type checking when creating instances of the dynamic classes.

The following `asv` metrics provide support for this proposed change:

Capture

In particular, note the performance gains when creating common `iris` first-class objects, such as:
* `iris.coords.AncillaryVariable` (significant)
* `iris.coords.AuxCoord` (significant)
* `iris.coords.DimCoord` (meh)
* `iris.coords.CellMeasure` (significant)
* `iris.cube.Cube` (significant)

---

### DIFF
--- a/lib/iris/common/metadata.py
+++ b/lib/iris/common/metadata.py
@@ -13,7 +13,7 @@ from abc import ABCMeta
 from collections import namedtuple
 from collections.abc import Iterable, Mapping
 from copy import deepcopy
-from functools import wraps
+from functools import lru_cache, wraps
 import re
 
 import numpy as np
@@ -1338,31 +1338,8 @@ class DimCoordMetadata(CoordMetadata):
         return super().equal(other, lenient=lenient)
 
 
-_FACTORY_CACHE = {}
-
-
-def metadata_manager_factory(cls, **kwargs):
-    """
-    A class instance factory function responsible for manufacturing
-    metadata instances dynamically at runtime.
-
-    The factory instances returned by the factory are capable of managing
-    their metadata state, which can be proxied by the owning container.
-
-    Args:
-
-    * cls:
-        A subclass of :class:`~iris.common.metadata.BaseMetadata`, defining
-        the metadata to be managed.
-
-    Kwargs:
-
-    * kwargs:
-        Initial values for the manufactured metadata instance. Unspecified
-        fields will default to a value of 'None'.
-
-    """
-
+@lru_cache
+def _factory_cache(cls, **kwargs):
     def __init__(self, cls, **kwargs):
         # Restrict to only dealing with appropriate metadata classes.
         if not issubclass(cls, BaseMetadata):
@@ -1471,10 +1448,33 @@ def metadata_manager_factory(cls, **kwargs):
     if cls is CubeMetadata:
         namespace["_names"] = cls._names
 
-    # Dynamically create the class
-    Metadata = _FACTORY_CACHE.setdefault(
-        str(cls), type(name, bases, namespace)
-    )
+    return type(name, bases, namespace)
+
+
+def metadata_manager_factory(cls, **kwargs):
+    """
+    A class instance factory function responsible for manufacturing
+    metadata instances dynamically at runtime.
+
+    The factory instances returned by the factory are capable of managing
+    their metadata state, which can be proxied by the owning container.
+
+    Args:
+
+    * cls:
+        A subclass of :class:`~iris.common.metadata.BaseMetadata`, defining
+        the metadata to be managed.
+
+    Kwargs:
+
+    * kwargs:
+        Initial values for the manufactured metadata instance. Unspecified
+        fields will default to a value of 'None'.
+
+    """
+
+    # Dynamically create the class at runtime or get a cached version of it
+    Metadata = _factory_cache(cls, **kwargs)
 
     # Now manufacture an instance of that class.
     metadata = Metadata(cls, **kwargs)
--- a/lib/iris/common/metadata.py
+++ b/lib/iris/common/metadata.py
@@ -1344,10 +1344,12 @@ def _factory_cache(cls):
         #: The metadata class to be manufactured by this factory.
         self.cls = cls
 
-        # Proxy for self.cls._fields
+        # Proxy for self.cls._fields for later internal use, as this
+        # saves on indirect property lookup via self.cls
         self._fields = cls._fields
 
         # Initialise the metadata class fields in the instance.
+        # Use cls directly here since it's available.
         for field in cls._fields:
             setattr(self, field, None)
 
@@ -1411,12 +1413,7 @@ def _factory_cache(cls):
         fields = {field: getattr(self, field) for field in self._fields}
         return self.cls(**fields)
 
-    # Restrict factory to appropriate metadata classes only.
-    if not issubclass(cls, BaseMetadata):
-        emsg = "Require a subclass of {!r}, got {!r}."
-        raise TypeError(emsg.format(BaseMetadata.__name__, cls))
-
-    # Define the name, (inheritance) bases and namespace of the dynamic class.
+    # Define the name, (inheritance) bases, and namespace of the dynamic class.
     name = "MetadataManager"
     bases = ()
     namespace = {
--- a/lib/iris/common/metadata.py
+++ b/lib/iris/common/metadata.py
@@ -1440,7 +1440,10 @@ def _factory_cache(cls):
     if cls is CubeMetadata:
         namespace["_names"] = cls._names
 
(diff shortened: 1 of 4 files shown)


lib/iris/common/metadata.py
```diff
@@ -1338,39 +1338,14 @@ def equal(self, other, lenient=None):
...
-
-    """
-
+@lru_cache(maxsize=None)
+def _factory_cache(cls):
     def __init__(self, cls, **kwargs):
-        # Restrict to only dealing with appropriate metadata classes.
-        if not issubclass(cls, BaseMetadata):
-            emsg = "Require a subclass of {!r}, got {!r}."
-            raise TypeError(emsg.format(BaseMetadata.__name__, cls))
```
bjlittle, line 1367: Type checking is expensive, but thankfully we can safely remove this redundant `cls` type check at metadata manager instance creation time, as it has already been performed by the `metadata_manager_factory` function (see lines 1412-1414)

lib/iris/common/metadata.py
```diff
@@ -1338,39 +1338,14 @@ def equal(self, other, lenient=None):
...
-
         #: The metadata class to be manufactured by this factory.
         self.cls = cls
 
         # Initialise the metadata class fields in the instance.
-        for field in self.fields:
+        for field in cls._fields:
```
bjlittle, line 1348: Yup, it actually matters to avoid the convenience of property lookup.

lib/iris/common/metadata.py
```diff
@@ -1438,14 +1413,6 @@ def values(self):
...
         raise TypeError(emsg.format(BaseMetadata.__name__, cls))
 
-    # Check whether kwargs have valid fields for the specified metadata.
-    if kwargs:
-        extra = [field for field in kwargs.keys() if field not in cls._fields]
-        if extra:
-            bad = ", ".join(map(lambda field: "{!r}".format(field), extra))
-            emsg = "Invalid {!r} field parameters, got {}."
-            raise ValueError(emsg.format(cls.__name__, bad))
-
```
bjlittle, line 1448: Avoid checking the `kwargs` at metadata manager instance creation time, and instead do it within the `metadata_manager_factory`. This also deftly avoids the potential issue of non hashable `kwargs` being passed through `lru_cache`.

lib/iris/common/metadata.py
```diff
@@ -1338,39 +1338,14 @@ def equal(self, other, lenient=None):
...
-    Kwargs:
-
-    * kwargs:
-        Initial values for the manufactured metadata instance. Unspecified
-        fields will default to a value of 'None'.
-
-    """
-
+@lru_cache(maxsize=None)
```
bjlittle, line 1341: Explicitly set `maxsize=None` to allow `functools.lru_cache` to work for both [py37](https://docs.python.org/3.7/library/functools.html#functools.lru_cache) and [py38](https://docs.python.org/3.8/library/functools.html#functools.lru_cache)

lib/iris/common/metadata.py
jamesp: Not sure I understand this comment - from the code above it looks like the lru_cache is decorating a function of single argument `cls`.  How does the `kwargs` affect the cache?  Is there some closure I'm not spotting?

jamesp: Ah, are you saying that you've moved `kwargs` to `metadata_manager_factory`, so it never sees the cache?

jamesp:
Lovely stuff @bjlittle, speed ups look significant!

lib/iris/common/metadata.py
bjlittle: I was considering that approach, as in, be sensible and pass in the right class otherwise all bets are off.

It is a public function but I can't ever see users calling it, and if they did, then they really need to know what their doing... so yeah, I'm happy to purge that check and the associated test. Nice one :+1:

lib/iris/common/metadata.py
```diff
@@ -1344,6 +1344,9 @@ def __init__(self, cls, **kwargs):
         #: The metadata class to be manufactured by this factory.
         self.cls = cls
 
+        # Proxy for self.cls._fields
+        self._fields = cls._fields
+
```
jamesp, line 1349: Probably worth a comment in the code here to explain whey you're creating a proxy and then on the next line not using it!

lib/iris/common/metadata.py
bjlittle: Yeah, it's a bit subtle, but there's no need to pass the `kwargs` into `_factory_cache`. I wanted to avoid potentially passing in non-hashable `kwargs` (as `lru_cache` enforces this), plus I also want things to fail fast i.e., within the factory and not the cache. Does that help?

jamesp:
Probably worth another quick set of eyes on this from someone who is more familiar with metadata than I am... @trexfeathers or @pp-mo?  

Tests are passing and the code looks sensible so I'm happy.

lib/iris/common/metadata.py
```diff
@@ -1468,12 +1435,52 @@ def values(self):
...
+        A manager instance for the provided metadata ``cls``.
+
+    """
+    # Check whether kwargs have valid fields for the specified metadata.
+    if kwargs:
+        extra = [field for field in kwargs.keys() if field not in cls._fields]
+        if extra:
+            bad = ", ".join(map(lambda field: "{!r}".format(field), extra))
+            emsg = "Invalid {!r} field parameters, got {}."
+            raise ValueError(emsg.format(cls.__name__, bad))
```
bsherratt, line 1474: I think it would be cleaner to do this error handling in `Manager.__init__`, at the same time as assigning them, eg
```python
for field in cls._fields:
    setattr(self, field, kwargs.pop(field, None))

if kwargs:
    bad = ", ".join(kwargs.keys())
    raise TypeError(f"Invalid {type(self)!r} parameters: {bad}")
```

bsherratt (commented):
I had a bit of a look at this function a while ago (following your comment on #4142) so possibly qualified to comment if you don't mind? I'm excited to see these changes anyway, it *was* a little convoluted.

The bit that stood out most was definitely that `metadata_manager_factory` went to the trouble of defining a whole new class, just to return a single instance of it and forget about the class. Which I see is addressed by factoring out `_factory_cache`.

Personally, I think I wouldn't have used `lru_cache` at all, just created all managers upfront
```python
for cls in (CubeMetadata, CoordMetadata, ...):
    manager = _manager_class_factory(cls)
    globals()[manager.__name__] = manager
```
But that's back to the old question of how much should happen at import time vs when actually needed so down to you. The cache is definitely nicer than assigning to `globals` directly; that's a different tradeoff, for being able to use slightly more idiomatic code everywhere else, ie
```python
self.metadata = iris.common.CubeMetadataManager(**kwargs)
```
(Assuming the error handling of `kwargs` is also moved per inline comment)

bjlittle:
Here's the `asv` report for the latest commit, with additional metric coverage:

asv

Which still looks encouraging :smile:

bjlittle:
> Personally, I think I wouldn't have used `lru_cache` at all, just created all managers upfront

@bsherratt There are several ways to skin this particular cat, so to speak... but I'm still attracted to on-demand creation of the metadata manager classes. In particular, not building them upfront has the advantage that there isn't a definitive list that goes stale when new metadata classes are created/removed/changed. I was trying to hit the sweet spot for less maintenance burden on developers and I think this is reasonably close (but I'm sure everyone has an opinion on that :smile:) ... and this approach has already paid dividends recently when we plugged in new metadata classes for our forthcoming UGRID support.

To be honest, I'm not wedded to this implementation, and there are certainly further optimisations to be had for sure. As long as we cycle back around and start to polish and optimise (as per this PR), then I think ultimately we'll tend towards a reasonably robust and optimal solution.

lbdreyer (approved):
A nice, simple improvement to our metadata code with proven speed ups :+1: 

This should definitely go into Iris 3.1 so that our users can benefit from the optimisations.

By merging this in, we by no means close the door on this! We can reconsider #4241 (or indeed other alternatives) for Iris 3.2, but they will require more discussion and performance evaluations (via asv) so we compare them properly, and I don't want to hold up getting the optimisations to the user. We can always refine it later.

So I'm going to merge this in, and put #4241 under Iris 3.2.

PR #4124, opened 2021-05-07
Widen cube printout for long ancil or cell-measure names.

pp-mo:
### Description
Fixes #4123 

NOTE: adding  the test exposed ***another*** bug in cube printout : When widening, it stopped showing the cube units !
I have fixed that as well, but it did cause a knock-on in one other test (see separate commit with results-file change).

I would really rather have tidied up first, with #3998 
 -- after all, this bug has probably happened ***because of*** the unreasonable complexity of the Cube.summary code !
But this one is quite a simple fix, and if we do #3998 later it can check that it still passes this test.
And there is still a good deal of work needed to inject good levels of testing into #3998 

---
[Consult Iris pull request check list]( https://scitools-iris.readthedocs.io/en/latest/developers_guide/contributing_pull_request_checklist.html)

PR #4230, opened 2021-07-07
Wide cubestr fix 3v0vx

pp-mo:
### Fix bugs in cube printout, including critical failure for ancillaries with overlong names
Include the #4124 changes as a bugfix to v3
(also added whatsnew)

This is already on 'main', but there I missed off the whatsnew.
No need to address that I think, because this code has no longterm future, due to #4206 

~**WIP** because ...~
  * ~probably do ***not*** want  to consider this for v3.0.3~
  * ~wait for 3.0.4~
 (update : following [v3.0.3 release](https://github.com/SciTools/iris/releases/tag/v3.0.3) --  now **good to go !** )

I see we have no `whatsnew/latest.rst` on the bugfix branch.
So I hope creating a '3.0.4' whatsnew is the correct way with this.

---

### DIFF
--- a/docs/iris/src/whatsnew/3.0.4.rst
+++ b/docs/iris/src/whatsnew/3.0.4.rst
@@ -1,0 +1,592 @@
+.. include:: ../common_links.inc
+
+v3.0.4 (?? ??August 2021)
+*************************
+
+This document explains the changes made to Iris for this release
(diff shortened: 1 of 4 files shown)


pp-mo:
Replaced by #4233

PR #4233, opened 2021-07-08
Wide cubestr fix 3v0vx v2

pp-mo:
### Fix bugs in cube printout, including critical failure for ancillaries with overlong names
Include the #4124 changes as a bugfix to v3
(also added whatsnew)

This is already on 'main', but there I missed off the whatsnew.
No need to address that I think, because this code has no longterm future, due to #4206 
I see we have no `whatsnew/latest.rst` on the bugfix branch.
So I hope creating a '3.0.4' whatsnew is the correct way with this.

### DIFF
--- a/docs/iris/src/whatsnew/3.0.4.rst
+++ b/docs/iris/src/whatsnew/3.0.4.rst
@@ -28,7 +28,6 @@ This document explains the changes made to Iris for this release
    :title: text-primary text-center font-weight-bold
    :body: bg-light
    :animate: fade-in
-   :open:
 
    The patches included in this release include:
 
--- a/docs/iris/src/whatsnew/3.0.4.rst
+++ b/docs/iris/src/whatsnew/3.0.4.rst
@@ -1,0 +1,592 @@
+.. include:: ../common_links.inc
+
+v3.0.4 (?? ??August 2021)
+*************************
+
+This document explains the changes made to Iris for this release
+(:doc:`View all changes <index>`.)
+
+
+.. dropdown:: :opticon:`alert` v3.0.4 Patches
+   :container: + shadow
+   :title: text-primary text-center font-weight-bold
+   :body: bg-light
+   :animate: fade-in
+   :open:
+
+   The patches included in this release include:
+
+   🐛 **Bugs Fixed**
+
+   #. `@lbdreyer`_ modified :meth:`~iris.cube.Cube.intersection` to use a tolerant 
+      equality check, when looking for cells that straddle the wrapping point.
+      (:pull:`4220`)
+
+
+.. dropdown:: :opticon:`alert` v3.0.2 Patches
+   :container: + shadow
+   :title: text-primary text-center font-weight-bold
+   :body: bg-light
+   :animate: fade-in
+
+   The patches included in this release include:
+
+   🐛 **Bugs Fixed**
+
+   #. `@jonseddon`_ handled a malformed ``um_stash_source`` CF variable attribute in
+      a netCDF file rather than raising a ``ValueError``. (:pull:`4035`)
+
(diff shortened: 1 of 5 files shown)


docs/iris/src/whatsnew/3.0.4.rst
```diff
@@ -0,0 +1,592 @@
...
+
+   The patches included in this release include:
+
+   🐛 **Bugs Fixed**
+
+   #. `@lbdreyer`_ modified :meth:`~iris.cube.Cube.intersection` to use a tolerant 
+      equality check, when looking for cells that straddle the wrapping point.
+      (:pull:`4220`)
+
+
```
trexfeathers, line 25: Missing a `v3.0.3` dropdown, so I'm guessing this file wasn't based originally on [`3.0.3.rst`](https://github.com/SciTools/iris/blob/82dcba7fb6e33f70069800f7a40c3ae6171f522f/docs/iris/src/whatsnew/3.0.3.rst). Can you address this?

pp-mo: Not at all, it **_was_** based on 3.0.3, but I didn't understand what I was doing !

I'm now seeing that the different entries are aranged into a section for each point release,
but there is only one "v3.x.x (date)" subheading at the top.

Hope this is now correct ?

trexfeathers (approved):
Looks good - the What's New is what I'd expect, and the rest is a faithful cherry pick of #4124 .

pp-mo:
Great, thanks @trexfeathers .  A relief to get this done + out of the way !

bjlittle:
@trexfeathers We really didn't want to merge this now. That's why #4222 is sitting in draft.

We were holding off **all** merges to `3.0.x` post 3.0.3 in case we needed to cut another release with minimal bug fixes, otherwise I would have continued to bank PRs on 3.0.x after the 3.0.3 release.

docs/iris/src/whatsnew/3.0.4.rst
```diff
@@ -0,0 +1,597 @@
...
+   :animate: fade-in
+   :open:
+
+   The patches included in this release include:
+
+   🐛 **Bugs Fixed**
+
+   #. `@pp-mo`_ reverted a change made previously in (:pull:`3659`) to
+      :meth:`iris.fileformats.pp.PPDataProxy.__getitem__`. The check for empty slicings
+      is no longer needed since (:pull:`4135`) was added. (:pull:`4141`)
```
bjlittle, line 23: @pp-mo Why have you moved this `whatsnew` entry?

It was part of the 3.0.2 release. We can't change patch release history.

I was expecting a whatnew entry relating to this PR #4233 or #4124.

trexfeathers:
🙄 I really dropped the ball on this one. @bjlittle should we back it out again?

bjlittle:
No worries. Just hang fire as is.

I was giving a 2 week window between 3.0.3 and 3.0.4.

I should have been clearer in communicating that we didn't want to bank anything in between. It's a bit of an odd situation, but not the end of the world :+1:

docs/iris/src/whatsnew/3.0.4.rst
pp-mo: Sorry I seem to have messed this up somehow with in a merge/rebase process.
Also the original proposed new entry ("bugs in cube printout") has completely vanished.
Will address.

pp-mo: I must say, having a continously-augmented "lastest.rst" would avoid these problems.
Why can we not use that within bugfix branches as we do in the main branch ??

pp-mo:
Above comments addressed, I hope.
Needs re-think, since we already committed it.

PR #4235, opened 2021-07-08
pre-commit isort and black --check only for cirrus-ci

### LINKED ISSUE #4226: pre-commit, black and isort
## 📰 Custom Issue

When committing a change to a local branch, `black` and `isort` now tell you which files need fixing, but don't go ahead and fix them.  Is this intentional?  I think they were changed to do a `--check` at #4181, which makes sense within the context of the CI, but it's not obviously so helpful when committing a change.

In order to commit my change, I installed `black` and `isort` to run them directly.  I looked in [.pre-commit-config.yaml](https://github.com/SciTools/iris/blob/main/.pre-commit-config.yaml) to find out what versions I needed.  `conda` couldn't seem to find the specified `black` version, so installed that with `pip` instead.  This all feels unnecessarily complicated - is there a simpler way to install the right versions of these tools?

bjlittle:
### Description
This PR re-instates the `pre-commit` behaviour for the `black` and `isort` hooks, whereby they will automatically correct files in-place on `git commit` rather than performing a `--check`. This means that developers no longer require to explicitly install `black` or `isort`; this will all be managed by `pre-commit`.

Subsequently, the linting `cirrus-ci` task customises the `pre-commit-config.yaml` to ensure that `black` and `isort` only perform a `--check`, and will fail when a non-compliance is detected, as required.

Closes #4226 

---

### DIFF
--- a/noxfile.py
+++ b/noxfile.py
@@ -8,6 +8,8 @@ For further details, see https://nox.thea.codes/en/stable/#
 import hashlib
 import os
 from pathlib import Path
+from tempfile import NamedTemporaryFile
+from urllib.parse import unquote, urlsplit
 
 import nox
 from nox.logger import logger
@@ -173,10 +175,28 @@ def lint(session: nox.sessions.Session):
         A `nox.sessions.Session` object.
 
     """
+    import yaml
+
     # Pip install the session requirements.
     session.install("pre-commit")
-    # Execute the pre-commit linting tasks.
-    cmd = ["pre-commit", "run", "--all-files"]
+
+    # Load the pre-commit configuration YAML file.
+    with open(".pre-commit-config.yaml", "r") as fi:
+        config = yaml.load(fi, Loader=yaml.FullLoader)
+
+    # Ensure black and isort only perform a status check
+    # with a custom pre-commit configuration.
+    for entry in config["repos"]:
+        package = Path(urlsplit(unquote(entry["repo"])).path).name
+        if package in ["black", "isort"]:
+            entry["hooks"][0]["args"].insert(0, "--check")
+
+    with NamedTemporaryFile(mode="w+t", delete=False) as temp:
+        yaml.dump(config, temp)
+
+    # Execute the pre-commit linting tasks with the custom
+    # pre-commit configuration.
+    cmd = ["pre-commit", "run", "--all-files", f"--config={temp.name}"]
     hooks = ["black", "blacken-docs", "flake8", "isort"]
     for hook in hooks:
         session.run(*cmd, hook)
--- a/noxfile.py
+++ b/noxfile.py
@@ -182,15 +182,19 @@ def lint(session: nox.sessions.Session):
     with open(".pre-commit-config.yaml", "r") as fi:
         config = yaml.load(fi, Loader=yaml.FullLoader)
 
-    # Enumerate singleton pre-commit hooks.
-    hooks = [
+    # Enumerate the ids of pre-commit hooks we want to run.
+    # We're only capturing the hook id of pre-commit repos with *one*
+    # registered hook. This is a simple approach to filtering out
+    # the "https://github.com/pre-commit/pre-commit-hooks" hooks,
+    # all of which we don't want to run within nox.
+    ids = [
         entry["hooks"][0]["id"]
         for entry in config["repos"]
         if len(entry["hooks"]) == 1
     ]
 
     # Execute the pre-commit linting hooks.
-    [session.run("pre-commit", "run", "--all-files", hook) for hook in hooks]
+    [session.run("pre-commit", "run", "--all-files", id) for id in ids]
 
 
 @nox.session(python=PY_VER, venv_backend="conda")
--- a/noxfile.py
+++ b/noxfile.py
@@ -182,18 +182,18 @@ def lint(session: nox.sessions.Session):
     with open(".pre-commit-config.yaml", "r") as fi:
         config = yaml.load(fi, Loader=yaml.FullLoader)
 
-    # Enumerate the ids of pre-commit hooks we want to run.
-    # We're only capturing the hook id of pre-commit repos with *one*
-    # registered hook. This is a simple approach to filtering out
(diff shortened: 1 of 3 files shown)


noxfile.py
```diff
@@ -173,13 +175,34 @@ def lint(session: nox.sessions.Session):
...
+
+    # Ensure black and isort only perform a status check
+    # with a custom pre-commit configuration.
+    hooks = []
+    for entry in config["repos"]:
+        hook = Path(urlsplit(unquote(entry["repo"])).path).name
+        hooks.append(hook)
+        if hook in ["black", "isort"]:
+            # Enforce a "--check" for the hook.
+            entry["hooks"][0]["args"].insert(0, "--check")
```
trexfeathers, line 195: Previously, when we just had Black, it worked perfectly well to allow a 'full' Black operation, since that raises errors anyway and the file modification was isolated to the Cirrus instance. I'd much prefer to go back to that if possible, since it doesn't require this extra complexity. Is there a problem with this now we're also using isort?

bjlittle: For CI we definitely only want to do a `--check` for both `black` and `isort` across all files.

There is no change in behaviour here for CI for this PR. We want `pre-commit` to return a status code failure for both the `black` and `isort` hooks when a non-compliance is detected, not for them to fix the non-compliancies, otherwise failures will go unnoticed.

However, the `black` and `isort` `pre-commit` hooks will automagically fix non-compliances for the developer on `git commit`.

trexfeathers: >We want `pre-commit` to return a status code failure for both the `black` and `isort` hooks when a non-compliance is detected

That's also achievable without the `--check` option, in which case there's no need for the YAML editing. I appreciate that without `--check` it will also modify the files, but that's irrelevant since it's on a Cirrus machine.

trexfeathers: I believe a key distinction here is that **pre-commit** raises exit code `1` on modification, even if the tool (**Black** or **isort**) does not.

bjlittle: Okay, so I've verified the exact behaviour here, and `pre-commit` without the `--check` argument to both `black` and `isort` fixes the files but **also** returns a non-zero return code for the `pre-commit run --all-files` use case.

So I'm happy that it's safe not to inject the `--check` argument but still get the same behaviour i.e., a non-compliance failure.

Also, I guess if the developer is using `nox` to lint the code, then they get the same behaviour if they'd installed `pre-commit` as part of their workflow... so that now aligns.

Parsing the `pre-commit` config file still has value as we can now add or remove hooks and the `lint` task behaves appropriately, rather then only running fixed hooks from a static list - which can go stale.

noxfile.py
```diff
@@ -173,13 +174,25 @@ def lint(session: nox.sessions.Session):
...
+
+    # Load the pre-commit configuration YAML file.
+    with open(".pre-commit-config.yaml", "r") as fi:
+        config = yaml.load(fi, Loader=yaml.FullLoader)
+
+    # Enumerate the pre-commit hooks.
+    hooks = [
+        Path(urlsplit(unquote(entry["repo"])).path).name
+        for entry in config["repos"]
+    ]
```
trexfeathers, line 190: It it safe to depend on the URL? Presumably pre-commit is actually attempting to match your `hook` argument below to an `id` value under `hooks`?

bjlittle: TBH I'm not totally mad keen on parsing a URL, although it is pretty easy to do and I think stable enough for our needs.

I'll have a play and drill down to the `id`, but some repos provide more than one hook, but that's shouldn't be a blocker... let's see :+1:

bjlittle: Yeah, I think we've tended towards a nice approach now... happy to go with that now?

trexfeathers: > but some repos provide more than one hook

Yeah, that's unfortunate, but I think it's a pre-commit API limitation rather than something we can solve with the right incantation. The final argument is [specifying a single hook id](https://pre-commit.com/#pre-commit-run), so we're already stuck running just one hook whether we extract the hook's id from a URL or from somewhere else.

bjlittle: The `id` is guaranteed per hook, and `alias` is optional, so I think this is pretty robust, https://pre-commit.com/#pre-commit-configyaml---hooks

noxfile.py
```diff
@@ -173,13 +173,24 @@ def lint(session: nox.sessions.Session):
...
-    cmd = ["pre-commit", "run", "--all-files"]
-    hooks = ["black", "blacken-docs", "flake8", "isort"]
-    for hook in hooks:
-        session.run(*cmd, hook)
+
+    # Load the pre-commit configuration YAML file.
+    with open(".pre-commit-config.yaml", "r") as fi:
+        config = yaml.load(fi, Loader=yaml.FullLoader)
+
+    # Enumerate singleton pre-commit hooks.
```
trexfeathers, line 185: Please can you stick a comment on explaining _why_ we're limiting ourselves to singletons? Thanks

trexfeathers (changes requested):

trexfeathers (approved):

noxfile.py
bjlittle: Done :+1: 

That said... (sorry for stringing this out :stuck_out_tongue_winking_eye:) looking at the `pre-commit/pre-commit-hooks`, they all actually seem like reasonable hooks to run except for `no-commit-to-branch`.

We could simply opt to have a hook id exclude list, just containing `no-commit-to-branch`, and run everything but that hook. It'll get us away for the weak-ish approach of only including `id`s from a `repo` with **one** `hook`.

Whay say ye?

trexfeathers: I'm up for that, particularly given you've said that we have some devs who rely on the Nox session for their checks. Not sure if the correct session name would still be `lint` in that case! I'll hold off merging.

bjlittle:
@trexfeathers Right, I'm done now, honest... I'm stepping away from the keyboard
:runner: :keyboard: :rofl:

rcomer:
Thanks guys 🤩

bjlittle:
@trexfeathers Awesome, thanks :clinking_glasses:

PR #4241, opened 2021-07-13
Explicit metadata manager classes.

pp-mo:
This is intended as an alternative to #4227

At present, each metadata 'manager' object is created in its own unique class by the `iris.common.metadata.metadata_manager_factory` factory function.
But _in practice,_ there is only one such class definition for each type of metadata class (aka BaseMetadata subclass),
( actually, only 5 of these ).  All the manager types are identical re--creations of those.

#4227 notes this as performance problem, and adds caching to avoid re-creating a manager class each time.
So, effectively, this just creates 5 reference classes + re-uses them.

Quite apart from performance issues, I think that the way managers all have unique types is already pretty odd, and also the code is rather unusual and hard to follow.
As I think @bsherratt also suggests [there](https://github.com/SciTools/iris/pull/4227#pullrequestreview-700776584)

The _**reason**_ for the code complexity is the effort to avoid boilerplate repetition in the BaseMetadata class definitions and the related manager classes.

So, I feel that maybe a problem with #4277 is that the caching adds yet another "layer" to the code.
However, the basic performance problem it solves just goes back to defining each of the 5 manager classes once, instead of creating them dynamically.

Seeing that, and carefully considering the code + @bsherratt comments, I came to the conclusion that we can replace the factory approach with a more conventional class definition for each of the "manager classes".  
So that's what this attempts to do.

I'm keen to see if we can do that in a more explicit and easy-to-follow way, without re-introducing too much repetition or boilerplate.

N.B. the appropriate manager class is then referenced directly by containing objects, instead of calling the factory function.
( see the small changes in cube.py / coords.py / aux_factory.py )

### DIFF
--- a/lib/iris/aux_factory.py
+++ b/lib/iris/aux_factory.py
@@ -15,11 +15,8 @@ import cf_units
 import dask.array as da
 import numpy as np
 
-from iris.common import (
-    CFVariableMixin,
-    CoordMetadata,
-    metadata_manager_factory,
-)
+from iris.common import CFVariableMixin
+from iris.common.metadata import CoordMetadataManager
 import iris.coords
 
 
@@ -40,7 +37,7 @@ class AuxCoordFactory(CFVariableMixin, metaclass=ABCMeta):
     def __init__(self):
         # Configure the metadata manager.
         if not hasattr(self, "_metadata_manager"):
-            self._metadata_manager = metadata_manager_factory(CoordMetadata)
+            self._metadata_manager = CoordMetadataManager()
 
         #: Descriptive name of the coordinate made by the factory
         self.long_name = None
@@ -390,7 +387,7 @@ class HybridHeightFactory(AuxCoordFactory):
 
         """
         # Configure the metadata manager.
-        self._metadata_manager = metadata_manager_factory(CoordMetadata)
+        self._metadata_manager = CoordMetadataManager()
         super().__init__()
 
         if delta and delta.nbounds not in (0, 2):
@@ -579,7 +576,7 @@ class HybridPressureFactory(AuxCoordFactory):
 
         """
         # Configure the metadata manager.
-        self._metadata_manager = metadata_manager_factory(CoordMetadata)
+        self._metadata_manager = CoordMetadataManager()
         super().__init__()
 
         # Check that provided coords meet necessary conditions.
@@ -788,7 +785,7 @@ class OceanSigmaZFactory(AuxCoordFactory):
 
         """
         # Configure the metadata manager.
-        self._metadata_manager = metadata_manager_factory(CoordMetadata)
+        self._metadata_manager = CoordMetadataManager()
         super().__init__()
 
         # Check that provided coordinates meet necessary conditions.
@@ -1093,7 +1090,7 @@ class OceanSigmaFactory(AuxCoordFactory):
 
         """
         # Configure the metadata manager.
-        self._metadata_manager = metadata_manager_factory(CoordMetadata)
+        self._metadata_manager = CoordMetadataManager()
         super().__init__()
 
         # Check that provided coordinates meet necessary conditions.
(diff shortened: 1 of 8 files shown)


bsherratt:
Turning out to be quite a tricky bit of code isn't it...

Just to clarify, my comment wasn't against using `type` to create a class. It's certainly unconventional so it does add to the complexity, but not necessarily the first thing I'd think to change. After all, a `class` block *is* literally syntactic sugar for `type(name, bases, namespace)` (well, plus some magic to avoid creating a closure over the class body).

I was just observing that we only need one Manager class per Metadata class, but the current code creates many more than that (one manager class per required manager instance). And furthermore that we could easily enumerate all the Managers and expose those in the module namespace instead of the factory. (Hence the code snippet of what that could look like)

> we can replace the factory approach with a more conventional class definition for each of the "manager classes"
> I'm keen to see if we can do that in a more explicit and easy-to-follow way, without re-introducing too much repetition or boilerplate.

I think these are mutually exclusive. If there isn't a factory function that takes a Metadata class and returns a Manager class, then the only alternative is that every new metadata class requires a corresponding incantation
```python
@manager_class
class SomeNewMetadataManager(MetadataManager):
    metadata_class = SomeNewMetdata
```
It's not *much* boilerplate, but if it's valuable to easily be able to add new metadata classes, then surely it's already too much?

pp-mo:
> @bsherratt If there isn't a factory function ...  every new metadata class requires a corresponding incantation
> ... It's not *much* boilerplate ...

Of course you're right.  My feeling was, it looks a bit clunky but there is virtually nothing there to get *wrong*.
( Given that, when creating + using a new metadata type, you are not very likely to miss the fact that you need a new manager type, as the [container boilerplate for that]() is now pretty obvious )

My first commit here did retain the factory approach, and the [manager classes creation there](https://github.com/SciTools/iris/blob/2056b1cd601aeb30572de4cccc8f21363271473c/lib/iris/common/metadata.py#L1464-L1472) is notably simpler.

It's a plain choice.
But after consideration, I think this (later) way is less clever, less pretty, but just that much simpler to understand (and therefore maintain).

pp-mo:
Another point worth noting : we removed the need for the setstate/getstate/reduce to make picking work, which is a definite win.
That works whether or not you use the original 
`manager_class = factory_class(metadata_class)` approach, 
or the `@decorator: class MyManager(MetadataManager): metadata_class=MyMeta" one.

FWIW 
I did try to replace the factory/decorator approach with an embedded class definition
(i.e. a "class ..." within the factory call, replacing the 'type' usage)
But that re-introduces the pickling difficulty (it doesn't associate the generated class directly with the module).
I'd love to get rid of that, I expect there's a way, but I couldn't work out how.

bsherratt:
Alright, I'll admit you intrigued me with the pickling issue

So it turns out that pickling custom objects requires that its type's [`__qualname__`](https://docs.python.org/3/library/stdtypes.html#special-attributes) resolves to the type object.

It also turns out that there's another little bit of magic that a `class` block has over `type`: it sets that `__qualname__` by appending `__name__` to the qualname of the current scope. `type` just leaves it the same as `__name__`. But each function call has its own scope, called eg `funcname.<locals>`, so the class qualname is nothing useful and it can't be (automatically) pickled.

I suppose messing with `__qualname__` could open up any other wacky combination we wanted (`MyMetadata.Manager` anyone?) but that's clearly even more of a maintenance headache.

bsherratt:
I didn't have much else to add by the way, I'm happy with "clunky but obvious" if everyone else is. Or indeed #4227, but for me I think this wins (sorry!)

bjlittle (changes requested):
@pp-mo You need to add performance metrics for this PR akin to #4227 so that we can compare and contrast.

So basically, you've effectively replaced a one line change with this whole PR. You've also exposed manager classes publicly, which we really shouldn't do. Everything should be private, akin to the data manager. Also you've now forced several extra steps as a maintenance burden on the developer i.e., if this PR is merged to `main`, and then you update the `mesh-data-model` feature branch, you'll break it and you'll have work to do to fix it. Just sayin.

Whereas before this all wasn't a concern.

IMHO what this all comes down to is essentially performance, regardless of whether you want to be *idiomatic* or *convoluted*, balanced along with the cost of ongoing maintenance. 

Anyways hopefully you can publish some accompanying performance metrics, then we can make a judgement call on this one.

Nice one 👍

PR #4246, opened 2021-07-21
Fix aggregated_by mdtol/masked constant problem

### LINKED ISSUE #3190: problem with aggregated_by and mdtol and masks with iris2.1 and numpy1.15
When I run these commands in python2.7 with iris 2.1 and numpy 1.15, I get the error messages in traceback below. 

PYTHON commands:
```python
import iris
cube = iris.load_cube('example.nc')
#check to see if the data is there:
cube.data
cube = cube.aggregated_by(["year", "day_of_year"], iris.analysis.MEAN, mdtol=0.03)
```

The 20MB example.nc NETCDF file is available here: 
https://drive.google.com/open?id=1hZNWXs7wOtBq2dXWM1zbAJnVMjy_DXc7

```python-traceback
  File "test10.py", line 10, in <module>
    cube = cube.aggregated_by(["year", "day_of_year"], iris.analysis.MEAN, mdtol=0.03)
  File "/Users/pmcguire/fluxnet1/iris/cube.py", line 3414, in aggregated_by
    **kwargs)
  File "/Users/pmcguire/fluxnet1/iris/analysis/__init__.py", line 524, in aggregate
    result.mask = result.mask | mask_update
  File "/Users/pmcguire/.local/lib/python2.7/site-packages/numpy/ma/core.py", line 6339, in __setattr__
    "attributes of {!r} are not writeable".format(self))
AttributeError: attributes of masked are not writeable
```
-------------------
I have tracked this down I think.

I could get test_fluxnet_evaluation.py to work if I comment out the three lines below from the _Aggregator(object).aggregate method in iris/analysis/__init__.py (near line 514) of a copy of iris2.1. 
The result.mask below was not defined for some reason even though ma.isMaskedArray(result) is true.
```python
      result = self.call_func(data, axis=axis, **kwargs)
        if (mdtol is not None and ma.isMaskedArray(data)):
            fraction_not_missing = data.count(axis=axis) / data.shape[axis]
            mask_update = 1 - mdtol > fraction_not_missing
            #if ma.isMaskedArray(result):
            #    result.mask = result.mask | mask_update
            #else:
            result = ma.array(result, mask=mask_update)
```
In version Iris1.13, this problem does not exist.
I can reproduce this problem if I upgrade from Iris1.13 to Iris2.2.0dev0, including all dependencies.

But then when I downgrade only Iris2.2 -> Iris.1.13, keeping the updated dependencies, the bug persists.
But then when I downgrade only Iris2.2 -> Iris.1.13 and numpy1.15->numpy1.12.1, keeping the updated dependencies, the bug goes away.
So I think it's something in numpy.

When it works with Iris1.13 and numpy1.12.1, Aggregrate (with iris.analysis.Mean) function tries to override the (non-array) mask value for the day, and it can do this, because the mask is writeable. But when it doesn't work with iris2.1 and numpy1.15, this function fails because this mask is not writeable.

2)A quick perusal of the numpy1.15 documentation suggests that they changed the mask behavior: https://docs.scipy.org/doc/numpy/release.html
"np.ma.masked is no longer writeable:
Attempts to mutate the masked constant now error, as the underlying arrays are marked readonly. In the past, it was possible to get away with...".
I am not sure if this is the change which is affecting us or not.

Can someone fix this problem in iris? Should I just maintain a separate iris branch (with those three lines commented out) for my own purposes till this is fixed? Or does anybody have a better solution? I am not sure if I should downgrade numpy or not.

Thanks,
Patrick McGuire

Department of Meteorology and Department of Geography & Environmental Science
University of Reading
Personal Web Page: http://www.personal.reading.ac.uk/~gn916173

rcomer:
### Description

Fixes #3190.  Note that the examples provided there all involve 1D cubes and by the time `Aggregator.aggregate` is called, `Cube.aggregated_by`  has sliced out and passed the array for one particular section.  So the problem boils down to the issue illustrated in the new test:  if you average an entirely masked array, you get the `ma.masked` constant, whose `mask` attribute is read-only.  In this situation, there is no point doing anything extra for `mdtol` anyway, so just skip the loop.

I have verified that the three examples from #3190 all work with this change.

---

### DIFF
--- a/lib/iris/analysis/__init__.py
+++ b/lib/iris/analysis/__init__.py
@@ -582,7 +582,11 @@ class _Aggregator:
         mdtol = kwargs.pop("mdtol", None)
 
         result = self.call_func(data, axis=axis, **kwargs)
-        if mdtol is not None and ma.isMaskedArray(data):
+        if (
+            mdtol is not None
+            and ma.is_masked(data)
+            and result is not ma.masked
+        ):
             fraction_not_missing = data.count(axis=axis) / data.shape[axis]
             mask_update = 1 - mdtol > fraction_not_missing
             if ma.isMaskedArray(result):
--- a/lib/iris/tests/unit/analysis/test_Aggregator.py
+++ b/lib/iris/tests/unit/analysis/test_Aggregator.py
@@ -156,6 +156,19 @@ class Test_aggregate(tests.IrisTest):
             self.assertArrayAlmostEqual(result, mock_return.copy())
         mock_method.assert_called_once_with(data, axis=axis)
 
+    def test_allmasked_1D_with_mdtol(self):
+        data = ma.masked_all((3,))
+        axis = 0
+        mdtol = 0.5
+        mock_return = ma.masked
+        with mock.patch.object(
+            self.TEST, "call_func", return_value=mock_return
+        ) as mock_method:
+            result = self.TEST.aggregate(data, axis, mdtol=mdtol)
+
+        self.assertIs(result, mock_return.copy())
+        mock_method.assert_called_once_with(data, axis=axis)
+
     def test_returning_scalar_mdtol(self):
         # Test the case when the data aggregation function returns a scalar and
         # turns it into a masked array.
--- a/lib/iris/tests/unit/analysis/test_Aggregator.py
+++ b/lib/iris/tests/unit/analysis/test_Aggregator.py
@@ -166,7 +166,7 @@ class Test_aggregate(tests.IrisTest):
         ) as mock_method:
             result = self.TEST.aggregate(data, axis, mdtol=mdtol)
 
-        self.assertIs(result, mock_return.copy())
+        self.assertIs(result, mock_return)
         mock_method.assert_called_once_with(data, axis=axis)
 
     def test_returning_scalar_mdtol(self):
--- a/docs/src/whatsnew/dev.rst
+++ b/docs/src/whatsnew/dev.rst
@@ -42,6 +42,10 @@ This document explains the changes made to Iris for this release
    :func:`iris.plot.plot` no longer defaults to placing a "Y" coordinate (e.g.
    latitude) on the y-axis of the plot. (:issue:`4493`, :pull:`4601`)
 
+#. `@rcomer`_ fixed :meth:`~iris.cube.Cube.aggregated_by` with `mdtol` for 1D
+   cubes where an aggregated section is entirely masked, reported at
+   :issue:`3190`.  (:pull:`4246`)
+
 
 💣 Incompatible Changes
 =======================


lib/iris/analysis/__init__.py
```diff
@@ -582,7 +582,11 @@ def aggregate(self, data, axis, **kwargs):
         mdtol = kwargs.pop("mdtol", None)
 
         result = self.call_func(data, axis=axis, **kwargs)
-        if mdtol is not None and ma.isMaskedArray(data):
+        if (
+            mdtol is not None
+            and ma.is_masked(data)
```
rcomer, line 587: Changing from `isMaskedArray` to `is_masked` isn’t part of the fix, but seems like the better check: if there is an all-False mask, then the loop won’t have any effect, so more efficient to just skip it.

lib/iris/tests/unit/analysis/test_Aggregator.py
```diff
@@ -156,6 +156,19 @@ def test_unmasked(self):
             self.assertArrayAlmostEqual(result, mock_return.copy())
         mock_method.assert_called_once_with(data, axis=axis)
 
+    def test_allmasked_1D_with_mdtol(self):
+        data = ma.masked_all((3,))
+        axis = 0
+        mdtol = 0.5
+        mock_return = ma.masked
+        with mock.patch.object(
```
wjbenfold, line 164: Iris.Test has a patch built in, so you can use `mock_method = self.patch.object(...` I think (slight benefit of reduced indenting)

wjbenfold (commented):
Looks good :) Made a small suggestion and have a testing suggestion below (but only if I've understood the problem properly).

Is it worth testing that a 2d array with all of one "row" masked (where the "row" runs in the direction of `axis`) is successfully aggregated (and thus that it's handling `ma.masked` in the code that calls `aggregate`)?

lib/iris/tests/unit/analysis/test_Aggregator.py
rcomer: Thanks, I didn't know about this method.  However I can't seem to get it to work in this case.  From the docstring, the method seems to want to patch the class rather than the instance, so I tried
```python
mock_method = self.patch("iris.analysis.Aggregator.call_func", return_value=mock_return)
```
but got 
```
AttributeError: <class 'iris.analysis.Aggregator'> does not have the attribute 'call_func'
```
because `call_func` is added to the aggregator at instantiation.

I also tried patching the instance by
```python
mock_method = self.patch("self.TEST.call_func", return_value=mock_return)
```
but that gave me
```
ModuleNotFoundError: No module named 'self'
```

rcomer (commented):
Thanks for the review @wjbenfold.  I don't think we need to worry about the 2D case, as the masked constant (`ma.masked`) is only returned when you completely collapse an array.  If you partially collapse a masked array you still get a masked array, which can be manipulated in the way the loop was designed for  E.g.

```
In [1]: import numpy.ma as ma

In [3]: arr = ma.array([[1, 2], [3, 4]], mask=[[1, 1], [1, 1]])

In [4]: arr
Out[4]: 
masked_array(
  data=[[--, --],
        [--, --]],
  mask=[[ True,  True],
        [ True,  True]],
  fill_value=999999,
  dtype=int64)

In [5]: arr.mean(axis=1)
Out[5]: 
masked_array(data=[--, --],
             mask=[ True,  True],
       fill_value=1e+20,
            dtype=float64)

In [6]: arr.mean()
Out[6]: masked

```

Or have I misunderstood what you're saying?

lib/iris/tests/unit/analysis/test_Aggregator.py
wjbenfold: Huh, turns out our implementation of `patch` doesn't have an `object` attribute, whereas `unittest.mock.patch` does. Seems good to leave this as it is then.

wjbenfold (approved):
> I don't think we need to worry about the 2D case, as the masked constant (`ma.masked`) is only returned when you completely collapse an array

I was worried about

> by the time Aggregator.aggregate is called, Cube.aggregated_by has sliced out and passed the array for one particular section.

and that that might mean that `aggregated_by` is expecting back an array but gets an `ma.masked`?

~~I remembered there being something in the aggregation higher up that would do a 2D aggregation by doing one axis and then the other, (I think `cube.aggregated_by` does that) and was concerned that depending how the rows were combined it might be odd. You're right though that it's just a matter of applying to an n-d array of all the data, so I'm happy that this is fine without an extra test.~~

rcomer:
> > I don't think we need to worry about the 2D case, as the masked constant (`ma.masked`) is only returned when you completely collapse an array
> 
> I was worried about
> 
> > by the time Aggregator.aggregate is called, Cube.aggregated_by has sliced out and passed the array for one particular section.
> 
> and that that might mean that `aggregated_by` is expecting back an array but gets an `ma.masked`?
> 

My understanding is that `aggregated_by` will slice along the aggregation axis, so if you start with an array that is (5, 20) and aggregate along dimension 1, it will pass a series of arrays that are (5, x).  Then the aggregator passes back an array of shape (5,) for each.

PR #4249, opened 2021-07-22
Test wrangling vs v3.0.x

rcomer:
### Description

Replacement for #4192, targeting the v3.0.x branch.

This PR gets our tests running consistently under `pytest` and `nose`.  We had some `__init__.py` files missing so some tests have not been running under `nose`/`setuptools`/`nox`.  Once reinstated, three of the tests for `PartialDateTime` failed, so this PR also includes changes to address that.  See inline comments for explanations of individual changes.

I have verified that pointing `pytest` at this new branch produces 5216 passess and 22 skips.  `python setup.py test` also finds 5148 tests, but some are falling over with memory allocation errors 😕.  So let's see how the CI fairs...

---

### DIFF
--- a/lib/iris/tests/system_test.py
+++ b/lib/iris/tests/system_test.py
@@ -22,7 +22,7 @@ import iris.tests as tests
 
 
 class SystemInitialTest(tests.IrisTest):
-    def system_test_supported_filetypes(self):
+    def test_supported_filetypes(self):
         nx, ny = 60, 60
         data = np.arange(nx * ny, dtype=">f4").reshape(nx, ny)
 
@@ -74,7 +74,7 @@ class SystemInitialTest(tests.IrisTest):
                 new_cube, ("system", "supported_filetype_%s.cml" % filetype)
             )
 
-    def system_test_imports_general(self):
+    def test_imports_general(self):
         if tests.MPL_AVAILABLE:
             import matplotlib  # noqa
         import netCDF4  # noqa
--- a/lib/iris/tests/test_merge.py
+++ b/lib/iris/tests/test_merge.py
@@ -25,7 +25,7 @@ import iris.exceptions
 import iris.tests.stock
 
 
-class TestMixin:
+class MergeMixin:
     """
     Mix-in class for attributes & utilities common to these test cases.
 
@@ -55,7 +55,7 @@ class TestMixin:
 
 
 @tests.skip_data
-class TestSingleCube(tests.IrisTest, TestMixin):
+class TestSingleCube(tests.IrisTest, MergeMixin):
     def setUp(self):
         self._data_path = tests.get_data_path(("PP", "globClim1", "theta.pp"))
         self._num_cubes = 1
@@ -63,7 +63,7 @@ class TestSingleCube(tests.IrisTest, TestMixin):
 
 
 @tests.skip_data
-class TestMultiCube(tests.IrisTest, TestMixin):
+class TestMultiCube(tests.IrisTest, MergeMixin):
     def setUp(self):
         self._data_path = tests.get_data_path(
             ("PP", "globClim1", "dec_subset.pp")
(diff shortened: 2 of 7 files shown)


lib/iris/tests/system_test.py
```diff
@@ -22,7 +22,7 @@
 
 
 class SystemInitialTest(tests.IrisTest):
-    def system_test_supported_filetypes(self):
+    def test_supported_filetypes(self):
```
rcomer, line 25: `pytest` requires test names to start with "test" or it just won't discover them.

lib/iris/tests/system_test.py
```diff
@@ -74,7 +74,7 @@ def horiz_cs():
                 new_cube, ("system", "supported_filetype_%s.cml" % filetype)
             )
 
-    def system_test_imports_general(self):
+    def test_imports_general(self):
```
rcomer, line 77: `pytest` requires test names to start with "test" or it just won't discover them

lib/iris/tests/test_merge.py
```diff
@@ -25,7 +25,7 @@
 import iris.tests.stock
 
 
-class TestMixin:
+class MergeMixin:
```
rcomer, line 28: `pytest` tried to run this as a test class, so errored all three tests when it couldn't find `_data_path`. Removing "Test" from the class name stops `pytest` trying to do that.

lib/iris/tests/unit/time/test_PartialDateTime.py
```diff
@@ -224,12 +224,12 @@ def setUp(self):
         self.expected_value = EQ_EXPECTATIONS
 
     def test_cftime_equal(self):
-        pdt = PartialDateTime(month=3, microsecond=2)
+        pdt = PartialDateTime(month=3, second=2)
```
rcomer, line 227: This was failing because the equality operator returned `False`. The microsecond comparison is skipped if `other` doesn't have a microsecond attribute:
https://github.com/SciTools/iris/blob/a6cd220f64f7d2fd2af8308f814d81b00eb1e05e/lib/iris/time.py#L152-L157

My guess is that, when this test was written, `cftime.datetime`s' predecessors did not have a microsecond attribute, but they do now.

lib/iris/time.py
```diff
@@ -155,23 +155,21 @@ def __eq__(self, other):
...
             except AttributeError:
-                result = NotImplemented
+                result = other.__eq__(self)
+                if result is NotImplemented:
+                    # Equaliy is undefined between these objects.  We don't
+                    # want Python to fall back to the default `object`
+                    # behaviour (which compares using object IDs), so we raise
+                    # an exception here instead.
+                    fmt = "unable to compare PartialDateTime with {}"
+                    raise TypeError(fmt.format(type(other)))
```
rcomer, line 167: We previously used the `__cmp__` method to stop the equality comparison falling back to object identities.  But this method is [not a thing in python3](https://docs.python.org/3/whatsnew/3.0.html#ordering-comparisons).

My above solution works, but feels like a bit of a hack. Alternative solutions welcome!

bjlittle: @rcomer Rings a bell... see https://github.com/Unidata/cftime/pull/81 :wink: 

This might explain this change in behaviour :thinking:

rcomer: @bjlittle yup the failures I saw were just for `==` and `!=` operators.  So I think `>=`, etc. were still raising the `TypeError` as there's no way to check that based on identities.

bjlittle: @rcomer Okay, I'm fine with this.

It seems completely reasonable to me and maintains prior behaviour both within `iris` and for users :+1:

bjlittle (changes requested):
@rcomer Looks good to me.

I'm going to hold this PR until #4222 lands in the `v3.0.x` release feature branch. 

At which point please:
* rebase with the latest `v3.0.x`
* add a `whatsnew` entry.

You're blocked until #4222 is merged, then we're good to get this across the line :+1:

bjlittle (approved):
@rcomer LGTM, thanks!

I'll merge when `cirrus-ci` is happy :+1:

pp-mo:
This is really awesome work, long overdue, and I hope means we can get serious about migrating to PyTest.

Praise @bjlittle @rcomer 💐 🍻

Issue #4211, opened 2021-06-24
Review/revise ordering of cube-summary sections


pp-mo:
## 📰 Custom Issue
Following after #4206 (and before updating for unstructured cubes ?)
I think that it would at least be better to have "Attributes" last, and "Cell Methods" further up.
See [example here](https://github.com/SciTools/iris/blob/a6669092e11c5d02eb33d835a46b6a96c7388363/lib/iris/cube.py#L741-L757)

PR #4251, opened 2021-07-22
Make cube summary list attributes last.

pp-mo:
### Description
Proposing to move the position of cell-methods in a Cube summary.

At present, if present, cell-methods appear last.
But attributes are a much more common content, and they include global attributes, which **_always_** appear last in an ncdump.
So I think it's less confusing if attributes always appear last in a cube summary too.

**If we want to do this, we should definitely include it alongside the new-cube summary in #4206**

Closes : #4211

---

### DIFF
--- a/lib/iris/_representation.py
+++ b/lib/iris/_representation.py
@@ -311,7 +311,7 @@ class CubeSummary:
             "Scalar cell measures:",
             scalar_cell_measures,
         )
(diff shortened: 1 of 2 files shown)


bjlittle:
@pp-mo see #4252

pp-mo:
Merged this change into #4206  instead

PR #4252, opened 2021-07-22
rename cirrus-ci tasks with name override

bjlittle:
### Description
This PR addresses `cirrus-ci` parsing issues, which appear to have resulted from a recent update on the `cirrus-ci` server-side.

cirrus-ci

---

### DIFF
--- a/.cirrus.yml
+++ b/.cirrus.yml
@@ -105,7 +105,7 @@ iris_test_data_template: &IRIS_TEST_DATA_TEMPLATE
 #
 # Linting
 #
-precommit_task:
+task:
   only_if: ${SKIP_LINT_TASK} == ""
   << : *CREDITS_TEMPLATE
   auto_cancellation: true
@@ -129,7 +129,7 @@ precommit_task:
 #
 # Testing (Linux)
 #
-test_task:
+task:
   only_if: ${SKIP_TEST_TASK} == ""
   << : *CREDITS_TEMPLATE
   matrix:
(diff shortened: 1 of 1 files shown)


.cirrus.yml
```diff
@@ -129,7 +129,7 @@ precommit_task:
 #
 # Testing (Linux)
 #
-test_task:
+task:
```
pp-mo, line 132: I know this is what the Cirrus warning suggests, but it makes me uncomfortable because it is not really good YAML -- because you probably shouldn't use duplicate mapping keys.  
For details, see [this discussion](https://github.com/SciTools/iris-grib/pull/283#issuecomment-874570061) ...

So I'd love to find a way that **_doesn't_** use a repeated key, if we could ?

bjlittle: Oddly, that's precisely what we had before... it's almost like `cirrus-ci`s YAML parser is complaining unnecessarily

bjlittle: So it seems like YAML, but YAML according to `cirrus-ci` :thinking:

pp-mo: > Oddly, that's precisely what we had before

I'm sure they changed it.

> YAML according to cirrus-ci

My reading, yes.
I couldn't work out whether they are actually **_using_** https://github.com/cirruslabs/yaml.
If so not sure why, as they don't seem to have changed anything significant.  Perhaps it is just a "pin".
It seems that the [parent original](https://github.com/go-yaml/yaml) has removed the API that let you use it in this way, but they are [reconsidering whether to reintroduce something similar](https://github.com/go-yaml/yaml/pull/484)

pp-mo: I agree, as Cirrus recommend this, we have to believe in it.
https://cirrus-ci.org/guide/writing-tasks/

pp-mo (approved):
Maybe not ideal, but it works + agrees with Cirrus' way on it.

PR #4253, opened 2021-07-22
CI Performance Testing

jamesp:
### Description
An MVP for CI performance testing PRs.

Things in this PR:

#### `benchmarks/benchmarks/*`
These are directly lifted from the offline `metrics` repository.  No edits made.

#### `benchmarks/conda_lock_plugin.py`
This is a low-touch change to `asv.plugins.conda.Conda` to use a lockfile as the environment source. 
The trickiness here is that `asv` doesn't really have great support for per-commit environment changes.
It wants to make one environment and then run a whole load of commits in this same environment.  
So this environment plugin moves the setting of environment packages to the build stage of testing
a specific commit. 
This may have a poor interaction with parallel processing of environments.

#### `.github/workflows/benchmark.yml`
This workflow is configured to run on pull requests.
It runs asv on the head of the pull request and the head of the target branch.  
The results are compared and output as an artifact from the run.  i.e. you can download the `.asv/results`
folder and interrogate in more detail if needed.
If the comparison finds performance degradation that hits ASV's criteria (currently default criteria)
then the run is marked as failed and will show up in the pull request checks section as a failed check.
See https://github.com/jamesp/iris/pull/9 for an example of this.
If there are no changes above the threshold, the check is a success.

---


benchmarks/conda_lock_plugin.py
```diff
@@ -0,0 +1,48 @@
...
+
+"""
+from asv.console import log
+from asv.plugins.conda import Conda, _find_conda
+
+class CondaLock(Conda):
+    """
+    Create the environment based on a **version-controlled** lockfile.
+
+    Creating the environment instance is deferred until ``install_project`` time,
```
trexfeathers, line 13:
```suggestion
    Creating the environment instance is deferred until ``build_project`` time,
```

.github/workflows/benchmark.yml
```diff
@@ -0,0 +1,70 @@
...
+    runs-on: ubuntu-latest
+
+    # Steps represent a sequence of tasks that will be executed as part of the job
+    steps:
+      # Checks-out your repository under $GITHUB_WORKSPACE, so your job can access it
+      - uses: actions/checkout@v2
+        with:
+          fetch-depth: 0
+
+      # This is too complicated to get working for now.
```
trexfeathers, line 27: Is [the way Cirrus does this](https://cirrus-ci.org/guide/tips-and-tricks/#custom-clone-command) any help/inspiration?

.github/workflows/benchmark.yml
```diff
@@ -0,0 +1,70 @@
...
+      # - name: Checkout PR and target branch
+      #   run: |
+      #     git fetch --no-tags --depth=1 origin ${{ github.base_ref }}
+      #     git fetch --no-tags --depth=1 origin ${{ github.head_ref }}
+
+      - name: Setup asv
+        run: |
+          pip install asv
+          cd benchmarks
+          asv machine --yes --machine gh-actions
```
trexfeathers, line 40: I'm nervous about `--machine gh-actions`. It would be good to make it clear that a set of results come from a specific GHA run, given the likelihood that GHA's `ubuntu-latest` spec will evolve over time, and the likelihood that we will subject results to increased scrutiny over time (e.g. through the ASV dashboard, or by downloading the `.asv/results` content.

So I'd prefer to use whatever machine name GHA provides. What do you think?

.github/workflows/benchmark.yml
```diff
@@ -0,0 +1,70 @@
...
+      - name: Run benchmarks on source and target
+        run: |
+          cd benchmarks
+          export BASE=origin/${{ github.base_ref }}
+          export HEAD=origin/${{ github.head_ref }}
+          # run asv benchmarks on the target
+          # -s 1 ensures only one "step" (the tip of the branch) is run
+          asv run $BASE -s 1
+          # run asv on the head of this PR
+          asv run $HEAD -s 1
```
trexfeathers, line 51:
```suggestion
          asv continuous $BASE $HEAD
```

This is designed specifically to error if regressions are found, so that you wouldn't need the `Fail if performance degraded` step.

Obviously some re-jigging would be needed so that we could still get `compare.txt`, or even to just have the raw `asv compare` output in the CI log - either seem valuable. But it would overall be simpler, and use ASV's intended API rather than parsing their text output.

.github/workflows/benchmark.yml
```diff
@@ -0,0 +1,70 @@
...
+            benchmarks/.asv/compare.txt
+
+      - name: Fail if performance degraded
+        run: |
+          cd benchmarks
+          cat .asv/compare.txt
+          if grep -q "Benchmarks that have got worse" .asv/compare.txt; then
+            echo "::error::Performance degradation.  See action artifact for full results."
+            exit 1
+          fi
```
trexfeathers, line 69: See the comment in the `Run benchmarks on source and target` step.

.github/workflows/benchmark.yml
```diff
@@ -0,0 +1,70 @@
...
+          asv run $HEAD -s 1
+          asv compare -s $BASE $HEAD > .asv/compare.txt
+
+      - name: Archive asv results
+        uses: actions/upload-artifact@v2
+        with:
+          name: asv-report
+          path: |
+            benchmarks/.asv/results
+            benchmarks/.asv/compare.txt
```
trexfeathers, line 60: I love this! Chapeau 🎩

trexfeathers (changes requested):
@jamesp thanks so much for taking this on. Makes me happy to see benchmarking moving forward. Some comments for you, and some stuff that won't fit as code annotations:

* [x] Also need to remove the [`tests/benchmarking` directory](https://github.com/SciTools/iris/tree/4850fda8d6faf1879b4cfd06f8bbcadc7cebe152/lib/iris/tests/benchmarking). The single benchmark in there is part of our offline suite, although we will need to sort out file loading before it come into this setup.

* [x] Could we add `.gitignore` and `pyproject.toml` exclusions for `benchmarks/.asv`? I've been burned before!

* [ ] What do you think of making a Nox session for this? Too early? I would want to eventually - avoids everyone having to know the correct `asv run ...` incantation if they want to run benchmarks manually.

* [ ] As discussed offline, it would be great if we could soon replicate the ASV environment plugin introduced in SciTools-incubator/iris-esmf-regrid#76, to truly defer environment management to a single common place. Something for me to capture in an issue once merged.

benchmarks/conda_lock_plugin.py
```diff
@@ -0,0 +1,48 @@
...
+            # been made before, otherwise there is no python to use to uninstall
+            super()._uninstall_project()
+            # TODO: we probably want to conda uninstall all the packages too
+            #       something like:
+            #       conda list --no-pip | sed /^#/d | cut -f 1 -d " " | xargs conda uninstall
+
+    def _setup(self):
+        # create the shell of a conda environment, that includes no packages
+        log.info("Creating conda environment for {0}".format(self.name))
+        self.run_executable(_find_conda(), ['create', "-y", '-p', self._path, '--force'])
```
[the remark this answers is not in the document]
jamesp: That is true, but I'm not sure that is a valid use case for this kind of environment.  a conda-lock file is explicit about which version of numpy it uses, it can't be easily changed out without re-resolving the whole environment.

I suppose the way to do that kind of testing would be to have a list of lockfiles which pin different sets of dependencies... but that seems like it's better served by the exisiting Conda or VirtualEnv approaches?

benchmarks/conda_lock_plugin.py
```diff
@@ -0,0 +1,48 @@
...
+    tool_name = "conda-lock"
+
+    def __init__(self, conf, python, requirements):
+        self._lockfile_path = conf.conda_lockfile
+        super().__init__(conf, python, requirements)
+
+    def _uninstall_project(self):
+        if self._get_installed_commit_hash():
+            # we can only run the uninstall command if an environment has already
+            # been made before, otherwise there is no python to use to uninstall
```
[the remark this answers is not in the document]
jamesp: The parent method expects there to be *some* version of python available in the environment.  But when starting with an empty conda env this is not true, so the default command will fail.

benchmarks/conda_lock_plugin.py
```diff
@@ -0,0 +1,48 @@
...
+            super()._uninstall_project()
+            # TODO: we probably want to conda uninstall all the packages too
+            #       something like:
+            #       conda list --no-pip | sed /^#/d | cut -f 1 -d " " | xargs conda uninstall
```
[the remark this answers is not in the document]
jamesp: > I assume you're thinking here about the small risk that a dependency might be removed between commits, perhaps if it were causing a problem, but it would not be removed from the ASV environment? Or does installing from a lock file mean that existing packages are removed?

Yes, exactly.

As it stands, this is how the environment management looks for running several commits, lets call them A->B, one after the other, each contains an updated lockfile e.b. B.lock has different packages than A.lock:

1. asv creates a new, empty conda environment `.asv/env/hash`, let's call this $ENV
2. asv checks out A.
3. asv does a `conda install --file A.lock` in $ENV.  A.lock is an explicit conda requirements file so conda doesn't resolve anything, it just installs all the listed pacakges into $ENV.
4. asv runs the benchmark for A and saves the output.
5. asv checks out B.
6. asv does a `conda install --file B.lock` in $ENV. **What happens to the packages in $ENV?** I think that if they have the same name, they are just overwritten.  Packages that were in A.lock but not in B.lock will not be removed, they just sit there.  This shouldn't be a problem, but it could be.  But I think if this was to cause and error, this should be picked up in testing, not in benchmarking.  What I'm suggesting is that we could be explicit here and clean down the environment before we do the next run.
7. asv runs benchmark for B and saves output.

I think you're right, if benchmarking could be improved to use the nox environment creation logic, this problem may go away.

benchmarks/conda_lock_plugin.py
```diff
@@ -0,0 +1,48 @@
...
+        log.info("Creating conda environment for {0}".format(self.name))
+        self.run_executable(_find_conda(), ['create', "-y", '-p', self._path, '--force'])
+
+    def _build_project(self, repo, commit_hash, build_dir):
+        # at "build" time, we build the environment from the provided lockfile
+        self.run_executable(_find_conda(), ["install", "-y", "-p", self._path, "--file", f"{build_dir}/{self._lockfile_path}"])
+        # this is set to warning as the asv.commands.run._do_build function
+        # explicitly raises the log level to WARN, and I want to see the environment being updated
+        # in the stdout log.
+        log.warning(f"Environment {self.name} updated to spec at {commit_hash[:8]}")
```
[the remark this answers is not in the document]
jamesp: Sure, I'm happy with that.

.github/workflows/benchmark.yml
jamesp: Thanks for the link, I'll play with a version of that and see if I can get it to work.  asv calls a variety of git commands when it is writing out the benchmarks and not all of them have been happy with a shallow clone.

.github/workflows/benchmark.yml
jamesp: I agree for storing long term results.  But this CI check only needs to be consistent with itself - everything is being thrown away at the end.

jamesp: But i'm happy to use the automatically ascribed name if that's simpler :+1:

.github/workflows/benchmark.yml
jamesp: how did I miss that...!!

benchmarks/conda_lock_plugin.py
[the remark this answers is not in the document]
trexfeathers: My idea would be to generate the environment from a lock file, then if any `matrix` is in `asv.conf.json`: install the required package as a second step.

[Here is an example](https://github.com/SciTools-incubator/iris-esmf-regrid/blob/ee8fdecfc8633208cab7c133dec98cef6a57a4d8/benchmarks/nox_asv_plugin.py#L210), which I _think_ I tested while writing it. Certainly wouldn't be the only way of achieving the idea.

benchmarks/conda_lock_plugin.py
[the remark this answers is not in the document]
trexfeathers: Thanks, glad I'm on the same page as you.

.github/workflows/benchmark.yml
trexfeathers: Yes I get that it's not currently a problem, I'm just worried about establishing an anti-pattern for later when we might forget or not be the ones expanding our ASV use.

.github/workflows/benchmark.yml
```diff
@@ -44,27 +44,19 @@ jobs:
...
           export HEAD=origin/${{ github.head_ref }}
-          # run asv benchmarks on the target
-          # -s 1 ensures only one "step" (the tip of the branch) is run
-          asv run $BASE -s 1
-          # run asv on the head of this PR
-          asv run $HEAD -s 1
+          asv continuous $BASE $HEAD
+
+      - name: Write stats to archive on failure
+        if: ${{ failure() }}
```
trexfeathers, line 50: I'm not worried either way, but I'm still curious: why only store and upload the results on a failure?

benchmarks/conda_lock_plugin.py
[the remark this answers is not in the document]
trexfeathers: My suggestion wasn't good enough here, as the comment is now out of sync with the code!

.github/workflows/benchmark.yml
jamesp: I'm happy to make this an always if there is utility in access to the data 👍

jamesp: This is now `always()` so results will be uploaded on both success and failure

jamesp:
Ok I think the checkout logic is way too naive, it will fail with branches and other forks.  I'll review it and come up with something better

jamesp:
ok @trexfeathers, I think this might be ready for final review.

trexfeathers:
I just took a look at `compare.txt` for the latest run. All the benchmarks for the 'previous' commit (`main` in this case) failed. Something to do with that checkout perhaps?

```
Benchmarks that have improved:

       before           after         ratio
     [149164ef]       [c1c7a17f]
           failed              n/a      n/a  aux_factory.FactoryCommon.time_create
           failed              n/a      n/a  aux_factory.FactoryCommon.time_return
           failed       14.6±0.1μs      n/a  aux_factory.HybridHeightFactory.time_create
           failed         81.4±2ns      n/a  aux_factory.HybridHeightFactory.time_return
           failed      16.2±0.05μs      n/a  coords.AncillaryVariable.time_create
           failed         82.8±5ns      n/a  coords.AncillaryVariable.time_return
           etc etc...
```

jamesp:
Should be resolved

.github/workflows/benchmark.yml
```diff
@@ -9,19 +9,17 @@ on:
...
     steps:
       # Checks-out your repository under $GITHUB_WORKSPACE, so your job can access it
       - uses: actions/checkout@v2
 
-      # fetch the target branch from origin too (just the head)
-      # and check it out into a local branch
-      - name: Checkout PR and target branch
-        run: |
-          git fetch --no-tags --depth=1 origin $BASE
+      - run: |
```
trexfeathers, line 17: How come this one doesn't have a `name`?

jamesp: They don't need names, but I've added one now anyway

trexfeathers:
Fancy adding this to the README?
[asv](http://your-url-here/)

trexfeathers (approved):

trexfeathers:
Thanks for your persistence @jamesp

PR #4262, opened 2021-07-29
Mesh Loading (AVD-1813)

trexfeathers:
### Description

Load `Mesh`es without associated `Cube`s.

TODO:
* [x] Integration tests
* ~~[ ] Performance tests~~
---

### DIFF
--- a/lib/iris/experimental/ugrid/__init__.py
+++ b/lib/iris/experimental/ugrid/__init__.py
@@ -3335,12 +3335,12 @@ def load_meshes(uris, var_name=None):
     uris : str or iterable of str
         One or more filenames/URI's. Any URI's must support OpenDAP.
     var_name : str, optional
-        Only return :class:`Mesh`es that have var_names matching this value.
+        Only return :class:`Mesh`'s that have var_names matching this value.
 
     Returns
     -------
     dict
-        A dictionary of file paths/URL's and lists of the :class:`Mesh`es
+        A dictionary of file paths/URL's and lists of the :class:`Mesh`'s
         returned from each.
 
     """
--- a/lib/iris/experimental/ugrid/__init__.py
+++ b/lib/iris/experimental/ugrid/__init__.py
@@ -205,9 +205,9 @@ class Connectivity(_DimensionalMetadata):
             for Fortran and legacy NetCDF files).
         * src_dim (int):
             Either ``0`` or ``1``. Default is ``0``. Denotes which dimension
-            of :attr:`indices` varies over the :attr:`src_location`'s (the
+            of :attr:`indices` varies over the :attr:`src_location`\\ s (the
             alternate dimension therefore varying within individual
-            :attr:`src_location`'s). (This parameter allows support for fastest varying index being
+            :attr:`src_location`\\ s). (This parameter allows support for fastest varying index being
             either first or last).
             E.g. for ``face_node_connectivity``, for 10 faces:
             ``indices.shape[src_dim] = 10``.
@@ -358,7 +358,7 @@ class Connectivity(_DimensionalMetadata):
     def src_dim(self):
         """
         The dimension of the connectivity's :attr:`indices` array that varies
-        over the connectivity's :attr:`src_location`'s. Either ``0`` or ``1``.
+        over the connectivity's :attr:`src_location`\\ s. Either ``0`` or ``1``.
         **Read-only** - validity of :attr:`indices` is dependent on
         :attr:`src_dim`. Use :meth:`transpose` to create a new, transposed
         :class:`Connectivity` if a different :attr:`src_dim` is needed.
@@ -372,7 +372,7 @@ class Connectivity(_DimensionalMetadata):
         Derived as the alternate value of :attr:`src_dim` - each must equal
         either ``0`` or ``1``.
         The dimension of the connectivity's :attr:`indices` array that varies
-        within the connectivity's individual :attr:`src_location`'s.
+        within the connectivity's individual :attr:`src_location`\\ s.
 
         """
         return self._tgt_dim
@@ -496,7 +496,7 @@ class Connectivity(_DimensionalMetadata):
         """
         Perform a thorough validity check of this connectivity's
         :attr:`indices`. Includes checking the sizes of individual
-        :attr:`src_location`'s (specified using masks on the
+        :attr:`src_location`\\ s (specified using masks on the
         :attr:`indices` array) against the :attr:`cf_role`.
 
         Raises a ``ValueError`` if any problems are encountered, otherwise
@@ -1933,7 +1933,7 @@ class Mesh(CFVariableMixin):
 
     def to_MeshCoords(self, location):
         """
-        Generate a tuple of :class:`MeshCoord`'s, each referencing the current
+        Generate a tuple of :class:`MeshCoord`\\ s, each referencing the current
         :class:`Mesh`, one for each :attr:`AXES` value, passing through the
         ``location`` argument.
 
@@ -1947,7 +1947,7 @@ class Mesh(CFVariableMixin):
             The ``location`` argument for :class:`MeshCoord` instantiation.
 
         Returns:
-            tuple of :class:`MeshCoord`'s referencing the current :class:`Mesh`.
+            tuple of :class:`MeshCoord`\\ s referencing the current :class:`Mesh`.
             One for each value in :attr:`AXES`, using the value for the
             ``axis`` argument.
 
@@ -3335,12 +3335,12 @@ def load_meshes(uris, var_name=None):
     uris : str or iterable of str
         One or more filenames/URI's. Any URI's must support OpenDAP.
     var_name : str, optional
-        Only return :class:`Mesh`'s that have var_names matching this value.
+        Only return :class:`Mesh`\\ es that have var_names matching this value.
 
     Returns
     -------
     dict
-        A dictionary of file paths/URL's and lists of the :class:`Mesh`'s
+        A dictionary of file paths/URL's and lists of the :class:`Mesh`\\ es
         returned from each.
 
     """
--- a/lib/iris/experimental/ugrid/__init__.py
+++ b/lib/iris/experimental/ugrid/__init__.py
@@ -3281,7 +3281,7 @@ class ParseUGridOnLoad(threading.local):
 PARSE_UGRID_ON_LOAD = ParseUGridOnLoad()
 
 
-def meshes_from_cf(cf_reader):
+def _meshes_from_cf(cf_reader):
     # TODO: docstring
 
     # Mesh instances are shared between file phenomena.
@@ -3363,7 +3363,7 @@ def load_meshes(uris, var_name=None):
 
     result = {}
     for source in valid_sources:
-        meshes_dict = meshes_from_cf(CFUGridReader(source))
+        meshes_dict = _meshes_from_cf(CFUGridReader(source))
         meshes = meshes_dict.values()
         if var_name is not None:
             meshes = filter(lambda m: m.var_name == var_name, meshes)
--- a/lib/iris/experimental/ugrid/__init__.py
+++ b/lib/iris/experimental/ugrid/__init__.py
@@ -3284,8 +3284,12 @@ PARSE_UGRID_ON_LOAD = ParseUGridOnLoad()
 
 
 def _meshes_from_cf(cf_reader):
-    # TODO: docstring
+    """
+    Common behaviour for extracting meshes from a CFReader.
 
+    Simple now, but expected to increase in complexity as Mesh sharing develops.
+
+    """
     # Mesh instances are shared between file phenomena.
     # TODO: more sophisticated Mesh sharing between files.
     # TODO: access external Mesh cache?
@@ -3298,7 +3302,23 @@ def _meshes_from_cf(cf_reader):
 
 
 def load_mesh(uris, var_name=None):
-    # TODO: docstring
+    """
+    Create a single :class:`Mesh` object from one or more NetCDF files.
+
+    Raises an error if more/less than one :class:`Mesh` is found.
+
+    Parameters
+    ----------
+    uris : str or iterable of str
+        One or more filenames/URI's. Any URI's must support OpenDAP.
+    var_name : str, optional
+        Only return a :class:`Mesh` if its var_name matches this value.
+
+    Returns
+    -------
+    :class:`Mesh`
+
+    """
     meshes_result = load_meshes(uris, var_name)
     result = [mesh for file in meshes_result.values() for mesh in file]
     mesh_count = len(result)
@@ -3311,7 +3331,23 @@ def load_mesh(uris, var_name=None):
 
 
 def load_meshes(uris, var_name=None):
-    # TODO: docstring
+    """
+    Create :class:`Mesh` objects from one or more NetCDF files.
+
+    Parameters
+    ----------
+    uris : str or iterable of str
+        One or more filenames/URI's. Any URI's must support OpenDAP.
+    var_name : str, optional
+        Only return a :class:`Mesh` if its var_name matches this value.
+
+    Returns
+    -------
+    dict
+        A dictionary of file paths/URL's and lists of the :class:`Mesh`es
+        returned from each.
+
+    """
     # No constraints or callbacks supported - these assume they are operating
     #  on a Cube.
 
--- a/lib/iris/experimental/ugrid/__init__.py
+++ b/lib/iris/experimental/ugrid/__init__.py
@@ -3348,13 +3348,12 @@ def load_meshes(uris, var_name=None):
         returned from each.
 
     """
+    # TODO: rationalise UGRID/mesh handling once experimental.ugrid is folded
+    #  into standard behaviour.
     # No constraints or callbacks supported - these assume they are operating
     #  on a Cube.
 
     from iris.fileformats import FORMAT_AGENT
-
-    # TODO: rationalise UGRID/mesh handling once experimental.ugrid is folded
-    #  into standard behaviour.
 
     if not PARSE_UGRID_ON_LOAD:
         # Explicit behaviour, consistent with netcdf.load_cubes(), rather than
--- a/lib/iris/experimental/ugrid/__init__.py
+++ b/lib/iris/experimental/ugrid/__init__.py
@@ -38,11 +38,7 @@ from ...common.metadata import (
 from ...common.mixin import CFVariableMixin
 from ...config import get_logger
 from ...coords import AuxCoord, _DimensionalMetadata
-from ...exceptions import (
-    ConnectivityNotFoundError,
-    ConstraintMismatchError,
-    CoordinateNotFoundError,
-)
+from ...exceptions import ConnectivityNotFoundError, CoordinateNotFoundError
 from ...fileformats import cf, netcdf
 from ...fileformats._nc_load_rules.helpers import get_attr_units, get_names
 from ...io import decode_uri, expand_filespecs
@@ -3326,7 +3322,7 @@ def load_mesh(uris, var_name=None):
         message = (
             f"Expecting 1 mesh, but input file(s) produced: {mesh_count} ."
         )
-        raise ConstraintMismatchError(message)
+        raise ValueError(message)
     return result[0]
 
 
--- a/lib/iris/experimental/ugrid/__init__.py
+++ b/lib/iris/experimental/ugrid/__init__.py
@@ -52,6 +52,8 @@ __all__ = [
     "CFUGridReader",
     "Connectivity",
     "ConnectivityMetadata",
+    "load_mesh",
+    "load_meshes",
     "Mesh",
     "Mesh1DConnectivities",
     "Mesh1DCoords",
@@ -3364,11 +3366,11 @@ def load_meshes(uris, var_name=None):
     result = {}
     for source in valid_sources:
         meshes_dict = _meshes_from_cf(CFUGridReader(source))
-        meshes = meshes_dict.values()
+        meshes = list(meshes_dict.values())
         if var_name is not None:
-            meshes = filter(lambda m: m.var_name == var_name, meshes)
+            meshes = list(filter(lambda m: m.var_name == var_name, meshes))
         if meshes:
-            result[source] = list(meshes)
+            result[source] = meshes
 
     return result
 
--- a/lib/iris/experimental/ugrid/__init__.py
+++ b/lib/iris/experimental/ugrid/__init__.py
@@ -3317,7 +3317,7 @@ def load_mesh(uris, var_name=None):
 
     """
     meshes_result = load_meshes(uris, var_name)
-    result = [mesh for file in meshes_result.values() for mesh in file]
+    result = set([mesh for file in meshes_result.values() for mesh in file])
     mesh_count = len(result)
     if mesh_count != 1:
         message = (
--- a/lib/iris/experimental/ugrid/__init__.py
+++ b/lib/iris/experimental/ugrid/__init__.py
@@ -3299,7 +3299,7 @@ def _meshes_from_cf(cf_reader):
 
 def load_mesh(uris, var_name=None):
     """
-    Create a single :class:`Mesh` object from one or more NetCDF files.
+    Load a single :class:`Mesh` object from one or more NetCDF files.
 
     Raises an error if more/less than one :class:`Mesh` is found.
 
@@ -3329,7 +3329,7 @@ def load_mesh(uris, var_name=None):
 
 def load_meshes(uris, var_name=None):
     """
-    Create :class:`Mesh` objects from one or more NetCDF files.
+    Load :class:`Mesh` objects from one or more NetCDF files.
 
     Parameters
     ----------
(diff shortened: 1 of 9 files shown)


lib/iris/experimental/ugrid/__init__.py
```diff
@@ -3469,7 +3606,17 @@ def identify(cls, variables, ignore=None, target=None, warn=True):
...
 
         # Identify all CF-UGRID mesh variables.
+        all_vars = target == variables
         for nc_var_name, nc_var in target.items():
+            if all_vars:
+                # SPECIAL BEHAVIOUR FOR MESH VARIABLES.
+                # We are looking for all mesh variables. Check if THIS variable
+                #  is a mesh using its own attributes.
+                if getattr(nc_var, "cf_role", "") == "mesh_topology":
+                    result[nc_var_name] = CFUGridMeshVariable(
```
pp-mo, line 3616: This is useful, but there are now 2 routes which lead to a CFUGridMeshVariable being constructed : Here, and below in line 3641.
Which is fine, given that..
 (a) id via a link-to is (now) no longer required to id a mesh in a given file (e.g.  for in a meshfile, or if it got missed), which we absolutely now want, _and_
 (b) we will also (now) tolerate a missing or miss-spelt 'cf_role' and still build a mesh
 
But.. I think this code will now, in the 'normal' case, be creating every mesh twice and setting `result[mesh-var-name]` twice.  
So it would be good to **check for a pre-existing entry in result[same-name]** and (a) check it matches (b) not re-create and over-write

**Also**, when creating "from a link", we should now be **checking `cf_role` and warning if it is not good**.
(i.e. around line 3641)

lib/iris/fileformats/netcdf.py
```diff
@@ -808,15 +808,7 @@ def load_cubes(filenames, callback=None):
...
-
-            # Mesh instances are shared between file phenomena.
-            # TODO: more sophisticated Mesh sharing between files.
-            # TODO: access external Mesh cache?
-            mesh_vars = cf.cf_group.meshes
-            meshes = {
-                name: _build_mesh(cf, var, filename)
-                for name, var in mesh_vars.items()
-            }
+            meshes = _meshes_from_cf(cf)
```
pp-mo, line 811: Nice !!

lib/iris/experimental/ugrid/__init__.py
```diff
@@ -3274,6 +3283,134 @@ def context(self):
...
+
+    """
+    meshes_result = load_meshes(uris, var_name)
+    result = [mesh for file in meshes_result.values() for mesh in file]
+    mesh_count = len(result)
+    if mesh_count != 1:
+        message = (
+            f"Expecting 1 mesh, but input file(s) produced: {mesh_count} ."
+        )
+        raise ConstraintMismatchError(message)
```
pp-mo, line 3329: I can see that this is what `iris.load_cube` uses, but I'm uncormfortable with the context.. there are no constraints involved here at all.
In fact, there isn't anything else that works in quite the same area.
So maybe just a simple ValueError would be better?

lib/iris/experimental/ugrid/__init__.py
```diff
@@ -3274,6 +3283,134 @@ def context(self):
...
+def load_meshes(uris, var_name=None):
+    """
+    Create :class:`Mesh` objects from one or more NetCDF files.
+
+    Parameters
+    ----------
+    uris : str or iterable of str
+        One or more filenames/URI's. Any URI's must support OpenDAP.
+    var_name : str, optional
+        Only return a :class:`Mesh` if its var_name matches this value.
```
pp-mo, line 3342:
```suggestion
        Only return :class:`Mesh`'s whose var_name matches this.
```
?maybe slightly clearer?

pp-mo (changes requested):
Some changes suggested, but no show-stoppers.  Feel free to mark things as 'resolved' in my absence.
I've only looked properly at the implementation code : I haven't time now to scan all the tests for correct coverage + someone else will have to take this on
But it's looking good to me !

lib/iris/experimental/ugrid/__init__.py
trexfeathers: 8a981a7997008e3ec815a1f04d8a5b69cd19b995

lib/iris/experimental/ugrid/__init__.py
trexfeathers: 5a29d44a8f40d3090fdf8b65a3e475efa1b14b81

lib/iris/experimental/ugrid/__init__.py
trexfeathers: #### Creating meshes twice

This is already the case for any variables that are referenced more than once e.g. files that have many data variables but which share coordinates/meshes/etc. If this hasn't been considered a problem so far in Iris' life, I'm reluctant to confuse the codebase by making a specific allowance here. Of course if it _has_ always been a problem we ought to fix it for every `identify()` method.

#### Bad `cf_role` warning/tolerance

Excellent point. I've mirrored what you did with `topology_dimension` in terms of warning and correcting the loaded value: 48c6d145b9ccfb144c9d67b1765c5e792e942fd6

trexfeathers:
This branch doesn't yet support performance testing, so that will be run locally for now.

lib/iris/tests/unit/experimental/ugrid/test_CFUGridMeshVariable.py
```diff
@@ -31,6 +31,22 @@ class TestIdentify(tests.IrisTest):
     def setUp(self):
         self.cf_identity = "mesh"
 
+    def test_cf_role(self):
+        match_name = "match"
```
pp-mo, line 35: Worth some comments, I think, explaining in one line what the different testcases (existing and new) are doing.
If I have it right, something like:
  * test_cf_identity : a variable is IDd as a mesh-variable when it appears in some `data-variable::mesh`, even if it does not have a `var::cf_role == 'mesh_topology'`
  * test_cf_role : a variable is IDd as a mesh-variable when it has `var::cf_role == 'mesh_topology'`, even if it does **_not_** appear in some `data-variable::mesh`
  * test_cf_role_and_identity : a variable is IDd as a mesh-variable when it has `var::cf_role == 'mesh_topology;`, **_and_** appears in some `data-variable::mesh`

lib/iris/tests/unit/experimental/ugrid/test_load_mesh.py
```diff
@@ -0,0 +1,54 @@
...
+
+from iris.experimental.ugrid import PARSE_UGRID_ON_LOAD, load_mesh
+
+
+class Tests(tests.IrisTest):
+    # All 'real' tests have been done for load_meshes(). Here we just check
+    #  that load_mesh() works with load_meshes() correctly, using mocking.
+    def setUp(self):
+        patcher = mock.patch("iris.experimental.ugrid.load_meshes")
+        self.addCleanup(patcher.stop)
```
pp-mo, line 25: `patcher = mock.patch ... ; self.addCleanup(patcher.stop) ... ; patcher.start() `
All this can be replaced with use of `self.patch()`, which does all those things  (provided in `IrisTest`).

lib/iris/tests/unit/experimental/ugrid/test_load_meshes.py
```diff
@@ -0,0 +1,218 @@
...
+                    int levels(levels) ;
+                    float node_data(levels, node) ;
+                        node_data:coordinates = "node_x node_y" ;
+                        node_data:location = "node" ;
+                        node_data:mesh = "mesh" ;
+                    float face_data(levels, face) ;
+                        face_data:coordinates = "face_x face_y" ;
+                        face_data:location = "face" ;
+                        face_data:mesh = "mesh" ;
+                data:
```
pp-mo, line 83: You probably don't need to define any actual variable data values to make this work?
Unless possibly some of the load chain checks for no-missing-points ?  : I haven't actually checked that

lib/iris/tests/unit/experimental/ugrid/test_load_meshes.py
```diff
@@ -0,0 +1,218 @@
...
+                        mesh:node_coordinates = "node_x node_y" ;
+                        mesh:face_coordinates = "face_x face_y" ;
+                        mesh:face_node_connectivity = "face_nodes" ;
+                    float node_x(node) ;
+                        node_x:standard_name = "longitude" ;
+                    float node_y(node) ;
+                        node_y:standard_name = "latitude" ;
+                    float face_x(face) ;
+                        face_x:standard_name = "longitude" ;
+                    float face_y(face) ;
```
pp-mo, line 69: I suspect that face coordinates are redundant for what you are testing here.
If so, I would omit them for minimalism + brevity

lib/iris/tests/unit/experimental/ugrid/test_load_meshes.py
```diff
@@ -0,0 +1,218 @@
...
+                )
+        self.assertDictEqual({}, meshes)
+
+
+class TestsHttp(tests.IrisTest):
+    # Tests of HTTP (OpenDAP) loading need mocking since we can't have tests
+    #  that rely on 3rd party servers.
+    def setUp(self):
+        patcher = mock.patch("iris.fileformats.FORMAT_AGENT.get_spec")
+        self.addCleanup(patcher.stop)
```
pp-mo, line 197: See previous note on using `self.patch` for this.

pp-mo (commented):
Basically all quite nice, but I raised a few questions.
So here's some things to think about, anyway.
Unfortunately I'm still not really quite done -- I need to think about the coverage of the main new tests a bit more.

lib/iris/tests/integration/experimental/test_ugrid_load.py
```diff
@@ -178,6 +182,47 @@ def test_mesh_no_topology_dimension(self):
...
+        log_regex = r"no cf_role attribute"
+        with self.assertLogs(logger, level="WARNING", msg_regex=log_regex):
+            template = "minimal_bad_mesh_cf_role"
+            dim_line = ""
+            _ = self.create_synthetic_test_cube(
+                template=template, subs=dict(CF_ROLE_DEFINITION=dim_line)
+            )
+
+
+@tests.skip_data
```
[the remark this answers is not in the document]
pp-mo: (Afterthought : arguably though, the no-mesh cases are covered OK by the unit tests?)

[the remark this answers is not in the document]
trexfeathers: I was just looking at this - your uncertainty makes me wonder if I've got the wrong idea about integration tests and coverage. I've always _thought_ that coverage is delivered by unit tests, while integration tests' role is to simulate a real use case - to make sure that nothing bad happens when everything is 'strung together'.

`load_meshes()` does barely anything more than `load_mesh()`. What value is there in having an integration test specifically for this, other than to increase the test run time?

Also note that we would need to deviate from the other integration tests here and use a synthetic UGRID file - we have so far been unable to source a UGRID file with multiple meshes that isn't also too big for `iris-test-data`.

lib/iris/experimental/ugrid/__init__.py
```diff
@@ -3274,6 +3279,133 @@ def context(self):
...
+
+
+def load_meshes(uris, var_name=None):
+    """
+    Create :class:`Mesh` objects from one or more NetCDF files.
+
+    Parameters
+    ----------
+    uris : str or iterable of str
+        One or more filenames/URI's. Any URI's must support OpenDAP.
```
pp-mo, line 3336: Could also mention the possibility of file wildcards here.
Though, for the main load/load_cube/load_cubes routines, this in not in their docstrings but is covered [in the iris module docstring](https://github.com/SciTools/iris/blob/main/lib/iris/__init__.py#L36-L45).

lib/iris/tests/unit/experimental/ugrid/test_load_meshes.py
trexfeathers: 436f5f737bc1c1b8fc08631717bcf64fab5f054e

lib/iris/tests/unit/experimental/ugrid/test_load_mesh.py
trexfeathers: 436f5f737bc1c1b8fc08631717bcf64fab5f054e

lib/iris/tests/unit/experimental/ugrid/test_CFUGridMeshVariable.py
trexfeathers: ff73cb0308cba274c1ab50af990cf2dd835c4833

lib/iris/tests/integration/experimental/test_ugrid_load.py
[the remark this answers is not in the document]
trexfeathers: In the vein: I **do** see the value in checking that a full process flow doesn't fall over when loading a file without meshes: a14317b686e6ef32629006c2c9aa6f25309c485a.

lib/iris/experimental/ugrid/__init__.py
```diff
@@ -3274,6 +3279,133 @@ def context(self):
...
+    var_name : str, optional
+        Only return :class:`Mesh`'s that have var_names matching this value.
```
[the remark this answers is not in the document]
trexfeathers: I agree it's not great. Thankfully I now know how to 'officially' get Sphinx to play nicely with pluralisation:
7316ec62fbf40b07006814c0f1c99f39ef67ad59

lib/iris/experimental/ugrid/__init__.py
```diff
@@ -3274,6 +3279,133 @@ def context(self):
...
+
+    Returns
+    -------
+    dict
+        A dictionary of file paths/URL's and lists of the :class:`Mesh`'s
```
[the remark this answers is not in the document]
trexfeathers: Thanks, that's much nicer.
5ffc3d6606f71c0a1efde3d28fa36d25ad6bb4d9
7316ec62fbf40b07006814c0f1c99f39ef67ad59

pp-mo (commented):
Ok, that's my part#2
and I think that's now all 
**_except_** that I'm still thinking about the duplication of the CFUGridMeshVariable creation I mentioned before : I'm still thinking about whether I think that is something different from the existing cases, or not.  But that shouldn't take me much longer, and I'll let you know shortly..

lib/iris/tests/unit/experimental/ugrid/test_load_meshes.py
trexfeathers: I'd like to keep this. Tests are intimidating when revisiting, so it's important to me to minimise confusion wherever possible, and a good example is having valid CDL that one would actually expect to see.

Brevity isn't always good for clarity, even if it is in many cases!

lib/iris/tests/unit/experimental/ugrid/test_load_meshes.py
trexfeathers: https://github.com/SciTools/iris/pull/4262#discussion_r692873231

lib/iris/experimental/ugrid/__init__.py
```diff
@@ -3274,6 +3279,133 @@ def context(self):
...
+    # TODO: access external Mesh cache?
+    mesh_vars = cf_reader.cf_group.meshes
+    meshes = {
+        name: _build_mesh(cf_reader, var, cf_reader.filename)
+        for name, var in mesh_vars.items()
+    }
+    return meshes
+
+
+def load_mesh(uris, var_name=None):
```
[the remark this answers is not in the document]
trexfeathers: https://github.com/SciTools/iris/pull/4262#discussion_r692854595 nullifies this concern, yes?

lib/iris/experimental/ugrid/__init__.py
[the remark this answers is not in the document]
pp-mo: I would still say this ought to say "mapping each file path/URL **which contains :class:`Mesh`es** to a list ..."
Because otherwise it still sound like it should be 1-2-1 with the input  _which_ `_it_` is not.

lib/iris/tests/unit/experimental/ugrid/test_load_meshes.py
pp-mo: But it **_is_** valid CDL.  It just makes a file with all-missing variable content.

lib/iris/experimental/ugrid/__init__.py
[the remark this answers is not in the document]
pp-mo: Sorry, just realised I already noted this in my #2 pass..

lib/iris/tests/unit/experimental/ugrid/test_load_meshes.py
trexfeathers: OK "valid" is a loaded word. Would you ever expect to encounter such a file?

lib/iris/experimental/ugrid/__init__.py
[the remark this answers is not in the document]
pp-mo: Sorry, not quite correct....
> if they all contain the same (shared) Mesh, load_meshes would return a dict with a single key matching the arg, but the value will be a list of duplicates of the Mesh (assuming they all had the same)

Should actually be 
"if they all contain the same (shared) Mesh, then load_meshes would return a key per file, and each value will be a duplicate of the single Mesh (assuming they all had the same)"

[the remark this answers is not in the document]
trexfeathers: > unexpected

Actually deliberate(!).

I wrote it this way specifically because we haven't [yet] implemented `Mesh` sharing. Part of the acceptance criteria was to return a list of `Mesh`es from a collection of files. But without `Mesh` sharing, we know that plenty of use cases would return a list of duplicate `Mesh`es, which would be confusing and might appear erroneous. Instead, I wanted to provide better context by mapping each `Mesh` to its source file.

I'm struggling with design philosophy here. One should write code for the world as it currently is, which suggests I did the write thing to help clarify outputs in the absence of `Mesh` sharing. But if we anticipate `Mesh` sharing in future, the correct API then will be different - probably returning a `list` instead of a `dict`, but that basically means I am consciously forcing a future breaking change.

---

`load_mesh()` is as close as reasonably possible to `load_cube()`, and despite the different cases it still has some of `load_cube()`'s utility, specifically the ability to extract a single named `Mesh` from a collection of files by using the `var_name` parameter. But perhaps for every bit of utility there is a greater bit of potential confusion?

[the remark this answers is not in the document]
trexfeathers: TL;DR I'm on the fence!

lib/iris/experimental/ugrid/__init__.py
trexfeathers: b28fbd80bdb8007dc0c71831cd62e85756c5fc4f

lib/iris/experimental/ugrid/__init__.py
```diff
@@ -3274,6 +3279,133 @@ def context(self):
...
+    """
+    Create a single :class:`Mesh` object from one or more NetCDF files.
+
+    Raises an error if more/less than one :class:`Mesh` is found.
+
+    Parameters
+    ----------
+    uris : str or iterable of str
+        One or more filenames/URI's. Any URI's must support OpenDAP.
```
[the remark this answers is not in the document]
trexfeathers: b28fbd80bdb8007dc0c71831cd62e85756c5fc4f

lib/iris/experimental/ugrid/__init__.py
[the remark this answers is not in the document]
trexfeathers: b28fbd80bdb8007dc0c71831cd62e85756c5fc4f

lib/iris/tests/unit/experimental/ugrid/test_load_meshes.py
```diff
@@ -0,0 +1,218 @@
...
+            meshes = load_meshes(nc_path)
+
+        files = list(meshes.keys())
+        self.assertEqual(1, len(files))
+        file_meshes = meshes[files[0]]
+        self.assertEqual(1, len(file_meshes))
+        mesh = file_meshes[0]
+        self.assertEqual("mesh", mesh.var_name)
+
+    def test_var_name(self):
```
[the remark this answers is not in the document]
trexfeathers: bb30d2ec40b7b0912b6b1312bb9491c50923a3d4

lib/iris/tests/unit/experimental/ugrid/test_load_meshes.py
```diff
@@ -0,0 +1,218 @@
...
+
+        files = list(meshes.keys())
+        self.assertEqual(1, len(files))
+        file_meshes = meshes[files[0]]
+        self.assertEqual(2, len(file_meshes))
+        mesh_names = [mesh.var_name for mesh in file_meshes]
+        self.assertIn("mesh", mesh_names)
+        self.assertIn("mesh2", mesh_names)
+
+    def test_no_parsing(self):
```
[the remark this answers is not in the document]
trexfeathers: 2a4237f04cf52472deb976e3e6d4c0aed93362cb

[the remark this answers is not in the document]
pp-mo: :+1:

lib/iris/experimental/ugrid/__init__.py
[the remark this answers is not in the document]
pp-mo: > I wanted to provide better context by mapping each Mesh to its source file.

Absolutely no disagreement with that.  I think `load_meshes` is fine just as it is.
I just thought that `load_mesh` might require a bit of a re-think, as it isn't quite what I expected.

But actually I can see the aim, of checking that there is only one Mesh in multiple files, even though we can't actually do that yet.
As it already says in the docstring, "Create a single :class:`Mesh` object from one or more NetCDF files."
In that case, rather than restricting to a single input, we can leave the args as-is and document the intent.

So, thinking a bit more, I think we can generate a more "future-proof" way of actually doing this by just collecting all the meshes from the lists-of-meshes in dict values, and using a set to remove duplicates.
I.E. just change "result = [mesh for file in meshes_result.values() for mesh in file]",
to "result = **set(**[mesh for file in meshes_result.values() for mesh in file]**)**"

That uses on Mesh.\_\_eq__, so it will work with mesh-sharing as at present, but also across files once we fix mesh comparison.

lib/iris/experimental/ugrid/__init__.py
```diff
@@ -3274,6 +3279,135 @@ def context(self):
...
+    meshes = {
+        name: _build_mesh(cf_reader, var, cf_reader.filename)
+        for name, var in mesh_vars.items()
+    }
+    return meshes
+
+
+def load_mesh(uris, var_name=None):
+    """
+    Create a single :class:`Mesh` object from one or more NetCDF files.
```
pp-mo, line 3302: Small point, but I think 
```suggestion
    Load a single :class:`Mesh` object from one or more NetCDF files.
```
is a more suggestive of what is does with the input source(s)

lib/iris/experimental/ugrid/__init__.py
```diff
@@ -3274,6 +3279,135 @@ def context(self):
...
+        message = (
+            f"Expecting 1 mesh, but input file(s) produced: {mesh_count} ."
+        )
+        raise ValueError(message)
+    return result[0]
+
+
+def load_meshes(uris, var_name=None):
+    """
+    Create :class:`Mesh` objects from one or more NetCDF files.
```
pp-mo, line 3332: Small point, as above
```suggestion
    Load :class:`Mesh` objects from one or more NetCDF files.
```

lib/iris/tests/unit/experimental/ugrid/test_load_meshes.py
trexfeathers: d8809da8baec16684f98f281574ace3b8bc7438d

lib/iris/tests/unit/experimental/ugrid/test_load_meshes.py
trexfeathers: Have discussed offline, leaving a minimal dataset in.

lib/iris/experimental/ugrid/__init__.py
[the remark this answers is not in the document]
trexfeathers: bb30d2ec40b7b0912b6b1312bb9491c50923a3d4

lib/iris/experimental/ugrid/__init__.py
trexfeathers: ec55e07404debf2623283b28b6591c5bea4ae39c

lib/iris/experimental/ugrid/__init__.py
trexfeathers: ec55e07404debf2623283b28b6591c5bea4ae39c

pp-mo:
It looks like we need to make Meshes hashable.
See : https://github.com/trexfeathers/iris/pull/5

Logical arguments against, exactly as for Cube, but I think it's forgiveable.

pp-mo (approved):

trexfeathers:
Thanks @pp-mo!

PR #4150, opened 2021-05-23
Contourf Antialias Fix

rcomer:
### Description

Fixes #4086 

As noted at #4086, our antialiasing handling exposed a matplotlib bug for a pretty specific edge case.  Regardless of this bug, it's a bit redundant to call `pyplot.contour` when the data do not span the requested levels, so we should plausibly get a tiny efficiency gain by skipping it when not needed.

---
[Consult Iris pull request check list]( https://scitools-iris.readthedocs.io/en/latest/developers_guide/contributing_pull_request_checklist.html)

Issue #4086, opened 2021-04-08
Antialiased contourf plot shows unexpected artifacts


jamesp:
## 🐛 Bug Report

When plotting an iris cube with `contourf` in some circumstances an additional contour is shown.  This appears to be related to the logic included by the `antialiased` option in `iris.plot.contouf`.

## How To Reproduce
Steps to reproduce the behaviour:

1. Contact @jamesp for data and reproduction script 

## Expected behaviour
Antialiased data should not show additional artifacts when plotting.

## Screenshots
![image](https://user-images.githubusercontent.com/22805/114049558-7110db80-9883-11eb-9d5c-d45bfe866c98.png)

## Environment 
 - OS & Version: Linux
 - Iris Version: 3.0.1.  Affects Iris >= 2.3

bjlittle:
@jamesp What version of `mpl` are you using?

jamesp:
The antialiased artifact (additional contour) is present on all stack versions since Iris 1.13. 

I've tested on matplotlib 2.2.3 and 3.3.4, both show the same additional contour.  Setting `antialiased=False` restores the same behaviour as 1.13, which is what the user was running on previously.

I've tried to create a standardised test that reproduces the issue, but haven't achieved it yet, there does appear to be something pathological about the input data

rcomer:
Here is a minimal example:

```python
import iris
import iris.quickplot as qplt
import numpy as np
import matplotlib.pyplot as plt


lats = np.linspace(-10, 10, 20)
lons_2d, lats_2d = np.meshgrid(lats, lats)

data = np.cos(lons_2d * np.pi/10.) + np.cos(lats_2d * np.pi/10.)
np.place(data, data < 0., 0.)

cube = iris.cube.Cube(data)
for dim, name in enumerate(['longitude', 'latitude']):
    cube.add_dim_coord(iris.coords.DimCoord(lats, standard_name=name, units='degrees'), dim)

levels = [0.2, 2, 20, 200]
colors = ['b', 'r', 'y']

qplt.contourf(cube, levels=levels, colors=colors)
```
![contourf_antialias1](https://user-images.githubusercontent.com/10599679/117773219-f5160480-b22f-11eb-8458-97cbd4dbe649.png)
The problem seems to be triggered by a combination of
* having zeros everywhere other than our desired feature
* data values all being below the second level

If we pick level so that all the colours are used, all seems well

```python
levels = [0.2, 0.5, 1, 2]
[...]

PR #4263, opened 2021-07-29
Contourf bugfix in the presence of nans

### LINKED ISSUE #4261: Contourf plots look faulty if array contains nans
This plot was created with `contourf`. It does not look right in regions with gradients. I can put together an MWE if this is necessary, but I just wanted to ask if you are aware of this problem and if it will be fixed.

OS: Ubuntu 20.04
iris: 3.1.dev0
matpltolib: 3.4.2
cartopy: 0.19.0.post2
20210729_6_a m _Test_surface_current_3

MHBalsmeier:
### Description

Fixes #4261 by using `np.nanmax(cube.data)` instead of `cube.data.max()`.


MHBalsmeier:
I have checked, it does not break the problem mentioned in #4086.

MHBalsmeier:
I don't think I can get the performance test to pass.

jamesp:
Thanks for the PR fix @MHBalsmeier.  Don't worry about the performance regression, that CI test is still work in progress, no reason this change should affect the benchmark that triggered.  I've re-run the check.

@rcomer, do you want to just confirm you're happy with this fix?  Seems like a reasonable change to me :+1:

rcomer:
Yes thanks @MHBalsmeier for this.  Nice detective work!  I have to admit that `nan`s had not even crossed my mind when I wrote the original fix.  The proposed change looks good to me 👍

MHBalsmeier:
Okay I think this is ready for the final review.

rcomer:
I’ve added this to the v3.1 project, as we clearly don’t want to introduce the bug in the next release!

My only question is whether this should have a new test, or whether that would be overkill for such a minor change.  If we did want a test, I think it could follow the same pattern as the one I added at #4150.  Just need to insert a `nan` somewhere in `cube.data` and choose the levels so that the loop is triggered.  Then `ax.collections` should contain `LineCollection` instances, as they are added by `contour`.  It might be enough to just check that the last element of `ax.collections` is a `LineCollection`.

What do you think @MHBalsmeier, @jamesp?

MHBalsmeier:
I am not sure if I get your point correctly because why would we introduce a bug with this ... I thought this solves #4086 as well as #4261.

As for the test, why not, but I never ran these tests and you seem to have the better idea of how to test this ...
I personally do not think this is really necessary.

rcomer:
@MHBalsmeier sorry I meant the bug that I introduced at #4150, that you are now fixing.  We need to include your fix for the v3.1 release.

MHBalsmeier:
@rcomer okay I see

rcomer:
We have now added a test.  I verified that the new test fails against `main`.  The new `setUp` and `tearDown` ensure that the two tests on the class don't plot on the same figure/axes and interfere with each other's results.

The new failures appear to be because an external file has moved.  I have opened #4266 to address that.

jamesp:
Thanks for fixing the doc failures @rcomer.  @MHBalsmeier this just needs a rebase and should be good to go 👍

MHBalsmeier:
Okay, lets see if I did that correctly.

jamesp (approved):
Thanks so much @MHBalsmeier for sticking with it and getting this fix in, much appreciated!

PR #4266, opened 2021-08-02
Fix scipy intersphinx mapping

rcomer:
### Description

It looks like this mapping object has moved.  See #4263 where it broke the linkcheck and doctests.

---


jamesp (approved):
Looks good, thanks for tracking this down @rcomer

PR #4268, opened 2021-08-03
add stale gha

bjlittle:
### Description
This PR adds the `stale` GHA in order to help maintain a healthy issues and prs backlog.

---

### DIFF
--- a/.github/workflows/stale.yml
+++ b/.github/workflows/stale.yml
@@ -1,0 +1,70 @@
+name: Stale issues and pull-requests
+on:
+  schedule:
+    - cron: 0 * * * *
+
+jobs:
+  stale:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/stale@v4.0.0
+        with:
+          repo-token: ${{ secrets.GITHUB_TOKEN }}
+
+          # Idle number of days before marking issues/prs stale.
+          days-before-stale: 1000
+
+          # Idle number of days before closing stale issues/prs.
+          days-before-close: 28
+
+          # Comment on the staled issues.
+          stale-issue-message: |
+            In order to maintain a backlog of relevant issues, we automatically label them as stale after 1,000 days of inactivity.
+
+            If this issue is still important to you, then please comment on this issue or remove the stale label.
+
+            Otherwise this issue will be automatically closed in 28 days time.
+
+          # Comment on the staled prs.
+          stale-pr-message: | 
+            In order to maintain a backlog of relevant PRs, we automatically label them as stale after 1,000 days of inactivity.
+
+            If this PR is still important to you, then please comment on this PR or remove the stale label.
+
+            Otherwise this PR will be automatically closed in 28 days time.
+
+          # Comment on the staled issues while closed.
+          close-issue-message: |
+            This stale issue has been automatically closed due to a lack of community activity.
+            
+            If you still care about this issue, then please either:
+              * Re-open this issue, if you have sufficient permissions, or
+              * Add a comment pinging `@SciTools/iris-devs` who will re-open on your behalf.
+
+          # Comment on the staled prs while closed.
+          close-pr-message: |
+            This stale PR has been automatically closed due to a lack of community activity.
+            
+            If you still care about this PR, then please either:
+              * Re-open this PR, if you have sufficient permissions, or
+              * Add a comment pinging `@SciTools/iris-devs` who will re-open on your behalf.
+
+          # Label to apply on staled issues. 
+          stale-issue-label: Stale
+
+          # Label to apply on staled prs.
+          stale-pr-label: Stale
+
+          # Labels on issues exempted from stale.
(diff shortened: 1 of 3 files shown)


bjlittle:
I guess the `days-before-stale` and `days-before-close` are up for debate.

If it helps, here is a plot of number of issues that would be not be considered state for the given dates or days since now `3 Aug 2021` e.g., for the proposed `days-before-stale` of `1000` which is `7 Nov 2018`, approximately `159` issues would remain **not** marked as `stale` out of `295`.

stale_issues

.github/workflows/stale.yml
```diff
@@ -0,0 +1,70 @@
...
+
+            Otherwise this PR will be automatically closed in 28 days time.
+
+          # Comment on the staled issues while closed.
+          close-issue-message: |
+            This stale issue has been automatically closed due to a lack of community activity.
+            
+            If you still care about this issue, then please either:
+              * Re-open this issue, if you have sufficient permissions, or
+              * Add a comment pinging `@SciTools/iris-devs` who will re-open on your behalf.
```
bjlittle, line 42: To be honest, we could have a GHA `bot` that auto-opens closed issues/prs whenever the user adds a comment like `@SciTools/iris-devs, please re-open` akin to `@conda-forge-admin, please rerender` on a `conda-forge` feedstock.

jamesp:
I think this is a great idea.  I'd be all in favour of a shorter `days-before-stale`. Realistically, if we haven't addressed it in 365 days it's probably never going to get addressed, and the project will have moved on significantly from this point of raising.

.github/workflows/stale.yml
```diff
@@ -0,0 +1,70 @@
...
+
+          # Label to apply on staled issues. 
+          stale-issue-label: Stale
+
+          # Label to apply on staled prs.
+          stale-pr-label: Stale
+
+          # Labels on issues exempted from stale.
+          exempt-issue-labels: |
+            "Status: Blocked,Status: Decision Required,Status: Needs Info,Status: Stalled,Peloton 🚴‍️"
```
jamesp, line 60: I would argue that "Needs Info" and "Stalled" should not be in this list.  They should be staled out if they go 1000 days without info or unstalling

bjlittle: Cool, I'd go with that... did I miss out any labels that you'd consider should be in the list?

bjlittle: Done :+1:

jamesp (approved):

.github/workflows/stale.yml
```diff
@@ -0,0 +1,70 @@
+name: Stale issues and pull-requests
+on:
+  schedule:
+    - cron: 0 * * * *
```
jamesp, line 4: Does this need to run every hour?  Could we run every day?

.github/workflows/stale.yml
```diff
@@ -0,0 +1,70 @@
...
+          days-before-stale: 500
+
+          # Idle number of days before closing stale issues/prs.
+          days-before-close: 28
+
+          # Comment on the staled issues.
+          stale-issue-message: |
+            In order to maintain a backlog of relevant issues, we automatically label them as stale after 500 days of inactivity.
+
+            If this issue is still important to you, then please comment on this issue or remove the stale label.
```
rcomer, line 24: I think only core devs can remove the label.  How about "If this issue is still important to you, then please comment on this issue and the stale label will be removed." ?

.github/workflows/stale.yml
```diff
@@ -0,0 +1,70 @@
...
+
+            Otherwise this issue will be automatically closed in 28 days time.
+
+          # Comment on the staled prs.
+          stale-pr-message: | 
+            In order to maintain a backlog of relevant PRs, we automatically label them as stale after 500 days of inactivity.
+
+            If this PR is still important to you, then please comment on this PR or remove the stale label.
```
rcomer, line 32: As above.

.github/workflows/stale.yml
```diff
@@ -0,0 +1,70 @@
+name: Stale issues and pull-requests
+on:
+  schedule:
+    - cron: 0 0 * * *
```
bjlittle, line 4: Run GHA daily, see https://crontab.guru/every-day

rcomer (commented):
Over on the discussion, there was talk of starting with a 3 year lead, and then gradually reducing.  I think starting with a smaller number of stale issues makes sense, because we can see what kind of activity (if any) is generated by people re-engaging.  It would be a shame if someone did re-engage but we didn't have time to address their comments.

On the gap between stale and closed, I would have said at least a month, as it gives time for people who are on holiday or otherwise away.  So 28 days seems OK to me, but I wouldn't want to go any lower.

bjlittle:
@jamesp Did you want a `whatsnew` entry? I'm really not fussed...

jamesp:
No I don't think it's necessary, thanks @bjlittle!

PR #4269, opened 2021-08-03
CI Benchmarks are too sensitive

jamesp:
We're getting too many false positives from CI benchmarks.  This increases the tolerance to a 50% slowdown.

### Description

---

### DIFF
--- a/.github/workflows/benchmark.yml
+++ b/.github/workflows/benchmark.yml
@@ -30,7 +30,7 @@ jobs:
       - name: Run benchmarks on source and target
         run: |
           cd benchmarks
-          asv continuous --factor 1.2 _base _head
+          asv continuous --factor 1.5 _base _head
 
       - name: Write a compare file to the output folder
         if: ${{ always() }}


trexfeathers:
I'm a little nervous about this. I have in the past ignored factor ~2 slowdowns, only to find they **were** problematic for users since they were using the same operation many times - multiplying the 2 many times.

I wonder if we could look for an alternative quick-ish fix. Is it the same benchmark(s) that are failing each time?

jamesp:
Yep fair point, it may be better to have a smaller subset of benchmarks that we know to be more appropriate for CI testing, and then leave the more sensitive to the main asv profiling routine

trexfeathers:
> leave the more sensitive to the main asv profiling routine

Certainly an option. I was actually envisaging applying the wisdom from ASV's [Tuning timing measurements](https://asv.readthedocs.io/en/stable/tuning.html) to problem benchmarks. Lots of options 🙂

trexfeathers:
Closing so we can come up with a better plan.

PR #4271, opened 2021-08-04
ascending issues/prs order for gha stale

bjlittle:
### Description
This PR updates the `stale` GHA to enforce the processing of `ascending` issues/PRs i.e., oldest first.

---

### DIFF
--- a/.github/workflows/stale.yml
+++ b/.github/workflows/stale.yml
@@ -1,3 +1,5 @@
+# See https://github.com/actions/stale
+
 name: Stale issues and pull-requests
 on:
   schedule:
@@ -57,14 +59,20 @@ jobs:
 
           # Labels on issues exempted from stale.
           exempt-issue-labels: |
-            "Status: Blocked,Status: Decision Required,Peloton 🚴‍️"
+            "Status: Blocked,Status: Decision Required,Peloton 🚴‍♂️"
 
           # Labels on prs exempted from stale.
           exempt-pr-labels: |
-            "Status: Blocked,Status: Decision Required,Peloton 🚴‍"
+            "Status: Blocked,Status: Decision Required,Peloton 🚴‍♂️"
+
+          # Max number of operations per run.
+          operations-per-run: 30
 
           # Remove stale label from issues/prs on updates/comments.
           remove-stale-when-updated: true
+
+          # Order to get issues/PRs.
+          ascending: true
 
           # Exempt all issues/prs with milestones from stale.
           exempt-all-milestones: true


.github/workflows/stale.yml
```diff
@@ -57,14 +59,20 @@ jobs:
 
           # Labels on issues exempted from stale.
           exempt-issue-labels: |
-            "Status: Blocked,Status: Decision Required,Peloton 🚴‍️"
+            "Status: Blocked,Status: Decision Required,Peloton 🚴‍♂️"
```
bjlittle, line 62: I appear to have not originally copied the peloton label correctly :blush:

.github/workflows/stale.yml
```diff
@@ -57,14 +59,20 @@ jobs:
...
 
           # Labels on prs exempted from stale.
           exempt-pr-labels: |
-            "Status: Blocked,Status: Decision Required,Peloton 🚴‍"
+            "Status: Blocked,Status: Decision Required,Peloton 🚴‍♂️"
+
+          # Max number of operations per run.
+          operations-per-run: 30
```
bjlittle, line 69: Let's make this explicit, even though it's the default value, as it may be something that we want to change given the observed behaviour of the GHA

.github/workflows/stale.yml
```diff
@@ -57,14 +59,20 @@ jobs:
...
 
           # Remove stale label from issues/prs on updates/comments.
           remove-stale-when-updated: true
 
+          # Order to get issues/PRs.
+          ascending: true
```
bjlittle, line 75: Let's go oldest first... seems like the best behaviour to adopt for a higher stale hit ratio.

.github/workflows/stale.yml
rcomer: I genuinely can't see what the difference is here 😕

trexfeathers: Looks like the emoji difference visible on Linux but not Windows. This isn't the first time that confusion has come up - worth keeping in mind!

trexfeathers (approved):
LGTM!

.github/workflows/stale.yml
bjlittle: It's the *gender* of the cyclist, which can be male or female... apparently it matters :neutral_face: 

It appears to work by appending the gender unicode character after the cyclist character i.e., 🚴 + ♂️ = 🚴‍♂️ (person-biking + male-sign = man-biking) but whether you see the difference or not depends on the platform

rcomer: 🤔🧐

PR #4273, opened 2021-08-05
Preserve position when replacing Axes with GeoAxes

### LINKED ISSUE #1157: Plotting cube data in inset axes
There is a really niche bug in Iris that causes plotting to fail with an `AttributeError` in the very specific case of plotting mappable\* data into an inset axes*\* that was not originally defined as a Cartopy `GeoAxes` instance with use of the `projection` kwarg.

\* By mappable data I mean any data that will be drawn to a Cartopy `GeoAxes` instance.
** By an inset axes I mean a new axes instance with a positional `rect` defined.

The following code demonstrates the problem:

``` python
import iris
import iris.plot as iplt

fname = iris.sample_data_path("air_temp.pp")
cube = iris.load_cube(fname)

fig = plt.figure()
ax1 = plt.axes()
ax2 = plt.axes([.17, .65, .2, .2])

plt.sca(ax1)
iplt.contourf(cube)

plt.sca(ax2)
iplt.pcolormesh(cube)

plt.show()
```

which produces the following traceback:

```
  File "test_code.py", line 92, in <module>
    main()
  File "test_code.py", line 87, in main
    a_different_approach(cube, x, y, c)
  File "test_code.py", line 53, in a_different_approach
    iplt.pcolormesh(cube)
  File "/opt/ukmo/iris/default/linux64/site-packages/iris/plot.py", line 783, in pcolormesh
    result = _draw_2d_from_bounds('pcolormesh', cube, *args, **kwargs)
  File "/opt/ukmo/iris/default/linux64/site-packages/iris/plot.py", line 235, in _draw_2d_from_bounds
    cube, plot_defn, *args, **kwargs)
  File "/opt/ukmo/iris/default/linux64/site-packages/iris/plot.py", line 528, in _map_common
    kwargs)
  File "/opt/ukmo/iris/default/linux64/site-packages/iris/plot.py", line 467, in _geoaxes_draw_method_and_kwargs
    draw_method = getattr(ax, draw_method_name)
AttributeError: 'NoneType' object has no attribute 'pcolormesh'
```

Adding a `projection` kwarg to the definition of `ax2` works around the problem because the problem code in `iris.plot._replace_axes_with_cartopy_axes` is not run to return a `NoneType` that causes the AttributeError.

rcomer:
### Description

Fixes #1157.  The example from that issue now produces:

inset_example

This fix seems almost too easy, so I am somewhat worried I'm missing something 👀 

I couldn't find any existing tests for this helper function (I guess it is well tested indirectly though), so added a new test module.

---

### DIFF
--- a/lib/iris/plot.py
+++ b/lib/iris/plot.py
@@ -843,7 +843,7 @@ def _replace_axes_with_cartopy_axes(cartopy_proj):
                 ylabel=ax.get_ylabel(),
             )
         else:
-            position = ax.get_position()
+            position = ax.get_position(original=True)
             _ = fig.add_axes(
                 position,
                 projection=cartopy_proj,
--- a/lib/iris/plot.py
+++ b/lib/iris/plot.py
@@ -843,7 +843,9 @@ def _replace_axes_with_cartopy_axes(cartopy_proj):
                 ylabel=ax.get_ylabel(),
             )
         else:
+            position = ax.get_position()
             _ = fig.add_axes(
+                position,
                 projection=cartopy_proj,
                 title=ax.get_title(),
                 xlabel=ax.get_xlabel(),
--- a/lib/iris/tests/unit/plot/test__replace_axes_with_cartopy_axes.py
+++ b/lib/iris/tests/unit/plot/test__replace_axes_with_cartopy_axes.py
@@ -1,0 +1,43 @@
+# Copyright Iris contributors
+#
+# This file is part of Iris and is released under the LGPL license.
+# See COPYING and COPYING.LESSER in the root of the repository for full
+# licensing details.
+"""Unit tests for the `iris.plot.__replace_axes_with_cartopy_axes` function."""
+
+# Import iris.tests first so that some things can be initialised before
+# importing anything else.
+import iris.tests as tests  # isort:skip
+
+import cartopy.crs as ccrs
+import matplotlib.pyplot as plt
+
+from iris.plot import _replace_axes_with_cartopy_axes
+
+
+@tests.skip_plot
+class Test_replace_axes_with_cartopy_axes(tests.IrisTest):
+    def setUp(self):
+        self.fig = plt.figure()
+
+    def test_preserve_position(self):
+        position = [0.17, 0.65, 0.2, 0.2]
+        projection = ccrs.PlateCarree()
+
+        plt.axes(position)
+        _replace_axes_with_cartopy_axes(projection)
+        result = plt.gca()
+
+        # result should be the same as an axes created directly with the projection.
+        expected = plt.axes(position, projection=projection)
+
+        self.assertEqual(
+            expected.get_position().bounds, result.get_position().bounds
+        )
+
+    def tearDown(self):
+        plt.close(self.fig)
+
+
+if __name__ == "__main__":
+    tests.main()
--- a/lib/iris/tests/unit/plot/test__replace_axes_with_cartopy_axes.py
+++ b/lib/iris/tests/unit/plot/test__replace_axes_with_cartopy_axes.py
@@ -31,6 +31,8 @@ class Test_replace_axes_with_cartopy_axes(tests.IrisTest):
         # result should be the same as an axes created directly with the projection.
         expected = plt.axes(position, projection=projection)
 
+        # get_position returns mpl.transforms.Bbox object, for which equality does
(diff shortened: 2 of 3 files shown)


lib/iris/plot.py
```diff
@@ -844,7 +844,9 @@ def _replace_axes_with_cartopy_axes(cartopy_proj):
                 ylabel=ax.get_ylabel(),
             )
         else:
+            position = ax.get_position()
```
QuLogic, line 847: You may want the original position?
```suggestion
            position = ax.get_position(original=True)
```

rcomer: Thanks @QuLogic!

lib/iris/plot.py
```diff
@@ -845,7 +845,9 @@ def _replace_axes_with_cartopy_axes(cartopy_proj):
                 ylabel=ax.get_ylabel(),
             )
         else:
+            position = ax.get_position(original=True)
```
wjbenfold, line 848: I'd have thought that if the axes have been moved we'd want to keep that change in the new axes?

lib/iris/plot.py
```diff
@@ -845,7 +845,9 @@ def _replace_axes_with_cartopy_axes(cartopy_proj):
...
             _ = fig.add_axes(
+                position,
                 projection=cartopy_proj,
                 title=ax.get_title(),
                 xlabel=ax.get_xlabel(),
```
wjbenfold, line 853: I find it odd that these arguments aren't exhaustive, and that cartopy doesn't offer a `GeoAxes.from_mpl_axes` method, but that's not relevant for this PR I don't think. Though the original issue mentions that title font isn't preserved by this currently either.

wjbenfold (changes requested):
Github wouldn't let me comment on the original discussion of `original` so I made a new comment. Also not sure about if we want to be more comprehensive with this whilst we're here.

lib/iris/tests/unit/plot/test__replace_axes_with_cartopy_axes.py
```diff
@@ -0,0 +1,45 @@
...
+import iris.tests as tests  # isort:skip
+
+import cartopy.crs as ccrs
+import matplotlib.pyplot as plt
+
+from iris.plot import _replace_axes_with_cartopy_axes
+
+
+@tests.skip_plot
+class Test_replace_axes_with_cartopy_axes(tests.IrisTest):
```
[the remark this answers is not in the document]
rcomer: I'm in two minds on this one.  On the one hand you could argue that anything that isn't about the position is out of scope for this PR.  On the other hand, we often do things that aren't strictly within scope as they come up (we certainly have for #3901!)  I think in this case I'm leaning towards not adding the extra tests, as they could be added just as easily in a follow-up PR if we wanted.

lib/iris/plot.py
rcomer: I have to admit that, before @QuLogic's comment, I hadn't realised an Axes had more than one position.  From the [docs](https://matplotlib.org/stable/api/_as_gen/matplotlib.axes.Axes.set_position.html#matplotlib.axes.Axes.set_position) it seems the only way they would be different is if you have set an aspect ratio.  Cartopy will set its own aspect ratio to make the map look sensible.  So I think it does make sense to override any "active" position and revert to the "original" position.

lib/iris/plot.py
rcomer: I've no idea why these specific attributes were chosen.  I think this function dates back to the very early days of Iris.  In my own use-cases, I think I always call `iplt`/`qplt` functions before doing any titles/labels anyway, so it's only really the position/gridspec that would matter.  Of course I may not be representative of the average user!

> cartopy doesn't offer a GeoAxes.from_mpl_axes method

I did raise a feature request in cartopy for that, but it seems it's not a goer https://github.com/SciTools/cartopy/issues/1756

rcomer:
Rebased and fixed the whatsnew :smirk:

wjbenfold (approved):

lib/iris/plot.py
wjbenfold: Issue raised (#4664)

rcomer:
Rebased again in hope of triggering the tests.  Looks like it worked.

rcomer:
Thanks @wjbenfold 🎉

PR #4285, opened 2021-08-17
Integrated whatsnew for v3.1 release (rc0)

pp-mo:
### Description
I think this replicates the way we have done it in the past : the "interim" separate whatsnew files 3.0.<x>.rst are combined into a single 3.1.rst covering all history since 3.0
My only precedent for this is what we did at [v2.2.1](https://github.com/SciTools/iris/releases/tag/v2.2.1)

---


lib/iris/__init__.py
```diff
@@ -103,7 +103,7 @@ def callback(cube, field, filename):
 
 
 # Iris revision.
-__version__ = "3.1.dev0"
+__version__ = "3.1.0rc0"
```
lbdreyer, line 106: I'm confused by this diff. You already changed the version string yesterday. The 3.1.x branch already has 3.1.0rc0(https://github.com/SciTools/iris/blob/v3.1.x/lib/iris/__init__.py#L106) so why is this trying to change is again?

lbdreyer: I see there are some conflicting files. Maybe try rebasing onto the 3.1.x branch ?

pp-mo: Sorry, I uses the previous branch, so I think it's the probably squash+merge.  I'll rebase onto 3.1.x

pp-mo: .. yes, I think that fixed it

pp-mo:
@lbdreyer sorry,  it looks like I failed to bring some links across.
I will rebuild + run checks locally this time.
Just fixing that ...

docs/src/whatsnew/3.1.rst
```diff
@@ -257,11 +312,44 @@ This document explains the changes made to Iris for this release
...
+
+#. `@rcomer`_ modified test modules so they run consistently under ``pytest`` and
+   ``nose``, and also fixed some minor issues with :class:`~iris.time.PartialDateTime`.
+   (:pull:`4249`)
+
+========================================
+Internal changes in v3.0.2 (27 May 2021)
+========================================
+
+#. `@jamesp`_ updated a test to the latest numpy version (:pull:`3977`) [``pre-v3.1.0``]
```
lbdreyer, line 334: should these still have the `pre-v3.1.0` ?

docs/src/whatsnew/3.1.rst
```diff
@@ -257,11 +312,44 @@ This document explains the changes made to Iris for this release
...
+Internal changes in v3.0.4 (22 July 2021)
+=========================================
+#. `@bjlittle`_ Unpinned the `cftime`_ package dependency within Iris in order
+   to allow use of the latest versions of `cftime`_, `cf-units`_ and `nc-time-axis`_.
+   (:pull:`4222`)
+
+#. `@rcomer`_ modified test modules so they run consistently under ``pytest`` and
+   ``nose``, and also fixed some minor issues with :class:`~iris.time.PartialDateTime`.
+   (:pull:`4249`)
+
```
lbdreyer, line 329: You have lost the note:
```
   Note that, we are forced to drop support for ``Python 3.6`` in this patch due to
   the third-party package dependencies required by (:pull:`4222`).
```
?

docs/src/whatsnew/3.1.rst
lbdreyer: Maybe include the note
```
   Note that, the above contributions labelled with ``pre-v3.1.0`` are part of the forthcoming
   Iris v3.1.0 release, but require to be included in this patch release.
```

docs/src/whatsnew/3.1.rst
pp-mo: There is a note about this in the release highlights, so I thought that might be enough.
I don't think we really need to explain _why_  :  this note in the bugfix release was only there because it had changed requirements from the original v3.0.0

docs/src/whatsnew/3.1.rst
pp-mo: Good spot !
I hadn't properly understood what these were there for.

docs/src/whatsnew/3.1.rst
```diff
@@ -93,6 +93,53 @@ This document explains the changes made to Iris for this release
 #. `@trexfeathers`_ fixed the "anomaly_log_colouring" gallery example to be compatible with the latest Matplotlib usage
    (:pull:`4115`)
 
+===================================
+Bugs fixed in v3.0.4 (22 July 2021)
+===================================
```
lbdreyer, line 98: Should these not be in the 3.0.rst whats new not the 3.1.rst?

docs/src/whatsnew/3.1.rst
pp-mo: I think it means that these entries should be in the 'main' release part, not the bugfix sections.
Doing that now.

docs/src/whatsnew/3.1.rst
pp-mo: Oh dear.
I thought not, but now I look back at what we did at 2.2.1 / 2.3.0 I think I was wrong.
So yes, I suppose the bug fix sections should all go into an updated version of the 3.0 whatsnew.

But I find it a  bit odd that the release notes do **_not_** explain everything that changed since the previous release
( you have to read the _updated_ version of the previous release notes, too )
Does this seem right to you @lbdreyer ?

docs/src/whatsnew/3.1.rst
```diff
@@ -180,6 +179,9 @@ This document explains the changes made to Iris for this release
 #. `@MHBalsmeier`_ updated and simplified non-conda installation on Debian-based distros.
    (:pull:`4260`)
 
+#. `@bjlittle`_ updated the ``intersphinx_mapping`` and fixed documentation
+   to use ``stable`` URLs for `matplotlib`_. (:pull:`4003`)
+
```
lbdreyer, line 184: I'm a bit confused by this...it makes sense that its in the 3.1. whats new, but we also included it in the 3.0.4 release so shouldn't it also be included in the 3.0 whats new?

pp-mo: No, I **_think_** that is what the `[pre-3.1.0]` labels were for.
If I get it right, they mark changes which was brought in from main + included in (e.g.) 3.0.4, because it was ~too hard to merge in the fixes without, but strictly aren't part of a bugfix and so belong with the 3.1 updates.

pp-mo: Though, I suppose we could have it in **_both_** places ???

lbdreyer: I'm still a bit conflicted about this (in future maybe we write down the intention behind these sort of labels, to avoid confusion!) but as they only concern internal changes/a small docs change, I'm not majorly concerned by it and happy to just keep them in the 3.1 whats new

docs/src/whatsnew/3.1.rst
lbdreyer: When I look at the v3.0 what's new I expect to see the changes for all v3.0 releases. This includes 3.0.0, 3.0.1, 3.0.2, 3.0.3, 3.0.4.
If after having cut 3.1, we have to go back and cut another 3.0 patch release (v3.0.5), I would expect to see these changes as part of the 3.0 what's new.

docs/src/whatsnew/index.rst
```diff
@@ -10,11 +10,7 @@ Iris versions.
 .. toctree::
    :maxdepth: 1
 
-   3.1.0.rst
-   3.0.4.rst
-   3.0.3.rst
-   3.0.2.rst
-   3.0.1.rst
```
lbdreyer, line 17: @pp-mo  Do you know why we had separate what's news for each bug release?
Was it just there to draw attention to the fact that there multiple 3.0.x releases?

lbdreyer (approved):
Thanks for all your work on this @pp-mo!
Once the CI is finished, I'll merge this in...

docs/src/whatsnew/3.1.rst
```diff
@@ -1,7 +1,7 @@
 .. include:: ../common_links.inc
 
-|iris_version| |build_date|
-***************************
+v3.1 (17 Aug 2021) [unreleased]
+*******************************
```
lbdreyer, line 4: I notice for the Iris 3 rc we included a date:
https://scitools-iris.readthedocs.io/en/v3.0.0rc0/whatsnew/3.0.html

lbdreyer: Sorry wasn't clear. I meant to say _just_ a date no `unreleased`

docs/src/whatsnew/index.rst
pp-mo: I don't know really.
However, I did find that we did _not_ do this with the 2.2.1 bugfix release : there we just has a [modified 2.2.rst file](https://github.com/SciTools/iris/tree/v2.2.1/docs/iris/src/whatsnew)

pp-mo: Also, the 3.0.4.rst incorporates all the entries from the other 3

docs/src/whatsnew/3.1.rst
pp-mo: Yes, I think that's effectively a _preliminary_ date -- the date of the RC.
Which we would change again we we make it a full release
(imho)

pp-mo: I've done the same, but left in the '[unreleased]' tag, because I thought that made sense..

pp-mo: You might also consider that it makes sense to label this with the 'rc0', 
but, again, we haven't done so in the past

PR #4291, opened 2021-08-23
Separate netcdf load and save

pp-mo:
### Description
Split the 'iris.fileformats.netcdf' module into separate load + save submodules.

---


pp-mo:
I thought this would be a useful time to clean this up, while I'm looking into UGRID saving.
But I'm now having second thoughts, as 
(a) it's rather more work than I thought, and
(b) doing this while 'main' and `mesh-data-model` branches are so far apart is bound to make merging hard.
So I'm going to bank these ideas here, close this PR + come back to it at a better time.

pp-mo:
I thought this would be a useful time to clean this up, while I'm looking into UGRID saving.
But I'm now having second thoughts, as
(a) it's rather more work than I thought, and
(b) doing this while 'main' and mesh-data-model branches are so far apart is bound to make merging hard.
So I'm going to bank these ideas here, close this PR + come back to it at a better time.

PR #4294, opened 2021-08-24
Some type annotations for `coords.py`

### LINKED ISSUE #3653: cube.collapsed fails with multi-dimensional string coordinates
Hi, I have a cube like that below, with with data from multi-ensemble member climate model runs covering different years. The 'Expt ID' coordinate contains the run ID corresponding to each ensemble member for each year. I get an error when I do cube.collapsed('year', iris.analysis.MEAN).

I did previously collapse the cube over a 'season' coordinate (since removed), where each season had three time values, so perhaps this issue only arises when an entire dimension is collapsed?

    print cube
    air_temperature / (K)               (time: 30; Ens member: 15; latitude: 145; longitude: 192)
         Dimension coordinates:
              time                           x              -             -               -
              Ens member                     -              x             -               -
              latitude                       -              -             x               -
              longitude                      -              -             -               x
         Auxiliary coordinates:
              season_year                    x              -             -               -
              year                           x              -             -               -
              Expt ID                        x              x             -               -

    cube_mean=cube.collapsed('year',iris.analysis.MEAN)
    
The full error message is given below. It seems that the problem is that iris tries joining the strings in 'Expt ID' into one long string, and then finds that this does not have the same size as the 'Ens member' dimension.
```
---------------------------------------------------------------------------
ValueError                                Traceback (most recent call last)
<ipython-input-13-43764f0e4a7d> in <module>()
----> 1 cube_seasmean.collapsed('year',iris.analysis.MEAN)

/network/home/aopp/watson/anaconda2/envs/main/lib/python2.7/site-packages/iris/cube.pyc in collapsed(self, coords, aggregator, **kwargs)
   3253                 local_dims = [coord_dims.index(dim) for dim in
   3254                               dims_to_collapse if dim in coord_dims]
-> 3255                 collapsed_cube.replace_coord(coord.collapsed(local_dims))
   3256 
   3257         untouched_dims = sorted(untouched_dims)

/network/home/aopp/watson/anaconda2/envs/main/lib/python2.7/site-packages/iris/cube.pyc in replace_coord(self, new_coord)
   1181             self.add_dim_coord(new_coord, dims[0])
   1182         else:
-> 1183             self.add_aux_coord(new_coord, dims)
   1184 
   1185         for factory in self.aux_factories:

/network/home/aopp/watson/anaconda2/envs/main/lib/python2.7/site-packages/iris/cube.pyc in add_aux_coord(self, coord, data_dims)
    964         if self.coords(coord):  # TODO: just fail on duplicate object
    965             raise ValueError('Duplicate coordinates are not permitted.')
--> 966         self._add_unique_aux_coord(coord, data_dims)
    967 
    968     def _check_multi_dim_metadata(self, metadata, data_dims):

/network/home/aopp/watson/anaconda2/envs/main/lib/python2.7/site-packages/iris/cube.pyc in _add_unique_aux_coord(self, coord, data_dims)
    996 
    997     def _add_unique_aux_coord(self, coord, data_dims):
--> 998         data_dims = self._check_multi_dim_metadata(coord, data_dims)
    999         self._aux_coords_and_dims.append([coord, data_dims])
   1000 

/network/home/aopp/watson/anaconda2/envs/main/lib/python2.7/site-packages/iris/cube.pyc in _check_multi_dim_metadata(self, metadata, data_dims)
    988                     raise ValueError(msg.format(dim, self.shape[dim],
    989                                                 metadata.name(), i,
--> 990                                                 metadata.shape[i]))
    991         elif metadata.shape != (1,):
    992             msg = 'Missing data dimensions for multi-valued {} {!r}'

ValueError: Unequal lengths. Cube dimension 0 => 15; metadata 'Expt ID' dimension 0 => 1.
```

A quick way of coming up with a solution is just to explicitly make a string that is the join of the individual strings, removing the string coord and re-adding the joined string as a scalar coord (I'm not good enough with iris to know if this would be very robust, but it seems to work for my case).
```
for coord in cube.aux_coords:
    if coord.ndim>1 and coord.dtype.char=='S':
        new_str='|'.join(coord.points.ravel())
        new_coord=iris.coords.AuxCoord(new_str, attributes=coord.attributes, long_name=coord.long_name, standard_name=coord.standard_name, units=coord.units, var_name=coord.var_name)
        cube.remove_coord(coord)
        cube.add_aux_coord(new_coord)
        
cube_mean=cube_.collapsed('year', iris.analysis.MEAN)  #now this works
```
It would be nice if iris would do something like this in cube.collapsed(). Even better would be a method that only collapses 'Expt ID' here along the dimension being collapsed, so the association of 'Expt ID' values with the 'Ens member' dimension would be maintained.

rcomer:
### Description

Added some type hints relevant to the changes in #5955 

---

### DIFF
--- a/lib/iris/coords.py
+++ b/lib/iris/coords.py
@@ -1591,7 +1591,7 @@ class Coord(_DimensionalMetadata):
         self._values = points
 
     @property
-    def bounds(self) -> RealData:
+    def bounds(self) -> Optional[RealData]:
         """Coordinate bounds values.
 
         The coordinate bounds values, as a NumPy array,
@@ -1727,7 +1727,7 @@ class Coord(_DimensionalMetadata):
         """Core points array at the core of this coord, which may be a NumPy array or a dask array."""
         return super()._core_values()
 
-    def core_bounds(self) -> RealOrLazyData:
+    def core_bounds(self) -> Optional[RealOrLazyData]:
         """Core bounds. The points array at the core of this coord, which may be a NumPy array or a dask array."""
         result = None
         if self.has_bounds():
--- a/lib/iris/coords.py
+++ b/lib/iris/coords.py
@@ -2107,8 +2107,7 @@ class Coord(_DimensionalMetadata):
         return Cell(point, bound)
 
     def collapsed(self, dims_to_collapse: Optional[Dims] = None) -> "Coord":
-        """
-        Returns a copy of this coordinate, which has been collapsed along the specified dimensions.
+        """Returns a copy of this coordinate, which has been collapsed along the specified dimensions.
 
         Replaces the points & bounds with a simple bounded region.
         """
--- a/lib/iris/coords.py
+++ b/lib/iris/coords.py
@@ -11,7 +11,7 @@ import copy
 from functools import lru_cache
 from itertools import zip_longest
 import operator
-from typing import Sequence, Union
+from typing import Iterable, Optional, Union
 import warnings
 import zlib
 
@@ -34,10 +34,14 @@ from iris.common import (
 import iris.exceptions
 import iris.time
 import iris.util
-import iris.warnings
 
 #: The default value for ignore_axis which controls guess_coord_axis' behaviour
 DEFAULT_IGNORE_AXIS = False
+
+# Define some typing aliases.
+Dims = Union[int, Iterable[int]]
+RealData = Union[np.ndarray, ma.MaskedArray]
+RealOrLazyData = Union[RealData, da.Array]
 
 
 class _DimensionalMetadata(CFVariableMixin, metaclass=ABCMeta):
@@ -240,24 +244,8 @@ class _DimensionalMetadata(CFVariableMixin, metaclass=ABCMeta):
         """Return a lazy array representing the dimensional metadata values."""
         return self._values_dm.lazy_data()
 
-<<<<<<< HEAD
-    def _core_values(self):
+    def _core_values(self) -> RealOrLazyData:
         """Value array of this dimensional metadata which may be a NumPy array or a dask array."""
-||||||| constructed merge base
-    def _core_values(self):
-        """
-        The values array of this dimensional metadata which may be a NumPy
-        array or a dask array.
-
-        """
-=======
-    def _core_values(self) -> Union[npt.NDArray, "da.Array"]:
-        """
-        The values array of this dimensional metadata which may be a NumPy
-        array or a dask array.
-
-        """
->>>>>>> add type hints
         result = self._values_dm.core_data()
         if not _lazy.is_lazy_data(result):
             result = result.view()
@@ -789,24 +777,8 @@ class _DimensionalMetadata(CFVariableMixin, metaclass=ABCMeta):
         return self._values_dm.dtype
 
     @property
-<<<<<<< HEAD
-    def ndim(self):
-        """Return the number of dimensions of the current dimensional metadata object."""
-||||||| constructed merge base
-    def ndim(self):
-        """
-        Return the number of dimensions of the current dimensional metadata
-        object.
-
-        """
-=======
     def ndim(self) -> int:
-        """
-        Return the number of dimensions of the current dimensional metadata
-        object.
-
-        """
->>>>>>> add type hints
+        """Return the number of dimensions of the current dimensional metadata object."""
         return self._values_dm.ndim
 
     def has_bounds(self):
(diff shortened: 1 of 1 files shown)


lib/iris/coords.py
```diff
@@ -1948,24 +1948,35 @@ def collapsed(self, dims_to_collapse=None):
...
+
+                return np.array(joined).reshape(out_shape)
 
             bounds = None
             if self.has_bounds():
-                shape = self._bounds_dm.shape[1:]
                 bounds = []
-                for index in np.ndindex(shape):
-                    index_slice = (slice(None),) + tuple(index)
-                    bounds.append(serialize(self.bounds[index_slice]))
```
rcomer, line 1960: If I've understood the old slicing method, it and the new method are consistent for the basic case of 2D bounds (n_points, n_bounds).  For higher dimensions, the old method doesn't look right to me: if we had a 2D (n, m) coordinated with bounds (n, m, n_bounds), it looks like we would get `m * n_bounds` elements in the list.  But I **think** we want just `n_bounds` elements.

I didn't even know bounded string coords were a thing!

lib/iris/coords.py
```diff
@@ -1948,24 +1948,35 @@ def collapsed(self, dims_to_collapse=None):
...
+                for index in range(self.nbounds):
+                    bounds_slice = np.take(self.bounds, index, axis=-1)
+                    serialized = serialize(bounds_slice, dims_to_collapse)
+                    bounds.append(serialized)
+                # Make sure bounds dim comes last.
+                bounds = np.moveaxis(bounds, 0, -1)
+            points = serialize(self.points, dims_to_collapse)
             # Create the new collapsed coordinate.
-            coord = self.copy(
-                points=np.array(points, dtype=dtype), bounds=bounds
```
rcomer, line 1967: I don't know why it was necessary to set the `dtype`, but the existing tests pass fine without this now.  The old code is 8 years old, so I guess numpy has moved on and is likely doing more for us automatically.

lib/iris/coords.py
```diff
@@ -1948,24 +1948,38 @@ def collapsed(self, dims_to_collapse=None):
         if np.issubdtype(self.dtype, np.str_):
             # Collapse the coordinate by serializing the points and
             # bounds as strings.
-            def serialize(x):
-                return "|".join([str(i) for i in x.flatten()])
+            def serialize(x, axis):
+                if axis is None or len(axis) == x.ndim:
```
rcomer, line 1952: I'm not 100% sure about this `len(axis) == x.ndim`.  Is there any way there could be repeated values in `axis`?

rcomer: Removed.

rcomer:
Rebased.  CI is green.  Ready for review 😎

rcomer:
Added some type hints for the method and function I modified, as well as ones required to tidy up PyRight complaints within the `collapsed` method.  The form `import numpy.typing as npt` is taken from [examples in the numpy docs](https://numpy.org/doc/1.21/reference/typing.html).

rcomer:
Some screenshots of the built docs:

ndim_property
bounds_property
core_bounds_method
collapsed_method

lib/iris/coords.py
```diff
@@ -1933,7 +1935,9 @@ def cell(self, index):
 
         return Cell(point, bound)
 
-    def collapsed(self, dims_to_collapse=None):
+    def collapsed(
+        self, dims_to_collapse: Union[int, Sequence[int], None] = None
```
bjlittle, line 1939: @rcomer For this you could use...

```python
from typing import Iterable, Union

# type aliases
Dims = Union[int, Iterable[int]]

def collapsed(
    self, dims_to_collapse: Optional[Dims] = None
```
Which is slightly neater. See https://docs.python.org/3/library/typing.html#typing.Optional

IMHO with typing there certainly seems to be several ways to skin the one cat...

rcomer:
Made some changes to the type hints following a chat with @bjlittle this morning, and taking some inspiration from @jamesp's [stubs](https://github.com/jamesp/iris-stubs/blob/main/iris-stubs/cube.pyi).  The rendered docs look better:

bounds
core_points_and_bounds
collapsed

bjlittle:
@bjlittle get your finger out... <nudge>

bjlittle:
@rcomer Fancy doing a rebase?

rcomer:
Now rebased.

@bjlittle it's all on you now 😛

rcomer:
See #5955 for a version of this without the type-hinting rabbit hole.

rcomer:
The bugfix has now been merged in #5955 so I have rebased this to only include the type changes.  It seems that mypy is unhappy though and I'm not sure if/when I will be motivated to investigate that.

PR #4297, opened 2021-08-27
Pre ugridsave refactor

pp-mo:
### Description
Refactoring netcdf variable creation routines.
Now squashed down to 2 commits :
  1. add lost of extra assertCDL-like testing, for ~anything that uses iris.save
  2. do some refactoring
---
[Consult Iris pull request check list]( https://scitools-iris.readthedocs.io/en/latest/developers_guide/contributing_pull_request_checklist.html)

PR #4301, opened 2021-08-31
Refactor some netcdf save code

pp-mo:
### Description
Initial "tidying" to routines in `netcdf.py`, to prepare for the addition of UGRID saving support.

This unifies the handling codepaths for different types of cube components, 
to make it easier to add the new ones which are coming (mesh-element coords and connectivities).

**This is a _pure refactor_ :** by design, should not change any exisiting behaviour/results.

These same code changes have been proven against extra CDL test snapshots in #4297
 .. but we don't ever intend to **_merge_** that extra testing, as it is excessive and somewhat incomplete/unpolished

---

### DIFF
--- a/lib/iris/fileformats/netcdf.py
+++ b/lib/iris/fileformats/netcdf.py
@@ -40,6 +40,7 @@ from iris.aux_factory import (
 import iris.config
 import iris.coord_systems
 import iris.coords
+from iris.coords import AncillaryVariable, AuxCoord, CellMeasure, DimCoord
 import iris.exceptions
 import iris.fileformats.cf
 import iris.io
@@ -1356,40 +1357,50 @@ class Saver:
                 self._dataset.createDimension(dim_name, size)
 
     def _add_inner_related_vars(
-        self,
-        cube,
-        cf_var_cube,
-        dimension_names,
-        coordlike_elements,
-        role_attribute_name,
+        self, cube, cf_var_cube, dimension_names, coordlike_elements
     ):
-        # Common method to create a set of file variables and attach them to
-        # the parent data variable.
-        element_names = []
-        # Add CF-netCDF variables for the associated auxiliary coordinates.
-        for element in sorted(
-            coordlike_elements, key=lambda element: element.name()
-        ):
-            # Create the associated CF-netCDF variable.
-            if element not in self._name_coord_map.coords:
-                cf_name = self._create_generic_cf_array_var(
-                    cube, dimension_names, element
-                )
-                self._name_coord_map.append(cf_name, element)
+        """
+        Create a set of variables for aux-coords, ancillaries or cell-measures,
+        and attach them to the parent data variable.
+
+        """
+        if coordlike_elements:
+            # Choose the approriate parent attribute
+            elem_type = type(coordlike_elements[0])
+            if elem_type in (AuxCoord, DimCoord):
+                role_attribute_name = "coordinates"
+            elif elem_type == AncillaryVariable:
+                role_attribute_name = "ancillary_variables"
             else:
-                cf_name = self._name_coord_map.name(element)
+                # We *only* handle aux-coords, cell-measures and ancillaries
+                assert elem_type == CellMeasure
+                role_attribute_name = "cell_measures"
 
-            if cf_name is not None:
-                if role_attribute_name == "cell_measures":
-                    # In the case of cell-measures, the attribute entries are not just
-                    # a var_name, but each have the form "<measure>: <varname>".
-                    cf_name = "{}: {}".format(element.measure, cf_name)
-                element_names.append(cf_name)
+            # Add CF-netCDF variables for the given cube components.
+            element_names = []
+            for element in sorted(
+                coordlike_elements, key=lambda element: element.name()
+            ):
+                # Create the associated CF-netCDF variable.
+                if element not in self._name_coord_map.coords:
+                    cf_name = self._create_generic_cf_array_var(
+                        cube, dimension_names, element
+                    )
+                    self._name_coord_map.append(cf_name, element)
+                else:
+                    cf_name = self._name_coord_map.name(element)
 
-        # Add CF-netCDF references to the primary data variable.
-        if element_names:
-            variable_names = " ".join(sorted(element_names))
-            _setncattr(cf_var_cube, role_attribute_name, variable_names)
+                if cf_name is not None:
+                    if role_attribute_name == "cell_measures":
+                        # In the case of cell-measures, the attribute entries are not just
+                        # a var_name, but each have the form "<measure>: <varname>".
+                        cf_name = "{}: {}".format(element.measure, cf_name)
+                    element_names.append(cf_name)
+
+            # Add CF-netCDF references to the primary data variable.
+            if element_names:
+                variable_names = " ".join(sorted(element_names))
+                _setncattr(cf_var_cube, role_attribute_name, variable_names)
 
     def _add_aux_coords(self, cube, cf_var_cube, dimension_names):
         """
@@ -1410,7 +1421,6 @@ class Saver:
             cf_var_cube,
             dimension_names,
             cube.aux_coords,
-            "coordinates",
         )
 
     def _add_cell_measures(self, cube, cf_var_cube, dimension_names):
@@ -1432,7 +1442,6 @@ class Saver:
             cf_var_cube,
             dimension_names,
             cube.cell_measures(),
-            "cell_measures",
         )
 
     def _add_ancillary_variables(self, cube, cf_var_cube, dimension_names):
@@ -1455,7 +1464,6 @@ class Saver:
             cf_var_cube,
             dimension_names,
             cube.ancillary_variables(),
-            "ancillary_variables",
         )
 
     def _add_dim_coords(self, cube, dimension_names):
@@ -1941,14 +1949,11 @@ class Saver:
                 if axis is not None and axis.lower() in SPATIO_TEMPORAL_AXES:
                     _setncattr(cf_var, "axis", axis.upper())
 
-            # Add the data to the CF-netCDF variable.
-            cf_var[:] = data  # TODO: support dask streaming
-
             # Create the associated CF-netCDF bounds variable, if any.
             self._create_cf_bounds(element, cf_var, cf_name)
 
         # Add the data to the CF-netCDF variable.
-        cf_var[:] = data
+        cf_var[:] = data  # TODO: support dask streaming
 
         # Deal with CF-netCDF units and standard name.
         if isinstance(element, iris.coords.Coord):
--- a/lib/iris/fileformats/netcdf.py
+++ b/lib/iris/fileformats/netcdf.py
@@ -1805,7 +1805,7 @@ class Saver:
 
         * cube (:class:`iris.cube.Cube`):
             The cube that contains the given coordinate.
-        * coord (:class:`iris.coords.DimensionalMetadata`):
+        * coord (:class:`iris.coords._DimensionalMetadata`):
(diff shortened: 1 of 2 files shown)


lib/iris/fileformats/netcdf.py
```diff
@@ -1361,7 +1361,6 @@ def _add_inner_related_vars(
         cf_var_cube,
         dimension_names,
         coordlike_elements,
-        saver_create_method,
         role_attribute_name,
```
bjlittle, line 1364: @pp-mo We could take the opportunity to drop the `role_attribute_name` argument, and instead use the `element` within `_add_inner_related_vars` to lookup the `_ELEMENT_TYPE_NAMES`?

lib/iris/fileformats/netcdf.py
```diff
@@ -1825,166 +1825,57 @@ def _get_coord_variable_name(self, cube, coord):
...
+        element_dims = [
+            cube_dim_names[dim] for dim in element.cube_dims(cube)
+        ]  # NB using 'cube_dims' as this works for any type of element
+
+        # Get the data values, in a way which works for any element type, as
+        # all are subclasses of _DimensionalMetadata.
+        # (e.g. =points if a coord, =data if an ancillary, etc)
+        data = element._values
+
+        if np.issubdtype(data.dtype, np.str_):
```
bjlittle, line 1874: @pp-mo This is potentially new behaviour for `cell-measures` and `ancillary-variables`, right?

lib/iris/fileformats/netcdf.py
```diff
@@ -1994,77 +1885,97 @@ def _create_cf_coord_variable(self, cube, dimension_names, coord):
...
-            if coord in cf_coordinates:
-                axis = iris.util.guess_coord_axis(coord)
+            if is_dimcoord:
+                axis = iris.util.guess_coord_axis(element)
                 if axis is not None and axis.lower() in SPATIO_TEMPORAL_AXES:
                     _setncattr(cf_var, "axis", axis.upper())
 
             # Add the data to the CF-netCDF variable.
-            cf_var[:] = points
+            cf_var[:] = data  # TODO: support dask streaming
```
bjlittle, line 1945: @pp-mo This data is being written twice :scream: 

See `line +1951`... right?

bjlittle: If you agree, then your `TODO` applies to `line +1951` instead

lib/iris/fileformats/netcdf.py
```diff
@@ -1994,77 +1885,97 @@ def _create_cf_coord_variable(self, cube, dimension_names, coord):
...
 
+        long_name = element.long_name
         if long_name is not None:
             _setncattr(cf_var, "long_name", long_name)
 
         # Add the CF-netCDF calendar attribute.
-        if coord.units.calendar:
-            _setncattr(cf_var, "calendar", coord.units.calendar)
+        if element.units.calendar:
+            _setncattr(cf_var, "calendar", str(element.units.calendar))
```
bjlittle, line 1974: @pp-mo I *think* a `calendar` attribute is only applicable to a `time` coordinate, and not a cell-measure or ancillary variable, see https://cfconventions.org/cf-conventions/cf-conventions.html#calendar

bjlittle (changes requested):
@pp-mo Some comments to service, but this is looking good.

Thanks for taking the time to rationalise and tee things up for #4303 :+1:

pp-mo:
Re-opened, and merged from _new_ update 'mesh-data-model'.
So, this ought to pass tests ..

lib/iris/fileformats/netcdf.py
```diff
@@ -1793,15 +1791,15 @@ def _get_cube_variable_name(self, cube):
...
+        Returns a CF-netCDF variable name for a given coordinate-like element.
 
         Args:
 
         * cube (:class:`iris.cube.Cube`):
             The cube that contains the given coordinate.
-        * coord (:class:`iris.coords.Coord`):
-            An instance of a coordinate for which a CF-netCDF variable
-            name is required.
+        * coord (:class:`iris.coords.DimensionalMetadata`):
```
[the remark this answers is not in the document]
pp-mo: No, they really _are_ all ultimately _DimensionalMetadata, by inheritance
 \-- though I may have missed out a dunder ??

This is to be used with Coord, CellMeasure and Ancillary, 
and in future will add mesh coordinates (AuxCoord) and connectivities (Connectivity)
Inheritance is:
  * coords.Coord -> _DimensionalMetadata
  * coords.AncillaryVariable -> _DimensionalMetadata
  * coords.CellMeasure -> AncillaryVariable (-> _DimensionalMetadata)
  * ugrid.Connectivity -> _DimensionalMetadata

lib/iris/fileformats/netcdf.py
```diff
@@ -1812,6 +1810,8 @@ def _get_coord_variable_name(self, cube, coord):
         else:
             name = coord.standard_name or coord.long_name
             if not name or set(name).intersection(string.whitespace):
+                # NB don't know how to fix this if it is not a Coord or similar
+                assert isinstance(coord, iris.coords.Coord)
```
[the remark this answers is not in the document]
pp-mo: Because we are about to use 'coord_dims'.
We could turn that around and use 'element.cube_dims(cube)', but this still doesn't work for Connectivities
Because Connectivities, of course, **aren't** mapped to cube dims.

In fact, anything which is a _DimensionalMetadata [has a cube_dims](https://github.com/SciTools/iris/blob/6c2dc49122d4c862e4d6a3e16bd0eeb4c49d09b0/lib/iris/coords.py#L185-L199), 
but we made a (maybe slightly nasty) exception by [disabling the cube_dims function on Connectivities](https://github.com/SciTools/iris/blob/6c2dc49122d4c862e4d6a3e16bd0eeb4c49d09b0/lib/iris/experimental/ugrid/__init__.py#L628-L630)

lib/iris/fileformats/netcdf.py
```diff
@@ -1825,166 +1825,57 @@ def _get_coord_variable_name(self, cube, coord):
...
             The associated cube being saved to CF-netCDF file.
-        * dimension_names (list):
-            Names for each dimension of the cube.
-        * coord (:class:`iris.coords.Coord`):
-            The coordinate to be saved to CF-netCDF file.
+        * cube_dim_names (list of string):
+            The name of each dimension of the cube.
+        * element:
+            An Iris :class:`iris.coords.DimensionalMetadata`, belonging to the
+            cube.  Provides data, units and standard/long/var names.
```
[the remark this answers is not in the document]
pp-mo: No, I think this **_is_** the most general superclass that we require.  See previous.

lib/iris/fileformats/netcdf.py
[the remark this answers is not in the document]
pp-mo: Actually, a lot of the legacy code is still use 'coord' pretty loosely, where it now deals with other ~similar types of data. That is both in variable naming and docstrings.

If you think this is important, we can have a general blitz on it.  But I suggest we leave it until the last, so as to avoid generating a blizzard of extra changes to review.

lib/iris/fileformats/netcdf.py
```diff
@@ -1825,166 +1825,57 @@ def _get_coord_variable_name(self, cube, coord):
...
-        # Note: currently shares variable creation code with ancillary-variables.
-        return self._inner_create_cf_cellmeasure_or_ancil_variable(
-            cube, dimension_names, cell_measure
-        )
+    _ELEMENT_TYPE_NAMES = {
+        iris.coords.DimCoord: "coordinate",
+        iris.coords.AuxCoord: "coordinate",
+        iris.coords.CellMeasure: "cell-measure",
+        iris.coords.AncillaryVariable: "ancillary-variable",
+    }
```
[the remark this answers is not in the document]
pp-mo: Good Question.
In subsequent development, I have now got rid of this + just used the class names themselves,
since it's only actually used to generate warning messages inside '_ensure_valid_dtype`.

lib/iris/fileformats/netcdf.py
pp-mo: True, but I thought in that case the unit simply wouldn't ever have a calendar ?

lib/iris/fileformats/netcdf.py
pp-mo: :+1:  will fix this

lib/iris/fileformats/netcdf.py
pp-mo: Potentially yes.
This code is now unified usage for coordinates, cell-measures and ancillaries (and in future will be applied to connectivities and mesh-location coordinates).
From unifying the code, there is in fact a fair bit of code here that used to be applied only to coordinates but now can trigger with other types of things, so we need to consider that carefully on a case-by-case basis.

In this particular case, I think string-typed  ancillaries are a reasonable possibility, and this ought to handle them OK.
I don't think string cell-measures are at all likely.
However I _think_ the logic here is pretty general : when you have a array of data to process, and it might sometimes by strings.

Other cases like this :
  * units calendar handling : see comment 
  * the odd sections I made special to dim-coords : should be pretty safe.
  * special handling of STASH attributes : could apply to AuxCoords in some cases, I think.  
    We could narrow the scope when this is applied, but I don't think it's very dangerous as-is.

lib/iris/fileformats/netcdf.py
pp-mo: :+1: I've now added code to test the type of the arguments

lib/iris/fileformats/netcdf.py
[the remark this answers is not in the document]
bjlittle: @pp-mo Cool. Understood. Thanks :+1: 

So it's `iris.coords._DimensionalMetadata` then, and not `iris.coords.DimensionalMetadata`, which doesn't exist.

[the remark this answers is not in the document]
pp-mo: :+1: will check that out + fix it..

lib/iris/fileformats/netcdf.py
bjlittle: @pp-mo Cool. I guess we'll see if use cases arrive that exercise this, but at this moment in time I'm not worried if you're not.

I just wanted to confirm that this would be new behaviour for `cell-measures` and `ancillary-variables`, thanks for clarifying :+1:

lib/iris/fileformats/netcdf.py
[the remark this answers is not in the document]
bjlittle: TBH I'm not the biggest fan of the `cube_dims` method, and it exposes all sorts of questionable behavioural awkwardness/weirdness for `Connectivity`... so it would be kinda nice not to build on that; we'll see I guess.

lib/iris/fileformats/netcdf.py
```diff
@@ -1825,166 +1833,57 @@ def _get_coord_variable_name(self, cube, coord):
...
         * cube (:class:`iris.cube.Cube`):
             The associated cube being saved to CF-netCDF file.
-        * dimension_names (list):
-            Names for each dimension of the cube.
-        * coord (:class:`iris.coords.Coord`):
-            The coordinate to be saved to CF-netCDF file.
+        * cube_dim_names (list of string):
+            The name of each dimension of the cube.
+        * element:
+            An Iris :class:`iris.coords.DimensionalMetadata`, belonging to the
```
bjlittle, line 1859:
```suggestion
            An Iris :class:`iris.coords._DimensionalMetadata`, belonging to the
```

bjlittle (approved):

pp-mo:
:man_dancing: :rocket: :sun_with_face:

PR #4309, opened 2021-09-12
Made clear we only test on Linux.

tkknight:
### Description

Updated the **documentation** to make clear that the **testing** we do is based on **Linux** (and not Windows or macOS).

I'm leaning to not create a what's news for this minor change.  If you think otherwise please comment.

---


bjlittle (approved):

trexfeathers:
>I'm leaning to not create a what's news for this minor change. If you think otherwise please comment.

I certainly wouldn't insist on it. It's something I personally like to do since I find it informative to see just how many iterative changes we make between releases, and I hope that it makes users realise how hard we work even if we haven't addressed a particular concern of theirs.

PR #4313, opened 2021-09-14
Merge `mesh-data-model`

trexfeathers:
The long awaited merge of `mesh-data-model` into `main`!

### Should this be a squash or a merge commit?


pp-mo (approved):
Looks as we would expect.  Lots of noise from isort !

PR #4315, opened 2021-09-15
cube.aggregated_by output bounds

### LINKED ISSUE #1528: cube.aggregated_by produces wrong bounds in some cases
Try the following:

```python
import iris
import iris.cube
import numpy as np

cube = iris.cube.Cube(np.zeros((12, 2, 2)), standard_name='wind_speed', units='m s-1')
co_y = iris.coords.AuxCoord([1996, 1997, 1997, 1996, 1997, 1997, 1996, 1997, 1997, 1996, 1997, 1997], long_name='year')
co_m = iris.coords.AuxCoord(['Dec', 'Jan', 'Feb']*4, long_name='month')
co_r = iris.coords.AuxCoord([2, 2, 2, 3, 3, 3, 4, 4, 4, 1, 1, 1], standard_name=u'realization')
cube.add_aux_coord(co_y, 0)
cube.add_aux_coord(co_m, 0)
cube.add_aux_coord(co_r, 0)

print(cube)

means = cube.aggregated_by(['month','year'],iris.analysis.MEAN)
print('means.coord(\'realization\'):\n', means.coord('realization'))

```

Result is:

```
>>> print cube
wind_speed / (m s-1)                (-- : 12; -- : 2; -- : 2)
     Auxiliary coordinates:
          month                         x        -       -
          realization                   x        -       -
          year                          x        -       -
>>> 
>>> means = cube.aggregated_by(['month','year'],iris.analysis.MEAN)
>>> print 'means.coord(\'realization\'):\n', means.coord('realization')
means.coord('realization'):
AuxCoord(array([ 1.5,  1.5,  1.5]), bounds=array([[2, 1],
       [2, 1],
       [2, 1]]), standard_name=u'realization', units=Unit('1'))
>>> 
```

**But we expected** 
- all points == 2 (instead of 1.5)
- all bounds == [1, 4] (instead of [2, 1])

**Note:** 
Trying subsets of the data suggests that it is the last 3 realization values being out-of-order which triggers the problem...

```
>>> means = cube[:9].aggregated_by(['month','year'],iris.analysis.MEAN)
>>> print means.coord('realization')
AuxCoord(array([ 3.,  3.,  3.]), bounds=array([[2, 4],
       [2, 4],
       [2, 4]]), standard_name=u'realization', units=Unit('1'))
```

```
>>> means = cube[6:].aggregated_by(['month','year'],iris.analysis.MEAN)
>>> print means.coord('realization')
AuxCoord(array([ 2.5,  2.5,  2.5]), bounds=array([[4, 1],
       [4, 1],
       [4, 1]]), standard_name=u'realization', units=Unit('1'))
```

rcomer:
### Description

I've had this one in my sights for some time, as I have workarounds for it in my user code.

A step towards closing #1528.  If a coordinate's cells are not monotonic, we now calculate the new bounds as the minimum and maximum from the relevant slice, rather than first and last cell from the slice.  The example given in #1528 now gives us the bounds we expect
```
means.coord('realization'):
 AuxCoord(array([2.5, 2.5, 2.5]), bounds=array([[1, 4],
       [1, 4],
       [1, 4]]), standard_name='realization', units=Unit('unknown'))
```
Note that @pp-mo also said in #1528 that we expect the points to all be 2.  That would be consistent with what we get if we make the same calculation via `extract` and `collapsed`:
```python
for year, month in zip([1996, 1997, 1997], ['Dec', 'Jan', 'Feb']):
    constraint = iris.Constraint(year=year, month=month)
    mean = cube.extract(constraint).collapsed('realization', iris.analysis.MEAN)
    print(mean.coord('realization'))
```
```
AuxCoord(array([2]), bounds=array([[1, 4]]), standard_name='realization', units=Unit('unknown'))
AuxCoord(array([2]), bounds=array([[1, 4]]), standard_name='realization', units=Unit('unknown'))
AuxCoord(array([2]), bounds=array([[1, 4]]), standard_name='realization', units=Unit('unknown'))
```

I think it we should make `collapsed` and `aggregated_by` consistent on this, but I'm unclear which of them currently has it right.  I also suspect that making changes to the output points of either method should wait for a major release: I did try updating the points dtype to match the bounds dtype within `aggregated_by`, and got 9 test failures.

My current proposed change to bounds handling only affects coordinates that started out in no particular order.  It's hard to see how the existing behaviour is useful in those cases, so seems safe enough to call this a bugfix.

Note that the first two commits here are just a refactor of the existing code to reduce repetition, and do not affect functionality.

---


lib/iris/tests/unit/cube/test_Cube.py
```diff
@@ -757,7 +757,29 @@ def test_agg_by_label(self):
         res_cube = self.cube.aggregated_by("label", self.mock_agg)
         val_coord = AuxCoord(
             np.array([1.0, 0.5, 1.0]),
-            bounds=np.array([[0, 2], [0, 1], [2, 0]]),
+            bounds=np.array([[0, 2], [0, 1], [0, 2]]),
```
lbdreyer, line 760: Looks much better!!

lbdreyer (approved):
This is great! Thanks @rcomer !

I completely agree with your assessment; the handling of the bounds is a bug fix, whereas deciding how aggregated_by and extract/collapse should handle the points is more likely to be considered a breaking change and should wait for a major release.

rcomer:
Thanks @lbdreyer!  I have opened #4455 to capture the remaining points issue and so closed #1528.

PR #4318, opened 2021-09-15
Ugrid save

pp-mo:
### Description
Support for saving cubes with meshes.
(No "save_mesh" yet, easy to add that later)

### Current status : 
Good for review.  

Some additional questions to consider ..
  - [x] ~should we minimise the number of small/bounds dimensions created ?~ See [this comment](https://github.com/SciTools/iris/pull/4318#discussion_r709210943)
  - [x] ~should we be augmenting the global 'Conventions' attribute See [this comment](https://github.com/SciTools/iris/pull/4318#issuecomment-924051616)~
  - [x] ~should the cube-summary changes be pulled out into a separate PR ?~

---


lib/iris/tests/results/integration/experimental/ugrid_save/TestBasicSave/ugrid_ex2_2d_triangular.cdl
```diff
@@ -0,0 +1,75 @@
+dimensions:
+	Mesh2_edge_N_faces = 2 ;
```
pp-mo, line 2: N.B. this use of separate dims for the last dim of different connectivity types is by design,
but it could seem wasteful.
we could change this, even create all "small" dimension values only once + re-use them

lib/iris/tests/results/integration/experimental/ugrid_save/TestBasicSave/ugrid_ex2_2d_triangular.cdl
```diff
@@ -0,0 +1,75 @@
...
+	Mesh2_edge_N_nodes = 2 ;
+	Mesh2_face_N_edges = 3 ;
+	Mesh2_face_N_faces = 3 ;
+	Mesh2_face_N_nodes = 3 ;
```
pp-mo, line 6: N.B. this use of separate dims for the last dim of different connectivity types is by design,
but it could seem wasteful.
We could change this, even create all "small" dimension values only once + re-use them

lib/iris/_representation/cube_summary.py
```diff
@@ -259,8 +259,17 @@ def __init__(self, cube, shorten=False, name_padding=35):
         vector_dim_coords = [
             coord for coord in dim_coords if id(coord) not in scalar_coord_ids
         ]
+        if cube.mesh is None:
```
pp-mo, line 262: These changes I've just done this for now, as it was irritating me + making debugging hard (not to easily see when there is a mesh or not).

**TODO: pull out these changes (1 commit) and put them in their own PR.**

lib/iris/experimental/ugrid/__init__.py
```diff
@@ -522,6 +522,8 @@ def __eq__(self, other):
             if hasattr(other, "metadata"):
                 # metadata comparison
                 eq = self.metadata == other.metadata
+                if eq:
```
pp-mo, line 525: This fixes it so that connectivities of different shapes can compare correctly (i.e. returns False).
Without this, I was getting errors from the next code lines due to problems in array comparison.

pp-mo:
General question : **should we be adding some kind of UGRID reference to the global 'Conventions' attribute ?**

 * Iris currently just uses `Conventions = "CF=1.7"` : see [here](https://github.com/SciTools/iris/blob/v3.1.0/lib/iris/fileformats/netcdf.py#L109) and [here](https://github.com/SciTools/iris/blob/v3.1.0/lib/iris/fileformats/netcdf.py#L2812-L2813)
 * LFRic/XIOS output files currently have `Conventions = "UGRID"` : see [here](https://github.com/SciTools/iris/blob/main/lib/iris/tests/stock/file_headers/xios_2D_face_half_levels.cdl#L57)
 * in pure netcdf terms, we can apparently have "CF UGRID", but there is no special mention of of meaning for _versions_ : See [here](https://www.unidata.ucar.edu/software/netcdf/conventions.html)
 * there is no particular agreement within either CF or UGRID for how to do this, or refer to the other
   * i.e. does "UGRID" **_imply_** CF ?  That doesn't seem to work as the versions are explicitly connected (at least yet)
   * there is discussion [here](https://github.com/cf-convention/cf-conventions/issues/153) but no clear winning solution yet

lib/iris/_representation/cube_summary.py
bjlittle: @pp-mo I can see why you done this here, makes total sense.

You've added testing in this PR to cover this change, so I don't really see the advantage in separating concerns here to end up at the same destination.

I think we're inevitably going to have to add extra code (e.g., updating the cube summary) when we integrate UGRID features into `iris`. 

So thanks for making the change, and I'm happy with this as it stands, if you are?

lib/iris/fileformats/netcdf.py
```diff
@@ -977,6 +977,11 @@ def __setitem__(self, keys, arr):
         self.target[keys] = arr
 
 
+# NOTE : this matches :class:`iris.experimental.ugrid.Mesh.LOCATIONS`,
+# but in a specific preferred order.
```
bjlittle, line 981: @pp-mo Begs the obvious question... what's the reasoning behind the specific order?

If a developer was to change the order, what would it mean?

pp-mo: It really just controls which order the coords + connectivities appear in the file.
Let me try fixing the comment to make that clear...

lib/iris/_representation/cube_summary.py
pp-mo: Yes if you're happy.
I have already included the minimal test additions + changes I thought necessary.

The need for this really became apparent when I was collapsing the mesh on cubes to allow cube comparisons in the round-trip tests :  The resulting replacement of MeshCoords with AuxCoords is completely invisible -- the cube printout was just the same.

It's worth considering, though, that this is still a "minimal" approach, compared with ways we have considered making a structured cube visible in the past :  It does not show anyt mesh details -- most notably, it **_doesn't_** say what mesh location the cube maps to.  Nor does it let you judge whether 2 cubes _share_ a mesh (e.g. by a mesh name)

lib/iris/fileformats/netcdf.py
pp-mo:
```suggestion
# but in the preferred order for coord/connectivity variables in the file.
```

pp-mo: ? ok ?

lib/iris/tests/unit/fileformats/netcdf/test_Saver__ugrid.py
```diff
@@ -42,6 +42,19 @@ def build_mesh(
     Mesh has faces edges, face-coords and edge-coords, numbers of which can be
     controlled.
 
+    Args:
+
+    * n_nodes, n_faces, n_edges (int):
+        Basic dimensions of mesh components.  Zero means no such location.
+    * nodecoord_xyargs, nodecoord_xyargs, nodecoord_xyargs (pair of dict):
```
pp-mo, line 49:
```suggestion
    * nodecoord_xyargs, edgecoord_xyargs, facecoord_xyargs (pair of dict):
```

lib/iris/fileformats/netcdf.py
```diff
@@ -1593,57 +1725,158 @@ def _get_dim_names(self, cube):
...
-            List of dimension names with length equal the number of dimensions
-            in the cube.
+            cube_dimensions, mesh_dimensions
+            * cube_dimensions (list of string):
+                A lists of dimension names for each dimension of the cube
+            * mesh_dimensions (list of string):
+                A list of the mesh dimensions of the attached mesh, if any.
+
+        ..note::
+            One of the mesh dimensions will generally also appear in the cube
```
bjlittle, line 1735:
```suggestion
            One of the mesh dimensions will typically also appear in the cube
```

lib/iris/fileformats/netcdf.py
```diff
@@ -1593,57 +1725,158 @@ def _get_dim_names(self, cube):
...
+            # Add new coords (mesh or dim) to the already-seen list.
+            for coord in matching_coords:
+                if coord not in self._dim_names_and_coords.coords:
+                    self._dim_names_and_coords.append(dim_name, coord)
+
+            # Add the latest name to the returned list
+            dimension_names.append(dim_name)
+
+        # Get info on mesh, first.
+        dimension_names.clear()
```
bjlittle, line 1765: @pp-mo This is a NOP, right? If so, then we can drop it

lib/iris/fileformats/netcdf.py
pp-mo: :+1:

lib/iris/fileformats/netcdf.py
pp-mo: :+1:
Definitely does nothing

lib/iris/fileformats/netcdf.py
```diff
@@ -1593,57 +1725,158 @@ def _get_dim_names(self, cube):
...
-                    dimension_names.append(dim_name)
+                    # For face/edge, use the non-optional connectivity variable.
+                    cf_role = f"{location}_node_connectivity"
+                    dim_coords = mesh.connectivities(cf_role=cf_role)
+                if len(dim_coords) > 0:
+                    # As the mesh contains this location, we want to include this
+                    # dim in our returned mesh dims.
+                    # We should have 1 identifying variable (of either type).
+                    assert len(dim_coords) == 1
+                    mesh_coord = dim_coords[0]
```
pp-mo, line 1793: Potentially a rather confusing name for it -- definitely not a MeshCoord.
Consider changing this ?

lib/iris/fileformats/netcdf.py
```diff
@@ -1593,57 +1725,158 @@ def _get_dim_names(self, cube):
...
-            in the cube.
+            cube_dimensions, mesh_dimensions
+            * cube_dimensions (list of string):
+                A lists of dimension names for each dimension of the cube
+            * mesh_dimensions (list of string):
+                A list of the mesh dimensions of the attached mesh, if any.
+
+        ..note::
+            One of the mesh dimensions will typically also appear in the cube
+            dimensions.
```
pp-mo, line 1736: As noted in discussion, may be worth noting that : the ordering of the return values is the intended order in the file.
And for consistency with that, this should probably return `(mesh_dimensions, cube_dimensions)` instead of the other way around.

lib/iris/fileformats/netcdf.py
```diff
@@ -1593,57 +1725,158 @@ def _get_dim_names(self, cube):
...
+                                or dim_name in self._name_coord_map.names
+                            ):
+                                dim_name = self._increment_name(dim_name)
+
+            # Record the dimension.
+            record_dimension(dim_name, cube.shape[dim], dim_coords)
+
+        cube_dimensions = dimension_names.copy()
+
+        return cube_dimensions, mesh_dimensions
```
bjlittle, line 1879: Minor point, but this could just be:
```python
return dimension_names, mesh_dimensions
```
there's no need for the `copy` or `cube_dimensions`... but if you think it's clearer as is, I'm okay with that too

lib/iris/fileformats/netcdf.py
```diff
@@ -1593,57 +1725,158 @@ def _get_dim_names(self, cube):
...
-                    # Increment name if conflicted with one already existing.
-                    if self._existing_dim[dim_name] != cube.shape[dim]:
+                # Get a name from the dim-coord (if any).
+                dim_coords = cube.coords(dimensions=dim, dim_coords=True)
+                if dim_coords:
+                    # Derive a dim name from a coord.
+                    coord = dim_coords[0]  # always have at least one
+
+                    # Add only dimensions that have not already been added.
+                    if coord in self._dim_names_and_coords.coords:
```
bjlittle, line 1846: @pp-mo We could refine this behaviour (as discussed) here by the `coord` and `name` methods on the `CFNameCoordMap` returning `None` rather than raising a `KeyError`.

That change is all self contained within this function (I think)... I'll leave it for your to consider

pp-mo: :+1:
as discussed offline, quite possible + looks like it gives a simpler approach
TBD

lib/iris/fileformats/netcdf.py
```diff
@@ -1593,57 +1725,158 @@ def _get_dim_names(self, cube):
...
-                else:
-                    # Update dictionary with new entry
-                    self._existing_dim[dim_name] = cube.shape[dim]
 
-                dimension_names.append(dim_name)
-        return dimension_names
+                else:
+                    # No CF-netCDF coordinates describe this data dimension.
+                    # Make up a new, distinct dimension name
+                    dim_name = "dim%d" % dim
```
bjlittle, line 1862:
```suggestion
                    dim_name = f"dim{dim}"
```

lib/iris/fileformats/netcdf.py
```diff
@@ -1593,57 +1725,158 @@ def _get_dim_names(self, cube):
...
+            If the dimension has been seen already, check that it's length
+            matches the earlier finding.
+
+            """
+            if dim_name not in self._existing_dim:
+                self._existing_dim[dim_name] = length
+            else:
+                # Just make sure we never re-write one.
+                # TODO: possibly merits a proper Exception with message
+                assert self._existing_dim[dim_name] == length
```
bjlittle, line 1754: @pp-mo You could perhaps adopt the `assert expression, msg` here

lib/iris/fileformats/netcdf.py
```diff
@@ -1593,57 +1725,158 @@ def _get_dim_names(self, cube):
...
+
+            # Add new coords (mesh or dim) to the already-seen list.
+            for coord in matching_coords:
+                if coord not in self._dim_names_and_coords.coords:
```
bjlittle, line 1758: @pp-mo Minor - You could cache `self._dim_names_and_coords` outside the loop

lib/iris/fileformats/netcdf.py
```diff
@@ -1593,57 +1725,158 @@ def _get_dim_names(self, cube):
...
+        # Get info on mesh, first.
+        dimension_names.clear()
+        mesh = cube.mesh
+        if mesh is None:
+            cube_mesh_dim = None
+        else:
+            # Identify all the mesh dimensions.
+            # NOTE: one of these will be a cube dimension, but that one does not
+            # get any special handling.  We *do* want to list/create them in a
+            # definite order (face,edge,node), and before non-mesh dimensions.
```
bjlittle, line 1773: @pp-mo Here you document that the `definite order` is `face,edge,node`, and yet `MESH_LOCATIONS = ("node", "edge", "face")`

lib/iris/fileformats/netcdf.py
```diff
@@ -1593,57 +1725,158 @@ def _get_dim_names(self, cube):
...
+                if location == "node":
+                    # For nodes, identify the dim with a coordinate variable.
+                    # Selecting the X-axis one for definiteness.
+                    dim_coords = mesh.coords(include_nodes=True, axis="x")
                 else:
-                    # Return the dim_name associated with the existing
-                    # coordinate.
-                    dim_name = self._name_coord_map.name(coord)
-                    dimension_names.append(dim_name)
+                    # For face/edge, use the non-optional connectivity variable.
```
bjlittle, line 1785: @pp-mo `face_node_connectivity` is a required attribute, but `edge_node_connectivity` is *optionally* required... but I'm splitting hairs here.

lib/iris/fileformats/netcdf.py
```diff
@@ -1593,57 +1725,158 @@ def _get_dim_names(self, cube):
...
+                    cf_role = f"{location}_node_connectivity"
+                    dim_coords = mesh.connectivities(cf_role=cf_role)
+                if len(dim_coords) > 0:
+                    # As the mesh contains this location, we want to include this
+                    # dim in our returned mesh dims.
+                    # We should have 1 identifying variable (of either type).
+                    assert len(dim_coords) == 1
+                    mesh_coord = dim_coords[0]
+                    match = mesh_coord in self._dim_names_and_coords.coords
```
bjlittle, line 1794: @pp-mo You may consider refactoring here, given this [comment](https://github.com/SciTools/iris/pull/4318/files#r714082409)

lib/iris/fileformats/netcdf.py
```diff
@@ -1593,57 +1725,158 @@ def _get_dim_names(self, cube):
...
+                    if match:
+                        # For mesh-identifying coords, we require the *same*
+                        # coord, not an identical one (i.e. "is" not "==")
+                        name = self._dim_names_and_coords.name(mesh_coord)
+                        stored_coord = self._dim_names_and_coords.coord(name)
+                        match = mesh_coord is stored_coord
```
bjlittle, line 1800: @pp-mo Can you give an example where you would find an *identical* coord and not the *same* one?

lib/iris/fileformats/netcdf.py
```diff
@@ -1593,57 +1725,158 @@ def _get_dim_names(self, cube):
...
+                    if match:
+                        # Use the previous name for this dim of this mesh, from
+                        # a previously saved cube with the same mesh.
+                        dim_name = self._dim_names_and_coords.name(mesh_coord)
```
bjlittle, line 1804: @pp-mo Surely this is simply `dim_name = name`, right?

lib/iris/fileformats/netcdf.py
```diff
@@ -1593,57 +1725,158 @@ def _get_dim_names(self, cube):
...
+                        while dim_name in self._existing_dim:
+                            dim_name = self._increment_name(dim_name)
+
+                        record_dimension(dim_name, dim_length, dim_coords)
 
+                    # Store the mesh dims indexed by location
+                    mesh_location_dimnames[location] = dim_name
+
+            # Identify the cube dimension which maps to the mesh.
+            (cube_mesh_dim,) = cube.coords(mesh_coords=True)[0].cube_dims(cube)
```
bjlittle, line 1825: :scream:

lib/iris/fileformats/netcdf.py
```diff
@@ -1593,57 +1725,158 @@ def _get_dim_names(self, cube):
...
+                        if location == "node":
+                            # always 1-d
+                            (dim_length,) = mesh_coord.shape
+                        else:
+                            # extract source dim, respecting dim-ordering
+                            dim_length = mesh_coord.shape[mesh_coord.src_dim]
+                        location_dim_attr = f"{location}_dimension"
+                        dim_name = getattr(mesh, location_dim_attr)
+                        if dim_name is None:
+                            dim_name = f"{mesh.name()}_{location}"
```
bjlittle, line 1815: @pp-mo The use of `mesh.name()` could result in an invalid `dim_name`, right? e.g., it could be based on a space filled `long_name`

lib/iris/fileformats/netcdf.py
```diff
@@ -1416,11 +1544,15 @@ def _add_aux_coords(self, cube, cf_var_cube, dimension_names):
             Names associated with the dimensions of the cube.
 
         """
+        # Exclude any mesh coords, which are bundled in with the aux-coords.
+        aux_coords_no_mesh = [
+            coord for coord in cube.aux_coords if not hasattr(coord, "mesh")
+        ]
```
bjlittle, line 1550: Nice one :+1:

lib/iris/fileformats/netcdf.py
```diff
@@ -1818,29 +2051,149 @@ def _get_coord_variable_name(self, cube, coord):
...
+            if not cf_name:
+                # Auto-generate a name based on mesh properties.
+                cf_name = f"Mesh_{mesh.topology_dimension}d"
+
+        # Ensure valid form for var-name.
+        cf_name = self.cf_valid_var_name(cf_name)
+        return cf_name
 
-    def _create_generic_cf_array_var(self, cube, cube_dim_names, element):
+    def _create_mesh(self, cube, mesh):
```
bjlittle, line 2116: @pp-mo The `cube` argument is not used within `_create_mesh`

lib/iris/fileformats/netcdf.py
```diff
@@ -1818,29 +2051,149 @@ def _get_coord_variable_name(self, cube, coord):
...
+            required.
+
+        Returns:
+            A CF-netCDF variable name as a string.
+
+        """
+        cf_name = mesh.var_name
+        if mesh.var_name is None:
+            # Prefer var-name, but accept long/standard as aliases.
+            cf_name = mesh.var_name or mesh.standard_name or mesh.long_name
```
bjlittle, line 2107: @pp-mo `mesh.var_name` is already `None` here... and personally I'd probably stray away from using the `standard_name` and `long_name`.

Why not just simply go for `Mesh_{...}d` here?

lib/iris/fileformats/netcdf.py
```diff
@@ -1818,29 +2051,149 @@ def _get_coord_variable_name(self, cube, coord):
...
+        )
+
+        # Add the basic essential attributes
+        _setncattr(cf_mesh_var, "cf_role", "mesh_topology")
+        _setncattr(
+            cf_mesh_var,
+            "topology_dimension",
+            np.int32(mesh.topology_dimension),
+        )
+        # Add 'standard' names + units atributes
```
bjlittle, line 2151:
```suggestion
        # Add 'standard' names + units attributes
```

lib/iris/fileformats/netcdf.py
```diff
@@ -1356,6 +1381,109 @@ def _create_cf_dimensions(
...
+            * cf_mesh_name (string or None):
+            The name of the mesh variable created, or None if the cube does not
+            have a mesh.
+
+        """
+        cf_mesh_name = None
+        mesh = cube.mesh
+        if mesh:
+            if mesh in self._name_coord_map.coords:
+                cf_mesh_name = self._name_coord_map.name(mesh)
```
bjlittle, line 1409: @pp-mo I just need some clarity here...

If the `mesh` is in the `self._name_coord_map`, then it must have already been created (`_create_mesh`) for it to be in the map... therefore surely we don't want to do anything further, right?

lib/iris/fileformats/netcdf.py
```diff
@@ -1356,6 +1381,109 @@ def _create_cf_dimensions(
...
+                cf_mesh_name = self._create_mesh(cube, mesh)
+                self._name_coord_map.append(cf_mesh_name, mesh)
+
+            cf_mesh_var = self._dataset.variables[cf_mesh_name]
+
+            # Get the mesh-element dim names.
+            mesh_dims = self._mesh_dims[mesh]
+
+            # Add all the element coordinate variables.
+            for location in MESH_LOCATIONS:
```
bjlittle, line 1420: @pp-mo Could you use `mesh.all_coords` instead? Then take advantage of the returned `namedtuple` API?

lib/iris/fileformats/netcdf.py
```diff
@@ -1356,6 +1381,109 @@ def _create_cf_dimensions(
...
+                            element_dims=(mesh_dims[location],),
+                        )
+                        coord_names.append(coord_name)
+                    # Record the coordinates (if any) on the mesh variable.
+                    if coord_names:
+                        coord_names = " ".join(coord_names)
+                        _setncattr(cf_mesh_var, coords_file_attr, coord_names)
+
+            # Add all the connectivity variables.
+            # pre-fetch the set + ignore "None"s -- looks like a bug ?
```
bjlittle, line 1442: @pp-mo It's not a bug, it's returning a `namedtuple`, which will have populated slots, or a `None`.

lib/iris/fileformats/netcdf.py
```diff
@@ -1818,29 +2051,149 @@ def _get_coord_variable_name(self, cube, coord):
...
-                # Handle scalar coordinate (dims == ()).
-                if not name:
-                    name = "unknown_scalar"
+                # We need to invent a name, based on its associated dimensions.
+                if cube.coords(coord):
+                    # It is a regular cube coordinate.
+                    # Auto-generate a name based on the dims.
+                    name = ""
+                    for dim in cube.coord_dims(coord):
+                        name += "dim{}".format(dim)
```
bjlittle, line 2060:
```suggestion
                        name += f"dim{dim}"
```

lib/iris/fileformats/netcdf.py
```diff
@@ -1818,29 +2051,149 @@ def _get_coord_variable_name(self, cube, coord):
...
+                    from iris.experimental.ugrid import Connectivity
+
+                    if isinstance(coord, Connectivity):
+                        location = coord.cf_role.split("_")[0]
+                    else:
+                        # Must be a mesh-element coordinate.
+                        location = None
+                        for test_location in MESH_LOCATIONS:
+                            include_key = f"include_{test_location}s"
+                            if coord in cube.mesh.coords({include_key: True}):
```
bjlittle, line 2077: `**{include_key: True}` ?

bjlittle (changes requested):
@pp-mo Over to you. It's looking good, considering what you inherited :wink: 

A few review comments to service, but nothing major. I've still to sweep through the tests, so I'll do that in the meantime.

lib/iris/fileformats/netcdf.py
pp-mo: :+1:
Since our talk yesterday, I now have a plan for simplifying this

lib/iris/fileformats/netcdf.py
pp-mo: :+1:

lib/iris/fileformats/netcdf.py
pp-mo: :+1: Ok will address.  I think an explicit 'raise' is probably as good as anything.
"assert" in production code is such bad style !!

lib/iris/fileformats/netcdf.py
pp-mo: :+1: oops

lib/iris/fileformats/netcdf.py
pp-mo: :+1:

lib/iris/fileformats/netcdf.py
pp-mo: :+1: I only just re-discovered the "optionally required" term they use.  It is a useful distinction + I will adopt it here.

lib/iris/fileformats/netcdf.py
pp-mo: :+1:

lib/iris/fileformats/netcdf.py
pp-mo: Ok I could sugar this one a bit ...

lib/iris/fileformats/netcdf.py
pp-mo: Well, certainly if we are saving cubes loaded from similar input files, but not the same one.
I'm not sure this is the place to explain that though.  What do you think?

lib/iris/fileformats/netcdf.py
pp-mo: :+1: good spot
Should have noticed PyCharm telling me this.

lib/iris/fileformats/netcdf.py
pp-mo: > mesh.var_name is already None here

Whoops, yes !

> personally I'd probably stray away from using the standard_name and long_name... Why not just simply go for Mesh_{...}d 

I think it's entirely logical that a Mesh can never really have a standard_name, or units for that matter.
I will fix this.
The alternative mechanism seems sensible in itself, **_but_** I was trying to be consistent with usage elsewhere.  E.G. coords, cell measures etc will get a varname from the long-name if there is one.  So I think I would prefer to keep it doing that on "least surprise" grounds.

lib/iris/fileformats/netcdf.py
pp-mo: Well, mostly I want to clearly control the order I'm doing them in.
I also still have doubts about the hard-wiring of terms like 'node_x', 'node_y' in the Mesh APIs :  I wish we could be more UGRID-generic.
Which this way --i.e. iterating over the coords of a specific location-- does deliver OK

lib/iris/fileformats/netcdf.py
pp-mo: :+1: will amend

lib/iris/fileformats/netcdf.py
pp-mo: :+1:

lib/iris/fileformats/netcdf.py
bjlittle: That might be better. I certainly had to think about it a couple of times when I first came across this... so perhaps your suggestion would be more natural. Plus it reinforces the intended ordering relationship within the file :+1:

lib/iris/fileformats/netcdf.py
pp-mo: As in e.g. `mesh.coords(include_nodes=True)`

I do wish that Mesh had an API giving access to these things programmatically, without needing to construct a keyword name from a location string.
So in this case, I think we should be able to say e.g. `mesh.coords(location='edge')`
And likewise in the case of connectivities, it would need something like `mesh.connectivities(from='face', to='node', axis='x')`
(but where  keys are variables, not code literals which we can already do)
Whereas I am currently using [keyword construction for that too](https://github.com/SciTools/iris/blob/813bfce4c1c1e4c763f9f9ffab9dac286ceccd77/lib/iris/fileformats/netcdf.py#L1786-L1787).

lib/iris/fileformats/netcdf.py
pp-mo: Yes, it's just like the regular dim-coords that `self._name_coord_map` was originally used for.
The logic is :  if you've seen it before, it was already created : so lookup the name that it was already given

lib/iris/fileformats/netcdf.py
pp-mo: :+1:

lib/iris/fileformats/netcdf.py
pp-mo: This is tricky.
Essentially you're right, and I have added a testcase including an unsuitable long name.
Unfortunately, at least at present this code can *never* runs, because a Mesh always invents a dimension name if it didn't have one.
So for now, I've fixed this, added an assert that it never happens, _and_ a comment to explain it (!phew!)

pp-mo: > a Mesh always invents a dimension name if it doesn't have one

so the line that calls `mesh.name()` can never actually run.
I have now added an assert + and explanation for this.

lib/iris/fileformats/netcdf.py
pp-mo: Fixed in the "new world", where we always fetch a match + then see if it is None.

pp-mo:
Ok, that was a useful stage to review the code + my changes.
Multiple tweaks.  
@bjlittle I hope this addressed everything raised so far (but I pushed back on a couple) : **please re-review**

lib/iris/tests/unit/fileformats/netcdf/test_Saver__ugrid.py
```diff
@@ -932,44 +932,52 @@ def _namestext(names):
...
                 ("air_temperature", None, None),
-                ("air_temperature", None, "air_temperature"),
+                ("air_temperature", None, "Mesh_2d"),
             ),
             # long_name only
             (
                 (None, "my_long_name", None),
                 (None, "my_long_name", "my_long_name"),
             ),
+            # long_name that needs "fixing"
```
pp-mo, line 960: Whoops.  I think we are now testing this twice.

lib/iris/_representation/cube_summary.py
bjlittle: Yeah, I think we're in a better position to enrich the cube summary. Going back to before Xmas, I was keen to include minimal mesh information in the cube summary... and this is a good (self) use case to push that forwards (seems like)

lib/iris/fileformats/netcdf.py
bjlittle: No worries, I just asked because I was in that space and was curious... I can appreciate that it's more of a discussion point.

All good, nothing to see here... happily moving along :wink:

lib/iris/fileformats/netcdf.py
bjlittle: @pp-mo You still need to change this, right? i.e., you can't just pass a dictionary here

lib/iris/tests/unit/fileformats/netcdf/test_Saver__ugrid.py
pp-mo: (removed now : https://github.com/SciTools/iris/pull/4318/commits/9ae1ea39e58dd7ed2eced2e9a58cef78822e9670)

lib/iris/fileformats/netcdf.py
pp-mo: Ok I see what it was.
`test_location_coord_names` was supposed to cover this, but we can't actually have mesh location coords with no standard_name, so it is unreachable.

Removed this code + added some notes in case we relax this limitation later.

bjlittle: Cool, thanks :+1:

lib/iris/tests/stock/netcdf.py
```diff
@@ -72,11 +74,13 @@ def _add_standard_data(nc_path, unlimited_dim_size=0):
             unlimited_dim_size if dim == unlimited_dim_names[0] else size
             for dim, size in zip(dims, shape)
         ]
-        data = np.zeros(shape, dtype=var.dtype)
+        data = np.ones(shape, dtype=var.dtype)  # Do not use zero
```
bjlittle, line 77: @pp-mo `Do not use zero`... why not?

Add a smidge of context to help the reader? :+1:

lib/iris/tests/unit/fileformats/netcdf/test_Saver__ugrid.py
```diff
@@ -0,0 +1,1171 @@
...
+
+    def apply_xyargs(coords, xyargs):
+        if xyargs:
+            for coord, kwargs in zip(coords, xyargs):
+                # coords and xyargs both iterables : implicitly=(x,y)
+                applyargs(coord, kwargs)
+
+    node_coords = [
+        AuxCoord(np.arange(n_nodes), standard_name=name)
+        for loc, name in zip(XY_LOCS, XY_NAMES)
```
bjlittle, line 74: @pp-mo `loc` (`XY_LOCS`) isn't used

lib/iris/tests/unit/fileformats/netcdf/test_Saver__ugrid.py
```diff
@@ -0,0 +1,1171 @@
...
+
+    """
+    ds = nc.Dataset(filepath)
+    # dims dict is {name: len}
+    dimsdict = {name: dim.size for name, dim in ds.dimensions.items()}
+    # vars dict is {name: {attr:val}}
+    varsdict = {}
+    for name, var in ds.variables.items():
+        varsdict[name] = {prop: getattr(var, prop) for prop in var.ncattrs()}
+        varsdict[name]["_DIMS"] = list(var.dimensions)
```
bjlittle, line 260: @pp-mo Perhaps consider making `_DIMS` a constant?

lib/iris/tests/unit/fileformats/netcdf/test_Saver__ugrid.py
```diff
@@ -0,0 +1,1171 @@
...
+        tempfile_path = self.temp_dir / Path(tempfile_path).name
+
+        # Save data to the file.
+        save(data, tempfile_path)
+
+        return tempfile_path
+
+    def test_basic_mesh(self):
+        # Save a small mesh example and check aspects of the resulting file.
+        data = make_cube()  # A simple face-mapped data example.
```
bjlittle, line 368: `data` -> `cube` ? (and below, if you agree)

lib/iris/tests/unit/fileformats/netcdf/test_Saver__ugrid.py
```diff
@@ -0,0 +1,1171 @@
...
+        If None, use 'default_mesh()'
+    * location (string):
+        Which mesh element to map the cube to.
+    * kwargs (dict):
+        Additional property settings to apply to the cube (after creation).
+
+    """
+    if mesh is None:
+        mesh = default_mesh()
+    dim = mesh_location_size(mesh, location)
```
bjlittle, line 216: This isn't a `dim`, but a `size` or dimension depth/extent

lib/iris/tests/unit/fileformats/netcdf/test_Saver__ugrid.py
```diff
@@ -0,0 +1,1171 @@
...
+    cube = Cube(np.zeros(dim, np.float32))
+    for meshco in mesh.to_MeshCoords(location):
+        cube.add_aux_coord(meshco, (0,))
+    for key, val in kwargs.items():
+        setattr(cube, key, val)
+    return cube
+
+
+def add_height_dim(cube):
+    """Add an extra inital 'height' dimension onto a cube."""
```
bjlittle, line 226:
```suggestion
    """Add an extra initial 'height' dimension onto a cube."""
```

lib/iris/tests/unit/fileformats/netcdf/test_Saver__ugrid.py
```diff
@@ -0,0 +1,1171 @@
...
+
+        # data maps to the height + mesh dims
+        self.assertEqual(data_props["_DIMS"], ["height", face_dim])
+        self.assertEqual(data_props["mesh"], mesh_name)
+        self.assertEqual(data_props["location"], "face")
+
+    def test_nonmesh_hybrid_dim(self):
+        # Check a case with a hybrid non-mesh dimension
+        cube = realistic_4d()
+        # Strip off the time and longtude dims, to make it simpler.
```
bjlittle, line 577:
```suggestion
        # Strip off the time and longitude dims, to make it simpler.
```

lib/iris/tests/unit/fileformats/netcdf/test_Saver__ugrid.py
```diff
@@ -0,0 +1,1171 @@
...
+        cube = add_height_dim(cube)
+
+        # Save and snapshot the result
+        tempfile_path = self.check_save_cubes(cube)
+        dims, vars = scan_dataset(tempfile_path)
+
+        # have just 1 mesh, including a face and node coordinates.
+        (mesh_name,) = vars_meshnames(vars)
+        face_dim = vars_meshdim(vars, "face", mesh_name)
+        _ = vars_meshdim(vars, "node", mesh_name)
```
bjlittle, line 564: @pp-mo Is it worthwhile checking that `face_dim == vars_meshdim(vars, "node", mesh_name)` ?

lib/iris/tests/unit/fileformats/netcdf/test_Saver__ugrid.py
```diff
@@ -0,0 +1,1171 @@
...
+
+        # have just 1 mesh, including face and node coordinates.
+        (mesh_name,) = vars_meshnames(vars)
+        face_dim = vars_meshdim(vars, "face", mesh_name)
+        _ = vars_meshdim(vars, "node", mesh_name)
+
+        # have hybrid vertical dimension, with all the usual term variables.
+        self.assertIn("model_level_number", dims)
+        vert_vars = list(vars_w_dims(vars, ["model_level_number"]).keys())
+        # The list of file variables mapping the vertical dimensio:
```
bjlittle, line 612:
```suggestion
        # The list of file variables mapping the vertical dimension:
```

lib/iris/tests/unit/fileformats/netcdf/test_Saver__ugrid.py
```diff
@@ -0,0 +1,1171 @@
...
+        # 2. non- (plain) lonlat are NOT converted
+        # 3. other names remain as whatever was given
+        # 4. no units on input --> none on output
+        self.assertEqual(node_x["units"], "degrees")
+        self.assertEqual(node_y["units"], "ms-1")
+        self.assertNotIn("units", face_x)
+        self.assertEqual(face_y["units"], "degrees_north")
+
+        # N.B. we also do not automatically add 'axis' properties, nor recognise
+        # them on input :  Because we *only* do that for dim-coords.
```
bjlittle, line 921: @pp-mo Just to clarify... If I recall (could be mistaken) but we don't ever store the `axis`, we only infer it, right?

So this comment about dim coords is a little misleading... or am I confused / missing something?

lib/iris/tests/unit/fileformats/netcdf/test_Saver__ugrid.py
```diff
@@ -0,0 +1,1171 @@
...
+            )
+            self.assertEqual(expected_names, result_names, fail_msg)
+
+    def test_location_coord_names(self):
+        # Check the selection of mesh-element coordinate names.
+        # Check the selection of mesh-variables names.
+        # N.B. this is basically centralised in Saver._get_mesh_variable_name,
+        # but we test in an implementation-neutral way (as it's fairly easy).
+
+        # Options heres are limited because the Mesh relies on guess_axis so,
```
bjlittle, line 1013:
```suggestion
        # Options here are limited because the Mesh relies on guess_axis so,
```

lib/iris/tests/stock/netcdf.py
pp-mo: :+1:

lib/iris/tests/unit/fileformats/netcdf/test_Saver__ugrid.py
pp-mo: :+1: nice spot.
I'm surprised flake8 didn't find this though.

lib/iris/tests/unit/fileformats/netcdf/test_Saver__ugrid.py
pp-mo: :+1:

bjlittle (approved):

bjlittle:
@pp-mo Truly super stuff, thanks :tada:

trexfeathers:
I have 105 email notifications about this - you two must have worked really hard! 🏆🏆

PR #4319, opened 2021-09-15
Skip TestConstrainedLoad if data missing

wjbenfold:
### Description

Added the skip data decorator to TestConstrainedLoad in the integration tests for NetCDF

Running the tests without access to the test data was causing test failures when they should have been skipped (specifically from the integration tests for NetCDF, in the TestConstrainedLoad class). This is no longer the case.

---

### DIFF
--- a/docs/src/whatsnew/latest.rst
+++ b/docs/src/whatsnew/latest.rst
@@ -104,6 +104,9 @@ This document explains the changes made to Iris for this release
    http://www.nationalarchives.gov.uk/doc/open-government-licence since this
    always works locally, but never within CI. (:pull:`4307`)
 
+#. `@wjbenfold`_ netCDF integration tests now skip TestConstrainedLoad if
+   test data is missing (:pull:`4319`)
+
 
 .. comment
     Whatsnew author names (@github name) in alphabetical order. Note that,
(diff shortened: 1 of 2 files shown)


docs/src/whatsnew/latest.rst
```diff
@@ -104,6 +104,9 @@ This document explains the changes made to Iris for this release
    http://www.nationalarchives.gov.uk/doc/open-government-licence since this
    always works locally, but never within CI. (:pull:`4307`)
 
+#. `@wjbenfold`_ netCDF integration tests now skip TestConstrainedLoad if
```
bjlittle, line 107: @wjbenfold We can't use `sphinx` to link to test classes, as they're not publicly documented as per the `iris` API, but perhaps put back ticks around `TestConstrainedLoad` to highlight it as a class?

bjlittle (changes requested):
@wjbenfold Lovely catch 👀 

A minor suggestion to service, then we're good to go 👍

docs/src/whatsnew/latest.rst
bjlittle:
```suggestion
#. `@wjbenfold`_ netCDF integration tests now skip ``TestConstrainedLoad`` if
```

bjlittle (approved):

Issue #4091, opened 2021-04-09
Implement OR for Iris ConstraintCombinations


jamesp:
## ✨ Feature Request
Implement the OR operator for Constraints

## Motivation
Constraints can be combined to give a `ConstraintCombination` object.  Only the `__and__` operation has been implemented (i.e. `c1 & c2`), it would be useful to create a combination of `c1 | c2`.

rcomer:
[Interesting](https://stackoverflow.com/questions/45412581/combine-iris-constraints-with-or), though I've long since forgotten what I wanted to do with them!

jamesp:
Interesting discussion, thanks for the link @rcomer.  I can see the argument both ways on this one.

PR #4321, opened 2021-09-16
Update to loading docs to cover absence of 'or' for constraints

wjbenfold:
### Description

Added documentation to Iris Constrained Loading docs to highlight that Constraints don't support ¦ (and explain why)

Whilst addressing a recent support request it seemed that there should be an ability to logically "or" Constraints as well as to "and" them. The reasons for this weren't immediately obvious, so documenting it seems useful for the future. I've done so.

I'd note, I'm not 100% convinced that this is an improvement to the quality of the docs, as I can see the argument that in catering for an edge case of misunderstanding the docs may become less easily understood by those without this misunderstanding (particularly as loading is likely to be one of the first bits of documentation a new user reads). Happy to discuss the merits of this and/or axe this PR on that basis.

---

### DIFF
--- a/docs/src/userguide/loading_iris_cubes.rst
+++ b/docs/src/userguide/loading_iris_cubes.rst
@@ -217,11 +217,15 @@ constraint to ``load``::
 
 .. note::
 
-    Whilst ``&`` is supported, the ``|`` that might reasonably be expected
-    is not. This is because "or" constraints could lead to a cube which did
-    not cover a hyper-rectangular region if these constraints were on
-    different coordinates. If an "or" type constraint on a single coordinate
-    is useful then it may be obtained by passing a function to :class:`iris.Constraint`.
+    Whilst ``&`` is supported, the ``|`` that might reasonably be expected is
+    not. Explanation as to why is in the :class:`iris.Constraint` reference
+    documentation.
+
+    For an example of constraining to multiple ranges of the same coordinate to
+    generate one cube, see the :class:`iris.Constraint` reference documentation.
+
+    To generate multiple cubes, each constrained to a different range of the
+    same coordinate, use :py:func:`iris.load_cubes`.
 
 As well as being able to combine constraints using ``&``,
 the :class:`iris.Constraint` class can accept multiple arguments,
--- a/docs/src/userguide/loading_iris_cubes.rst
+++ b/docs/src/userguide/loading_iris_cubes.rst
@@ -215,6 +215,14 @@ constraint to ``load``::
     level_10 = iris.Constraint(model_level_number=10)
     cubes = iris.load(filename, forecast_6 & level_10)
 
+.. note::
+
+    Whilst ``&`` is supported, the ``|`` that might reasonably be expected
+    is not. This is because "or" constraints could lead to a cube which did
+    not cover a hyper-rectangular region if these constraints were on
+    different coordinates. If an "or" type constraint on a single coordinate
+    is useful then it may be obtained by passing a function to :class:`iris.Constraint`.
+
 As well as being able to combine constraints using ``&``,
 the :class:`iris.Constraint` class can accept multiple arguments,
 and a list of values can be given to constrain a coordinate to one of
--- a/lib/iris/_constraints.py
+++ b/lib/iris/_constraints.py
@@ -83,7 +83,8 @@ class Constraint:
             ambiguous what should be extracted.
 
             To generate multiple cubes, each constrained to a different range of
-            the same coordinate, use :py:func:`iris.load_cubes`.
+            the same coordinate, use :py:func:`iris.load_cubes` or
+            :py:func:`iris.cube.CubeList.extract_cubes`.
 
             A constraint covering multiple ranges of the same coordinate may be
             used to generate a single cube like so::
--- a/lib/iris/_constraints.py
+++ b/lib/iris/_constraints.py
@@ -87,7 +87,7 @@ class Constraint:
             :py:func:`iris.cube.CubeList.extract_cubes`.
 
             A constraint covering multiple ranges of the same coordinate may be
-            used to generate a single cube like so::
+            created like so::
 
                 def latitude_bands(cell):
                     return (0 < cell < 30) or (60 < cell < 90)
--- a/lib/iris/_constraints.py
+++ b/lib/iris/_constraints.py
@@ -73,6 +73,26 @@ class Constraint:
                        model_level_number=[10, 12])
                        & Constraint(ensemble_member=2)
 
+        .. note::
+            Whilst ``&`` is supported, the ``|`` that might reasonably be expected
+            is not. This is because each constraint describes a boxlike region, and
+            thus the intersection of these constraints (obtained with ``&``) will
+            also describe a boxlike region. Allowing the union of two constraints
+            (with the ``|`` symbol) would allow the description of a non-boxlike
+            region. These are difficult to describe with cubes and so it would be
+            ambiguous what should be extracted.
+
+            To generate multiple cubes, each constrained to a different range of
+            the same coordinate, use :py:func:`iris.load_cubes`.
+
+            A constraint covering multiple ranges of the same coordinate may be
+            used to generate a single cube like so::
+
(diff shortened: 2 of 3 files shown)


jamesp:
Nice addition @wjbenfold.  This should probably close #4091 as not going to be implemented.

docs/src/userguide/loading_iris_cubes.rst
```diff
@@ -215,6 +215,14 @@ constraint to ``load``::
...
     cubes = iris.load(filename, forecast_6 & level_10)
 
+.. note::
+
+    Whilst ``&`` is supported, the ``|`` that might reasonably be expected
+    is not. This is because "or" constraints could lead to a cube which did
+    not cover a hyper-rectangular region if these constraints were on
+    different coordinates. If an "or" type constraint on a single coordinate
+    is useful then it may be obtained by passing a function to :class:`iris.Constraint`.
+
```
pp-mo, line 225: I think it might also be worth explaining that you can select for for A and B constraints separately, using iris.load_cubes and CubeList.extract_cubes.
E.G. `my_cubes = iris.load_cubes(sourcefile, [cons_sel_A. cons_sel_B, cons_sel_C])`
or, (often) maybe : `cube_A, cube_B, cube_C = iris.load_cubes(sourcefile, [cons_sel_A. cons_sel_B, cons_sel_C])`

So, this is really _not_ an 'OR', but it is similar as long as your separate selections each make a single cube.

Is this any use ??

docs/src/userguide/loading_iris_cubes.rst
```diff
@@ -215,6 +215,14 @@ constraint to ``load``::
     level_10 = iris.Constraint(model_level_number=10)
     cubes = iris.load(filename, forecast_6 & level_10)
 
+.. note::
+
+    Whilst ``&`` is supported, the ``|`` that might reasonably be expected
+    is not. This is because "or" constraints could lead to a cube which did
```
stephenworsley, line 221: It might make sense to go into the details of why "|"/"or" is not supported in the docstring of the Constraint class and refer to that documentation here, thus saving space in the user guide and giving you freedom to expand the explanation of Constaint combination.

docs/src/userguide/loading_iris_cubes.rst
```diff
@@ -215,6 +215,14 @@ constraint to ``load``::
...
+    not cover a hyper-rectangular region if these constraints were on
```
stephenworsley, line 222: "hyper-rectangular" is perhaps a tad technical, consider just "rectangular" or "boxlike".

docs/src/userguide/loading_iris_cubes.rst
wjbenfold: Yes, that's helpful as another place people might want to end up from the initial query that starts them off here. I'll build a reference to that into the final resolution.

docs/src/userguide/loading_iris_cubes.rst
```diff
@@ -215,6 +215,14 @@ constraint to ``load``::
...
+    different coordinates. If an "or" type constraint on a single coordinate
```
stephenworsley, line 223: This alternative approach is worth keeping in the user guide. An example here would be quite helpful.

pp-mo:
> @jamesp This should probably close #4091 as not going to be implemented.

Hint: worth trying -- you can add it to "Linked issues" on the rhs ...

stephenworsley (changes requested):
Looks good, just a couple suggestions.

docs/src/userguide/loading_iris_cubes.rst
wjbenfold: Rather than adding an example specifically to cover this, I included a link to the load_cubes method with a description of the use case, which I think is the concise but helpful way to go.

docs/src/userguide/loading_iris_cubes.rst
wjbenfold: Have done this

docs/src/userguide/loading_iris_cubes.rst
wjbenfold: Done, good suggestion, thanks!

docs/src/userguide/loading_iris_cubes.rst
wjbenfold: Included an example in the Constraint docs

lib/iris/_constraints.py
```diff
@@ -73,6 +73,26 @@ def __init__(self, name=None, cube_func=None, coord_values=None, **kwargs):
...
+            Whilst ``&`` is supported, the ``|`` that might reasonably be expected
+            is not. This is because each constraint describes a boxlike region, and
+            thus the intersection of these constraints (obtained with ``&``) will
+            also describe a boxlike region. Allowing the union of two constraints
+            (with the ``|`` symbol) would allow the description of a non-boxlike
+            region. These are difficult to describe with cubes and so it would be
+            ambiguous what should be extracted.
+
+            To generate multiple cubes, each constrained to a different range of
+            the same coordinate, use :py:func:`iris.load_cubes`.
```
pp-mo, line 86:
```suggestion
            the same coordinate, use :py:func:`iris.load_cubes` or :meth:`iris.cube.CubeList.merge_cubes`.
```

Showing that constraints are used in the same way for extracting as for loading.

wjbenfold: I can't see a method called "iris.cube.CubeList.merge_cubes" in the docs, am I missing something? Did you mean "extract_cubes"?

pp-mo: Sorry, yes, I did mean 'extract_cubes' .
And I was _so_ careful to check the import path, too ...

wjbenfold: Committed this addition

lib/iris/_constraints.py
```diff
@@ -73,6 +73,27 @@ def __init__(self, name=None, cube_func=None, coord_values=None, **kwargs):
...
+            (with the ``|`` symbol) would allow the description of a non-boxlike
+            region. These are difficult to describe with cubes and so it would be
+            ambiguous what should be extracted.
+
+            To generate multiple cubes, each constrained to a different range of
+            the same coordinate, use :py:func:`iris.load_cubes` or
+            :py:func:`iris.cube.CubeList.extract_cubes`.
+
+            A constraint covering multiple ranges of the same coordinate may be
+            used to generate a single cube like so::
```
stephenworsley, line 90: This agrees with the existing pattern for including example code in docstrings found here for example: https://github.com/SciTools/iris/blob/06de63873dcd3108cf3a0b848c0c110e8cc3a7f9/lib/iris/cube.py#L682-L692

lib/iris/_constraints.py
```diff
@@ -73,6 +73,27 @@ def __init__(self, name=None, cube_func=None, coord_values=None, **kwargs):
...
+
+                def latitude_bands(cell):
+                    return (0 < cell < 30) or (60 < cell < 90)
+
+                Constraint(cube_func=latitude_bands)
```
stephenworsley, line 95: Just a small quibble, this example does not go as far as generating a cube as the above comment would suggest, rather it describes a constraint which is capable of generating a single cube. Perhaps something more like this would be more accurate:
"A cube can be constrained to multiple ranges within the same coordinate using something like the following constraint::"

stephenworsley (changes requested):
Looks good, just a tiny comment and I think this is good to go.

lib/iris/_constraints.py
wjbenfold: Is this just confirmation, or an action?

lib/iris/_constraints.py
wjbenfold: This is now done

wjbenfold:
Rebased onto master

lib/iris/_constraints.py
stephenworsley: Confirmation. I was curious about how this fit into our existing patterns so I decided to check and leave the comment for the record.

stephenworsley (approved):
Looks good to me.

PR #4320, opened 2021-09-16
refactor 3.0 whatsnew

bjlittle:
### Description
This PR offers an alternative to #4285 the the `v3.1.x` feature release branch.

For context see #4316 and this [comment](https://github.com/SciTools/iris/pull/4316#issuecomment-920408310).

If this approach is adopted, then I'll subsequently update the `latest.rst.template` in `main`.

The rendered `whatsnew` documentation changes for this PR are available [here](https://bjlittle-iris.readthedocs.io/en/latest/whatsnew/3.0.html).

---
[Consult Iris pull request check list]( https://scitools-iris.readthedocs.io/en/latest/developers_guide/contributing_pull_request_checklist.html)

PR #4323, opened 2021-09-16
update latest.rst.template

bjlittle:
### Description
This PR plays forward the minor pattern change in the `whatsnew` to support patches, as adopted on the `v3.1.x` release feature branch.

For context, see #4320.

---

### DIFF
--- a/docs/src/whatsnew/latest.rst.template
+++ b/docs/src/whatsnew/latest.rst.template
@@ -7,7 +7,7 @@ This document explains the changes made to Iris for this release
 (:doc:`View all changes <index>`.)
 
 
-.. dropdown:: :opticon:`report` v3.X.X Release Highlights
+.. dropdown:: :opticon:`report` |iris_version| Release Highlights
    :container: + shadow
    :title: text-primary text-center font-weight-bold
    :body: bg-light
--- a/docs/src/whatsnew/latest.rst.template
+++ b/docs/src/whatsnew/latest.rst.template
@@ -39,6 +39,11 @@ v3.X.X (DD MMM YYYY)
 
    #. N/A
 
+NOTE: section above is a template for bugfix patches
+====================================================
+    (Please remove this section when creating an initial 'latest.rst')
+    
+
 
 📢 Announcements
 ================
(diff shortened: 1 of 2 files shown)


docs/src/whatsnew/latest.rst.template
```diff
@@ -7,31 +7,33 @@ This document explains the changes made to Iris for this release
...
+   The highlights for this major/minor release of Iris include:
+
+   * N/A
+
+   And finally, get in touch with us on `GitHub`_ if you have any issues or
+   feature requests for improving Iris. Enjoy!
 
-   #. N/A
 
+v3.X.X (DD MMM YYYY)
```
pp-mo, line 25:
```suggestion
NOTE: section below is a template for bugfix patches
====================================================
    (Please remove this section when creating an initial 'latest.rst')

v3.X.X (DD MMM YYYY)
```
Comparing this with the current `latest.rst`, I see that we have no initial use for this section when we are creating a new `latest.rst`.
So, this section is actually a further reference to be used _later_ -- it's' a sort of template within the template.

So, _**can we use a header like this to point this out ?**_

docs/src/whatsnew/latest.rst.template
```diff
@@ -7,31 +7,33 @@ This document explains the changes made to Iris for this release
...
-   :open:
 
-   The highlights for this major/minor release of Iris include:
-
-   * N/A
+   The patches in this release of Iris include:
 
-   And finally, get in touch with us on `GitHub`_ if you have any issues or
-   feature requests for improving Iris. Enjoy!
+   #. N/A
```
pp-mo, line 36:
```suggestion
   #. N/A

NOTE: section above is a template for bugfix patches
====================================================
    (Please remove this section when creating an initial 'latest.rst')
    
```

matching the previous suggestion

pp-mo (approved):
Looks nicer
 -- but I made a suggestion about the 'bugfix' section content.
 
 Also, I see that our current `latest.rst` isn't quite in line with this now
  \-- just because, it doesn't have the version number in the "Release Highlights" box heading.  
So, could we fix that in this PR also ?

bjlittle:
> So, **_could we fix that in this PR also ?_**

Yup, great idea... I'll push that now in this PR, thanks.

pp-mo:
Nice, useful step forward !

PR #4324, opened 2021-09-17
fix trove classifier

bjlittle:
### Description
This PR fixes the malformed [PyPI trove classifer](https://pypi.org/classifiers/), which is preventing the `iris` sdist and binary wheel being uploaded to PyPI due to stricter metadata checks.

The `iris` v3.1.0 release date has also been updated. We can re-tag and re-release afresh after this PR is merged.

I'll purge the existing tag on `16 Sept 2021`, which is safe to do, as we've not packaged the release yet (so no need to bump versions or perform a `post` release).

---

### DIFF
--- a/docs/src/whatsnew/3.1.rst
+++ b/docs/src/whatsnew/3.1.rst
@@ -1,6 +1,6 @@
 .. include:: ../common_links.inc
 
-v3.1 (16 Sep 2021)
+v3.1 (17 Sep 2021)
 ******************
 
 This document explains the changes made to Iris for this release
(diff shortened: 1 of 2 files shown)


pp-mo:
:+1: will merge when checks done

bjlittle:
@pp-mo I've now deleted the `v3.1.0` tag, so the release goes automatically into `draft`.

All you need to do is re-tag and then re-publish and we should be good to re-spin the packaging :+1:

bjlittle:
@pp-mo I also checked and verified the other PyPI trove classifiers with the [trove-classifiers](https://pypi.org/project/trove-classifiers/) package, and they're all fine :+1:

PR #4326, opened 2021-09-17
Added mesh_from_coords (AVD-1877)

trexfeathers:
### Description

Adds the ability to construct a `Mesh` from two `Coords` in `iris.experimental.ugrid`.

**I expect this to fail on doctests until `iris-sample-data` `v2.4.0` is available for use by Miniconda.** I was able to manually demonstrate this would work on my local machine by replacing the lockfile URL with the [front-end URL](https://anaconda.org/conda-forge/iris-sample-data/2.4.0/download/noarch/iris-sample-data-2.4.0-pyhd8ed1ab_0.tar.bz2) instead. Once `v2.4.0` is available the doctests should be re-spun (check using `conda search iris-esmf-regrid -c conda-forge`).

---

### DIFF
--- a/lib/iris/experimental/ugrid/__init__.py
+++ b/lib/iris/experimental/ugrid/__init__.py
@@ -2818,11 +2818,7 @@ def mesh_from_coords(coord_1, coord_2):
 
         >>> print(cube_w_mesh.mesh.name())
         Topology data of 2D unstructured mesh
-        >>> mesh_coord_names = [
-        ...     coord.name()
-        ...     for coord in cube_w_mesh.coords()
-        ...     if isinstance(coord, MeshCoord)
-        ... ]
+        >>> mesh_coord_names = cube_w_mesh.coords(mesh_coords=True)
         >>> print(f"MeshCoords: {mesh_coord_names}")
         MeshCoords: ['latitude', 'longitude']
 
--- a/lib/iris/experimental/ugrid/__init__.py
+++ b/lib/iris/experimental/ugrid/__init__.py
@@ -37,7 +37,7 @@ from ...common.metadata import (
 )
 from ...common.mixin import CFVariableMixin
 from ...config import get_logger
-from ...coords import AuxCoord, _DimensionalMetadata
+from ...coords import AuxCoord, Coord, _DimensionalMetadata
 from ...exceptions import ConnectivityNotFoundError, CoordinateNotFoundError
 from ...fileformats import cf, netcdf
 from ...fileformats._nc_load_rules.helpers import get_attr_units, get_names
@@ -2792,25 +2792,16 @@ def mesh_from_coords(coord_1, coord_2):
 
     .. testsetup::
 
-        from iris import load_cube
+        from iris import load_cube, sample_data_path
         from iris.experimental.ugrid import (
             PARSE_UGRID_ON_LOAD,
             MeshCoord,
             mesh_from_coords,
         )
 
-        # TODO: Replace with a file in iris-sample-data, as soon as we can
-        #  generate something adequately small and representative.
-        from iris.tests import get_data_path
-        file_path = get_data_path(
-            [
-                "NetCDF",
-                "unstructured_grid",
-                "lfric_ngvat_2D_1t_face_half_levels_main_conv_rain.nc",
-            ]
-        )
+        file_path = sample_data_path("mesh_C4_synthetic_float.nc")
         with PARSE_UGRID_ON_LOAD.context():
-            cube_w_mesh = load_cube(file_path, "conv_rain")
+            cube_w_mesh = load_cube(file_path)
 
     For example::
 
@@ -2818,7 +2809,9 @@ def mesh_from_coords(coord_1, coord_2):
 
         >>> print(cube_w_mesh.mesh.name())
         Topology data of 2D unstructured mesh
-        >>> mesh_coord_names = cube_w_mesh.coords(mesh_coords=True)
+        >>> mesh_coord_names = [
+        ...     coord.name() for coord in cube_w_mesh.coords(mesh_coords=True)
+        ... ]
         >>> print(f"MeshCoords: {mesh_coord_names}")
         MeshCoords: ['latitude', 'longitude']
 
@@ -2839,8 +2832,8 @@ def mesh_from_coords(coord_1, coord_2):
 
         # Replace the AuxCoords with MeshCoords.
         >>> for ix in range(2):
-        ...     cube_sub.remove(orig_mesh_coords[ix])
-        ...     cube_sub.add_coord(new_mesh_coords[ix], cube_w_mesh.mesh_dim())
+        ...     cube_sub.remove_coord(orig_mesh_coords[ix])
+        ...     cube_sub.add_aux_coord(new_mesh_coords[ix], cube_w_mesh.mesh_dim())
 
         >>> print(cube_sub.mesh.name())
         Topology data of 2D unstructured mesh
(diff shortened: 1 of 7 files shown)


lib/iris/experimental/ugrid/__init__.py
```diff
@@ -2746,6 +2747,239 @@ def face_node(self):
...
+        Topology data of 2D unstructured mesh
+        >>> for coord_name in mesh_coord_names:
+        ...     coord = cube_sub.coord(coord_name)
+        ...     print(f"{coord_name}: {type(coord).__name__}")
+        latitude: MeshCoord
+        longitude: MeshCoord
+
+    """
+    # Validate coord essentials.
+    coord_names = {coord_1: "coord_1", coord_2: "coord_2"}
```
stephenworsley, line 2848: It seems like bad practice to use a mutable object as a key in a dictionary. It might make more sense to reverse the order of key/item here

lib/iris/experimental/ugrid/__init__.py
```diff
@@ -2746,6 +2747,239 @@ def face_node(self):
...
+        invalid_bounds = False
+        sub_message = ""
+        if not coord.has_bounds():
+            invalid_bounds = True
+            sub_message = "None"
+        else:
+            bounds_shape = coord.core_bounds().shape
+            # 1D coord check guarantees bounds are ndim==2.
+            if bounds_shape[1] < 2:
+                invalid_bounds = 2
```
stephenworsley, line 2867: Shouldn't this be `invalid_bounds = True`?

lib/iris/experimental/ugrid/__init__.py
```diff
@@ -2746,6 +2747,239 @@ def face_node(self):
...
+                sub_message = f"shape {bounds_shape}. Must be (n, >=2)."
+        if invalid_bounds:
+            message = f"{name} has invalid bounds: " + sub_message
+            raise ValueError(message)
+
+    # Validate orthogonality.
+    coord_axes = {
+        coord: guess_coord_axis(coord) for coord in (coord_1, coord_2)
```
stephenworsley, line 2875: As suggested above, this may make more sense as a list or tuple.

lib/iris/experimental/ugrid/__init__.py
```diff
@@ -2746,6 +2747,239 @@ def face_node(self):
...
+        topology_dimension = 2
+        coord_centring = "face"
+        conn_cf_role = "face_node_connectivity"
+
+    # Create connectivity.
+    if coord_1.has_lazy_bounds():
+        array_lib = da
+    else:
+        array_lib = np
+    indices = array_lib.arange(np.prod(bounds_shape)).reshape(bounds_shape)
```
stephenworsley, line 2921: When the bounds array is masked this seems like it would cause the existence of nodes without location coordinates. I can't recall if we're directly supporting mixed size cells, but if we are, it's probably worth at least raising an appropriate error in this case.

lib/iris/experimental/ugrid/__init__.py
pp-mo: Well, Coord is one of those classes [specifically adjusted](https://github.com/SciTools/iris/blob/v3.1.0/lib/iris/coords.py#L1610-L1617) so that we _can_ use them as dictionary keys and sets.

Frankly, the more I see of this, the less it worries me -- it's already a kluge in Python itself that generic objects can be used in this way.
And I think it basically just makes `dict[obj] = value` a shorthand for `dict[id(obj)] = value`.

lib/iris/experimental/ugrid/__init__.py
pp-mo: After some thought, I think we **_can_** support this quite happily, and without much further difficulty.

The key point is that building the indices like this is treating every corner of every face as an independent node
(to avoid the cost of searching the bounds arrays for duplicate values, to work out where an original node is re-visited)

In that case, I think there are just 2 problems

(1) in this connectivity array, we need to replace a 'normal' index value with a masked point,  wherever the bounds values are masked. 
I think that can just be something like...
`indices = array_lib.ma.masked_array(indices, array_lib.ma.getmaskarray(coord_1.core_bounds()))`
(i.e. apply the bounds mask to the connectivity array, but lazily if need be)

(2) in the node arrays, we will have some values which derived from masked points in the bounds array.  But we don't support masked data in these, so we should at least fill those points with something safe.
Which again is pretty easy, just use an `array_lib.ma.filled(coord.core_bounds(), 0.0)`

I think it helps to consider what a simple mixed-shape mesh looks like in the original, and then what happens when you make meshcoords for it,  and then re-construct a mesh from the meshcoord bounds.
Here is my teeny example :
```
Suppose we have a square + a triangle joined like this ...
 1--2
 |  |\
 |  | \
 3--4--5

this translates as the following
    node_x = [x1, x2, x3, x4, x5]
    face_node_connectivity = [[1, 2, 3, 4], [2, 5, 4, --]]
    face_x = [fx1, fx2]
    
(NOTE : skipping all Y values -- they behave just like the X ones )

So, if you now create MeshCoords from this, you get (X only) ...
    mesh_x.points = [fx1, fx2]
    mesh_x.bounds = [[x1, x2, x3, x4], [x2, x5, x4, --]]

If we now reconstruct a new mesh from this, it leads to the following... 
USING CAPS to denote the NEW mesh data, 
    FACE_X = [fx1, fx2]
    NODE_X = [X1=x1, X2=x2, X3=x3, X4=x4, X5=x2, X6=x5, X7=x4, ***X8=0.0*** ]  # 0.0 at "masked node"
    FACE_NODE_CONNECTIVITY = [[1, 2, 3, 4], [5, 6, 7, ***--*** ]]   # "--"==missing for unused slot
```
N.B. in this, the thing we *really* need to fix is the masked point in the connectivity array : we can't use a real index pointing to a 'masked node', as we need the missing points encoded in the connectivity itself.

lib/iris/experimental/ugrid/__init__.py
trexfeathers: >It seems like bad practice to use a mutable object as a key in a dictionary.

Pretty sure `Coord`s are immutable. Indeed that's why I was comfortable with this, but it sounds like that's irrelevant anyway!

lib/iris/experimental/ugrid/__init__.py
pp-mo: P.S. examples like the above we can calculate for real : 
We can use an indexing something like `cube = cube[..., 0:]` to remove the mesh + replace the mesh-coords with auxcoords.

lib/iris/experimental/ugrid/__init__.py
trexfeathers: https://github.com/SciTools/iris/pull/4326/files#r715711539

lib/iris/experimental/ugrid/__init__.py
trexfeathers: >P.S. examples like the above we can calculate for real : 
>We can use an indexing something like `cube = cube[..., 0:]` to remove the mesh + replace the mesh-coords with auxcoords.

Yes, I did something like that in the doctest:

https://github.com/SciTools/iris/blob/764e44f8883cef55ef1109ee1041700fd7c8b1c5/lib/iris/experimental/ugrid/__init__.py#L2819-L2821

lib/iris/experimental/ugrid/__init__.py
trexfeathers: e8cf2e1ef1d3b3ac25c487f6c475142c9df01b4b

lib/iris/experimental/ugrid/__init__.py
trexfeathers: 13f98539462a612dda5236c0eccdbb0bb8de02a1

lib/iris/experimental/ugrid/__init__.py
```diff
@@ -2746,6 +2747,239 @@ def face_node(self):
...
+        # Neither X nor Y guessed, use in original order.
+        coord_x = coord_1
+        coord_y = coord_2
+    else:
+        # Find X/Y if guessed, otherwise use the single un-guessed coord.
+        coord_x = axis_coords.get("X") or axis_coords["other"]
+        coord_y = axis_coords.get("Y") or axis_coords["other"]
+
+    node_x = AuxCoord(
+        points=coord_x.core_bounds().flatten(),
```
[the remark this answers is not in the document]
trexfeathers: 13f98539462a612dda5236c0eccdbb0bb8de02a1

lib/iris/tests/unit/experimental/ugrid/test_mesh_from_coords.py
```diff
@@ -233,6 +233,28 @@ def test_centre_metadata(self):
...
+
+        lat_bounds = np.array([[0, 1, 1, 0], [1, 2, 2, 1], [2, 2, 3, 999]])
+        lat_bounds = np.ma.masked_equal(lat_bounds, 999)
+        lat = AuxCoord(
+            points=[0.5, 1.5, 2.5], bounds=lat_bounds, standard_name="latitude"
+        )
+
+        mesh = mesh_from_coords(lon, lat)
+        self.assertArrayEqual(
+            mesh.face_node_connectivity.src_lengths(), [4, 4, 3]
```
pp-mo, line 253: Nice !!

lib/iris/experimental/ugrid/__init__.py
```diff
@@ -2748,6 +2749,241 @@ def face_node(self):
         return self._members["face_node_connectivity"]
 
 
+def mesh_from_coords(coord_1, coord_2):
```
pp-mo, line 2752: I think we might consider presenting this as a class method `Mesh.from_coords`, similar to `Coord.from_coord` or `DimCoord.from_regular`.
Does that seem slightly neater ?

pp-mo (changes requested):
Some slightly random issues to consider, 
also including my attempt at a summary of our long discussion earlier.

lib/iris/tests/unit/experimental/ugrid/test_mesh_from_coords.py
```diff
@@ -0,0 +1,288 @@
...
+            "long_name",
+            "var_name",
+        ):
+            self.assertIsNone(getattr(mesh, attr))
+        self.assertTrue(mesh.units.is_unknown())
+        self.assertDictEqual({}, mesh.attributes)
+
+    def test_lazy(self):
+        self.lon = AuxCoord(
+            points=self.lon.lazy_points(),
```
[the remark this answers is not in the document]
trexfeathers: Not possible in this case since one of the coords is a `DimCoord` (to prove that `DimCoord`s also work fine).

lib/iris/experimental/ugrid/__init__.py
```diff
@@ -2748,6 +2749,241 @@ def face_node(self):
...
+        >>> print(cube_sub.mesh.name())
+        Topology data of 2D unstructured mesh
+        >>> for coord_name in mesh_coord_names:
+        ...     coord = cube_sub.coord(coord_name)
+        ...     print(f"{coord_name}: {type(coord).__name__}")
+        latitude: MeshCoord
+        longitude: MeshCoord
+
+    """
+    # Validate coord essentials.
```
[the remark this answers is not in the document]
trexfeathers: eb868ea06be373c188c492e3339c1385ecc08f29

lib/iris/tests/unit/experimental/ugrid/test_mesh_from_coords.py
```diff
@@ -0,0 +1,288 @@
...
+        self.lat = replace_standard_name(self.lon, None)
+        with self.assertRaisesRegex(
+            ValueError, ".*Both coords' metadata identical."
+        ):
+            _ = self.create()
+
+
+class Test2Dim(Test1Dim):
+    def setUp(self):
+        self.lon = DimCoord(
```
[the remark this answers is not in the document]
trexfeathers: ccc410af84b0960bd1949c8cae19cfecf0360254

lib/iris/tests/unit/experimental/ugrid/test_mesh_from_coords.py
```diff
@@ -0,0 +1,288 @@
...
+
+        self.lat = lat_orig.copy(
+            points=lat_orig.points[-1], bounds=lat_orig.bounds[-1]
+        )
+        with self.assertRaisesRegex(ValueError, ".*points shape must be"):
+            _ = self.create()
+
+    def test_coords_not_orthogonal(self):
+        def replace_standard_name(coord, standard_name):
+            return coord.__class__(
```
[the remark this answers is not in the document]
trexfeathers: ccc410af84b0960bd1949c8cae19cfecf0360254

lib/iris/tests/unit/experimental/ugrid/test_mesh_from_coords.py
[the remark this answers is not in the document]
trexfeathers: ccc410af84b0960bd1949c8cae19cfecf0360254

PR #4330, opened 2021-09-20
Suppress warning in animation test

rcomer:
### Description

Occasionally we get test failures because `assertWarns` finds a warning about animation deletion when it is expecting something else, in tests that have nothing to do with animation.  Running this single test module prior to this change shows that the warning is thrown here.  I thought it might be worth suppressing it, but obviously I have no way to check whether it will cure the problem of interfering with other tests, as it only happens sporadically.

Worth a try?

---

### DIFF
--- a/docs/src/whatsnew/latest.rst
+++ b/docs/src/whatsnew/latest.rst
@@ -124,6 +124,9 @@ This document explains the changes made to Iris for this release
    Useful for development purposes only.  For more information see
    :ref:`contributing.documentation.building` the documentation. (:pull:`4333`)
 
+#. `@rcomer`_ modified the ``animation`` test to prevent it throwing a warning
+   that sometimes interferes with unrelated tests. (:pull:`4330`)
+
 .. comment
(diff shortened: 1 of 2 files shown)


rcomer:
Another option might be to modify the `_draw_was_started` attribute before the test method returns.

https://github.com/matplotlib/matplotlib/blob/2b15797272f53215b65b37424d34535ff43777fd/lib/matplotlib/animation.py#L887-L895

bjlittle:
> Another option might be to modify the `_draw_was_started` attribute before the test method returns.
> 
> [](https://github.com/matplotlib/matplotlib/blob/2b15797272f53215b65b37424d34535ff43777fd/lib/matplotlib/animation.py#L887-L895)

@rcomer Did you want to give that a try instead?

rcomer:
@bjlittle yeah, I was leaning towards that too.  How's this?  I can confirm that the warning is not thrown when I just run the one test module.

bjlittle (approved):
@rcomer LGTM. Awesome thanks :beers:

PR #4339, opened 2021-09-24
Ugrid save mesh

pp-mo:
### Description

---


lib/iris/fileformats/netcdf.py
```diff
@@ -1766,9 +1784,20 @@ def record_dimension(names_list, dim_name, length, matching_coords=[]):
             # Add the latest name to the list passed in.
             names_list.append(dim_name)
 
+        # Choose cube or mesh saving.
+        from iris.cube import Cube
+
+        if isinstance(cube_or_mesh, Cube):
```
pp-mo, line 1790: I have added this (or something very similar) in several routines now.
It would be possible, and maybe less boilerplate, to pass cube + mesh separately instead, but this way does preserve some essential logic of the invocation :
  1. we _either_ have a mesh with no cube, a cube with no mesh, or a cube with a mesh.
  2. (we never have no cube _and_ no mesh)
  2. if we have _both_ mesh _and_ cube, then `mesh == cube.mesh`

bjlittle (changes requested):
@pp-mo This is lovely... slots in quite neatly to easily extend the capability.

Minor suggestion on the `save_mesh` API location to service, so consider that, otherwise we're good to go :beers:

bjlittle:
@pp-mo Could you also add a `whatsnew` entry for this PR, and also #4318 ?

lib/iris/fileformats/netcdf.py
```diff
@@ -3119,3 +3181,51 @@ def is_valid_packspec(p):
 
         # Add conventions attribute.
         sman.update_global_attributes(Conventions=conventions)
+
+
+def save_mesh(
```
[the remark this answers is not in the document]
pp-mo: :+1: yes that makes good sense.

lib/iris/fileformats/netcdf.py
```diff
@@ -3119,3 +3181,51 @@ def is_valid_packspec(p):
...
+    mesh_or_meshes,
```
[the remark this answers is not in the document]
pp-mo: :+1:

pp-mo:
>  a whatsnew entry for this PR, and also #4318 

Whoops yes, absolutely.
I'll add them both in this PR, I guess that is good enough.

lib/iris/experimental/ugrid/__init__.py
```diff
@@ -45,11 +45,11 @@
 from ...util import guess_coord_axis
 
 __all__ = [
-    "CFUGridReader",
```
pp-mo, line 48: NOTE: I don't think this belongs or is needed in public API, so I removed it.
It is imported (in deferred usage) within netcdf.py, but I don't think that warrants including here :  once the two are re-integrated, that would become a local reference.

pp-mo:
Thanks @bjlittle !

Now addressed above comments (I hope).
Please re-consider.. :pray:

docs/src/whatsnew/latest.rst
```diff
@@ -49,9 +49,21 @@ This document explains the changes made to Iris for this release
...
    returned with a :class:`~iris.experimental.ugrid.Mesh` attached.
+   (:pull:`4058`).
+
+#. `@pp-mo`_ added support to save cubes with meshes to netcdf files, using the
+   `UGRID`_ conventions.
+   The existing :meth:`iris.save` function now does this, when saving cubes with meshes.
+   A routine :meth:`iris.experimental.ugrid.save_mesh` allows saving
+   :class:`iris.experimental.ugrid.Mesh` objects to netcdf _without_ any associated data
+   (i.e. not attached to cubes).
+   (:pull:`4318` and :pull:`4339`).
```
bjlittle, line 66: @pp-mo This renders like this...
image
Is that what you intended? Or were you aiming for it to be in italics? ... if so, instead use `*without*`, or `**without**` for bold

lib/iris/experimental/ugrid/__init__.py
```diff
@@ -45,11 +45,11 @@
...
     "Connectivity",
     "ConnectivityMetadata",
     "load_mesh",
     "load_meshes",
+    "save_mesh",
```
bjlittle, line 52: @pp-mo I think this should come after `PARSE_UGRID_ON_LOAD`, but this is due to be refactored imminently, so not bothered either way

bjlittle (approved):

Issue #4336, opened 2021-09-22
Run tests on Windows


jamesp:
## ✨ Feature Request

We provide iris 3 for Windows via conda-forge - see https://anaconda.org/conda-forge/iris. There are known users of iris on Windows. We should setup CI to run the test suite on Windows.

trexfeathers:
This could definitely be good thing, but the way people have talked before: it also sounds like a significant extra commitment. This is presumably a combination of extra potential problems arising at any given time due to a 'wider CI net', plus the extra expertise we may need to keep it working.

tkknight:
We can at least try to get the CI working, if it highlights many issues then it is more of a known issue as right now it is an unknown unknown.  Suggest any effort to this will managed and prioritised, unless we have help from contributors outside the Met Office.

wjbenfold:
I've started making changes to .cirrus.yml that run the tests against Windows to try this

wjbenfold:
A confluence page exists to document the progress that has been made on understanding this

PR #4340, opened 2021-09-27
Initial addition of windows tests to ci

wjbenfold:
### Description

Adds Windows tests to CI (see issue #4336)

Changes made to .cirrus.yml to include extra tests for python 3.7 and 3.8 on Windows.

---

### DIFF
--- a/.cirrus.yml
+++ b/.cirrus.yml
@@ -280,7 +280,7 @@ task:
     - conda create -n xxhash-test --file requirements\ci\nox.lock\py37-win-64.lock
     - conda init cmd.exe
     - conda activate xxhash-test
-    - python -c "import sys; print(sys.path)"
+    - python -c "import sys; sys.path.append('c:\\users\\containeradministrator\\appdata\\local\\temp\\cirrus-ci-build\\lib'); print(sys.path)"
     - python -c "import xxhash"
     - conda activate base
     - nox --session tests -- --verbose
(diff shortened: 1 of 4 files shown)


wjbenfold:
Would mark as WIP but I believe that prevents the CI that I'm trying to test, so not doing so

tkknight:
Suggest to disable/comment out the non windows test to speed up for CI testing.  Also see if you can run the windows tests locally first if possible.

bjlittle:
@wjbenfold Marking a PR as `draft` shouldn't stop CI. It hasn't for me in the past :thinking:

trexfeathers:
I don't know if this is true or not, but if Cirrus is turning out to be a big blocker I'd like to draw your attention to #4193.

jamesp:
@wjbenfold I'm hugely impressed with the way you've approached this (ever-expanding) problem, clearly an iceberg! 

Looking at the test failures there is a bunch we can already learn about how we both write iris and how we test it. 
e.g. int32 vs int64. Should these errors be fixed by making iris enforce loading data in the same way (a bunch of infrastructure code to support multiple os that is entirely orthogonal to the problem domain) or should our tests be more tolerant of differences in defaults in the underlying os?

I think it's worth a write-up of what you've discovered (a lot!). There may be quick wins in here the rest of the dev team could jump in and act on to help you over the next hurdle.  Great work.

wjbenfold:
A writeup now exists in confluence of the understanding gained

Issue #4342, opened 2021-09-28
Confusing error message when applying "concatenate_cube" to multi-variable cube list


tinyendian:
## 🐛 Bug Report
Method `concatenate_cube` in source file `cube.py` can produce a confusing error message when it is (mistakenly) used on a multi-variable cube list. It reports identical cube names, e.g., `Cube names differ: air_pressure != air_pressure` when the cube list contains multiple instances of `air_pressure` cubes (along with cubes for other variables, such as `air_temperature`), which will probably confuse users (it did confuse me :-)).

## How To Reproduce
Steps to reproduce the behaviour:

1. Prepare two data files with two variables, with time steps of one variable stored across both files and with some differences in their metadata (so that cubes cannot be automatically concatenated and the variable will thus appear as multiple cubes when loaded by Iris)
2. Load the data using `cubes = iris.load(['file1', 'file2'], ['var1', 'var2'])`, with `cubes` now containing, e.g., `[var1, var1, var2]`
3. Concatenate the cubes using `cubes.concatenate_cube()`
4. As expected, a `ConcatenateError` message appears, which may list the same variable names as different:
```
iris.exceptions.ConcatenateError: failed to concatenate into a single cube.
  Cube names differ: var1 != var1
```
The error message might (correctly) say `var != var2` if cube order in `cubes` is different.

## Expected behaviour
The error message should read
```
iris.exceptions.ConcatenateError: failed to concatenate into a single cube.
  Cube names differ: var1 != var2
```
instead, to make it clear that the user erroneously tried to apply `concatenate_cube()` on a cube list that contains multiple variables.

## Environment 
 - OS & Version: Ubuntu 18.04.4 LTS
 - Iris Version: 3.1.0

## Additional context

Looking at the code (line 558 in `cube.py`), it seems to me that
```
"Cube names differ: {} != {}".format(names[0], names[1])
```
should be replaced by
```
"Cube names differ: {} != {}".format(unique_names[0], unique_names[1])
```
as `unique_names` is guaranteed to contain different names at positions `0` and `1`.

trexfeathers:
This is a great idea @tinyendian, thanks for suggesting it! We believe the fix should be as simple as what you have suggested; do you fancy taking it on?

If not, this is an ideal opportunity for _anyone_ looking to get involved with Iris 🙂

tinyendian:
Thanks, @trexfeathers, I'll create a pull request (and I'll work my way through the contributor's guidelines 🙂).

PR #4345, opened 2021-09-30
Clearer error message when using concatenate_cube on multi-variable cubes list

tinyendian:
### Description
This pull request addresses issue #4342: cube method `concatenate_cube()` can produce confusing error messages, reporting that `var1 != var1` when a user (erroneously) attempts to concatenate a multi-variable cube list consisting of, e.g., multiple time steps of var1 and var2  with `cubes = [var1, var1, var2, ...]`.

The reason for this behaviour is that the error message in the method quotes the first two elements of list `names`, which can be identical in the above scenario. This pull request replaces `names` with list `unique_names` when generating the error message - the latter list is used as a criterion for `concatenate_cube()` to reject concatenation, and it is guaranteed to produce non-identical variable names.  This should give the user better feedback on why concatenation failed.

It should not be necessary to handle the cases of length-zero or length-1 list, since `len(unique_names)` should always be `>=2`, given that the method will fail earlier on an empty cube list, and it will accept length 1 for concatenation.

---

### DIFF
--- a/docs/src/whatsnew/latest.rst
+++ b/docs/src/whatsnew/latest.rst
@@ -146,6 +146,7 @@ This document explains the changes made to Iris for this release
     Whatsnew author names (@github name) in alphabetical order. Note that,
     core dev names are automatically included by the common_links.inc:
 
+.. _@tinyendian: https://github.com/tinyendian
 
 
 
--- a/docs/src/whatsnew/latest.rst
+++ b/docs/src/whatsnew/latest.rst
@@ -74,6 +74,10 @@ This document explains the changes made to Iris for this release
    one cell's bounds align with the requested maximum and negative minimum, fixing
    :issue:`4221`. (:pull:`4278`)
 
+#. `@tinyendian`_ fixed the error message produced by :meth:`~iris.cube.CubeList.concatenate_cube`
+   when a cube list contains cubes with different names, which will no longer report
+   "Cube names differ: var1 != var1" if var1 appears multiple times in the list
+   (:issue:`4342`, :pull:`4345`)
 
 💣 Incompatible Changes
 =======================
--- a/lib/iris/cube.py
+++ b/lib/iris/cube.py
@@ -555,7 +555,9 @@ class CubeList(list):
         else:
             msgs = []
             msgs.append(
-                "Cube names differ: {} != {}".format(unique_names[0], unique_names[1])
+                "Cube names differ: {} != {}".format(
+                    unique_names[0], unique_names[1]
+                )
             )
             raise iris.exceptions.ConcatenateError(msgs)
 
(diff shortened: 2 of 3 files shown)


tinyendian:
Note that I just signed the SciTools CLA.

rcomer:
Hi @tinyendian, thanks for your contribution!

We have some [unit tests that check concatenate throws the right error](https://github.com/SciTools/iris/blob/2dc22e4d940d3b81b1cc51b023b42a058f21fcc2/lib/iris/tests/unit/concatenate/test_concatenate.py#L105-L216) in various circumstances.  Would you be happy to add a test there to cover the case you've highlighted?

Also, so that we can advertise your contribution in the next release, could you add an entry under "Bugs Fixed" in the [whatsnew/latest.rst](https://github.com/SciTools/iris/blob/main/docs/src/whatsnew/latest.rst) file?  There is guidance about writing a "whatsnew" entry [here](https://scitools-iris.readthedocs.io/en/stable/developers_guide/documenting/whats_new_contributions.html#writing-a-contribution).

Do let us know if any of this isn't clear, or if you need help with anything.

rcomer:
Sorry, I just realised I pointed to the wrong set of tests in my last comment 🤦‍♀️.  The "names differ" error is raised within `CubeList.concatenate_cube()` itself, so a test to cover this case would need to be added to [this class](https://github.com/SciTools/iris/blob/2dc22e4d940d3b81b1cc51b023b42a058f21fcc2/lib/iris/tests/unit/cube/test_CubeList.py#L27-L55).

tinyendian:
Hi @rcomer, I added a unit test that replicates the multi-variable scenario mentioned in the issue and checks the error message, as well as a "whatsnew" entry in the list of bug fixes. I hope that both match the required styles.

The doctests and link check are failing with error message "Unknown target name: "@tinyendian", does my GitHub name need to be added somewhere?

trexfeathers:
>"Unknown target name: "@tinyendian", does my GitHub name need to be added somewhere?

Correct!

https://github.com/SciTools/iris/blob/a07e20dd88f0d8367c908cba986d413869cdc50f/docs/src/whatsnew/latest.rst#L145-L149

See here for a populated example from `v3.1`:

https://github.com/SciTools/iris/blob/b15386e49d7e9597f69adf870481d7cee579fa9e/docs/src/whatsnew/3.1.rst#L300-L305

rcomer (approved):
Thanks @tinyendian, the new test looks perfect to me.  It fails without the fix and passes with it.  Just need to add the GitHub link for your handle, as @trexfeathers shows above, and then I think this will be ready to merge.

tinyendian:
Thanks, @rcomer, working on it...

rcomer:
Awesome - thanks @tinyendian!

tinyendian:
Thanks for your help getting this over the line, @rcomer and @trexfeathers!

PR #4347, opened 2021-09-30
Refactor `iris.experimental.ugrid`

trexfeathers:
### Description

Splitting the module into submodules. Looking forward to the conflicts!

---


lib/iris/experimental/ugrid/__init__.py
```diff
@@ -8,4029 +8,29 @@
...
-from ...fileformats._nc_load_rules.helpers import get_attr_units, get_names
-from ...io import decode_uri, expand_filespecs
-from ...util import guess_coord_axis
 
 __all__ = [
     "Connectivity",
-    "ConnectivityMetadata",
     "load_mesh",
     "load_meshes",
     "save_mesh",
```
bjlittle, line 32: @trexfeathers You've inherited this... but I *think* we're sorting alphabetical (exclusive of case), so `save_mesh` should be last, right?

lib/iris/experimental/ugrid/__init__.py
```diff
@@ -8,4029 +8,29 @@
...
-from collections import namedtuple
-from collections.abc import Iterable
-from contextlib import contextmanager
-from functools import wraps
-from itertools import groupby
-import logging
-from pathlib import Path
-import re
-import threading
+logger = get_logger(__name__, fmt="[%(cls)s.%(funcName)s]")
```
bjlittle, line 20: @trexfeathers This isn't quite right... but I'm tempted to bank this as is, and we fix it in a separate PR.

bjlittle (changes requested):
@trexfeathers Thanks for taking this refactor on... it's been well overdue :+1: 

Just a minor comment to service about `__all__` sort order, and the logging we can fix in a follow-up PR; this PR is noisy enough.

Also, I think this is worthy of a `whatsnew` entry... fancy adding one? :wink:

lib/iris/experimental/ugrid/__init__.py
trexfeathers: c71b88751da2d0c73ebf482ddc4e9ef6127a1934

bjlittle (approved):

PR #4351, opened 2021-10-01
propagate ugrid logger and address `np.int` deprecations

bjlittle:
### Description
This PR is a follow-up from #4347 and configures the `iris.experimental.ugrid.logger` to be a root logger.

The descendant modules `iris.experimental.ugrid.cf`, `iris.experimental.ugrid.load` and `iris.experimental.ugrid.mesh` now all configure their loggers to propagating their messages upwards to the root logger for appropriate handling.

Raising the `level` of the `iris.experimental.ugrid.logger` root logger above `NOTSET` will filter descendant logger messages appropriately; typically, this should not be necessary. Currently, the root logger is configured to report all descendant logger messages.

Alter the descendant logger `level`, as normal, to control logging e.g.,
```python
from iris.experimental.ugrid.load import logger

logger.setLevel("DEBUG")
```
will enable `iris.experimental.ugrid.load.logger` debug messages to be reported by the root logger (after propagation) using its stream handler.

---

### DIFF
--- a/lib/iris/config.py
+++ b/lib/iris/config.py
@@ -37,7 +37,9 @@ import os.path
 import warnings
 
 
-def get_logger(name, datefmt=None, fmt=None, level=None, propagate=None):
+def get_logger(
+    name, datefmt=None, fmt=None, level=None, propagate=None, handler=True
+):
     """
     Create a :class:`logging.Logger` with a :class:`logging.StreamHandler`
     and custom :class:`logging.Formatter`.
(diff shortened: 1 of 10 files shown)


lib/iris/tests/unit/experimental/ugrid/cf/test_CFUGridAuxiliaryCoordinateVariable.py
```diff
@@ -16,16 +16,18 @@
...
+)
 from iris.tests.unit.experimental.ugrid.cf.test_CFUGridReader import (
     netcdf_ugrid_variable,
 )
 
 
 def named_variable(name):
     # Don't need to worry about dimensions or dtype for these tests.
-    return netcdf_ugrid_variable(name, "", np.int)
+    return netcdf_ugrid_variable(name, "", int)
```
bjlittle, line 30: The use of `np.int` as an alias for the builtin `int` was [deprecated](https://numpy.org/doc/stable/release/1.20.0-notes.html#using-the-aliases-of-builtin-types-like-np-int-is-deprecated) in `numpy` v1.20.0

lib/iris/config.py
```diff
@@ -91,16 +91,25 @@ def get_logger(name, datefmt=None, fmt=None, level=None, propagate=None):
     logger.setLevel(level)
     logger.propagate = propagate
 
-    # Create a formatter.
-    formatter = logging.Formatter(fmt=fmt, datefmt=datefmt)
+    # Create and add the handler to the logger, if required.
+    if handler:
```
trexfeathers, line 95: [The Python logging HOWTO](https://docs.python.org/3/howto/logging.html#loggers) provides an explanation for this change:

>Child loggers propagate messages up to the handlers associated with their ancestor loggers. Because of this, it is unnecessary to define and configure handlers for all the loggers an application uses. It is sufficient to configure handlers for a top-level logger and create child loggers as needed. (You can, however, turn off propagation by setting the _propagate_ attribute of a logger to `False`.)

`handler` therefore needs to be a configurable option so that it can be disabled whenever `propagate` is set to `True`.

trexfeathers (approved):
Looks good, thanks @bjlittle 🙂

PR #4353, opened 2021-10-02
pre-commit sort-all

bjlittle:
### Description
This PR enables the [sort-all](https://github.com/aio-libs/sort-all) `pre-commit` hook, which automatically sorts `__all__` lists alphabetically.

Recently I've seen a few PRs where authors had to move entries manually on request by a reviewer, which is a wee bit haphazard and tiresome. There have also been debates regarding sort order etc.

This will no longer be an issue, thanks to `pre-commit` automation and the opinionated `sort-all` hook. It also means that there's one less thing for the PR author/reviewer to think about :partying_face: 

Note that, `sort-all` can be manually installed and executed from the command line, if devs so wish. However, it's only available on [PyPI](https://pypi.org/project/sort-all/) atm

---

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -54,3 +54,9 @@ repos:
     -   id: blacken-docs
         types: [file, rst]
         additional_dependencies: [black==21.6b0]
+
+-   repo: https://github.com/aio-libs/sort-all
+    rev: v1.1.0
+    hooks:
+    -   id: sort-all
+        types: [file, python]
--- a/docs/src/whatsnew/latest.rst
+++ b/docs/src/whatsnew/latest.rst
@@ -148,6 +148,9 @@ This document explains the changes made to Iris for this release
 #. `@trexfeathers`_ refactored :mod:`iris.experimental.ugrid` into sub-modules.
    (:pull:`4347`).
 
+#. `@bjlittle`_ enabled the `sort-all`_ `pre-commit`_ hook to automatically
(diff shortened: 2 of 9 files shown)


trexfeathers (commented):
I'm not sure about this one. There's no PEP for it, so I can't see why it matters enough to add architecture to manage it.

If a module's `__all__` is long enough that it needs sorting before it is readable then the module should probably be sub-modules anyway.

Anyone else have an opinion?

trexfeathers:
[Even numpy doesn't do this](https://github.com/numpy/numpy/blob/79a8986b3caf55ffe5caf715b09faa055e38d342/numpy/core/numeric.py#L40-L57), despite some rather long `__all__`s.

bjlittle:
> There's no PEP for it

We don't need a PEP to justify this. There's also no PEP for `black` or `isort`, and yet we still use them, so I don't see that as a  reason not to adopt this.

The win is that there is one less thing to worry about as a PR reviewer, regardless of what `numpy` does or doesn't do.

tkknight:
A small bit of automation that looks useful.  The work has already been done so I see no reason not to bank this.

@bjlittle, I assume you have tested this already?  If so I am happy to approve and merge.

bjlittle:
@tkknight Seems to do what it says on the tin.

If it's troublesome, we simply remove it, just as easy as we've added it 😉

tkknight (approved):

PR #4358, opened 2021-10-07
Gitignore test results

trexfeathers:
### Description

Something a bit silly, but I get fed up with seeing these files in my Git unstaged list after every local test run!

---


bjlittle:
Nice one. But even better, let's just drop support for `dot` :rocket:

PR #4360, opened 2021-10-07
rationalise logging support

bjlittle:
## 🚀 Pull Request

This PR rationalises and consolidates the `logging` within `iris`.

The initial setup is a single top-level root logger i.e., `iris.logger`, to which sub-module child loggers propagate their logging.

The root logger has a convenience formatter, which will always append the calling function name, and the class, if supplied; reducing boiler plate logging patterns being used by callers.

The logging message format is unified, but customisable. We can extend customisation as we see fit moving forwards, when required.

---


jamesp:
Let's go through this at the next peloton, worthy of a proper discussion

Issue #2168, opened 2016-10-06
iris.tests.test_trajectory.TestTrajectory.test_tri_polar


bjlittle:
The "integration" test `iris.tests.test_trajectory.TestTrajectory.test_tri_polar` is chronically s--l--o--w.

It's interpolating a 90 point trajectory over the following cube ...

```
sea_water_potential_temperature / (degC) (time: 1; depth: 31; -- : 148; -- : 180)
     Dimension coordinates:
          time                                x         -        -         -
          depth                               -         x        -         -
     Auxiliary coordinates:
          latitude                            -         -        x         x
          longitude                           -         -        x         x
```

... using `iris.analysis.trajectory.interpolate` to do all the hard work.

We should investigate to diagnose the bottleneck. 

I'd be pretty interested in user feedback on their experience of using `iris.analysis.trajectory.interpolate` and whether it's performance is fit for purpose (within the context of its use).

ajdawson:
#1767 has details and links to user stories regarding this, seems like an issue for others too.

PR #4366, opened 2021-10-12
Increase speed of iris.analysis.trajectory.interpolate with linear interpolation

wjbenfold:
### Description

Moved call to cube.interpolate out of a loop in iris.analysis.trajectory.interpolate so that rather than instantiating an object every time, that object only gets made once. This leads to a significant speed improvement over the method.

The previous slow speed is noted in issue #1767. I don't think the linked #2168 is affected because that applies to a test that uses nearest neighbour rather than linear interpolation.

### Questions for reviewer

- Is there a reason why this wasn't done previously, and therefore shouldn't be done now?
- It seems like it should be possible to use advanced indexing, or some other form of cube manipulation, to make the cube that results from the call to cube.interpolate the right size, but I'm yet to work out how. This would obviate the need for the loop entirely.
- Are there further obvious time savings that can be made by altering my current implementation?
- Should a benchmark be added to cover this?

---

### DIFF
--- a/benchmarks/benchmarks/trajectory.py
+++ b/benchmarks/benchmarks/trajectory.py
@@ -1,0 +1,48 @@
+# Copyright Iris contributors
+#
+# This file is part of Iris and is released under the LGPL license.
+# See COPYING and COPYING.LESSER in the root of the repository for full
+# licensing details.
+"""
+Trajectory benchmark test
+
+"""
+
+# import iris tests first so that some things can be initialised before
+# importing anything else
+from iris import tests  # isort:skip
+
+import numpy as np
+
+import iris
+from iris.analysis.trajectory import interpolate
+
+
+class TrajectoryInterpolation:
+    def setup(self) -> None:
+        # Prepare a cube and a template
+
+        cube_file_path = tests.get_data_path(
+            ["NetCDF", "regrid", "regrid_xyt.nc"]
+        )
+        self.cube = iris.load_cube(cube_file_path)
+
+        trajectory = np.array(
+            [np.array((-50 + i, -50 + i)) for i in range(100)]
+        )
+        self.sample_points = [
+            ("longitude", trajectory[:, 0]),
+            ("latitude", trajectory[:, 1]),
+        ]
+
+    def time_trajectory_linear(self) -> None:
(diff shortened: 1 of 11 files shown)


wjbenfold:
image

wjbenfold:
There are now unit tests for linear interpolation in `test_interpolate.py`. The two unit tests below in `test_Trajectory.py` seem like they might sit better in `test_interpolate.py` too given that they're testing `interpolate` rather than `Trajectory`, and `Trajectory` has a test to make sure it's calling interpolate?

https://github.com/SciTools/iris/blob/05c1850376df1c145adc92cd73d9a94aeaf2fa87/lib/iris/tests/unit/analysis/trajectory/test_Trajectory.py#L134-L177

wjbenfold:
To do:

- [ ] Consider moving `test_Trajectory.py` tests into `test_interpolate.py`
- [ ] Fix CML failures (seem to be dtype based)

wjbenfold:
The failing tests are due to dtypes now being preserved, where previously they weren't. This is causing

* float32 cubes to come out as float32 rather than float64. Ideally as we're interpolating we'd use the higher precision but given that we were previously casting the result of the underlying function to float64 rather than maintaining a higher precision, we're not losing anything by returning the same dtype as went in

* float64 coordinates are coming out as float64, rather than being reduced to float32. This seems like an improvement.

If the reviewer is happy, and I've understood the fix for this right, I'll delete the relevant .CML files locally and then run the tests (which will regenerate them with the new dtypes) and then push those changes to the test data repo

bjlittle:
@wjbenfold Fancy rebasing to resolve your conflicts :+1:

bjlittle:
@wjbenfold I'm keen to bank this before it goes even more stale... where we at with it?

Just need a rebase (again)?

wjbenfold:
Have rebased. I think we need @trexfeathers to agree I added some tests to `test_interpolate.py` then we're good to go?

bjlittle (approved):
@wjbenfold Awesome, looks good to me.

Keen to bank this and get some mileage on it :+1:

trexfeathers (approved):

PR #4367, opened 2021-10-12
Only try to work out the differences between points for multiple

### LINKED ISSUE #4250: Avoid warning when creating a coordinate with one point
## 🐛 Bug Report
When constructing a coordinate's points, Iris checks whether the points are regularly spaced (see https://github.com/SciTools/iris/blob/main/lib/iris/util.py#L1458-L1464). If the coordinate has a single point, `np.mean(diffs)` is performed on an empty ndarray, and raises a numpy warning.

## How To Reproduce

```
DimCoord.from_regular(zeroth=0, step=20, count=1)
````

(with numpy v1.19.2) Returns:
```
/path_to_env/lib/python3.6/site-packages/numpy/core/fromnumeric.py:3373: RuntimeWarning: Mean of empty slice.
  out=out, **kwargs)
/path_to_env/lib/python3.6/site-packages/numpy/core/_methods.py:170: RuntimeWarning: invalid value encountered in true_divide
  ret = ret.dtype.type(ret / rcount)

DimCoord(array([20.], dtype=float32), standard_name=None, units=Unit('unknown'))
```

## Expected behaviour
Perhaps skip the check that the points are regularly spaced if the points are of length 1?

wjbenfold:
### Description

Causes `iris.util.points_step` to check whether there are multiple points to check before trying (thus averting a warning is caused by only one point being provided).

Fixes #4250.

### Questions for reviewer

- Is it better to apply the fix here, or to apply it in `coords.from_regular` and thus leave this so that it warns when called with a single point (albeit cryptically) so that people calling it are aware that they may not be doing what they meant? I personally think that the np.nan return value of avdiff (which matches the current behaviour) is sufficient.

---

### DIFF
--- a/lib/iris/tests/unit/util/test__coord_regular.py
+++ b/lib/iris/tests/unit/util/test__coord_regular.py
@@ -94,6 +94,18 @@ class Test_points_step(tests.IrisTest):
         self.assertEqual(exp_avdiff, result_avdiff)
         self.assertFalse(result)
 
+    def test_single_point(self):
+        lone_point = np.array([4])
+        result_avdiff, result = points_step(lone_point)
+        self.assertTrue(np.isnan(result_avdiff))
+        self.assertTrue(result)
+
+    def test_no_points(self):
+        no_points = np.array([])
+        result_avdiff, result = points_step(no_points)
+        self.assertTrue(np.isnan(result_avdiff))
+        self.assertTrue(result)
+
 
 if __name__ == "__main__":
     tests.main()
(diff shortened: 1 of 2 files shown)


wjbenfold:
Changes made as requested. Unit tests raised questions of what the correct behaviour with an empty array supplied is.

Will rebase then fix whatsnew once @bjlittle is happy with everything else

lbdreyer:
This looks good to go! I'd like to get it merged. @wjbenfold could I ask you to rebase to resolve the conflicting files?

bjlittle (approved):

rcomer:
Thanks All for sorting this 😊

PR #4377, opened 2021-10-15
Use partial to make cube pickleable

### LINKED ISSUE #4354: convert_units breaks pickle/multiprocessing for lazy data
## 🐛 Bug Report

When working with lazy data the Cube.convert_units method creates a temporary function for the unit conversion which cannot be pickled. As it cannot be pickled it also break multiprocessing.

1. cube = (load a cube)
2. cube.convert_units(new_units)
3. pickle.dumps(cube)

## Expected behaviour
In this case, Pickle should dump a string representation of the cube.

## Actual Behviour ##
Exception raised: 
`AttributeError: Can't pickle local object 'Cube.convert_units.<locals>.pointwise_convert'`

## Environment 
 - OS & Version: RHEL7
 - Iris Version: 2.4.0

## Additional context

<details>

I originally encountered this issue trying to make plots with a library that splits the cube up for batch plotting over multiple processors. However, after running a few tests, this does seem to be a problem caused by pickling cubes.

For now, I am working around this by realizing the data before calling the plotting routines, but it would (hypothetically) be more memory efficient to call the routines with lazy data.

</details>

wjbenfold:
### Description

Where previously a function was defined within cube.convert_units if the cube had lazy data, now a partial object is created instead.

The definition of a function at runtime was preventing pickling of a cube that had lazy data and was awaiting a convert_units operation. This is no longer the case. (fixes #4354)

---

### DIFF
--- a/docs/src/whatsnew/latest.rst
+++ b/docs/src/whatsnew/latest.rst
@@ -140,6 +140,10 @@ This document explains the changes made to Iris for this release
    coordinate bounds using minimum and maximum for unordered coordinates,
    fixing :issue:`1528`. (:pull:`4315`)
 
+#. `@wjbenfold`_ changed how a delayed unit conversion is performed on a cube
+   so that a cube with lazy data awaiting a unit conversion can be pickled.
+   (:issue:`4354 `, :pull:`4377`)
+
 
 💣 Incompatible Changes
 =======================
@@ -150,7 +154,9 @@ This document explains the changes made to Iris for this release
 🚀 Performance Enhancements
 ===========================
 
-#. N/A
+#. `@wjbenfold`_ resolved an issue that previously caused regridding with lazy
+   data to take significantly longer than with real data. Relevant benchmark
+   shows a time decrease from >10s to 625ms. (:issue:`4280`, :pull:`4400`)
 
 
 🔥 Deprecations
--- a/docs/src/whatsnew/latest.rst
+++ b/docs/src/whatsnew/latest.rst
@@ -152,7 +152,7 @@ This document explains the changes made to Iris for this release
 
 #. `@wjbenfold`_ resolved an issue that previously caused regridding with lazy
    data to take significantly longer than with real data. Benchmark
-   :class:`HorizontalChunkedRegridding.time_regrid_area_w` shows a time decrease
+   :class:`benchmarks.HorizontalChunkedRegridding` shows a time decrease
    from >10s to 625ms. (:issue:`4280`, :pull:`4400`)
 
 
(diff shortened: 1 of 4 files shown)


wjbenfold:
Not sure this is the right place to put these tests - it's a bit odd that we're not testing a function of cube.

Checked that the third test fails against the version of cube.py on main.

Not sure if it's worth checking any other cube methods preserve the ability to pickle, none stood out as likely to affect it.

Will rebase and update what's new when @bjlittle is happy with everything else

lbdreyer:
@wjbenfold Do you mind also resolving the conflicts on this PR?

wjbenfold:
After more faffing than strictly necessary, this is now rebased

docs/src/whatsnew/latest.rst
```diff
@@ -150,7 +154,9 @@ This document explains the changes made to Iris for this release
 🚀 Performance Enhancements
 ===========================
 
-#. N/A
+#. `@wjbenfold`_ resolved an issue that previously caused regridding with lazy
+   data to take significantly longer than with real data. Relevant benchmark
+   shows a time decrease from >10s to 625ms. (:issue:`4280`, :pull:`4400`)
```
bjlittle, line 159: @wjbenfold I assume that you're cheekily slipping this forgotten `whatsnew` entry into the mix?

bjlittle: Can you cite the relevant benchmark somehow?

wjbenfold: If I recall correctly, I was moving it from "bug fixes" to "performance enhancements". Looks like I just duplicated it though (or at least once merges were all resolved that's where we got to). Where do you reckon it belongs?

bjlittle: @wjbenfold I reckon `Performance Enhancements` is the right section, but drop it from `Bugs Fixed`... we tend not to duplicate entries and I'd see it more as an optimistation rather than a bug fix IMHO

docs/src/whatsnew/latest.rst
```diff
@@ -150,7 +150,10 @@ This document explains the changes made to Iris for this release
 🚀 Performance Enhancements
 ===========================
 
-#. N/A
+#. `@wjbenfold`_ resolved an issue that previously caused regridding with lazy
+   data to take significantly longer than with real data. Benchmark
+   :class:`HorizontalChunkedRegridding.time_regrid_area_w` shows a time decrease
```
bjlittle, line 155: @wjbenfold We can't actually link to this, but I guess it should be :class:\`benchmarks.HorizontalChunkedRegridding` instead?

bjlittle, line 155: We can't actually link to this, but I guess it should be this instead...
```suggestion
   :class:`benchmarks.HorizontalChunkedRegridding` shows a time decrease
```

bjlittle (approved):

PR #4380, opened 2021-10-18
adopt flake8 maccabe complexity metric

bjlittle:
### Description
This PR introduces the `flake8` [mccabe](https://github.com/PyCQA/mccabe) algorithm to provide a metric on `iris` code complexity.

This empirical metric can then form part of the discussion between a PR author and reviewer/s with regards to the complexity of code being proposed.

According to [McCabe](https://github.com/PyCQA/mccabe#plugin-for-flake8), code with a score greater than `10` is deemed too complex. Currently `iris` has a `mccabe` score of `50`.

Adopting the `maccabe` metric will allow us to target and reduce complexity within `iris`, but also be aware of any potential rise in complexity.

As per `flake8`, the complexity warnings/errors issued by `mccabe` can be explicitly ignored, if so deemed suitable and appropriate. 

---

### DIFF
--- a/setup.cfg
+++ b/setup.cfg
@@ -97,6 +97,7 @@ all =
 #   https://pycodestyle.readthedocs.io/en/latest/intro.html#error-codes
 
 max-line-length = 80
+max-complexity = 50
 select = C,E,F,W,B,B950
 ignore =
     # E203: whitespace before ':'


bjlittle:
Bumping the complexity down to `1` makes interesting reading...

jamesp:
Interesting! Is it possible to get some examples from the iris codebase along with their complexity scores? Reading the wikipedia link @rcomer provided:
> Cyclomatic complexity may also be applied to individual functions, modules, methods or classes within a program.

Do we know what level this tool applies? From the output shown on the [project page](https://github.com/PyCQA/mccabe#plugin-for-flake8) it certainly does per-function analysis

bjlittle:
Great question...

It's early days, and this is all new to us. I'd happily opt for just rolling with it on a PR by PR basis, making in-situ judgement calls, and watching out for patterns on how this affects our workflow.

If we take this all under the remit of the Peloton and review on occasion how were fairing, that seems like a heathy and reasonable approach to take.

What say ye?

trexfeathers:
> From @bjlittle’s posted output, there is only one offender with a score over 30.  So if something else goes over 50, that would indicate a genuine problem I think.

@rcomer sure, for now. But I hope over time we are able to drop the threshold and then it will become more pertinent to think about "accepting failures". I'm happy to go with @bjlittle's approach of rolling with the punches for now; personally I'll be pushing for `# noqa` if it comes up.

trexfeathers:
>have to dual with the reviewer

Sounds much more collaborating and less adversarial than "have a duel"!

>I'm also hoping that I can always play the "reduce complexity card" without jumping through hoops, otherwise then it becomes a author/reviewer discussion i.e., a strong signal that it might be harder to get the pull-request merged without very good reason.

👍 There are so many things we all agree we want, but it's often hard to agree what that looks like. So having an objective vehicle for such things is a positive step, like `black` does for readable code style. As you say, there will still be discussions at times, but hopefully fewer.

PR #4387, opened 2021-10-22
Make concat mismatch warning for scalar coords more accurate

### LINKED ISSUE #4096: Concatenate: wrong error message about scalar coords
## How To Reproduce

Attempt to concatenate cubes that are identical apart from the coordinate points.

```python
import iris

print(iris.__version__)

cube1 = iris.cube.Cube(
    range(3),
    dim_coords_and_dims=[(iris.coords.DimCoord(range(3), long_name='foo'), 0)],
    aux_coords_and_dims=[(iris.coords.AuxCoord(1, long_name='bar'), None)])

cube2 = iris.cube.Cube(
    range(3),
    dim_coords_and_dims=[(iris.coords.DimCoord(range(3, 6), long_name='foo'), 0)],
    aux_coords_and_dims=[(iris.coords.AuxCoord(2, long_name='bar'), None)])

print(iris.cube.CubeList([cube1, cube2]).concatenate_cube())
```
Get an error about mismatched **metadata**:

```
3.0.1
  File "concatenate_error.py", line 15, in <module>
    print(iris.cube.CubeList([cube1, cube2]).concatenate_cube())
  File "[site-packages-path]/iris/cube.py", line 551, in concatenate_cube
    check_ancils=check_ancils,
  File "[site-packages-path]/iris/_concatenate.py", line 324, in concatenate
    check_ancils,
  File "[site-packages-path]/iris/_concatenate.py", line 839, in register
    match = self._cube_signature.match(cube_signature, error_on_mismatch)
  File "[site-packages-path]/iris/_concatenate.py", line 567, in match
    raise iris.exceptions.ConcatenateError(msgs)
iris.exceptions.ConcatenateError: failed to concatenate into a single cube.
  Scalar coordinates metadata differ: bar != bar
```

## Expected behaviour

The error should be about the points, not the metadata.

## Environment 
 -  OS & Version: RHEL7.9
 -  Iris Version: 3.0.1

wjbenfold:
### Description

Added a "reason" keyword argument to `_coordinate_differences` as currently it attaches the reason, and whilst it can diagnose some of the problem, it doesn't know if it's just been given metadata or also values to check, whereas the calling method does.

Fixes #4096

Question for reviewer:
- Is this the best place to implement this fix? Should it be shallower or deeper instead?

---

### DIFF
--- a/lib/iris/_concatenate.py
+++ b/lib/iris/_concatenate.py
@@ -456,8 +456,8 @@ class _CubeSignature:
             always, testing metadata)
 
         Returns:
-            Tuple of the names of attributes that differ between `self` and
-            `other`.
+            Tuple of a descriptive error message and the names of attributes
+            that differ between `self` and `other`.
 
         """
         # Set up {name: attribute} dictionaries.
--- a/lib/iris/_concatenate.py
+++ b/lib/iris/_concatenate.py
@@ -437,7 +437,7 @@ class _CubeSignature:
             av_and_dims = _CoordAndDims(av, tuple(dims))
             self.ancillary_variables_and_dims.append(av_and_dims)
 
-    def _coordinate_differences(self, other, attr):
+    def _coordinate_differences(self, other, attr, reason="metadata"):
         """
         Determine the names of the coordinates that differ between `self` and
         `other` for a coordinate attribute on a _CubeSignature.
@@ -450,6 +450,10 @@ class _CubeSignature:
         * attr (string):
             The _CubeSignature attribute within which differences exist
             between `self` and `other`.
+
+        * reason (string):
+            The reason to give for mismatch (function is normally, but not
+            always, testing metadata)
 
         Returns:
             Tuple of the names of attributes that differ between `self` and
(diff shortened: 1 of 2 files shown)


rcomer:
Hi @wjbenfold, I’m happy to have a look at this, though I don’t think I’m qualified to answer this question:

> Is this the best place to implement this fix? Should it be shallower or deeper instead?

jamesp (commented):
Seems reasonable to me.  How many places is this function called in the codebase? Is it just the two you point out in #4096?
Do we need an additional test that picks up the default case?

wjbenfold:
>  How many places is this function called in the codebase?

It's called repeatedly in the method where it's called on scalar coordinates, but nowhere else I can see:

https://github.com/SciTools/iris/blob/2d53161af562b513bb36438f47ed315b292bac65/lib/iris/_concatenate.py#L525-L554

> Do we need an additional test that picks up the default case?

This is already covered by the other tests in that file, e.g.

https://github.com/SciTools/iris/blob/2d53161af562b513bb36438f47ed315b292bac65/lib/iris/tests/unit/concatenate/test_concatenate.py#L137-L143

jamesp (approved):

rcomer:
Thanks for doing this @wjbenfold and @jamesp 🎉

Issue #4382, opened 2021-10-18
Make a statement about adoption of type hints in Iris 3.2


pp-mo:
## 📚 Documentation
We should maybe come up with a policy for future use of typing in Iris code, especially API methods + functions.
See also discussion : https://github.com/SciTools/iris/discussions/4383

trexfeathers:
Closed by #4390

PR #4390, opened 2021-10-27
Suggest type hinting

trexfeathers:
### Description

A simple note in the documentation to suggest type hinting when contributing to the codebase. Moves us forward with #4272 / #4382 / #4383, since everyone seems in agreement.

---


docs/src/developers_guide/contributing_code_formatting.rst
```diff
@@ -57,5 +57,12 @@ will look similar to::
 .. note:: You can also run `black`_ and `flake8`_ manually.  Please see the
           their officially documentation for more information.
 
+Type Hinting
+------------
+Iris is gradually adding
+`type hints <https://docs.python.org/3/library/typing.html>`_ into the
+codebase. Where possible please include/add type hints in code that you are
```
rcomer, line 64:
```suggestion
codebase. If you feel confident to do so, please include/add type hints in code that you are
```

rcomer (commented):
I haven’t got involved in the discussions about type hinting, basically because I know very little about it.  For the same reason, I think it would be quite a learning curve for me to include it in my PRs (and more work for the reviewer to set me straight when I get it wrong!)

That said, if a contributor is used to using type hints in other projects then it’s probably No Big Deal for them to also add them in Iris.  So maybe just relax the language to indicate that type hints are encouraged but not compulsory?

docs/src/developers_guide/contributing_code_formatting.rst
trexfeathers: I get what you're saying, but I feel it's something worth pushing for to a certain degree, despite the anticipated pain. It has been explained to me that typing is more than a helpful decoration - widespread typing makes the codebase more robust, and even allows for a smaller test suite.

So I'm cautious about making the language **too** forgiving. Can we have a third opinion @SciTools/iris-devs?

pp-mo: Can we put it onto the reviewer ?
suggest that reviewer "will consider"

pp-mo: Also @bjlittle suggests : especially new code

trexfeathers: 1212d263be1002f7b66e816220c8c9ecf7f907be

lbdreyer:
This is looking good to me. Unless there are any objections I will merge this in by the end of the week

lbdreyer:
Oh dear! I was just about to merge this but I see there is now a conflict! @trexfeathers do you mind rebasing?

lbdreyer (approved):

PR #4392, opened 2021-10-29
Show acceptable image test results in the docs

jamesp:
### Description

There are times when I am running the tests on awkward remote platforms and I get failures due to hamming distance of graphical tests failing. It would be helpful in these circumstances to see the contents of [scitools/test-iris-imagehash](https://github.com/SciTools/test-iris-imagehash) ordered by test, rather than hash.

This PR adds autogenerated pages to the iris docs that show, by test, the acceptable image results.

The pages added:

1. An index page of all graphical tests listed in `imagerepo.json`:
image
2. A page per test with all acceptable images linked inline:
image

The images are linked and loaded directly from the scitools/test-iris-imagehash repository.

---


bjlittle (changes requested):
@jamesp This is really awesome! Thanks :+1:  

My only comment is that Python files under `docs/src/sphinxext` are explicitly not checked by `black`, `flake8` or `isort`... happy to live with that if you are?

The alternative is either:

1. we add exceptions to the rule for the linters et al configurations
1. manually lint (I'm assuming that this code will be pretty stable and not subject to much change)
1. we move `docs/src/sphinxext/image_test_output.py` to live somewhere else and _somehow_ hook it in

jamesp:
I'm happy to do a manual run of those tools over this code  👍

bjlittle (approved):

bjlittle:
@jamesp Works for me dude, nice one 🤟

PR #4397, opened 2021-11-03
Don't realise lazy coords in netcdf save.

pp-mo:
### Description
Performance testing revealed that the [recent changes to stream all lazy data in netcdf saves](https://github.com/SciTools/iris/pull/4375) didn't quite fix the problem of realising lazy coordinates, as intended
I.E. nc-save is still loading all lazy aux-coords into memory.

So this fixes that, and adds a test to ensure the behaviour.

---

### DIFF
--- a/lib/iris/fileformats/netcdf.py
+++ b/lib/iris/fileformats/netcdf.py
@@ -1190,9 +1190,9 @@ class Saver:
 
         # Ensure that attributes are CF compliant and if possible to make them
         # compliant.
-        self.check_attribute_compliance(cube, cube.lazy_data())
+        self.check_attribute_compliance(cube, cube.dtype)
         for coord in cube.coords():
-            self.check_attribute_compliance(coord, coord.points)
+            self.check_attribute_compliance(coord, coord.dtype)
 
         # Get suitable dimension names.
         mesh_dimensions, cube_dimensions = self._get_dim_names(cube)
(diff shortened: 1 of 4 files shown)


lib/iris/fileformats/netcdf.py
```diff
@@ -1190,9 +1190,9 @@ def write(
 
         # Ensure that attributes are CF compliant and if possible to make them
         # compliant.
-        self.check_attribute_compliance(cube, cube.lazy_data())
+        self.check_attribute_compliance(cube, cube.dtype)
         for coord in cube.coords():
-            self.check_attribute_compliance(coord, coord.points)
```
pp-mo, line 1195: This was the problem line that realised all the points
(though bounds were ok, interestingly)

I think we **_could_** just replace the `points` here with `core_points()`
However, I also realised that `check_attribute_compliance` call doesn't actually _use_ the data array at all, it only uses the _dtype_ of it.
So, I've replaced that throughout, as an alternative to passing a 'safe' version of the data in the problem location.
Unfortunately that does causes a lot of noisy changes in the tests (though very little in functional code).

lib/iris/tests/integration/test_netcdf.py
```diff
@@ -273,10 +273,16 @@ def test_lazy_preserved_save(self):
...
         self.assertTrue(acube.has_lazy_data())
+        # Also check a coord with lazy points + bounds.
+        self.assertTrue(acube.coord("forecast_period").has_lazy_points())
+        self.assertTrue(acube.coord("forecast_period").has_lazy_bounds())
         with self.temp_filename(".nc") as nc_path:
             with Saver(nc_path, "NETCDF4") as saver:
                 saver.write(acube)
+        # Check that cube data is not realised, also coord points + bounds.
         self.assertTrue(acube.has_lazy_data())
+        self.assertTrue(acube.coord("forecast_period").has_lazy_points())
```
pp-mo, line 284: NOTE: this check fails with the older code.

lib/iris/tests/unit/fileformats/netcdf/test_Saver__lazy.py
```diff
@@ -122,3 +122,7 @@ def test_lazy_streamed_bounds(self):
         self.cube.replace_coord(lazy_coord)
         self.save_common(self.cube)
         self.assertTrue(self.store_watch.called)
+
+
+if __name__ == "__main__":
```
pp-mo, line 127: Just for consistency, so you can run these tests directly.

trexfeathers (approved):
Nice catch @pp-mo

PR #4402, opened 2021-11-08
Support test data in benchmark workflows

wjbenfold:
### Description

Adds test data support to github action benchmarks (by downloading the iris test data and setting the `OVERRIDE_TEST_DATA_REPOSITORY` environment variable).
Adds an example benchmark (of area weighted regridding) that uses the test data

Currently github action benchmarks don't have access to the test data. I'm aiming to improve this.

To do:

- [x] Enable caching (by uncommenting) once the test data download is tested
- [x] Revert noxfile logging changes
- [x] Change the example benchmark to use new test data that shows significant slowness (once it's in the iris-test-data repo)
- [x] What's new

---

### DIFF
--- a/.github/workflows/benchmark.yml
+++ b/.github/workflows/benchmark.yml
@@ -14,7 +14,7 @@ jobs:
       IRIS_TEST_DATA_PATH: /benchmarks/iris-test-data
       IRIS_TEST_DATA_VERSION: "2.5"
       # Lets us manually bump the cache to rebuild
-      TEST_DATA_CACHE_BUILD: "0"
+      TEST_DATA_CACHE_BUILD: "1"
 
     steps:
       # Checks-out your repository under $GITHUB_WORKSPACE, so your job can access it
@@ -49,9 +49,9 @@ jobs:
         uses: actions/cache@v2
         with:
           path: |
-            ${{ env.IRIS_TEST_DATA_PATH }}
-          key: 
-            test-data-${{ env.IRIS_TEST_DATA_VERSION }}-${{ env.TEST_DATA_CACHE_BUILD}}
+            ${{ env.GITHUB_WORKSPACE }}${{ env.IRIS_TEST_DATA_PATH }}
+          key:
+            test-data-${{ env.IRIS_TEST_DATA_VERSION }}-${{ env.TEST_DATA_CACHE_BUILD }}
 
       - name: Fetch the test data
         if: steps.cache-test-data.outputs.cache-hit != 'true'
--- a/.github/workflows/benchmark.yml
+++ b/.github/workflows/benchmark.yml
@@ -11,6 +11,7 @@ jobs:
     runs-on: ubuntu-latest
 
     env:
+      IRIS_TEST_DATA_LOC_PATH: benchmarks
       IRIS_TEST_DATA_PATH: benchmarks/iris-test-data
       IRIS_TEST_DATA_VERSION: "2.5"
       # Lets us manually bump the cache to rebuild
@@ -58,9 +59,8 @@ jobs:
         run: |
           wget --quiet https://github.com/SciTools/iris-test-data/archive/v${IRIS_TEST_DATA_VERSION}.zip -O iris-test-data.zip
           unzip -q iris-test-data.zip
-          mkdir --parents ${GITHUB_WORKSPACE}/${IRIS_TEST_DATA_PATH}
-          mv iris-test-data-${IRIS_TEST_DATA_VERSION} iris-test-data
-          mv iris-test-data ${GITHUB_WORKSPACE}/benchmarks
+          mkdir --parents ${GITHUB_WORKSPACE}/${IRIS_TEST_DATA_LOC_PATH}
+          mv iris-test-data-${IRIS_TEST_DATA_VERSION} ${GITHUB_WORKSPACE}/${IRIS_TEST_DATA_PATH}
           
       - name: Set test data var
         run: |
--- a/.github/workflows/benchmark.yml
+++ b/.github/workflows/benchmark.yml
@@ -52,7 +52,7 @@ jobs:
             test-data-${{ env.IRIS_TEST_DATA_VERSION }}
 
       - name: Fetch the test data
-        # if: steps.cache-test-data.outputs.cache-hit != 'true'
+        if: steps.cache-test-data.outputs.cache-hit != 'true'
         run: |
           wget --quiet https://github.com/SciTools/iris-test-data/archive/v${IRIS_TEST_DATA_VERSION}.zip -O iris-test-data.zip
           unzip -q iris-test-data.zip
--- a/.github/workflows/benchmark.yml
+++ b/.github/workflows/benchmark.yml
@@ -13,6 +13,7 @@ jobs:
     env:
       IRIS_TEST_DATA_DIR: benchmarks/iris-test-data
       OVERRIDE_TEST_DATA_REPOSITORY: benchmarks/iris-test-data/test_data
+      IRIS_TEST_DATA_VERSION: "2.4"
 
     steps:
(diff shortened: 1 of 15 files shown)


.github/workflows/benchmark.yml
```diff
@@ -38,6 +42,27 @@ jobs:
           key: ${{ runner.os }}-${{ github.sha }}
           restore-keys: ${{ runner.os }}
 
+      - name: Get test data version
+        run: |
+          echo "IRIS_TEST_DATA_VERSION=$(wget --quiet -O - https://raw.githubusercontent.com/SciTools/iris-test-data/master/version.txt)" >> $GITHUB_ENV
```
trexfeathers, line 47: I'm uncomfortable that the versioning is managed differently depending on the test run in question:

https://github.com/SciTools/iris/blob/2202793568ca1eb23f1f8e9be42afae0f1383f34/.cirrus.yml#L40-L41

Also note that the use of release versions of iris-test-data was a deliberate decision taken in the last year for iris and iris-esmf-regrid. And before that we used commit hashes for some years, rather than just taking the latest. I can't recall the arguments for not using the very latest (if you're interested you might be able to follow the blame of `travis.yml` back when we used that).

Not sure if there's a sensible way to centralise this? At the very least we should make sure the versions are the same.

wjbenfold: It was not knowing how to centralise it that pushed me towards grabbing it out of the repo itself. Maybe it belongs in iris.config?

I can see that there will be times it's best not to use the latest version - if only so that someone committing to a different repo can't silently affect your test results

trexfeathers: >someone committing to a different repo can't silently affect your test results

Oh yeah of course! And we have in the past discussed putting routines in there too which would be a proper injection risk.

>It was not knowing how to centralise it that pushed me towards grabbing it out of the repo itself. Maybe it belongs in iris.config?

I think this is a temporary concern and the easy solution is the right one. #4193 suggests there are a lot of reasons for switching everything to GHA and then there are probably many more possibilities for shared caches/artifacts/etc.

trexfeathers:
>Change the example benchmark to use new test data that shows significant slowness (once it's in the iris-test-data repo)

@wjbenfold if you figure out a decent pattern to host full size PP files online that will potentially solve some other benchmarking headaches so I'd be interested in the result!

wjbenfold:
I think this PR is now ready for review (save a what's new).

Reviewer questions:

- Are the changes to `benchmarks/nox_asv_plugin.py`, `benchmarks/benchmarks/mixin.py` and `benchmarks/benchmarks/metadata_manager_factory.py` as a result of changing `.pre-commit-config.yaml` acceptable? (This was done because the ci precommit was checking more aggressively than my local precommit and thus forcing me to pull the changes down to my local copy.)
- I've used the test data to access files for benchmarking, ignoring the code in `benchmarks/benchmarks/__init__.py` that sets a data path. Is this ok, and if so should `__init__.py` be tidied at all (given I don't think it's doing anything to do with test data for us)?

.github/workflows/benchmark.yml
```diff
@@ -38,6 +44,28 @@ jobs:
...
+            test-data-${{ env.IRIS_TEST_DATA_VERSION }}-${{ env.TEST_DATA_CACHE_BUILD }}
+
+      - name: Fetch the test data
+        if: steps.cache-test-data.outputs.cache-hit != 'true'
+        run: |
+          wget --quiet https://github.com/SciTools/iris-test-data/archive/v${IRIS_TEST_DATA_VERSION}.zip -O iris-test-data.zip
+          unzip -q iris-test-data.zip
+          mkdir --parents ${GITHUB_WORKSPACE}/${IRIS_TEST_DATA_PATH}
+          mv iris-test-data-${IRIS_TEST_DATA_VERSION} iris-test-data
+          mv iris-test-data ${GITHUB_WORKSPACE}/benchmarks
```
trexfeathers, line 63:
```suggestion
          mv iris-test-data-${IRIS_TEST_DATA_VERSION} ${GITHUB_WORKSPACE}/${IRIS_TEST_DATA_PATH}
```

There's presumably no risk that it might already exist in this scenario. If so you could also add `rm -rf ${GITHUB_WORKSPACE}/${IRIS_TEST_DATA_PATH}` first.

trexfeathers (changes requested):
Thanks @wjbenfold! Some comments for you.

I don't get how[ an performance improvement has been detected](https://github.com/SciTools/iris/runs/4164171350?check_suite_focus=true#step:9:464) for this PR, which makes no changes to the regridding code? It may be a sign that performance of this one is highly variable; I'll try re-running it.

.github/workflows/benchmark.yml
```diff
@@ -38,6 +44,28 @@ jobs:
...
+          
+      - name: Set test data var
+        run: |
+          echo "OVERRIDE_TEST_DATA_REPOSITORY=${GITHUB_WORKSPACE}/${IRIS_TEST_DATA_PATH}/test_data" >> $GITHUB_ENV
+
       - name: Run CI benchmarks
         run: |
           mkdir --parents benchmarks/.asv
```
trexfeathers, line 71: During reviewing of this PR I've discovered that the below piping hides the exit status so that benchmarking failures are not detected!

```suggestion
          mkdir --parents benchmarks/.asv
          set -o pipefail
```

trexfeathers:
>I don't get how[ an performance improvement has been detected](https://github.com/SciTools/iris/runs/4164171350?check_suite_focus=true#step:9:464) for this PR, which makes no changes to the regridding code? It may be a sign that performance of this one is highly variable; I'll try re-running it.

From investigation this looks like it was an aberration, so no changes needed.

(For the record I was going to suggest adjusting the [`repeat` attribute](https://asv.readthedocs.io/en/stable/benchmarks.html#timing-benchmarks) to force ASV to spend longer / do more repeats. Not needed here but a good tool to have).

benchmarks/benchmarks/regridding.py
```diff
@@ -0,0 +1,41 @@
...
+# import iris tests first so that some things can be initialised before
+# importing anything else
+from iris import tests  # isort:skip
+
+import iris
+from iris.analysis import AreaWeighted
+
+
+@tests.skip_data
+class RegriddingTests:
```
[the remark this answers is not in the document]
wjbenfold: Done

benchmarks/benchmarks/regridding.py
```diff
@@ -0,0 +1,41 @@
...
+
+# import iris tests first so that some things can be initialised before
+# importing anything else
+from iris import tests  # isort:skip
+
+import iris
+from iris.analysis import AreaWeighted
+
+
+@tests.skip_data
```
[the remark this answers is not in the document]
wjbenfold: Done

.github/workflows/benchmark.yml
```diff
@@ -10,6 +10,12 @@ jobs:
   benchmark:
     runs-on: ubuntu-latest
 
+    env:
+      IRIS_TEST_DATA_PATH: benchmarks/iris-test-data
+      IRIS_TEST_DATA_VERSION: "2.5"
```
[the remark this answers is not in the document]
wjbenfold: Done

.github/workflows/benchmark.yml
wjbenfold: I need in some way to make sure the parent directory exists before moving the directory. I've done this by introducing a another env variable, which seems cleanest of the things I've thought of but still seems a bit messy

trexfeathers (approved):

trexfeathers (changes requested):
I'm happy that you've actioned my comments. Still need that What's New entry though!

trexfeathers (approved):

PR #4407, opened 2021-11-10
Netcdf saving: workaround known netcdf bug.

pp-mo:
## 🚀 Pull Request

A response to having encountered [this known netcdf bug](https://github.com/Unidata/netcdf-c/issues/1772)

I think this should *not* upset existing behaviour.
We might need to modify some test results, but I expect **_not_**, since any cases that trigger the code changes would have failed with the HDF error it is trying to avoid.

There is another question as to test coverage : 
I've added one test, which fails with the older code.
However, I have implemented "the fix" in multiple code locations, wherever dimensions are created -- or in the case of 'get_dim_names' method, where the names get chosen, prior to creation.
So in principle, for full coverage, I **_could_** add more tests to trigger each of those cases.
I'm not yet sure whether I think that is worth it ...

### DIFF
--- a/lib/iris/tests/unit/fileformats/netcdf/test_save.py
+++ b/lib/iris/tests/unit/fileformats/netcdf/test_save.py
@@ -273,26 +273,24 @@ class Test_HdfSaveBug(tests.IrisTest):
         """
         A special tolerant cube compare.
 
-        Either both are unstructured, or neither.
-        Strip off any Meshes (as we can't compare unstructured cubes).
         Ignore any 'Conventions' attributes.
         Ignore all var-names.
 
         """
-        # Either they both have a mesh, or neither.
-        self.assertEqual(cube1.mesh is None, cube2.mesh is None)
 
         def clean_cube(cube):
             cube = cube.copy()  # dont modify the original
-            if cube.mesh:
-                # "Strip" off mesh, reducing MeshCoords to AuxCoords
-                cube = cube[..., 0:]  # magic "non-trivial" slice form
             # Remove any 'Conventions' attributes
             cube.attributes.pop("Conventions", None)
             # Remove var-names (as original mesh components wouldn't have them)
             cube.var_name = None
             for coord in cube.coords():
                 coord.var_name = None
+            mesh = cube.mesh
+            if mesh:
+                mesh.var_name = None
+                for component in mesh.coords() + mesh.connectivities():
+                    component.var_name = None
 
             return cube
 
@@ -312,7 +310,6 @@ class Test_HdfSaveBug(tests.IrisTest):
     def test_anonymous_dim_varname_collision(self):
         # Second cube is going to name an anonymous dim.
         cube_2 = Cube([0, 1], var_name="cube_2")
-        # cube_2 = Cube(0, var_name="cube_2")
         # First cube has a varname which collides with the dim-name.
         cube_1 = Cube([0, 1], long_name="cube_1", var_name="dim0")
         # Add a dimcoord to prevent the *first* cube having an anonymous dim.
--- a/lib/iris/tests/unit/fileformats/netcdf/test_save.py
+++ b/lib/iris/tests/unit/fileformats/netcdf/test_save.py
@@ -18,10 +18,12 @@ import netCDF4 as nc
 import numpy as np
 
 import iris
-from iris.coords import DimCoord
+from iris.coords import AuxCoord, DimCoord
 from iris.cube import Cube, CubeList
+from iris.experimental.ugrid import PARSE_UGRID_ON_LOAD
 from iris.fileformats.netcdf import CF_CONVENTIONS_VERSION, save
 from iris.tests.stock import lat_lon_cube
+from iris.tests.stock.mesh import sample_mesh_cube
 
 
 class Test_conventions(tests.IrisTest):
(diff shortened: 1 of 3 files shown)


trexfeathers (changes requested):
Thanks @pp-mo! I can confirm this fixes the code where the problem was first observed. Couple of minor changes to make things more readable.

trexfeathers:
>So in principle, for full coverage, I **_could_** add more tests to trigger each of those cases.
>I'm not yet sure whether I think that is worth it ...

This is a case of the 80-20 rule IMO. We already have good coverage, and we don't want to bloat the test run more than necessary, so I would leave as-is.

pp-mo:
Unfortunately, I decided that as I couldn't work out which code change was actually fixing the testcase, I though I should create more anyway.  And in fact, in the process, I discovered there is a problem anyway !!

So, pending my fixes ...

pp-mo:
> in the process, I discovered there is a problem anyway !!

I also discovered that cubes with string data don't seem to "work", and possibly haven't done for a long time.
 https://github.com/SciTools/iris/issues/4412

pp-mo:
Here it is, out of draft again after rather a lot of tiresome fiddling.
@trexfeathers I think we could now claim to have banked the learning here now, so there is really no great hurry getting this signed off.

The work that uncovered the problem is basically https://github.com/SciTools-incubator/iris-esmf-regrid/pull/130
This issue is not a blocker to that, since it adopted a simple workaround.

pp-mo:
Ping @trexfeathers I think this is good to go now ?

lib/iris/tests/unit/fileformats/netcdf/test_save.py
```diff
@@ -222,37 +224,140 @@ class Test_HdfSaveBug(tests.IrisTest):
...
 
+        # Return result cubes for any additional checks.
+        return result
+
+    def assertSameCubes(self, cube1, cube2):
+        """
+        A special tolerant cube compare.
+
+        Either both are unstructured, or neither.
+        Strip off any Meshes (as we can't compare unstructured cubes).
```
trexfeathers, line 277: No longer true! 🎉

lib/iris/tests/unit/fileformats/netcdf/test_save.py
```diff
@@ -222,37 +224,140 @@ class Test_HdfSaveBug(tests.IrisTest):
...
+
+        """
+        # Either they both have a mesh, or neither.
+        self.assertEqual(cube1.mesh is None, cube2.mesh is None)
+
+        def clean_cube(cube):
+            cube = cube.copy()  # dont modify the original
+            if cube.mesh:
+                # "Strip" off mesh, reducing MeshCoords to AuxCoords
+                cube = cube[..., 0:]  # magic "non-trivial" slice form
```
trexfeathers, line 289: We can ditch this now?

lib/iris/tests/unit/fileformats/netcdf/test_save.py
```diff
@@ -222,37 +224,140 @@ class Test_HdfSaveBug(tests.IrisTest):
...
+        cube_1 = Cube([0, 1], long_name="cube_1", var_name="dimco_name")
+        # Test save + loadback
+        reload_1, reload_2 = self._check_save_and_reload([cube_1, cube_2])
+        # As re-loaded, the coord will have a different varname.
+        self.assertEqual(reload_2.coord("dim_x").var_name, "dimco_name_0")
+
+    def test_anonymous_dim_varname_collision(self):
+        # Second cube is going to name an anonymous dim.
+        cube_2 = Cube([0, 1], var_name="cube_2")
+        # cube_2 = Cube(0, var_name="cube_2")
```
trexfeathers, line 315: Commented code

lib/iris/tests/unit/fileformats/netcdf/test_save.py
```diff
@@ -222,37 +224,140 @@ class Test_HdfSaveBug(tests.IrisTest):
...
+
+            # Load them back for roundtrip testing.
+            with PARSE_UGRID_ON_LOAD.context():
+                new_cubes = iris.load(str(filepath))
+
+            # There should definitely still be the same number of cubes.
+            self.assertEqual(len(new_cubes), len(cubes))
+
+            # Get results in the input order, matching by var_names.
+            result = [new_cubes.extract_cube(cube.var_name) for cube in cubes]
```
trexfeathers, line 258: Could you add a clause to confirm that the cube ordering has indeed been altered - evidencing that a name clash has been handled?

The test cases in this class are all based on assumptions of how Iris will name variables. If these conventions are ever altered without this test class being altered too (seems likely), the tests will continue to pass but will not be providing the coverage they were written for - there won't be any name clashes that need handling!

I assume this suggestion would break the string test until #4412 is solved? Not sure what to do about that.

trexfeathers (changes requested):
Sorry for the delay @pp-mo. I've reviewed the new changes since my last review.

lib/iris/tests/unit/fileformats/netcdf/test_save.py
trexfeathers: @pp-mo has explained to me that a change in `Cube` ordering is not symptomatic of a name clash having been handled.

There is _always_ a chance that `Cube`s will load in a different order to which they were saved. That's the real reason that this line of code exists.

trexfeathers (approved):
Nice @pp-mo, I know this is ready to come out of Draft so I've done that.

PR #4411, opened 2021-11-11
NAME loader fixes

bsherratt:
### Description

I recently discovered that as part of unpinning cftime in version 3.0.4 (#4222), a bug was introduced that makes it impossible to load a NAME file that contains only one time point. (Because `np.float_(arr)` on an array with one element returns a scalar, instead of another array with one element.)

I took the opportunity to fix some long-standing annoyances that we've so far just worked around:
- Vertical coordinates were never given the attribute `positive = up`, so the "axis" was not guessed as "Z".
- Iris tries to identify lat-lon coordinates in a column header with a regex for a float, but more recent versions of NAME may not use a decimal point.

(draft while I check whether there are other small things like that)

---

### DIFF
--- a/docs/src/whatsnew/latest.rst
+++ b/docs/src/whatsnew/latest.rst
@@ -88,7 +88,6 @@ This document explains the changes made to Iris for this release
 🐛 Bugs Fixed
 =============
 
-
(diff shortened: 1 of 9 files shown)


trexfeathers (approved):
Looks good, thanks for your diligence @bsherratt! I'm just re-running the py38 tests.

PR #4418, opened 2021-11-15
Remove newline to satisfy setuptools

wjbenfold:
### Description

In 59.0.0, on 12/11/21, setuptools introduced a [breaking change](https://setuptools.pypa.io/en/latest/history.html?highlight=description%2059%20breaking#breaking-changes) that no longer allows multi-line descriptions. This has caused `nox` to fail when running the tests with a `ValueError`.

---

### DIFF
--- a/setup.cfg
+++ b/setup.cfg
@@ -17,8 +17,7 @@ classifiers =
     Topic :: Scientific/Engineering
     Topic :: Scientific/Engineering :: Atmospheric Science
     Topic :: Scientific/Engineering :: Visualization
-description =
-    A powerful, format-agnostic, community-driven Python package for analysing and visualising Earth science data
+description = A powerful, format-agnostic, community-driven Python package for analysing and visualising Earth science data
 download_url = https://github.com/SciTools/iris
 keywords =
     cf-metadata


jamesp (approved):
Approved, we might want to change this back later for stylistic reasons when the setuptools bug has been fixed.  @wjbenfold can you add a todo to check back on it in a few weeks?

wjbenfold:
> Approved, we might want to change this back later for stylistic reasons when the setuptools bug has been fixed. @wjbenfold can you add a todo to check back on it in a few weeks?

Will do

wjbenfold:
> Approved, we might want to change this back later for stylistic reasons when the setuptools bug has been fixed. @wjbenfold can you add a todo to check back on it in a few weeks?

Doesn't look like setuptools have any [intention to revert or change this](https://github.com/pypa/setuptools/pull/2870) - we could propose a change to their code but this is minor enough that it doesn't seem worth it

trexfeathers:
>we could propose a change to their code but this is minor enough that it doesn't seem worth it

Agreed, particularly since the conversation elsewhere has generally been highlighting that the chance to act was during the long period while `DeprecationWarning` was being raised.

PR #4421, opened 2021-11-16
Removed addition of period from wrap_lons

akuhnregnier:
Updated affected tests using assertArrayAllClose following #3993.

### Description
Please see #3994.

### DIFF
--- a/docs/src/whatsnew/latest.rst
+++ b/docs/src/whatsnew/latest.rst
@@ -117,7 +117,7 @@ This document explains the changes made to Iris for this release
    as well as some long-standing bugs with vertical coordinates and number
    formats. (:pull:`4411`)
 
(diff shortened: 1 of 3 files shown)


akuhnregnier:
Assuming the tests run for this PR, do let me know if you want me to edit the `latest.rst` file again to reflect the updated pull request (or remove it entirely from this PR).

wjbenfold (changes requested):
Still looks good, just like the identical PR did. Having checked the values that came out of this fix, it actually improved the precision of the test result. The tests are now doing a more sane check on the floating point values.
Do please update `latest.rst` to name the right PR, just to avoid confusing anyone

wjbenfold (approved):
Looks good to me. Something slightly odd where it thinks line 120 of the whatsnew has just appeared, but that's probably from force pushing rather than pushing in some way. Will just check that as it gets merged.

trexfeathers (approved):

PR #4428, opened 2021-11-22
Changes for Matplotlib v3.5

rcomer:
### Description

Changes to get tests passing with Matplotlib v3.5.

* New image hash for one gallery example (see https://github.com/SciTools/test-iris-imagehash/pull/56)
* The antialias tests basically check whether `plt.contour` has been called, in addition to `plt.contourf`.  The tests previously relied on the fact that the two functions added different artist types to the axes, but this is no longer the case (https://github.com/matplotlib/matplotlib/issues/20906).  So have updated these tests to use `mock` instead.

---

### DIFF
--- a/docs/src/whatsnew/latest.rst
+++ b/docs/src/whatsnew/latest.rst
@@ -222,6 +222,8 @@ This document explains the changes made to Iris for this release
    :func:`~iris.analysis.cartography.wrap_lons` and updated affected tests
    using assertArrayAllClose following :issue:`3993`.
    (:pull:`4421`)
+   
+#. `@rcomer`_ updated some tests to work with Matplotlib v3.5. (:pull:`4428`)
 
 #. `@rcomer`_ applied minor fixes to some regridding tests. (:pull:`4432`)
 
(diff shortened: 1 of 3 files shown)


rcomer:
See #4424 for failing tests.

rcomer:
The replaced commit on the `auto-update-lockfiles` branch gave me a bunch of conflicts in the lockfiles as I had the original commit (https://github.com/SciTools/iris/commit/774f9cf15f49dd8ee444f8ef6c2b96d5b6115d1e) which no longer exists in the target branch.  I fIxed this by dropping that commit with an interactive rebase before rebasing back onto `upstream/auto-update-lockfiles`.

Do we know if it's possible for the lockfile action to add a new commit to the branch rather than replace the existing commit?  That would make this a little more straightforward I think.  Probably a question for @jamesp.

jamesp:
I think that we included some logic in the GHA to stop the bot from clobbering any additional commits already on the auto-update-lockfiles branch, but not for the workflow you're describing @rcomer.

The workflow creates/updates the PR using this action: https://github.com/peter-evans/create-pull-request

It might be somewhere in the options but I can't recognise it from the descriptions given. We did also consider making a new PR for each run of the action - that would prevent this situation, but possibly result in a profusion of lockfile update PRs if we don't get to merging in each week.

rcomer:
Thanks for looking @jamesp.  At least the new commit didn’t introduce new failures!

trexfeathers (approved):
Looks good, especially the switch to mocking - big improvement. I have re-spun the failed tests to prove that the failed image test should be fine now.

rcomer:
Thanks @trexfeathers!

PR #4432, opened 2021-11-22
Minor fixes for regridding tests

rcomer:
### Description

Running the tests under `pytest`, I got `TypeError`s from these four tests, apparently because `save` expected a string but got an int.  These tests will have been missed by `nose` within the CI because the `__init__.py` was missing.

---


wjbenfold (approved):
Looks good, thanks @rcomer. I hadn't realised we've got our own way to make a tempfile on the test base class or that we needed an `__init__.py`

PR #4434, opened 2021-11-23
Use dask-core instead of dask in ci

aaronspring:
### Description

Tried to reduce dependencies from installing dask via conda which installs like pip install dask[complete]. dask-core is like pip install dask. https://github.com/xgcm/xhistogram/pull/71#discussion_r752738286

Why? dask[complete] includes bokeh etc which are not needed here and likely speed up CI setup/install times

consider also for feedstock https://github.com/conda-forge/iris-feedstock/pull/77

---

### DIFF
--- a/requirements/ci/py37.yml
+++ b/requirements/ci/py37.yml
@@ -13,7 +13,7 @@ dependencies:
   - cartopy >=0.20
   - cf-units >=3
   - cftime >=1.5
-  - dask >=2
+  - dask-core >=2
   - matplotlib
   - netcdf4
   - numpy >=1.19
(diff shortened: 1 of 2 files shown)


aaronspring:
Signed CLA now

lbdreyer:
Thanks for signing the CLA @aaronspring ! I'm not sure why the cla checker hasn't updated but sometimes this can be fixed by closing and reopening the PR. I'm going to do that now to see if that helps!

lbdreyer (changes requested):
Thanks for this @aaronspring ! It would be good to trim the dependencies that get installed.

Could I also ask you to add an entry to the [whats new](https://github.com/SciTools/iris/blob/main/docs/src/whatsnew/latest.rst#-dependencies) summarising this change? There is some guidance in the docs on how to do this (see https://scitools-iris.readthedocs.io/en/stable/developers_guide/documenting/whats_new_contributions.html)

lbdreyer (approved):
Brilliant! Thanks @aaronspring !

PR #4436, opened 2021-11-24
Prevent truncation of cell method descriptions with nested brackets

### LINKED ISSUE #4427: truncated comments in cell_methods
## 🐛 Bug Report
In code that sequentially aggregated observational data it seems that `cell_methods` comments may be truncated under certain circumstances. This is an excerpt from a file produced by iris 3.1.0:

```
 !ncdump -h myfile.nc 
     float tasmin(y, x) ;
        tasmin:standard_name = "air_temperature" ;
        tasmin:long_name = "Daily Minimum Near-Surface Air Temperature" ;
        tasmin:units = "K" ;
        tasmin:cell_methods = "time: minimum within days (comment: 18h(day-1)-18h) time: mean over days time: mean" ;
        tasmin:grid_mapping = "lambert_conformal_conic" ;
        tasmin:coordinates = "height lat lon time" ;
```
Note the parenthesis within the `cell_method` comment.  The cell methods represent 
1.  minimum within day (i.e., 18 UTC the day before to 18 UTC)
2.  mean over days [within the month] of the daily minima
3.  mean over months [within the year] of the daily minima

where the aim is to compact 2 and 3 to 
2.  mean over days [within the year] of the daily minima

However, this is how it looks when reading this file into iris:
```
 air_temperature / (K)                                  (projection_y_coordinate: 823; projection_x_coordinate: 567)
    ....
    Cell methods:
        minimum within days                              time (18h(day-1)
        mean over days                                   time
        mean                                             time

### PR DESCRIPTION
**Pull request by wjbenfold**
### Description

Improves the parsing of netcdf cell methods to not rely solely on regular expressions and thus fixes #4427. Initially created a broken test using the example from #4427 and then fixed it.

Questions for reviewer:
* Is there a canonical list of what can make up cell methods (e.g. whether nested brackets are allowed) that could be used to check this code works on them all?
* The previous code would fail silently on a cell method that didn't match the regex whereas now it's possible to identify strings that should have given a cell method but didn't match. Should these be warnings or exceptions?

To do:

- [ ] Warnings when a cell method can't be parsed
- [ ] What's new

---

### CODE REVIEW

**Comment by rcomer** [comment]
> Should these be warnings or exceptions?

I do not think an exception would be helpful here, as (as far as I know) there would be no way for the user to get around it within Iris.  All they could do is find some other tool to modify the NetCDF file before attempting to load it again.

**Comment by larsbarring** [comment]
Regarding your first point, on a canonical list, I think there is none. In CF 1.9 section [7.3.2 Recording the spacing of the original data and other information](https://cfconventions.org/Data/cf-conventions/cf-conventions-1.9/cf-conventions.html#recording-spacing-original-data) details are given how to combine standardized elements and non-standardized elements. But I did not find any guidance as to what can or can not go into the non-standardized part.

Will your PR handle nested brackets and other more or less difficult (in python sense) strings? If not it might be useful to have a check of the string before writing to file in the first case (cf. my comment in #4427).

**Comment by wjbenfold** [comment]
> Will your PR handle nested brackets and other more or less difficult (in python sense) strings?

It currently has to assume that you've closed all the brackets you've opened, otherwise there's no way to distinguish comment content from actual new methods.

The only way I can think of that Iris could check if it can interpret a cell method string is to try reading it once it's formed it, which is basically what happens when you attempt the load again, so I'm not sure warning then is any better than warning on load? I don't think Iris should do anything stronger than just warning because it can't read it back, because you might really want to write that for some other program.

**Comment by larsbarring** [comment]
> It currently has to assume that you've closed all the brackets you've opened, otherwise there's no way to distinguish comment content from actual new methods.

Fair enough :+1: 

> The only way I can think of that Iris could check if it can interpret a cell method string is to try reading it once it's formed it, which is basically what happens when you attempt the load again, so I'm not sure warning then is any better than warning on load? I don't think Iris should do anything stronger than just warning because it can't read it back, because you might really want to write that for some other program.

Warning is perfect :+1: 

Thanks for taking this on so quickly!

**Comment by wjbenfold** [comment]
Any advice on better use of warnings, mocking and patching warmly welcomed

File: lib/iris/tests/unit/fileformats/netcdf/test_parse_cell_methods.py [hunk 1]
```diff
@@ -85,6 +99,53 @@ def test_comment(self):
...
+            res = parse_cell_methods(cell_method_str)
+            self.assertEqual(res, expected)
+
+    def test_comment_bracket_mismatch_warning(self):
+        cell_method_strings = [
+            "time: minimum within days (comment: 18h day-1)-18h)",
+            "time : minimum within days (comment: 18h day-1)-18h)",
+        ]
+        for cell_method_str in cell_method_strings:
+            with mock.patch("warnings.warn") as warn:
```
rcomer, line 125: Would `self.assertWarnsRegexp` do the job here?

https://github.com/SciTools/iris/blob/91b01be11ebd439668c3f140f66350b9e65db59e/lib/iris/tests/__init__.py#L601-L609

wjbenfold: I'd gone for copying the code below what I'd inserted (below) but don't know if one of these is more up-to-date or preferred. I think either works?

https://github.com/SciTools/iris/blob/a7869293fb8a94765a2508ff126dfd1fd3f42679/lib/iris/tests/unit/fileformats/netcdf/test_parse_cell_methods.py#L187-L193

rcomer: Looking at the git blame I _think_ the `assertWarnsRegexp` is newer.  I generally prefer to have a single function/method that is called in multiple places, rather than copy/paste.  That way, if I think of a better way to do it later, I only need to change it in one place to get the improvement all over.

I would defer to alternative advice from anyone in your team though!

wjbenfold: Turns out assertWarns takes a msg argument, so I've used that in both places. Thanks for the prompt to dig into this!

rcomer: Looks like [assertWarns and assertWarnsRegex were introduced into unittest at python3.2](https://docs.python.org/3/library/unittest.html#unittest.TestCase.assertWarns).  Our own `assertWarnsRegex` was presumably written when we still supported python2.7, and is now redundant 🤔

wjbenfold: Ah that explains why assertWarnsRegex and assertWarnsRegexp were both showing up as options in VSCode

wjbenfold:
@rcomer apart from a whatsnew, which I'd pop in a separate PR to avoid conflicts, is there anything more to do here?

rcomer:
@wjbenfold I'm afraid I've not reviewed the bulk of this.  The `re` package is completely alien to me!

jamesp (approved):
Discussed with @wjbenfold, looks good, LGTM!

larsbarring:
:+1: :+1: :+1:

PR #4437, opened 2021-11-25
Mesh recombine

pp-mo:
First-version for "recombining" sub-mesh information.

It's worth noting that supporting location-index-sets would make this easier, as then the sub-region cubes would simply be referred to a common mesh+location, and the awkward 'special index coords' would not be required.

Without that, though, various alternative APIs are feasible + I could still be persuaded to change.
In particular, we _could_ replace 'full_mesh_cube' with 'mesh' + 'location' args -- it's effectively more logical, as we can deduce everything else from the region cubes.
So, that might be slightly neater, but more work is required .  Where elements could map both mesh and non-mesh dims, it could get complicated.

TODOs:
   - [x] tests for all the consistency checks
   - [x] tests for other features + functions
      - [x] transposed forms
      - [x] zero + multiple non-mesh (structured) dimensions
      - [x] dtypes
      - [x] mask behaviours
      - [x] NaNs
   - [x] tests for lazy behaviour
   - [ ] benchmarks showing sensible use of memory in lazy evaluation (as this is a known practical requirement)

### DIFF
--- a/lib/iris/experimental/ugrid/utils.py
+++ b/lib/iris/experimental/ugrid/utils.py
@@ -8,7 +8,7 @@
 Utility operations specific to unstructured data.
 
 """
-from typing import AnyStr, Iterable
+from typing import AnyStr, Iterable, Union
 
 import dask.array as da
 import numpy as np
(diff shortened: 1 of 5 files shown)


lib/iris/experimental/ugrid/utils.py
```diff
@@ -0,0 +1,322 @@
...
+                        f"{reg_cube_str} has an index coord "
+                        f'"{index_coord_name}" whose ".metadata" does not '
+                        "match that on 'full_mesh_cube' :  "
+                        f"{regdim_metadata} != {fulldim_metadata}."
+                    )
+
+                # At this point, we know we *have* an index coord, and it does not
+                # conflict with the one on 'full_mesh_cube' (if any).
+                # Now check for matches between the region cubes.
+                if indexcoord_metadata is None:
```
bjlittle, line 205: @pp-mo You only want to perform this `if ... elif` block when `err is None`

lib/iris/experimental/ugrid/utils.py
```diff
@@ -0,0 +1,322 @@
...
+    else:
+        dim_range = np.arange(full_mesh_cube.ndim, dtype=int)
+        # Transpose all inputs to mesh-last order.
+        tranpose_dims = [i_dim for i_dim in dim_range if i_dim != mesh_dim] + [
+            mesh_dim
+        ]  # chop out mesh_dim + put it at the end.
+
+        def transposed_copy(cube, dim_order):
+            cube = cube.copy()
+            cube.transpose()
```
bjlittle, line 234: @pp-mo Hmm I think that you've forgotten to use the `dim_order` to transpose :fire:

i.e., `cube.transpose(dim_order)`

lib/iris/experimental/ugrid/utils.py
```diff
@@ -0,0 +1,322 @@
...
+            datarr,
+            indarr,
+            dtype=result_array.dtype,
+            meta=np.ndarray,
+        )
+
+    # Construct the result cube.
+    result_cube = full_mesh_cube.copy()
+    result_cube.data = result_array
+    # Copy names, units + attributes from region data (N.B. but not var_name)
```
bjlittle, line 316: @pp-mo If you're convinced that it's the right thing to do to clear the `var_name`, then you should definitely highlight this to the user in the doc-string :+1:

lib/iris/experimental/ugrid/utils.py
```diff
@@ -0,0 +1,322 @@
...
+    # Notes on resultant calculation properties:
+    # 1. map_blocks is chunk-mapped, so it is parallelisable and space-saving
+    # 2. However, fetching less than a whole chunk is not efficient
+    for region_cube in region_cubes:
+        # Lazy data array from the region cube
+        datarr = region_cube.lazy_data()
+
+        # Lazy indices from the mesh-dim coord.
+        mesh_dimcoord = region_cube.coord(
+            name_or_coord=index_coord_name, dimensions=region_cube.ndim - 1
```
bjlittle, line 293: @pp-mo Minor, but you could have something like an `rc_mesh_dim = region_cube.ndim - 1`, as you use it a couple of times in this loop i.e., `line+299` also

lib/iris/experimental/ugrid/utils.py
```diff
@@ -0,0 +1,322 @@
...
+        # Apply the operation to paste from one region into the target.
+        # N.B. replacing 'result_array' each time around the loop.
+        result_array = da.map_blocks(
+            fill_region,
+            result_array,
+            datarr,
+            indarr,
+            dtype=result_array.dtype,
+            meta=np.ndarray,
+        )
```
bjlittle, line 311: @pp-mo Do we need to care about `result_array` and `datarr` having the same chunks?

bjlittle (changes requested):
@pp-mo Looks good... still need to go through the tests, but I can sweep through them later when they're all there :+1:

lib/iris/experimental/ugrid/utils.py
```diff
@@ -0,0 +1,322 @@
...
+"""
+from typing import AnyStr, Iterable
+
+import dask.array as da
+import numpy as np
+
+from iris.cube import Cube
+
+
+def recombine_regions(
```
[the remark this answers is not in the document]
pp-mo: I don't much like 'region_cube' for the big one, it sounds to me like it is already "only a part of something".
How about 'mesh_cube' + 'submesh_cubes' ?

lib/iris/experimental/ugrid/utils.py
```diff
@@ -0,0 +1,322 @@
...
+        same mesh and location, but with its data and ".metadata" replaced by
+        that from the 'region_cubes'.
+        Where regions overlap, the result data comes from the last-listed of the
+        original region cubes which contain that location.
+        Where no region contains a datapoint, it will be masked in the result.
+        HINT: alternatively, values covered by no region can be taken from the
+        original 'full_mesh_cube' data, if 'full_mesh_cube' is *also* passed
+        as the first of the 'region_cubes'.
+
+    """
```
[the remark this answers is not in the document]
pp-mo: Ok, I was going to ask you what standards we now wanted
:+1: will do

lib/iris/experimental/ugrid/utils.py
```diff
@@ -0,0 +1,322 @@
...
+
+    def metadata_no_varname(cube_or_coord):
+        # Get a metadata object but omit any var_name.
+        metadata = cube_or_coord.metadata
+        fields = metadata._asdict()
+        fields["var_name"] = None
+        result = metadata.__class__(**fields)
+        return result
+
+    n_regions = len(region_cubes)
```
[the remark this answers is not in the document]
pp-mo: :+1: will do

lib/iris/experimental/ugrid/utils.py
```diff
@@ -0,0 +1,322 @@
...
+    n_dims = full_mesh_cube.ndim
+    regioncube_metadata = None
+    indexcoord_metadata = None
+    for i_region, region_cube in enumerate(region_cubes):
+        reg_cube_str = (
+            f'Region cube #{i_region}/{n_regions}, "{region_cube.name()}"'
```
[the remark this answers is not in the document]
pp-mo: :+1: good spot

lib/iris/experimental/ugrid/utils.py
```diff
@@ -0,0 +1,322 @@
...
+    # Perform consistency checks on all the region-cubes.
+    #
+
+    def metadata_no_varname(cube_or_coord):
+        # Get a metadata object but omit any var_name.
+        metadata = cube_or_coord.metadata
+        fields = metadata._asdict()
+        fields["var_name"] = None
+        result = metadata.__class__(**fields)
+        return result
```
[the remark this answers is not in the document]
pp-mo: :+1: ok I just failed to find find the necessary metadata method !

lib/iris/experimental/ugrid/utils.py
```diff
@@ -0,0 +1,322 @@
...
+        # Check dimensionality.
+        if reg_ndims != n_dims:
+            err = (
+                f"{reg_cube_str} has {reg_ndims} dimensions, but "
+                f"'full_mesh_cube' has {n_dims}."
+            )
+            raise ValueError(err)
+
+        # Get region_cube metadata, which will apply to the result..
+        region_cube_metadata = metadata_no_varname(region_cube)
```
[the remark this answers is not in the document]
pp-mo: The problem is that one is the metadata of **this** region_cube, while the other is the 'common' value which **all** the cubes are "supposed to" have.  I'll find a better naming scheme

lib/iris/experimental/ugrid/utils.py
```diff
@@ -0,0 +1,322 @@
...
+                    )
+
+                # At this point, we know we *have* an index coord, and it does not
+                # conflict with the one on 'full_mesh_cube' (if any).
+                # Now check for matches between the region cubes.
+                if indexcoord_metadata is None:
+                    # Store first occurrence (from first region-cube)
+                    indexcoord_metadata = regdim_metadata
+                elif regdim_metadata != indexcoord_metadata:
+                    # Compare subsequent occurences (from other region-cubes)
```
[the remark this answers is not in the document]
pp-mo: :+1:

lib/iris/experimental/ugrid/utils.py
```diff
@@ -0,0 +1,322 @@
...
+        # Get region_cube metadata, which will apply to the result..
+        region_cube_metadata = metadata_no_varname(region_cube)
+        if regioncube_metadata is None:
+            # Store the first region-cube metadata as a reference
+            regioncube_metadata = region_cube_metadata
+        elif region_cube_metadata != regioncube_metadata:
+            # Check subsequent region-cubes metadata against the first.
+            err = (
+                f"{reg_cube_str} has metadata {region_cube_metadata}, "
+                "which does not match that of the first region region_cube, "
```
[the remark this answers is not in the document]
pp-mo: :+1:

lib/iris/experimental/ugrid/utils.py
```diff
@@ -0,0 +1,322 @@
...
+                # Now check for matches between the region cubes.
+                if indexcoord_metadata is None:
+                    # Store first occurrence (from first region-cube)
+                    indexcoord_metadata = regdim_metadata
+                elif regdim_metadata != indexcoord_metadata:
+                    # Compare subsequent occurences (from other region-cubes)
+                    err = (
+                        f"{reg_cube_str} has an index coord "
+                        f'"{index_coord_name}" whose ".metadata" does not '
+                        f"match that of the first region-cube :  "
```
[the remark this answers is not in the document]
pp-mo: :+1:

lib/iris/experimental/ugrid/utils.py
```diff
@@ -0,0 +1,322 @@
...
+                        f"{regdim_metadata} != {indexcoord_metadata}."
+                    )
+
+        if err:
+            raise ValueError(err)
+
+    # Use the mesh_dim to transpose inputs + outputs, if required, as it is
+    # simpler for all the array operations to always have the mesh dim *last*.
+    if mesh_dim == full_mesh_cube.ndim - 1:
+        # Mesh dim is already the last one : no tranposes required
```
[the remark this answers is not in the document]
pp-mo: :+1:

lib/iris/experimental/ugrid/utils.py
pp-mo: Oops ! :+1:
Not yet tested it , of course..

lib/iris/experimental/ugrid/utils.py
pp-mo: Ok, but it does make for tough reading :grimacing: 

BTW for me this is really highlighting, once again, why 'var_name' should never have been a part of Iris object identity !

lib/iris/experimental/ugrid/utils.py
pp-mo: Ok! :+1:

lib/iris/experimental/ugrid/utils.py
pp-mo: I don't know whether this code is actually needed (or if it happens automatically), but I think we definitely *do* need this as an outcome, to get efficient calculation in Dask -- so we get control over the chunking of the result.

I'm can't be  sure though, as as I've still not got a successfully performance demonstration of this, but I'm working on it ...

pp-mo:
A point that just occurred to me..
I have realised that the indices could be checked against the original mesh ~fairly cheaply.
I.E. as everything has coords, we could test that each submesh cube has 
`all(submesh.coord(xyname).points == meshcube.coord(xyname)[submesh.indices].points)`

It would be sort-of logical to do this, to test the user contract that the submesh-s 'belong' to the original mesh.
But I'm thinking we don't have much enthusiasm for this, even though the runtime cost is probably not very high ??

lib/iris/experimental/ugrid/utils.py
```diff
@@ -0,0 +1,322 @@
...
+
+        # Check dimensionality.
+        if reg_ndims != n_dims:
+            err = (
+                f"{reg_cube_str} has {reg_ndims} dimensions, but "
+                f"'full_mesh_cube' has {n_dims}."
+            )
+            raise ValueError(err)
+
+        # Get region_cube metadata, which will apply to the result..
```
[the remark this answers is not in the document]
pp-mo: :+1:

lib/iris/experimental/ugrid/utils.py
```diff
@@ -0,0 +1,322 @@
...
+                        f"the mesh dimension (dimension {mesh_dim})."
+                    )
+                elif fulldim and regdim_metadata != fulldim_metadata:
+                    # May *not* have full-cube index, but if so it must match
+                    err = (
+                        f"{reg_cube_str} has an index coord "
+                        f'"{index_coord_name}" whose ".metadata" does not '
+                        "match that on 'full_mesh_cube' :  "
+                        f"{regdim_metadata} != {fulldim_metadata}."
+                    )
```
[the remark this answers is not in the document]
pp-mo: No I don't think there's anything to check in that case.
We allow the `mesh_cube` to not have an index coord (since it would have trival values).
But in that case, we can _only_ check this coord it against the other index-coords -- which is what happens next.

lib/iris/experimental/ugrid/utils.py
```diff
@@ -0,0 +1,322 @@
...
+        # Get a metadata object but omit any var_name.
+        metadata = cube_or_coord.metadata
+        fields = metadata._asdict()
+        fields["var_name"] = None
+        result = metadata.__class__(**fields)
+        return result
+
+    n_regions = len(region_cubes)
+    n_dims = full_mesh_cube.ndim
+    regioncube_metadata = None
```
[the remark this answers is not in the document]
pp-mo: For ref, here's what I wound up with :

|  original  |  suggested  |  modified  |
|  regioncube_metadata  |  rc_metadata  |  metadata  |
|  reg_cube_str  |  rc_str  |  sub_str  |
|  reg_ndims  |  rc_ndims  |  (gone)  |

    fulldim -> full_coord
    full_dimname -> full_coord_name
    fulldim_metadata -> full_coord_metadata
    regdim -> rc_coord
    reg_dimname -> rc_coord_name
    regdim_metadata -> rc_coord_metadata
    datarr -> rc_data

```

File: lib/iris/experimental/ugrid/utils.py (hunk 1 above)
Comment (pp-mo) [reply lib/iris/experimental/ugrid/utils.py:205]: :+1: nice !

File: lib/iris/experimental/ugrid/utils.py (hunk 4 above)
Comment (pp-mo) [reply lib/iris/experimental/ugrid/utils.py:293]: Actually that's a bit horrible, because in this usage, we are using "cube.ndim - 1" as the "index of the last data dimension".
But in the next case below, we are using "cube.ndim - 1" as the "number of data dimensions, minus one", which we multiply by to get a (1,1,1,n) tuple.
So if you name this as 'mesh_dimension' (effectively, an index), then _multipying_ by it does look very odd !

File: lib/iris/experimental/ugrid/utils.py [hunk 19]
```diff
@@ -0,0 +1,322 @@
...
+
+    """
+    if not region_cubes:
+        raise ValueError("'region_cubes' must be non-empty.")
+
+    mesh_dim = full_mesh_cube.mesh_dim()
+    if mesh_dim is None:
+        raise ValueError("'full_mesh_cube' has no \".mesh\".")
+
+    # Check the basic required properties of the input.
```
pp-mo, line 80: NOTE: this is not wanted, since I changed my mind about the 'full_mesh' having to have a dim-coord on the mesh.
Now removed :

pp-mo:
Ok, some rather noisy changes there.
I think I covered everything.
Please re-review @bjlittle

lib/iris/experimental/ugrid/utils.py
```diff
@@ -0,0 +1,321 @@
...
+
+    # Create an initially 'empty' (all-masked) dask array matching the input.
+    # N.B. this does not use the mesh_cube.lazy_data() array, but only its
+    # shape and dtype, since the data itself is not used in the calculation.
+    # N.B. chunking matches the input cube, allowing performance control.
+    input_data = mesh_cube.lazy_data()
+    result_array = da.ma.masked_array(
+        da.zeros(
+            input_data.shape,
+            dtype=input_data.dtype,
```
pp-mo, line 273: NOTE: just spotted, this should be "result_dtype".
To be fixed in forthcoming commits, with added testcases

bjlittle: @pp-mo I don't think this matters :thinking: 

At this point we know that all the `dtypes` are equal between the `mesh_cube` and the `submesh_cubes`, right?

pp-mo: No, that's not required.
The submesh data has it's own phenomenon type, e.g. name + units,  which could be different from 'mesh_cube' as that's only a template (logically, just for the mesh itself, but we fudged that point).
So if it  has its own name+units, it makes sense to include dtype with that -- e.g. it could be a category number of a classification, or a COUNT statistic

pp-mo: So.. I just added a test where the submesh_cube info is all ints.

pp-mo:
Update: added the rest of the tests.

I think this is now GTG

bjlittle (approved):
@pp-mo This is a lovely contributions, thanks :beers:

PR #4439, opened 2021-11-26
Mesh full comparison

trexfeathers:
### Description

Allows checking of the equality of two `Mesh`es, fully inspecting their `Coord` and `Connectivity` membership.

Due to the way `MeshCoord.__eq__` is written, this also allows us to check the equality of two `Cube`s with `Mesh`es attached.

Acceptable performance also required making the equality of `_DimensionalMetadata` and `Connectivity` preserve array laziness.

---

### DIFF
--- a/lib/iris/tests/unit/experimental/ugrid/mesh/test_Mesh.py
+++ b/lib/iris/tests/unit/experimental/ugrid/mesh/test_Mesh.py
@@ -132,9 +132,10 @@ class TestProperties1D(TestMeshCommon):
         self.assertEqual(equivalent, self.mesh)
 
     def test_different(self):
-        # 2021-11-26 currently not possible to have a metadata-only difference
-        #  - only topology_dimension makes a difference and that also mandates
-        #  different connectivities.
+        different_kwargs = self.kwargs.copy()
+        different_kwargs["long_name"] = "new_name"
+        different = mesh.Mesh(**different_kwargs)
+        self.assertNotEqual(different, self.mesh)
 
         different_kwargs = self.kwargs.copy()
         ncaa = self.kwargs["node_coords_and_axes"]
--- a/lib/iris/tests/unit/experimental/ugrid/mesh/test_Mesh.py
+++ b/lib/iris/tests/unit/experimental/ugrid/mesh/test_Mesh.py
@@ -81,7 +81,7 @@ class TestProperties1D(TestMeshCommon):
         cls.kwargs = {
             "topology_dimension": 1,
             "node_coords_and_axes": ((cls.NODE_LON, "x"), (cls.NODE_LAT, "y")),
-            "connectivities": cls.EDGE_NODE,
+            "connectivities": [cls.EDGE_NODE],
             "long_name": "my_topology_mesh",
             "var_name": "mesh",
             "attributes": {"notes": "this is a test"},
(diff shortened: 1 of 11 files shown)


lib/iris/tests/unit/experimental/ugrid/mesh/test_Mesh.py
```diff
@@ -124,6 +124,33 @@ def test___repr__(self):
...
+        # The dimension names do not participate in equality.
+        equivalent_kwargs = self.kwargs.copy()
+        equivalent_kwargs["node_dimension"] = "something_else"
+        equivalent = mesh.Mesh(**equivalent_kwargs)
+        self.assertEqual(equivalent, self.mesh)
+
+    def test_different(self):
+        # 2021-11-26 currently not possible to have a metadata-only difference
+        #  - only topology_dimension makes a difference and that also mandates
+        #  different connectivities.
```
stephenworsley, line 137: I'm not sure what you mean by this. Could you elaborate?

trexfeathers: I would ideally have also tested that a difference would be detected if ONLY the metadata were different and the `Coord`s and `Connectivity`s were identical. But this isn't possible since the only possible metadata difference is in `topology_dimension`, and different `topology_dimension`s require different `Connectivity`s.

Does that help? Do you have any ideas for a phrasing that's easier to understand?

stephenworsley: So do you mean the metadata of the Mesh itself? Because I imagine testing for a different `start_index` for example could also be worthwhile, which is in the `Connectivity` metadata.

trexfeathers: >So do you mean the metadata of the Mesh itself?

Yes I specifically mean the contents of the `MeshMetadata` instance attached to the `Mesh`.

---

>Because I imagine testing for a different `start_index` for example could also be worthwhile, which is in the `Connectivity` metadata.

That certainly would qualify as a difference, but such a check is deferred to `Connectivity.__eq__()` so from the point of view of this test it's no different from the existing lines:

https://github.com/SciTools/iris/blob/3d0d0aa00c9f2ac6723dcbd8fab3d57597fb25f2/lib/iris/tests/unit/experimental/ugrid/mesh/test_Mesh.py#L147-L152

stephenworsley: That makes sense. In which case I think it's worth expanding the test in the connectivity equality check that makes sure it is able to handle metadata in equality properly. https://github.com/SciTools/iris/blob/b03aa7c16cba7e4947c93301e99494736c0bd977/lib/iris/tests/unit/experimental/ugrid/mesh/test_Connectivity.py#L129-L133

trexfeathers: Found a way to have a metadata difference: e5b1d2385c889acd8151aca808fcf1c895f32fdf

We have agreed that the missing `Connectivity` test coverage mentioned above is beyond the scope of this PR.

stephenworsley (approved):
Looks good  to merge.

PR #4450, opened 2021-12-06
test__hybrid_formulae.py nose discovery

### LINKED ISSUE #4431: `nose` missing some tests
## 📰 Custom Issue

We have some tests for NetCDF loading that are created by using `setattr` on the test class:

https://github.com/SciTools/iris/blob/0a0cb8722153440a4db2a7bb53ba3998a507cd2f/lib/iris/tests/unit/fileformats/nc_load_rules/actions/test__hybrid_formulae.py#L242-L278

`nose` does not seem to be discovering these.  They run fine if you run the module directly, and they are also discovered by `pytest`.  So this issue would be solved by implementing #3835 🙂

wjbenfold:
### Description

Fixes #4431 which highlights that the method used to construct some tests in `iris/tests/unit/fileformats/nc_load_rules/actions/test__hybrid_formulae.py` wasn't allowing them to be picked up by `nose`.  I've expanded the short method that was being used to make the tests into a repetitive setup that should be picked up by `nose`, hopefully exactly replicating the previous tests.

---

### DIFF
--- a/lib/iris/tests/unit/fileformats/nc_load_rules/actions/test__hybrid_formulae.py
+++ b/lib/iris/tests/unit/fileformats/nc_load_rules/actions/test__hybrid_formulae.py
@@ -238,44 +238,75 @@ variables:
             formula_terms=["a", "b", "depth", "eta", "orog", "sigma"],
         )
 
-
-# Add in tests methods to exercise each (supported) vertical coordinate type
-# individually.
-# NOTE: hh.CF_COORD_VERTICAL lists all the valid types, but we don't yet
-# support all of them.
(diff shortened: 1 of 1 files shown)


rcomer (approved):
Thanks @wjbenfold!

Confirmed that the CI is now finding 5977 tests, whereas the[ last commit to main](https://github.com/SciTools/iris/runs/4429591097) only found 5970.

I have convinced myself that the new tests are equivalent to the generated ones, and that each of the new tests is checking a different `hybrid_type`.

LGTM

rcomer:
Oh, it should probably have a whatsnew though!

PR #4457, opened 2021-12-10
Fix array_equal behaviour for masked arrays

stephenworsley:
Current behaviour will compare the underlying data of a masked array and will sometimes compare masked arrays as unequal when they are otherwise equal for all unmasked points. This would cause an error in an iris-esmf-regrid PR as described here:
https://github.com/SciTools-incubator/iris-esmf-regrid/pull/138#issuecomment-991099896

### DIFF
--- a/lib/iris/util.py
+++ b/lib/iris/util.py
@@ -363,7 +363,8 @@ def array_equal(array1, array2, withnans=False):
 
     def normalise_array(array):
         if not is_lazy_data(array):
-            array = np.asarray(array)
+            if not ma.isMaskedArray(array):
+                array = np.asarray(array)
         return array
 
     array1, array2 = normalise_array(array1), normalise_array(array2)


rcomer:
Seems like ignoring the masks was deliberate, but only because “the chosen approach fixes the problem [of all-masked data raising errors], is simpler, and reflects numpy behaviour.” #905

Perhaps using something like [ma.filled](https://numpy.org/doc/stable/reference/generated/numpy.ma.filled.html) would help?

bjlittle:
Related to https://github.com/numpy/numpy/pull/16022#issuecomment-1174794735, which was merged into `numpy`, but then reverted (as it broke `scipy` and `astropy`)

lib/iris/util.py
```diff
@@ -363,7 +363,8 @@ def array_equal(array1, array2, withnans=False):
 
     def normalise_array(array):
         if not is_lazy_data(array):
-            array = np.asarray(array)
+            if not ma.isMaskedArray(array):
+                array = np.asarray(array)
```
pp-mo, line 367: Would `np.asanyarray` maybe be a better approach  here ?
I think it would achieve the same thing, and maybe a more standard approach

rcomer:
There is also a relevant thread at https://github.com/numpy/numpy/issues/14624

I think what Iris does in testing is sensible:

https://github.com/SciTools/iris/blob/56a97d66b8404f96ef61fe2d52433eb58373f338/lib/iris/tests/__init__.py#L590-L668

stephenworsley:
> There is also a relevant thread at [numpy/numpy#14624](https://github.com/numpy/numpy/issues/14624)

@rcomer It's interesting that there is an inconsistency between different numpy functions at the moment, though given that currently `iris.util.array_equal` behaves in the same way as `np.array_equal`, that could be enough of a reason to keep this function behaving the same. There's still potentially a question of how we use `iris.util.array_equal` within iris. It might make sense in places (such ass coord equality) to use an approach of calling `ma.filled` as you suggest before calling `iris.util.array_equal` and also checking that the mask is equal.

pp-mo:
As discussed @trexfeathers @stephenworsley @bjlittle @pp-mo  ...
This is probably still relevant
Causes problem here : https://github.com/SciTools-incubator/iris-esmf-regrid/blob/main/esmf_regrid/schemes.py#L1348-L1366

lib/iris/util.py
```diff
@@ -406,17 +406,24 @@ def array_equal(array1, array2, withnans=False):
 
     def normalise_array(array):
         if not is_lazy_data(array):
-            array = np.asarray(array)
+            if not ma.isMaskedArray(array):
```
pp-mo, line 409: I think that the behaviour changes to this function can reasonably be considered as "fixes" and not "breaking change".

But, can you please add something to the docstring to make the operation on masked arrays explicit ?
The existing docstring simply doesn't cover it, and suggests it behaves like `np.array_equal`, which is not really true any more, at least in this respect :
```
>>> np.array_equal([5, 6], np.ma.masked_array([5, 6], mask=[0,1]))
True
```

So, I think the whats-new comment is totally clear about the masked behaviour, but the method docstring is not (!)

lib/iris/tests/unit/util/test_array_equal.py
```diff
@@ -44,9 +44,14 @@ def test_nd(self):
         self.assertTrue(array_equal(array_a, array_b))
         self.assertFalse(array_equal(array_a, array_c))
 
-    def test_masked_is_ignored(self):
+    def test_masked_is_not_ignored(self):
         array_a = ma.masked_array([1, 2, 3], mask=[1, 0, 1])
         array_b = ma.masked_array([2, 2, 2], mask=[1, 0, 1])
+        self.assertTrue(array_equal(array_a, array_b))
+
+    def test_masked_is_different(self):
```
pp-mo, line 52: This is good stuff, but I'd like us also to be clear what happens when you compare masked arrays with "regular" ones.
E.G. :
   * array([1,0]) == ma.masked_array([1, 0])
   * array([1,0]) == ma.masked_array([1, 0], mask=[0, 0])  # (not necessarily the same as mask=False)
   * array([1,0]) != ma.masked_array([1, 0], mask=[0, 1])

lib/iris/tests/unit/util/test_array_equal.py
```diff
@@ -54,6 +54,18 @@ def test_masked_is_different(self):
         array_b = ma.masked_array([1, 2, 2], mask=[0, 0, 1])
         self.assertFalse(array_equal(array_a, array_b))
```
pp-mo, line 55: Sorry I just spotted another thing : I'm not sure this case is now adding anything useful, since `tset_masked_isnt_unmasked` is already ensuring masks **_must_** be equal (even when all the data is equal).

So, unless I'm missing something, you could just remove this testcase ?

pp-mo (changes requested):
All done. apart from the extra point I just added.

lib/iris/tests/unit/util/test_array_equal.py
stephenworsley: The one case which this covers but other tests doesn't is when precisely:

- Both cubes have masks.
- Those cubes are defferent.

`test_masked_isnt_unmasked` tests the case where a mask on one cube is different to the mask which is implied on another cube _without_ an explicit mask. There's maybe a case that this justifies another test even if it's a very narrow case. I'm happy to remove this if you think it's overkill though.

pp-mo: OK, I can see we're making the masks explicit in this case.

lib/iris/tests/unit/util/test_array_equal.py
```diff
@@ -44,11 +44,28 @@ def test_nd(self):
...
 
-    def test_masked_is_ignored(self):
+    def test_masked_is_not_ignored(self):
         array_a = ma.masked_array([1, 2, 3], mask=[1, 0, 1])
         array_b = ma.masked_array([2, 2, 2], mask=[1, 0, 1])
+        self.assertTrue(array_equal(array_a, array_b))
+
+    def test_masked_is_different(self):
+        array_a = ma.masked_array([1, 2, 3], mask=[1, 0, 1])
+        array_b = ma.masked_array([1, 2, 2], mask=[0, 0, 1])
```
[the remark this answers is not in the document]
stephenworsley: yeah, that seems fair.

pp-mo:
Great, thanks for your patience @stephenworsley !

pp-mo (approved):

Issue #4469, opened 2021-12-20
Review changed netcdf-load behaviours, due to replacement of Pyke rules


pp-mo:
In the course of #4470 , I found that there are some worrying breaking changes in behaviour, which were introduced by #4198 

I now think that within #4198 I may have oversimplified the logic and behaviours of the old Pyke rules, and that should probably be reviewed.

For a notable example,the recognition of a latitude/longitude dimcoord **_used_** to behave differently depending on whether or not there is a grid-mapping, but the new code doesn't allow for that :  which may well be a "breaking change too far".

The test harness added in #4470 only addresses the specifics of how latitude/longitude dim-coords are recognised and built.
However, even within that small scope I found that the behaviour depends on at least four separate factors (from file content), as shown in the [principal arguments to the `_make_testcase_cdl` routine](https://github.com/SciTools/iris/blob/4f9794727284b9f78a95cac7b4e808ee5666f52c/lib/iris/tests/unit/fileformats/nc_load_rules/actions/test__latlon_dimcoords.py#L43-L51) : (standard-name, units, axis, grid_mapping)

I [identified some likely+significant values of these four factors](https://github.com/pp-mo/iris/blob/nc_load_latlon_further_RETRO/lib/iris/tests/unit/fileformats/test__latlon_dimcoords_STANDALONE.py#L336-L352), and [tested those](https://github.com/pp-mo/iris/blob/nc_load_latlon_further_RETRO/lib/iris/tests/unit/fileformats/test__latlon_dimcoords_STANDALONE.py#L256-L260) in a [matrix of all possible combinations](https://github.com/pp-mo/iris/blob/nc_load_latlon_further_RETRO/lib/iris/tests/unit/fileformats/test__latlon_dimcoords_STANDALONE.py#L384-L387).
As you might expect, this throws up a whole lot more differences between the old + new behaviour.  
I then made an attempt to [summarise the older behaviour](https://github.com/pp-mo/iris/blob/nc_load_latlon_further_RETRO/lib/iris/tests/unit/fileformats/test__latlon_dimcoords_STANDALONE.py#L266-L334), but this proved overly difficult, and begs the question as to **_whether this is really what we really want_** (i.e. all this, and only this).  
So I'm now thinking that the future of this is to decide on more sensible behaviour, and then code tests for that.

SO... 
I think the way forward is to cautiously re-review this, in a reasonably comprehensive way.
[...]

PR #4470, opened 2021-12-20
Nc load latlon fix

### LINKED ISSUE #4460: Error reading NetCDF file dimensions with units degrees
## 🐛 Bug Report
(Edited to show the error is associated with the coordinates having units 'degrees' rather than no units.)
I can create a NetCDF file of a field with 'longitude' and 'latitude' coordinates with units 'degrees' using the netCDF4 module version 1.5.8. Its metadata appears correct according to ncdump and the data it contains looks correct according to ncview. When I read the file using iris 3.1.0, however, it assigns the name 'grid_latitude' to both coordinates.

## How To Reproduce

Creating the file, reading it in with iris and printing the cube:
```
import numpy as np
import iris
import netCDF4

lat_vals=np.arange(-90,90)
lon_vals=np.arange(360)
data_out=np.random.normal(0,1,(len(lat_vals),len(lon_vals)))

fname='test.nc'
f = netCDF4.Dataset(fname, 'w')
lat_dim=f.createDimension('latitude', len(lat_vals))
lats=f.createVariable('latitude', 'f4', ('latitude',))
lats[:]=lat_vals
lon_dim=f.createDimension('longitude', len(lon_vals))
lons=f.createVariable('longitude', 'f4', ('longitude',))
lons[:]=lon_vals
data=f.createVariable('data', 'f4', ('latitude','longitude'))
data[:]=data_out

#Commenting these next two lines makes iris read the file correctly
lats.units='degrees'
lons.units='degrees'

f.close()

cube=iris.load_cube(fname)
print(cube)
```

This gives the result
```
data / (unknown)                    (grid_latitude: 180; grid_latitude: 360)
    Dimension coordinates:
        grid_latitude                             x                   -
        grid_latitude                             -                   x
```

## Expected behaviour
That iris would name the coordinates "latitude" and "longitude".

## Environment 
 - OS & Version: CentOS Linux 7
 - Iris Version: 3.1.0

pp-mo:
### Description
Closes #4460 

I realised that the reported problem was due to a mistake which I introduced in #4198 :
I had retained unchanged the trusted older routines, 'is_latitude/longitude' and 'is_rotated_latitude/longitude' , but changed the way in which they are used :  As written, 'is_rotated' relies on your having already ensured the truth of 'is_latitude_longitude'.
The small changes here to `iris.fileformats._nc_load_rules.actions` fix that.

Since there were no specific tests before, I have now developed a test harness for this aspect of the translation process.
Having encountered the problems described at #4469 (see below), I have only included simpler testcases, and particularly excluded some more complex ones which show potentially problematic changes in behaviour.  However, I think the remaining tests do show utility, and demonstrate that the problem raised was fixed.

(
Note: 
During this process I stumbled over a whole lot more unexpected complexity.  
It turns out that within #4198 I may have oversimplified the logic and behaviours of the old Pyke rules, and that should now  probably be reviewed : see #4469
)

### DIFF
--- a/lib/iris/fileformats/_nc_load_rules/actions.py
+++ b/lib/iris/fileformats/_nc_load_rules/actions.py
@@ -185,8 +185,9 @@ def action_provides_coordinate(engine, dimcoord_fact):
     # Identify the "type" of a coordinate variable
     coord_type = None
 
-    # NEW FORM
     if hh.is_latitude(engine, var_name):
+        # N.B. result of 'is_rotated_lat/lon' checks are valid ONLY when the
+        # relevant 'is_lat/lon' is also True.
         if hh.is_rotated_latitude(engine, var_name):
             coord_type = "rotated_latitude"
         else:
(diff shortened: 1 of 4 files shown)


bjlittle (approved):
@pp-mo Looks good to me.

The removal of PyKE and it's re-implementation was a BIG change, so it's inevitable that these types of issues will start to surface from users reporting unexpected behaviour. 

This is something that we're going to have to rely on somewhat as we're only as good as our test coverage, so thanks to @PAGWatson for letting us know :+1: 

@pp-mo Thanks again for the fix :beers:

bjlittle:
@pp-mo Forgot to mention that this is totally worthy of a `whatsnew` entry... fancy pushing a follow-on PR and I'll bank that for you?

PR #4476, opened 2022-01-04
rtd with latest mambaforge image for faster building

bjlittle:
### Description
This PR adopts the latest `mambaforge` image workflow from RTD to build our documentation, replacing our previous use of the RTD [Legacy build specification](https://docs.readthedocs.io/en/stable/config-file/v2.html#legacy-build-specification).

When using the legacy RTD specification, typically I've seen documentation builds taking between `670-730 seconds`.

Using the latest `mambaforge` image, this PR built its documentation in `454 seconds`... which is quite an improvement :rocket: 

(although, this is only a sample of one :wink:)

The built documentation associated with this PR can be viewed [here](https://bjlittle-iris.readthedocs.io/en/latest/), and the raw RTD build log can be viewed [here](https://readthedocs.org/api/v2/build/15679017.txt).

---

### DIFF
--- a/.readthedocs.yml
+++ b/.readthedocs.yml
@@ -18,7 +18,3 @@ python:
       path: .
       extra_requirements:
         - docs
-
-formats:
-  - htmlzip
-  - pdf
--- a/.readthedocs.yml
+++ b/.readthedocs.yml
@@ -1,20 +1,24 @@
 version: 2
 
 build:
-    image: latest
+  os: ubuntu-20.04
+  tools:
+    python: mambaforge-4.10
 
 conda:
-    environment: requirements/ci/readthedocs.yml
+  environment: requirements/ci/readthedocs.yml
 
 sphinx:
-    configuration: docs/src/conf.py
-    fail_on_warning: false
+  configuration: docs/src/conf.py
+  fail_on_warning: false
 
 python:
-    install: 
-        - method: setuptools
-          path: .
+  install:
+    - method: pip
+      path: .
+      extra_requirements:
+        - docs
 
 formats:
-    - htmlzip
-    - pdf
+  - htmlzip
+  - pdf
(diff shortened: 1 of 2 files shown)


.readthedocs.yml
```diff
@@ -1,20 +1,24 @@
 version: 2
 
 build:
-    image: latest
+  os: ubuntu-20.04
+  tools:
+    python: mambaforge-4.10
```
bjlittle, line 6: See the associated RTD documentation [Making builds faster with mamba](https://docs.readthedocs.io/en/stable/guides/conda.html#making-builds-faster-with-mamba)... certainly seems to deliver :+1:

bjlittle:
@tkknight Also I question our need to support `htmlzip` and `pdf` formats of the documentation.

It adds to the built time... and who actually needs or uses these formats? Do we know? :thinking:

.readthedocs.yml
tkknight: Looks like the RTD docs have been updated since I last looked at it.  Speedier build is welcome.

tkknight:
> 

I enable the html and pdf options as it was easy to do so and thought some may find it useful.  A quick look at other RTD hosted docs didnt show much uptake of it either - so yeah +1 for removing these two build targets.

tkknight (approved):
Looks good.

bjlittle:
@tkknight Thanks. I'll remove the `htmlzip` and `pdf` building, and also add a `whatsnew` entry

bjlittle:
So we're now down to about `385 seconds` to build the docs... which is about half-ish the build time :+1:

bjlittle:
Back to you @tkknight after a wierd CI barf

bjlittle:
... the days of `cirrus-ci` are numbered

bjlittle:
Awesome, thanks @tkknight :smile:

PR #4477, opened 2022-01-05
Loading Benchmarks

trexfeathers:
### Description

* Adds our tried and tested (offline and in iris-esmf-regrid) framework for stable generation of test data for benchmarks.
* Introduces a method (`create_um_files()`) for generating synthetic FF/PP/NetCDF files for benchmarking, using Mule.
  * This avoids reliance on [potentially enormous] real files.
  * We can size the files appropriately for benchmarking purposes.
  * A useful tool when following up on performance regressions that need more info.
* Ports our existing loading benchmarks from the offline benchmark suite.
  With some adaptations to use the new synthetic data generation.

I believe synthetic FF file generation removes the last major blocker to fully migrating our remaining offline benchmarks onto GitHub 🤞

---


trexfeathers:
Have decided against caching the generated benchmark data in the GHA job - GHA relies on working out whether or not to cache at the _start_ of the job, it is not possible to check for changes during/after. This therefore forces us to _always_ cache, which adds time/resource when 95% of the time it would be unnecessary.

Given benchmarks run during CI will necessarily be quick - i.e. small synthetic data - it is quicker and simpler to generate data files fresh at the start of the GHA job. The benchmarks at least have the intelligence to re-use the files once they are there.

jamesp:
I don't know how complicated your data gen scripts are, but it might be the case that regen is actually _faster_ than restoring from cache if CPU gen speed > network IO speed.

trexfeathers:
>I don't know how complicated your data gen scripts are, but it might be the case that regen is actually faster than restoring from cache if CPU gen speed > network IO speed.

Exactly. My thoughts, but more succinct!

wjbenfold:
Does it make sense to use real data in some of these benchmarks as well as synthetic (as a sanity check perhaps?). I believe that the benchmarks already have access to the test data

benchmarks/benchmarks/loading.py
```diff
@@ -0,0 +1,186 @@
...
+from iris.cube import Cube
+from iris.fileformats.um import structured_um_loading
+
+from .generate_data import BENCHMARK_DATA, REUSE_DATA, run_function_elsewhere
+from .generate_data.um_files import create_um_files
+
+
+class LoadAndRealise:
+    params = [
+        [(2, 2, 2), (1280, 960, 5)],
```
wjbenfold, line 28: We've seen a few issues where a pp file with lots of layers is significantly less performant than one with a few - could we try another shape (if I understand what's here correctly) that's not very wide but quite deep?

trexfeathers:
>Does it make sense to use real data in some of these benchmarks as well as synthetic (as a sanity check perhaps?). I believe that the benchmarks already have access to the test data

>We've seen a few issues where a pp file with lots of layers is significantly less performant than one with a few - could we try another shape (if I understand what's here correctly) that's not very wide but quite deep?

@wjbenfold even if I make no more changes to this PR, I'm already worried about:

1. The CI runtime
2. The size of the changes in this PR

Better coverage is always good in principle, but benchmarks take a long time to run. So to solve Point 1 I have some future plans for splitting benchmarks into categories: CI / semi-regular / on-demand, but the complexity of that runs up against Point 2. So I'd like to lay that groundwork in a future PR, not this one, after which the whole team will hopefully find it much easier to expand benchmark coverage without my help/interference.

benchmarks/benchmarks/loading.py
trexfeathers: https://github.com/SciTools/iris/pull/4477#issuecomment-1025613155

benchmarks/benchmarks/generate_data/__init__.py
```diff
@@ -0,0 +1,91 @@
...
+except KeyError:
+    error = "Env variable DATA_GEN_PYTHON not defined."
+    raise KeyError(error)
+except (CalledProcessError, FileNotFoundError, PermissionError):
+    error = (
+        "Env variable DATA_GEN_PYTHON not a runnable python executable path."
+    )
+    raise ValueError(error)
+
+default_data_dir = (Path(__file__).parents[2] / ".data").resolve()
```
pp-mo, line 40:
```suggestion
# Get default test-files location, for CI runs.
default_data_dir = (Path(__file__).parents[2] / ".data").resolve()
```

benchmarks/benchmarks/generate_data/__init__.py
```diff
@@ -0,0 +1,91 @@
...
+BENCHMARK_DATA = Path(environ.get("BENCHMARK_DATA", default_data_dir))
```
pp-mo, line 41:
```suggestion
# N.B. env variable is an override for manual testing.
BENCHMARK_DATA = Path(environ.get("BENCHMARK_DATA", default_data_dir))
```

benchmarks/benchmarks/generate_data/um_files.py
```diff
@@ -0,0 +1,212 @@
...
+    len_y: int,
+    len_z: int,
+    len_t: int,
+    compress: bool,
+    file_types: list,
+) -> dict:
+    """
+    Generate FF-based FF / PP / NetCDF files with specified shape and compression.
+
+    Saved to a directory for all files that shape. A dictionary of the saved
```
pp-mo, line 187: Bit of a glitch in the descrtiption here:
>  "all files that shape".

benchmarks/benchmarks/generate_data/um_files.py
```diff
@@ -0,0 +1,212 @@
...
+        },
+        "real_constants": {},
+        "level_dependent_constants": {"dims": (len_z + 1, None)},
+    }
+    new_ff = FieldsFile.from_template(deepcopy(template))
+
+    data_array = np.arange(len_x * len_y).reshape(len_x, len_y)
+    array_provider = ArrayDataProvider(data_array)
+
+    def add_field(level_: int, time_step_: int) -> None:
```
pp-mo, line 67: I don't understand why the arg names `level_`, `time_step_`  have the final underscores.
Does this have a particular meaning that I'm missing ?

pp-mo (changes requested):
Thanks @trexfeathers looking rather good.

Unfortunately, I'm not done with this yet, but I thought I should complete this review now so you can see my initial comments + get started on that.
Sorry for the delay, I'll get some more done this afternoon.

noxfile.py
```diff
@@ -289,15 +289,15 @@ def linkcheck(session: nox.sessions.Session):
     )
 
 
-@nox.session(python=PY_VER[-1], venv_backend="conda")
+@nox.session(python=PY_VER, venv_backend="conda")
 @nox.parametrize(
     ["ci_mode"],
     [True, False],
     ids=["ci compare", "full"],
```
pp-mo, line 296: N.B. I *was* going to ask for the options here to be clarified a bit, in terms of what you see in the "nox -l" lisiting.
But from offline conversation with @trexfeathers, there is a subsequent improvement to that coming soon in another PR.

trexfeathers: #4583 🙂

benchmarks/benchmarks/generate_data/um_files.py
trexfeathers: Some IDE's (I'm using PyCharm) warn about 'shadowing outer scope' - where variables have the same name as the next level up.

I assume this is from the risk of confusion where one could delete the 'inner' variable and the script might still run fine, just not in the way intended.

noxfile.py
```diff
@@ -315,6 +315,43 @@ def benchmarks(session: nox.sessions.Session, ci_mode: bool):
 
     """
     session.install("asv", "nox")
+
+    data_gen_var = "DATA_GEN_PYTHON"
+    if data_gen_var in os.environ:
+        print("Using existing data generation environment.")
+    else:
+        print("Setting up the data generation environment...")
+        session.run_always(
```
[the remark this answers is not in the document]
trexfeathers: 9fbcb336393e19c6d2ed3146b14509eb5808e374

noxfile.py
```diff
@@ -315,6 +315,43 @@ def benchmarks(session: nox.sessions.Session, ci_mode: bool):
...
+            "nox",
+            "--session=tests",
+            "--install-only",
+            f"--python={session.python}",
+        )
+        data_gen_python = next(
```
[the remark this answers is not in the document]
trexfeathers: 9fbcb336393e19c6d2ed3146b14509eb5808e374

noxfile.py
```diff
@@ -315,6 +315,43 @@ def benchmarks(session: nox.sessions.Session, ci_mode: bool):
...
+            Path(".nox").rglob(f"tests*/bin/python{session.python}")
+        ).resolve()
+        session.env[data_gen_var] = data_gen_python
+
+        print("Installing Mule into data generation environment...")
+        mule_dir = data_gen_python.parents[1] / "resources" / "mule"
+        if not mule_dir.is_dir():
+            session.run_always(
```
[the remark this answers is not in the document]
trexfeathers: 9fbcb336393e19c6d2ed3146b14509eb5808e374

benchmarks/benchmarks/generate_data/__init__.py
trexfeathers: 9fbcb336393e19c6d2ed3146b14509eb5808e374

benchmarks/benchmarks/generate_data/__init__.py
trexfeathers: 9fbcb336393e19c6d2ed3146b14509eb5808e374

benchmarks/benchmarks/generate_data/um_files.py
trexfeathers: 9fbcb336393e19c6d2ed3146b14509eb5808e374

pp-mo (changes requested):
Just a few points, most of them cosmetic.
Feel free to resolve whatever you may consider + reject.
Overall looking good ! :bouquet:

benchmarks/benchmarks/generate_data/__init__.py
```diff
@@ -0,0 +1,91 @@
...
+if BENCHMARK_DATA == default_data_dir:
+    BENCHMARK_DATA.mkdir(exist_ok=True)
+elif not BENCHMARK_DATA.is_dir():
+    message = f"Not a directory: {BENCHMARK_DATA} ."
+    raise ValueError(message)
+
+# Manual flag to allow the rebuilding of synthetic data.
+REUSE_DATA = True
```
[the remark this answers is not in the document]
trexfeathers: 9fbcb336393e19c6d2ed3146b14509eb5808e374

benchmarks/benchmarks/loading.py
```diff
@@ -0,0 +1,186 @@
...
+    def time_load(self, _, __, ___, ____) -> None:
+        _ = self.load()
+
+    def time_realise(self, _, __, ___, ____) -> None:
+        # Cache the original data object and re-apply after realisation, which
+        #  restores original state for the next repeat. The cache/apply steps
+        #  add negligible time to the benchmark result.
+        data_original = self.cube.core_data()
+        _ = self.cube.data
+        self.cube.data = data_original
```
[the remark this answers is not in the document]
trexfeathers: >I think I would use `self.cube.core_data().compute()`

Is your suggestion in place of `_ = self.cube.data`? It's hard to tell! If so, I supposed I was using the syntax that we normally recommend to the user.

>I'm puzzled by the need for this.

This was actually something you originally discovered and rectified. ASV performs re-runs between `self.setup()` calls, meaning that it can only time realisation once, then the data is already realised at the start subsequent re-runs. So we restore the original state before finishing the benchmark so that it's ready for the next re-run (providing that this action doesn't have a big effect on the benchmark result).

benchmarks/benchmarks/loading.py
```diff
@@ -0,0 +1,186 @@
...
+        import numpy as np
+
+        from iris import save
+        from iris.coords import AuxCoord
+        from iris.cube import Cube
+
+        data_len = 8
+        data = np.arange(data_len)
+        cube = Cube(data, units="unknown")
+        extra_vars = 80
```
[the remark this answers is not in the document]
trexfeathers: Maybe, but here I'm mostly interested in migrating the coverage we already had offline, rather than improving coverage at the same time. Do you think it's important?

benchmarks/benchmarks/loading.py
```diff
@@ -0,0 +1,186 @@
...
+        names = ["coord_" + str(i) for i in range(extra_vars)]
+        for name in names:
+            coord = AuxCoord(data, long_name=name, units="unknown")
+            cube.add_aux_coord(coord, 0)
+        save(cube, save_path)
+
+    def setup(self) -> None:
```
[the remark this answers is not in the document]
trexfeathers: 79a822867219acbfdcdd84ffc827d811b1d6918b

Too much copy-paste on the last day before Christmas!

benchmarks/benchmarks/generate_data/um_files.py
```diff
@@ -0,0 +1,212 @@
...
+    len_z: int,
+    len_t: int,
+    compress: bool,
+    file_types: list,
+) -> dict:
+    """
+    Generate FF-based FF / PP / NetCDF files with specified shape and compression.
+
+    Saved to a directory for all files that shape. A dictionary of the saved
+    paths is returned.
```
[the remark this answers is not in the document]
trexfeathers: b69a388437fa6e136b709d01a5c15eeb8f8bded0

benchmarks/benchmarks/loading.py
wjbenfold: I'm happy that this shouldn't be done here, but think our benchmark coverage should (relatively soon) be in a place where it accounts for this.

trexfeathers: Yes, agreed

benchmarks/benchmarks/loading.py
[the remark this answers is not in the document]
pp-mo: > migrating the coverage we already had offline, rather than improving coverage at the same time

No, I think I have to agree there is no great need, and making it comparable to the existing has it's own value anyway.

benchmarks/benchmarks/generate_data/um_files.py
pp-mo: Ok thanks, wasn't really aware of that.
Interesting : I'm sure I do write code that might trigger this -- so, my PyCharm had that warning turned off for some reason.
Anyway it's a reasonable approach to avoid those warnings. :+1:

benchmarks/benchmarks/loading.py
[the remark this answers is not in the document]
pp-mo: **Update**: well **_I_** still think it would be simpler to ditch all the comment + explicitly fetch the lazy content + compute it.
But we can discuss.

pp-mo (changes requested):
All good now except remaining question on https://github.com/SciTools/iris/pull/4477#discussion_r805323502
We can discuss that + either change it or not.
Will then merge !

benchmarks/benchmarks/loading.py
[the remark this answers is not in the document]
trexfeathers: f96480db26e38cafecb2bcdfbc5b5446c0842b11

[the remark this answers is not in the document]
pp-mo: :+1: for explaining the issues nicely !

pp-mo (approved):

trexfeathers:
Thanks @pp-mo!

PR #4484, opened 2022-01-07
Remove all Metarelate dependencies.

pp-mo:
Changes to decouple Iris metadata translation from dependency on Metarelate (http://www.metarelate.net/).

This is a slightly bold step, but I think it makes sense.
It means that we can freely manage metadata translations by editing the source in Iris : the files which were previously "auto-generated" are now themselves "first-class citizens", managed directly as sourcecode in the Iris repo.

N.B. there is really no obstacle to reverting this at a future date, if Metarelate does achieve wider general usage.

The drivers for this are the following:
  * Although the _idea_ of an independent, collaboratively managed metadata relationship database is sound, and the implementation is viable, it appears to be getting no attention whatsoever from the target (Science) community
  * Nothing has changed on the main Metarelate repo since [Nov 29, 2019](https://github.com/metarelate/metOcean/commit/448f2ef5e676edaaa27408b9f3ddbecbf05e3289)
  * Efforts to promote to ordinary (Science) users, and to further document the editing processes,  appear to have made no progress for a similar period
  * maintenance support for the Metarelate infrastructure is virtually non-existent, and we have been warned that if/when it breaks we will be "stuck"
  * The mechanism for editing the database is highly complex and under-documented
  * The necessary software environment to run the editor is rather complex
    * it requires Java as well Python
    * it might not work (has not been tested) in Python3, without further fixes
    * a site.cfg is required, the content of which is not obvious
    * when last exercised, operation was very slow + flaky using the necessary web-based 'remote requests'


bjlittle (commented):
@pp-mo Thanks for taking this on.

I second this approach. Perhaps the status of `metarelate` changes at a future date, at which point we can review the situation again, otherwise we require to decouple ourselves from the `metarelate` tooling.

trexfeathers (commented):
Pleasing improvements to codebase simplicity!

Two possible documentation additions (excuse my ignorance):

### [Releases](https://github.com/SciTools/iris/blob/main/docs/src/developers_guide/release.rst)

Do `_ff_cross_references.py` and `um_cf_map.py` need to be checked for alignment with the updated standard names table?

https://github.com/SciTools/iris/blob/20e0caeca1918d27a1115624dd966fb9f4079066/docs/src/developers_guide/release.rst#L39-L46

### [Contributing Changes](https://github.com/SciTools/iris/blob/main/docs/src/developers_guide/contributing_changes.rst)

A small new page describing what files users should change in a PR if they want to propose changes to the existing mapping?

lbdreyer (approved):
I agree that this is a good move. As it stands, going through metarelate is a bit of a barrier to updating these mappings so, with Metarelate being a bit stale, this is a problem.

And as others have mentioned, if metarelate gets picked up again, we can always start using it again.

lbdreyer:
> Do _ff_cross_references.py and um_cf_map.py need to be checked for alignment with the updated standard names table?
It's an interesting thought and not one I've really considered doing when reviewing updates to standard name tables. Could we add it as an extra test in our test suite (in a separate PR if possible though. I think we can still merge this in without that)?

bjlittle:
@pp-mo Fancy resolving the conflicts in the `whatsnew` so that we can bank this PR :+1:

pp-mo:
> @trexfeathers Releases ... Do _ff_cross_references.py and um_cf_map.py need to be checked for alignment with the updated standard names table?

My feeling is **_no_** :  
  * Firstly, I don't think updated standard-names are ever likely to practically affect the translations, not least because they are generally managed to be backwards-compatible, i.e. the CF community avoid modifying existing terms at almost all costs.
  * Secondly, I don't see any _other_ need to make this a standard consideration at release time, now that we don't need a regular prompt to re-load from Metarelate :  The translation information is logically separate from other things, and can generally updated as-and-when requests are raised.  In practice, there has actually been very little change to it for some years.

pp-mo:
Now rebased, should go OK !

PR #4481, opened 2022-01-07
nep29: drop py37 support

bjlittle:
### Description
This PR drops support for `py37` as per the [NEP29 backwards compatibility and deprecation policy support schedule](https://numpy.org/neps/nep-0029-deprecation_policy.html#support-table).

Also see [Python endoflife](https://endoflife.date/python).

---
[Consult Iris pull request check list]( https://scitools-iris.readthedocs.io/en/latest/developers_guide/contributing_pull_request_checklist.html)

PR #4486, opened 2022-01-10
address np printoptions pre/post v1.22

bjlittle:
### Description
This PR addresses the test failures within #4480 when updating to the latest `numpy` v1.22.0.

We've now dropped support within `iris` for `py37` (#4481) but we still require to support `numpy` >= v1.19 ([NEP29 Support Table](https://numpy.org/neps/nep-0029-deprecation_policy.html#support-table)). However https://github.com/numpy/numpy/pull/19686 introduced a change that impacts `iris.util.format_array` due to our use of the internal `numpy.core.arrayprint._formatArray` function.

This PR maintains our continuing `numpy` v1.13 legacy array printing format for `py38` and `numpy` >= 1.19.

As a separate concern, it would be wise to commit to addressing #3048, by _at least_ bumping to the more recent `numpy` v1.21 legacy array printing.

---

### DIFF
--- a/lib/iris/util.py
+++ b/lib/iris/util.py
@@ -1059,13 +1059,14 @@ def format_array(arr):
     ffunc = str
     max_line_len = 50
 
-    formatArray = np.core.arrayprint._formatArray
-    format_options = np.core.arrayprint._format_options
+    # Format the array with version 1.13 legacy behaviour
+    with np.printoptions(legacy="1.13"):
+        # Use this (private) routine for more control.
+        formatArray = np.core.arrayprint._formatArray
+        # N.B. the 'legacy' arg had different forms in different numpy versions
+        # -- fetch the required form from the internal options dict
+        legacy_key = np.core.arrayprint._format_options["legacy"]
 
-    options = np.get_printoptions()
-    options["legacy"] = "1.13"
-
-    with np.printoptions(**options):
         result = formatArray(
             arr,
             ffunc,
@@ -1074,7 +1075,7 @@ def format_array(arr):
             separator=", ",
             edge_items=edge_items,
             summary_insert=summary_insert,
-            legacy=format_options["legacy"],
+            legacy=legacy_key,
         )
 
     return result
--- a/lib/iris/util.py
+++ b/lib/iris/util.py
@@ -1065,7 +1065,7 @@ def format_array(arr):
         formatArray = np.core.arrayprint._formatArray
         # N.B. the 'legacy' arg had different forms in different numpy versions
         # -- fetch the required form from the internal options dict
-        legacy_key = np.core.arrayprint._format_options["legacy"]
+        format_options_legacy = np.core.arrayprint._format_options["legacy"]
 
         result = formatArray(
             arr,
@@ -1075,7 +1075,7 @@ def format_array(arr):
             separator=", ",
             edge_items=edge_items,
             summary_insert=summary_insert,
-            legacy=legacy_key,
+            legacy=format_options_legacy,
         )
 
     return result
--- a/lib/iris/util.py
+++ b/lib/iris/util.py
@@ -10,7 +10,6 @@ Miscellaneous utility functions.
 
 from abc import ABCMeta, abstractmethod
 from collections.abc import Hashable, Iterable
-from contextlib import contextmanager
 import copy
 import functools
 import inspect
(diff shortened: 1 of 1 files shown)


bjlittle:
At the moment this PR is testing with `py38` and `numpy` 1.21.5.

However, this PR change also supports `py38` and `numpy` 1.22.0.

I propose that if we're happy with this PR, bank it, then I'll create a follow-on PR that incorporates the `nox.lock/py38-linux-64.lock`  from #4480 (and close #4480).

lib/iris/util.py
```diff
@@ -1074,29 +1074,12 @@ def format_array(arr):
             separator=", ",
             edge_items=edge_items,
             summary_insert=summary_insert,
-            legacy=legacy,
+            legacy=format_options["legacy"],
```
trexfeathers, line 1077:
```suggestion
            legacy=False,
```

Numpy is programmed with handling for `False`, and this would avoid us importing the private `_format_options`. I can be convinced otherwise!

trexfeathers (commented):
I have confirmed for myself that this will fix the #4480 failures, and I _think_ I understand what's going on! I just had one comment about avoiding a private import.

lib/iris/util.py
trexfeathers: Note all our tests still pass when this change is included. Is that evidence enough?

bjlittle: @trexfeathers We don't want `legacy=False` here.

You can see that from the different implementations of `numpy.core.arrayprint._formatArray` for 1.21.5 vs 1.22.0

We can use the safe route of setting the `legacy` option through the public API of `numpy.printoptions` context manager and rely on it correctly translating it to the correct internal value within the `numpy.core.arrayprint._format_options` dictionary for the `"legacy"` key.

bjlittle: > Note all our tests still pass when this change is included. Is that evidence enough?

I think it is. What else is there to do?

We're not in the business of testing a matrix of `numpy` versions against different `python` and `iris` versions.

... but I do know that this PRs changes work for `numpy` 1.21.5 **and** 1.22.0. That should be sufficient, right?

trexfeathers: But since it's within the context manager already, providing the key doesn't make any difference?

trexfeathers: >I think it is. What else is there to do?

I was referring in this case to the tests still passing even when using `legacy=False`.

pp-mo: I think I can see the point of doing this if the new code would remain compatible with **_older_** versions of numpy.
**But surely we have already burned that bridge by adopting the context-manager form ??**
It seems a lot to assume that this will save us from _future_ changes.

TBH I'd much rather we had a stable, supported mechanism for this printed output.
It's caused so much work over the years, and it's absolutely bound to change again.
But equally there is probably a lot of clever stuff in the formatter, and I think even a minimal replacement would require a lot of work.

bjlittle: > I was referring in this case to the tests still passing even when using `legacy=False`.

What does this matter? True, this is an interesting curiosity, but setting `legacy=False` is not the correct thing to do here. It's simply not future proof, particularly if we come to change `iris.util.format_array` and pin it to a future legacy style, such as `1.21`, as suggested as an option in `numpy` 1.22.0

The behaviour that we see from `format_array` is directly dependent on the data that is being printed. Setting `legacy=False` does not guarantee that the data of a future `iris` test will be truly 1.13 legacy compliant... regardless of whether all the `iris` tests are currently passing at the moment with `legacy=False`.

The doc-string for the `legacy` kwarg on `set_printoptions` is:
```
    legacy : string or `False`, optional
        If set to the string `'1.13'` enables 1.13 legacy printing mode. This
        approximates numpy 1.13 print output by including a space in the sign
        position of floats and different behavior for 0d arrays. If set to
        `False`, disables legacy mode. Unrecognized strings will be ignored
        with a warning for forward compatibility.
```
i.e., `legacy=False` means **disables legacy mode**. Note that, there is special treatment of the sign for floats and 0d arrays for `legacy="1.13"`. This is **NOT** the case for `legacy=False`, as is plain to see in the `numpy.core.arrayprint` code base for both `numpy` 1.21.5 and 1.22.0

Also, calling `numpy.core.arrayprint._formatArray` with `legacy=False` within the `numpy.printoptions` context-manager scope is just plain confusing to the reader, simply because the `legacy="1.13"` kwarg has just been passed to context-manager.

Note that I've attempted to introduce the minimum changes required here to continue successful `numpy` legacy 1.13 array printing for the `iris` supported versions of `numpy` (>=1.19) using `py38`. I'm not condoning the use of private `numpy` functionality, and I'm not going to attempt to re-write our approach to array printing, I'm simply attempting to unblock `iris` for the 3.2.0 release, given the diminishing time-scale.

If you don't want to merge this PR, then I see the following options:
1. the `iris` 3.2.0 can go ahead, but we can't test successfully against `numpy` +1.22.0, and `main` is also pinned to `numpy` <1.22.0 [:-1: from me]
2. invest more time and effort with a different strategy and PR for `iris` array printing of `numpy` arrays [personally, I'm not going to invest effort to do that at time]

Your call.

pp-mo: Following discussion, I think this will do for now.

I think we shouldn't try to come up with a better way for this now, but **_should_** review it after the release, as outlined in [this recent comment on the longstanding issue about it](https://github.com/SciTools/iris/issues/3048#issuecomment-1009744727)

pp-mo (approved):
Good enough for now, as noted in comments.

bjlittle:
Thanks @pp-mo and @trexfeathers :beers: :partying_face: :100: 

There was a lot of history and tech debt associated with this one. 

Hopefully it's raised the priority on us committing to resolving https://github.com/SciTools/iris/issues/3048 and moving away from what we have at the moment

Issue #3528, opened 2019-11-13
Remove experimental equalise_cubes


bjlittle:
Follow-up on #3527, and remove the `iris.experimental.equalise_cubes` module, akin to #3523.

Target this for the `3.1.0` release.

bjlittle:
Closes #3342

larsbarring:
Now, when this function becomes a 'regular member of iris', would it be possible to have it returning the attributes ( etc) that had to be removed to equalise the cubes?

bjlittle:
Hey @larsbarring,

Good to hear from you. I'm assuming that this is because you want to reattach some of the metadata that was removed?

larsbarring:
Yes, either reattaching as for example with tracking_id, or just to know what is not consistent when I expected it to be.
Lars

bjlittle:
Thanks @larsbarring, much appreciated.

My head is in the space of considering how to make `iris` more lenient and customisable for users, particularly when performing such operations as this, as well as concatenation, merging, arithmetic and equality.

You should probably also be aware of https://github.com/SciTools/iris/issues/3325 which intends to address the issue of `iris` being opinionated about what `attributes` it cares about.

Anyways, I appreciated the unnecessary hoops that users are forced to jump through, so your use case here is valuable, thanks for taking the time to feedback 😃

PR #4496, opened 2022-01-12
Delete equalise_cubes.py and add whatsnew

wjbenfold:
### Description

Remove `iris.experimental.equalise_cube`, closing #3528

---


wjbenfold:
Not sure where in what's new this belongs - is it Internal or a Deprecation (which I assumed were things we're deprecating now, not removing having previously deprecated them)

lbdreyer:
I'm tempted to say it may make more sense for the whats new to be added to deprecations. We state that something will be deprecated, and then follow this up by confirming it has actually been removed, and no longer available. As a user, if I wanted to check when it was actually removed, my first port of call would be to look under the "deprecations" section as it is related to that. 
I wouldn't think a user should necessarily have to concern themself with what goes in "internal" so they may not look there

lbdreyer:
Don't think I made myself so clear in the lat comment, so to follow up: I think removing code is _part of the deprecation_ process so the what's new should live under the deprecations heading

rcomer (approved):
I agree it makes sense to have the whatsnew in internal.  The user will see no functional change (other than a different error message if they try to import this module).

Edit: x-post with @lbdreyer and will defer to her!

lbdreyer:
> Edit: x-post with @lbdreyer and will defer to her!

Whoops! danger of commenting at the same time! But I suppose this proves there's no correct answer; both have valid reasons.

Another point I wanted to mention though: I think this is a bit of a strange deprecation, in that we changed the code of the old function. Normally I would expect, when using the function, to get a deprecation warning, but to still be able to use it, until its eventual removal. In that example, I would say it definitely belongs in deprecations

lbdreyer:
Thanks @wjbenfold !

PR #4497, opened 2022-01-12
GeogCS.as_cartopy_projection handling non-Earth planets

### LINKED ISSUE #4408: Conversion between projections & coordinate systems does not always preserve the ellipsoid.
## 🐛 Bug Report

Conversion between `iris` and `cartopy` projections/coordinate systems does not always preserve the ellipsoid.

I noticed this behaviour when I was trying to rotate winds to a rotated pole coordinate system. The crux of the issue is that the original coordinate system attached to the cubes of winds is defined with a non-Earth semi-major axis (I am modelling a different planet). However, when I try to apply `iris.analysis.cartography.rotate_winds()`, I get this error:
```python
ProjError: Error creating Transformer from CRS.: (Internal Proj Error: proj_create_operations: Source and target ellipsoid do not belong to the same celestial body)
```
This apparently happens at this step in the `rotate_winds()` function:
```python
src_crs = x_coord.coord_system.as_cartopy_projection()
```
because `GeogCS.as_cartopy_projection()` just calls `ccrs.PlateCarree()` without passing any information about the globe.

My code that used these operations used to work (~ a 1-2 months ago), so it's likely related to some changes in the recent versions of `iris` , `cartopy` or `pyproj`.

## How To Reproduce

```python
from iris.coord_systems import GeogCS

planet_crs = GeogCS(semi_major_axis=123456)
print(planet_crs.as_cartopy_projection().proj4_init)
```
result:
```
'+ellps=WGS84 +a=6378137.0 +proj=eqc +lon_0=0.0 +to_meter=111319.4907932736 +vto_meter=1 +no_defs'
```

## Expected behaviour

```
'+ellps=WGS84 +a=123456.0 +proj=eqc +lon_0=0.0 +to_meter=2154.7136813421193 +vto_meter=1 +no_defs'
```

I think the conversion between coordinate systems should respect the ellipsoid and not default to Earth (or at least have a warning).

## Environment 
 - OS & Version: e.g., Ubuntu 20.04 LTS
 - Iris Version: 3.1.0
 - Cartopy Version: 0.20.1
 - Pyproj Version: 3.2.1

wjbenfold:
### Description

Fixes #4408 by passing the `GeogCS` as a globe to the `cartopy.crs.PlateCarree` constructor

---

### DIFF
--- a/lib/iris/tests/test_coordsystem.py
+++ b/lib/iris/tests/test_coordsystem.py
@@ -188,6 +188,23 @@ class Test_GeogCS_as_cartopy_globe(tests.IrisTest):
         self.assertEqual(res, expected)
 
 
+class Test_GeogCS_as_cartopy_projection(tests.IrisTest):
+    def test_as_cartopy_projection(self):
+        geogcs_args = {
+            "semi_major_axis": 6543210,
+            "semi_minor_axis": 6500000,
+            "longitude_of_prime_meridian": 30,
+        }
+        cs = GeogCS(**geogcs_args)
+        cs_projection = cs.as_cartopy_projection()
+        res = {
+            "semi_major_axis": cs_projection.globe.semimajor_axis,
+            "semi_minor_axis": cs_projection.globe.semiminor_axis,
+            "longitude_of_prime_meridian": cs_projection.prime_meridian.longitude,
+        }
+        self.assertEqual(res, geogcs_args)
+
+
 class Test_GeogCS_as_cartopy_crs(tests.IrisTest):
     def test_as_cartopy_crs(self):
         cs = GeogCS(6543210, 6500000)
--- a/lib/iris/tests/test_coordsystem.py
+++ b/lib/iris/tests/test_coordsystem.py
@@ -196,14 +196,19 @@ class Test_GeogCS_as_cartopy_projection(tests.IrisTest):
             "longitude_of_prime_meridian": 30,
         }
         cs = GeogCS(**geogcs_args)
-        cs_projection = cs.as_cartopy_projection()
-        exps = [
-            f"a={geogcs_args['semi_major_axis']}",
-            f"b={geogcs_args['semi_minor_axis']}",
-            f"lon_0={geogcs_args['longitude_of_prime_meridian']}",
-        ]
-        for exp in exps:
-            self.assertTrue(exp in cs_projection.srs)
+        res = cs.as_cartopy_projection()
+
+        globe = ccrs.Globe(
+            semimajor_axis=geogcs_args["semi_major_axis"],
+            semiminor_axis=geogcs_args["semi_minor_axis"],
+            ellipse=None,
+        )
+        expected = ccrs.PlateCarree(
+            globe=globe,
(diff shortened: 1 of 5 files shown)


wjbenfold:
Currently the introduced test fails because Cartopy doesn't appear to pick up the central longitude all the way through (see example below). It's sort of there, but not fully. Don't know whether to make the test less stringent, stop trying to pass through the central longitude at all, or do something else.

```
>>> gcs = GeogCS(6000000, 6000000, longitude_of_prime_meridian=30)
>>> gcs.__dict__
{'semi_major_axis': 6000000.0, 'semi_minor_axis': 6000000.0, 'inverse_flattening': 0.0, 'longitude_of_prime_meridian': 30.0}
>>> pc = gcs.as_cartopy_projection()
>>> pc
<Derived Projected CRS: +proj=eqc +a=6000000.0 +b=6000000.0 +lon_0=30.0 +t ...>
Name: unknown
Axis Info [cartesian]:
- E[east]: Easting (unknown)
- N[north]: Northing (unknown)
- h[up]: Ellipsoidal height (metre)
Area of Use:
- undefined
Coordinate Operation:
- name: unknown
- method: Equidistant Cylindrical (Spherical)
Datum: unknown
- Ellipsoid: unknown
- Prime Meridian: Reference meridian

>>> pc.prime_meridian
PRIMEM["Reference meridian",0,
    ANGLEUNIT["degree",0.0174532925199433,
        ID["EPSG",9122]]]
>>> pc.prime_meridian.longitude
0.0
```

wjbenfold:
As well as the test I know is failing (that's my new one), there's also one failing in `test_mapping` because it relies on getting a vanilla PlateCarree back from `as_cartopy_projection` rather than one the right size for the cube. What's the best fix for this?

wjbenfold:
The fix for the failing test in `test_mapping` seems somewhat circular, in that the test now uses very similar logic to the function it's testing. Seems fine though? The other option is to just remove that assert from the test, though maybe it's important to check that PlateCarree is the default?

lib/iris/tests/test_coordsystem.py
```diff
@@ -188,6 +188,24 @@ def test_as_cartopy_globe(self):
...
+        }
+        cs = GeogCS(**geogcs_args)
+        cs_projection = cs.as_cartopy_projection()
+        exps = [
+            f"a={geogcs_args['semi_major_axis']}",
+            f"b={geogcs_args['semi_minor_axis']}",
+            f"lon_0={geogcs_args['longitude_of_prime_meridian']}",
+        ]
+        for exp in exps:
+            self.assertTrue(exp in cs_projection.srs)
```
stephenworsley, line 206: The test below (Test_GeogCS_as_cartopy_crs) compares cartopy projections directly. Is there an advantage to checking the attributes directly here instead? I could see the below approach perhaps being less sensative to changes to cartopy.

wjbenfold: I've changed this as suggested, though I prefer a slightly more explicit way of entering values than the one in the test below so that it's clear I expect the same value, not just a number that's the same.

stephenworsley (approved):
Looks good to me.

PR #4504, opened 2022-01-14
Mesh Documentation

### LINKED ISSUE #4489: Document UGRID Support
## 📚 Documentation

Iris `v3.2` will be the first release that includes UGRID support (`iris.experimental.ugrid`). This is a large topic, with several new concepts, so needs introducing to the user.

### [What's New](https://github.com/SciTools/iris/blob/34ee90ae1dd24a06d612c527ad4cc7dbd5855327/docs/src/whatsnew/latest.rst)

The What's New page already includes several mentions of the continuing UGRID development. Once all features are complete this should be revised as necessary, particularly the `Release Highlights` section.

### [Further Topics](https://github.com/SciTools/iris/tree/34ee90ae1dd24a06d612c527ad4cc7dbd5855327/docs/src/further_topics)

Need a few pages to explain the new concepts. In `Further Topics` since it's inappropriate for the `User Guide` until out of `experimental`.

#### Format

- Kept brief
  - Want to minimise the changes needed as we continue to evolve the codebase
  - Aim for 5 pages (challenge!)
  - Accept that some stuff won't make the cut!
- Try using more worked examples in place of prose

#### Include

- Explanation of the **experimental** status
- What it doesn't do yet
- A UGRID/Iris-UGRID data model explanation including diagram (and printed cubes?)
- Relationship with iris-esmf-regrid and GeoVista, which provide some of the required functionality
- Operations (using worked examples):
  - Load/save
  - Plotting - i.e. use GeoVista
  - Regional extraction
    (and lack of indexing)
  - Regridding - i.e. use iris-esmf-regrid
  - Equality?
  - Recombination
- What's coming next

## The Plan

Writing: @trexfeathers
Reviewing: @pp-mo
Fresh perspective: @wjbenfold fancy this?

trexfeathers:
# [Click here for rendered docs](https://trexfeathers-iris.readthedocs.io/en/ugrid_docs/further_topics/ugrid/index.html)

### Description

Closes #4489. Adds a Mesh section to the documentation `Further Topics`.

To do:
* [x] Flesh out `data_model.rst` .
* [x] Flesh out `partner_packages.rst`.
* [x] Flesh out `operations.rst`

---


docs/src/further_topics/ugrid/operations.rst
```diff
@@ -0,0 +1,56 @@
...
+..
+    Highlight the uselessness of indexing.
+
+Regridding
+----------
+
+Equality
+--------
+..
+    Is this worth mentioning, given it just works the way it always has?
```
wjbenfold, line 44: I think so, reassurance is nice

docs/src/whatsnew/latest.rst
```diff
@@ -16,7 +16,7 @@ This document explains the changes made to Iris for this release
 
    The highlights for this minor release of Iris include:
 
-   * We've added experimental support for `UGRID`_ meshes which can now be loaded
+   * We've added experimental support for `CF-UGRID`_ meshes which can now be loaded
```
wjbenfold, line 19: Docs link here too?

lbdreyer:
Looking really great to me so far! The layout seems really easy to follow, and it looks like you are planning on covering all the most important parts, without overwhelming the user with info! 
I definitely agree with your plan to use examples as much as possible, I was just wondering about the practicalities of getting this working. In particular, do you have data that can be shared publicly for the examples?

docs/src/further_topics/ugrid/data_model.rst
```diff
@@ -0,0 +1,202 @@
+.. include:: ../../common_links.inc
+
+.. _ugrid model:
```
pp-mo, line 3: Looking good.
But overall I wish we could say 'unstructured data' a bit more, and 'UGRID' a bit less.
The way I see it, we have borrowed UGRID as a template for our data model, enabling us to represent unstructured data.  But we aim to support unstructured data, per se, whatever form we got it in.  
So we want to make the point that this is (or at least will be) a general facility in Iris.
We kind-of proved that _already_, by interoperating with ESMF grids in esmf-regrid and VTK data in Geovista.

docs/src/further_topics/ugrid/partner_packages.rst
```diff
@@ -0,0 +1,93 @@
...
+   :width: 300
+   :class: no-scaled-link
+
+.. rubric:: "A collection of structured and unstructured ESMF regridding schemes for Iris"
+
+ESMF provide a sophisticated, performant regridding utility that supports a
+variety of regridding types with both structured grids and unstructured meshes.
+A Python interface does exist - ESMPy - however working with Iris
+:class:`~iris.cube.Cube`\s and :class:`~iris.experimental.ugrid.Mesh`\es is
+sub-optimal. iris-esmf-regrid provides easy Iris-ESMPy interoperability, as
```
stephenworsley, line 81: A minor point, but the use of the word sub-optimal here could suggest that the `Cube` and `Mesh` objects have inherent flaws which make working with them sub-optimal, or that ESMF has inherent shortcomings beyond not supporting dask and the API not knowing about iris.
This could be better phrased as something like:
"however, this is not designed to work directly with `Cube`s and `Mesh`es."

trexfeathers: Yes, I was a bit uncertain about this myself.

c670b92d1d6c1a89b32ed220c743300b4a9290f4

docs/src/further_topics/ugrid/data_model.rst
trexfeathers: 77e3f95aa4bf4b57e1c60788fd0462191e6645ce

docs/src/further_topics/ugrid/index.rst
```diff
@@ -0,0 +1,54 @@
...
+----
+
+Meshes are different
+  Mesh-located data is fundamentally different to grid-located data.
+  Many of Iris' existing operations need adapting before they can work with
+  mesh-located data, and in some cases entirely new concepts are needed.
+  **Read the detail in these pages before jumping into your own code.**
+Iris' mesh support is experimental
+  This is a rapidly evolving part of the codebase at time of writing
+  (``Jan 2021``), as we continually expand the operations that work with mesh
```
jamesp, line 27: Jan 2022

docs/src/further_topics/ugrid/data_model.rst
```diff
@@ -0,0 +1,319 @@
...
+        This page is intended to summarise the essentials that Iris users need
+        to know about meshes. For exhaustive details on UGRID itself:
+        `visit the official UGRID conventions site`__.
+
+Evolution, not revolution
+=========================
+Mesh support has been designed wherever possible to fit within the existing
+Iris model. Meshes concern only the spatial location of data, and can
+optionally be limited to just the horizontal location (e.g. X and Y). Other
+dimensions such as time or experimental re-runs (and often vertical levels)
```
jamesp, line 19:
```suggestion
dimensions such as time or ensemble member (and often vertical levels)
```

jamesp: I found the "suggest changes" button! Most will be super picky/pedantic - feel free to ignore.

trexfeathers:
`operations.rst` is likely to get rather large, given the agreed effectiveness of example code. To compensate for this I would quite like to introduce the [Sphinx `collapse` directive](https://sphinx-toolbox.readthedocs.io/en/stable/extensions/collapse.html), but I believe it would be the first third party extension we've used so I'm not sure if that's OK.

Any input @tkknight?

docs/src/further_topics/ugrid/data_model.rst
```diff
@@ -0,0 +1,319 @@
...
+~~~~~~~~~~~~~~~~
+Assigning data to locations using a structured grid is essentially an act of
+matching coordinate arrays to each dimension of the data array. The data can
+also be represented as an area (instead of a point) by including a bounds array
+for each coordinate array.
+
+.. figure:: images/data_structured_grid.svg
+   :alt: Diagram of how data is represented on a structured grid
+
+   Data on a structured grid
```
jamesp, line 54: Are there references in the text pointing to these images? I'm not sure I fully follow what the circles and crosses mean, e.g. are there x-y axes on the cross diagrams, but index axes on the dot diagrams? And the diamonds are scalar values mapped to those coords?

image

docs/src/further_topics/ugrid/data_model.rst
```diff
@@ -0,0 +1,319 @@
...
+
+Unstructured Meshes
+~~~~~~~~~~~~~~~~~~~
+The most basic element in a mesh is the 0-dimensional **node**: a single
+location in space. Every node in the mesh is defined by indexing the
+1-dimensional X and Y (and optionally Z) coordinate arrays (the
+``node_coordinates``) - e.g. ``(x[3], y[3])`` gives the position of the fourth
+node. Note that this means each node has its own coordinates, independent of
+every other node. Since nodes can be anywhere in the space - **unstructured** -
+the position in the array has nothing to do with spatial position.
```
jamesp, line 66:
```suggestion
the position in the array (pink dot in Fig 25) has nothing to do with spatial position (corresponding pink cross in Fig 25).
```

jamesp:
It's looking great! 👍

docs/src/further_topics/ugrid/data_model.rst
trexfeathers: I've been trying to keep the prose and diagrams self-contained - for accessibility and ease of maintenance.

And I had also hoped that the diagrams would be self-describing where possible. But from yours and others' feedback there are parts that aren't as obvious as I had thought! I'll see what I can think of to make things clearer within the diagrams, while avoiding information overload.

docs/src/further_topics/ugrid/data_model.rst
trexfeathers: https://github.com/SciTools/iris/pull/4504#discussion_r790679955

docs/src/further_topics/ugrid/data_model.rst
trexfeathers: >But from yours and others' feedback there are parts that aren't as obvious as I had thought! I'll see what I can think of to make things clearer within the diagrams, while avoiding information overload.

Other opinions on this would be gratefully received @SciTools/developers

docs/src/further_topics/ugrid/data_model.rst
trexfeathers: 1d94cc892fcefdc2da2abdd32e593d1fb659d3b8

docs/src/further_topics/ugrid/index.rst
trexfeathers: 1d94cc892fcefdc2da2abdd32e593d1fb659d3b8

docs/src/further_topics/ugrid/data_model.rst
```diff
@@ -0,0 +1,326 @@
...
+
+If data is assigned to node location it must be stored in a 1D array of equal
+length to the number of nodes in the mesh. ``data[3]`` is at the position:
+``(x[3], y[3])``.
```
pp-mo, line 70: This sentence starting with "data[3]" seemed to me to appear somewhat 'out of the blue', as a bit of a shock.
Possibly all it needs is a "Thus, for example... " before it ?

docs/src/further_topics/ugrid/data_model.rst
```diff
@@ -0,0 +1,326 @@
...
+
+* | :class:`iris.experimental.ugrid.Mesh`
+  | Contains all information about the mesh.
+  | Includes:
+
+  * 1-3 collections of :class:`iris.coords.AuxCoord`\s:
+
+    * | **Required**: :attr:`~iris.experimental.ugrid.Mesh.node_coords`
+      | The nodes that are the basis for the mesh.
+    * | Optional: :attr:`~iris.experimental.ugrid.Mesh.edge_coords`,
```
pp-mo, line 258: It may be worth a special note that we still ***require*** these at present (a specific limitation of the current Mesh implementation)

docs/src/further_topics/ugrid/data_model.rst
```diff
@@ -0,0 +1,326 @@
...
+:attr:`~iris.experimental.ugrid.Mesh.edge_node_connectivity`/
+:attr:`~iris.experimental.ugrid.Mesh.face_node_connectivity` and
+:attr:`~iris.experimental.ugrid.Mesh.edge_coords`/
+:attr:`~iris.experimental.ugrid.Mesh.face_coords`
+to produce a :class:`~iris.coords.Coord`
+:attr:`~iris.coords.Coord.points` and :attr:`~iris.coords.Coord.bounds`
+representation of all the :class:`~iris.experimental.ugrid.Mesh`\'s
+nodes/edges/faces for the given axis.
+
+A :class:`~iris.experimental.ugrid.Mesh`
```
pp-mo, line 319: Reads a bit awkwardly.  I think "The method :meth:`~iris.experimental.ugrid.Mesh.to_MeshCoords` is available ... *given* only the ``location`` argument."

docs/src/further_topics/ugrid/partner_packages.rst
```diff
@@ -0,0 +1,94 @@
...
+advantage of having a specific use-case - regridding Iris
+:class:`~iris.cube.Cube`\s - to provide ESMPy-Iris wrappers that make the
+process as easy as possible, with highly optimised performance.
+
+Applications
+------------
+* Regrid structured to unstructured.
+* Regrid unstructured to structured.
+* Regrid while maintaining data laziness.
+* Save a prepared regridder for re-use in subsequent runs.
```
pp-mo, line 90: For me, I think the important feature of iris-esmf-regrid over plain ESMF is not just that the regridders are re-usable, but they are also faster.  Because IIUC, we use ESMF to calculate weights, but we can perform the lazy array multiplications faster elsewhere.
Does this make sense @stephenworsley ?

docs/src/further_topics/ugrid/operations.rst
```diff
@@ -0,0 +1,228 @@
...
+    >>> from iris.coords import AuxCoord
+    >>> from iris.experimental.ugrid import Connectivity, Mesh
+
+    # Going to create the following mesh
+    #  (node indices are shown to aid understanding):
+    #
+    #  0----1
+    #  |    |\
+    #  | +  |+\
+    #  2----3--4
```
pp-mo, line 44: Could you not use the very-similar example in UGRID itself here ?
It's in [this section](https://ugrid-conventions.github.io/ugrid-conventions/#2d-flexible-mesh-mixed-triangles-quadrilaterals-etc-topology) - but you have to scroll down to the CDL example, as they don't provide many anchor points on the page.

tkknight:
> `operations.rst` is likely to get rather large, given the agreed effectiveness of example code. To compensate for this I would quite like to introduce the [Sphinx `collapse` directive](https://sphinx-toolbox.readthedocs.io/en/stable/extensions/collapse.html), but I believe it would be the first third party extension we've used so I'm not sure if that's OK.
> 
> Any input @tkknight?

Sorry, I missed the original mention.  It is unclear if the extension needs to be specifically added as a dependency or just used.  Either way aslong as it the env can be created, ideally using conda then yes, lets try the collapse extension.

For reference you can also use panels for big expandable blocks, like we have used on the [whats new](https://scitools-iris.readthedocs.io/en/latest/whatsnew/latest.html) page.

trexfeathers:
> For reference you can also use panels for big expandable blocks, like we have used on the [whats new](https://scitools-iris.readthedocs.io/en/latest/whatsnew/latest.html) page.

Thanks for pointing that out @tkknight. The panels have served my purpose nicely (73f3f1b221ff59be5d102b4413da4be48afae22a) without me needing to introduce any more extensions.

docs/src/further_topics/ugrid/data_model.rst
```diff
@@ -0,0 +1,328 @@
...
+or **volumes**. These elements are constructed by connecting nodes together
+using a 2-dimensional **connectivity** array. One dimension varies over each
+element, while the other dimension varies over each node that makes up that
+element; the values in the array are the node indices. E.g. we could make 2
+square faces from 6 nodes using this ``face_node_connectivity``:
+``[[0, 1, 3, 2], [2, 3, 5, 4]]``. Remember that meshes are **unstructured**, so
+there is no significance to the order of the faces in the array. Data assigned
+to a higher dimensional location must be stored in a 1D array of equal length
+to the number of elements in the mesh, e.g. ``my_data = [0.33, 4.02]`` for our
+example.
```
wjbenfold, line 82: I was tracking well through this page, then this paragraph completely stumped me - the concept of "node" and "element" weren't solid enough yet that I could work out what was happening.
If it's rows and columns, can we refer to rows and columns? E.g. "Each row describes an element (edge, face or volume), by giving the nodes (points) that make it up." I think that "we could make 2 square faces, each from 4 nodes" would be much easier to understand for the example too, because otherwise I have to work out where the 6 maps to, which isn't immediately obvious.

On "Remember", I'd split to a new paragraph so that I can have a break as a user by being satisfied that I got the previous concept. I'd then reword to something like

> Data assigned to the elements described by a **connectivity** array must be stored in a 1D array with one datum per element, e.g. ``my_data = [0.33, 4.02]`` for our example because we have two square elements. Remember that meshes are **unstructured**, so there is no significance to the order of the faces in the array.

docs/src/further_topics/ugrid/data_model.rst
wjbenfold: Is the vague highlighting of the top-left point intentional?
I think I'd prefer it wasn't highlighted, and just one in the middle was? I'd also add some form of background colouring so that it's clearer what's the data, what's the objects that are part of the cube and what's just conceptual process.

docs/src/further_topics/ugrid/data_model.rst
```diff
@@ -0,0 +1,328 @@
...
+
+.. note::
+
+        Connectivities also exist to connect the higher dimensional elements,
+        e.g. ``face_edge_connectivity``. These are optional conveniences to
```
wjbenfold, line 87: Do we not hit problems where only having node-face connectivity without node-edge and edge-face means that we define faces ambiguously because we could reorder the nodes used? Or does order in the connectivity array define that for us? Also, how do we define "inside"?

docs/src/further_topics/ugrid/data_model.rst
wjbenfold: Ah, I see now that it was. Can we make the two highlight types more distinct?

docs/src/further_topics/ugrid/data_model.rst
```diff
@@ -0,0 +1,328 @@
...
+   :download:`full size image <images/data_ugrid_mesh.svg>`
+
+----
+
+The mesh model also supports edges/faces/volumes having associated 'centre'
+coordinates - to allow point data to be assigned to these elements. 'Centre' is
+just a convenience term - the points can exist anywhere within their respective
+elements.
+
+.. figure:: images/ugrid_element_centres.svg
```
wjbenfold, line 104: I'd find it easier with a key here that tells me those are the "centres" and the "nodes". Also somewhere, maybe in the text, link those centres to the "face_coordinates" so I don't have to count to work out what's linked to what?

docs/src/further_topics/ugrid/data_model.rst
```diff
@@ -0,0 +1,328 @@
...
+   :alt: Diagram demonstrating mesh face-centred data.
+
+   Data can be assigned to mesh edge/face/volume 'centres'
+
+Mesh Flexibility
```
wjbenfold, line 109: This feels different to the sections above - we're out of "what's going on" and into "why is it better". Maybe this moves into "What this means"?

docs/src/further_topics/ugrid/data_model.rst
```diff
@@ -0,0 +1,328 @@
...
+----------
+..
+    The diagram images are SVG's, so editable by any graphical software
+     (e.g. Inkscape). They were originally made in MS PowerPoint.
+
+    Uses the IBM Colour Blind Palette (see
+     http://ibm-design-language.eu-de.mybluemix.net/design/language/resources/color-library
+     )
+
+Structured Grids
```
wjbenfold, line 44: I'd appreciate this having a subtitle or something, e.g. "Structured Grids (the old world)" to contrast with "Unstructured meshes (the new world)" so that I know I'm not expecting surprises in the structured grids bit (given I'm not familiar with referring to them as such - to me they're just how the Iris world works).

docs/src/further_topics/ugrid/data_model.rst
```diff
@@ -0,0 +1,328 @@
...
+The Iris :class:`~iris.cube.Cube` has several new members:
+
+* | :attr:`~iris.cube.Cube.mesh`
+  | The :class:`iris.experimental.ugrid.Mesh` that describes the
+    :class:`~iris.cube.Cube`\'s horizontal geography.
+* | :attr:`~iris.cube.Cube.location`
+  | ``node``/``edge``/``face`` - the mesh element type with which this
+    :class:`~iris.cube.Cube`\'s :attr:`~iris.cube.Cube.data` is associated.
+* | :meth:`~iris.cube.Cube.mesh_dim`
+  | The :class:`~iris.cube.Cube`\'s **unstructured dimension** - the one that
```
wjbenfold, line 231: This is the first use of the phrase "unstructured dimension" and is near the bottom of the page. Can we define this earlier (it might make some of the explanation easier to define this early then use it, and will make it less unexpected here)?

wjbenfold (commented):
Done first two pages - I'm broadly amazed how well you've got the concepts across. I've aimed to highlight where I think there's more significant gains to be made or where I got actually confused, rather than minor wording/formatting that I think is less important for getting these docs out there.

docs/src/further_topics/ugrid/partner_packages.rst
trexfeathers: I have made some mentions of the performance enhancements, but perhaps not enough. I'm totally happy to change if anyone can suggest the specific phrasing that would be better.

docs/src/further_topics/ugrid/data_model.rst
trexfeathers: a0b405c3b4432e6e172616620cb6a1563b083a57

docs/src/further_topics/ugrid/data_model.rst
trexfeathers: a0b405c3b4432e6e172616620cb6a1563b083a57

docs/src/further_topics/ugrid/data_model.rst
trexfeathers: a0b405c3b4432e6e172616620cb6a1563b083a57

docs/src/further_topics/ugrid/data_model.rst
trexfeathers: >Also somewhere, maybe in the text, link those centres to the "face_coordinates" so I don't have to count to work out what's linked to what?

Aren't the labelled array sizes and the name `face_coordinates` enough for that?

docs/src/further_topics/ugrid/data_model.rst
trexfeathers: Are you thinking it would be better to rename this section to `Why use Meshes?` and promote it to the higher heading level?

trexfeathers: Scratch that, it definitely belongs in a subheading concerning `The Detail`. I never meant to imply that `The Detail` should be limited to "what's going on" - I just wanted to isolate the deep explanation so that those who want the 'cliff notes' can skip it.

docs/src/further_topics/ugrid/data_model.rst
wjbenfold: I thought from `face_coordinates` that I'd be getting the corners, not the centre. I also initially thought you'd made an error only having 12 not 15, until I worked it out.

If you're worried it'll clog the image or add confusion then it's probably fine to leave out, as it can be worked out, but it took some spare brain capacity to do so which meant I couldn't use it all on the overall concepts.

trexfeathers: That's a tricky compromise. I have indeed been trying to keep the diagrams sparse - both for clarity and also to require readers to 'engage' with what they are seeing - but whether that's sensible is another matter...

docs/src/further_topics/ugrid/data_model.rst
trexfeathers: >Or even that the face coords always lie strictly within the face ??

Yes I had that same moment of pause while writing the docs 😅

trexfeathers: I can address enough of the uncertainty here by mentioning the anti-clockwise thing.

docs/src/further_topics/ugrid/data_model.rst
trexfeathers: I've had a go at addressing both:

* The distinction between arrays and geography - by including an Africa outline to indicate geography.
* The highlighting of two distinct points - bolder dash markings for the 'secondary' point.

3c0e5f85bff8e008226dc9d8a88da47b9b6f4db1

docs/src/further_topics/ugrid/data_model.rst
trexfeathers: 3c0e5f85bff8e008226dc9d8a88da47b9b6f4db1

docs/src/further_topics/ugrid/data_model.rst
trexfeathers: Have put simple labels against one face coordinate and one node coordinate.

3c0e5f85bff8e008226dc9d8a88da47b9b6f4db1

docs/src/further_topics/ugrid/data_model.rst
trexfeathers: 3c0e5f85bff8e008226dc9d8a88da47b9b6f4db1

docs/src/further_topics/ugrid/data_model.rst
trexfeathers: 3c0e5f85bff8e008226dc9d8a88da47b9b6f4db1

docs/src/further_topics/ugrid/data_model.rst
trexfeathers: 3c0e5f85bff8e008226dc9d8a88da47b9b6f4db1

docs/src/further_topics/ugrid/partner_packages.rst
```diff
@@ -0,0 +1,94 @@
+.. _ugrid partners:
+
+Iris' Mesh Partner Packages
+****************************
+Python is an easy to use language and has formed a very strong collaborative
+scientific community, which is why Iris is written in Python. But it is not as
+computationally strong as other languages, and this is a problem for working
+with unstructured meshes where computations are higher in volume and complexity
+(:ref:`see the data model detail <ugrid implications>`). So there are
+operations - which Python can perform acceptably on a structured grids - which
```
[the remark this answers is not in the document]
trexfeathers: 3c0e5f85bff8e008226dc9d8a88da47b9b6f4db1

docs/src/further_topics/ugrid/data_model.rst
```diff
@@ -0,0 +1,361 @@
...
+
+Indexing a mesh data array cannot be used for:
+
+#. Region selection
+#. Neighbour identification
+
+This is because - unlike with a structured data array - relative position in
+a mesh's 1-dimensional data arrays has no relation to relative position in
+space. We must instead perform specialised operations using the information in
+the connectivities present, or by translating the mesh into a format designed
```
lbdreyer, line 224: I may be reading this incorrectly, but I would have thought this should be:

> We must instead perform specialised operations using the information **present** in the connectivities

docs/src/further_topics/ugrid/data_model.rst
```diff
@@ -0,0 +1,361 @@
...
+
+These members will all be ``None`` for a :class:`~iris.cube.Cube` with no
+associated :class:`~iris.experimental.ugrid.Mesh`.
+
+This :class:`~iris.cube.Cube`\'s unstructured dimension has multiple attached
+:class:`iris.experimental.ugrid.MeshCoord`\s (one for each axis e.g.
+``x``/``y``), which can be used to infer the points and bounds of any index on
+the :class:`~iris.cube.Cube`\'s unstructured dimension.
+
+.. todo: Cube printout
```
lbdreyer, line 272: This was the one thing I felt was missing from this page so this will be a great addition

docs/src/further_topics/ugrid/data_model.rst
lbdreyer: Considering there is a lot going on in this, I think it would be quite difficult for a user to follow this diagram without a bit of prose explaining
 what is going on. Maybe include something like the following in the figure caption:

> 1 dimensional Coordinate Arrays, x and y (pink circles)  are combined to construct a structured grid of points (pink crosses). If bounds arrays (blue circles) are present, the x and y bounds construct 4-sided cells (outlined with blue). The data array (orange circles) assigns data with the points (orange diamonds) if no bounds are present) or with the bounded cells (orange quadrialterals) when bounds exist. The above example highlighs the location (x=25, y=20) which has bounds (x=[23,28], y=[19,21]) 

I would also think the other figures would benefit from a more detailed caption

docs/src/further_topics/ugrid/data_model.rst
```diff
@@ -0,0 +1,361 @@
...
+
+* Every node is completely independent - every one can have unique X and
+  Y (and Z) coordinate values.
+
+.. figure:: images/ugrid_node_independence.svg
+   :alt: Diagram demonstrating the independence of each mesh node
+   :align: center
+   :width: 300
+
+   Every mesh node is completely independent
```
lbdreyer, line 150: I don't understand what this figure is intending to represent. Particularly the arrows. Are they to depicting the transition from old world (structured grids) to new world (unstructured grids)
image

trexfeathers: I wanted to introduce the new concepts gradually. So the initial introduction of the 'new world' was replicating the 'old world' example. I wanted to demonstrate that arrays of the **_same shape and size_** could also represent something unstructured.

I hope you'll agree that's an important step. But presumably there's a better way to represent this (or maybe just including the explanation I've just given you)?

docs/src/further_topics/ugrid/data_model.rst
trexfeathers: I was actually thinking that if there were more connectivities than the bare minimum (e.g. `face_edge_connectivity`), it would be possible to write something that quickly finds neighbours or maybe identifies regions.

This will probably do:

```suggestion
the mesh's connectivities, or by translating the mesh into a format designed
```

docs/src/further_topics/ugrid/data_model.rst
jamesp: > I've been trying to keep the prose and diagrams self-contained - for accessibility and ease of maintenance.
> 
> And I had also hoped that the diagrams would be self-describing where possible. But from yours and others' feedback there are parts that aren't as obvious as I had thought! I'll see what I can think of to make things clearer within the diagrams, while avoiding information overload.

I understand that rationale, makes a lot of sense so feel free to ignore this advice if you feel those concerns trump this.

I know this isn't a paper or publication, but I was taught (and I think it is a good general rule of thumb) that all figures in a document should be referenced from the text, otherwise what is the context, why are they there?
A brief search for a guide to this I found the APA style guide, see https://guides.lib.monash.edu/citing-referencing/apa-tables-figures:

> * All Tables and Figures must be referred to in the main body of the text.

docs/src/further_topics/ugrid/partner_packages.rst
stephenworsley: So I don't think I could say in full confidence that there is a performance improvement in every case. ESMF seems to have its own way to handle parallelism which I could never quite figure out. I will say that there are difinitely circumstances where we are able to provide significantly better performance. I think one of the biggest improvements is the fact that a regridder can be reused on sources with different masks, where in ESMF you would have to recalculate the weights every time. I believe that regridding sources with large non-horizontal coordinates also performs better (at the moment, this is the only case where parallelism is possible), though performance here also depends on how the array is chunked.

In terms of what is worth adding here as a feature, I think "Regrid while maintaining data laziness" already sort of implies dask integration. Maybe the performance could be emphasised as: "Regrid with dask integration, computing in parallel and maintaining data laziness."
The handling of masked data could also be emphasised with a bullet point something like: "Improved handling of masked data", If you think this is the place to expand on this you could add "(unlike ESMF, you don't need to have the same mask for all data given to an initialised regridder)."

docs/src/further_topics/ugrid/data_model.rst
trexfeathers: We're gonna go with some more explicit captions on each figure, and I'll think about some cross-referencing with the text blocks.

docs/src/further_topics/ugrid/data_model.rst
trexfeathers: https://github.com/SciTools/iris/pull/4504#discussion_r793486314

docs/src/further_topics/ugrid/data_model.rst
wjbenfold: > represented by either the coordinate or connectivity arrays detailed above

This bit of the `important` block doesn't make sense to me. My attempt to make this make sense below:

Elements are enumerated along a single unstructured dimension (present in the coordinate or connectivity arrays detailed above). An element’s position along this dimension has nothing to do with its spatial position.

docs/src/further_topics/ugrid/partner_packages.rst
```diff
@@ -0,0 +1,98 @@
+.. _ugrid partners:
+
+Iris' Mesh Partner Packages
+****************************
+Python is an easy to use language and has formed a very strong collaborative
+scientific community, which is why Iris is written in Python. But it is not as
+computationally strong as other languages, and this is a problem for working
+with unstructured meshes where computations are higher in volume and complexity
+(:ref:`see the data model detail <ugrid implications>`). So there are
```
wjbenfold, line 9: I'm not sure I'm comfortable saying this about Python - naively written Python is slow, and calls down to C are a way that Python combines speed with convenience. I'd say we're continuing a known strength of Python rather than working around a weakness (though I guess people working with meshes are probably fairly sold on Python so maybe I don't need to worry). I'm happy to write a variant of this if that would be useful

docs/src/further_topics/ugrid/operations.rst
```diff
@@ -0,0 +1,791 @@
+.. _ugrid operations:
+
+Working with Mesh Data
+***********************
+
+.. note:: Several of the operations below rely on the optional dependencies
+          mentioned in :doc:`partner_packages`.
+
```
wjbenfold, line 8: Title? "Operation summary"? Would make it easier to link to when people ask us about them too

docs/src/further_topics/ugrid/operations.rst
```diff
@@ -0,0 +1,791 @@
...
+      - |tagline: equality|
+    * - `Combining Cubes`_
+      - |tagline: combining cubes|
+    * - `Arithmetic`_
+      - |tagline: arithmetic|
+
+..
+    Below: use demo code over prose wherever workable. Headings aren't an
+     exhaustive list (can you think of any other popular operations?).
+
```
wjbenfold, line 37: All the way through the below, I would reduce the number of links out to the rest of the documentation. It may be that I read documentation oddly, but my tendency is to Ctrl-click out on each link so I can learn more in a minute. The repeated linking to "Cube" and "Mesh" makes this hard and breaks up readability. I'd suggest linking to things the first time that they're referenced in a given operation but not afterwards (when I'd just use the `stuff` quotes).
This would make it easier to spot when something new is referenced. It's not going to affect conceptual understanding though, so happy if you ignore this in the name of time.

docs/src/further_topics/ugrid/data_model.rst
trexfeathers: https://github.com/SciTools/iris/pull/4504/commits/3692c1cfa51f15181437b22088f3f223d0037a93

docs/src/further_topics/ugrid/data_model.rst
trexfeathers: 3692c1cfa51f15181437b22088f3f223d0037a93

docs/src/further_topics/ugrid/data_model.rst
trexfeathers: 3692c1cfa51f15181437b22088f3f223d0037a93

docs/src/further_topics/ugrid/data_model.rst
trexfeathers: 3692c1cfa51f15181437b22088f3f223d0037a93

docs/src/further_topics/ugrid/operations.rst
trexfeathers: I think on balance I'd like to keep `Working with Mesh Data`.

docs/src/further_topics/ugrid/operations.rst
trexfeathers: >ignore this in the name of time

I think I might have to, that's rather fiddly! It's an interesting question though, I've so far adopted the approach that **any** reference to the API should use a Sphinx domain, but perhaps others feel differently @SciTools/iris-devs?

docs/src/further_topics/ugrid/partner_packages.rst
trexfeathers: a31546293bde29313aa7d24dd647ac27476c6604

docs/src/further_topics/ugrid/data_model.rst
trexfeathers: a31546293bde29313aa7d24dd647ac27476c6604

docs/src/whatsnew/latest.rst
trexfeathers: a31546293bde29313aa7d24dd647ac27476c6604

docs/src/further_topics/ugrid/operations.rst
wjbenfold: Oh, sorry, I meant add an extra subtitle for this table so I know what I'm looking at (under the `Working with Mesh Data` heading).

docs/src/further_topics/ugrid/operations.rst
wjbenfold: I tend to look out for this since I really struggled reading this page of the docs: https://scitools-iris.readthedocs.io/en/stable/developers_guide/contributing_getting_involved.html - I'd be interested to know it other people find something similar.

trexfeathers: >I tend to look out for this since I really struggled reading this page of the docs: https://scitools-iris.readthedocs.io/en/stable/developers_guide/contributing_getting_involved.html

FWIW I'm with you on that one.

I guess I must have a different subconscious contract when I see an API object - I see it mostly as formatting and don't feel the same obligation/compulsion to click as with a normal link.

wjbenfold: That's fair - it does help that `Cube` always looks like `Cube` rather than being replaced by other text that hides that it's the same link

docs/src/further_topics/ugrid/operations.rst
trexfeathers: ef47b6b7d0aad7ec7e5b28ea5375f1cfae9db106

docs/src/further_topics/ugrid/partner_packages.rst
trexfeathers: 20113ca5305a99af4e0e3e3ce76d83729dfe9705

trexfeathers:
I believe I've addressed everyone's comments, either by making a change or by replying. I consider this ready to go.

docs/src/further_topics/ugrid/operations.rst
```diff
@@ -0,0 +1,953 @@
...
+          N Cells:	2
+          N Points:	5
+          X Bounds:	9.903e-01, 1.000e+00
+          Y Bounds:	0.000e+00, 1.392e-01
+          Z Bounds:	6.123e-17, 5.234e-02
+          N Arrays:	2
+
+        # Create the GeoVista plotter and add our mesh+data to it.
+        >>> my_plotter = GeoPlotter()
+        >>> my_plotter.add_coastlines(color="black")
```
bjlittle, line 482:
```suggestion
        >>> my_plotter.add_coastlines()
```
@trexfeathers The associated rendered image has `white` coastlines (default), and yet you specify `black` here in the example... I guess the code and image have got out of sync.

trexfeathers: Tut tut @trexfeathers 

4c2b19a76783d5dd5ca6d1af91e1887bb59b5058

docs/src/further_topics/ugrid/data_model.rst
```diff
@@ -0,0 +1,565 @@
...
+    :alt: Diagram demonstrating mesh face-centred data.
+    :align: right
+    :width: 1280
+
+    Data can be assigned to mesh edge/face/volume 'centres'
+
+    1D *node* coordinate arrays (pink circles) describe node positions in
+    space (pink crosses). A 2D connectivity array (blue circles) describes
+    faces by connecting four nodes into a face outline (blue outlines on the
+    map). Further 1D *face* coordinate arrays (pink circles) describe a
```
pp-mo, line 168: Sorry to be late to the table here, but I think it seems unnecessarily confusing here to use the same colour for face coordinates as you do for node coordinates.  Is there a particular logic to the colour coding, which I've missed, which forces that choice ?

trexfeathers: 1. Using more than 3 colours from the [IBM colour-blind palette](https://davidmathlogic.com/colorblind/#%23648FFF-%23785EF0-%23DC267F-%23FE6100-%23FFB000) puts them too close together under some types of colour-blindness.
2. Although number 1 was my original reason, I also like that they're both coordinates and therefore are more similar to each other than the other concepts in the diagrams.

trexfeathers: There are perhaps other palettes that allow four distinct colours but I wanted to pick one that was well recognised.

docs/src/further_topics/ugrid/data_model.rst
```diff
@@ -0,0 +1,565 @@
...
+                    <AuxCoord: latitude / (degrees_north)  [...]  shape(2,)>
+            long_name: 'my_mesh'
+
+* | :class:`iris.experimental.ugrid.MeshCoord`
+  | Described in detail in `MeshCoords`_.
+  | Stores the following information:
+
+    * | :attr:`~iris.experimental.ugrid.MeshCoord.mesh`
+      | The :class:`~iris.experimental.ugrid.Mesh` associated with this
+        :class:`~iris.experimental.ugrid.MeshCoord`. Mirrored by the
```
pp-mo, line 485: Might it be clearer to say
In place of "Mirrored by " ... the  /attr/ attribute of any Cube this is attached to.
Could have
 "This becomes"... or "This defines" ... or "This determines"

docs/src/further_topics/ugrid/operations.rst
```diff
@@ -0,0 +1,953 @@
...
+        ... )
+
+        >>> my_mesh = Mesh(
+        ...     long_name="my_mesh",
+        ...     topology_dimension=2,  # Supports 2D (face) elements.
+        ...     node_coords_and_axes=[(node_x, "x"), (node_y, "y")],
+        ...     connectivities=[edge_node_c, face_node_c],
+        ...     edge_coords_and_axes=[(edge_x, "x"), (edge_y, "y")],
+        ...     face_coords_and_axes=[(face_x, "x"), (face_y, "y")],
+        ... )
```
pp-mo, line 111: Is it worth also printing the Mesh out at this point ?

docs/src/further_topics/ugrid/operations.rst
```diff
@@ -0,0 +1,953 @@
...
+        ...            units="K",
+        ...            dim_coords_and_dims=[(vertical_levels, 1)],
+        ...            aux_coords_and_dims=[(mesh_coord_x, 0), (mesh_coord_y, 0)],
+        ...        )
+        ...    )
+
+        >>> print(my_cubelist)
+        0: edge_data / (K)                     (-- : 6; height: 3)
+        1: face_data / (K)                     (-- : 2; height: 3)
+
```
pp-mo, line 156: Is it worth also printing a cube at this point, to show how MeshCoords map to the dims?

trexfeathers: d2e504f66a1f46335ae441c9494956ffb2aafb09

docs/src/further_topics/ugrid/operations.rst
trexfeathers: d2e504f66a1f46335ae441c9494956ffb2aafb09

docs/src/further_topics/ugrid/data_model.rst
trexfeathers: 8bc8c8488f912d10662f26edf94bff6c315df0e0

docs/src/further_topics/ugrid/data_model.rst
trexfeathers: Also I hadn't spotted that the example coordinates were unbelievable in a couple of ways.

67b1eb3765829eb70b5d4f2e581a8b8dde09b498

docs/src/further_topics/ugrid/operations.rst
```diff
@@ -0,0 +1,953 @@
...
+          Z Bounds:	6.123e-17, 5.234e-02
+          N Arrays:	2
+
+        # Create the GeoVista plotter and add our mesh+data to it.
+        >>> my_plotter = GeoPlotter()
+        >>> my_plotter.add_coastlines(color="black")
+        >>> my_plotter.add_base_layer(color="grey")
+        >>> my_plotter.add_mesh(face_polydata)
+
+        # Centre the camera on the data.
```
pp-mo, line 518: These lines here about the camera are just touching over the boilerplate tolerance level, for me.
So I hope this isn't done for showiness, but rather is essential to ensure the resulting plot is useful out of the box ??
If there was any way of reducing the "extra stuff" a bit here, I'd really like that

pp-mo: I see that the next example in line is much cleaner-looking.

trexfeathers: Yes, that's why it's there.

Are you basically saying you'd prefer it if we set the camera position on a single line, using raw coordinates?

docs/src/further_topics/ugrid/operations.rst
```diff
@@ -0,0 +1,953 @@
...
+        >>> print(indices[:10])
+        [ 6  7  8  9 10 11 18 19 20 21]
+
+        # Use the face indices to subset the global cube.
+        >>> region_cube = global_cube[:, indices]
+
+        # In this case we **know** the indices correspond to a contiguous
+        #  region, so we will convert the sub-setted Cube back into a
+        #  Cube-with-Mesh.
+        >>> new_mesh = Mesh.from_coords(*region_cube.coords(dimensions=1))
```
pp-mo, line 713: This seems almost a "sly" introduction of from_coords.
Which is nevertheless a potentially really useful operation to the user.
Is there anyway of highlighting this a bit more, I wondered ?

docs/src/further_topics/ugrid/data_model.rst
```diff
@@ -0,0 +1,565 @@
...
+.. todo:
+        a third example!
+
+Mesh 'payload' is much larger than with structured grids
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+Coordinates are recorded per-node, and connectivities are recorded per-element.
+This is opposed to a structured grid, where a single coordinate value is shared
+by every data point/area along that line.
+
+For example: representing a cubed-sphere using a mesh leads to coordinates and
```
stephenworsley, line 252: I think it's worth noting here that this is true when you are representing a cubed-sphere *with no other coordinates*.

docs/src/further_topics/ugrid/operations.rst
pp-mo: I'm not really capable of making any concrete proposal here  -- I'm just a little disappointed that the very first example makes it look so complicated

pp-mo: Maybe you can just pull out one specific point + centre it on that ??

pp-mo: ... I don't really understand the whole "region" thing here though

docs/src/further_topics/ugrid/operations.rst
```diff
@@ -0,0 +1,985 @@
...
+        >>> new_mesh_coords = new_mesh.to_MeshCoords(global_cube.location)
+        >>> for coord in new_mesh_coords:
+        ...     region_cube.remove_coord(coord.name())
+        ...     region_cube.add_aux_coord(coord, 1)
+
+        # A Mesh-Cube with a subset (101) of the original 864 faces.
```
pp-mo, line 719: This seems like a heroic success -- it's pretty neat for such a complicated operation.
I think it might improve the impact to also plot it
 -- though I guess that is cross-linking the example categories, in a way you might not favour ?

docs/src/further_topics/ugrid/partner_packages.rst
```diff
@@ -0,0 +1,99 @@
+.. _ugrid partners:
+
+Iris' Mesh Partner Packages
+****************************
+Python is an easy to use language and has formed a very strong collaborative
+scientific community, which is why Iris is written in Python. *Performant*
+Python relies on calls down to C, which is ideal for structured grid work since
```
stephenworsley, line 7: Since ESMF uses Fortran, this could be better as "Python relies on calls down to low level languages like C".

docs/src/further_topics/ugrid/operations.rst
```diff
@@ -0,0 +1,985 @@
...
+
+Regridding to or from a mesh requires different logic than Iris' existing
+regridders, which are designed for structured grids. For this we recommend
+ESMF's powerful regridding tools, which integrate with Iris' mesh data model
+via the :ref:`ugrid iris-esmf-regrid` package.
+
+.. todo: inter-sphinx links when available.
+
+Regridding is achieved via the
+:class:`esmf_regrid.experimental.unstructured_scheme.MeshToGridESMFRegridder`
```
pp-mo, line 754: Some of this is sort-of boilerplate, as all Iris regrid schemes basically work the same way.
I also think we need to emphasize that the regridders **_are_** regular Iris "schemes", so the operation of them should be familiar if you've used Iris regridding before
 -- i.e. you don't actually need to know anything much more, except that iris-esmf-regrid holds the functionality you need .  And maybe not to mess about with the olders ones, but I think you already made that clear enough.

docs/src/further_topics/ugrid/operations.rst
trexfeathers: I noticed last year that our demonstrating this via plots was starting to create a false association that _all_ GeoVista operations could take a lot of resource / need a GPU because that's true of plotting.

So here I wanted to deliberately break any possible association by demonstrating GeoVista operations without a plot.

It's also a lot more real estate.

docs/src/further_topics/ugrid/operations.rst
trexfeathers: @stephenworsley help me

docs/src/further_topics/ugrid/operations.rst
```diff
@@ -0,0 +1,985 @@
...
+                Conventions                 UGRID
+                description                 Created by xios
+                interval_operation          300 s
+                interval_write              1 d
+                name                        lfric_surface
+                online_operation            average
+                timeStamp                   2020-Feb-07 16:23:14 GMT
+                title                       Created by xios
+                uuid                        489bcef5-3d1c-4529-be42-4ab5f8c8497b
+
```
pp-mo, line 847: Again, it seems a slight shame not to _plot_ results, to show before+after,
but again it mixes up the example topics, which I guess you wanted to rule out ?

stephenworsley (changes requested):
Looking good, just some very small things, but otherwise I think this is looking very close.

docs/src/further_topics/ugrid/operations.rst
trexfeathers: I've looked at this some more and the alternative would be:

```python
my_plotter.camera_position = [3, 0.183, 0.079]
```

Which, while shorter, is unhelpful for the user when replicating with their own data. I expect camera centre-ing to be a very common operation with any non-global data, and until GeoVista includes conveniences for it I feel it's important to include lines that show how to generically achieve this, so I'm not going to action this one.

docs/src/further_topics/ugrid/operations.rst
trexfeathers: @stephenworsley thoughts?

docs/src/further_topics/ugrid/operations.rst
trexfeathers: OK thanks. As it stands I'm not planning to make any changes (?)

docs/src/further_topics/ugrid/operations.rst
trexfeathers: 970db3ca5ea0b1c32a3ae7730b51a1df7db0d847

docs/src/further_topics/ugrid/partner_packages.rst
trexfeathers: be5f51bc6a379de5f020ea418f0c88f57ecf5527

docs/src/further_topics/ugrid/data_model.rst
trexfeathers: 3d18be7cdcd8937d45bec9fd83046b2c8dc497d0

docs/src/further_topics/ugrid/operations.rst
trexfeathers: TBH this might add some value but in the interests of time this feels like an 80-20 thing that we could get away without.

stephenworsley: yeah, pictures would be nice but no necessarily essential, and I think putting in the code to print them could distract from the regridding process.

stephenworsley (approved):

docs/src/further_topics/ugrid/operations.rst
pp-mo: > manually keeping the images and code in sync,

:+1: Ok I didn't realise how this was being done.

I'm beginning to think this stuff should really be maintained somewhere else, with all the necessary dependencies pulled in, and perhaps using notebooks too.
A problem for another day !

PR #4510, opened 2022-01-18
Turn off documentation type hints

rcomer:
### Description

From https://github.com/SciTools/iris/discussions/4383#discussioncomment-1917571 and replies, there seems to be a consensus that we don't want to display type annotations in the docs.  There are not many examples in Iris so far, but for [recombine_submeshes](https://scitools-iris.readthedocs.io/en/latest/generated/api/iris/experimental/ugrid/utils.html#iris.experimental.ugrid.utils.recombine_submeshes) this PR turns

with_type_hint
 into
without_type_hint

---

### DIFF
--- a/docs/src/conf.py
+++ b/docs/src/conf.py
@@ -209,6 +209,7 @@ todo_include_todos = True
 # api generation configuration
 autodoc_member_order = "groupwise"
 autodoc_default_flags = ["show-inheritance"]
+autodoc_typehints = "none"
 autosummary_generate = True
 autosummary_imported_members = True
 autopackage_name = ["iris"]


bjlittle (approved):
@rcomer Agreed. We need to consider this a tad more as for me it renders the docs very hard to read

pp-mo:
I've been (elsewhere) trialling the option `autodoc_typehints = "description"` 
Not sure if that option post-dates this PR ??
It puts the type-hints into the parameter / returns section, but not the signature, which in practice is much more readable IMHO.

----
Using this, you can omit the type definitions in the docstring, and sphinx-napoleon will then fill in from the type hints.
For example ...
```
def move_one_dimension(content: np.ndarray, from_dim: int, to_dim: int) -> np.ndarray:
    """
    Move one dimension of an array.

    Transpose the input, to shift the selected dimension from one place to another in
    the index order.

    Parameters
    ----------
    content
        an array to be transposed
    from_dim
        the dimension number to 'move'.  ``-1`` means the last
    to_dim
        index of the 'moved' dimension within the output.
        ``0`` means to the first dimension, ``-1`` to the last.

    Returns
    -------
    array
        the result of an appropriate ``content.transpose()`` operation.
    """
```
Results in ,,,
napoleon_typehinted

**I think this is great, as it needs only one DRY statement of types, and it is in a strict vocabulary based on typing, instead of the rather loose numpy docstring format.**

----

It can go wrong though, as when I type the first arg as "np.typing.ArrayLike", this renders as ...
```_SupportsArray[dtype[Any]] | _NestedSequence[_SupportsArray[dtype[Any]]] | bool | int | float | complex | str | bytes | _NestedSequence[bool | int | float | complex | str | bytes]```

According to the Python command-line, this object prints as 
```typing.Union[numpy._typing._array_like._SupportsArray[numpy.dtype[typing.Any]], numpy._typing._nested_sequence._NestedSequence[numpy._typing._array_like._SupportsArray[numpy.dtype[typing.Any]]], bool, int, float, complex, str, bytes, numpy._typing._nested_sequence._NestedSequence[typing.Union[bool, int, float, complex, str, bytes]]]```

Probably there is a solution to this, but I don't know what as yet.

What I find you **_can_** do is revert to putting ": :class:`~np.typing.ArrayLike`" **_within the docstring_**  (as well as the type hint) :
That "works", then rendering as `content (ArrayLike) – an array to be transposed`
\-- in that, the 'ArrayLike' is then highlighted _but not clickable_.

pp-mo:
More on the above now added [here](https://github.com/SciTools/iris/discussions/4272#discussioncomment-7465178)

Issue #4394, opened 2021-11-01
Iris 3.2.0 Deprecations


bjlittle:
## 📰 Custom Issue
We require to consider and include any functional deprecations that we want to include for the forthcoming `iris` 3.2.0 release. 

**Existing Deprecations**
- [x] [iris.util.as_compatible_shape](https://scitools-iris.readthedocs.io/en/latest/generated/api/iris/util.html#iris.util.as_compatible_shape). Deprecated in 3.0.0 (#4513)

**Proposed New Deprecations for 3.2**
- [x] [iris.util.approx_equal](https://scitools-iris.readthedocs.io/en/stable/generated/api/iris/util.html#iris.util.approx_equal).
     Proposed by @rcomer, see [comment](https://github.com/SciTools/iris/issues/4394#issuecomment-996780194) below. Agreed
     (#4514)
- [x] [iris.fileformats.abf](https://scitools-iris.readthedocs.io/en/latest/generated/api/iris/fileformats/abf.html).
     Proposed by @bjlittle. Agreed (remove - ask community if this is an issue)
     Also see https://github.com/SciTools/iris-abf and https://github.com/SciTools/iris/pull/3657
     (#4515)
- [x] [iris.fileformats.dot](https://scitools-iris.readthedocs.io/en/latest/generated/api/iris/fileformats/dot.html).
     Proposed by @bjlittle. Agreed (remove - as above)
     (#4515)
- [x] iris.experimental.raster. 
     Proposed by @bjlittle. Agreed (remove - as above)
     (#4525)
     (#4545)
- [x] `iris.experimental.regrid_conservative` (and optional dependency `esmpy`).
     Proposed by @pp-mo. Agreed (remove - as above) Functionality available/equivalent in `iris-esmf-regrid`.
     (#4551)
- [x] `iris.experimental.regrid`
    Proposed by @pp-mo. Agreed (this will be moved)
     (#4548)

**Investigate**
- [x] @rcomer proposal for `intersection_of_cubes` vs `extract_overlapping`, see [comment](https://github.com/SciTools/iris/issues/4394#issuecomment-1011240999) below.
     (#4541)

**Related**
- See https://github.com/SciTools/iris/issues/3515

**Further**
[...]

PR #4513, opened 2022-01-20
Removed iris.util.as_compatible_shape and now unused imports

wjbenfold:
### Description

Deprecation as per #4394
Removed unused imports in addition to the deprecated function.

---


lib/iris/tests/test_util.py
```diff
@@ -276,103 +276,5 @@ def test_output_file(self):
...
-
-    def test_collapsed(self):
-        cube = tests.stock.realistic_4d()
-        collapsed = cube.collapsed("model_level_number", iris.analysis.MEAN)
-        expected_shape = list(cube.shape)
-        expected_shape[1] = 1
-        expected_data = collapsed.data.reshape(expected_shape)
-        res = iris.util.as_compatible_shape(collapsed, cube)
-        self.assertCML(
-            res, ("util", "as_compatible_shape_collapsed.cml"), checksum=False
```
lbdreyer, line 312: If nothing else is using this [CML result file](https://github.com/SciTools/iris/blob/main/lib/iris/tests/results/util/as_compatible_shape_collapsed.cml), you could delete it, since there would be no reason to keep it

lib/iris/util.py
```diff
@@ -23,7 +22,6 @@
 import numpy as np
 import numpy.ma as ma
 
-from iris._deprecation import warn_deprecated
```
lbdreyer, line 26: I think a bit of a merge conflict has happened here. It looks like warn_deprecated is now being used since https://github.com/SciTools/iris/pull/4514 was merged in, so this shouldn't be removed.

I think the original tests would have passed since 4514 hadn't been merged in yet? So it didn't catch this

wjbenfold: Ah, I thought that at the time but missed it as a consequence of the merge. Fixed now

lbdreyer (approved):

lbdreyer:
Thanks @wjbenfold !

PR #4514, opened 2022-01-20
Deprecate iris.util.approx_equal

wjbenfold:
### Description

As per #4394 
Will conflict with #4513 so will need a rebase once that's merged

---


pp-mo (approved):
LGTM
But I would like to be sure @lbdreyer has had a chance to comment, before merging.

lbdreyer:
>But I would like to be sure @lbdreyer has had a chance to comment, before merging.

I have no objections to this being removed. It doesn't make sense to live in Iris' codebase and I can't find it being using anywhere within Iris

PR #4515, opened 2022-01-20
Deprecate abf and dot

wjbenfold:
### Description

See #4394 

To do:
- [x] Work out whether this is the right way to deprecate a module - the warning for abf at least currently shows up on import of fileformats, which we may want to avoid. Not sure where to put it instead though - in every function?
- [x] Check wording of message and consider mentioning future mitigations (undo deprecation, use plugin architecture?)
- [x] Fix doctest that's failing due to the deprecation warnings

---

### DIFF
--- a/docs/src/whatsnew/latest.rst
+++ b/docs/src/whatsnew/latest.rst
@@ -195,6 +195,11 @@ This document explains the changes made to Iris for this release
    believed to still be in use. The deprecation warnings invite users to contact
    the Iris Developers if this isn't the case. (:pull:`4525`)
 
+#. `@wjbenfold`_ deprecated :mod:`iris.fileformats.abf` and
+   :mod:`iris.fileformats.dot` as they are not believed to still be in use. The
+   deprecation warnings invite users to contact the Iris Developers if this
+   isn't the case. (:pull:`4515`)
+
 
 🔗 Dependencies
 ===============
--- a/lib/iris/fileformats/__init__.py
+++ b/lib/iris/fileformats/__init__.py
@@ -17,7 +17,7 @@ from iris.io.format_picker import (
     UriProtocol,
 )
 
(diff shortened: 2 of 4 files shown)


pp-mo (changes requested):
Looking good, but I agree we need to avoid the message when "just" importing iris.fileformats.
As with the other deprecations, I'd also like to be sure @lbdreyer had a chance to OK this.

That warning on any use of iris.fileformats happens because Iris imports abf that to populate its file format handler structures.

I think it still makes sense to issue the warning whenever abf is imported, so I think the 
nice way is to use a deferred import within a wrapper function.
I think I will propose something similar as a PR to your branch...

pp-mo:
> I think I will propose something

See https://github.com/wjbenfold/iris/pull/2

wjbenfold:
> > I think I will propose something
> 
> See [wjbenfold#2](https://github.com/wjbenfold/iris/pull/2)

I love the elegance of this solution

lbdreyer:
>As with the other deprecations, I'd also like to be sure @lbdreyer had a chance to OK this.

I'm fine with this. If we are happy to reverse the decision (or perhaps come up with an alternative!) should someone say they are still using it then that is fine.

pp-mo:
hi @wjbenfold 
something weird has happened here -- although you merged my suggestion [wjbenfold#2 ](https://github.com/wjbenfold/iris/pull/2) there is no change included here to `iris.fileformats.__init__.py`
So I think something was wrong with the merge operation + you need to address it to we have what we need here

wjbenfold:
Ah, didn't pull before the rebase, will do that

wjbenfold:
I believe that this is ready now @pp-mo

pp-mo (approved):
Ok now!

PR #4525, opened 2022-01-24
Deprecate experimental.raster

wjbenfold:
### Description

Deprecation as per #4394. If it turns out that this is used, we can reverse the deprecation or make it a plug-in.

---


pp-mo (approved):
LGTM!

PR #4499, opened 2022-01-13
Coord print

pp-mo:
Closes #4488, #4494
Linked to https://github.com/SciTools/iris/discussions/4501

This is just an initial draft version, for discussing the new behaviour + printed output format

**Please review for behaviours (output results), and not the code (yet!)**

Some general notes :
  * I tried to mimic the general form of the `Cube.__str__` output
  * the `__repr__` produces a minimal one-line form, and the `__str__` a detailed multiline printout, like the Cube ones

PR #4528, opened 2022-01-25
Improve Coord and Cube attributes printout

pp-mo:
During development of #4499, and discussion in https://github.com/SciTools/iris/discussions/4501, it was [suggested](https://github.com/SciTools/iris/discussions/4501#discussioncomment-1986399) that when printing cube components with names/units/attributes (i.e. derived from CFVariableMixin) we should be listing any attributes in a one-per-line fashion, as `Cube.__str__` already does.

This wasn't initially done in #4499 as I was concerned it introduced more change to an already complex PR.

On subsequent investigation though, it isn't really that complex, so this PR introduces that.
Results go from looking (schematically) like
```
DimCoord :  name / (units)
 . . .
    long_name: 'thing'
    dtype: float64
    attributes: {'string': 'this', 'array': [1.0 2.0 3.0], 'really_long_string': 'long long long long thing', 'float':3.2}
    circular: True
```
to
```
DimCoord :  thing / (units)
 . . .
    long_name: 'thing'
    dtype: float64
    attributes:
        string     'this'
        array      [1.0 2.0 3.0]
        long_name  'long long long long thing'
        float      3.2
    circular: True
```

Comparing the output of attributes in Cube listings to those in Coords, I also came to the conclusion that it would be better to always quote strings when they appear as attribute values -- that is, in Cube listings also.
Luckily that is very easy to do -- though there were quite a few tests to fix, so it's a bit noisy.


lib/iris/coords.py
```diff
@@ -571,10 +571,23 @@ def reindent_data_string(text, n_indent):
...
+                        # Use a multi-line form for this.
+                        add_output(newline_indent)
+                        add_output("attributes:", section="attributes")
+                        max_attname_len = max(len(attr) for attr in val.keys())
+                        for attrname, attrval in val.items():
+                            attrname = attrname.ljust(max_attname_len)
+                            if isinstance(attrval, str):
+                                # quote strings
+                                attrval = repr(attrval)
+                            attr_string = f"{attrname}  {attrval}"
```
pp-mo, line 584: See comment below...

lib/iris/_representation/cube_summary.py
```diff
@@ -207,7 +207,7 @@ def __init__(self, title, attributes):
         self.values = []
         self.contents = []
         for name, value in sorted(attributes.items()):
-            value = value_repr(value)
+            value = value_repr(value, quote_strings=True)
```
pp-mo, line 210: See other comment below ...

lib/iris/experimental/ugrid/mesh.py
```diff
@@ -1044,7 +1044,19 @@ def line(text, i_indent=0):
...
+                    if name == "attributes":
+                        # Use a multi-line form for this.
+                        line("attributes:", 1)
+                        max_attname_len = max(len(attr) for attr in val.keys())
+                        for attrname, attrval in val.items():
+                            attrname = attrname.ljust(max_attname_len)
+                            if isinstance(attrval, str):
+                                # quote strings
+                                attrval = repr(attrval)
+                            attr_string = f"{attrname}  {attrval}"
```
pp-mo, line 1056: This is essentially the same as the code in _DimensionalMetadata.summary, [above](https://github.com/SciTools/iris/pull/4528/files?file-filters%5B%5D=.py&show-viewed-files=true#r792715108).

So there is potentially an option to combine these approaches, and possibly even with that in CubeSummary [above#2](https://github.com/SciTools/iris/pull/4528/files?file-filters%5B%5D=.py&show-viewed-files=true#r792717644).

It is also notable that we ***could*** be clipping over-long strings here, as Cube.summary does.
Not sure if that is really needed, but if we combine them it would make sense.

pp-mo:
Added code to clip over-long strings, as in previous comments.

lbdreyer (approved):
Looks good! Thanks @pp-mo !

PR #4540, opened 2022-01-26
docs: move whatsnew

bjlittle:
### Description
This PR moves the `What's New in Iris` documentation menu item to a move prevalent position i.e., from buried within the lower `Reference` section up into its own `What's New in Iris` section at the top of the LHS menu panel.

This make it a wee bit more obvious for the reader to find and access. 

Also, as a convenience, this PR also includes a one-click hit that takes the reader to the latest development or stable documentation on RTD, or an archive of all the (dev and stable) doc versions.

A rendered version of the documentation is available [here](https://bjlittle-iris.readthedocs.io/en/latest/) to view e.g.,
image

---
[Consult Iris pull request check list]( https://scitools-iris.readthedocs.io/en/latest/developers_guide/contributing_pull_request_checklist.html)

PR #4548, opened 2022-01-27
Deprecate regrids

pp-mo:
Deprecate everything in `iris.experimental.regrid`, as noted in https://github.com/SciTools/iris/issues/4394

Some of the content has already moved; some will vanish; some can be accessed in a better way.

In this case, as for `iris.experimental.raster`, merely importing it will give a warning.
But I have also put deprecation warnings in the docstrings, **_and_** a runtime warning, for all of the public features.


lib/iris/experimental/regrid.py
```diff
@@ -6,8 +6,18 @@
...
+.. note::
+
+    .. deprecated:: 3.2.0
+
+    This package will be removed in a future release.
+    The PointInCell class has now moved to :class:`iris.analysis.PointInCell`.
+    All the other content will be withdrawn.
+
+    If you still use any of this, please contact the Iris Developers to
+    discuss how to replace it or to retain it.
```
wjbenfold, line 18: I think that this will be triggered by a regrid of a cube via `cube.regrid`, `iris.analysis.AreaWeighted` and

https://github.com/SciTools/iris/blob/9ba2adde25e6a720e6b8afd7f738ebc8663d449a/lib/iris/analysis/_area_weighted.py#L46-L48

I don't know where that leaves us?

pp-mo: Sorry I hadn't thought of that.
I will fix it ...

pp-mo: I have shifted the still-used code out of experimental.
Not sure I've got it right yet, though.

pp-mo:
Latest change : triggered by the observation that we must avoid the deprecation message in normal use of AreaWeightedRegridder.
I might have broken something in the test imports here, so will just see what the CI tests say...

pp-mo:
Latest probably fixed tests.
But .. we seem to have an import circularity problem that is specific to the docs-build ??
Not clear how that can be ...

lbdreyer:
Hey @pp-mo is this close to being ready? Once this is in we can merge #4540, then we should be ready to cut the v3.2.x branch

pp-mo:
> Hey @pp-mo is this close to being ready? Once this is in we can merge #4540, then we should be ready to cut the v3.2.x branch

I think the main job is done, but there is an outstanding CI problem, and I'm not clear what is wrong.
It seems to be an import glitch (like a circularity import problem?) that only happens when running the doctests.
So, that will need a bit of investigation.

pp-mo:
> It seems to be an import glitch (like a circularity import problem?) that only happens when running the doctests.

Update : rather luckily it is also failing in the benchmarks, which is a lot easier to unpick.
So basically "from iris import analysis" or "import iris.analysis" fails, but it's ok if you import "iris.cube" _first_.

Search goes on ...

pp-mo:
OK, that fixed it !
I can't quite understand why neither PyCharm or flake8 spotted that the import of iris.cube was not necessary, yet don't complain when it is removed.

Whatever, I have now moved all the non-obsolete support code used by  AreaWeighted into iris.analysis._area_weighted, and it all seems to function, and without emitting a DeprecationWarning.

lbdreyer (approved):
All looks good to me! Thanks @pp-mo !

Issue #4538, opened 2022-01-26
Degraded PP load times in later Iris versions


wjbenfold:
## 📰 Custom Issue

Iris has got slower at loading pp files into cubes (though not noticeably in realising them). There was a massive slowdown in realisation in the default-current environment, but that seems to have resolved in default-next.

![image](https://user-images.githubusercontent.com/69585101/151176168-8ff4e9b6-140e-4221-9c79-25ef02c755db.png)

x-axis is number of time steps in the pp file
y-axis is time in seconds

wjbenfold:
`pstats` from `profile` for default-next for comparison
```
   ncalls  tottime  percall  cumtime  percall filename:lineno(function)
        1    0.000    0.000  143.879  143.879 profile:0(main())
   1066/1    0.039    0.000  143.878  143.878 :0(exec)
    288/1    0.005    0.000  143.878  143.878 <string>:1(<module>)
        1    0.000    0.000  143.874  143.874 profiler.py:9(main)
        1    0.044    0.044  131.555  131.555 .../lib/python3.7/site-packages/iris/__init__.py:278(load)
        1    0.000    0.000   84.038   84.038 .../lib/python3.7/site-packages/iris/cube.py:142(merged)
        1    0.000    0.000   84.038   84.038 .../lib/python3.7/site-packages/iris/cube.py:155(<listcomp>)
        1    0.003    0.003   84.038   84.038 .../lib/python3.7/site-packages/iris/cube.py:84(merged)
        1    0.063    0.063   84.036   84.036 .../lib/python3.7/site-packages/iris/cube.py:408(merge)
     4263    0.101    0.000   81.865    0.019 .../lib/python3.7/site-packages/iris/_merge.py:1330(register)
     4264    0.948    0.000   71.788    0.017 .../lib/python3.7/site-packages/iris/_merge.py:1758(_extract_coord_payload)
    34112    0.325    0.000   64.963    0.002 .../lib/python3.7/site-packages/iris/cube.py:1411(coord_dims)
    34112    0.426    0.000   64.546    0.002 .../lib/python3.7/site-packages/iris/cube.py:1785(coord)
    34112    0.817    0.000   64.002    0.002 .../lib/python3.7/site-packages/iris/cube.py:1585(coords)
    34112    0.800    0.000   57.509    0.002 .../lib/python3.7/site-packages/iris/cube.py:1742(<listcomp>)
[...]

Issue #4530, opened 2022-01-26
Why are ScalarCoords listed in cube.aux_coords?


kaedonkers:
## 🐛 Bug Report
The tuple that is `cube.aux_coords` contains scalar coords. Why? Surely ScalarCoords are not the same as AuxCoords? This presents a problem when trying to remove AuxCoords from a cube, for instance to deal with errors when using `cube.coords(dimensions=[0])` (see #3324). 

## How To Reproduce
Inspect `cube.aux_coords` for a cube containing both AuxCoords and ScalarCoords. You will find that both types of coords are present in the tuple.

## Expected behaviour
`cube.aux_coords` returns a tuple containing **only** AuxCoords.

## Environment 
 - OS & Version: RHEL7
 - Iris Version: 3.1.0

pp-mo:
In Iris terms, scalar coords **_are_** aux-coords :  
That is, they are the among the coords of a cube which are not dimension coords, or derived coords.  
If you build a cube directly with Cube(dim_coords_and_dims=xx, aux_coords_and_dims=yy), they are included with the other aux-coords.

They are listed separately from Auxiliary Coordinate in a cube printout, but that is really just for visual convenience :  
Internally, that is in term of how they are stored, they _are_ aux-coords, but just ones which have no cube dims.
https://github.com/SciTools/iris/blob/main/lib/iris/_representation/cube_summary.py#L251-L255

Possibly a confusion is due to dimensionality/shape :  
For technical reasons, although they don't map any cube dimension, `coord.cube_dims() == ()`, they still always have a shape `coord.shape == (1,)` and `coord.ndim == 1`, rather than shape=() and ndim=0 as you might expect.  
I seem to recall there was a good techincal reason for this ... but not what it actually was !!  Probably something to do with numpy ??

pp-mo:
Also in CF terms...
They are auxiliary coordinates because they are attached to a variable via its 'coordinates' attribute
https://cfconventions.org/Data/cf-conventions/cf-conventions-1.9/cf-conventions.html#scalar-coordinate-variables
(this is what distinguishes dimension coords from auxiliary coordinates)

pp-mo:
Sorry for all this noise, but.... 
@kaedonkers I forgot to ask what was the actual problem here (!)
  -- you need to deal with scalar coords as a separate sub-class in some way?
Obviously you can get that info, but what do you think is **_missing_** ?

kaedonkers:
[...]

PR #4549, opened 2022-01-27
minor optimisations

bjlittle:
### Description
This PR introduces an optimisation to `iris.cube.Cube.coord_dims`, and also a lesser optimisation to merge.  

This change contributes, in part, to addressing the use case raised by issue #4538. As a result, I've seen a rough 25% speed-up of the worst case scenario given in #4538 (on my desktop) i.e., loading a 24,000 field PP file (time series) has gone from ~80secs to ~60secs. 

Essentially, a cheap win saving can be achieved by not blindly calling `iris.cube.Cube.coord` within `iris.cube.Cube.coord_dims`, particularly when the actual cube coordinate that exists within the cube is provided to the `coord_dims` method.

The philosophy of `coord_dims` is to avoid coordinate equivalence, as it is a relatively expensive operation, but rather use object identity instead, if possible.

---

### DIFF
--- a/lib/iris/_merge.py
+++ b/lib/iris/_merge.py
@@ -1809,7 +1809,8 @@ class ProtoCube:
 
         # Order the coordinates by hints, axis, and definition.
         for coord in sorted(coords, key=key_func):
-            if not cube.coord_dims(coord) and coord.shape == (1,):
+            dims = tuple(cube.coord_dims(coord))
+            if not dims and coord.shape == (1,):
                 # Extract the scalar coordinate data and metadata.
                 scalar_defns.append(coord.metadata)
                 # Because we know there's a single Cell in the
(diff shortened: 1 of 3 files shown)


bjlittle:
The benchmarks have registered a significant increase in performance, which is reassuring... Although they don't specifically target the use case. That said, this change seems like a general good win

bjlittle:
I'll add an appropriate `whatsnew` entry...

wjbenfold (approved):
Changes to `_merge` and `cube` look good to me

wjbenfold:
FWIW, I can't see an explicit test of the `cube.coord_dims` function (though there are a lot of tests that use it)

bjlittle:
I don't believe we need an explicit test for this, it's more than exercised with existing tests.

But that's just my opinion 😉

bjlittle:
BTW I'm pretty much AWTK today, so don't let a lack of `whatsnew` entry block this PR @wjbenfold 

If you've got capacity, and think it's worthwhile, perhaps add one as a follow on?

wjbenfold:
Would like to see unit testing of `Cube.coord_dims` but don't feel it needs to block this given the incidental integration test coverage by other tests. Checked this view with @lbdreyer 

Raised #4552

lib/iris/_merge.py
```diff
@@ -1809,7 +1809,8 @@ def key_func(coord):
 
         # Order the coordinates by hints, axis, and definition.
         for coord in sorted(coords, key=key_func):
-            if not cube.coord_dims(coord) and coord.shape == (1,):
+            dims = tuple(cube.coord_dims(coord))
+            if not dims and coord.shape == (1,):
```
pp-mo, line 1813: FWIW I was looking at this only yesterday, in [quite another context](https://github.com/SciTools/iris/issues/4530#issuecomment-1022220908)
And, I think that the shape **_didn't_** really need testing : 
if it has no cube-dims, it **must** be scalar, no?

PR #4557, opened 2022-01-31
v3.2.0rc0 whats new and version string

lbdreyer:
### Description
Updates version string to 3.2.0rc0
Moved whats new dev.rst -> 3.2.rst

---


pp-mo:
A couple of quick queries here ...
  1. is it really necessary + desirable to delete `docs/whatsnew/dev.template` ?
      I know the dev docs say to do that, but then we have to re-introduce it post-release.  I'm not convinced it does any harm to just leave it in ?
  2. I get that #4540 basically renamed 'latest.rst' to  'dev.rst'.  We have also changed the ref in 'whatsnew/index.rst'.
      And here when cutting a new release version, we rename 'dev.rst' to '3.2.rst' in the index.
      but then why do we still need 'latest.rst' _at all_  -- which is now just a link to 'dev.rst' ?
      (I know this is really a question for #4540, but I missed the boat there)

stephenworsley (approved):
Looks good.

pp-mo (changes requested):
A couple of quick queries here ...

  1.  is it really necessary + desirable to delete docs/whatsnew/dev.template ?
       I know the dev docs say to do that, but then we have to re-introduce it post-release. I'm not convinced it does any harm to just leave it in ?
  2.  I get that #4540 basically renamed 'latest.rst' to 'dev.rst'. We have also changed the ref in 'whatsnew/index.rst'.
     And here when cutting a new release version, we rename 'dev.rst' to '3.2.rst' in the index.
     but then why do we still need 'latest.rst' at all -- which is now just a link to 'dev.rst' ?
     (I know this is really a question for docs: move whatsnew #4540, but I missed the boat there)

Otherwise, :+1: LGTM !

wjbenfold:
> 2\. but then why do we still need 'latest.rst' at all -- which is now just a link to 'dev.rst' ?

Presumably because we want to just update the symlink rather than the toctree, but I think @bjlittle added it so might have another view?

lbdreyer:
>is it really necessary + desirable to delete docs/whatsnew/dev.template ?
Not sure really! I am doing it because that's what the docs say and that is what you did for the 3.1.0rc0

PR #4558, opened 2022-01-31
Revert "v3.2.0rc0 whats new and version string"

lbdreyer:
Reverts SciTools/iris#4557

This shouldn't have gone onto main!!
This was supposed to target the v3.2.x branch


pp-mo:
Oh gosh, I should have spotted that  !

Would it not be better to reset the main, than to add a reversion commit ?

lbdreyer:
>Would it not be better to reset the main, than to add a reversion commit ? 

Probably! I added the reversion commit as that it what github creates for you when you click "revert" on a merged PR.

I will try resetting main

pp-mo (changes requested):
I'd prefer to try to reset the branch first, if we can.
It seems to need an admin though, so let's wait + see if @bjlittle can help.

Hi @bjlittle 😉 !

bjlittle:
@lbdreyer Just to let you know that @pp-mo and I successfully reset the `main` branch, so I'm going to close this PR now

Thanks :+1:

lbdreyer:
Something still seems broken on main. The latest [commit](https://github.com/SciTools/iris/commit/d503ce62c41e9d827b91121e4f6395d1fcb6b644) included the pre-commit change and my changes

PR #4559, opened 2022-01-31
3.2 version and whats new

lbdreyer:
### Description
Replace #4557 

---


pp-mo (approved):
OK, same commits as #4557 so nothing to see here.
Will merge when it (re-)checks out.

PR #4561, opened 2022-02-02
gha: update refresh-lockfiles

bjlittle:
### Description
This PR includes some minor updates to the `refresh-lockfiles` GHA.

---

### DIFF
--- a/.github/workflows/refresh-lockfiles.yml
+++ b/.github/workflows/refresh-lockfiles.yml
@@ -108,6 +108,7 @@ jobs:
           rm -r artifacts
         
       - name: Create Pull Request
+        id: cpr
         uses: peter-evans/create-pull-request@f22a7da129c901513876a2380e2dae9f8e145330
         with:
           commit-message: Updated environment lockfiles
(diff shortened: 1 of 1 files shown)


.github/workflows/refresh-lockfiles.yml
```diff
@@ -108,13 +108,24 @@ jobs:
           rm -r artifacts
         
       - name: Create Pull Request
-        uses: peter-evans/create-pull-request@052fc72b4198ba9fbc81b818c6e1859f747d49a8
+        id: cpr
+        uses: peter-evans/create-pull-request@v3
```
bjlittle, line 112: We can pin to the latest stable release... unless there's a specific reason we need to use the previously pinned commit

trexfeathers: This is considered unsafe practice with third party actions (i.e. those not made by GitHub), as a bad actor could re-label a new malicious commit as `v3`.

bjlittle: Well, at the very least we can bump it to a later specific commit

bjlittle: also we should start thinking about using dependabot

jamesp: Dependabot is ace but doesn't support conda files

jamesp: https://github.com/dependabot/dependabot-core/issues/2227

bjlittle: @jamesp I was thinking more of using dependabot to automate (or tell us) when the `create-pull-request` GHA itself can be updated to a later release (rather than using it to resolve/check for later releases of conda packages in the lock files)

It should be able to do that, right?

jamesp: ah yes, good idea. it looks like dependabot scans any gha in our yaml files. https://docs.github.com/en/code-security/supply-chain-security/understanding-your-software-supply-chain/about-the-dependency-graph#dependencies-included
So long as the GHA code uses one of the supported ecosystems, I guess

bjlittle: As a starter, see #4568

PR #4562, opened 2022-02-02
Final offline benchmark migration

trexfeathers:
### Description

Applies all remaining improvements from the offline benchmark suite.

Wherever possible no changes have been made other than copying the changes across. Expecting plenty of improvements in future PR's, rather than here, to keep the complexity down.

The main extra change has been converting **all** benchmarks to use the correct method for synthetic data generation introduced by #4477.

---


trexfeathers:
With the extra coverage adding so much to the run-time, have been forced to implement #4583 earlier than planned. So this PR is now waiting on #4477 _and_ #4583!

pp-mo (approved):
Seems like a complex lift+shift from the old 'metrics' code.
I checked everthing against the originals, and ultimately I couldn't find anything to call out.

PR #4572, opened 2022-02-08
Dask chunking control for netcdf loading.

pp-mo:
Aims to address #3333 

Still a bit preliminary, testing and documentation probably need more doing.
But here it is for discussion.  I basically set out to do this because I couldn't stop thinking about it.
@cpelley @TomekTrzeciak

I will also add some comments in the nature of a discussion...
Some of that as pointers for things we might want changed, and some as a replacement for missing documentation (?!)

### DIFF
--- a/lib/iris/_lazy_data.py
+++ b/lib/iris/_lazy_data.py
@@ -47,7 +47,9 @@ def is_lazy_data(data):
     return result
 
 
-def _optimum_chunksize(chunks, shape, limit=None, dtype=np.dtype("f4")):
+def _optimum_chunksize(
+    chunks, shape, limit=None, dtype=np.dtype("f4"), dims_fixed=None
+):
     """
     Reduce or increase an initial chunk shape to get close to a chosen ideal
     size, while prioritising the splitting of the earlier (outer) dimensions
@@ -55,7 +57,7 @@ def _optimum_chunksize(chunks, shape, limit=None, dtype=np.dtype("f4")):
 
     Args:
 
-    * chunks (tuple of int, or None):
+    * chunks (tuple of int):
         Pre-existing chunk shape of the target data : None if unknown.
     * shape (tuple of int):
         The full array shape of the target data.
@@ -63,6 +65,11 @@ def _optimum_chunksize(chunks, shape, limit=None, dtype=np.dtype("f4")):
         The 'ideal' target chunk size, in bytes.  Default from dask.config.
     * dtype (np.dtype):
         Numpy dtype of target data.
+    * dims_fixed (list of bool):
+        If set, a list of values equal in length to 'chunks' or 'shape'.
+        'True' values indicate a dimension that can not be changed, i.e. that
+        element of the result must equal the corresponding value in 'chunks' or
+        data.shape.
 
     Returns:
     * chunk (tuple of int):
@@ -83,6 +90,9 @@ def _optimum_chunksize(chunks, shape, limit=None, dtype=np.dtype("f4")):
         "chunks = [c[0] for c in normalise_chunks('auto', ...)]".
 
     """
+    if chunks is None:
+        chunks = list(shape)
+
     # Set the chunksize limit.
     if limit is None:
         # Fetch the default 'optimal' chunksize from the dask config.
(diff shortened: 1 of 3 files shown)


pp-mo:
## API discussion

The API (reference docs-build [here]()) is actually rather different from what I had expected to end up with.
Here are some reasons..

I reasoned that if loading several pieces of data, it will often make sense to align the chunking according to the dimensions.
That could be if there are several datacubes wiith same dimensions.  But also applies to the cube components, e.g. multiple AuxCoords.
The logic of using dimensions is that different AuxCoords (for instance) may not all have the same dimensions, but if arithmetic might be performed between different ones, aligning chunks will make good sense.

Assuming that totally dimension-oriented approach is _not_ always correct though, you need a way to give  different settings for different variables.  Hence the var-name selective behaviour.

It then made sense to me to apply common dimension settings to individual result cubes, which results in the logic shown.

I had expected to add loader kwargs, and involve the https://github.com/SciTools/iris/pull/3720
However, I took ideas from the [unstructured data loading control](https://scitools-iris.readthedocs.io/en/latest/further_topics/ugrid/operations.html#load) to provide a context-manager approach instead.
One big advantage of this is that you can just invoke existing code, with maybe multiple load calls, without modifying the existing code at all.

pp-mo:
## Usage examples.
This should really go in docs "somewhere"
For now, I put usage examples in some stopgap [integration tests](https://github.com/SciTools/iris/blob/93b0917bd8152df163e5accaada75c6344e30c78/lib/iris/tests/integration/test_netcdf__chunk_control.py#L48-L90).

An example that is working for me is, to solve the problem referrred to [here](https://github.com/SciTools/iris/issues/3333#issuecomment-1031391208)

In that specific example ...
  * data was loaded from multiple 1-timestep files
    * with dims `(realization: 18; pressure: 33; latitude: 960; longitude: 1280)`
  * by loading multiple files (and auto-merge in load), we get a single cube
    * with dims `(time:10, realization: 18; pressure: 33; latitude: 960; longitude: 1280)`
    * which is chunked as `(1, 1, 17, 960, 1280)`
      * (the '17' is determined by the dask default chunksize).
  * the user then wants only one pressure-level, so cube[:, 0]
    *  --> `(time:10, realization: 18; latitude: 960; longitude: 1280)`
    * chunked as `(1, 1, 960, 1280)`
  * but attempting to fetch that ~200Mb of data (in ~1Mb chunks) consumes several Gb of data, and can crash the machine
  * .. alternatively, saving it to a file (dask streaming) works but is extremely slow (20x similar data)

When applying `CHUNK_CONTROL.set(pressure=1)`, the chunking of the original merged cube changes
from `(1, 1, 17, 960, 1280)`
to `(1, 17, 1, 960, 1280)`
So that a chunk spans several realization points, but only one pressure level.
This then works much better.

lib/iris/_lazy_data.py
```diff
@@ -92,61 +102,93 @@ def _optimum_chunksize(chunks, shape, limit=None, dtype=np.dtype("f4")):
...
+    if dims_fixed is None:
+        # Get initial result chunks, starting with a copy of the input.
+        working = list(chunks)
+    else:
+        # Adjust the operation to ignore the 'fixed' dims.
+        # (We reconstruct the original later, before return).
+        chunks = np.array(chunks)
+        dims_fixed_arr = np.array(dims_fixed)
+        # Reduce the target size by the fixed size of all the 'fixed' dims.
+        point_size_limit = point_size_limit // np.prod(chunks[dims_fixed_arr])
```
TomekTrzeciak, line 118: It would be nice to somehow allow Dask's [special chunk values](https://docs.dask.org/en/stable/array-chunks.html#automatic-chunking) (`-1`, `'auto'` and `None`), but I guess this is at odds with doing this logic on Iris side. Have you tried to inquire if Dask would be interested to add/extend chunking logic to support such use cases? I think it would make the most sense to boot this out of Iris as per the note [here](https://github.com/SciTools/iris/blob/1d5eb3eaf9d183af4e1849e82798bedf876ab53e/lib/iris/_lazy_data.py#L77-L78).

lib/iris/fileformats/netcdf.py
```diff
@@ -3206,3 +3245,114 @@ def is_valid_packspec(p):
...
+
+# Note: the CHUNK_CONTROL object controls chunk sizing in the
+# :meth:`_get_cf_var_data` method.
+# N.B. :meth:`_load_cube` also modifies this when loading each cube,
+# introducing an additional context in which any cube-specific settings are
+# 'promoted' into being global ones.
+
+#: A :class:`ChunkControl` object providing user-control of Dask chunking
+#: when Iris loads netcdf files.
+CHUNK_CONTROL: ChunkControl = ChunkControl()
```
TomekTrzeciak, line 3358: Nitpick: I would go with lower case here, `CHUNK_CONTROL.set(...)` looks a bit shouty.

TomekTrzeciak (commented):
> I basically set out to do this because I couldn't stop thinking about it.
 
That's always a good motivator 😉

First off, really glad to see progress on this. I think context manager makes a lot of sense for this given the many layers that these options would have to pass through. If I could have a wish it would be for some way to disable Iris' logic entirely and fallback on Dask and its machinery to manage the chunking.

lib/iris/fileformats/netcdf.py
TomekTrzeciak: Isn't it mostly for simple things with "fixed" value, rather than objects with complex behaviour like context managers?

trexfeathers: Oh. In that case I've messed up elsewhere! Look forward to seeing what @pp-mo has to say 😅

jamesp: I'm going to do a drive-by nitpick as I'm getting all the email alerts for this thread in my inbox :)

There should be no need for the type hint here, Python type hinter should be smart enough to know that 

`CHUNK_CONTROL = ChunkControl()`

is indeed a `ChunkControl` instance.

pp-mo: >  @TomekTrzeciak Nitpick

As regards captilisation, I was basically mimicking [iris.FUTURE](https://scitools-iris.readthedocs.io/en/latest/generated/api/iris.html#iris.FUTURE).
Which makes _some_ kind of sense IMHO as this is very much the same style of thing.

On the other hand it's possible that, when that was introduced, we were slightly conflicted about the difference between a "constant value" and a "global variable", as you suggest, since in Python there is essentially no actual distinction.

----
> @jamesp drive-by nitpick

It wasn't to enable type hinting, but trying to inject type into the docstring representation.  
https://pp-iris.readthedocs.io/en/nc_chunk_control/generated/api/iris/fileformats/netcdf.html#iris.fileformats.netcdf.CHUNK_CONTROL
Hence the colon.  But it doesn't seem to actually do so, presumably because Sphinx is not so configured 😞

lib/iris/_lazy_data.py
pp-mo: ### (background, not really an answer!)

The problem with the existing Dask built-in strategy is that is treats all dimensions as equal and tries to even out the chunking across all dims.  Which is pretty clearly unsuitable in most of our cases, given the known contiguity factors in file storage, and especially when the file contains 2d-fields which cannot be efficiently sub-indexed.
Which is why we provided the Iris version, to prioritise splitting of the "outer" dims and keeping the "inner" ones together.

**But it's certainly true that we could feedback and engage, as-per the note you mention.**
Unfortunately though, the dask code is really complicated and engagement seems a bit of an uphill struggle IMHO.
By which I mean ... I have repeatedly attempted to engage with the dask codebase, and maybe contribute, but I've found it very hard to understand, with a lot of unexplained terminology and framework concepts.  
So it seems to me that there is a problem with contributing unless you are an "expert", because there is a lot that you need to know beyond what the "ordinary" user sees.

In fact, as a very relevant example, I considered addressing this same chunking control problem by picking apart the graph of a loaded cube to identify and adjust (or replace) the `iris.fileformats.netcdf.NetCDFDataProxy` objects which are in there.  It's perfectly do-able, but you need to understand the structure of the dask graph and employ various assumptions about it :  However, there is _no public account_ of that, or any apparent intention for "ordinary users" to manipulate them in this way.  So it would seem that a dask graph is really an "opaque" object, and the internal structure might get refactored and change completely, in which case such a solution, being outside intended usage, would simply need re-writing.  So effectively, AFAICT only a dask "expert" is expected to work at that level and, crucially, such a solution only really makes sense _if the code lives in the dask codebase_.

pp-mo: > @TomekTrzeciak It would be nice to somehow allow Dask's [special chunk values](https://docs.dask.org/en/stable/array-chunks.html#automatic-chunking) (-1, 'auto' and None),

Reminder of how that is described :
```
Chunks also includes three special values:
   * -1: no chunking along this dimension
   * None: no change to the chunking along this dimension (useful for rechunk)
   * "auto": allow the chunking in this dimension to accommodate ideal chunk sizes
```


From a technical POV, I guess we could respect -1, but the others are a little tricky :  
"Auto" is sort-of what we are already doing, with the dims we don't have an explicit control for.
But it isn't quite what Dask means my it since we will _still_ want to apply a dimension ordering.
**"None"** I think could only really apply if we treat chunking specified on a file variable as the initial setting.  At present, we do use that as a starting point anyway, so I'm not sure how this would differ from "auto" in this usage.

TomekTrzeciak: > From a technical POV, I guess we could respect -1, but the others are a little tricky :
> **"Auto"** is sort-of what we are already doing, with the dims that _don't_ have an explicit control given. But it isn't quite what Dask means, since we will still want to apply our dimension ordering priority.
> **"None"** I think could only really apply if we treat chunking specified on a file variable as the initial setting. At present, we do use that as a starting point anyway, so I'm not sure how this would differ from "auto" in this usage.
> 
> **Can you suggest what the interpretations would be -- how would this work ?**

For **None** the only sensible interpretation I can think of is to keep the chunking from the netcdf file itself as you suggest, but dunno what Dask is doing in that case.

I reckon **Auto** would require repeating the logic that Dask is doing but restricted only to the specified dimensions. Seem quite tricky, probably best not to support it if the implementation is going to live on the Iris' side.

TomekTrzeciak: > > **"None"** I think could only really apply if we treat chunking specified on a file variable as the initial setting. At present, we do use that as a starting point anyway, so I'm not sure how this would differ from "auto" in this usage.
> 
> For **None** the only sensible interpretation I can think of is to keep the chunking from the netcdf file itself as you suggest, but dunno what Dask is doing in that case.

Sorry, I've re-read your comment and I think this is not what you suggest. My interpretation of **None** is that it behaves the same as if you would explicitly specify chunks equal to those already existing on the variable, i.e., this is their final rather than initial value.

lib/iris/fileformats/netcdf.py
```diff
@@ -3206,3 +3245,114 @@ def is_valid_packspec(p):
...
+                    )
+                    raise ValueError(msg)
+                dim_chunks = self.var_dim_chunksizes.setdefault(var_name, {})
+                for dim_name, chunksize in dimension_chunksizes.items():
+                    if not (
+                        isinstance(dim_name, str)
+                        and isinstance(chunksize, int)
+                    ):
+                        msg = (
+                            "'dimension_chunksizes' kwargs should be an iterable "
```
pp-mo, line 3340: TODO: fix this : it's not an iterable in the API, it's a dictionary..

lib/iris/fileformats/netcdf.py
```diff
@@ -3206,3 +3245,114 @@ def is_valid_packspec(p):
...
+        """
+        Control the Dask chunksizes applied to netcdf variables during loading.
+
+        Parameters
+        ----------
+        var_names : str or list of str
+            apply the ``dimension_chunksizes`` controls only to these variables,
+            or when building cubes from these data variables.
+            If None (the default), settings apply to all loaded variables.
+        dimension_chunksizes : dict: str --> int
```
pp-mo, line 3289: TODO: this might be extended to more complex chunking controls, if useful?
Probably could do nonuniform sizes within a dimension (e.g. (3, 5, 5, 2)), but _not_ across multiple dims (not even sure how you say that, but I think is possible).
Plus None/auto/-1 as [already suggested](https://github.com/SciTools/iris/pull/4572#discussion_r810056992)

lib/iris/tests/integration/test_netcdf__chunk_control.py
```diff
@@ -0,0 +1,94 @@
...
+        self.assertEqual((3, 4, 5, 6), cube.lazy_data().chunksize)
+        sigma = cube.coord("sigma")
+        self.assertEqual((4,), sigma.shape)
+        self.assertEqual((2,), sigma.lazy_points().chunksize)
+        # N.B. this does not apply to bounds array
+        self.assertEqual((4, 2), sigma.lazy_bounds().chunksize)
+
+    def test_control_cube_var(self):
+        with CHUNK_CONTROL.set(self.cube_varname, model_level_number=2):
+            cube = iris.load_cube(self.tempfile_path, self.cube_varname)
```
cpelley, line 71: :+1:

trexfeathers:
We're going to make this happen! But since this PR has gone rather stale, and @pp-mo has a lot of assigned work, we're going to close this PR and aim to apply the code to the latest Iris. See #5398 for the task breakdown.

PR #4580, opened 2022-02-10
Fix load_http bug, extend testing, and note to docs

lbdreyer:
### Description
This addresses two issues relating to loading files via OPeNDAP

**Fix iris.io.load_http and add extra testing**
It looks like `iris.io.load_http` broke in Iris 3.1 when we moved imports around. I have added the import (`from iris.fileformats import FORMAT_AGENT`)
Unfortunately a lack of testing meant we didn't catch it! Currently, the only OPeNDAP related testing we have is:
* we test load/load_cube/load_cubes/load_raw call iris.io.load_http
* we check that the format_picker selects the correct FormatSpecification ('NetCDF OPeNDAP'). 

You can see these two tests in [this commit](https://github.com/SciTools/iris/pull/422/commits/f1a9f0603c4b283921e99f1bec381c40bc570af1).
Testing with http files is quite difficult as we don't want to be tied to trying to access an OPeNDAP server which may go down! I have proposed a testing strategy which is a bit hacky but at least will test all operation Iris performs before it call the netCDF library (which is all we really need to test)

**Add documentation**
A user raised the issue that someone could quite reasonably read the documentation, see Iris.load takes a URI and supply it with a non-OPeNDAP URL. I have added clarification to say we only support OPeNDAP URLS.
The trouble with this is that the restriction of OPeNDAP is due to NetCDF (since we just pass the URL onto NetCDF) so if NetCDF were to start support something else, our documentation may be out of date. Should we be concerned about this?

Would be nice to get this into Iris 3.2 before the release if possible?
---

### DIFF
--- a/lib/iris/__init__.py
+++ b/lib/iris/__init__.py
@@ -44,6 +44,8 @@ All the load functions share very similar arguments:
         standard library function :func:`os.path.expanduser` and
         module :mod:`fnmatch` for more details.
 
+        If supplying a URL, only OPeNDAP Data Sources are supported.
+
     * constraints:
         Either a single constraint, or an iterable of constraints.
         Each constraint can be either a string, an instance of
@@ -287,6 +289,7 @@ def load(uris, constraints=None, callback=None):
 
     * uris:
         One or more filenames/URIs, as a string or :class:`pathlib.PurePath`.
+        If supplying a URL, only OPeNDAP Data Sources are supported.
 
     Kwargs:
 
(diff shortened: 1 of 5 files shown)


lib/iris/__init__.py
```diff
@@ -44,6 +44,8 @@
         standard library function :func:`os.path.expanduser` and
         module :mod:`fnmatch` for more details.
 
+        If supplying a URL, only OPeNDAP Data Sources are supported.
```
bjlittle, line 47: @lbdreyer Great to highlight this to the user :100: 

Perhaps we could also add a `.. warning::` directive in the above main body of the doc-string to ensure this information can't be missed in the rendered `sphinx` docs?

bjlittle (changes requested):
@lbdreyer Awesome, nice work. Thanks for doing this :beers: 

Just a minor suggestion to service, otherwise let's get this goodness banked :+1:

bjlittle (approved):

lbdreyer:
Thanks @bjlittle  !

Issue #4574, opened 2022-02-09
`DeprecationWarning`s


rcomer:
## 📰 Custom Issue

A couple of `DeprecationWarning`s I've spotted that should probably be addressed at some point.  They can be reproduced by running test modules, but don't seem to show up in the CI.

### Steps to reproduce
* Check out a current branch
* Activate an up-to-date `iris-dev` environment

- [ ] Regridding Issue #4598
`python lib/iris/tests/integration/analysis/test_area_weighted.py`   produces
```
[git-path]/iris/lib/iris/analysis/_area_weighted.py:606: DeprecationWarning: Out of bound index found. This was previously ignored when the indexing result contained no elements. In the future the index error will be raised. This error occurs either due to an empty slice, or if an array has zero elements even before indexing.
  (Use `warnings.simplefilter('error')` to turn this DeprecationWarning into an error and get more details on the invalid index.)
    src_area_datas_square = src_data[
```
- [x] Plotting Issue #4584
`python lib/iris/tests/integration/plot/test_netcdftime.py` produces
```
[conda-path]/iris-dev/lib/python3.8/site-packages/nc_time_axis/__init__.py:41: DeprecationWarning: CalendarDateTime is obsolete and will be deprecated in nc_time_axis version 1.5.  Please consider switching to plotting instances or subclasses of cftime.datetime directly.
```

rcomer:
@pp-mo I'm not seeing that from the Iris tests, but testing my user code with the v3.2 release candidate I got some like this:

```
<frozen importlib._bootstrap>:228: RuntimeWarning: numpy.ndarray size changed, may indicate binary incompatibility. Expected 16 from C header, got 88 from PyObject
```

```
<frozen importlib._bootstrap>:241: RuntimeWarning: numpy.ndarray size changed, may indicate binary incompatibility. Expected 16 from C header, got 96 from PyObject
```

I can't say I even attempted to understand them!

pp-mo:
> I can't say I even attempted to understand them!

😄 
I haven't seen it in our tests, either.
I'd like to chase it, but initially it seems very hard to work out what is responsible .
My usage was based on PyTest,. There you can use a magic "-W" setting to cause the warning to error+stasckdump.  From which, it seems to occur when netCDF4 is first imported.  Beyond that I ran out of time / patience / enthusiasm ...

wjbenfold:
> [ ] Regridding Issue

[...]

PR #4584, opened 2022-02-11
Stop using nc_time_axis.CalendarDateTime

rcomer:
### Description

`nc_time_axis.CalendarDateTime` was deprecated at v1.4 of `nc-time-axis` and we are encouraged to pass the `cftime` objects directly, which is nicer.  Fixes the easy part of #4574.

References
https://github.com/SciTools/nc-time-axis/pull/80 new plotting functionality introduced.
https://github.com/SciTools/nc-time-axis/pull/86 `CalendarDateTime`s deprecated.

I will add a whatsnew when I know that merge is imminent.  Or in a separate PR if the reviewer prefers.

---

### DIFF
--- a/requirements/ci/py38.yml
+++ b/requirements/ci/py38.yml
@@ -25,7 +25,7 @@ dependencies:
   - graphviz
   - iris-sample-data >=2.4.0
   - mo_pack
-  - nc-time-axis >=1.3
+  - nc-time-axis >=1.4
   - pandas
   - pip
   - python-stratify
--- a/docs/src/whatsnew/dev.rst
+++ b/docs/src/whatsnew/dev.rst
@@ -61,7 +61,9 @@ This document explains the changes made to Iris for this release
 🔗 Dependencies
 ===============
 
-#. N/A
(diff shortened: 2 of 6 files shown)


requirements/ci/py38.yml
```diff
@@ -25,7 +25,7 @@ dependencies:
   - graphviz
   - iris-sample-data >=2.4.0
   - mo_pack
-  - nc-time-axis >=1.3
+  - nc-time-axis >=1.4
```
rcomer, line 28: Not sure whether it's better to do this or to add some handling so we still use `CalendarDateTime` with nc_time_axis v1.3?

lbdreyer: I agree with the approach you have done here.
I don't see the benefit of adding support for older version nc_time_axis. There isn't a reason user's shouldn't be able to use the latest version of nc_time_axis (it is after all a small + simple package so it should bring dependency conflicts).

lbdreyer (approved):
This all looks good to me! Thanks for fixing this @rcomer !
If you add a whats new I will merge it in

rcomer:
Thanks @lbdreyer!  I have now added the whatsnew, and it renders like this:
calendardatetime_whatsnew

PR #4589, opened 2022-02-15
Weighted `aggregated_by`

### LINKED ISSUE #4581: Weighted `aggregated_by`
## ✨ Feature Request

Similar to `iris.cube.Cube.collapsed`, it would be nice if `iris.cube.Cube.aggregated_by` could also support `weights`.

## Motivation

`iris.cube.Cube.aggregated_by` currently [does not support weights](https://github.com/SciTools/iris/blob/e3fbd843d0c64a027d5ff460a5564fdd3a87439d/lib/iris/cube.py#L3895-L3901). Using weights in `aggregated_by` might be helpful in lots of use cases, e.g., when calculating annual means for monthly data where the months have a different number of days.

## Additional context

I tried to implement this for the non-lazy case (see [this commit](https://github.com/schlunma/iris/commit/d020bd2f9db5fa9d3d54619b7555173f1daafe51)). It actually doesn't require a lot of changes. If you think this feature is helpful, I can come up with a solution for the lazy case and open a pull request.

Thanks!!

schlunma:
### Description

This pull request implements weighted `aggregated_by`, e.g.,

```py
annual_mean_cube = monthly_mean_cube.aggregated_by('year', iris.analysis.MEAN, weights=n_days_per_month)
```

The implementation allows lazy and non-lazy aggregation, and fully supports the keyword `returned=True`. I also added a large number of tests to extensively test this feature; all seems to work fine.

Closes #4581.

---

### DIFF
--- a/docs/src/whatsnew/latest.rst
+++ b/docs/src/whatsnew/latest.rst
@@ -31,6 +31,11 @@ This document explains the changes made to Iris for this release
 ✨ Features
 ===========
 
+#. `@schlunma`_ added weighted aggregation over "group coordinates":
+   :meth:`~iris.cube.Cube.aggregated_by` now accepts the keyword `weights` if a
+   :class:`~iris.analysis.WeightedAggregator` is used. (:issue:`4581`,
+   :pull:`4589`)
+
 #. `@wjbenfold`_ added support for ``false_easting`` and ``false_northing`` to
    :class:`~iris.coord_system.Mercator`. (:issue:`3107`, :pull:`4524`)
 
--- a/lib/iris/tests/test_aggregate_by.py
+++ b/lib/iris/tests/test_aggregate_by.py
@@ -17,6 +17,7 @@ import iris
 import iris.analysis
 import iris.coord_systems
 import iris.coords
+from iris._lazy_data import as_concrete_data
 
 
 class TestAggregateBy(tests.IrisTest):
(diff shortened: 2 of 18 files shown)


lib/iris/tests/test_aggregate_by.py
```diff
@@ -621,18 +1064,73 @@ def test_single_missing(self):
...
+        )
+
+        # Realize data so that tests with lazy data pass (see
+        # test_lazy_aggregate_by.py).
+        aggregateby_cube.data = as_concrete_data(aggregateby_cube.data)
+        self.assertCML(
+            aggregateby_cube,
+            ("analysis", "aggregated_by", "single_missing.cml"),
+            checksum=False,
+        )
```
trexfeathers, line 1126:
```suggestion
        self.assertCML(
            aggregateby_cube,
            ("analysis", "aggregated_by", "single_missing.cml"),
        )
```

I _believe_ there are no longer concerns with masks messing with `checksum`, and that we should therefore remove all instances of `checksum=False` in this module. This also removes any need for `as_concrete_data()` since the data gets realised during `assertCML`.

trexfeathers (changes requested):
Sorry for the delay @schlunma, this was a very difficult review for me since I haven't previously done any work with `aggregated_by()`.

Well done on contributing such a valuable change. Got just a few changes for you.

schlunma:
Thanks for your review @trexfeathers! I implemented all your comments :+1: 

Not sure about the failing tests though, I can't really find out what's going wrong here. Is there anything I can do to fix these?

trexfeathers:
>Not sure about the failing tests though, I can't really find out what's going wrong here. Is there anything I can do to fix these?

Both failures are because a warning has been generated during documentation builds:

```
<frozen importlib._bootstrap>:219: RuntimeWarning: During parsing environment variable 'NPY_DISABLE_CPU_FEATURES':
```

Not sure why yet. `NPY_DISABLE_CPU_FEATURES` is something we've encountered before but I don't see why it's relevant here. I'm going to see if I can replicate it locally.

docs/src/whatsnew/latest.rst
```diff
@@ -31,6 +31,11 @@ This document explains the changes made to Iris for this release
 ✨ Features
 ===========
 
+#. `@schlunma`_ added weighted aggregation over "group coordinates":
```
trexfeathers, line 34: Please could you also add a `@schlunma` entry here:

https://github.com/SciTools/iris/blob/766e9e758ed8a48461deb9069e385f72f7a2c7c6/docs/src/common_links.inc#L65-L66

That will fix the documentation failures.

trexfeathers (changes requested):
This looks great, thanks @schlunma!

I've discovered the source of the documentation failure, so that's the only remaining change you need to make.

trexfeathers (approved):

trexfeathers:
Well done @schlunma, this has been a big piece of work.

schlunma:
Awesome, thanks for the review!!

PR #4592, opened 2022-02-16
Utility class in netcdf loader should not be public.

pp-mo:
Just spotted that our  API docs include the class `iris.fileformats.netcdf.OrderedAddableList`, which is unintended.

This was introduced in Iris v3.1.0, when we replaced the Pyke rules, but it seems has mistakenly been published as part of the public API + appears in the API docs, e.g. :
https://scitools-iris.readthedocs.io/en/v3.1.0/generated/api/iris/fileformats/netcdf.html#iris.fileformats.netcdf.OrderedAddableList

I don't think it's a big deal to just say this was a mistake + withdraw it, i.e. no deprecation or warnings required.

"Ideally" this might be combined with a desirable reorganisation of the netcdf package #4291, 
but that is basically for developers' benefit, whereas this is about the public API, so let's keep it separate.

### DIFF
--- a/lib/iris/fileformats/_nc_load_rules/actions.py
+++ b/lib/iris/fileformats/_nc_load_rules/actions.py
@@ -18,7 +18,7 @@ not :
 
 3) Iris-specific info is (still) stored in additional properties created on
    the engine object :
-       engine.cf_var, .cube, .cube_parts, .requires, .rule_triggered, .filename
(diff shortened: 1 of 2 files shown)


bjlittle (approved):

bjlittle:
@pp-mo Agreed. This was unintentional and there is no real value to users having this class part of the public API.

No harm, no foul.

Good spot, glad we're addressing this now 👍

Issue #4582, opened 2022-02-10
Coordinate systems with non-Earth radii do not work properly in iris 3.2.rc0


dennissergeev:
## 🐛 Bug Report

Unfortunately there are still bugs related to #4408 in iris 3.2.0rc0, even with @wjbenfold's fix (#4497, thanks for that by the way). I tested the `iris.analysis.cartography.rotate_winds()` function in the release candidate today and discovered that conversion between iris & cartopy projections / CRS continues to assume an Earth radius, failing for coordinate systems with semi-major axes different to `iris.fileformats.pp.EARTH_RADIUS`.

## How To Reproduce
Steps to reproduce the behaviour:

There are actually 2 seemingly related bugs, both of which trigger an error in a call of the `transform_points()` method of a CRS.

1. (mostly `iris`-side) Using `rotate_winds()` on cubes with a non-Earth radius `GeogCS` - see the error message below. It seems to be caused by [this line](https://github.com/SciTools/iris/blob/main/lib/iris/analysis/cartography.py#L930) in the cartography module. To reproduce:
```python
import cartopy.crs as ccrs
from iris.coord_systems import GeogCS
import numpy as np

x, y = np.meshgrid([1,2,3], [4,5,6])

non_earth_cs = GeogCS(123456.7)
crs_latlon = ccrs.Geodetic(globe=ccrs.Globe(ellipse="sphere"))  # Same as L930 in cartography.py
crs_latlon.transform_points(non_earth_cs.as_cartopy_projection(), x, y)
```
2. (mostly `cartopy`-side) Just calling `as_cartopy_projection()` on a non-Earth radius CS results in an error in Jupyter Lab when the `Projection._repr_html_(self)` method is called, which in turn calls the `transform_points()` method in the same way as above.
```python
from iris.coord_systems import GeogCS

non_earth_cs = GeogCS(123456.7)
non_earth_cs.as_cartopy_projection()
```

## Environment 
- OS & Version: e.g., Ubuntu 20.04 LTS
- Python Version: 3.10.2
- Iris Version: 3.2.0.rc0
- Cartopy Version: 0.20.2
- Pyproj Version: 3.3.0

## Additional context

<details>
<summary><b>Error message</b></summary>

```python
---------------------------------------------------------------------------
ProjError                                 Traceback (most recent call last)
Input In [169], in <module>
----> 1 tl_u, tl_v = rotate_winds(u, v, tl_cs)

[...]

PR #4596, opened 2022-02-22
Use crs.globe in _crs_distance_differentials() instead of creating a new Globe

dennissergeev:
### Description

Addresses the first point in #4582 by making `iris.analysis.cartography._crs_distance_differentials()` use the `.globe` attribute of the input `crs` parameter instead of creating a new `ccrs.Globe()` object.

All tests pass on my computer.

---


lib/iris/analysis/cartography.py
```diff
@@ -927,7 +927,7 @@ def _crs_distance_differentials(crs, x, y):
 
     """
     # Make a true-latlon coordinate system for distance calculations.
-    crs_latlon = ccrs.Geodetic(globe=ccrs.Globe(ellipse="sphere"))
+    crs_latlon = ccrs.Geodetic(globe=crs.globe)
```
wjbenfold, line 930: Sweet and simple, love it

wjbenfold (changes requested):
Thanks for this PR @dennissergeev!

Just one addition it would be good to get in:

### Testing

It's great that the tests all pass, though given they passed beforehand that's probably indicative of our test coverage rather than anything having improved. Let's change that by adding a test - my vote below, though if somewhere else seems more suitable then go for it.

I'd add the following or similar to lib/iris/tests/unit/analysis/cartography/test_rotate_winds.py

```
class TestNonEarthPlanet:
    def test_different_planet(self):
        u, v = uv_cubes()
        u.coord("grid_latitude").coord_system = iris.coord_systems.GeogCS(1)
        u.coord("grid_longitude").coord_system = iris.coord_systems.GeogCS(1)
        v.coord("grid_latitude").coord_system = iris.coord_systems.GeogCS(1)
        v.coord("grid_longitude").coord_system = iris.coord_systems.GeogCS(1)
        rotate_winds(u, v, iris.coord_systems.GeogCS(1))
```
I considered testing closer to the changed function but it doesn't have coverage currently as it's a private function so I think checking the rotate_winds function is a good way to go.

Before we merge this we'll also want to retarget the branch to a patch release to Iris 3.2 (probably 3.2.1) and add a what's new entry in the appropriate place to let people know what's changed, but we can worry about that another day.

dennissergeev:
Thanks for reviewing my PR @wjbenfold!

At your suggestion, I've added a test case, I hope it's good enough.

wjbenfold (approved):
Yep, better than my suggestion and it fails without your change but passes with it.

I'll check about putting it in a patch release rather than straight onto Iris main so we can get the benefit out there earlier, so don't merge it yet.

dennissergeev:
Many thanks for your help @wjbenfold!

lbdreyer:
@dennissergeev could you retarget this to target the v3.2.x branch instead of main?

 And could you also add a whats new entry to the "Bugs fixed" section in the ["3.2.1" Patches drop down box](https://github.com/SciTools/iris/blob/v3.2.x/docs/src/whatsnew/3.2.rst#v321-build_date-unreleased). You may need to rebase your branch onto the v3.2.x to do this.

Once those two are done it looks likes this would be ready to be merged

dennissergeev:
@lbdreyer sorry, what's the best way of rebasing my PR to the `v3.2.x` branch?

lbdreyer:
The commands I use are:
```
git fetch upstream
git rebase --onto upstream/v3.2.x
```

Where upstream refers to the Scitools/iris repo
```
$ git remote -v
origin	git@github.com:lbdreyer/iris.git (fetch)
origin	git@github.com:lbdreyer/iris.git (push)
upstream	git@github.com:SciTools/iris.git (fetch)
upstream	READONLY (push)
```

Assuming that works, you should then be able to edit the whats new and commit that change. Then when it comes to pushing your changes, the following command may not work:
```
git push origin <name_of_branch>
```
this is because your local branch is different (due to it now being based off v3.2.x rather than main) compared to the remote branch you are trying to update. If this is the case, you will get an error saying it failed with a message, something like "To prevent you from losing history, non-fast-forward updates were rejected ". In this case you can add the `--force` option:
```
git push origin <name_of_branch> --force
```
And that will force it to push you branch

Hope that helps!

dennissergeev:
Thanks for your help @lbdreyer, but I think something went wrong... I only now noticed I was using the `master` branch instead of `main` because I had a fork of iris before you renamed the main branch. When I force-pushed the master branch, the commits changed and my changes seem to be lost...

dennissergeev:
I'm going to do it again starting from the `v.3.2.x` branch.

lbdreyer:
>I'm going to do it again starting from the v.3.2.x branch.

Sounds like a good plan :) In case you can't remember what you changed the two commits were:
https://github.com/SciTools/iris/commit/e58a3157c61bb8a710fbd2f2b7cea6cb20fb84d8 and https://github.com/SciTools/iris/commit/bc8e1c1dd749ee676905bdfbb738b06db0378b25

PR #4597, opened 2022-02-22
Add lbcode 9 support to pp_load_rules

wjbenfold:
### Description

In response to support query, adding support for lbcode 9 (cartesian grid). Doing so by not assigning a coordinate system to the coordinates. Units in metres as per convention.

Had to stop flake checking `_all_other_rules` because it was too complicated for it and this didn't seem the time to fix that.

To do:
- [ ] Check with @pp-mo (or anyone else more familiar with this code than me) if this method is legit or if I need to do more for this to be done
- [ ] Qu: should I be setting a different name to `standard_name`, and should it be set to something different?
- [ ] Add tests (can ask original proposer for example file)
- [ ] Add whatsnew

---

### DIFF
--- a/lib/iris/fileformats/pp_load_rules.py
+++ b/lib/iris/fileformats/pp_load_rules.py
@@ -956,7 +956,7 @@ def convert(f):
     )
 
 
-def _all_other_rules(f):
+def _all_other_rules(f):  # noqa C901
     """
     This deals with all the other rules that have not been factored into any of
     the other convert_scalar_coordinate functions above.
@@ -1156,6 +1156,27 @@ def _all_other_rules(f):
         )
 
     if (
+        f.bdx != 0.0
+        and f.bdx != f.bmdi
+        and len(f.lbcode) != 5
+        and f.lbcode[0] == 9
+    ):
+        dim_coords_and_dims.append(
+            (
+                DimCoord.from_regular(
+                    f.bzx,
+                    f.bdx,
+                    f.lbnpt,
+                    standard_name="projection_x_coordinate",
+                    units="metres",
+                    circular=(f.lbhem in [0, 4]),
+                    coord_system=None,
+                ),
+                1,
+            )
+        )
+
+    if (
         f.bdy != 0.0
         and f.bdy != f.bmdi
         and len(f.lbcode) != 5
@@ -1191,6 +1212,26 @@ def _all_other_rules(f):
                     units="degrees",
                     coord_system=f.coord_system(),
                     with_bounds=True,
+                ),
+                0,
+            )
+        )
+
+    if (
+        f.bdy != 0.0
+        and f.bdy != f.bmdi
+        and len(f.lbcode) != 5
+        and f.lbcode[0] == 9
+    ):
+        dim_coords_and_dims.append(
+            (
+                DimCoord.from_regular(
+                    f.bzy,
+                    f.bdy,
+                    f.lbrow,
+                    standard_name="projection_y_coordinate",
+                    units="metres",
+                    coord_system=None,
                 ),
                 0,
             )


lib/iris/fileformats/pp_load_rules.py
```diff
@@ -1155,6 +1155,27 @@ def _all_other_rules(f):
...
+        and len(f.lbcode) != 5
+        and f.lbcode[0] == 9
+    ):
+        dim_coords_and_dims.append(
+            (
+                DimCoord.from_regular(
+                    f.bzx,
+                    f.bdx,
+                    f.lbnpt,
+                    standard_name="projection_x_coordinate",
```
lbdreyer, line 1170: The [CF standard name table](http://cfconventions.org/Data/cf-standard-names/current/build/cf-standard-name-table.html) states that "projection_x_coordinate" refers to

> "x" indicates a vector component along the grid x-axis, when this is not true longitude, positive with increasing x. Projection coordinates are distances in the x- and y-directions on a plane onto which the surface of the Earth has been projected according to a map projection. The relationship between the projection coordinates and latitude and longitude is described by the grid_mapping.

So that sounds about right to me!

lib/iris/fileformats/pp_load_rules.py
```diff
@@ -1155,6 +1155,27 @@ def _all_other_rules(f):
...
+                    units="metres",
+                    circular=(f.lbhem in [0, 4]),
+                    coord_system=None,
```
lbdreyer, line 1173: I'm unsure about this. I was wondering if this should be a type of Mercator projection?

wjbenfold: I think if it were Mercator we'd use a different code? The spec specifies Mercator as different. We'd also need a globe underlying the Mercator CS I think, which wouldn't make sense for a plane?

Will check with the user, though they seemed to be mostly relying on the workaround we gave them in the first place rather than sure what the correct answer was

lbdreyer: I see you're right about the lbcode for a mercator coord system. Although I was more trying to refer to the family of mercator (so also including transverse mercator, pseudo mercator, utm, and maybe some more that I'm not familiar with?) as those would have been my best guess for what the coord system would be, but as PP has an encoding for Mercator my guess seems unlikely!

I'm a bit uncomfortable to set the coord_system as None. Coordinates are mostly meaningless unless you have that information. Iris also makes some interesting assumptions when a coord_system is missing, for example in plotting, if no coord_system is present, it will assume PlateCarree:
https://github.com/SciTools/iris/blob/main/lib/iris/plot.py#L870-L874
 
I'm also a little concerned that in their earlier emails, the user mentions "Note the Cartesian grid can be wrap around in both X and Y," and mentions lbhem as the determining factor, but I can't see how you could interpret when int wraps in both X and Y and when it wraps only x and not Y. Also I'm not so sure what how extensive Iris' handling of non X-like circular coordinates is

wjbenfold: I think because we're dealing with idealised models (according to the user) there just isn't a projection.

I think all Iris coordinate systems (and at least all that the pp loader can load) are Geographic Coordinate Systems (and thus can't be about a flat surface). Which seems to constrain us to a) leave None there, b) put in an incorrect coordinate system, c) don't support LBCODE 9?

I can get more information on how LBHEM relates to x wrapping and y wrapping, which can inform us on how this should be written if we implement it.

In terms of whether the support is there in the rest of Iris to handle this, it should probably either refuse or try its best when asked to plot it, unless we go through everything building this case in. I've asked for information from the user about how much it's used.

wjbenfold:
Closing this, as there's a lot of underlying work that would need doing before it could be considered a complete feature.

Issue #4544, opened 2022-01-27
Document CML auto-generation


wjbenfold:
## 📚 Documentation

Currently it's quite difficult to work out that a .cml file can be recreated when running the tests if you know it should change, which leads to a temptation to do this manually (which is slow).

The Iris documentation doesn't appear to cover IRIS_TEST_CREATE_MISSING or --create-missing at all. This should be addressed in the Developers Guide (or it should be easier to find if it's in there, I haven't managed on several searches.

PR #4600, opened 2022-02-24
Developer documentation on available testing conveniences

wjbenfold:
### Description

As per #4544 - documenting how to get the CMLs to regenerate, and whilst I was there giving a rundown of test tools that are available (whilst not pretending it's a full list that would need maintaining as such, just giving a flavour).

Latest build: https://www-avd/~wbenfold/iris/developers_guide/testing_tools.html

---


docs/src/developers_guide/testing_tools.rst
```diff
@@ -0,0 +1,70 @@
...
+also consistent with the rest of Iris (which makes it easier to work with the
+code). Most of these conveniences are accessed through the
+:class:`iris.tests.IrisTest` class, from which Iris' test classes inherit.
+
+Custom assertions
+=================
+
+:class:`iris.tests.IrisTest` supports a variety of custom unittest-style
+assertions, such as :meth:`~iris.tests.IrisTest_nometa.assertStringEqual`,
+:meth:`~iris.tests.IrisTest_nometa.assertArrayEqual`,
```
pp-mo, line 19: It's a nice idea, but none of these links will work, because we don't let Sphinx include the test routines in the built API docs -- I think that is fixed [here](https://github.com/SciTools/iris/blob/v3.2.0/docs/src/sphinxext/generate_package_rst.py#L284-L285) ?

So, you could maybe make all these links target the codebase directly instead?
E.G. 
`such as :meth:'~iris.tests.IrisTest_nometa.assertStringEqual'`
--> `such as 'assertStringEqual <https://github.com/SciTools/iris/blob/main/lib/iris/tests/__init__.py#L264>'_`

Unfortunately though, those line-numbers won't be stable against code changes.
Though I guess you could invent a sphinx plugin for that ?...

pp-mo (changes requested):
This is a great idea, but the broken links let it down.
Not currently sure about how to fix that.

docs/src/developers_guide/testing_tools.rst
rcomer: Does the `literalinclude` directive help?
https://www.sphinx-doc.org/en/master/usage/restructuredtext/directives.html#directive-literalinclude

wjbenfold: Does it matter if we have links from here, or can we just double tick quote it and assume that anyone looking at the testing tools is happy to find them themselves in the codebase?

I'm not sure `literalinclude` does help @rcomer, unless I've misunderstood what you mean to do with it?

pp-mo:
Interest : I tried _including_ the test packages in the documentation.
Had to exclude a few things that were crashing it, but the _size_ of the result was interesting :
The docs build is about 56Mb without the tests (normal), and 570Mb with :open_mouth:

trexfeathers:
> This is a great idea, but the broken links let it down.
> Not currently sure about how to fix that.

@pp-mo and I discussed the various problems with using URL links to the relevant lines on GitHub. But I've thought of one more!

I believe Sphinx domains will be covered by most/all automatic refactoring tools, but I'm not so certain when it comes to freetext links, so another opportunity for things to diverge.

docs/src/developers_guide/testing_tools.rst
rcomer: Aside: I believe `assertStringEqual` could be replaced with [assertMultiLineEqual](https://docs.python.org/3/library/unittest.html#unittest.TestCase.assertMultiLineEqual).  In fact I made a [branch](https://github.com/SciTools/iris/compare/main...rcomer:remove-assertwarnsregexp) that does that.

wjbenfold: As discussed in Peloton, have laid out the breadcrumbs for devs to follow in the absence of handy links.

wjbenfold:
Updated latest build (https://www-avd/~wbenfold/iris/developers_guide/testing_tools.html)

wjbenfold:
Rebased for what's new

jamesp (approved):
This all seems good @wjbenfold.

Will it need to be updated/added to when the new graphics checking functionality is merged?

wjbenfold:
> Will it need to be updated/added to when the new graphics checking functionality is merged?

It will, but I'll make those changes on that PR

docs/src/developers_guide/testing_tools.rst
jamesp: @pp-mo are you happy with this as it stands now to go in?

pp-mo:
Sorry @wjbenfold just realised I was blocking this with unresolved comment.
And now it needs a rebase 🤦

wjbenfold:
I've rebased. I think you're still blocking it with a "changes requested" review and you need to submit a new "approving" review @pp-mo

pp-mo (approved):

PR #4601, opened 2022-02-24
Revert plotting-vs-y

### LINKED ISSUE #4493: Shouldn't the 3.1 change to line plot defaults to plot the independent variable vertically if it's vertical or "y" have been only for vertical ones?
https://github.com/SciTools/iris/pull/3906
seems overdone to me.

https://scitools-iris.readthedocs.io/en/stable/whatsnew/3.1.html?highlight=vertical%20axis#features
says it was "to automatically place the cube on the x axis if the primary coordinate being plotted against is a vertical coordinate" for line plots.  That is such standard practice in Earth science that I am confident it would have been coded from the start in a package created by scientists from the field, so I'm totally in favour.

But what was actually implemented is, as stated at
https://github.com/SciTools/iris/pull/3906
to "Put cube data on the x axis if plotting just a cube against a vertical or y coordinate", & the addition of "y coordinate" is wrong in my opinion.

I can see how programmers could assume that a dimension that is called "y" by iris would be thought of as "y-axis-like" by scientists, but this is not true.  Graphs plotted against latitude conventionally have latitude as the x-axis, the conventional place for the independent variable.  (There is one important exception: when the graph against latitude is lined up beside a map, to show e.g. the zonal mean, when of course the map normally has latitude running vertically on the page - but this is a fairly small fraction of cases.  I can't remember ever doing it in my career.)

I think any look at the literature will confirm this.  Basically I suppose we see the "y axis" of line plots as a vertical, while python's "Y" is horizontal (in all my experience).  Where a 2D field is being plotted & one non-vertical dimension has to be vertical on paper, yes, latitude is normally the one, following the general convention for maps, but that's another issue.

So I reckon the "y" part should be reverted - preferably soon, not waiting till lots of people are using a version with the change, disliking it & saying so!  (I don't suggest testing for the exceptional case I mentioned as I doubt iris can access the details of a previous plot, let alone a following one, & it needs the zonal-mean plot much narrower than the map so the user will need to get into matplotlib.gridspec anyway, & the change to the call to iris plot to flip from the default is comparatively simple.)

rcomer:
### Description

Closes #4493.  Cube data should now automatically go on the x-axis of an `iplt.plot` only if plotting against a vertical coordinate.

Since this is reverting to previous behaviour, for the affected tests I just looked in #3906 for what the good hashes used to be, and reinstated those.  It seems to me a good idea to do this before @wjbenfold starts moving the image-hashing infrastructure around! 😅

---

### DIFF
--- a/docs/src/whatsnew/dev.rst
+++ b/docs/src/whatsnew/dev.rst
@@ -37,7 +37,9 @@ This document explains the changes made to Iris for this release
 🐛 Bugs Fixed
 =============
 
-#. N/A
+#. `@rcomer`_ reverted part of the change from :pull:`3906` so that
+   :func:`iris.plot.plot` no longer defaults to placing a "Y" coordinate (e.g.
+   latitude) on the y-axis of the plot. (:issue:`4493`, :pull:`4601`)
 
 
 💣 Incompatible Changes
--- a/lib/iris/plot.py
+++ b/lib/iris/plot.py
@@ -673,7 +673,7 @@ def _get_plot_objects(args):
         if (
             isinstance(v_object, iris.cube.Cube)
             and isinstance(u_object, iris.coords.Coord)
-            and iris.util.guess_coord_axis(u_object) in ["Y", "Z"]
+            and iris.util.guess_coord_axis(u_object) == "Z"
         ):
             u_object, v_object = v_object, u_object
             u, v = v, u
--- a/lib/iris/tests/results/imagerepo.json
+++ b/lib/iris/tests/results/imagerepo.json
@@ -684,7 +684,10 @@
         "https://scitools.github.io/test-iris-imagehash/images/v4/8bfe956b7c01c2f26300929dfc1e3c6690736f91817e3b0c84be6be5d1603ed1.png"
     ],
     "iris.tests.test_plot.TestPlot.test_y.0": [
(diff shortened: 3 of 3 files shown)


lbdreyer (approved):
I'm in support of this. It does make much more sense (to me at least!) for the default behaviour to only put z-type axis on the y-xis of a plot. If a user wants a `latitude`/`grid_latitude`/`projection_y_coordinate` coord on their y-axis they can always configure that manually.

Assuming there are no objections, this just needs a whatsnew then I would be happy to merge it

rcomer:
Thanks @lbdreyer, which category of the whatsnew should this go in?  Does it constitutes a bugfix if the previous change was deliberate (although we've changed our minds)?  Or should it go under "Features" or something else?

lbdreyer:
> Thanks @lbdreyer, which category of the whatsnew should this go in? Does it constitute a bugfix if the previous change was deliberate (although we've changed our minds)? Or should it go under "Features" or something else?

A very difficult question! I always find determining what is/isn't a bug fix a very disputable issue!

I am leaning towards a "bug fix", with the following reasoning: A bug is something that produces an incorrect result. We have now decided that defaulting to put y-like coords on the y-axis of a plot is the incorrect behaviour. I think our original decision was wrong and we are now fixing it and that doesn't sound appropriate for a feature (I think a user may find this a bit strange when reading through the what's new feature list)

Although, I do think you could easily argue for either case (bug fix/feature) so I'm not terribly fussed which you go for.

rcomer:
Thanks @lbdreyer whatsnew now added:

image

lbdreyer:
Thanks @rcomer !

PR #4610, opened 2022-03-01
Avoid dimensionality mismatch between data array and CF variables.

### LINKED ISSUE #4599: Collapsing over a lazy auxiliary coordinate results in a `netCDF4` error when the cube is saved as `*.nc` file.
## 🐛 Bug Report

Collapsing a cube over a lazy auxiliary coordinate results in the following error when the cube is save as `nc` file:

```
ValueError: slicing expression exceeds the number of dimensions of the variable
```

I tested this on two independent machines. Some additional hints:

- The error only appears in `iris=3.2.0.post0`, not for `iris=3.1.0`.
- Adding a `print(cube)` between the collapsing the the `save` somehow fixes this issue.
- Using a non-lazy auxiliary coordinate also fixes this issue.
- Using a lazy dimensional coordinate also does not trigger the error (if used without a lazy auxiliary coordinate).

## How To Reproduce

```py
import dask.array as da
import numpy as np
import os

import iris
from iris.coords import AuxCoord, DimCoord
from iris.cube import Cube


print("iris version:", iris.__version__)


# Create cube with lazy aux coord and aggregate over this dimension
dim_coord = DimCoord(np.arange(10), var_name='time')
aux_coord = AuxCoord(da.arange(10), var_name='year')  # the "da" is important here!
cube = Cube(np.arange(10),
            var_name='x',
            dim_coords_and_dims=[(dim_coord, 0)],
            aux_coords_and_dims=[(aux_coord, 0)],
           )

cube = cube.collapsed('time', iris.analysis.MEAN)


# Adding a print() somehow fixes this issue
# print(cube)


# Saving this cube gives the error
filename = os.path.expanduser('~/test_iris_32.nc')
iris.save(cube, filename)
```

## Expected behaviour

No error, similar to `iris=3.1.0`.

## Environment 
 - OS & Version: openSUSE Tumbleweed 20220221; Red Hat Enterprise Linux Server release 6.10
 - Iris Version: `3.2.0.post0`

## Additional context

<details>
<summary>Full traceback</summary>

```
  File "/home/manuel/Tresorit/DLR/scripts/iris/iris32bug.py", line 31, in <module>
    iris.save(cube, filename)
  File "/home/manuel/mambaforge/envs/xxx/lib/python3.10/site-packages/iris/io/__init__.py", line 436, in save
    saver(source, target, **kwargs)
  File "/home/manuel/mambaforge/envs/xxx/lib/python3.10/site-packages/iris/fileformats/netcdf.py", line 3167, in save
    sman.write(
  File "/home/manuel/mambaforge/envs/xxx/lib/python3.10/site-packages/iris/fileformats/netcdf.py", line 1242, in write
    self._add_aux_coords(cube, cf_var_cube, cube_dimensions)
  File "/home/manuel/mambaforge/envs/xxx/lib/python3.10/site-packages/iris/fileformats/netcdf.py", line 1569, in _add_aux_coords
    return self._add_inner_related_vars(
  File "/home/manuel/mambaforge/envs/xxx/lib/python3.10/site-packages/iris/fileformats/netcdf.py", line 1534, in _add_inner_related_vars
    cf_name = self._create_generic_cf_array_var(
  File "/home/manuel/mambaforge/envs/xxx/lib/python3.10/site-packages/iris/fileformats/netcdf.py", line 2407, in _create_generic_cf_array_var
    self._create_cf_bounds(element, cf_var, cf_name)
  File "/home/manuel/mambaforge/envs/xxx/lib/python3.10/site-packages/iris/fileformats/netcdf.py", line 2069, in _create_cf_bounds
    self._lazy_stream_data(
  File "/home/manuel/mambaforge/envs/xxx/lib/python3.10/site-packages/iris/fileformats/netcdf.py", line 2899, in _lazy_stream_data
    is_masked, contains_fill_value = store(
  File "/home/manuel/mambaforge/envs/xxx/lib/python3.10/site-packages/iris/fileformats/netcdf.py", line 2874, in store
    da.store([data], [target])
  File "/home/manuel/mambaforge/envs/xxx/lib/python3.10/site-packages/dask/array/core.py", line 1163, in store
    compute_as_if_collection(Array, store_dsk, map_keys, **kwargs)
  File "/home/manuel/mambaforge/envs/xxx/lib/python3.10/site-packages/dask/base.py", line 317, in compute_as_if_collection
    return schedule(dsk2, keys, **kwargs)
  File "/home/manuel/mambaforge/envs/xxx/lib/python3.10/site-packages/dask/threaded.py", line 81, in get
    results = get_async(
  File "/home/manuel/mambaforge/envs/xxx/lib/python3.10/site-packages/dask/local.py", line 506, in get_async
    raise_exception(exc, tb)
  File "/home/manuel/mambaforge/envs/xxx/lib/python3.10/site-packages/dask/local.py", line 314, in reraise
    raise exc
  File "/home/manuel/mambaforge/envs/xxx/lib/python3.10/site-packages/dask/local.py", line 219, in execute_task
    result = _execute_task(task, data)
  File "/home/manuel/mambaforge/envs/xxx/lib/python3.10/site-packages/dask/core.py", line 119, in _execute_task
    return func(*(_execute_task(a, cache) for a in args))
  File "/home/manuel/mambaforge/envs/xxx/lib/python3.10/site-packages/dask/array/core.py", line 4164, in store_chunk
    return load_store_chunk(x, out, index, lock, return_stored, False)
  File "/home/manuel/mambaforge/envs/xxx/lib/python3.10/site-packages/dask/array/core.py", line 4151, in load_store_chunk
    out[index] = x
  File "/home/manuel/mambaforge/envs/xxx/lib/python3.10/site-packages/iris/fileformats/netcdf.py", line 972, in __setitem__
    self.target[keys] = arr
  File "src/netCDF4/_netCDF4.pyx", line 4903, in netCDF4._netCDF4.Variable.__setitem__
  File "/home/manuel/mambaforge/envs/xxx/lib/python3.10/site-packages/netCDF4/utils.py", line 335, in _StartCountStride
    raise ValueError("slicing expression exceeds the number of dimensions of the variable")
ValueError: slicing expression exceeds the number of dimensions of the variable
```
</details>

trexfeathers:
### Description

Closes #4599.

Corrects a problem where bounds for a scalar coordinate - still in a 2D array - where hitting a 'broadcast ambiguity' trying to save to a 1D CF variable.

---

### DIFF
--- a/lib/iris/fileformats/netcdf.py
+++ b/lib/iris/fileformats/netcdf.py
@@ -2865,10 +2865,11 @@ class Saver:
 
     @staticmethod
     def _lazy_stream_data(data, fill_value, fill_warn, cf_var):
-        if data.shape == (1,) + cf_var.shape:
+        if hasattr(data, "shape") and data.shape == (1,) + cf_var.shape:
             # Reduce dimensionality where bounds data is for a scalar point -
             #  bounds data is 2D but contains just 1 row, which causes
             #  broadcast ambiguity between the data and its 1D cf_var.
+            # (Don't do this check for string data).
             data = np.atleast_1d(data[0])
 
         if is_lazy_data(data):
--- a/lib/iris/fileformats/netcdf.py
+++ b/lib/iris/fileformats/netcdf.py
@@ -2866,11 +2866,12 @@ class Saver:
     @staticmethod
     def _lazy_stream_data(data, fill_value, fill_warn, cf_var):
         if hasattr(data, "shape") and data.shape == (1,) + cf_var.shape:
-            # Reduce dimensionality where bounds data is for a scalar point -
-            #  bounds data is 2D but contains just 1 row, which causes
-            #  broadcast ambiguity between the data and its 1D cf_var.
             # (Don't do this check for string data).
-            data = np.atleast_1d(data[0])
+            # Reduce dimensionality where the data array has an extra dimension
+            #  versus the cf_var - to avoid a broadcasting ambiguity.
+            # Happens when bounds data is for a scalar point - array is 2D but
+            #  contains just 1 row, so the cf_var is 1D.
+            data = data.squeeze()
 
         if is_lazy_data(data):
 
--- a/lib/iris/fileformats/netcdf.py
+++ b/lib/iris/fileformats/netcdf.py
@@ -2865,6 +2865,12 @@ class Saver:
 
     @staticmethod
     def _lazy_stream_data(data, fill_value, fill_warn, cf_var):
+        if data.shape == (1,) + cf_var.shape:
+            # Reduce dimensionality where bounds data is for a scalar point -
+            #  bounds data is 2D but contains just 1 row, which causes
+            #  broadcast ambiguity between the data and its 1D cf_var.
+            data = np.atleast_1d(data[0])
+
         if is_lazy_data(data):
 
             def store(data, cf_var, fill_value):
--- a/lib/iris/fileformats/netcdf.py
+++ b/lib/iris/fileformats/netcdf.py
@@ -2871,7 +2871,7 @@ class Saver:
             #  versus the cf_var - to avoid a broadcasting ambiguity.
             # Happens when bounds data is for a scalar point - array is 2D but
             #  contains just 1 row, so the cf_var is 1D.
-            data = data.squeeze()
+            data = data.squeeze(axis=0)
 
         if is_lazy_data(data):
 
--- a/lib/iris/tests/unit/fileformats/netcdf/test_Saver.py
+++ b/lib/iris/tests/unit/fileformats/netcdf/test_Saver.py
@@ -30,7 +30,7 @@ from iris.coord_systems import (
     TransverseMercator,
     VerticalPerspective,
 )
-from iris.coords import DimCoord
+from iris.coords import AuxCoord, DimCoord
 from iris.cube import Cube
 from iris.fileformats.netcdf import Saver
 import iris.tests.stock as stock
@@ -298,6 +298,21 @@ class Test_write(tests.IrisTest):
             with Saver(nc_path, "NETCDF4") as saver:
                 saver.write(cube)
             self.assertCDL(nc_path)
+
+    def test_dimensional_to_scalar(self):
+        # Bounds for 1 point are still in a 2D array.
+        scalar_bounds = self.array_lib.arange(2).reshape(1, 2)
+        scalar_point = scalar_bounds.mean()
+        scalar_data = self.array_lib.zeros(1)
+        scalar_coord = AuxCoord(points=scalar_point, bounds=scalar_bounds)
+        cube = Cube(scalar_data, aux_coords_and_dims=[(scalar_coord, 0)])[0]
+        with self.temp_filename(".nc") as nc_path:
+            with Saver(nc_path, "NETCDF4") as saver:
+                saver.write(cube)
+            ds = nc.Dataset(nc_path)
+            # Confirm successful save of 2D bounds array into single dimension.
+            self.assertEqual(2, ds.dimensions["bnds"].size)
+            ds.close()
 
 
 class Test__create_cf_bounds(tests.IrisTest):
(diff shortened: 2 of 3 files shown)


lib/iris/fileformats/netcdf.py
```diff
@@ -2865,6 +2865,13 @@ def _increment_name(self, varname):
 
     @staticmethod
     def _lazy_stream_data(data, fill_value, fill_warn, cf_var):
+        if hasattr(data, "shape") and data.shape == (1,) + cf_var.shape:
+            # Reduce dimensionality where bounds data is for a scalar point -
+            #  bounds data is 2D but contains just 1 row, which causes
+            #  broadcast ambiguity between the data and its 1D cf_var.
+            # (Don't do this check for string data).
+            data = np.atleast_1d(data[0])
```
trexfeathers, line 2873: @wjbenfold tells me that numpy `squeeze` can do this.

pp-mo: Well it can, but it would reduce "any" length-1 dims.
So I'd say the existing code is clearer in terms of exactly what we need to do here.

But if we're now avoiding that 'string list' case, do we still need the `atleast_1d` call ?
It would be simpler still if we avoid the need for that.

trexfeathers: d786c7ac2c468b7b429a0066358c9ec67fae760d

rcomer: `numpy.squeeze` takes the `axis` keyword.  Would `np.squeeze(data, axis=0)` do what you need?

trexfeathers: Thanks all for your help this morning.

### Cases where `data[0]` creates a single, non-array value

@pp-mo wasn't expecting this, so this needed investigating to check nothing nasty is happening.

For scalar coordinates, when streaming the point data we get `data.shape == (1,)` and `cf_var.shape == ()`, which meets this check:

https://github.com/SciTools/iris/blob/19efa6582894bf4a9e3773463b40e6e8a0905c03/lib/iris/fileformats/netcdf.py#L2868

### Correct use of `squeeze()`

Considering the above, I still think the correct solution is to include insurance that the `data` will still be an array, even after removing a dimension. But I appreciate now that `data.squeeze()` could end up removing other axes that we weren't planning to mess with. So @rcomer's suggestion to include `axis=0` seems like the right fix - the `if` line makes sure we don't end up making some other, unexpected modification to `data`. 19efa6582894bf4a9e3773463b40e6e8a0905c03

lib/iris/tests/unit/fileformats/netcdf/test_Saver.py
```diff
@@ -299,6 +299,22 @@ def test_with_climatology(self):
                 saver.write(cube)
             self.assertCDL(nc_path)
 
+    def test_dimensional_to_scalar(self):
```
pp-mo, line 302: I'm finding that this test did not **_fail_** for me with iris 3.1.0, which it really ought to.
I may have got something wrong, but assuming not : can we fix that ??

trexfeathers: I introduced the bug in Iris 3.2. Before that we were not lazily streaming bounds data, so never encountered the problem.

trexfeathers: ... and I believe it would be impossible to replicate for `Cube.data` itself, due to the specific ways we handle that, so there's no test that can be written to fail for versions <`v3.2`.

pp-mo: OK sorry, I will check this again with latest-master ...

trexfeathers: Also: I expect this to always pass when run within `test_Saver.Test_write`, but without the fix it should fail as part of `test_Saver__lazy.Test_write`. Does that make sense?

pp-mo: Apologies, I see now that just running the new version of `test_Saver.py` does not exercise the correct test.
It is essential that I _replace_ `test_Saver.py` with the new one, and then run `test_Saver__lazy.py`.
Which is a bit subtle + I totally missed it !

pp-mo: Ah ok!
I think I hadn't seen that there were 2 instances of this, and was running the wrong one ...

trexfeathers:
Thanks @pp-mo!

PR #4616, opened 2022-03-02
Scalar Scatter Plot

rcomer:
### Description

This may be a slightly niche requirement, but it's also pretty easy to fix.  Matplotlib will happily let you make a scatter plot with a single point, e.g.

```python

plt.scatter(1, 1)
plt.show()
```

If you try the same with scalar Iris cubes

```python
import iris.plot as iplt
from iris.cube import Cube

iplt.scatter(Cube(1), Cube(1))
```

you currently get

<details><summary>Traceback</summary>

```
TypeError                                 Traceback (most recent call last)
Input In [6], in <module>
----> 1 iplt.scatter(Cube(1), Cube(1))

File [path]/iris/lib/iris/plot.py:1585, in scatter(x, y, *args, **kwargs)
   1583 args = (x, y) + args
   1584 _plot_args = None
-> 1585 return _draw_1d_from_points("scatter", _plot_args, *args, **kwargs)

File [path]/iris/lib/iris/plot.py:770, in _draw_1d_from_points(draw_method_name, arg_func, *args, **kwargs)
    763 def _draw_1d_from_points(draw_method_name, arg_func, *args, **kwargs):
    764     # NB. In the interests of clarity we use "u" to refer to the horizontal
    765     # axes on the matplotlib plot and "v" for the vertical axes.
   (...)
    768     # axes (cubes or coordinates) and their respective values, along with the
    769     # argument tuple with these objects removed
--> 770     u_object, v_object, u, v, args = _get_plot_objects(args)
    772     # Track numpy arrays to use for the actual plotting.
    773     plot_arrays = []

File [path]/iris/lib/iris/plot.py:655, in _get_plot_objects(args)
    653 u, v = _uv_from_u_object_v_object(u_object, v_object)
    654 args = args[2:]
--> 655 if len(u) != len(v):
    656     msg = (
    657         "The x and y-axis objects are not compatible. They should "
    658         "have equal sizes but got ({}: {}) and ({}: {})."
    659     )
    660     raise ValueError(
    661         msg.format(u_object.name(), len(u), v_object.name(), len(v))
    662     )

TypeError: len() of unsized object

```
</details>

This PR avoids that error.  [_uv_from_u_object_v_object](https://github.com/SciTools/iris/blob/c83722b0b5241958309c2843d9d715f012faa48b/lib/iris/plot.py#L625-L636) has checks to make sure our objects are <=1D and as far as I can tell, the outputs are always numpy arrays [type hinting would help here!]  So it seems safe to use the `.size` attribute instead of checking the length.

---

### DIFF
--- a/docs/src/whatsnew/dev.rst
+++ b/docs/src/whatsnew/dev.rst
@@ -41,6 +41,9 @@ This document explains the changes made to Iris for this release
 #. `@rcomer`_ reverted part of the change from :pull:`3906` so that
    :func:`iris.plot.plot` no longer defaults to placing a "Y" coordinate (e.g.
    latitude) on the y-axis of the plot. (:issue:`4493`, :pull:`4601`)
+   
+#. `@rcomer`_ enabled passing of scalar objects to :func:`~iris.plot.plot` and 
+   :func:`~iris.plot.scatter`. (:pull:`4616`)
 
 
 💣 Incompatible Changes


lib/iris/tests/unit/plot/test__get_plot_objects.py
```diff
@@ -0,0 +1,29 @@
...
+import iris.tests as tests  # isort:skip
+
+import iris.cube
+
+if tests.MPL_AVAILABLE:
+    from iris.plot import _get_plot_objects
+
+
+@tests.skip_plot
+class Test_get_plot_objects(tests.IrisTest):
```
wjbenfold, line 19: Should this have an extra `_` after `Test` for naming consistency?

wjbenfold, line 19: Given we're adding tests for `_get_plot_objects`, we could add test for the other modes it has too. Though those are presumably covered elsewhere so I don't mind.

lib/iris/plot.py
```diff
@@ -652,7 +652,7 @@ def _get_plot_objects(args):
         u_object, v_object = args[:2]
         u, v = _uv_from_u_object_v_object(u_object, v_object)
         args = args[2:]
-        if len(u) != len(v):
```
wjbenfold, line 655: Test breaks without this and doesn't with, which is good.

Also got a couple of image tests break when I checked out the version of this on `main` instead of your version though - any idea why?

wjbenfold (changes requested):
I think that https://github.com/SciTools/iris/blob/50b923d252539abeffa5941d932525d8288e7798/lib/iris/plot.py#L661 should use `.size` too?

rcomer:
> I think that
> 
> https://github.com/SciTools/iris/blob/50b923d252539abeffa5941d932525d8288e7798/lib/iris/plot.py#L661
> 
> should use `.size` too?

Ooops! 😦   Yes, excellent catch.  I've now added an extra test that proved this was necessary, and made that change.

lib/iris/tests/unit/plot/test__get_plot_objects.py
rcomer: I copied from another file in this directory but, now that I look at a few, I see there is no consistency.  So agreed.  Done.

rcomer: > presumably covered elsewhere

The fact that my initial change would have broken the error message rather suggests we don't have good coverage for this function.  I think I'd like to think about this more next week when I'm feeling more awake!

rcomer: Looks like [Test1dScatter](https://github.com/SciTools/iris/blob/35d97ed9bed0ccefa6d940d7d1d40a89b6ce0d24/lib/iris/tests/test_plot.py#L245) covers most of the eventualities.

lib/iris/plot.py
wjbenfold: Seems likely I hadn't updated main recently enough - I don't see these failures now

lib/iris/tests/unit/plot/test__get_plot_objects.py
rcomer: OK, I think I've convinced myself that all the various input permutations (one or two args, coords and/or cubes, different size inputs) for `_get_plot_objects` are exercised by [Test1dPlotMultiArgs](https://github.com/SciTools/iris/blob/35d97ed9bed0ccefa6d940d7d1d40a89b6ce0d24/lib/iris/tests/test_plot.py#L143) and [Test1dScatter](https://github.com/SciTools/iris/blob/35d97ed9bed0ccefa6d940d7d1d40a89b6ce0d24/lib/iris/tests/test_plot.py#L245).  The bit about plotting against a vertical coordinate was added quite recently (#3906) and, as well as updating some hashes for `Test1dPlotMultiArgs`, it [added specific quickplot tests](https://github.com/SciTools/iris/pull/3906/files#diff-66d543e25830127bbe1610fc11a967ab98604a3254f39ccb6851cb925dc0e94d).  So although it didn't have direct unit tests, I think we do have thorough coverage for this function.

This does make me wonder whether I've done the right thing in adding this module and should have added some tests for `plot` and/or `scatter` instead?

wjbenfold: I was tempted to suggest adding your new tests to Test1dScatter rather than making separate unit tests that aren't overly linked back to the main ones. I think I like unit testing this rather than just adding more integration tests (with images particularly) though, so happy with this choice.

lib/iris/tests/unit/plot/test__get_plot_objects.py
```diff
@@ -0,0 +1,37 @@
...
+@tests.skip_plot
+class Test__get_plot_objects(tests.IrisTest):
+    def test_scalar(self):
+        cube1 = iris.cube.Cube(1)
+        cube2 = iris.cube.Cube(1)
+        expected = (cube1, cube2, 1, 1, ())
+        result = _get_plot_objects((cube1, cube2))
+        self.assertTupleEqual(expected, result)
+
+    def test_mismatched_size(self):
```
wjbenfold, line 27: We could check that making either a scalar gives the right error message here (without failing)? Not too fussed though as the similar test in `test_plot` doesn't.

wjbenfold (approved):

rcomer:
Thanks for all your feedback @wjbenfold, I've added that last test and also a whatsnew as I assume we are now close 🙂

PR #4637, opened 2022-03-11
Mergeback the v3.2.x branch onto main

lbdreyer:
### Description
Mergeback after the v3.2.1 release.

Note: to merge this in we will need to switch off the setting that only allows "squash and commit" merges, as we have done with previous mergebacks

---


trexfeathers (approved):
This has everything I would expect to see, and nothing I wouldn't expect to see.

>Note: to merge this in we will need to switch off the setting that only allows "squash and commit" merges, as we have done with previous mergebacks

I don't have this power, so I'll just leave my approval here.

pp-mo:
> > Note: to merge this in we will need to switch off the setting that only allows "squash and commit" merges, as we have done with previous mergebacks

> I don't have this power, so I'll just leave my approval here.

Ping @bjlittle.  I think you are the one with "the power" ? 🏋️‍♂️

pp-mo:
Ping @bjlittle : I think you are the one with "the power" ? 🏋️

bjlittle (approved):
@lbdreyer Awesome, thanks! LGTM 👍

PR #4640, opened 2022-03-15
Graphics test change to rehash every time

wjbenfold:
### Description

This changes the graphics testing setup to re-phash images (found in the test data under test_data/images) every time to check them against the test results.

I've also pulled out the graphics testing infrastructure into a subfolder of iris/tests

Feedback requested on:

- Overall approach
- Structure of directories / files
- The current implementation of idiff checks that the phash isn't already in `repodata.json`. I've removed `repodata.json` so I'm checking the new file against being the same as any existing ones. This could probably be improved, but I'm not sure what's better to test against.

Known to-dos:

- Attach decorators to skip tests that use reference images when test data is missing
- Documentation (specifically developers guide) update
- What's new

---


rcomer:
> Update decorators so that tests marked skip_plot are also skipped if there's no test data

Note that not all tests marked `@skip_plot` use a reference image.  E.g none of these I think: https://github.com/SciTools/iris/blob/main/lib/iris/tests/unit/plot/test_contourf.py

wjbenfold:
> Note that not all tests marked `@skip_plot` use a reference image. E.g none of these I think: https://github.com/SciTools/iris/blob/main/lib/iris/tests/unit/plot/test_contourf.py

Good point - have changed the to-do

rcomer:
Or we could be really radical and drop both `@skip_plot` and `@skip_data` altogether.  We already removed the “minimal” version of the tests from the CI at #4218.  Matplotlib is listed as a core dependency so it seems unlikely that anyone would run the tests without it.  We could also just insist that if you’re going to run the tests, you need the test data.  I’m not sure under what circumstances you’d know it was safe to only use the tests that don’t need the test data.

wjbenfold:
> Matplotlib is listed as a core dependency

Why do we spend some much time checking if it exists then? :/

https://github.com/SciTools/iris/blob/main/setup.cfg#L47-L52

https://github.com/SciTools/iris/blob/main/lib/iris/tests/__init__.py#L58-L74

> I’m not sure under what circumstances you’d know it was safe to only use the tests that don’t need the test data.

I'm pretty sure I've done this when I've put Iris on a non-internetted machine and wanted to check the install was basically functional? Probably not a huge extra effort to grab the test data though

trexfeathers:
Just so you know... offline discussions have been developing the concept of 'core Iris' to the point that Matplotlib becomes an optional dependency - only some use cases include visualisation.

At the same time, the impending inclusion of [more optional dependencies for Mesh handling](https://scitools-iris.readthedocs.io/en/latest/further_topics/ugrid/partner_packages.html) - each of which brings a sizeable dependency stack - has lead to more granular tests being on the table again:

* Iris tests with core dependencies only
  Avoids spending too much resource on environment construction
* 1 or more test sessions of Iris working with its optional dependencies
  More resource on environment construction, but way fewer tests to complete

Of course if this became a reality then we'd probably need a more 'global' way of skipping optional dependencies, rather than the decorators we have now.

pp-mo:
N.B. https://github.com/SciTools/iris/discussions/4465 refers

bjlittle (changes requested):
@wjbenfold I really love your thinking and the way you've approached this, it seems like we're on the cusp of a much more robust graphical workflow for testing, but I have one high level recommendation for change, and one objection.

Firstly, given the accompanying PR https://github.com/SciTools/iris-test-data/pull/71, we could take this opportunity to completely rationalise the proliferation of references images that we care about. The genius game changer that you've tapped into there is that we version the `iris-test-data` repo. Therefore, there is now no reason to explicitly maintain several baseline images for a graphical test; only keep the latest baseline image i.e., one test = one image (we still need to support graphical tests that generate multiple images from multiple sub-tests within that one test). Previously we'd explicitly maintained baseline image history within the `imagerepo.json` file, but now we don't need to do that as the history of the baseline images per test is implicitly recorded for free by GH versioning the `iris-test-data` repo through time. All we need to do now is also lock down the version of the test data that is valid for the tests along with the underlying software dependencies that we're collectively running testing with and I think this will make a much more maintainable workflow. (As an aside, I've got thoughts on how we can automate the availability and verification of test data based on versioning - but that's for another time)

Secondly, I completely understand your motivation to dump the `imagerepo.json` file, and calculate the **expected** perceptual image hash at runtime. However, this will make graphical testing slower (which for me, is neither here nor there really, it shouldn't incur a massive additional cost, so whatevz really) BUT, more importantly, we're now exposing ourselves to **not knowing** when the **expected** perceptual image hash of a baseline image **changes**. That for me is a bit concerning, and after a bit of thought I'm not in favour of your proposed approach (happy to be convinced otherwise though). I think it's important for us to know exactly when the hash of the baseline image has unexpectedly changed and appropriately manage the fallout (and that typically manifests itself though a graphical test failure due to a larger hamming distance with the associated actual result image hash)... but this fallout will be massively easier to manage given your lovely proposal to version the baseline images. So, that said, I'd prefer that we maintained a single pre-computed expected perceptual image hash associated with each test (and each sub-test within a test) - IMHO there is still important value in doing that.

Anyways, happy to discuss

wjbenfold:
@bjlittle To refer back to what the imagehash was in the past, it sounds like a little json as a repository of our imagehashes might just do the job...

Said `imagerepo.json` (or another name if leaving it the same will cause confusion) would conveniently allow us to run the plotting tests even in the absence of the test data (though is such a state we wouldn't be robust to the version of imagehash/pillow having changed). Do you think that's a feature to add, or shall we just update it so that the change to an imagehash shows up in the diff but never read from it?

bjlittle:
@wjbenfold I'd recommend keeping state in `json`. In fact we're headed in that direction when we adopt `pytest-mpl`.

For me the goal here is to reduce the maintenance burden when imagehash/pillow causes changes, not for graphical tests to silently continue passing by adaptively changing the expected result on the fly. In hindsight, our historical mistake was not versioning the image repo to preserve known baseline snapshots, and also attempting to cope with such dependency changes through a maze of soft links, which compounds the problem.

We can certainly avoid all that, given your approach, but for me it's important that we still know when a change in dependencies causes graphical test failures, and in order to do that we need the hashes of the baseline images captured; that way we know exactly what we're aiming for.

I think this is a minor change, but makes a big difference.

wjbenfold:
Replaced by #4759

Issue #3312, opened 2019-05-14
Iris causes AttributeError when attempting to catch an OSError when loading a dataset


evertrol:
When attempting to open a NetCDF dataset in Iris, I wrap it in a `try: ... except OSError` block, to catch bad files or URLs. This can, however, cause an `AttributeError` instead.

An example (the corrupted file is truncated, while the HadCRUT dataset is not served by an OpenDAP server):

```
import iris
url = 'corrupted.nc'
url = 'https://crudata.uea.ac.uk/cru/data/temperature/HadCRUT.4.6.0.0.median.nc'
try:
    cube = iris.load(url)
except OSError:
    print('>>><<<')
print('continuing')
```

Note that while the `AttributeError` is raised, it is also ignored, and the script will continue.
The traceback and output for the corrupted file is:
```
HDF5-DIAG: Error detected in HDF5 (1.10.4) thread 4563412416:
  #000: H5F.c line 509 in H5Fopen(): unable to open file
    major: File accessibilty
    minor: Unable to open file
  #001: H5Fint.c line 1400 in H5F__open(): unable to open file
    major: File accessibilty
    minor: Unable to open file
  #002: H5Fint.c line 1700 in H5F_open(): unable to read superblock
    major: File accessibilty
    minor: Read failed
  #003: H5Fsuper.c line 623 in H5F__super_read(): truncated file: eof = 21365110, sblock->base_addr = 0, stored_eof = 21365116
    major: File accessibilty
    minor: File has been truncated
>>><<<
Exception ignored in: <function CFReader.__del__ at 0x1147f51e0>
Traceback (most recent call last):
  File "/usr/local/miniconda3/envs/testiris/lib/python3.7/site-packages/iris/fileformats/cf.py", line 1124, in __del__
    self._dataset.close()
AttributeError: 'CFReader' object has no attribute '_dataset'
continuing
```
For the HadCRUT dataset, it is a bit longer: essentially a HTML page that netCDF4 can't parse (presumably, it tries to access OpenDAP API entry points).

----

The likely cause of this is Python's garbage collection attempting to clean up intermediate variables. In this case, it runs [the `__del__` method of `CFReader`](https://github.com/SciTools/iris/blob/master/lib/iris/fileformats/cf.py#L1122), which will unconditionally perform `self._dataset.close()` (it is probably also the garbage collector that safely ignores the raised exception inside the `__del__` method).
[...]

PR #4646, opened 2022-03-18
Catch `AttributeError` from removing `CFReader._dataset` when it doesn't exist

wjbenfold:
### Description

Based on #3312 and taking @evertrol up on the offer of replicating the changes in #3336.

Rather than setting `_dataset` to `None` I've taken the approach of catching and ignoring the attribute error on file close (OSErrors etc. will still be raised).

---

### DIFF
--- a/lib/iris/fileformats/cf.py
+++ b/lib/iris/fileformats/cf.py
@@ -1044,6 +1044,7 @@ class CFReader:
     CFGroup = CFGroup
 
     def __init__(self, filename, warn=False, monotonic=False):
+        self._dataset = None
         self._filename = os.path.expanduser(filename)
 
         #: Collection of CF-netCDF variables associated with this netCDF file
@@ -1295,7 +1296,8 @@ class CFReader:
 
     def __del__(self):
         # Explicitly close dataset to prevent file remaining open.
-        self._dataset.close()
+        if self._dataset is not None:
+            self._dataset.close()
 
 
 def _getncattr(dataset, attr, default=None):
--- a/docs/src/whatsnew/latest.rst
+++ b/docs/src/whatsnew/latest.rst
@@ -71,6 +71,10 @@ This document explains the changes made to Iris for this release
    coordinate's direction is reversed. (:issue:`3249`, :issue:`423`,
    :issue:`4078`, :issue:`3756`, :pull:`4466`)
 
+#. `@wjbenfold`_ prevented an ``AttributeError`` being logged to ``stderr`` when
+   a :class:`~iris.fileformats.cf.CFReader` that fails to initialise is garbage
+   collected. (:issue:`3312`, :pull:`4646`)
+
 
 💣 Incompatible Changes
 =======================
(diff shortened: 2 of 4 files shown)


wjbenfold:
Closes #3312

lib/iris/fileformats/cf.py
```diff
@@ -1295,7 +1295,10 @@ def _reset(self):
 
     def __del__(self):
         # Explicitly close dataset to prevent file remaining open.
-        self._dataset.close()
+        try:
+            self._dataset.close()
+        except AttributeError:
```
pp-mo, line 1300: TBH I think you should ignore virtually **any** error here.
I can't think of anything it would be sensible to raise here, since there is no code context for what happened (just the garbage collector).

It's also fair to say that this belt+braces is a possibly bit odd, and it might be more logical to trust the dataset (or underlying file) to tidy up when _it_ is deferenced or garbage-collected.
Maybe there was at some time a problem with the netCDF4.Dataset that necessitated this, but I bet there isn't any more ?

evertrol: @pp-mo  That was initially my thought as well. See https://github.com/SciTools/iris/issues/3312#issuecomment-500452299 . See bjlittle's response below my comment, which is an explanation of why `__del__` is there.

pp-mo: Ok, but in that case we maybe need to be clear why we are swallowing some errors but not others.
It seems to be simple a bug in the CFReader -- it doesn't always have a `_dataset`.

So I think it would make far more sensible + robust to init `self._dataset = None` + test before closing.
[Like this]() ??

wjbenfold: That's what I had in [this commit](https://github.com/SciTools/iris/pull/4646/commits/0bf13f4825cbc69339c0c6c3ddc1fd15a3366288) and @evertrol had in their original PR (#3336) - I just thought it more pythonic to ask forgiveness rather than permission. I can put it back

evertrol: Then we're back at https://github.com/SciTools/iris/pull/3336 (which I closed for a different reason).

I'm still in favour of removing `__del__`, and let the garbage collector handling the destruction and closing of the dataset. If `netCDF4.Dataset` doesn't close its file pointer upon garbage collection/destruction, then that is the thing that should be fixed, imo.
Then again, I see no `__del__` method in `netCDF4.Dataset`, while `close()` does a lot of things there (there are also `__enter__` and `__exit__` methods, but these are not applicable here). But this related [netCDF4-python issue](https://github.com/Unidata/netcdf4-python/issues/218) does mention `__dealloc__` takes of the file closing. 
So once there are no references to the dataset, the underlying file get closed. Which suggests that letting Python garbage collect the CFReader instance, and thereby also its `_dataset` attribute, will close the file.

But someone else should decide; in particular as I haven't used Scitools for well over a year.

pp-mo (approved):
I'm good with this now.
I had doubts about whether the destructor would actually get called within a test evaluation, hence would the test ever fail.  But when tested against existing code, from the command-line, it does.  -- Though not if run from pytest.

PR #4647, opened 2022-03-18
Implement `iris.plot.fill_between`

### LINKED ISSUE #3493: Support `iris.plot.fill_between` and `iris.quickplot.fill_between`
It is fairly common to have `fill_between` plots as in [timeseries.pdf](https://github.com/SciTools/iris/files/3788002/timeseries.pdf) to show ranges or uncertainties.

It would be nice to support this in iris.

wjbenfold:
### Description

Closes #3493 by implementing `iris.plot.fill_between` and `iris.quickplot.fill_between`.

Notes for reviewer:

- I'm not proud of `_get_plot_objects` but not sure it's worth a general case version (maybe it is though?)
- I think these tests should give decent coverage, would accept additions though (and are they all sensible as they stand?)
- ~~Haven't added the image test results yet~~
- ~~Will add a what's new later~~

---


lib/iris/plot.py
```diff
@@ -645,7 +645,44 @@ def _u_object_from_v_object(v_object):
...
-    if len(args) > 1 and isinstance(
+    if len(args) > 2 and isinstance(
+        args[2], (iris.cube.Cube, iris.coords.Coord)
+    ):
+        # three arguments
+        u_object, v_object1, v_object2 = args[:3]
+        u1, v1 = _uv_from_u_object_v_object(u_object, v_object1)
+        u2, v2 = _uv_from_u_object_v_object(u_object, v_object2)
+        args = args[3:]
+        if not np.array_equal(u1, u2):
```
rcomer, line 656: Think this should be comparing `v1.size` and `v2.size`?  But I wonder if you can just roll these three checks into one with something like

```python
if u1.size != v1.size or u2.size != v2.size:
    msg = "The three objects are not compatible.  They should have equal sizes but got ({}: {}), ({}: {}) and ({}: {})"
    raise ValueError(
        msg.format(
            u_object.name(), u1.size, v_object1.name(), v1.size, v_object2.name(), v2.size))
```

wjbenfold: I think we also need a third check that u1 and u2 are the same size? Currently I am doing something odd in that I check them and then return v1 and v2, although those are likely to be the reason that u1 and u2 aren't the same. Maybe a threeway `or` then returning everything for the user to look at is the way forwards?

rcomer: Yes, sorry I missed that sometimes
```python
        u = np.arange(v.shape[0])
```

rcomer: Or may just do
```python
if u1.size != v1.size or u1.size != v2.size:
```
as those are the 3 arrays you end up using?

wjbenfold: Done

wjbenfold:
Images https://github.com/SciTools/iris-test-data/pull/78

bjlittle (approved):

PR #4649, opened 2022-03-21
Plotting circular data extends args as well as the data

### LINKED ISSUE #466: Distorted iplt.points with circular coordinate
`iplt.points()` distorts the output when longitude coordinate is circular.  The second plot below shows the correct behaviour by setting circular to false: `cube.coord('longitude').circular = False`

Also, the current API documentation suggests that `points()` requires a list of coords as arguments, but the current implementation expects a cube.

figure_1

```
import iris.tests as tests
import iris
import iris.plot as iplt

plt.switch_backend('tkagg')
pp_file = tests.get_data_path(('PP', 'globClim1', 'u_wind.pp'))
cube = iris.load(pp_file)[0][0]

plt.subplot(211)    
iplt.pcolormesh(cube, vmax=50)
iplt.points(cube, s=cube.data)
plt.gca().coastlines()
ax1 = plt.gca()

plt.subplot(212)
cube.coord('longitude').circular = False
iplt.pcolormesh(cube, vmax=50)
iplt.points(cube, s=cube.data)
plt.gca().coastlines()
ax2 = plt.gca()

plt.show()
```

wjbenfold:
### Description

Closes #466, which is caused by Iris extending cubes to handle the wrapping of circular coordinates when plotting but not doing the same for additional arguments to the plot.

Still to do:
* New test image
* What's new

---

### DIFF
--- a/lib/iris/plot.py
+++ b/lib/iris/plot.py
@@ -985,23 +985,25 @@ def _map_common(
         y = np.append(y, y[:, 0:1], axis=1)
         x = np.append(x, x[:, 0:1] + 360 * direction, axis=1)
         data = ma.concatenate([data, data[:, 0:1]], axis=1)
-        # if "_v_data" in kwargs:
-        #     v_data = kwargs["_v_data"]
-        #     v_data = ma.concatenate([v_data, v_data[:, 0:1]], axis=1)
-        #     kwargs["_v_data"] = v_data
 
+        # Having extended the data, we also need to extend extra kwargs for
+        # matplotlib (e.g. point colours)
         for key, val in kwargs.items():
             if (
                 isinstance(val, np.ndarray)
                 and val.ndim >= 2
                 and val.shape[1] == original_length
             ):
+                # Construct indices that take only the first column of the data,
+                # regardless of its current shape.
                 endslice = tuple(
                     [
                         slice(0, 1) if ii == 1 else slice(None)
                         for ii in range(val.ndim)
                     ]
                 )
+                # Concatenate the first column to the end of the data then
+                # update kwargs
                 val = ma.concatenate([val, val[endslice]], axis=1)
                 kwargs[key] = val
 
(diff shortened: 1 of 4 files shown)


lib/iris/plot.py
```diff
@@ -976,16 +976,32 @@ def _map_common(
...
-            v_data = ma.concatenate([v_data, v_data[:, 0:1]], axis=1)
-            kwargs["_v_data"] = v_data
+        # if "_v_data" in kwargs:
+        #     v_data = kwargs["_v_data"]
+        #     v_data = ma.concatenate([v_data, v_data[:, 0:1]], axis=1)
+        #     kwargs["_v_data"] = v_data
+
+        for key, val in kwargs.items():
+            if (
+                isinstance(val, np.ndarray)
```
jamesp, line 993: Does this have to handle the case of dask data or is it already realised at this point?

lib/iris/plot.py
```diff
@@ -976,16 +976,32 @@ def _map_common(
...
+                and val.ndim >= 2
+                and val.shape[1] == original_length
+            ):
+                endslice = tuple(
```
jamesp, line 997: This needs some comments, I'm not sure I understand why you're doing it!

jamesp (approved):
If it works and the tests pass I'm happy for this to go in.

@wjbenfold did you put the new test image into the test-data repo for this?

lib/iris/plot.py
wjbenfold: I don't think people should pass dask arrays in here, not sure if this a general enough check to catch masked data though? I think it should be.

To do: Write a test that we handle that

wjbenfold: I think I've changed my mind, a masked array is an instance of `ndarray` (I've checked the docs and tried it out). Could still do some testing of _map_common, but it's private so we don't tend to do that lots. What do you think @jamesp?

lib/iris/plot.py
wjbenfold: Comments added

wjbenfold:
https://github.com/SciTools/test-iris-imagehash/pull/62

wjbenfold:
Note from Jamie - check if the data has to be an array

Issue #4526, opened 2022-01-25
Converting a cube to a Pandas dataframe 


kaedonkers:
## ✨ Feature Request
Return a "long" DataFrame which retains all the metadata of the cube (either by default or using kwarg `table="long"`).

## Motivation
Currently `iris.pandas.as_data_frame` turns a 2D cube (a specification not present in the documentation) into a "pivot table"/"wide table" DataFrame with one of the former dim coord values as the column names and the other dim coord values as the indices. The values of the DataArray at the centre of the cube become the table values. 

I would argue that this result is unexpected for Pandas users, not particularly useful and loses lots of metadata in the process.

## Feature list
To keep track of ideas as I find them:
- [ ] Specify in the documentation that, currently, only a 2D cube can be converted to a DataFrame
- [ ] Remove dependence on `cube.coord(dimensions=[0/1])` as this throws unnecessary errors related to the presence of AuxCoords

## Proposed change
A better default behaviour would be to generate a "long" table in which all the coord values from the cube are in separate columns of the DataFrame, with the coord name as the name of the column. The DataArray values would be in another column named after the cube. Attributes could also be included as their own columns for maximum metadata retention, although this might want a toggle kwarg as it could clutter the resulting DataFrame. 

This would also allow the conversion to handle more than 2D data (which should really be added to the current documentation as a requirement).

For example:
- **Current situation**
```python
cube
>>> precipitation_amount / (kg m-2 day-1) (ensemble_member: 12; time: 7200)
        Dimension coordinates:
            ensemble_member                               x         -
            time                                          -         x
        Cell methods:
            mean                          time
        Attributes:
            Conventions                   CF-1.7

iris.pandas.as_data_frame(cube)
>>>     1980-12-01 12:00:00  1980-12-02 12:00:00  1980-12-03 12:00:00  ...  2000-11-30 12:00:00
    1              1.707561             0.000042             0.003254  ...             0.000034
    4              0.000037             0.001898             0.002343  ...             0.052522
    5              2.079478             0.080967             0.156937  ...             0.051903
    6             16.845169             3.951329             0.034234  ...             0.000010
    7              1.636697             0.000034             0.000013  ...             0.000002
    8              4.940392            15.944072             7.649919  ...             0.001480
    9              0.000037             0.458525             0.122213  ...             1.298610
    10            11.951665             0.038251             3.440227  ...             0.000024
    11             0.001125             0.000270             0.000000  ...             0.007789
[...]

PR #4669, opened 2022-03-29
Improve iris.pandas cube -> data.frame

hsteptoe:
### Description
Improve the conversion of a Cube to a Pandas data.frame. Aims to address #4526

Aims to deal with:

- [x] An arbitrary number of dimensions -> [d8c5b40](https://github.com/SciTools/iris/pull/4669/commits/d8c5b4060bac16642f3f60feb8d5002a1a0a5caa)
- [x] Missing cube dimensions -> [a46c08a](https://github.com/SciTools/iris/pull/4669/commits/a46c08a9e9db883108b0cf08a3190e55886fb2d0)
- [x] Add auxiliary coordinate information to dataframe -> [1df3072](https://github.com/SciTools/iris/pull/4669/commits/1df3072ccb5d3b281ca35f2791bd8a29d521f687)
- [x] Add global attribute data to dataframe -> [2c2ce69](https://github.com/SciTools/iris/pull/4669/commits/2c2ce69f11563f245ab14fe3b82914bee9f2fac7)
- [x] Deal with `copy` issue
- [x] Update doc strings
- [x] Update tests (but not unit & integration testing)
- [x] Update Whats New
- [x] Check alignment with @trexfeathers  #4890

### Major proposed (possibly breaking) changes

- Having functions to convert to a Series and a DataFrame seems superfluous. A Series is arguably a special case of a DataFrame, so I propose having **only** a function to do the DataFrame conversion, and then let folk further convert to a Series if they want to via standard `pandas` functions. This is done in [61b801a](https://github.com/SciTools/iris/pull/4669/commits/61b801aba20130c84782d938700fb99e10fc492d).
-~~I don't think the `copy` argument is valid any more. Working with 'long' format DataFrames there is no continuity between the cube.data array and the DataFrame in memory. `copy=True` is the default, but there is not option for `copy=False` any more. This is done in [071d07f](https://github.com/SciTools/iris/pull/4669/commits/071d07f699fcf7da10c283b42335a6e07f6f7bb9).~~ Copy ability retained.

**WORK IN PROGRESS**


hsteptoe:
Good discussion with @trexfeathers about our joint work on Pandas-Iris bridging. We agreed that he would focus (and separately PR) on the DataFrame -> Cube, so this PR will only focus on Cube -> DataFrame

lib/iris/pandas.py
```diff
@@ -192,39 +134,42 @@ def as_data_frame(cube, copy=True):
 
     Kwargs:
 
-        * copy - Whether to make a copy of the data.
-                 Defaults to True. Must be True for masked data
-                 and some data types (see notes below).
```
trexfeathers, line 197: We can potentially bring this back.

lib/iris/pandas.py
```diff
@@ -192,39 +134,42 @@ def as_data_frame(cube, copy=True):
...
+                    coords[i] = _as_pandas_coord(c)
+                    coord_names[i] = c.name()
+                else:
+                    pass
+    else:
+        coord_names = list(map(lambda x: x.name(), cube.dim_coords))
+        coords = list(map(lambda x: _as_pandas_coord(x), cube.dim_coords))
+
+    index = pandas.MultiIndex.from_product(coords, names=coord_names)
+    data_frame = pandas.DataFrame({cube.name(): data.flatten()}, index=index)
```
trexfeathers, line 168:
```suggestion
    data_frame = pandas.DataFrame(data.ravel(), columns=[cube.name()], index=index)
```

This allows `data_frame` to directly use `cube.data` i.e. `copy=False`.

lib/iris/pandas.py
```diff
@@ -192,39 +134,42 @@ def as_data_frame(cube, copy=True):
...
 
-    data_frame = pandas.DataFrame(data, index, columns)
-    if not copy:
-        _assert_shared(data, data_frame)
+    if dropna:
+        data_frame.dropna(inplace=True)
```
trexfeathers, line 171: This line ALWAYS copies the `data_frame` values i.e. `copy=True`. If you like my suggestion for the `pandas.DataFrame(...` line, then the user has the option to avoid copying data by setting `dropna=False`, if that's important to them.

We'd need a docstring caveat explaining this, similar to the original one against the `copy` argument.

lib/iris/pandas.py
```diff
@@ -124,65 +124,7 @@ def _as_pandas_coord(coord):
...
-
-    series = pandas.Series(data, index)
-    if not copy:
-        _assert_shared(data, series)
-
-    return series
-
-
-def as_data_frame(cube, copy=True):
+def as_data_frame(cube, dropna=True, asmultiindex=False):
```
trexfeathers, line 127:
```suggestion
def as_data_frame(cube, dropna=False, asmultiindex=False):
```

Might be better, depending on what you think of my other comments in this review. Although I'm still not certain.

trexfeathers (commented):
Having learnt more about Pandas' copy behaviour, I think it will be possible to retain this modules' original `copy` option. It certainly has some value in cases where a large in-memory `Cube` is being converted, since copying will double the memory demand and could even blow the machine's memory. So if there is a prospect of avoiding copying I reckon it's worth exploring.

See my comments below.

lib/iris/pandas.py
```diff
@@ -167,6 +167,19 @@ def as_data_frame(cube, dropna=True, asmultiindex=False):
     index = pandas.MultiIndex.from_product(coords, names=coord_names)
     data_frame = pandas.DataFrame({cube.name(): data.flatten()}, index=index)
 
+    # Add aux coord information
+    if add_aux_coord:
+        aux_coord_names = list(map(lambda x: x.name(), cube.aux_coords))
+        for acoord in add_aux_coord:
+            assert acoord in aux_coord_names, f'\"{acoord}\" not in cube' # Check aux coord exists
```
hsteptoe: Are `assert` statements too harsh?  Should an `AssertionError` be raised if `as_data_frame` can't find user specified `AuxCoords` or attributes in the cube, or pass silently and return a `DataFrame` but without the missing coords/attributes?

lib/iris/pandas.py
```diff
@@ -180,6 +180,13 @@ def as_data_frame(cube, dropna=True, asmultiindex=False, add_aux_coord=None):
             # Join to main data frame
             data_frame = data_frame.join(acoord_df, on=np.array(coord_names)[coord_bool][0])
 
+    # Add data from global attributes
+    if add_global_attributes:
+        global_attribute_names = list(rcp26.attributes.keys())
+        for global_attribute in add_global_attributes:
+            assert global_attribute in global_attribute_names, f'\"{global_attribute}\" not in cube' # Check global attribute exists
```
hsteptoe: As before, are `assert` statements too harsh here?

lib/iris/pandas.py
trexfeathers: Last week we discussed providing the basics but leaving it to the user to handle the more nuanced bits. In that vein I think the best course would be to add ALL the `AuxCoord`s - the vast majority of use cases would presumably be covered by either all or none, and those few who want some-not-all can easily remove unwanted columns downstream.

trexfeathers: Also I read last year that `assert` is only intended for things that you ALWAYS expect to be true - they're primarily a developer convenience. Anything that's there to inform the user should be in an `if:` block, raising the most appropriate `Error` class for that case.

lib/iris/pandas.py
hsteptoe: OK, I'm inclined to remove this as an option then to keep with the 'basics' principal.

lib/iris/pandas.py
hsteptoe: Agreed. Back in [332e260](https://github.com/SciTools/iris/pull/4669/commits/332e2606bbdfa1eed9e853f9028f94223483bc18)

lib/iris/pandas.py
```diff
@@ -124,107 +124,100 @@ def _as_pandas_coord(coord):
...
-
-    .. note::
+    # Extract dim coord information
+    if cube.ndim != len(cube.dim_coords):
+        # Create dummy dim coord information if dim coords not defined
+        coord_names = ["dim" + str(n) for n in range(cube.ndim)]
+        coords = [range(dim) for dim in cube.shape]
+        for onecoord in cube.dim_coords:
+            for i, dummyc in enumerate(coords):
+                if len(dummyc) == len(onecoord.points):
```
stephenworsley, line 171: Wouldn't this logic cause confusion when a cube has two DimCoords of the same length?

hsteptoe: Yes, although I originally thought this would be a problem, as the mapping would be the same (wrongly).  So, is there a better way to establish which DimCoord the AuxCoord maps to?

stephenworsley: I imagine you might want to identify each dim_coord by its dimension. You could try something like:
```
coord_lists = [cube.coords(dimensions=n, dim_coords=True) for n in range(len(cube.ndims))]
coords = [list[0] for list in coord_lists if len(list)>0 else None for list in coord_lists]
```
which should give you a list of DimCoords indexed by their associated dimension.

hsteptoe: I think [8301caa](https://github.com/SciTools/iris/pull/4669/commits/8301caaedd6b48470136effb708d96e60c93960b) now has better logic

hsteptoe:
Reference @trexfeathers alignment issues https://github.com/SciTools/iris/pull/4890#issuecomment-1217820004

bjlittle:
@hsteptoe Fancy doing a rebase to resolve these conflicts? :+1:

hsteptoe:
ci-tests/doctest and ci-test/linkcheck errors aren't especially obvious to me...

hsteptoe:
@trexfeathers Any chance this might make it into [v3.4.0](https://github.com/orgs/SciTools/projects/3) #4433

lib/iris/pandas.py
```diff
@@ -518,88 +518,298 @@ def _get_base(array):
...
+    419899    298.779938
+    419900    298.913147
+    419901           NaN
+    419902           NaN
+    419903    298.995148
+    Name: surface_temperature, Length: 419904, dtype: float32
 
     """
-    data = cube.data
+    # Checks
```
trexfeathers, line 726: ## Using `FUTURE`

Unfortunately, since this a behaviour change being released in a minor release, we need to preserve the old behaviour and provide the new behaviour as an opt-in.

### The good news

I don't expect any complications with input nor output as you haven't deleted any parameters, and both old and new behaviours return a `DataFrame`.

### Update your branch

`iris.FUTURE.pandas_ndim` has just been merged into the `pandas_ndim` feature branch, so you'll need to bring that into your branch.

### Code changes

Just wrap everything like this:

```python
if iris.FUTURE.pandas_ndim:
    # New behaviour.
else:
    # Original behaviour.
```

### Warnings

The user should get a notification that the behaviour will change in future. Here is an existing example for a different future flag:

https://github.com/SciTools/iris/blob/d56824cbc892356302c135ee23d704ff01e12d71/lib/iris/fileformats/_nc_load_rules/helpers.py#L272-L280

The warning will need a test, too.

### Docstring

I think it's good to align the docstring with the new behaviour, as you have done 🙂. But we should definitely include a note about the `FUTURE` flag and the opt-in nature of the new behaviour. We could also encourage the reader to visit older copies of the documentation if they need to read more about the original behaviour.

The best section for this is probably the `Warns` section - [see the NumPy docs for detail](https://numpydoc.readthedocs.io/en/latest/format.html#warns).

### Testing

The existing tests should remain unchanged. I recommend creating a new class to contain the tests you have created/modified. The class can include a setup routine which switches the `FUTURE` flag ([here's a quick summary on Stack Overflow of different ways to achieve](https://stackoverflow.com/a/26414138)).

lib/iris/pandas.py
```diff
@@ -518,88 +518,298 @@ def _get_base(array):
...
+    # Make base DataFrame
+    index = pandas.MultiIndex.from_product(coords, names=coord_names)
+    data_frame = pandas.DataFrame(
+        data.ravel(), columns=[cube.name()], index=index
+    )
 
-    data_frame = pandas.DataFrame(data, index, columns)
-    if not copy:
-        _assert_shared(data, data_frame)
+    # Add AuxCoord & scalar coordinate information
```
trexfeathers, line 764: To align with the features of `iris.pandas.as_cubes()`: can the code for `AuxCoord` columns be generalised to do `CellMeasure`s and `AncillaryVariable`s, too? These all come from the same parent class so I would hope there would only be a few pieces to change.

trexfeathers (changes requested):
Thanks so much @hsteptoe! Don't let my many comments fool you, I think this is a solid piece of work at the core 💐.

#### Notes that don't fit in individual comments:

- The examples are fantastic ❤.
- Nice job with the tests.
- My docstring and What's New suggestions ought to fix the failing documentation CI.
- If you think you'll struggle to find the time, we can always merge into the feature branch with some outstanding commends and the core Iris team can pick up the baton.

hsteptoe:
> > @trexfeathers Any chance this might make it into [v3.4.0](https://github.com/orgs/SciTools/projects/3) #4433 ?
> 
> @hsteptoe I'd certainly like for this to happen. Hopefully my review today gives enough warning for you to action the various comments, and therefore not need much more time from the core Iris team (since we are each assigned to other things too!).

@trexfeathers Thanks for the review! I'll look at actioning it this week...

lib/iris/pandas.py
```diff
@@ -518,88 +518,298 @@ def _get_base(array):
...
-                 and some data types (see notes below).
+    Parameters
+    ----------
+    :class:`~iris.cube.Cube`
+        The :class:`Cube` to be converted to a Pandas `DataFrame`.
+    copy : bool, default=True
+        Whether the Pandas `DataFrame` is a copy of the the Cube :attr:`~iris.cube.Cube.data`.
+        This option is provided to help with memory size concerns.
+    add_aux_coord : bool, default=False
+        If True, add all :attr:`~iris.cube.Cube.aux_coords` to add to the returned `DataFrame`.
```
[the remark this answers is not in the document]
hsteptoe: Mainly for speed of execution.  But happy to default to `True` with an added note about trying `False` if conversion seems slow.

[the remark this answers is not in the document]
trexfeathers: Ah, good point, I hadn't considered performance. Happy to leave as-is.

If you action https://github.com/SciTools/iris/pull/4669#discussion_r1003373284, it would be neatest to have everything covered by a single parameter.

docs/src/whatsnew/latest.rst
```diff
@@ -65,7 +71,10 @@ This document explains the changes made to Iris for this release
 🔥 Deprecations
 ===============
 
-#. N/A
+#. `@hsteptoe`_ and `@trexfeathers`_ (reviewer) deprecated
```
hsteptoe, line 74: @trexfeathers  Just checking... does this cover your comment [above](https://github.com/SciTools/iris/pull/4669/files/bff3e3afaede61806cb99f65359c4cf4e49ff90c#r1003335525)

lib/iris/pandas.py
```diff
@@ -518,88 +518,298 @@ def _get_base(array):
...
+    # Checks
+    if not isinstance(cube, iris.cube.Cube):
+        raise ValueError("Input must be an iris.cube.Cube instance")
+    if add_global_attributes:
+        global_attribute_names = list(cube.attributes.keys())
+        for global_attribute in add_global_attributes:
+            if (
+                global_attribute not in global_attribute_names
+            ):  # Check global attribute exists
+                raise ValueError(
```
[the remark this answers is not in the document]
hsteptoe: Code removed as [above](https://github.com/SciTools/iris/pull/4669#discussion_r1003237614)

lib/iris/pandas.py
```diff
@@ -518,88 +518,298 @@ def _get_base(array):
...
+
+        # Add scalar coordinate information
+        scalar_coord_list = cube.coords(dimensions=(), dim_coords=False)
+        for scalar_coord in scalar_coord_list:
+            data_frame[scalar_coord.name()] = scalar_coord.points.squeeze()
+
+    # Add global attribute information
+    if add_global_attributes:
+        if isinstance(
+            cube.attributes[global_attribute],
```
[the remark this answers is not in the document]
hsteptoe: Code removed as [above](https://github.com/SciTools/iris/pull/4669#discussion_r1003237614)

lib/iris/tests/test_pandas.py
```diff
@@ -249,76 +213,138 @@ def test_time_360(self):
...
-        assert cube.data[0, 0] == 99
+        self.assertArrayEqual(expected_attribute, data_frame.sheep)
 
-    def test_copy_int64_false(self):
-        cube = Cube(
-            np.array([[0, 1, 2, 3, 4], [5, 6, 7, 8, 9]], dtype=np.int64),
-            long_name="foo",
+    def test_attribute_error(self):
+        cube = Cube(np.array([0, 1, 2, 3, 4]), long_name="foo")
+        self.assertRaises(
```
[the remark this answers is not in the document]
hsteptoe: Code removed along with add_global_attributes as [above](https://github.com/SciTools/iris/pull/4669#discussion_r1003237614)

lib/iris/pandas.py
```diff
@@ -518,88 +518,298 @@ def _get_base(array):
...
     """
-    data = cube.data
+    # Checks
+    if not isinstance(cube, iris.cube.Cube):
+        raise ValueError("Input must be an iris.cube.Cube instance")
+    if add_global_attributes:
+        global_attribute_names = list(cube.attributes.keys())
+        for global_attribute in add_global_attributes:
+            if (
+                global_attribute not in global_attribute_names
```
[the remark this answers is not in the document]
hsteptoe: Code removed as [above](https://github.com/SciTools/iris/pull/4669#discussion_r1003237614)

lib/iris/pandas.py
```diff
@@ -518,88 +518,298 @@ def _get_base(array):
...
+                        [n, coord]
+                        for coord in cube.coords(
+                            dimensions=n, dim_coords=False
+                        )
+                    ]
+                    for n in _get_dim_combinations(cube.ndim)
+                ]
+            )
+        )
+        for aux_coord_index, aux_coord in aux_coord_list:
```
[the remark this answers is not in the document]
hsteptoe: Now split out the list comprehension into its own function.

lib/iris/pandas.py
trexfeathers: The What's New entry would also need minor changes once this is actioned.

docs/src/whatsnew/latest.rst
trexfeathers: All good 👍

lib/iris/pandas.py
hsteptoe: Done, but now added additional kwargs for syntactic consistency with `as_cubes`

lib/iris/pandas.py
```diff
@@ -518,88 +518,298 @@ def _get_base(array):
...
+        ):
+            data_frame[global_attribute] = str(
+                cube.attributes[global_attribute]
+            )
+        else:
+            data_frame[global_attribute] = cube.attributes[global_attribute]
 
-    return data_frame
+    if copy:
+        return data_frame.reorder_levels(coord_names).sort_index()
```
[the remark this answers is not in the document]
hsteptoe: This was because the `pandas.merge` functions that underpins much of the dimension mapping of auxiliary data resorts the MultiIndex.  This is to make sure that the final data_frame index ordering matches the that of the cubes.

lib/iris/pandas.py
```diff
@@ -518,88 +518,366 @@ def _get_base(array):
...
+    copy=True,
+    add_aux_coord=False,
+    add_cell_measures=False,
+    add_ancillary_variables=None,
+):
     """
-    Convert a 2D cube to a Pandas DataFrame.
-
-    Args:
+    :attr:`~iris.cube.Cube.dim_coords` and :attr:`~iris.cube.Cube.data` are flattened into a long-style
```
trexfeathers, line 632: Could you reduce some of the docstring line lengths? We try and stay within `80`, with some leeway when necessary (`~88`).

### Exceptions

- The one-line descriptor at the top of the docstring.
- Examples (which would render terribly otherwise).

trexfeathers (changes requested):
Thanks @hsteptoe! Looking good.

A few more for you...

lib/iris/pandas.py
hsteptoe: We agreed that @trexfeathers would action this once merged into the feature branch

lib/iris/pandas.py
```diff
@@ -518,88 +518,366 @@ def _get_base(array):
...
+                    data_frame, acoord_df, left_index=True, right_index=True
+                )
+
+        # Add scalar coordinate information
+        for scalar_coord in cube.coords(dimensions=(), dim_coords=False):
+            data_frame[scalar_coord.name()] = _as_pandas_coord(
+                scalar_coord
+            ).squeeze()
+
+    if add_ancillary_variables:
```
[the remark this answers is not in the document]
hsteptoe: We agreed that would make `_make_list_*` more consistent in the way that they access coordinate info. Add `TODO` comment in relevant function for future refactoring.

lib/iris/pandas.py
```diff
@@ -518,88 +517,330 @@ def _get_base(array):
...
+                                  357.500000                  NaN
+                                  358.333313                  NaN
+                                  359.166656           298.995148
+
+    [419904 rows x 1 columns]
+
+    Using `add_aux_coord=True` maps :class:`~iris.coords.AuxCoord` and scalar
+    coordinate information to the :class:`~pandas.DataFrame`:
+
+    >>> df = ipd.as_data_frame(cube, add_aux_coord=True)
```
trexfeathers, line 707:
```suggestion
    Using `add_aux_coords=True` maps :class:`~iris.coords.AuxCoord` and scalar
    coordinate information to the :class:`~pandas.DataFrame`:

    >>> df = ipd.as_data_frame(cube, add_aux_coords=True)
```

lib/iris/pandas.py
```diff
@@ -518,88 +517,330 @@ def _get_base(array):
...
-    .. note::
+    Warnings
+    --------
+    Where the :class:`~iris.cube.Cube` contains masked values, these become
+    :data:`numpy.nan` in the returned :class:`~pandas.DataFrame`.
 
-        Pandas will sometimes make a copy of the array,
-        for example when creating from an int32 array.
-        Iris will detect this and raise an exception if copy=False.
+    Examples
```
trexfeathers, line 677: The remaining doctest failures are caused by a few different problems in this section. Basically the code provided below doesn't produce the expected results. It's quite difficult for me to debug this as I don't know how you produced the outputs originally, so it would be great if you could. 

### To debug

```bash
$ cd <your iris directory>
$ conda create -n iris-dev --file requirements/ci/nox.lock/py310-linux-64.lock
$ conda activate iris-dev
$ export PYTHONPATH=$PYTHONPATH:$PWD/lib
$ cd docs/src
$ make doctest

```

### Blank lines

If you get any failures because doctest is expecting `<BLANKLINE>`: just add in that exact text where it is expected.

trexfeathers (changes requested):
Thanks @hsteptoe! We're basically there, just two final comments:

- https://github.com/SciTools/iris/pull/4669#discussion_r1012094268
-

docs/src/whatsnew/latest.rst
```diff
@@ -35,6 +35,12 @@ This document explains the changes made to Iris for this release
    non-existing paths, and added expansion functionality to :func:`~iris.io.save`.
    (:issue:`4772`, :pull:`4913`)
 
+#. `@hsteptoe`_ and `@trexfeathers`_ (reviewer) added :func:`iris.pandas.as_data_frame`,
+   which provides improved conversion of :class:`~iris.cube.Cube`\s to 
+   :class:`~pandas.DataFrame`\s. This includes better handling of multiple 
+   :class:`~iris.cube.Cube`dimensions, auxiliary coordinates and attribute information.
+   (:issue:`4526`, :pull:`4669`)
```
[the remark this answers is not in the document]
hsteptoe: [213afb6](https://github.com/SciTools/iris/pull/4669/commits/213afb6fce333ea8d9b92a89a6f2c622bc6c74c3) and [d39b052](https://github.com/SciTools/iris/pull/4669/commits/d39b052aa3b72baa4f959866b538b6c89a93a1fe) were my attempt to fix this. Re-typing didn't really help as git didn't recognise that anything had changed, hence faffing with remove-commit-re-add-commit.

docs/src/whatsnew/latest.rst
```diff
@@ -35,6 +35,12 @@ This document explains the changes made to Iris for this release
    non-existing paths, and added expansion functionality to :func:`~iris.io.save`.
    (:issue:`4772`, :pull:`4913`)
 
+#. `@hsteptoe`_ and `@trexfeathers`_ (reviewer) added :func:`iris.pandas.as_data_frame`,
+   which provides improved conversion of :class:`~iris.cube.Cube`\s to 
+   :class:`~pandas.DataFrame`\s. This includes better handling of multiple 
+   :class:`~iris.cube.Cube`dimensions, auxiliary coordinates and attribute information.
```
trexfeathers, line 41:
```suggestion
   :class:`~iris.cube.Cube` dimensions, auxiliary coordinates and attribute information.
```

Thanks @ESadek-MO!

docs/src/whatsnew/latest.rst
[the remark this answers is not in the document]
trexfeathers: https://github.com/SciTools/iris/pull/4669#discussion_r1014246563

lib/iris/pandas.py
hsteptoe: Turns out that once I'd fixed the obvious typos and `<BLANKLINE>` issues, `black` was still trolling me... `pandas` printing of `DataFrames` included trailing whitespace which doctest was trying to match, but which kept being removed by `black`.   Fixed with `# doctest: +NORMALIZE_WHITESPACE`

trexfeathers: That's a useful trick to know about!

trexfeathers:
Well I'm perplexed. They all passed for me locally!

hsteptoe:
> Well I'm perplexed. They all passed for me locally!

And for me... but it looked like there was some difference in the number of cols being printed in the github tests vs my local tests... now checked the output and matched to github test version...

trexfeathers:
> > Well I'm perplexed. They all passed for me locally!
> 
> And for me... but it looked like there was some difference in the number of cols being printed in the github tests vs my local tests... now checked the output and matched to github test version...

I'm guessing Pandas is sensitive to available horizontal space when running. Rather frustrating as this gives us a situation where local behaves differently to CI. I won't let this get in the way of merging and I'll discuss with the team in future.

hsteptoe:
> 

Possibly some `pandas` options that we could explore: https://stackoverflow.com/a/11711637

hsteptoe:
> > > Well I'm perplexed. They all passed for me locally!
> > 
> > 
> > And for me... but it looked like there was some difference in the number of cols being printed in the github tests vs my local tests... now checked the output and matched to github test version...
> 
> I'm guessing Pandas is sensitive to available horizontal space when running. Rather frustrating as this gives us a situation where local behaves differently to CI. I won't let this get in the way of merging and I'll discuss with the team in future.

Possibly some `pandas` options that we could explore: https://stackoverflow.com/a/11711637

trexfeathers (approved):

trexfeathers:
Super! Thanks @hsteptoe. And we made it in under 100 commits 😂

hsteptoe:
> Super! Thanks @hsteptoe. And we made it in under 100 commits 😂

🎉 @trexfeathers Thanks for all your help!

lib/iris/pandas.py
trexfeathers: #5059

kaedonkers:
Awesome, thanks gents!

PR #4676, opened 2022-04-01
Add `iris.analysis.MAX_RUN` aggregator

wjbenfold:
### Description

Multiple users have raised that they would like to be able to calculate run statistics on cubes. I have implemented a maximum run length aggregator to address this need. This could be extended in future to other run statistics (e.g. average run length).

In order to make the use of this aggregator flexible, the condition for counting a cell in the run is provided as a callable argument that returns a boolean. Doing this does mean that it isn't possible for the aggregator to write the cell method and standard name in a CF-compliant fashion.

Another approach could be to constrain the possible conditions and thus output a CF-compliant result - would that be better?

---

### DIFF
--- a/lib/iris/analysis/__init__.py
+++ b/lib/iris/analysis/__init__.py
@@ -1043,9 +1043,10 @@ class Aggregator(_Aggregator):
             coord_names.append(coord.name())
 
         # Add a cell method.
-        method_name = self.cell_method.format(**kwargs)
-        cell_method = iris.coords.CellMethod(method_name, coord_names)
-        cube.add_cell_method(cell_method)
+        if self.cell_method is not None:
+            method_name = self.cell_method.format(**kwargs)
+            cell_method = iris.coords.CellMethod(method_name, coord_names)
+            cube.add_cell_method(cell_method)
 
 
 class WeightedAggregator(Aggregator):
@@ -1472,8 +1473,35 @@ def _proportion(array, function, axis, **kwargs):
     return result
 
 
-@_build_dask_mdtol_function
-def _lazy_max_run(array, axis=None, **kwargs):
+def _lazy_max_run(array, axis=-1, **kwargs):
+    """
+    Runs fairly clearly need the axis along which they're being calculated not
+    to be chunked. This wrapper uses
+    :meth:`~iris._lazy_data.map_complete_blocks` to guarantee that.
+    """
+
+    _partial_max_run_calc = functools.partial(
+        _lazy_max_run_calc,
+        axis=axis,
+        **kwargs,
+    )
+
+    result = iris._lazy_data.map_complete_blocks(
+        array, _partial_max_run_calc, (axis,), ()
+    )
+
+    # Check whether to reduce to a scalar result, as per the behaviour
+    # of other aggregators.
+    if result.shape == (1,):
+        result = da.squeeze(result)
+
+    return result
+
+
+def _lazy_max_run_calc(array, axis=-1, **kwargs):
+    """
+    Lazily perform the calculation of maximum run lengths along the given axis
+    """
     array = iris._lazy_data.as_lazy_data(array)
     func = kwargs.pop("function", None)
     if not callable(func):
@@ -1499,7 +1527,7 @@ def _lazy_max_run(array, axis=None, **kwargs):
         method="sequential",
         preop=None,
     )
-    run_lengths = da.diff(stepped_run_lengths)
+    run_lengths = da.diff(stepped_run_lengths, axis=axis)
     return da.max(run_lengths, axis=axis)
 
 
@@ -1694,10 +1722,10 @@ This aggregator handles masked data.
 
 
 MAX_RUN = Aggregator(
-    "max_run",
+    None,
     iris._lazy_data.non_lazy(_lazy_max_run),
     units_func=lambda units: 1,
-    lazy_func=_lazy_max_run,
+    lazy_func=_build_dask_mdtol_function(_lazy_max_run),
 )
 """
 An :class:`~iris.analysis.Aggregator` instance that finds the longest run of
@@ -1721,6 +1749,7 @@ each grid location could be calculated with::
 This aggregator handles masked data.
 
 """
+MAX_RUN.name = lambda: "max_run"
 
 
 GMEAN = Aggregator("geometric_mean", scipy.stats.mstats.gmean)
--- a/lib/iris/analysis/__init__.py
+++ b/lib/iris/analysis/__init__.py
@@ -1479,9 +1479,13 @@ def _lazy_max_run(array, axis=None, **kwargs):
     if not callable(func):
         emsg = "function must be a callable. Got {}."
         raise TypeError(emsg.format(type(func)))
+    bool_array = func(array)
+    bool_array = np.logical_and(
+        bool_array, np.logical_not(da.ma.getmaskarray(array))
+    )
     padding = [(0, 0)] * array.ndim
     padding[axis] = (0, 1)
-    ones_zeros = da.pad(func(array), padding).astype(int)
+    ones_zeros = da.pad(bool_array, padding).astype(int)
     cum_sum = da.cumsum(ones_zeros, axis=axis)
     run_totals = da.where(ones_zeros == 0, cum_sum, 0)
     stepped_run_lengths = da.reductions.cumreduction(
--- a/lib/iris/analysis/__init__.py
+++ b/lib/iris/analysis/__init__.py
@@ -1472,6 +1472,33 @@ def _proportion(array, function, axis, **kwargs):
     return result
 
 
+@_build_dask_mdtol_function
+def _lazy_max_run(array, axis=None, **kwargs):
+    array = iris._lazy_data.as_lazy_data(array)
+    func = kwargs.pop("function", None)
+    if not callable(func):
+        emsg = "function must be a callable. Got {}."
+        raise TypeError(emsg.format(type(func)))
+    padding = [(0, 0)] * array.ndim
+    padding[axis] = (0, 1)
+    ones_zeros = da.pad(func(array), padding).astype(int)
+    cum_sum = da.cumsum(ones_zeros, axis=axis)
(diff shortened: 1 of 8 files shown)


lib/iris/analysis/__init__.py
```diff
@@ -1662,6 +1721,37 @@ def interp_order(length):
 """
 
 
+MAX_RUN = Aggregator(
+    None,
+    iris._lazy_data.non_lazy(_lazy_max_run),
+    units_func=lambda units: 1,
```
stephenworsley, line 1727: Do you think there's a case that the units here ought to be the same as the coordinate you are collapsing over? If, as in the example, you are calculating the number of consecutive days that a condition holds, it seems like the correct unit ought to be in days.

I imagine that the such behaviour would be complicated to implement, depending on if the coord had bounds or if all the bounds were of uniform length. In the case where the bounds are not uniform length, this complicates what the correct mathematics ought to be. Given this complexity, I can see the appeal of a simpler more robust approach as taken here. Indeed, it seems as though the percintile aggregators also take a simpler approach which doesn't take into account unequal bound width in calculation. There may in fact be cases where you'd want to treat cells with different bound widths with the same weight. In this case a dimensionless unit may actually be the most appropriate one. It's probably worth thinking about and maybe clarifying in the documentation either way.

wjbenfold: I went with this based on it seeming like a form of count (which also uses units of 1). I can see the argument though. Given that the output of this as it stands isn't CF compliant I'm not sure there's anything to learn from there.

https://github.com/SciTools/iris/blob/6b273898c3889fd6897351b4c6d4d60660a9397d/lib/iris/analysis/__init__.py#L1692-L1697

stephenworsley: Yeah, this makes sense and seems consistent with what we have currently. I suppose the use of "data occurences" in each of the aggregator docstrings is enough to clarify what is actually being calculated.

lib/iris/analysis/__init__.py
```diff
@@ -1472,6 +1473,64 @@ def _proportion(array, function, axis, **kwargs):
     return result
 
 
+def _lazy_max_run(array, axis=-1, **kwargs):
```
stephenworsley, line 1476: I think the dask cumulative reductions are actually able to handle data being chunked along the axis being acted on. In fact I think all operations in `_lazy_max_run_calc` ought not to require `map_complete_blocks`. It's still probably worth doing the `da.squeeze` operation here mind you.

lib/iris/tests/unit/analysis/test_MAX_RUN.py
```diff
@@ -0,0 +1,300 @@
...
+        data = as_lazy_data(self.data_2d_axis1)
+        result = self.run_func(
+            data,
+            axis=1,
+            function=bool_func,
+        )
+        self.check_array(result, self.expected_2d_axis1)
+
+
+class TestLazyChunked(UnmaskedTest, LazyMixin):
```
stephenworsley, line 209: It might be worthwhile checking that the chunks in the returned array are as expected.

lib/iris/analysis/__init__.py
```diff
@@ -1662,6 +1721,37 @@ def interp_order(length):
...
+
+**For example**:
+
+The longest run of days with precipitation exceeding 10 (in cube data units) at
+each grid location could be calculated with::
+
+    result = precip_cube.collapsed('time', iris.analysis.MAX_RUN,
+                                   function=lambda values: values > 10)
+
+This aggregator handles masked data.
```
stephenworsley, line 1749: How does it treat masked data? Specifically, what happens when a masked point interupts a run?

lib/iris/tests/unit/analysis/test_MAX_RUN.py
```diff
@@ -0,0 +1,300 @@
...
+                    np.array([1, 1, 1, 3, 1, 3]), np.array([1, 1, 1, 0, 0, 0])
+                ),
+                1,
+            ),  # Mask max run
+            (
+                ma.masked_array(
+                    np.array([1, 1, 1, 3, 1, 3]), np.array([0, 0, 1, 0, 0, 0])
+                ),
+                2,
+            ),  # Partially mask run
```
stephenworsley, line 90: What happens when a mask interupts the middle of a run?
e.g.
```
ma.masked_array(
                    np.array([1, 1, 1, 3, 1, 3]), np.array([0, 1, 0, 0, 0, 0])
                )
```

stephenworsley (changes requested):
Looking good, just a few things to consider.

lib/iris/tests/unit/analysis/test_MAX_RUN.py
wjbenfold: I would expect a masked point to end a run, which is the current behaviour (though I'll put in a test for that). Do you think that should be the behaviour?

stephenworsley: I'm not sure masked points do end a run, at least not with the tests I've tried.

lib/iris/analysis/__init__.py
```diff
@@ -1472,6 +1473,64 @@ def _proportion(array, function, axis, **kwargs):
...
+    if not callable(func):
+        emsg = "function must be a callable. Got {}."
+        raise TypeError(emsg.format(type(func)))
+    bool_array = func(array)
+    bool_array = np.logical_and(
+        bool_array, np.logical_not(da.ma.getmaskarray(array))
+    )
+    padding = [(0, 0)] * array.ndim
+    padding[axis] = (0, 1)
+    ones_zeros = da.pad(bool_array, padding).astype(int)
```
stephenworsley, line 1516: In order to get masked points to end a run, I think you'd have to put something like this here:
`ones_zeros = da.ma.filled(ones_zeros, 0)`

lib/iris/tests/unit/analysis/test_MAX_RUN.py
wjbenfold: I concur, will fix

lib/iris/tests/unit/analysis/test_MAX_RUN.py
```diff
@@ -0,0 +1,300 @@
...
+            axis=1,
+            function=bool_func,
+        )
+        self.check_array(result, self.expected_2d_axis1)
+
+
+class TestLazy(UnmaskedTest, LazyMixin):
+    def test_1d(self):
+        for data, expected in self.data_1ds:
+            data = as_lazy_data(data)
```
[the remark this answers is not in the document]
wjbenfold: And used this instead of chunking

lib/iris/analysis/__init__.py
wjbenfold: Removed `map_complete_blocks`, kept the squeeze

lib/iris/tests/unit/analysis/test_MAX_RUN.py
wjbenfold: Now done

stephenworsley:
> I feel like my change to `map_complete_blocks` is useful, but it's also not necessary for this PR so I guess it's best removed?

You could always put it up as a draft PR for future reference. I have also copied modified the code in iris-esmf-regrid so theres a chance that `map_complete_blocks` gets altered in future as part of alligning regridding work. This would be nice to refer to if that happens.

lib/iris/analysis/__init__.py
stephenworsley: I think unmasking the array earlier in the code makes sense.

lib/iris/analysis/__init__.py
```diff
@@ -1472,6 +1473,46 @@ def _proportion(array, function, axis, **kwargs):
...
+def _lazy_max_run(array, axis=-1, **kwargs):
+    """
+    Lazily perform the calculation of maximum run lengths along the given axis
+    """
+    array = iris._lazy_data.as_lazy_data(array)
+    func = kwargs.pop("function", None)
+    if not callable(func):
+        emsg = "function must be a callable. Got {}."
+        raise TypeError(emsg.format(type(func)))
+    bool_array = da.ma.getdata(func(array))
```
stephenworsley, line 1485: So thinking about the proper way to do this sent me down a bit of a rabbit hole which ended up in me making a bug report to numpy (https://github.com/numpy/numpy/issues/21389).

Ultimately, I don't have full confidence in the line `bool_array = da.ma.getdata(func(array))`. Using `getdata` fetches the values underneath the mask, which seem to work for the test cases, but I'm not sure it would work for all reasonable cases of `func`. I think `da.ma.filled(func(array), 0)` should be more robust here as it ensures that every masked point _must_ be False.

From doing a bit of a dive into numpy, it appears that inequalities, which are probably one of the most common types of `func` we can expect, still have some issues with masked arrays (e.g. https://github.com/numpy/numpy/issues/10092), so I think it's worth being as robust as possible here.

stephenworsley (changes requested):

lib/iris/analysis/__init__.py
wjbenfold: Doesn't the line below cause us to ignore whatever applying `func` to the masking has done? I don't think we can have a problem with this unless we get an exception from applying `func` or what comes out doesn't have truthiness?

wjbenfold:
> You could always put it up as a draft PR for future reference. I have also copied modified the code in iris-esmf-regrid so theres a chance that `map_complete_blocks` gets altered in future as part of alligning regridding work. This would be nice to refer to if that happens.

Done in #4708

lib/iris/analysis/__init__.py
stephenworsley: Ah yes, I think you're right. I guess your aproach might actually be more robust in the edge case where `func` does not preserve the mask.

stephenworsley (approved):

rcomer:
This one still needs a whatsnew entry 👀

wjbenfold:
> This one still needs a whatsnew entry

You're the hero we need

Issue #4598, opened 2022-02-23
Use `meta` correctly in `map_blocks` to prevent dask from passing a 0d array through the regridder


wjbenfold:
### Context

`dask.array.map_blocks` will under some circumstances pass a 0d array through the provided function when initialised, as documented in https://docs.dask.org/en/latest/generated/dask.array.map_blocks.html
In https://github.com/SciTools/iris/blob/4abaa8f5d4be918b2e0b7bd42fbcc1e0a0196dd6/lib/iris/_lazy_data.py#L355-L388 we call map_blocks without `meta`, including when handing it an area weighted regridding function (and presumably other times) that won't pass through a 0d array. We do the same elsewhere in the same file too.

### Issues arising

* #4574 documents a deprecation warning seen when the Iris tests are run as the 0d array is passed in by dask and then indexed.
* I don't know if we get performance or safety improvements by adding this in, it's more that we're "doing it properly" / using dask as designed. Seems like a good way to make it easier to understand the codebase though.

### Suggestions

* Work out how to choose what the `meta` kwarg should be set to, and set it
* Consider whether the dtype argument should also be provided

wjbenfold:
Turns out I was wrong the other day when I told @trexfeathers that all this needed was working out, and that it didn't need deciding first...

## Investigation

`map_complete_blocks` is called twice within Iris:

### `_area_weighted`

https://github.com/SciTools/iris/blob/2fa7c598d4bcbe7e861b12b28a874360af974421/lib/iris/analysis/_area_weighted.py#L1109-L1111

The return dtype is set within area weighted regridding using `np.promote_types(src_data.dtype, np.float16)`. We could:

- have `_regrid_area_weighted_rectilinear_src_and_grid__perform` calculate the dtype it wants and pass this as an argument to `_regrid_area_weighted_array`.
- tightly couple `_regrid_area_weighted_rectilinear_src_and_grid__perform` and `_regrid_area_weighted_array` by computing the promotion twice, and then passing it as a kwarg to `map_complete_blocks`.
- have `_regrid_area_weighted_array` catch a 0d input array as obviously from dask and return the correct type of array to satisfy it.

### `_regrid`

https://github.com/SciTools/iris/blob/9268ca9c70bd160c35c5099025d632ae33faf858/lib/iris/analysis/_regrid.py#L1087-L1089

[...]

PR #4684, opened 2022-04-08
Update percentile tests

rcomer:
### Description

Follow up to #3901, with some updates to the percentile tests:

* I realized the tests for checking that our `alphap`, `betap` kwargs get passed to `scipy` weren't quite right for the lazy case because
  * Passing a non-lazy array means that `map_blocks` never actually gets called!
  * We _expect_ the `mquantiles` call to happen when the data is realized, so we need a call to `as_concrete_data`.  Note that the tests would currently pass without the `as_concrete_data` call, because `mquantiles` is initially applied to a 0D array, as described at #4598, but that might change at some point.

* Added basic tests to verify that `numpy.percentile` gets called as expected for fast method.

Happily, the tests all still pass 🙂

---

### DIFF
--- a/lib/iris/tests/unit/analysis/test_PERCENTILE.py
+++ b/lib/iris/tests/unit/analysis/test_PERCENTILE.py
@@ -11,6 +11,7 @@ import iris.tests as tests  # isort:skip
 
 from unittest import mock
 
+import dask.array as da
 import numpy as np
 import numpy.ma as ma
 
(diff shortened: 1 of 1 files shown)


lib/iris/tests/unit/analysis/test_PERCENTILE.py
```diff
@@ -147,23 +148,35 @@ def test_masked_2d_multi(self):
             data, axis, percent, expected, mdtol=mdtol, approx=True
         )
 
-    @mock.patch("scipy.stats.mstats.mquantiles")
+    @mock.patch("scipy.stats.mstats.mquantiles", return_value=[2, 4])
```
wjbenfold, line 151: I'm not sure that it makes sense for these two checks to be in the `MaskedAggregateMixin` because they don't use masked data? I think they belong in `Test_aggregate` instead.

lib/iris/tests/unit/analysis/test_PERCENTILE.py
```diff
@@ -147,23 +148,35 @@ def test_masked_2d_multi(self):
...
     def test_default_kwargs_passed(self, mocked_mquantiles):
         data = np.arange(5)
-        percent = 50
+        percent = [42, 75]
```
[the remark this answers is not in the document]
wjbenfold: Seems fine to me

wjbenfold (changes requested):

lib/iris/tests/unit/analysis/test_PERCENTILE.py
wjbenfold: Yeah, maybe that's the most straightforward way? (or a `SlowMixin`, but we're already pretty far into `Mixin`s so that may well get confusing).

rcomer: Or just rename this Mixin?

wjbenfold: Yeah - call it the `Slow` or the `Scipy` or something then the docstring can explain what belongs in here (masked and scipy mocking)?

rcomer: done

wjbenfold (approved):

wjbenfold:
It occurs to me (too late) - worth a what's new or is this just an extension of the previous percentile PR enough not to warrant its own?

rcomer:
Thanks @wjbenfold!  I'm thinking this is effectively part of #3901, so not worth it's own whatsnew.

PR #4699, opened 2022-04-21
Git ignore the benchmarks data folder

wjbenfold:
### Description

The benchmarks make a `.data` folder, I suggest that this should be git ignored

---

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -26,9 +26,9 @@ pip-cache
 .tox
 .pytest_cache
 
-# asv environments, results
+# asv data, environments, results
 .asv
-.data
+benchmarks/.data
 
 #Translations
 *.mo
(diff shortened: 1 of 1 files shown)


.gitignore
```diff
@@ -28,6 +28,7 @@ pip-cache
 
 # asv environments, results
 .asv
+.data
```
trexfeathers, line 31: It's misleading to state that _any_ directory called `.data` relates to "`asv environments, results`". We should either:

* Stick this line under a different comment heading.
* Use the specific path (so that would be `benchmarks/.data` I think)

trexfeathers (changes requested):

.gitignore
wjbenfold: [579358c](https://github.com/SciTools/iris/pull/4699/commits/579358c8106b3f5e5b8b947486f835d5af91a7ef)

trexfeathers (approved):

PR #4701, opened 2022-04-21
Change Cell equality to equate nans (fixes #4681)

### LINKED ISSUE #4681: Unexpected behaviour when merging cubelist with constant NaN-valued scalar coord
Merge seems to treat scalar coordinates differently when they have value NaN. I am trying to merge a list of cubes, each of which has a scalar coordinate with the same value. If this value is not NaN, the merge works as expected: the resulting cube has a scalar coordinate identical to the one in the input cubes. But if the value is NaN, this coordinate becomes 2-dimensional in the result. 

Example below:
```python
import iris
import iris.cube
import numpy as np
from iris.coords import DimCoord, AuxCoord

data = np.random.random((3, 3))
dim_coords_and_dims = [(DimCoord(np.arange(3), long_name="coord1"), 0)]

# aux_coord is constant non-nan value, merged cube contains aux_coord as a scalar coord as expected
aux_coord = AuxCoord([0], long_name="aux_coord")
cubelist = iris.cube.CubeList()
for i in range(2):
    for j in range(2):
        aux_coords_and_dims = [
            (DimCoord([i], long_name="merge_coord1"), None),
            (DimCoord([j], long_name="merge_coord2"), None),
            (aux_coord, None),
        ]
        c = iris.cube.Cube(
            data,
            dim_coords_and_dims=dim_coords_and_dims,
            aux_coords_and_dims=aux_coords_and_dims,
        )
        cubelist.append(c)

cube = cubelist.merge_cube()
print(cube)
print(cube.coord("aux_coord"))

# aux_coord is nan, merged cube contains aux_coord as a 2-dimensional auxiliary coord
aux_coord = iris.coords.AuxCoord([np.nan], long_name="aux_coord")
cubelist = iris.cube.CubeList()
for i in range(2):
    for j in range(2):
        aux_coords_and_dims = [
            (DimCoord([i], long_name="merge_coord1"), None),
            (DimCoord([j], long_name="merge_coord2"), None),
            (aux_coord, None),
        ]
        c = iris.cube.Cube(
            data,
            dim_coords_and_dims=dim_coords_and_dims,
            aux_coords_and_dims=aux_coords_and_dims,
        )
        cubelist.append(c)

cube = cubelist.merge_cube()
print(cube)
print(cube.coord("aux_coord"))
```

stephenworsley:
### Description

Fixes #4681 by making Cell equality handle nans in the same way as Coord equality, i.e. returning true when comparing a nan value with another value. This causes scalar coords containing nans to merge into one scalar coord rather than creating a long AuxCoord.

---

### DIFF
--- a/lib/iris/coords.py
+++ b/lib/iris/coords.py
@@ -1358,9 +1358,19 @@ class Cell(namedtuple("Cell", ["point", "bound"])):
         ):
             if self.bound is not None:
                 return self.contains_point(other)
+            elif isinstance(other, (float, np.number)) and np.isnan(other):
+                return isinstance(self.point, (float, np.number)) and np.isnan(
+                    self.point
+                )
             else:
                 return self.point == other
         elif isinstance(other, Cell):
+            if isinstance(other.point, (float, np.number)) and np.isnan(
+                other.point
+            ):
+                return isinstance(self.point, (float, np.number)) and np.isnan(
+                    self.point
+                )
             return (self.point == other.point) and (
                 self.bound == other.bound or self.bound == other.bound[::-1]
             )
--- a/lib/iris/coords.py
+++ b/lib/iris/coords.py
@@ -1353,24 +1353,27 @@ class Cell(namedtuple("Cell", ["point", "bound"])):
         compared.
 
         """
+
+        def nan_equality(x, y):
+            return (
+                isinstance(x, (float, np.number))
+                and np.isnan(x)
+                and isinstance(y, (float, np.number))
+                and np.isnan(y)
+            )
+
         if isinstance(other, (int, float, np.number)) or hasattr(
             other, "timetuple"
         ):
             if self.bound is not None:
                 return self.contains_point(other)
-            elif isinstance(other, (float, np.number)) and np.isnan(other):
-                return isinstance(self.point, (float, np.number)) and np.isnan(
-                    self.point
-                )
+            elif nan_equality(self.point, other):
+                return True
             else:
                 return self.point == other
         elif isinstance(other, Cell):
-            if isinstance(other.point, (float, np.number)) and np.isnan(
-                other.point
-            ):
-                return isinstance(self.point, (float, np.number)) and np.isnan(
-                    self.point
-                )
+            if nan_equality(self.point, other.point):
+                return True
             return (self.point == other.point) and (
                 self.bound == other.bound or self.bound == other.bound[::-1]
             )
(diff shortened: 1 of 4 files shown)


lib/iris/coords.py
```diff
@@ -1357,9 +1357,19 @@ def __eq__(self, other):
         ):
             if self.bound is not None:
                 return self.contains_point(other)
+            elif isinstance(other, (float, np.number)) and np.isnan(other):
```
trexfeathers, line 1360: You've done the same 4 times. This is crying out for a little convenience function?

lbdreyer:
The changes you have made look good. However I am unsure about whether this is the right thing to do. The change here is making NaN cell equal _by default_. I think I would prefer it if it required a user to "opt in" to nan equality.

I unfortunately don't know much about this, but it looks like IEEE-754 requires NaN to be not equal to any other float (incl. NaN) 
```
>>> np.nan == np.nan
False
```

I'm not sure how best to make it so that it is opt-in. It's more challenging than what was done in #3283. Perhaps a context manager?

pp-mo:
> @lbdreyer The changes you have made look good. However I am unsure about whether this is the right thing to do. The change here is **making NaN cell equal by default. I think I would prefer it if it required a user to "opt in" to nan equality.**

I can't really see why?

#3283 changed the behaviour of Coord equality regarding NaNs.
This changes Cell equality in a similar way.
I don't see a huge difference, and I can't forsee a case for which the existing behaviour would be 'correct' and the new one wrong.

@lbdreyer do you have a picture in your head of a case where this would be a bad idea ?

lbdreyer:
> https://github.com/SciTools/iris/pull/3283 changed the behaviour of Coord equality regarding NaNs.
This changes Cell equality in a similar way.

Admittedly I was only really focussing on on how #3283 adds the `withnans` kwarg to `iris.util.array_equal` rather than the intention behind the PR itself to compare coords.
As discussed just now, I am now worried we were wrong to do the previous PR but @pp-mo's point about how we are not performing arithmetic operations so the nan equality issue may not apply is starting to make sense. I'll just need to think about this a bit more

stephenworsley:
Assuming that NaNs shouldn't compare as equal for default iris behaviour, is there a case that the existing behaviour described in #4681 is not actually incorrect? If we decide to treat all NaNs as unkown but separate quantities, it seems like not merging them into one scalar would be the natural implication. Would it then make sense to have a context manager for NaN behaviour which you could use to switch all this behaviour?

lbdreyer (approved):

PR #4715, opened 2022-04-26
Benchmarking issues: put PR link in bullet point

rcomer:
### Description

I'm lazy and don't like to do any more clicking than necessary!  If a PR link appears in a bulleted list then the PR's title gets spelt out so you can see what it was without clicking through.

I don't know how to test this change to verify it will have the effect I expect...

---


trexfeathers (approved):
Thanks @rcomer 

>I don't know how to test this change to verify it will have the effect I expect...

We'll do it live!

Issue #1422, opened 2014-10-30
Unexpected datetimes obtained when producing monthly mean climatology data 


rockdoc:
Hi

I'm producing a cube of monthly mean climatology data from a multi-year time-series of monthly mean data values (ozone, as it happens). My sample netcdf dataset spans the 20 years 1990-01-16 to 2010-12-16. In reality the actual dataset will likely span >100 years.

I'm using `iris.coord_categorisation` to add a month number auxiliary coordinate. And then I'm calling `cube.aggregated_by` to aggregate across that coordinate, thus producing a resultant cube which contains 12 time coords, one for each month of the year. (The other dimensions of my data are latitude and height, but for the purposes of this issue I guess they could equally be some other axes.)

Now, if the time span of my sample data covers an _odd_ number of years, then the dates written to the aggregated cube are as expected, e.g. YYYY-01-16, YYYY-02-16, …, YYYY-12-16, where YYYY is the mid-year of the input time-series.

But if the time span of my sample data covers an _even_ number of years, then the dates written to the aggregated cube are rotated by 6 months, e.g. YYYY-07-16, YYYY-08-16, etc.  Assuming that the cube data was written in this same order, I was sorting it to get it into the normal CF climatology order of Jan to Dec.  However, it would appear that the data is in fact in the desired Jan-Dec order, but that the rotated datetimes are incorrect.

An examination of the time bounds (in the aggregated cube) would seem to corroborate this observation as they too run from Jan to Dec (although the upper bounds are not quite right, they should be one month later than the lower bound).

I have put together a small bit of test code (see [this gist](https://gist.github.com/rockdoc/e3a55de71e94377c7f1e)) which hopefully illustrates the scenario described above.

Phil

larsbarring:
I just ran into the same behaviour as described in the initial post (luckily I remembered reading this issue). Now, with the improved handling of climatologies in Iris, it would be good to bounce up this and make Iris produce a climatological time coordinate, as per CF section 7.4.

Issue #4665, opened 2022-03-28
aggregated_by() - time coords do not match expected output.


mokhodge:
## 🐛 Bug Report
This might link to #4455. When aggregate by month_number the time coords of the outputted cube are not always in the order you would expect yet the corresponding cube.data is.

## How To Reproduce
filepath included as an example, have tried on other files and same behaviour occurs

```
import iris.analysis
import matplotlib.pyplot as plt
import iris.quickplot as qplt
import iris.coord_categorisation
import iris.cube

fname = f'/home/h03/hadmi/Python/MedGOLD/Colombia/data_files/pr/pr_CHIRPS' \
        f'_1981-2010_regridded_colombia.nc'
cube = iris.load_cube(fname)
iris.coord_categorisation.add_month_number(cube, 'time',
                                               name='month_number')
ycon1 = iris.Constraint(
        time=lambda t: 1990 <= t.point.year <= 2010)
ycon2 = iris.Constraint(
        time=lambda t: 1991 <= t.point.year <= 2010)

cube1 = cube.extract(ycon1)
cube2 = cube.extract(ycon2)
monthly_clim1 = cube1.aggregated_by('month_number', iris.analysis.MEAN)
monthly_clim2 = cube2.aggregated_by('month_number', iris.analysis.MEAN)

fig = plt.figure()
ax = fig.add_subplot(2,2,1)
qplt.plot(monthly_clim1[:, 16, 22])
ax.set_title('Clim 1990 -2010')
ax.set_ylabel('mm/month') # this shows the monthly_clim1 cube starts in Jan as expected

bx = fig.add_subplot(2,2,2)
qplt.plot(monthly_clim2[:, 16, 22])
bx.set_title('Clim 1991 -2010')
bx.set_ylabel('mm/month') # this shows the monthly_clim2 cube starts in July (not expected). It appears the time coords just comes from the middle of the original time coords

plt.show()

#### Show the cube.data always starts from Jan even if cube.coord('time') says it is July
t_cube = cube2.copy()
t_cube = t_cube[:, 16, 22]
print(t_cube.coord('time')) #starts in Jan
``` 
DimCoord([1991-01-01 00:00:00, 1991-02-01 00:00:00, 1991-03-01 00:00:00,
       1991-04-01 00:00:00, 1991-05-01 00:00:00, 1991-06-01 00:00:00,
       1991-07-01 00:00:00, 1991-08-01 00:00:00, 1991-09-01 00:00:00,
[...]

PR #4723, opened 2022-04-29
More sensible climatological time points

### LINKED ISSUE #4098: Time points for daily climatology are not intuitive
## 📰 Custom Issue

Hello :)

I have a cube containing 4 full years of daily data. Using `aggregated_by`, I aggregate the cube using a "day" and "month" coordinate. This produces a daily climatology with time points that start in July 🤯 (the bounds are as expected).

Mathematically, this makes sense, since the mean of, e.g. (Jan 1, 2000), (Jan 1, 2001), (Jan 1, 2002) and (Jan 1, 2003) is in July 2001.

However, scientifically, this doesn't make sense, since a scientist would expect the first time point of a daily climatology created as above to be Jan 1.

The CF conventions state (at the end of paragraph 2 in [Section 7.4](http://cfconventions.org/cf-conventions/cf-conventions.html#climatological-statistics)): "The time coordinates should be values that are representative of the climatological time intervals, such that an application which does not recognise climatological time will nonetheless be able to make a reasonable interpretation.". I'm not sure this is happening in this case.

Would it be better to set the time points in the climatology file such that they are more representative of the climatological time intervals?

wjbenfold:
### Description

Closes #4098 and #4665 which both relate to points being chosen oddly in a climatological aggregation. Main issue arose from the average of Januarys in different years potentially being July, as identified by @trexfeathers on #4098.

I've introduced a "climatological" keyword to `iris.cube.Cube.aggregated_by` that makes it choose the point from the first cell in each climatology, and sets the `climatological` attribute of the resultant coord to `True` if it can.

I've also pulled the tests for `aggregated_by` out into their own file, because `test_Cube` is huge and this seems like the first step in rationalising that.

---


wjbenfold:
@rcomer - as someone with more experience of actually wanting to aggregate climatologies than me, what would you expect to happen to a 2D `AuxCoord` when one of its axes is the collapsing axis? Should that just fail the whole operation or should it get aggregated in some way?

rcomer:
Hi @wjbenfold, I think that for _any_ 2D coordinate, collapsing/aggregating on one of its dimensions should give something consistent to having sliced the coordinate and done the collapse/aggregate on 1D slices.  E.g. see this test for the standard numerical aggregated_by case:

https://github.com/SciTools/iris/blob/e3f3bb2d2027aab569db57b7368e0fd1447b03e5/lib/iris/tests/unit/cube/test_Cube.py#L755-L771

trexfeathers:
>I've also pulled the tests for `aggregated_by` out into their own file, because `test_Cube` is huge and this seems like the first step in rationalising that.

Just to further muddy the testing waters: I've just become aware of the extensive testing in [`tests/test_aggregate_by.py`](https://github.com/SciTools/iris/blob/main/lib/iris/tests/test_aggregate_by.py).

As far as I'm aware, tests in the `tests/` root is a deprecated practice - we now place all test modules in one of the appropriate subdirectories - so there's an opportunity for further rationalisation if you're feeling brave!

lib/iris/cube.py
```diff
@@ -4020,17 +4029,24 @@ def aggregated_by(self, coords, aggregator, **kwargs):
             dimensions=dimension_to_groupby, dim_coords=True
         ) or [None]
         for coord in groupby.coords:
+            new_coord = coord.copy()
+
+            try:
+                if climatological:
+                    new_coord.climatological = True
```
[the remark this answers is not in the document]
wjbenfold: I thought it looked like the coord needed to be copied and then this added to it because otherwise we were editing inplace, but I see now that it's already been copied by the call to `iris.util._strip_metadata_from_dims` so I think you're right that this can all move/go

rcomer:
> My intuition is that all this should only apply to time coordinates

Or maybe time-like coordinates.  I.e. `coord.units.is_time_reference()`.  The obvious example (other than “time”) is “forecast_reference_time”.  This would make sense for seasonal/decadal hindcasts: I have members initialised on 1st November each year for lots of years.  If I aggregate across these, it would make sense that the output “forecast_reference_time” is still 1st November.

rcomer:
This also closes #1422.  @wjbenfold the OP of #1422 links to a gist containing test cases.  Might be useful 👀.

wjbenfold:
> so there's an opportunity for further rationalisation if you're feeling brave!

I feel like bravery and spotting a quagmire are different, so I'll happily take the follow-up PR to rationalise them but I'd rather not drag them into this PR (in the hope that this one gets merged sometime soon)

wjbenfold:
> This also closes #1422. @wjbenfold the OP of #1422 links to a gist containing test cases. Might be useful 👀.

This was a nice check - I still get test failures but they fail how I expected because my view on what's sensible (give them the first point that was used to make the climatology) is different to the author's. Importantly, the months match (just not the years)

lib/iris/analysis/__init__.py
```diff
@@ -2186,6 +2188,11 @@ def __init__(self, groupby_coords, shared_coords=None):
             that share the same group-by coordinate axis.  The `int` identifies
             which dimension of the coord is on the group-by coordinate axis.
 
+        * climatological (bool):
+            Indicates whether the output is expected to be climatological. This
+            causes the climatological flag to be set on the resulting coord(s)
+            and the value of the collapsed coord to be set sensibly.
```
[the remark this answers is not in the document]
wjbenfold: Nice, I'll go for something similar

lib/iris/analysis/__init__.py
```diff
@@ -2410,12 +2419,16 @@ def _compute_shared_coords(self):
...
                     first_choices = coord.bounds.take(0, -1)
                     last_choices = coord.bounds.take(1, -1)
+
                 else:
                     # Derive new coord's bounds from points.
                     item = coord.points
                     maxmin_axis = dim
                     first_choices = last_choices = coord.points
 
+                if self.climatological:
```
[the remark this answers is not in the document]
wjbenfold: Yeah, that seems reasonable

lib/iris/analysis/__init__.py
```diff
@@ -2453,23 +2466,39 @@ def _compute_shared_coords(self):
                             ]
                         )
 
+                if self.climatological:
+                    for indices in groupby_indices:
+                        new_points.append([coord.points.take(indices, dim)])
```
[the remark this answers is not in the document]
wjbenfold: Done

lib/iris/analysis/__init__.py
```diff
@@ -2453,23 +2466,39 @@ def _compute_shared_coords(self):
...
                 # Bounds needs to be an array with the length 2 start-stop
                 # dimension last, and the aggregated dimension back in its
                 # original position.
                 new_bounds = np.moveaxis(
                     np.array(new_bounds), (0, 1), (dim, -1)
                 )
 
+                if self.climatological:
+                    new_points = np.moveaxis(
+                        np.array(new_points), (0, 1), (dim, -1)
```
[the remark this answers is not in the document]
wjbenfold: Cool

lib/iris/cube.py
```diff
@@ -3831,46 +3831,55 @@ def collapsed(self, coords, aggregator, **kwargs):
...
-
-        Kwargs:
-
-        * kwargs:
-            Aggregator and aggregation function keyword arguments.
-
-        Returns:
+        climatological : bool
+            Indicates whether the output is expected to be climatological. This
+            causes the points of the collapsed coord to be set to match the
```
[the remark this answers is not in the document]
wjbenfold: Matched to other docstring

lib/iris/cube.py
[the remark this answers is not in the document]
wjbenfold: Moved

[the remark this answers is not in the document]
wjbenfold: Turns out I can't do this as-is because the `self.coord_dims(coord)` call on line 4041 fails if the metadata has changed (including climatologicalness). I can hack it back long enough to do the check, or add the climatological after the copy, or carry the coord_dims info from earlier or do something more involved.

lib/iris/tests/unit/cube/test_Cube__aggregated_by.py
```diff
@@ -0,0 +1,516 @@
...
+import iris.analysis
+from iris.analysis import MEAN, Aggregator
+import iris.aux_factory
+import iris.coords
+from iris.coords import AncillaryVariable, AuxCoord, CellMeasure, DimCoord
+from iris.cube import Cube
+import iris.exceptions
+
+
+class Test_aggregated_by(tests.IrisTest):
```
wjbenfold, line 27: None of the tests in this file appear to be running in the CI and I don't know why

rcomer: Does CI rely on the ifmain?  E.g.

https://github.com/wjbenfold/iris/blob/239114f89118573880a1d4fa0799f2e9af8a4bd7/lib/iris/tests/unit/cube/test_Cube.py#L2878-L2879

lib/iris/cube.py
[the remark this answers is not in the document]
rcomer: Unfortunately yes, you can have coords with the same name.  Sorry I may have led you on a wild goose chase here and maybe your original approach was the simplest.  Or maybe @bjlittle may have a better suggestion as he is the Metadata King 🤴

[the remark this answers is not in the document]
wjbenfold: @rcomer has just told me that's not a good assumption, I'll find another way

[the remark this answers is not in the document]
wjbenfold: I'm now maintaining a lookup on the `_Groupby` object that records which cube coord the resulting coords came from

lib/iris/tests/unit/cube/test_Cube__aggregated_by.py
wjbenfold: I should probably have one anyway, I've added it

wjbenfold: Still not finding them

wjbenfold: Still not finding them

lib/iris/tests/unit/cube/test_Cube__aggregated_by.py
```diff
@@ -0,0 +1,771 @@
...
+            np.array([np.arange(10), np.arange(10, 20)]).T,
+        )
+        self.assertTrue(aligned_coord.climatological)
+
+        categorised_coord1 = result.coord("cat1")
+        self.assertArrayEqual(
+            categorised_coord1.points, np.tile(np.arange(2), 5)
+        )
+        self.assertIsNone(categorised_coord1.bounds)
+        # TODO: I assume the categorised coord shouldn't be climatological?
```
rcomer, line 744: 👍

lib/iris/tests/unit/cube/test_Cube__aggregated_by.py
```diff
@@ -0,0 +1,771 @@
...
+        self.assertFalse(categorised_coord1.climatological)
+
+        categorised_coord2 = result.coord("cat2")
+        self.assertArrayEqual(
+            categorised_coord2.points, np.tile(np.arange(5), 2)
+        )
+        self.assertIsNone(categorised_coord2.bounds)
+        # TODO: I assume the categorised coord shouldn't be climatological?
```
rcomer, line 752: 👍

lib/iris/analysis/__init__.py
```diff
@@ -2510,17 +2545,19 @@ def _compute_shared_coords(self):
...
-                    )
+                new_coord = iris.coords.AuxCoord.from_coord(coord).copy(
+                    points=new_points, bounds=new_bounds
                 )
 
+            if climatological_coord:
+                new_coord.climatological = True
+
+            self.coords.append(new_coord)
+            self.coord_replacement_mapping.append((coord, new_coord))
```
rcomer, line 2559: Maybe make `coord_replacement_mapping` a dictionary, then
```suggestion
            self.coord_replacement_mapping[new_coord] = coord
```

lib/iris/cube.py
```diff
@@ -4103,17 +4113,27 @@ def aggregated_by(self, coords, aggregator, **kwargs):
...
             else:
+                # The metadata may have changed (e.g. climatology), so check if
+                # there's a better coord to pass to self.coord_dims
+                lookup_coord = coord
+                for (
+                    cube_coord,
+                    groupby_coord,
+                ) in groupby.coord_replacement_mapping:
+                    if coord == groupby_coord:
+                        lookup_coord = cube_coord
```
rcomer, line 4133: If `coord_replacement_mapping` was a dictionary, could just do

```python
lookup_coord = groupby.coord_replacement_mapping.get(coord, coord)
```

rcomer (changes requested):
I was in so deep, I figured I might as well officially take this on!  For the tests, I only looked through the "climatological" test class as I assume the others are a straight cut/paste from `test_Cube.py`.

wjbenfold:
From discussion on 4/5, should use first bound rather than points of first cell because for e.g. season aggregation (month and season) we end up with Dec16 referring to Dec1 - Feb28

lib/iris/cube.py
wjbenfold: I don't think I can do that because a coord isn't hashable so can't be used as a dictionary key?

rcomer: Huh.  Right you are.

lib/iris/tests/unit/cube/test_Cube__aggregated_by.py
wjbenfold: I think I've hunted this down - the file was executable, which is apparently enough for the test runner to decide it can't possibly be a test

rcomer: 🤯 

Good hunting skills!

wjbenfold: "Once you eliminate every line in the file and the filename, whatever difference remains must be the bug" - Sherlock Holmes

rcomer (changes requested):
Hi @wjbenfold this looks in good shape now.  I have a few more comments but I think they are all easy to address.

lib/iris/tests/unit/cube/test_Cube__aggregated_by.py
```diff
@@ -0,0 +1,829 @@
...
+            categorised_coords,
+            self.aggregator,
+            climatological=climatological_op,
+        )
+
+        return out_cube
+
+    def test_basic(self):
+        # Check the least complicated version works (set climatological, set
+        # points correctly).
```
rcomer, line 658: Now that we are using `pytest` as the runner (#4734) the test paths will appear in the output regardless of whether the tests have docstrings.  So you could change these back to docstrings if you want.  I have no opinion either way.

wjbenfold: Done

lib/iris/cube.py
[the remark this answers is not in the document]
wjbenfold: Fixed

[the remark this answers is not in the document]
wjbenfold: (good spot)

rcomer (approved):

rcomer:
Thanks @wjbenfold.  LGTM.

PR #4727, opened 2022-05-03
Redirect whatsnew to latest

wjbenfold:
### Description

In #4540, @bjlittle gave us a one-click link from homepage to latest what's new, something that we have lost on the move to pydata and that was definitely useful for developers (and hopefully useful for users).

This PR is to restore that behaviour, by having the top bar link and quick link on the front page both point to the latest what's new (we don't really lose anything as the what's new index is on the left of any what's new page, so people looking for any other what's new are still only two clicks away).

I've managed to change the button by moving the tag at the top of the page, but I don't know enough about how pydata's building the top bar to work out what should change to redirect that. @tkknight?

I think this means that the what's new index page is now inaccessible, so it could be removed?

It's not clear to me how the developer guide docs on releasing should change based on this change because we don't have stable pydata-based docs yet - we'll need to remove the `_iris_whatsnew` label from any pages that aren't the current most up-to-date one.

---

### DIFF
--- a/docs/src/developers_guide/release.rst
+++ b/docs/src/developers_guide/release.rst
@@ -257,8 +257,9 @@ Post Release Steps
    `Read The Docs`_ to ensure that the appropriate versions are ``Active``
    and/or ``Hidden``. To do this ``Edit`` the appropriate version e.g.,
    see `Editing v3.0.0rc0`_ (must be logged into Read the Docs).
-#. Make a new `latest.rst` from `latest.rst.template` and update the include
-   statement and the toctree in `whatsnew.rst` to point at the new `latest.rst`.
+#. Make a new ``latest.rst`` from ``latest.rst.template`` and update the include
+   statement and the toctree in ``whatsnew.rst`` to point at the new
+   ``latest.rst``.
 #. Merge back to ``main``
 
 
--- a/docs/src/developers_guide/release.rst
+++ b/docs/src/developers_guide/release.rst
@@ -221,6 +221,8 @@ Release Steps
 
    * Use ``git`` to rename ``docs/src/whatsnew/latest.rst`` to the release
      version file ``v1.9.rst``
+   * Update ``docs/src/whatsnews/whatsnew.rst`` to rename ``latest.rst`` in the
+     include statement and toctree.
    * Use ``git`` to delete the ``docs/src/whatsnew/latest.rst.template`` file
    * In ``v1.9.rst`` remove the ``[unreleased]`` caption from the page title.
      Note that, the Iris version and release date are updated automatically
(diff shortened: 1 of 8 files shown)


tkknight:
@wjbenfold and I had a quick chat, main points:

* We can link to ``latest.rst`` from the top bar but should retain the ``What’s New in Iris`` link text (sourced from the toc in the index) 
* We should update the release process if needed - but not make it any more complicated, it is easy to keep changing the release process each time on these type of things.
* We should ensure a clean build of the docs with no warnings

docs/src/developers_guide/release.rst
```diff
@@ -255,6 +257,8 @@ Post Release Steps
    `Read The Docs`_ to ensure that the appropriate versions are ``Active``
    and/or ``Hidden``. To do this ``Edit`` the appropriate version e.g.,
    see `Editing v3.0.0rc0`_ (must be logged into Read the Docs).
+#. Make a new `latest.rst` from `latest.rst.template` and update the include
```
tkknight, line 260: should this be using double single quotes? (same on next line)

docs/src/whatsnew/whatsnew.rst
```diff
@@ -1,11 +1,11 @@
...
-******************
-
-These "What's new" pages describe the important changes between major
-Iris versions.
+------------------
 
+.. include:: latest.rst
 
 .. toctree::
    :maxdepth: 1
```
tkknight, line 11: When the page is rendered the latest info is shown via a single click  but the toc is then also displayed at the bottom which doesn't quite look right imho.  As we have the toc in the left sidebar I think we could hide the toc in the main body by using the ``:hidden:`` option under the toctree directive, I think that will solve it.

tkknight (changes requested):
Looks good, a nice way to ensure the one click to the latest whats new entry.  Just a couple minor of suggested changes.

docs/src/whatsnew/whatsnew.rst
wjbenfold: [aab13c4](https://github.com/SciTools/iris/pull/4727/commits/aab13c401ef69f3f359cbbcbfd9fbac11acf63ae)

docs/src/developers_guide/release.rst
wjbenfold: [aab13c4](https://github.com/SciTools/iris/pull/4727/commits/aab13c401ef69f3f359cbbcbfd9fbac11acf63ae)

wjbenfold:
Review fixes made, and a rename of a section title that always confuses me (can always drop that back out if you're not a fan)

tkknight (approved):

PR #4729, opened 2022-05-05
Cell comparison: remove ancient netcdftime/cftime workarounds

### LINKED ISSUE #4697: Problems comparing time coord cell objects
## 📰 Custom Issue
From iris v3, cells for time coordinates now use cftime.datetime objects. These objects cannot be sorted or compared in the guise of coordinate cells, e.g.

```python
import cftime
import iris.coords

t1 = cftime.datetime(2022, 3, 8, 12, 0, 0)
t2 = cftime.datetime(2022, 3, 9, 12, 0, 0)
c1 = iris.coords.Cell(point=t1)
c2 = iris.coords.Cell(point=t2)
cells = [c1, c2]
print(sorted(cells))
```
will fail with the error:

```
  File "./test_time_sort.py", line 19, in <module>
    print(sorted(cells))
  File "lib/python3.8/site-packages/iris/coords.py", line 1513, in __lt__
    return self.__common_cmp__(other, operator.lt)
  File "lib/python3.8/site-packages/iris/coords.py", line 1419, in __common_cmp__
    "Cannot determine the order of " "cftime.datetime objects"
TypeError: Cannot determine the order of cftime.datetime objects
```

This protection was probably added to the cells construct at a time when cftime.datatime objects could not be compared, but now they can and as such the protection should be removed, e.g. from the above t1 and t2 can be compared sensibly. I would hope that lines 1413 to 1420 of coords.py can be simply removed. However, a tranche of tests will also likely need to be re-engineered as well.

rcomer:
### Description

Closes #4697.  The example given there now produces

```
[Cell(point=cftime.datetime(2022, 3, 8, 12, 0, 0, 0, calendar='standard', has_year_zero=False), bound=None), Cell(point=cftime.datetime(2022, 3, 9, 12, 0, 0, 0, calendar='standard', has_year_zero=False), bound=None)]
```

The first workaround was added at #1016.  From the comments on the tests I removed, the comparisons were then falling back to comparing object ids, so presumably rich comparison was not implemented in that version of netcdftime _at all_.  It is implemented now:
https://github.com/Unidata/cftime/blob/dc75368cd02bbcd1352dbecfef10404a58683f94/src/cftime/_cftime.pyx#L1314

The second workaround was added at #2440.  The comment there says that `NotImplemented` was not always being returned by the comparison method.  If I've understood correctly, @bjlittle put in a fix for this at https://github.com/Unidata/cftime/pull/81 where, as far as I can tell, python2 was the main complication.  @bjlittle's fix was merged in 2018 and we are now [pinned to cftime >= 1.5](https://github.com/SciTools/iris/blob/7db0bdb6acad80537ac33d3b051155dabfeb5e39/requirements/ci/py38.yml#L15) which is [only a year old](https://github.com/Unidata/cftime/releases/tag/v1.5.0rel).  So I think we should be safe.

I'm currently targetting `main`, but I wonder if this should actually go in a patch release?  #4697 suggests that this has now become a problem because time cell points have changed from being `datetime.datetime` to `cftime.datetime`.  This was a [breaking change in cf-units v3.0](https://github.com/SciTools/cf-units/releases/tag/v3.0.0).

---

### DIFF
--- a/lib/iris/tests/unit/coords/test_Cell.py
+++ b/lib/iris/tests/unit/coords/test_Cell.py
@@ -30,33 +30,6 @@ class Test___common_cmp__(tests.IrisTest):
         with self.assertRaisesRegex(exception_type, regexp):
             cell >= other
 
-    def test_cftime_cell(self):
-        # Check that cell comparison when the cell contains
-        # cftime.datetime objects raises an exception otherwise
-        # this will fall back to id comparison producing unreliable
-        # results.
-        cell = Cell(cftime.datetime(2010, 3, 21))
-        dt = mock.Mock(timetuple=mock.Mock())
-        self.assert_raises_on_comparison(
-            cell, dt, TypeError, "determine the order of cftime"
-        )
-        self.assert_raises_on_comparison(
-            cell, 23, TypeError, "determine the order of cftime"
-        )
-        self.assert_raises_on_comparison(
-            cell, "hello", TypeError, "Unexpected type.*str"
-        )
-
-    def test_cftime_other(self):
-        # Check that cell comparison to a cftime.datetime object
-        # raises an exception otherwise this will fall back to id comparison
-        # producing unreliable results.
-        dt = cftime.datetime(2010, 3, 21)
-        cell = Cell(mock.Mock(timetuple=mock.Mock()))
-        self.assert_raises_on_comparison(
-            cell, dt, TypeError, "determine the order of cftime"
-        )
-
     def test_PartialDateTime_bounded_cell(self):
         # Check that bounded comparisions to a PartialDateTime
         # raise an exception. These are not supported as they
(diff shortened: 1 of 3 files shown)


lib/iris/tests/unit/coords/test_Cell.py
```diff
@@ -85,10 +58,9 @@ def test_PartialDateTime_unbounded_cell(self):
     def test_datetime_unbounded_cell(self):
         # Check that cell comparison works with datetimes.
         dt = datetime.datetime(2000, 6, 15)
-        cell = Cell(datetime.datetime(2000, 1, 1))
-        # Note the absence of the inverse of these
-        # e.g. self.assertGreater(dt, cell).
-        # See http://bugs.python.org/issue8005
+        cell = Cell(cftime.datetime(2000, 1, 1))
```
rcomer, line 61: Changed this to `cftime.datetime` as I assume the usual way for a `datetime` object to appear in a cell is via `num2date`.  E.g.
https://github.com/SciTools/iris/blob/7db0bdb6acad80537ac33d3b051155dabfeb5e39/lib/iris/coords.py#L2180-L2183

trexfeathers:
@pdearnshaw we're not planning a bugfix release for this, so it would instead be released as part of [Iris `v3.3`](https://github.com/SciTools/iris/discussions/4300). Is that OK with you or is this a particularly urgent issue?

trexfeathers (changes requested):
@rcomer What's New and I'll merge 👍

trexfeathers (approved):

rcomer:
Thanks @trexfeathers!

PR #4734, opened 2022-05-09
Replace nose with pytest

lbdreyer:
We currently run the tests using nose, but as nose is effectively dead, I propose we replace this with pytest which [can run unittest tests](https://docs.pytest.org/en/6.2.x/unittest.html#unittest-testcase-support).

Note, pytest allows you to set the number of processors to "auto" however when I did this the workers kept being killed.[ pytest-xdist uses psutil to determine the cpu count](https://github.com/pytest-dev/pytest-xdist/blob/v2.5.0/src/xdist/plugin.py#L16-L44) and if that is not available it uses `multiprocessing.cpu_count` which on cirrus returns 30.
We can continue using the "Magic number" we were using to set number of processors (if not given), as I have done in this PR, or we need to add psutil as an extra test dependency.


lib/iris/tests/runner/_runner.py
```diff
@@ -129,35 +129,26 @@ def run(self):
                     "WARNING: Gallery path %s does not exist." % (gallery_path)
                 )
         if not tests:
-            tests.append("iris.tests")
-
-        regexp_pat = r"--match=^([Tt]est(?![Mm]ixin)|[Ss]ystem)"
```
trexfeathers, line 134: Might be worth renaming [`test_constraints.TestMixin`](https://github.com/SciTools/iris/blob/c047007228bafc7f38d78c0d2b5b54d32b387cb9/lib/iris/tests/test_constraints.py#L94)?

It doesn't contain any tests so won't be run, but we don't want to mislead users into thinking that it's OK to start a Mixin name with `Test`. If I understand correctly the only remaining protection against `Mixin` tests being run is that they don't start with `Test`. Please correct me if I'm wrong.

trexfeathers (changes requested):
Thanks @lbdreyer.

IMO `psutil` would be a good move. I just tried installing into the environment specified by this branch and got the below, which seems to indicate that `psutil` doesn't add dependencies:

<details>
<summary>Conda plan</summary>

```
The following NEW packages will be INSTALLED:

  psutil             conda-main/linux-64::psutil-5.8.0-py38h27cfd23_1

The following packages will be UPDATED:

  ca-certificates    conda-forge::ca-certificates-2021.10.~ --> conda-main::ca-certificates-2022.4.26-h06a4308_0
  libedit            conda-forge::libedit-3.1.20191231-he2~ --> conda-main::libedit-3.1.20210910-h7f8727e_0
  readline             conda-forge::readline-8.1-h46c0cb4_0 --> conda-main::readline-8.1.2-h7f8727e_1

The following packages will be SUPERSEDED by a higher-priority channel:

  openssl            conda-forge::openssl-1.1.1o-h166bdaf_0 --> conda-main::openssl-1.1.1o-h7f8727e_0

The following packages will be DOWNGRADED:

  esmf                                                      --> 
  esmpy                                                     --> 
  netcdf-fortran                                            --> 

```
</details>

lib/iris/tests/runner/_runner.py
rcomer: Yes.  See https://github.com/SciTools/iris/pull/4249#discussion_r674632721.

trexfeathers (approved):

trexfeathers:
Great! Sounds like it took a lot of experiments to get here, so thanks @lbdreyer for persevering.

lbdreyer:
To ensure we aren't going to lose some tests in the switch over due to the different test discovery in nose vs pytest I have compared the tests that are run in [this PR](https://cirrus-ci.com/build/6478271670910976) against the latest run on [master](https://cirrus-ci.com/build/5403419077771264) and I believe we are running all tests as before (plus some extra ones that oddly weren't being run with nose)

## Nose Results
### doctests and gallery task:
RESULT: `Ran 24 tests in 72.282s`

### test task: 
* "--system tests" 
RESULT: `Ran 2 tests in 0.819s`
* "--default-tests
RESULT: `Ran 6230 tests in 156.024s ... OK (SKIP=21)`

## Pytest Results
### doctests and gallery task:
RESULT: `================= 24 passed, 104 warnings in 92.11s (0:01:32) ==================`

### test task: 
* "--system tests" 
RESULT: `======================== 2 passed, 18 warnings in 8.16s ========================`
* "--default-tests
RESULT: `========= 6228 passed, 21 skipped, 1627 warnings in 184.96s (0:03:04) ==========`

## Analysis:
The doctests and system tests match but for default-tests, pytest is running more tests in total (6249) than nose (6230). I'm not sure why these tests weren't being run previously(!) but I think I am comfortable that we at least aren't losing anything with this change.

rcomer:
My previous experience for why `pytest` finds more tests than `nose` are

* Missing `__init__.py` in a test sub-package - `nose` needs this file but `pytest` doesn't care.
* Tests created using `setattr` on the class: #4431

lbdreyer:
Thanks for reviewing and merging @trexfeathers !

rcomer:
At https://github.com/SciTools/iris/pull/4723#discussion_r863729049, @wjbenfold established that nose also won't find test modules that are executable.  This is the case for

https://github.com/SciTools/iris/blob/main/lib/iris/tests/unit/analysis/test_MAX_RUN.py
https://github.com/SciTools/iris/blob/main/lib/iris/tests/unit/io/test_save.py
https://github.com/SciTools/iris/blob/main/lib/iris/tests/integration/test_Datums.py
https://github.com/SciTools/iris/blob/main/lib/iris/tests/unit/io/test__generate_cubes.py

which between them contain 19 tests.

wjbenfold:
I don't suppose you've run those tests to check they pass @rcomer? I'm pretty sure I've relied on at least some of them running in the CI to know PRs were good...

jamesp:
The log output from pytest is quite hard to read, I think because all the warnings get shown in the summary *after* the test results?  See here: https://cirrus-ci.com/task/4564444779905024?logs=tests as an example. It means the summary of the failures in the github checks window doesn't show the details of which tests have failed. Is there a config we could switch somewhere to either suppress that warning output, or switch the order they are printed?

rcomer:
Apparently the `—disable-warnings` flag is [not recommended](https://docs.pytest.org/en/7.1.x/how-to/capture-warnings.html?highlight=disable%20warnings#disabling-warnings-summary), though I use it all the time when running locally.  Lots of of other options in that section.

trexfeathers:
>The log output from pytest is quite hard to read, I think because all the warnings get shown in the summary after the test results?

I hope that one day this will be what we want (but not right this instant!): #4384

PR #4739, opened 2022-05-12
Mesh docs - include examples of converting different mesh formats.

trexfeathers:
### Description

This adds a page to demonstrate how to convert meshes of different formats into Iris' mesh data model, so that users can enjoy current and future mesh-specific features in Iris.

---


trexfeathers:
@pp-mo has suggested adding some PNG's to show readers the kind of data they are looking at.

lbdreyer:
Could you also add a what's new?

docs/src/further_topics/ugrid/other_meshes.rst
```diff
@@ -0,0 +1,223 @@
...
+        >>> print(temperature_cube.coord("longitude"))
+        AuxCoord :  longitude / (degrees)
+            points: <lazy>
+            bounds: <lazy>
+            shape: (126859,)  bounds(126859, 18)
+            dtype: float64
+            standard_name: 'longitude'
+            var_name: 'lon'
+
+        # Replace existing face AuxCoords with MeshCoords.
```
lbdreyer, line 62:
```suggestion
        # Replace existing face AuxCoords with new MeshCoords.
```

Also I think this step may be suprising to a new user, as in why do we have to remove the existing AuxCoords. It isn't immediately obvious why you can't have both. Maybe adding something like "to avoid coord name collision"?

trexfeathers: 08b58740da3b91129d6d32225dd4273dd54de3a0

lbdreyer (approved):
This great! And I like the addition of the image. They really help to visualise what's going on with the mesh (particularly the SMC grid!)

I'll just wait for the CI to finish then this will be ready to go :+1:

trexfeathers:
Thanks @lbdreyer 🙂

PR #4741, opened 2022-05-12
Gallery: update NEMO example

rcomer:
### Description

* Why use four lines when you can use one?
* It took me a while to understand why the time coordinate promotion was necessary, so expanded that comment.
---

### DIFF
--- a/docs/gallery_code/oceanography/plot_load_nemo.py
+++ b/docs/gallery_code/oceanography/plot_load_nemo.py
@@ -25,9 +25,11 @@ def main():
     # concatenation.
     equalise_attributes(cubes)
 
-    # The cubes still cannot be concatenated because their time dimension is
-    # time_counter, which has the same value for each cube, rather than time,
-    # which varies. time needs to be promoted to allow concatenation.
+    # The cubes still cannot be concatenated because their dimension coordinate
+    # is "time_counter", which has the same value for each cube.  concatenate
+    # needs distinct values in order to create a new DimCoord for the output
+    # cube.  Here, each cube has a "time" auxiliary coordinate, and these do
+    # have distinct values, so we can promote them to allow concatenation.
     for cube in cubes:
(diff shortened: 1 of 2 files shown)


docs/gallery_code/oceanography/plot_load_nemo.py
```diff
@@ -13,24 +13,21 @@
...
-            cube.attributes[attribute] = ""
+    # Some attributes are unique to each file and must be removed to allow
+    # concatenation.
+    equalise_attributes(cubes)
 
     # The cubes still cannot be concatenated because their time dimension is
-    # time_counter rather than time. time needs to be promoted to allow
-    # concatenation.
+    # time_counter, which has the same value for each cube, rather than time,
+    # which varies. time needs to be promoted to allow concatenation.
```
trexfeathers, line 30: I agree this was confusing. But I'm still finding it confusing!

Could you rephrase the whole thing to refer explicitly to `DimCoord`s and `AuxCoord`s? And it also hadn't occurred to me that promotion also demotes `time_counter` (perhaps I should have worked that one out); do you think it's worth mentioning that?

trexfeathers (changes requested):

docs/gallery_code/oceanography/plot_load_nemo.py
rcomer: Words are hard!  How's this? https://github.com/SciTools/iris/pull/4741/commits/848e53f1fcddae27f48f89fa78bfcc837dadda66

Note that, although the "time_counter" coord becomes an aux_coord on the cube, it is still of type `DimCoord` 🤪

rcomer:
Rebased to fix whatsnew conflict

trexfeathers (approved):

trexfeathers:
Nice bit of continuous improvement @rcomer 👍

PR #4755, opened 2022-05-19
Permit fast percentile with mdtol=0

### LINKED ISSUE #4735: Allow fast percentile method on masked arrays if `mdtol=0`
## ✨ Feature Request

Enable the `fast_percentile_method` of the `PERCENTILE` aggregator to work on masked arrays if `mdtol` is set to zero.

## Motivation

`fast_percentile_method` is, well, fast.  So it would be good to be able to use it in a wider range of contexts.  The particular use-case I'm thinking of is where all land points (and no other points) are masked and we aggregate over all the dimensions _except_ the spatial ones.  So for the grid cells we are interested in, we have a full (non-masked) set.

## Additional context

The `fast_percentile_method` of the `PERCENTILE` aggregator uses `numpy.percentile`, which just ignores any mask, so in general we'd get the wrong values out:

```python
import numpy as np
import numpy.ma as ma

arr = ma.array(np.tile(np.arange(6), 2).reshape(2, 6))
arr[0, 4:] = ma.masked

print(arr)
print(np.percentile(arr, 50, axis=1))
```

```
[[0 1 2 3 -- --]
 [0 1 2 3 4 5]]
/net/project/ukmo/scitools/opt_scitools/conda/deployments/default-2022_04_05/lib/python3.8/site-packages/numpy/lib/function_base.py:4650: UserWarning: Warning: 'partition' will ignore the 'mask' of the MaskedArray.
  arr.partition(
[2.5 2.5]
```

So we currently just don't allow it to run on masked arrays:

https://github.com/SciTools/iris/blob/1be3a54fc9efe0659d00989bed8933d7919d0a3a/lib/iris/analysis/__init__.py#L1283-L1286

However, if `mdtol` was set to zero, we would always add a mask over those wrong values in the result:

https://github.com/SciTools/iris/blob/1be3a54fc9efe0659d00989bed8933d7919d0a3a/lib/iris/analysis/__init__.py#L588-L598

or

https://github.com/SciTools/iris/blob/1be3a54fc9efe0659d00989bed8933d7919d0a3a/lib/iris/analysis/__init__.py#L1197

So I think it would be safe to relax the check so that this method can still run if `mdtol` is zero.

wjbenfold:
### Description

Closes #4735

We can't do the full check in the `PercentileAggregator.lazy_aggregate` method because we can't inspect the array to check if it's masked when it's lazy so I've left it where it was but done a check on `mdtol` out there where we know about it.

---

### DIFF
--- a/lib/iris/analysis/__init__.py
+++ b/lib/iris/analysis/__init__.py
@@ -720,6 +720,25 @@ class PercentileAggregator(_Aggregator):
             **kwargs,
         )
 
+    def _base_aggregate(self, data, axis, lazy, **kwargs):
+        """
+        Method to avoid duplication of checks in aggregate and lazy_aggregate.
+        """
+        msg = "{} aggregator requires the mandatory keyword argument {!r}."
+        for arg in self._args:
+            if arg not in kwargs:
+                raise ValueError(msg.format(self.name(), arg))
+
+        if kwargs.get("fast_percentile_method", False) and (
+            kwargs.get("mdtol", 1) != 0
+        ):
+            kwargs["error_on_masked"] = True
+
+        if lazy:
+            return _Aggregator.lazy_aggregate(self, data, axis, **kwargs)
+        else:
+            return _Aggregator.aggregate(self, data, axis, **kwargs)
+
     def aggregate(self, data, axis, **kwargs):
         """
         Perform the percentile aggregation over the given data.
@@ -755,12 +774,7 @@ class PercentileAggregator(_Aggregator):
 
         """
 
-        msg = "{} aggregator requires the mandatory keyword argument {!r}."
-        for arg in self._args:
-            if arg not in kwargs:
-                raise ValueError(msg.format(self.name(), arg))
-
-        return _Aggregator.aggregate(self, data, axis, **kwargs)
+        return self._base_aggregate(data, axis, lazy=False, **kwargs)
 
     def lazy_aggregate(self, data, axis, **kwargs):
         """
@@ -794,12 +808,7 @@ class PercentileAggregator(_Aggregator):
 
         """
 
-        msg = "{} aggregator requires the mandatory keyword argument {!r}."
-        for arg in self._args:
-            if arg not in kwargs:
-                raise ValueError(msg.format(self.name(), arg))
-
-        return _Aggregator.lazy_aggregate(self, data, axis, **kwargs)
+        return self._base_aggregate(data, axis, lazy=True, **kwargs)
 
     def post_process(self, collapsed_cube, data_result, coords, **kwargs):
         """
@@ -1281,9 +1290,13 @@ def _calc_percentile(data, percent, fast_percentile_method=False, **kwargs):
 
     """
     if fast_percentile_method:
-        msg = "Cannot use fast np.percentile method with masked array."
-        if ma.is_masked(data):
-            raise TypeError(msg)
+        if kwargs.pop("error_on_masked", False):
+            msg = (
+                "Cannot use fast np.percentile method with masked array unless"
+                " mdtol is 0."
+            )
+            if ma.is_masked(data):
+                raise TypeError(msg)
         result = np.percentile(data, percent, axis=-1)
         result = result.T
     else:
@@ -1965,7 +1978,8 @@ Additional kwargs associated with the use of this aggregator:
 * fast_percentile_method (boolean):
     When set to True, uses :func:`numpy.percentile` method as a faster
     alternative to the :func:`scipy.stats.mstats.mquantiles` method.  alphap and
-    betap are ignored. An exception is raised if the data are masked.
+    betap are ignored. An exception is raised if the data are masked and the
+    missing data tolerance is not 0.
     Defaults to False.
 
 **For example**:
--- a/lib/iris/analysis/__init__.py
+++ b/lib/iris/analysis/__init__.py
@@ -593,6 +593,12 @@ class _Aggregator:
         ):
             fraction_not_missing = data.count(axis=axis) / data.shape[axis]
             mask_update = 1 - mdtol > fraction_not_missing
+            if result.ndim > mask_update.ndim:
+                # call_func created trailing dimension.
+                mask_update = np.broadcast_to(
+                    mask_update.reshape(mask_update.shape + (1,)),
+                    result.shape,
+                )
             if ma.isMaskedArray(result):
                 result.mask = result.mask | mask_update
             else:
--- a/lib/iris/analysis/__init__.py
+++ b/lib/iris/analysis/__init__.py
@@ -592,8 +592,8 @@ class _Aggregator:
             and result is not ma.masked
         ):
             fraction_not_missing = data.count(axis=axis) / data.shape[axis]
-            mask_update = 1 - mdtol > fraction_not_missing
-            if np.array(result).ndim > np.array(mask_update).ndim:
+            mask_update = np.array(1 - mdtol > fraction_not_missing)
+            if np.array(result).ndim > mask_update.ndim:
                 # call_func created trailing dimension.
                 mask_update = np.broadcast_to(
                     mask_update.reshape(mask_update.shape + (1,)),
--- a/lib/iris/analysis/__init__.py
+++ b/lib/iris/analysis/__init__.py
@@ -593,11 +593,11 @@ class _Aggregator:
         ):
             fraction_not_missing = data.count(axis=axis) / data.shape[axis]
             mask_update = 1 - mdtol > fraction_not_missing
-            if result.ndim > mask_update.ndim:
+            if np.array(result).ndim > np.array(mask_update).ndim:
                 # call_func created trailing dimension.
                 mask_update = np.broadcast_to(
                     mask_update.reshape(mask_update.shape + (1,)),
-                    result.shape,
+                    np.array(result).shape,
                 )
             if ma.isMaskedArray(result):
                 result.mask = result.mask | mask_update
--- a/lib/iris/analysis/__init__.py
+++ b/lib/iris/analysis/__init__.py
@@ -39,6 +39,7 @@ from collections import OrderedDict
 from collections.abc import Iterable
 import functools
 from functools import wraps
+import warnings
 
 import dask.array as da
 import numpy as np
@@ -1297,7 +1298,12 @@ def _calc_percentile(data, percent, fast_percentile_method=False, **kwargs):
             )
             if ma.is_masked(data):
                 raise TypeError(msg)
-        result = np.percentile(data, percent, axis=-1)
+        with warnings.catch_warnings():
+            warnings.filterwarnings(
+                "ignore",
+                "Warning: 'partition' will ignore the 'mask' of the MaskedArray.",
+            )
+            result = np.percentile(data, percent, axis=-1)
         result = result.T
     else:
         quantiles = percent / 100.0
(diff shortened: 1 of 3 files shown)


lib/iris/tests/unit/analysis/test_PERCENTILE.py
```diff
@@ -208,12 +208,25 @@ def test_masked(self):
...
 
+    def test_masked_mdtol_0(self):
+        shape = (2, 11)
+        axis = 0
+        percent = 50
+        data = ma.arange(np.prod(shape)).reshape(shape)
+        data[0, ::2] = ma.masked
+        expected = np.arange(shape[-1]) + 5.5
+        expected[0] = ma.masked
+        self.check_percentile_calc(data, axis, percent, expected, mdtol=0)
```
rcomer, line 228: I think we need to do something cleverer here as `check_percentile_calc` uses `assertArrayEqual`, which ignores masks.  There is also [assertMaskedArrayEqual](https://github.com/SciTools/iris/blob/caae45075795c1e13fbb0f04488c3b99f15b60e8/lib/iris/tests/__init__.py#L686) which could be what we want.  I made a mini test script to check

```python
import iris.tests as tests

import numpy as np
import numpy.ma as ma

class TestDummy(tests.IrisTest):
    def setUp(self):
        self.arr1 = np.arange(3)
        self.arr2 = ma.arange(3)
        self.arr2[0] = ma.masked
        
    def test_equal(self):
        self.assertArrayEqual(self.arr1, self.arr2)
        
    def test_masked_equal(self):
        self.assertMaskedArrayEqual(self.arr1, self.arr2)
```

`pytest` gave me
```
FAILED array_equal_check.py::TestDummy::test_masked_equal - AssertionError: 
========================= 1 failed, 1 passed in 2.20s ==========================
```

lib/iris/tests/unit/analysis/test_PERCENTILE.py
```diff
@@ -208,12 +208,25 @@ def test_masked(self):
...
                 data, axis=0, percent=50, fast_percentile_method=True
             )
 
+    def test_masked_mdtol_0(self):
+        shape = (2, 11)
+        axis = 0
+        percent = 50
+        data = ma.arange(np.prod(shape)).reshape(shape)
+        data[0, ::2] = ma.masked
+        expected = np.arange(shape[-1]) + 5.5
```
rcomer, line 226:
```suggestion
        expected = ma.arange(shape[-1]) + 5.5
```

lib/iris/tests/unit/analysis/test_PERCENTILE.py
```diff
@@ -208,12 +208,25 @@ def test_masked(self):
...
+            "Cannot use fast np.percentile method with masked array unless "
+            "mdtol is 0."
+        )
         with self.assertRaisesRegex(TypeError, emsg):
             PERCENTILE.aggregate(
                 data, axis=0, percent=50, fast_percentile_method=True
             )
 
+    def test_masked_mdtol_0(self):
+        shape = (2, 11)
```
rcomer, line 221:
```suggestion
        shape = (3, 11)
```
I found that, with shape (2, 11), `np.percentile` returns a masked array anyway but doesn't with shape (3, 11).  Not sure why that should be.

rcomer (changes requested):
Hi @wjbenfold, thanks for this.  Looks good apart from some queries in the tests.

Also, I wonder if we want to suppress the warning that numpy raises saying that the mask is ignored?  See the example output in #4735

lib/iris/tests/unit/analysis/test_PERCENTILE.py
wjbenfold: Fixed (which also fixed the warnings I didn't understand about it ignoring this)

lib/iris/tests/unit/analysis/test_PERCENTILE.py
wjbenfold: Done (though I now have 4 test failures)

lib/iris/tests/unit/analysis/test_PERCENTILE.py
```diff
@@ -52,9 +52,9 @@ def check_percentile_calc(
             self.assertFalse(is_lazy)
 
         if approx:
-            self.assertArrayAlmostEqual(actual, expected)
+            self.assertMaskedArrayAlmostEqual(actual, expected)
         else:
-            self.assertArrayEqual(actual, expected)
+            self.assertMaskedArrayEqual(actual, expected)
```
wjbenfold, line 57: Adding this has caused test failures (e.g. `FAIL: test_masked_2d_multi (unit.analysis.test_PERCENTILE.Test_aggregate)`) @rcomer 

I think they're legitimate (see line 131) because we've got an `mdtol` of 0.1 that is exceeded by the 0.33 that we're sending in, so we should just be getting a masked array back.

What was your intention on this test when you were in this code previously? Should it just return a masked array, or do we want to check that the results come back right?

lib/iris/tests/unit/analysis/test_PERCENTILE.py
wjbenfold: I'm not sure I understand? Is it that I'm not doing a sufficient test of my code masking the data because `np.percentile` will do it anyway for a (2, 11) so I need to use a (3, 11) to make sure my code is doing it?

lib/iris/tests/unit/analysis/test_PERCENTILE.py
wjbenfold: I went for the best of both worlds, by masking half the columns then checking the array was part masked.

Imagine my surprise when one of those tests is now failing because when we do the lazy version, masking half the input points masks half the percentages in the output...

rcomer: I don't actually have any memory of this!  It looks like I did it deliberately when modifying the lazy mdtol handling: https://github.com/rcomer/iris/commit/46397cbd5c6e990d7f9cee97a8049fc9c5b41c00  So I really should have noticed it wasn't having the desired effect 😔

Maybe there should be two versions of this - with and without the `mdtol` keyword?

lib/iris/tests/unit/analysis/test_PERCENTILE.py
rcomer: Yes, that's what I meant.  Sorry to be ambiguous.

lib/iris/tests/unit/analysis/test_PERCENTILE.py
wjbenfold: > I really should have noticed it wasn't having the desired effect

We only started getting this failure since I added the `assertMaskedArrayEqual`, which you spotted as reviewer here when I missed it reviewing yours...

rcomer: > > I really should have noticed it wasn't having the desired effect
> 
> We only started getting this failure since I added the `assertMaskedArrayEqual`, which you spotted as reviewer here when I missed it reviewing yours...

It's all fine as long as noone else notices.  Good thing we're not doing all this development in public...

wjbenfold: Note to Monday @wjbenfold - line 594-599 of `iris/analysis/__init__.py` is decidedly suspect

rcomer: OK I just checked out f41fbbb37c54a093b8c7ca8ad698fb7a58ac3a26, which is the last commit on main before #3901.  I changed that version of the test to

```python
    def test_masked_2d_multi(self):
        shape = (3, 10)
        data = ma.arange(np.prod(shape)).reshape(shape)
        data[1, ::2] = ma.masked
        percent = np.array([10, 50, 70, 80])
        actual = PERCENTILE.aggregate(data, axis=0, percent=percent, mdtol=0.1)
        self.assertTupleEqual(actual.shape, (shape[-1], percent.size))
        expected = np.tile(ma.arange(shape[-1]), percent.size)
        expected = expected.reshape(percent.size, shape[-1]).T
        expected = expected + (percent / 10 * 2)
        expected[::2] = ma.masked
        self.assertMaskedArrayAlmostEqual(actual, expected)
```

which I think is equivalent to what you now have here.  It failed comparing the masks.  So this seems to have been broken before we ever touched it.

wjbenfold: https://github.com/SciTools/iris/pull/4755/commits/2a42c1891c05313ab07f705317dd30b1c12301fa

lib/iris/tests/unit/analysis/test_PERCENTILE.py
wjbenfold: https://github.com/SciTools/iris/pull/4755/commits/34719085840786f74976468bec790bd6c5f76437

lib/iris/analysis/__init__.py
```diff
@@ -592,6 +593,12 @@ def aggregate(self, data, axis, **kwargs):
         ):
             fraction_not_missing = data.count(axis=axis) / data.shape[axis]
             mask_update = 1 - mdtol > fraction_not_missing
+            if np.array(result).ndim > np.array(mask_update).ndim:
```
rcomer, line 596:
```suggestion
            mask_update = np.array(1 - mdtol > fraction_not_missing)
            if np.array(result).ndim > mask_update.ndim:
```

Making `mask_update` an array will ensure it also has `reshape` and `shape` used below.

lib/iris/tests/unit/analysis/test_PERCENTILE.py
```diff
@@ -293,6 +297,8 @@ def setUp(self):
         self.agg_method = PERCENTILE.lazy_aggregate
 
     def test_masked(self):
+        # Using (3,11) because np.percentile returns a masked array anyway with
+        # (2, 11)
         shape = (2, 11)
```
rcomer, line 302: Comment doesn't match `shape` here.  Since we're just checking that an exception is raised it probably doesn't matter whether it's (3, 11) or (2, 11).

lib/iris/tests/unit/analysis/test_PERCENTILE.py
```diff
@@ -307,14 +314,14 @@ def test_masked(self):
             as_concrete_data(actual)
 
     def test_masked_mdtol_0(self):
-        shape = (2, 11)
+        shape = (3, 11)
```
rcomer, line 317: Mismatched comment above may have been intended for here.

wjbenfold: https://github.com/SciTools/iris/pull/4755/commits/7af47a14fce0bad6269c234ece83c43b5f0d359f

lib/iris/tests/unit/analysis/test_PERCENTILE.py
wjbenfold: https://github.com/SciTools/iris/pull/4755/commits/7af47a14fce0bad6269c234ece83c43b5f0d359f

lib/iris/analysis/__init__.py
wjbenfold: https://github.com/SciTools/iris/pull/4755/commits/7af47a14fce0bad6269c234ece83c43b5f0d359f

rcomer (approved):
Thanks for persevering around the diversion @wjbenfold! 🎉

PR #4758, opened 2022-05-19
CPerf and SPerf Benchmark Fixes

trexfeathers:
### Description

Various small corrections after trying out C/SPerf benchmarks in reality (not just on paper).

* Warnings about especially large test files and their location.
* Use a consistent hashing algorithm `hashlib.sha256()` when preserving file generation arguments in filenames.
  Previous use of `hash()` meant a different file name each time - not what was intended when aiming for file re-use!
* Fewer time steps in SPerf test files.
  * Previous parameters generated files up to `700GB` in size, and over `4TB` total for all files.
  * New parameters use `400GB` total, with the largest file at `90GB`.
* Longer benchmark timeouts to allow more time for test file generation.
* Place SPerf `combine_regions` files in the standard `BENCHMARK_DATA` location - they are sizeable so using the standard temp file space risked blowing allocated space.
* `test.stock.netcdf` file generation uses dask for filling the data variable - allows creating files larger than memory.
* Minor amendments to the information provided by the Nox benchmarks session.
* Improvements to the Nox C/SPerf sessions:
  * Allow individual benchmark failures without the whole run failing - we expect some benchmarks to blow memory on certain machines.
  * No 'round' repetition - required for accuracy when doing commit comparison, but wasteful here with the longer running benchmarks as there is less noise and we don't want the run to take 4x as long when it's already long-running!

No what's new entry as that's already covered by the original entry about benchmarks.

---

### DIFF
--- a/benchmarks/benchmarks/sperf/combine_regions.py
+++ b/benchmarks/benchmarks/sperf/combine_regions.py
@@ -16,20 +16,21 @@ from iris.experimental.ugrid import PARSE_UGRID_ON_LOAD
 from iris.experimental.ugrid.utils import recombine_submeshes
 
 from .. import TrackAddedMemoryAllocation, on_demand_benchmark
-from ..generate_data.ugrid import make_cube_like_2d_cubesphere
+from ..generate_data.ugrid import BENCHMARK_DATA, make_cube_like_2d_cubesphere
 
 
 class Mixin:
     # Characterise time taken + memory-allocated, for various stages of combine
     # operations on cubesphere-like test data.
-    timeout = 180.0
+    timeout = 300.0
     params = [100, 200, 300, 500, 1000, 1668]
     param_names = ["cubesphere_C<N>"]
     # Fix result units for the tracking benchmarks.
     unit = "Mb"
+    temp_save_path = BENCHMARK_DATA / "tmp.nc"
 
     def _parametrised_cache_filename(self, n_cubesphere, content_name):
-        return f"cube_C{n_cubesphere}_{content_name}.nc"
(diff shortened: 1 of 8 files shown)


benchmarks/benchmarks/sperf/load.py
```diff
@@ -18,9 +18,6 @@ def time_load_cube(self, _, __, ___):
 
 @on_demand_benchmark
 class Realise(FileMixin):
-    # The larger files take a long time to realise.
-    timeout = 600.0
-
```
trexfeathers, line 23: This is now superceded by the `timeout` in the `FileMixin` parent class.

benchmarks/benchmarks/sperf/combine_regions.py
```diff
@@ -165,6 +169,9 @@ def recombine(self):
         )
         return result
 
+    def save(self):
+        save(self.recombined_cube, self.temp_save_path)
```
lbdreyer, line 173: I see you have two different function/methods called`save` here. It doesn't look like this is causing a namespace issue but it does make reading a bit more difficult. Maybe rename this method to something like `save_cube` or `save_recombined_cube`?

lbdreyer (commented):
This all looks good to me. I just have one minor comment

benchmarks/benchmarks/sperf/combine_regions.py
trexfeathers: ab467b5f5bd0ab9fbdef39c14d46dba6ff73d60e

lbdreyer (approved):

trexfeathers:
Thanks @lbdreyer!

PR #4765, opened 2022-05-24
Add coverage testing

lbdreyer:
### Description
This PR enables coverage testing using pytest-cov and coverage reporting using codacy.

There are a few different coverage reporting tools we can choose from:
* **coveralls** (used in cartopy),
* **codecov** (used in cf-units, iris-esmg-regrid, nc-time-axis, etc), and
* **codacy** (not currently used although we seem to have it [enabled but not running for cf-units](https://app.codacy.com/gh/SciTools/cf-units/dashboard))

I don't have a particularly strong preference between these tools, but a benefit of codacy (as @bjlittle pointed out) is that it provides extra features such as code quality inspection. It also has security scanning with bandit but that only seems to be available on the paid tiers unfortunately. Another benefit I have found when trialing these tools is that I have had difficulty with the codecov uploader and it would often fail silently, whereas codacy was quite simple to get working.

To implement codacy in this PR only required adding a single line to the .cirrus.yml file. Replacing this with a different coverage reporting tool (e.g. codecov) should be a simple change.

Outside of the changes in this PR, in order to get codacy working, we would also need to do the following:
- Add the Iris repo to codacy (requires admin permissions which I don't have)
- Add the CODACY_PROJECT_TOKEN to the Cirrus-CI settings (I tested this in my branch)

To see what it looks like on my branch you can see the [main dashboard here](https://app.codacy.com/gh/lbdreyer/iris/dashboard?branch=add_cov) and a [report of coverage per files here](https://app.codacy.com/gh/lbdreyer/iris/files?bid=29072991)


lib/iris/tests/runner/_runner.py
```diff
@@ -47,6 +47,7 @@ class TestRunner:
             "The number of processors used for running " "the tests.",
         ),
         ("create-missing", "m", "Create missing test result files."),
+        ("coverage", "l", "Enable coverage testing"),
```
lbdreyer, line 50: We're running out of letters! I chose "l" for "line coverage" as we already use c to mean coding tests

noxfile.py
```diff
@@ -222,13 +224,15 @@ def tests(session: nox.sessions.Session):
     prepare_venv(session)
     session.install("--no-deps", "--editable", ".")
     session.env.update(ENV)
-    session.run(
+    run_args = [
         "python",
         "-m",
         "iris.tests.runner",
         "--default-tests",
-        "--system-tests",
```
lbdreyer, line 230: I've dropped running the system tests because they already get run as part of "default-tests" so there doesn't seem to be much sense in running them twice.

trexfeathers (changes requested):
Thanks @lbdreyer! Codacy looks informative and worth a try.

A few minor points for you.

---

This definitely deserves a What's New entry. And while you're at it please could you do one for #4734 because we missed that one.

>* Add the CODACY_PROJECT_TOKEN to the Cirrus-CI settings (I tested this in my branch)

Assuming you mean via the Cirrus GUI, rather than having the token raw in a repo file?

>* Add the Iris repo to codacy (requires admin permissions which I don't have)

@bjlittle / @jamesp can you take care of this point? Thanks

bouweandela:
This would be a really nice feature to have!

In my experience, Codacy is good for linting pull requests (we've been using it for several years in the ESMValCore project now, see [here](https://app.codacy.com/gh/ESMValGroup/ESMValCore/dashboard)), but for code coverage, Codecov has been more useful to us because it makes the coverage and coverage changes much more visible by writing a comment on the pull request ([example](https://github.com/ESMValGroup/ESMValCore/pull/1750#issuecomment-1269663163)). Since we started using Codecov a bit over a year ago, the amount of code without coverage we have has halved ([see here](https://app.codecov.io/gh/ESMValGroup/ESMValCore?trend=all%20time)).

bouweandela:
Great to see some action again on this pull request!

lbdreyer:
> This would be a really nice feature to have!
> 
> In my experience, Codacy is good for linting pull requests (we've been using it for several years in the ESMValCore project now, see [here](https://app.codacy.com/gh/ESMValGroup/ESMValCore/dashboard)), but for code coverage, Codecov has been more useful to us because it makes the coverage and coverage changes much more visible by writing a comment on the pull request ([example](https://github.com/ESMValGroup/ESMValCore/pull/1750#issuecomment-1269663163)). Since we started using Codecov a bit over a year ago, the amount of code without coverage we have has halved ([see here](https://app.codecov.io/gh/ESMValGroup/ESMValCore?trend=all%20time)).

Thanks @bouweandela that is helpful to know about your experiences with codacy! 

With that in mind I propose we use codecov and if at a later date we wish to explore codacy a bit more, we can look at implementing that. But for now codecov gives us what we want: coverage reporting on the PR

noxfile.py
```diff
@@ -211,7 +212,8 @@ def precommit(session: nox.sessions.Session):
 @nox.session(python=PY_VER, venv_backend="conda")
 def tests(session: nox.sessions.Session):
     """
-    Perform iris system, integration and unit tests.
+    Perform iris system, integration and unit tests with coverage testing
+    enabled.
```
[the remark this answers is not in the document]
lbdreyer: Updated the doc string now to mention COVERAGE being set

pyproject.toml
```diff
@@ -37,3 +37,20 @@ extend_skip = [
...
+]
+omit = [
+    "lib/iris/tests/*",
+    "lib/iris/etc/*",
+]
+
+[tool.coverage.report]
+exclude_lines = [
+    "pragma: no cover",
+    "def __repr__",
```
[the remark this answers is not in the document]
lbdreyer: It's a good question. If I remember correctly, I had copied the cf-units config. I have also seen `__repr__ in the excludes section in many places, even in coverage's docs: https://coverage.readthedocs.io/en/6.4.4/excluding.html#advanced-exclusion
I had a look at our testing and it looks like we are testing __repr__ in quite a few places so it is probably a good idea to include in our coverage.
We can always remove it at a later date if we change our minds

lib/iris/tests/runner/_runner.py
lbdreyer: I've gone even more extreme: I've removed coding-tests as an option from the test runner. I don't really see the point in keeping it when you could just run the file directly
```
pytest test_coding_standards.py
```

I suppose it is there as a convenience, but I doubt it's being used at the moment?

noxfile.py
```diff
@@ -28,6 +28,7 @@
 
 #: Cirrus-CI environment variable hook.
 PY_VER = os.environ.get("PY_VER", _PY_VERSIONS_ALL)
+COVERAGE = os.environ.get("COVERAGE", False)
```
trexfeathers, line 31: We have discussed offline using [Nox `posargs`](https://nox.thea.codes/en/stable/config.html#passing-arguments-into-sessions) (also available with Tox, when we get there) rather than an environment variable.

trexfeathers (changes requested):
Everything looks good, thanks @lbdreyer! But I know there's one more thing you want to do.

noxfile.py
trexfeathers: @lbdreyer I've seen bea9a6837d4adabd60ba6ae10a8257c7daa41517, but the What's New entry also needs a minor change too

lbdreyer: Good point! I will update that too!

trexfeathers (approved):

lbdreyer:
Thanks @trexfeathers !

bouweandela:
Very nice! Now I can see which of the lines are covered by tests: https://github.com/SciTools/iris/pull/5015#issuecomment-1442499808. That will make it much easier to understand if I can assume everything is fine or if I need to write extra tests when contributing.

PR #4773, opened 2022-05-27
`PolarStereographic` implementation

### LINKED ISSUE #4770: Add Polar Stereographic CRS
## ✨ Feature Request
Allow Iris to correctly load data on a polar stereographic grid. 

## Motivation
We have acces to a high resolution digital terrain model of the south pole which we wish to convert to the UM input files for orography. We cannot use this data at the moment, as ANTS relies on Iris being able to load the data. Regridding to say a lat-lon grid is not appropriate as that either has very large numbers of points close together at the pole or we lose a lot of information at lower latitudes

## Additional context
<details>

Sample output from ncdump
```
netcdf REMA_1km_dem_filled {
dimensions:
	x = 5500 ;
	y = 4500 ;
variables:
	char polar_stereographic ;
		polar_stereographic:grid_mapping_name = "polar_stereographic" ;
		polar_stereographic:straight_vertical_longitude_from_pole = 0. ;
		polar_stereographic:false_easting = 0. ;
		polar_stereographic:false_northing = 0. ;
		polar_stereographic:latitude_of_projection_origin = -90. ;
		polar_stereographic:standard_parallel = -71. ;
		polar_stereographic:long_name = "CRS definition" ;
		polar_stereographic:longitude_of_prime_meridian = 0. ;
		polar_stereographic:semi_major_axis = 6378137. ;
		polar_stereographic:inverse_flattening = 298.257223563 ;
		polar_stereographic:spatial_ref = "PROJCS[\"unnamed\",GEOGCS[\"WGS 84\",DATUM[\"WGS_1984\",SPHEROID[\"WGS 84\",6378137,298.257223563,AUTHORITY[\"EPSG\",\"7030\"]],AUTHORITY[\"EPSG\",\"6326\"]],PRIMEM[\"Greenwich\",0],UNIT[\"degree\",0.0174532925199433],AUTHORITY[\"EPSG\",\"4326\"]],PROJECTION[\"Polar_Stereographic\"],PARAMETER[\"latitude_of_origin\",-71],PARAMETER[\"central_meridian\",0],PARAMETER[\"scale_factor\",1],PARAMETER[\"false_easting\",0],PARAMETER[\"false_northing\",0],UNIT[\"metre\",1,AUTHORITY[\"EPSG\",\"9001\"]]]" ;
		polar_stereographic:GeoTransform = "-2700000 1000 0 2300000 0 -1000 " ;
	double x(x) ;
		x:standard_name = "projection_x_coordinate" ;
		x:long_name = "x coordinate of projection" ;
		x:units = "m" ;
	double y(y) ;
		y:standard_name = "projection_y_coordinate" ;
		y:long_name = "y coordinate of projection" ;
		y:units = "m" ;
	float Band1(y, x) ;
		Band1:long_name = "GDAL Band Number 1" ;
		Band1:_FillValue = -9999.f ;
		Band1:grid_mapping = "polar_stereographic" ;
```
</details>

wjbenfold:
### Description

Closes #4770

Doesn't use `ccrs.NorthPolarStereo` or `ccrs.SouthPolarStereo` as they don't support `false_easting` and `false_northing`. Instead inherits from `iris.coord_systems.Stereographic` and thus is based on `ccrs.Stereographic`.

To do:
- ~Update test data version once test data merged~
- ~What's new~

---


pp-mo (changes requested):
Very comprehensive on the testing !
I think this is mostly good to go, but a couple of points to consider ...

lib/iris/coord_systems.py
```diff
@@ -1135,13 +1157,81 @@ def as_cartopy_crs(self):
...
+        scale_factor_at_projection_origin : float, optional
+            Scale factor at the origin of the projection
+
+        ellipsoid : :class:`GeogCS`, optional
+            If given, defines the ellipsoid.
+
+        Notes
+        -----
+        It is only valid to provide one of `true_scale_lat` and `scale_factor_at_projection_origin`.
+        TODO: What if you give neither?
```
wjbenfold, line 1216: Do you have an opinion on this @pp-mo? Can assume scale factor is 1, or refuse?

lib/iris/fileformats/netcdf.py
```diff
@@ -2685,6 +2685,28 @@ def add_ellipsoid(ellipsoid):
...
+                            cs.scale_factor_at_projection_origin
+                        )
+                    else:
+                        cf_var_grid.scale_factor_at_projection_origin = (
+                            1.0  # TODO: Is this right?
+                        )
+
                 # stereo
                 elif isinstance(cs, iris.coord_systems.Stereographic):
                     if cs.true_scale_lat is not None:
```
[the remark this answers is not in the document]
wjbenfold: Oops, missed that. Will do

lib/iris/fileformats/netcdf.py
```diff
@@ -2685,6 +2685,28 @@ def add_ellipsoid(ellipsoid):
...
+                    # Only one of these should be set
+                    if cs.true_scale_lat is not None:
+                        cf_var_grid.true_scale_lat = cs.true_scale_lat
+                    elif cs.scale_factor_at_projection_origin is not None:
+                        cf_var_grid.scale_factor_at_projection_origin = (
+                            cs.scale_factor_at_projection_origin
+                        )
+                    else:
+                        cf_var_grid.scale_factor_at_projection_origin = (
+                            1.0  # TODO: Is this right?
```
wjbenfold, line 2707: Also this @pp-mo

lib/iris/coord_systems.py
pp-mo: Before we added the new "scale_factor" keyword, the parent class supported only `true_scale_lat`, which was strictly optional.  So that surely means that it should now still be correct to set _neither_, which would have the same meaning.
And it makes sense to me that PolarStereo should do just what Stereo does in this respect.
I think it is **_not_** correct to default a scale-factor to 1.0 --  the correct default is 'None' :  Then, what that _actually_ does, or what number it is equivalent to, is up to the lower-level calls (i.e. cartopy or proj).

lib/iris/fileformats/netcdf.py
pp-mo: Hmm.  In this case, [the CF spec](https://cfconventions.org/Data/cf-conventions/cf-conventions-1.9/cf-conventions.html#polar-stereographic) does insist on having _one_ of scale-factor / standard-parallel.  
So I guess scale-factor is easiest, and 1.0 make some kind of sense.

lib/iris/fileformats/netcdf.py
```diff
@@ -2685,28 +2685,47 @@ def add_ellipsoid(ellipsoid):
...
+                    cf_var_grid.false_easting = cs.false_easting
+                    cf_var_grid.false_northing = cs.false_northing
+                    # Only one of these should be set
                     if cs.true_scale_lat is not None:
-                        warnings.warn(
-                            "Stereographic coordinate systems with "
-                            "true scale latitude specified are not "
-                            "yet handled"
+                        cf_var_grid.true_scale_lat = cs.true_scale_lat
+                    elif cs.scale_factor_at_projection_origin is not None:
```
pp-mo, line 2721: This is a bit tricky...
According to their CF descriptions, whereas a [PolarStereographic grid-mapping](https://cfconventions.org/Data/cf-conventions/cf-conventions-1.9/cf-conventions.html#polar-stereographic) can have _either_ a 'standard_parallel'  or 'scale_factor_at_projection_origin' attribute, the plain [Stereographic type](https://cfconventions.org/Data/cf-conventions/cf-conventions-1.9/cf-conventions.html#_stereographic) _only_ supports the 'scale_factor_at_projection_origin' one.

So, if we get a Stereographic with a standard-parallel, we can either try to _convert_ it to a scale-factor value, or we could just refuse to save it.  
In theory this can't happen with one loaded from CF, so  I think it's probably unlikely.  Rather than take responsibility for a conversion, my choice would be to error it.

Likewise, according to CF, a Stereographic grid-mapping _must_ have a scale factor, and a PolarStereo _must_ have either a scale-factor or a standard-parallel.
In those cases, if the coord-system has neither set, I think it is probably OK and more useful to set a scale-factor of 1.

lib/iris/fileformats/netcdf.py
```diff
@@ -2685,28 +2685,47 @@ def add_ellipsoid(ellipsoid):
...
                     else:
-                        if cs.ellipsoid:
-                            add_ellipsoid(cs.ellipsoid)
-                        cf_var_grid.longitude_of_projection_origin = (
-                            cs.central_lon
-                        )
-                        cf_var_grid.latitude_of_projection_origin = (
-                            cs.central_lat
+                        cf_var_grid.scale_factor_at_projection_origin = (
+                            1.0  # TODO: Is this right?
```
pp-mo, line 2727: Yes, this is what we were doing before your changes, and I think is OK.

pp-mo (changes requested):
Still some things to tweak, I think.
OWOK !

lib/iris/fileformats/netcdf.py
wjbenfold: Now raising a `ValueError` as that seems consistent with the nearby code. Could change that though

pp-mo (approved):

PR #4784, opened 2022-06-06
Fix whatsnew

wjbenfold:
### Description

The Iris 3.0 whatsnew referred explicitly to a renamed page. I've also reverted an accidental change to the permissions I made to the whatsnew file in my previous PR

---


wjbenfold:
I'm now wondering whether I should be using _iris_whatsnew instead of the full URL so that it doesn't break again in future, but that's already referenced in the whatsnew message so seems pointless repeating?

wjbenfold:
New docs build here: https://www-avd/~wbenfold/iris/whatsnew/latest.html

I think the linkcheck will be broken until this fix is merged, so we'll have to merge with that failing

docs/src/developers_guide/release.rst
```diff
@@ -258,7 +258,7 @@ Post Release Steps
    and/or ``Hidden``. To do this ``Edit`` the appropriate version e.g.,
    see `Editing v3.0.0rc0`_ (must be logged into Read the Docs).
 #. Make a new ``latest.rst`` from ``latest.rst.template`` and update the include
-   statement and the toctree in ``whatsnew.rst`` to point at the new
+   statement and the toctree in ``index.rst`` to point at the new
```
trexfeathers, line 261: Also line `224`

trexfeathers (approved):

trexfeathers:
Cirrus is asleep, but I'm confident that the `linux: py3.8 tests` result will be unchanged. Merging.

PR #4785, opened 2022-06-08
Stereographic load with scale factor

### LINKED ISSUE #3845: Netcdf load ignores 'scale_factor_at_projection_origin' of Stereographic grib-mapping
The CF conventions state that ['scale_factor_at_projection_origin' is a [required attribute for Stereographic grid-mapping](https://cfconventions.org/cf-conventions/cf-conventions.html#_stereographic)
But our code is [completely ignoring this](https://github.com/SciTools/iris/blob/a338ca530d5220330f136e58a70d502f62f52e6a/lib/iris/fileformats/_pyke_rules/fc_rules_cf.krb#L1479).  
( Note that the comment ["this is checked elsewhere"](https://github.com/SciTools/iris/blob/a338ca530d5220330f136e58a70d502f62f52e6a/lib/iris/fileformats/_pyke_rules/fc_rules_cf.krb#L1496) is totally untrue !  It couldn't be really, as this is the only place that we will inspect the file variable )

By ignoring this, we are mis-representing any input where this is different from 1.0.

Suggested action:
**We should either support this, or at least check it and raise an error/warning when our loaded version is incorrect.**

Note also:  when saving, we currently  [warn that the result is not exact + don't save anything](https://github.com/SciTools/iris/blob/a338ca530d5220330f136e58a70d502f62f52e6a/lib/iris/fileformats/netcdf.py#L2130).
We could fix that too, but at least it is not "wrong".

wjbenfold:
### Description

Closes #3845

---

### DIFF
--- a/lib/iris/fileformats/_nc_load_rules/helpers.py
+++ b/lib/iris/fileformats/_nc_load_rules/helpers.py
@@ -417,6 +417,10 @@ def build_stereographic_coordinate_system(engine, cf_grid_var):
     longitude_of_projection_origin = getattr(
         cf_grid_var, CF_ATTR_GRID_LON_OF_PROJ_ORIGIN, None
     )
+    scale_factor_at_projection_origin = getattr(
+        cf_grid_var, CF_ATTR_GRID_SCALE_FACTOR_AT_PROJ_ORIGIN, None
+    )
+
     false_easting = getattr(cf_grid_var, CF_ATTR_GRID_FALSE_EASTING, None)
     false_northing = getattr(cf_grid_var, CF_ATTR_GRID_FALSE_NORTHING, None)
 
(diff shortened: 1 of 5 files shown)


pp-mo (changes requested):
This is good stuff.

But unfortunately I think I missed something from the previous #4773, which is to test how the scale_factor is *saved* for a plain (i.e. not Polar) Stereo case.

I know it's not what the title here says, but I think it's important that we test saving it too.
Could you possibly add a test for the save (into `iris.tests.unit.fileformats.netcdf.test_Saver.Test_write`, 
(i.e. saving a scale-factor) 
along the lines of the existing `test_stereographic` and `test_stereographic_no_ellipsoid` already there ?
A round-trip wouldn't be a bad idea, either.  Not sure where that would live though.

pp-mo:
P.S. I was just starting to look into doing this -- but now you got there first !

wjbenfold:
We don't seem to have a roundtrip on any other coordinate system, so think I'm just happy adding a save test ([4c37df2](https://github.com/SciTools/iris/pull/4785/commits/4c37df20811f8e56852d4d72a080dc2f8a7e54dc))

PR #4791, opened 2022-06-09
Allow passing of method keyword to numpy.percentile

rcomer:
### Description

The `PERCENTILE` fast method will be more versatile if we can choose our interpolation method.

---

### DIFF
--- a/lib/iris/analysis/__init__.py
+++ b/lib/iris/analysis/__init__.py
@@ -1296,10 +1296,6 @@ def _calc_percentile(data, percent, fast_percentile_method=False, **kwargs):
     Calculate percentiles along the trailing axis of a 1D or 2D array.
 
     """
-    # Get hold of method keyword to pass it to numpy.percentile, or just remove it
-    # from kwargs so it's ignored for scipy mquantiles.  Note that kwargs always
-    # contains alphap and betap so we can't pass it to numpy.percentile.
-    method = kwargs.pop("method", None)
     if fast_percentile_method:
         if kwargs.pop("error_on_masked", False):
             msg = (
@@ -1313,15 +1309,13 @@ def _calc_percentile(data, percent, fast_percentile_method=False, **kwargs):
                 "ignore",
                 "Warning: 'partition' will ignore the 'mask' of the MaskedArray.",
             )
-            pc_kwargs = dict(axis=-1)
-            if method is not None:
-                pc_kwargs["method"] = method
-
-            result = np.percentile(data, percent, **pc_kwargs)
+            result = np.percentile(data, percent, axis=-1, **kwargs)
 
         result = result.T
     else:
         quantiles = percent / 100.0
+        for key in ["alphap", "betap"]:
+            kwargs.setdefault(key, 1)
         result = scipy.stats.mstats.mquantiles(
             data, quantiles, axis=-1, **kwargs
         )
@@ -1353,9 +1347,9 @@ def _percentile(data, percent, fast_percentile_method=False, **kwargs):
         alternative to the scipy.mstats.mquantiles method. Does not handle
         masked arrays.
 
-    **kwargs
+    **kwargs : dict, optional
         passed to scipy.stats.mstats.mquantiles if fast_percentile_method is
-        False.  Otherwise "method" is extracted and passed to numpy.percentile.
+        False.  Otherwise passed to numpy.percentile.
 
     """
     if not isinstance(percent, Iterable):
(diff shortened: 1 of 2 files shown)


rcomer:
@wjbenfold I promise that after this I am done messing with `PERCENTILE`!

wjbenfold:
Does it only make sense to allow passing "method"? It seems like it would be more general to pass all kwargs through, though I don't know enough about what the others do to tell if allowing that would only be letting people shoot themselves in the foot

rcomer:
@wjbenfold that’s a fair point and at the moment I think there is nothing to stop a user passing e.g. `limit` to the scipy function, even though we haven’t documented that you can.  I did wonder if it would be better to just pass all kwargs down and let the numpy/scipy function raise an error if they get the wrong ones.  The current approach of silently ignoring wrong kwargs could lead to more confusion for the user.

There was some relevant discussion at #3294, but it was somewhat inconclusive.

lib/iris/analysis/__init__.py
```diff
@@ -1296,6 +1296,10 @@ def _calc_percentile(data, percent, fast_percentile_method=False, **kwargs):
     Calculate percentiles along the trailing axis of a 1D or 2D array.
 
     """
+    # Get hold of method keyword to pass it to numpy.percentile, or just remove it
+    # from kwargs so it's ignored for scipy mquantiles.  Note that kwargs always
+    # contains alphap and betap so we can't pass it to numpy.percentile.
+    method = kwargs.pop("method", "linear")
```
wjbenfold, line 1302: Thinking about it, this couples us to `np.percentile`'s default - can we instead only set `method` in the `np.percentile` call if it's user specified and if not just not specify it? Means our wrapper is thinner, which seems like a good direction to go in if it's not too faffy

lib/iris/analysis/__init__.py
```diff
@@ -1309,7 +1313,12 @@ def _calc_percentile(data, percent, fast_percentile_method=False, **kwargs):
                 "ignore",
                 "Warning: 'partition' will ignore the 'mask' of the MaskedArray.",
             )
-            result = np.percentile(data, percent, axis=-1)
+            pc_kwargs = dict(axis=-1)
+            if method is not None:
+                pc_kwargs["method"] = method
```
rcomer, line 1318: This feels kinda ugly.  Is there a better way?

wjbenfold: I didn't think so, then I started wondering what happens if `scipy`'s `mquantiles` starts accepting a method argument, and now I think it might be better to pop `alphap` and `betap` out of the kwargs then pass the remainder wholesale to `np.percentile`?

wjbenfold: Ooh, that seems nice

lib/iris/analysis/__init__.py
```diff
@@ -1976,23 +1979,25 @@ def interp_order(length):
...
     When set to True, uses :func:`numpy.percentile` method as a faster
-    alternative to the :func:`scipy.stats.mstats.mquantiles` method.  alphap and
-    betap are ignored. An exception is raised if the data are masked and the
-    missing data tolerance is not 0.
-    Defaults to False.
+    alternative to the :func:`scipy.stats.mstats.mquantiles` method.  An
+    exception is raised if the data are masked and the missing data tolerance
+    is not 0.  Defaults to False.
+
+**kwargs : dict, optional
```
wjbenfold, line 1999: I think this should just be `kwargs` - that seems to be the complaint from the doctests

lib/iris/tests/unit/analysis/test_PERCENTILE.py
```diff
@@ -196,6 +196,13 @@ def test_missing_mandatory_kwarg(self):
         with self.assertRaisesRegex(ValueError, emsg):
             PERCENTILE.aggregate("dummy", axis=0)
 
+    def test_wrong_kwarg(self):
```
wjbenfold, line 199: This seems like a common user error - do we want a more informative message or are we happy that the docstring will help people enough?

wjbenfold (changes requested):

lib/iris/analysis/__init__.py
rcomer: Yep, local check confirms that.  I thought I was following numpy docstring style though 😕 
https://numpydoc.readthedocs.io/en/latest/format.html#parameters

lib/iris/tests/unit/analysis/test_PERCENTILE.py
rcomer: The full message is "mquantiles() got an unexpected keyword argument 'method'" but for some reason the assertion failed with a bad match when I pasted the lot in 😕

wjbenfold: Probably the brackets needing some form of escaping, seems fine as-is though

lib/iris/analysis/__init__.py
wjbenfold: fyi @tkknight

wjbenfold:
Happy despite failing test (see #4822)

PR #4792, opened 2022-06-10
pytest gallery tests

rcomer:
### Description

Uses pytest parametrization to remove the duplication from the gallery tests.  We now only need one test file instead of 24 🙂.  To get it working, I pulled the `check_graphic` method off the `GraphicsTest` class so we can call it as a stand-alone function.

Needs https://github.com/SciTools/iris-test-data/pull/79

~In draft for now because~

* ~Probably the context managers could be replaced with pytest fixtures.~ done
* ~I have inserted some deliberate fails into two examples so we can see that the tests work as expected.~ reverted

---


rcomer:
Hopefully @trexfeathers will be proud of me for using `nox --session doctest` to check this locally 😁

rcomer:
Note: 
* [this](https://github.com/SciTools/iris/blob/4fa3f06caf2175b8654fdb4a2bcd83080b7a1418/docs/Makefile#L58-L62) will not work so should either update or remove.
* [this advice](https://scitools-iris.readthedocs.io/en/stable/developers_guide/contributing_documentation_full.html#testing) may need updating (currently the Makefile in docs/src does not have gallerytest).

rcomer:
Possibly should have some version of the setUp and tearDown here.

https://github.com/SciTools/iris/blob/4fa3f06caf2175b8654fdb4a2bcd83080b7a1418/lib/iris/tests/__init__.py#L1220-L1236

rcomer:
> I pulled the `check_graphic` method off the `GraphicsTest` class so we can call it as a stand-alone function.

Huh.  I see @wjbenfold already did that at #4759.

docs/Makefile
```diff
@@ -55,8 +55,3 @@ linkcheck:
 		echo "Running linkcheck in $$i..."; \
 		(cd $$i; $(MAKE) $(MFLAGS) $(MYMAKEFLAGS) linkcheck); done
 
-gallerytest:
```
rcomer, line 58: I am suggesting running the tests directly with `pytest`, because that also allows you to test individual examples (see dev guide change).

rcomer (commented):
Inline comments explain what comes from where.

rcomer:
I had a rethink on the test strategy: previously we created each figure and called `check_graphic` on it before creating the next figure.  Current code creates all the figures for a particular example script and then loops through them to call `check_graphic`.  The new method is less efficient as we are holding more figures in memory and if the first one fails we have created the others unnecessarily.

However, I think the new method is simpler/clearer (relevant commit is https://github.com/SciTools/iris/pull/4792/commits/63cdd66b9419b981284284b78bca3fa7841456ff).  Also if/when we adopt [pytest-mpl](https://github.com/matplotlib/pytest-mpl) I think we would need to generate a sequence of tests that each return a single figure.  The new approach would probably easier to adapt into that.

rcomer:
Ready for review I think.  For examples of how the failure would look if an image changed or if functionality was deprecated, see [the CI for the penultimate commit](https://github.com/SciTools/iris/runs/6935773447?check_suite_focus=true#step:12:1693).

docs/gallery_tests/conftest.py
```diff
@@ -0,0 +1,67 @@
...
+    """
+
+    def no_show():
+        pass
+
+    monkeypatch.setattr(plt, "show", no_show)
+
+    for example_dir in GALLERY_DIR.iterdir():
+        if example_dir.is_dir():
+            monkeypatch.syspath_prepend(example_dir)
```
rcomer, line 49: This does the same job as the `add_gallery_to_path` context manager that was in `gallerytest_util.py`.

rcomer:
@SciTools/iris-devs is this one likely to be reviewed before `pillow` is unpinned?  I'm just wondering whether to fix the `imagerep.json` conflicts now or wait for it to change again.

wjbenfold:
> @SciTools/iris-devs is this one likely to be reviewed before `pillow` is unpinned? I'm just wondering whether to fix the `imagerep.json` conflicts now or wait for it to change again.

We're planning to unpin `pillow` pretty soon(tm), so I'd hang fire :)

rcomer:
Rebased following #4759 and #4826.  `recreate_imagerepo.py` came in very handy :sunglasses:

Needs https://github.com/SciTools/iris-test-data/pull/79

wjbenfold (changes requested):
This is a lot less code and works run locally, looks fundamentally good to me.

When I run with your test data PR and this PR checked out, plus latest `iris-dev` conda env, I get test errors in `plot_SOI_filtering` and `plot_lagged_ensemble` - does that happen for you too and can we fix it? I guess it might need a rebase given it's been around a little while?

rcomer:
Hi @wjbenfold great to “see” you again 😊

Thanks for the review.  I didn’t see any errors when I checked locally but maybe I missed something.  I suspect that #4871 needs prioritising so will wait for that to go in, rebase both branches, and check again.

rcomer:
OK I checked and yes, I do get failures for those examples.  They match the gallery test failures in #4846.  So once I rebase and pick up #4847, all should be well.

I'm kinda surprised the plot_lagged_ensemble didn't fall over in the pp-loading though :confused:

rcomer:
#4871 is abandoned for now, so I've gone ahead and rebased.  Both gallery tests and main tests are passing for me locally.  I've also added a whatsnew entry and bumped the `IRIS_TEST_DATA_VERSION` in anticipation of https://github.com/SciTools/iris-test-data/pull/79.

trexfeathers:
@bjlittle you're assigned to this: is this on your list before holidays or should someone else take a look?

wjbenfold:
When I run the tests I now get

```
lib/iris/__init__.py:104: in <module>
    from ._version import version as __version__  # noqa: F401
E   ModuleNotFoundError: No module named 'iris._version'
```

I'm sufficiently out of the loop to not understand what's broken here - I can't see a file named `_version`?

trexfeathers:
> When I run the tests I now get
> 
> ```
> lib/iris/__init__.py:104: in <module>
>     from ._version import version as __version__  # noqa: F401
> E   ModuleNotFoundError: No module named 'iris._version'
> ```
> 
> I'm sufficiently out of the loop to not understand what's broken here - I can't see a file named `_version`?

@wjbenfold this branch needs #4892

rcomer:
Thanks chaps.  I have rebased again.  Gallery tests all pass for me locally.  In the main tests I see two failures

* the regridding one discussed at #4893
* something to do with a dot file, which I think I've seen before and isn't generally reproducible.

So I think CI should pass once https://github.com/SciTools/iris-test-data/pull/79 is merged.

rcomer:
Rebased again.  Tests all passing for me locally.

wjbenfold (approved):
Tests are all passing :)

rcomer:
Thanks @wjbenfold!  Is there more to do here or can this be merged?

docs/gallery_tests/conftest.py
```diff
@@ -0,0 +1,68 @@
...
+    """
+    orig_sys_path = sys.path
+    sys.path = sys.path[:]
+    sys.path += GALLERY_DIRECTORIES
+    yield
+    sys.path = orig_sys_path
+
+
+@pytest.fixture
+def image_setup_teardown():
```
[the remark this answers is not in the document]
wjbenfold: If the lock experimentally isn't needed locally or in CI, it seems fine without it. The lock in the main image tests used to be able to make them hang if a test errored out so I'm not sad it's not there.

wjbenfold:
> Thanks @wjbenfold! Is there more to do here or can this be merged?

I was just checking through the conversations to check they were resolved. It's now in!

rcomer:
Thanks @wjbenfold! :tada:

trexfeathers:
Thanks so much @rcomer and @wjbenfold. We've been pretty thin on the ground during these summer weeks so your contribution is greatly appreciated 🥰

Issue #4754, opened 2022-05-19
Unify regridders


stephenworsley:
## 📰 Custom Issue

There are a collection of regridders in iris which each work slightly differently, but are similar enough that they could potentially benefit from unifying some of their behaviour and perhaps have them inherit from the same abstract class. This would also benefit regridders built outside of iris such as in iris-esmf-regrid. The current set of iris regridding schemes is:

- `Linear`
- `Nearest`
- `AreaWeighted`
- `UnstructuredNearest`
- `PointInCell`

There is also the deprecated regridder in iris.experimental.regrid_conservative which is functionally replaced by iris-esmf-regrid.

---
The current set of regridders have the following quirks:

### Linear and Nearest
- The `Linear` and `Nearest` regridding schemes create a `RectilinearRegridder` (from analysis._regrid). This uses `_RegularGridInterpolator` (from analysis._scipy_interpolate) to calculate weights and then use those weights, this derives from scipy and is under a scipy copyright, though the code has developed significantly since it was copied from scipy.
- Source and target grid must be lat/lon DimCoords.
- Weights are not cached on `RectilinearRegridder`, though `_RegularGridInterpolator` does offer weight caching.
- Weights matrix is a sparse csr_matrix for Linear, for Nearest, weights take the form of an index rather than a matrix.
- Loops through 2D slices for calculations (rather than using transposition and reshaping like in iris-esmf-regrid).
- Has `interpolator` method.
- Has powerful `_create_cube` method which handles derived coordinates. Even handles derived coordinates covering the grid dimension via the use of a regridder function callback.
- Lazy regridding is supported via `map_complete_blocks`.
- Additional keywords: `extrapolation_mode`

### AreaWeighted
- Creates `AreaWeightedRegridder` (from analysis._area_weighted).
- Source and target grid must be lat/lon DimCoords.
- Weights are cached on `__init__`.
- Uses python for loop for calculations rather than a sparse matrix.
- Uses `RectilinearRegridder._create_cube` with the callback being equivalent to Linear. This means that a different regridding method is acting on the AuxCoords and the data.
- Requires special handling of masked data in calculations.
- Lazy regridding is supported via `map_complete_blocks`.
- Additional keywords: `mdtol`

### UnstructuredNearest
- Creates `UnstructuredNearestNeighbourRegridder` (from analysis.trajectory) which uses `analysis.trajectory.interpolate` for calculations.
[...]

PR #4807, opened 2022-06-16
(Regridder unification) improve curvilinear regridding, generalise _create_cube

stephenworsley:
An attempt at demonstrating some of the ideas in #4754. That is, using a common function to create cubes from regridded data.

---
This addresses two of the bullet points in this issue:
- Rewrite the calculations in `CurvilinearRegridder` to work on whole cubes rather than slices.
- Rewrite the way callbacks work in `_create_cube` so that they can be consistent with the regridder they are being used in.

This PR replaces the cube creation code in the `RectilinearRegridder`, `AreaWeightedRegridder` and the `CurvilinearRegridder` so that they both call the same function, `_create_cube`. To accomplish this, surrounding code has been altered to be compatible with this new function. Whiole the behaviour of these regridders should otherwise remain the same, this has lead to some improvements in the functionality:
- The calculations in the `CurvilinearRegridder` ought to be more efficient now that they are running on full data rather than slices (TODO: check if this is a measurable performance improvement.)
- The `CurvilinearRegridder` is now able to handle derived coordinates in the same way that the `AreaWeightedRegridder` and `RectilinearRegridder` do (TODO: add tests to verify that this works).
- While this functionality is not yet in the `AreaWeightedRegridder`, the associated function `regrid_area_weighted_rectilinear_src_and_grid` is now able to handle derived coordinates when one of the grid coordinates is scalar.
- While this functionality is not yet in the `RectilinearRegridder`, the underlying functions should now be able to handle scalar grid coords. This should make it easier to add this in a subsequent PR.


bjlittle:
@stephenworsley Fancy rebasing from `main`?

This will enable CI for GHA :+1:

stephenworsley:
The goal with this PR is to demonstrate the possibility of a common structure that regridders can follow. That structure would look something like the pseudocode bellow. The `_prepare` and `regrid` functions would have to be specific to each regridder, most of the other functions may have small differences due to the types of "grids" they deal with, but there is room for code reuse. In particular, this PR tries to demonstrate that `create_cube` can be made more generic so that one function can be reused by many regridders (with the bonus that these regridders will then be able to handle derived coordinates).

### Rough proposed common structure of regridders:

```python

def regrid(data, dims, regrid_info):
  ...
  return new_data

def create_cube(data, src, src_dims, tgt_coords, callback):
  ...
  return result_cube

def _prepare(src, tgt):
  ...
  return regrid_info

def _perform(cube, regrid_info):
  ...
  dims = get_dims(cube)
  data = regrid(cube.data, dims, regrid_info)
  callback = functools.partial(regrid, regrid_info=regrid_info)
  return create_cube(data, src, dims, tgt_coords, callback)

class Regridder:
  def __init__(src, tgt):
    ...
    self._regrid_info = _prepare(src, tgt)
  def __call__(src):
    ...
    return _perform(src, self._regrid_info)
```

lib/iris/experimental/regrid_conservative.py
```diff
@@ -329,16 +331,23 @@ def _valid_units(coord):
 
     # Return result as a new cube based on the source.
     # TODO: please tidy this interface !!!
-    return RectilinearRegridder._create_cube(
-        fullcube_data,
-        src=source_cube,
-        x_dim=src_dims_xy[0],
-        y_dim=src_dims_xy[1],
+    _regrid_callback = functools.partial(
```
lbdreyer, line 334: In this module it says we have deprecated `iris.experimental.regrid_conservative` so this change is probably unnecessary?

lib/iris/analysis/_regrid.py
```diff
@@ -287,83 +288,105 @@ def _regrid_indices(cells, depth, points):
...
+    new_data_shape = list(data_shape)
+    for dim, length in zip(inds, grid_cube.shape):
+        new_data_shape[dim] = length
+    if len(dims) == 1:
+        new_data_shape.append(grid_cube.shape[1])
+        dims = (dims[0], dims[0] + 1)
+    if len(dims) > 2:
+        new_data_shape = new_data_shape[: 2 - len(dims)]
+        dims = dims[:2]
+    inds = [-2, -1]
```
lbdreyer, line 358: Was this hardcoding of `inds` leftover from some debugging, perhaps?

lbdreyer (commented):
This is a great improvement and a fantastic speed up!

I still have a few things I want to check, but I wanted to submit the comments I had so far in case you wanted to start working on them

lib/iris/experimental/regrid_conservative.py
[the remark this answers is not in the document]
stephenworsley: This was necessary in order to remove `RectilinearRegridder._regrid`. It also has the benefit of improving the test coverage for this approach.

lib/iris/analysis/_regrid.py
```diff
@@ -1089,20 +971,168 @@ def __call__(self, src):
...
+        The source Cube.
+    src_dims : tuple of int
+        The dimensions of the X and Y coordinate within the source Cube.
+    tgt_coords : tuple of :class:`iris.coords.Coord`\\ 's
+        Either two 1D :class:`iris.coords.DimCoord`\\ 's, two 1D
+        :class:`iris.experimental.ugrid.DimCoord`\\ 's or two ND
+        :class:`iris.coords.AuxCoord`\\ 's representing the new grid's
+        X and Y coordinates.
+    num_tgt_dims : int
+        The number of dimensions that the `tgt_coords` span.
```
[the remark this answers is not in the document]
stephenworsley: The idea here is that if you have a pair of 1D coords, these could be describing a 2D rectilinear grid or they could be on the same dimension and describe 1D unstructured data. This is inherited from iris-esmf-regrid where this is necessary. The idea here is that ultimately, iris-esmf-regrid could inherit this function directly from iris.

lib/iris/analysis/_regrid.py
stephenworsley: Yeah, I think this was rewritten a few times and I suppose it could be made clearer.

lib/iris/analysis/_regrid.py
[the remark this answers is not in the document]
stephenworsley: More relevantly, the `UnstructuredNearestNeigbourRegridder` would have to use this when that is refactored. This wasn't refactored as part of this PR since the existing structure is a bit harder to untangle so is worth handling in its own PR.

lib/iris/tests/unit/analysis/regrid/test__CurvilinearRegridder.py
```diff
@@ -168,6 +169,62 @@ def test_caching(self):
...
+        res_t = rg_t(src_t)
+        assert np.allclose(
+            res_t.coord("altitude").points,
+            np.moveaxis(src_t.coord("altitude").points, 1, 2),
+        )
+        src_t.transpose([0, 2, 1])
+        res_t = rg_t(src_t)
+        assert np.allclose(
+            res_t.coord("altitude").points,
+            np.moveaxis(src_t.coord("altitude").points, 1, 2),
```
lbdreyer, line 225: The tests look good although I think they need to be restructured. You're calling the regridder three times here which seems to go against the ethos of unit tests. I would prefer a separate call to the regridder per unit test. Could you perhaps move the set up of src and tgt into a fixture?

lbdreyer (approved):
Looking good to me! Thanks @stephenworsley

PR #4821, opened 2022-06-21
docs: cirrus-ci to gha

bjlittle:
### Description
This PR aligns the documentation with the recent migration from `cirrus-ci` to GHA, and is a follow-up to #4503.

---


bjlittle:
@wjbenfold We may want to bank this PR prior to #4759?

They're both stomping about in the same area of the docs.

Alternatively, happy to bank #4759 first and rebase this PR. Your call...

wjbenfold (changes requested):
Looking pretty good :) Made a couple of small suggestions

The `.cirrus.yml` link in `common_links.inc` is broken, it should be removed along with the references in `3.0.rst` and `3.1.rst`

bjlittle:
@wjbenfold Thanks.

I do find that retro-fixing past whatsnew's a bit weird... but there's no other way around it really. They're simply referencing things that now no longer exist. I guess in my head they where almost cast in stone once published, but that's clearly not the case.

wjbenfold:
> @wjbenfold Thanks.
> 
> I do find that retro-fixing past whatsnew's a bit weird... but there's no other way around it really. They're simply referencing things that now no longer exist. I guess in my head I thought of them being cast in stone once published, but that's clearly not the case.

Maybe it's best to not reference things from the whatsnew like that? Or to have the link checker skip old what's new pages?

bjlittle:
> Maybe it's best to not reference things from the whatsnew like that?

TBH perhaps the easiest approach :+1: 

I'd be hesitant about turning off the link-checker... almost seems like that would be setting ourselves up for a fall.

wjbenfold (approved):

bjlittle:
@wjbenfold awesome, thanks :beers:

PR #4827, opened 2022-06-23
Fix stock CDL headers for unlimited time dimension.

trexfeathers:
### Description

Was previously not possible to generate NetCDF-4 files using these headers - got stuck in an infinite loop somewhere.

---


pp-mo (approved):
Ok, this is somewhat exciting.
We can only really test by merging, so let's try it !

PR #4832, opened 2022-06-27
Python 3.9 CI & Combine CI workflows

bjlittle:
### Description
This PR combines the `ci-docs-tests` GHA workflow into the `ci-test` workflow.

It also unpins `python` for CI, so we now test on `python` 3.8 and 3.9.

---


.github/workflows/ci-tests.yml
```diff
@@ -35,8 +35,12 @@ jobs:
       fail-fast: false
       matrix:
         os: ["ubuntu-latest"]
-        python-version: ["3.8"]
-        session: ["tests"]
+        python-version: ["3.9"]
```
jamesp, line 38: So what does this matrix look like when unraveled? All test sessions on py3.9 and then only tests on py3.8?

trexfeathers: That seems to be the case based on the CI that has run on this PR

bjlittle: I'm assuming that we only want to run `doctest`, `gallery`, and `linkcheck` with the latest `python` and no other version, and `tests` across several versions that we care about.

I've used the `matrix` as a convenience to get the latest permutations without being explicit, and using `include` to add the exceptions to the pattern i.e., running `tests` on less modern versions of `python`.

Alternatively, we could fully load the `python-version` and go with an `exclude` pattern, or just be completely explicit.

Just whatever people are happiest with and doesn't look to contrived.

TBH I was also going to add `macos-latest` as a follow-on PR, and that might tip the scales in favour of being explicit... which is where I'm expecting we might end up, given the workflow behaviour that we want.

bjlittle: @jamesp Fancy merging and we'll find out? :wink:

jamesp (approved):

bjlittle:
@jamesp Awesome! Thanks :100: :partying_face:

Issue #4190, opened 2021-06-11
Support lazy saving


bouweandela:
## ✨ Feature Request

The `iris.save` function only supports saving a single file at a time and is not lazy. However, the [`dask.array.store`](https://docs.dask.org/en/latest/array-api.html#dask.array.store) function that backs the NetCDF saver supports delayed saving. For our use case, it would be computationally more efficient to have this supported by `iris.save`. Would it be possible to allow providing the `compute=False` option to the current `iris.save` function, so instead of saving directly it returns a `dask.delayed.Delayed` object that can be computed at a later time?

Alternatively, having a save function in iris that allows saving a list of cubes to a list of files, one cube per file (also similar to `da.store` but then working for cubes) would also work for us.

## Motivation

In our case, [multi model statistics in ESMValTool](https://docs.esmvaltool.org/projects/esmvalcore/en/latest/recipe/preprocessor.html#multi-model-statistics), we are interested in computing statistics (e.g. mean, median) over a number of climate models (cubes). Before we can compute those statistics, we need to load the data from disk and regrid the cubes to the same horizontal grid (and optionally to the same vertical levels). Then we merge all cubes into a single cube with a 'model' dimension and collapse along that dimension using e.g. `iris.analysis.MEAN` to compute the mean.

We want to store both the regridded input cubes and the cube(s) containing the statistics, each cube in it's own netCDF file according to the CMIP/CMOR conventions. Because `iris.save` only allows saving a single cube to a single file and is immediately executed, the load and regrid needs to be executed (1 + the number of statistics) times. Having support for delayed saving (or saving a list of cubes to a matching list of files) would save computational time, because the regridded chunks can be re-used for computing each statistic (as well as for storing the regridded cube), so we only need to load and regrid the chunk once.

## Additional context

<details>
<summary><b>Example script that shows the use case</b></summary>

This is an example script that demonstrates our workflow and how we could use the requested save function to speed up the multi-model statistics computation. Note that the script uses lazy multi-model statistics, which are still in development in https://github.com/ESMValGroup/ESMValCore/pull/968.

```python3
import os
import sys

import dask
import dask.array as da
import iris
from netCDF4 import Dataset

from esmvalcore.preprocessor import multi_model_statistics, regrid


def save(cube, target, compute):
    """Save the data from a 3D cube to file using da.store."""
    dataset = Dataset(target, "w")
    dataset.createDimension("time", cube.shape[0])
    dataset.createDimension("lat", cube.shape[1])
[...]

Issue #565, opened 2013-06-19
Add functionality to allow appending cubes to existing netcdf file


rsignell-usgs:
Currently writing Iris data to netcdf creates a new file.    In the case where we are using Iris to process a large amount of data (in my case a 30 year global hindcast), we need to be able process timestep-by-timestep and append the processed result at each timestep to the existing netcdf file, as we don't have enough computer memory to hold all the timesteps at once.   An example of what we would like to do is here, using NetCDF4 to append, but we would like to be able to accomplish this with Iris.

http://nbviewer.ipython.org/5777643

The relevant code snippet is:

```

nc = netCDF4.Dataset('cfsr.nc','r+')
for i in range(10):
    url='http://nomads.ncdc.noaa.gov/thredds/dodsC/modeldata/\
cmd_ocnh/2009/200905/200905%2.2d/ocnh01.gdas.200905%2.2d00.grb2' % (i+1,i+1)
    print url
    cubes = iris.load(url)
    t = cubes[4]
    slice=t.extract(iris.Constraint(longitude=lambda cell: -77.+360. < cell < -63.0+360.,latitude=lambda cell: 34. < cell < 46.0))
    nc.variables['Potential_temperature'][i,:,:,:]=slice.data
```

arulalant:
Is this #565 open ? 

cubes = [temperature, pressure, humidity]
for cube in cubes:
    iris.save(cube, 'myfile.nc', append=True)

Still I couldn't append cubes (list of different variables) into nc file.

Thanks,
Arulalan.T

rhattersley:
> Is this #565 still open ?

Yes, but I'm not aware of anyone working on a fix. Sorry!

pp-mo:
I've been meaning for ages to respond to this.

I spent quite a while a few months back trying to implement an append, mostly focussed on @rhattersley case #1, 
i.e. **_specifically_** appending to all the relevant variables along an unlimited dimension.

It's a natural usecase, e.g. to open a file and add "today's data" to it, especially as the necessary operation exists in netcdf (without requiring you to read+re-write all the existing, even by streaming it).

I spent quite a while climbing the mountain of "decoupling the saving code from actual file operations", so it instead produces an abstract representation of the required (new) data, 
With this, I could then either create new variables or extend existing ones.
Working code ideas here : https://github.com/pp-mo/iris/pull/51

[...]

PR #4835, opened 2022-06-28
read-write Xarray with nc-dataset adapter.

pp-mo:
Targets feature branch, just to unify testing.
This introduces a "netcdf dataset adapter" for xarray datasets, enabling to read/write xarray data as if it were a netcdf file.
ALSO combined here (and not really separate for now):  
  * a new loading "scheme" enabling iris to load from odd objects in place of file-specs:
    * [in `iris : _generate_cubes`](https://github.com/SciTools/iris/blob/5d2180425e111dd2796c953c7c7de531ddad5f11/lib/iris/__init__.py#L268-L271)
    * [in `iris.io : decode_uri`](https://github.com/SciTools/iris/blob/5d2180425e111dd2796c953c7c7de531ddad5f11/lib/iris/io/__init__.py#L138-L142)
    * [in `iris.io.format_picker`](https://github.com/SciTools/iris/blob/5d2180425e111dd2796c953c7c7de531ddad5f11/lib/iris/io/format_picker.py#L335-L351)
  * changes enabling the netcdf file format to [read from an open dataset (instead of filepath)](https://github.com/SciTools/iris/blob/5d2180425e111dd2796c953c7c7de531ddad5f11/lib/iris/fileformats/cf.py#L1048-L1058)
  * changes enabling the netcdf file format to [save to an open dataset (instead of filepath)](https://github.com/SciTools/iris/blob/5d2180425e111dd2796c953c7c7de531ddad5f11/lib/iris/fileformats/netcdf.py#L1109-L1125)

If this idea flies, it might be worth separating out the core-iris changes into a separate PR rather than doing it all at once?

Hence **_"WIP"_**

~Also, various TO-DOs noted on this old "private PR" : https://github.com/pp-mo/iris/pull/72~
(update: all done)


pp-mo:
**Status update**:
Almost there with CI.
Tests are functioning.
Now just some docs-build errors.

lib/iris/fileformats/netcdf.py
```diff
@@ -1103,20 +1107,23 @@ def __init__(self, filename, netcdf_format):
...
-            dir_name = os.path.dirname(filename)
-            if not os.path.isdir(dir_name):
-                msg = "No such file or directory: {}".format(dir_name)
-                raise IOError(msg)
-            if not os.access(dir_name, os.R_OK | os.W_OK):
-                msg = "Permission denied: {}".format(filename)
-                raise IOError(msg)
-            else:
-                raise
+        if hasattr(filename, "createVariable"):
```
jamesp, line 1110: Is it worth considering a runtime protocol to define, in code, what you mean by "supporting the netCF4.Dataset api"?

https://docs.python.org/3/library/typing.html#typing.runtime_checkable. Added in Python 3.8.

lib/iris/fileformats/netcdf.py
```diff
@@ -1061,6 +1064,7 @@ def __init__(self, filename, netcdf_format):
 
         * filename (string):
             Name of the netCDF file to save the cube.
+            OR a writeable object supporting the netCF4.Dataset api.
```
jamesp, line 1067:
```suggestion
            OR a writeable object supporting the netCDF4.Dataset api.
```

lib/iris/experimental/xarray_dataset_wrapper.py
```diff
@@ -0,0 +1,415 @@
...
+import netCDF4 as nc
+import numpy as np
+import xarray
+import xarray as xr
+
+
+class _XrMimic:
+    """
+    An netcdf object "mimic" wrapped around an xarray object, which will be
+    either a dim, var or dataset.
```
jamesp, line 29: By dim do you mean DimCoord, in iris parlance?

jamesp: might be good to be explicit here: var I assume is `xr.Variable`, dataset is `xr.Dataset`. dim is an object that doesn't exist in xarray, but you need it for netcdf compat?

lib/iris/experimental/xarray_dataset_wrapper.py
```diff
@@ -0,0 +1,415 @@
...
+    We also support object equality checks.
+
+    NOTE: a DimensionMimic, uniquely, does *NOT* in fact contain an xarray
+    object, so its self._xr == None.   See  DimensionMimic docstring.
+
+    """
+
+    def __init__(self, xr, name=None, group=None):
+        """
+        Create a mimic object wrapping a :class:`nco.Ncobj` component.
```
jamesp, line 45: What is `nco.Ncobj`? Old init string text?

lib/iris/experimental/xarray_dataset_wrapper.py
```diff
@@ -0,0 +1,415 @@
...
+    def datatype(self):
+        return self.dtype
+
+    @property
+    def dimensions(self):
+        return self._xr.dims
+
+    def __getitem__(self, keys):
+        if self.ndim == 0:
+            return self._xr.data
```
jamesp, line 161: `keys` is ununsed in this branch of the if. Is that right? Should `keys` have default value? Document with a comment either way I think.

lib/iris/experimental/xarray_dataset_wrapper.py
```diff
@@ -0,0 +1,415 @@
...
+            If None, initialise empty.
+            In either case, the result can be read or written like a
+            :class:`netcdf.Dataset`.  Or, an xarray equivalent can be
+            regenerated with the :meth:`to_xarray_dataset` method.
+
+        Notes
+        -----
+        Only a limited subset of the :mod:`netCDF4` APIs are currently
+        supported : just enough to allow Iris to read and write xarray datasets
+        in place of netcdf files.
```
jamesp, line 209: Again, this might be where a runtime protocol could help document which parts of the api are implemented so far.

jamesp (commented):
Not a complete review, just some questions that popped out at me as I read the code.

This is SUPER cool :D

pp-mo:
Here's an interesting follow-on possibility ... https://github.com/pp-mo/iris/pull/73 
(keeping this feature separate for now, just to simplify the discussion).

pp-mo:
## Important Context : things this can be used for ?
In addition to just "fixing data a bit" on either load or save,
I think this **_might_** possibly help with a number of awkward outstanding desired features relating to netcdf :
  * load chunks control : https://github.com/SciTools/iris/issues/3333
  * precise formatting round-trips : https://github.com/SciTools/iris/discussions/4215
  * lazy saving : https://github.com/SciTools/iris/issues/4190
  * output "append" mode : https://github.com/SciTools/iris/issues/565
  
(Sorry for late arrival : I initially added this note to the wrong ticket!)

pp-mo:
> Ping @dcherian

Thanks for making the link @bjlittle 

FWIW some other comments on this that you might also like to consider ..  

The current approach here is rather untidy, and grew out of an attempt to create a simple 'wrapper' object to an existing xarray dataset.  But that basic approach failed once writing is supported, since it is not possible to modify the dimensions of an existing xarray dataset, and the result is now a rather clunky "mixed" approach.
The principal limitation as it is, is that is expects an xarray dataset which does _not_ intpret cf/coords/time-data
  \- which is why those are [all turned off in the test example](https://github.com/SciTools/iris/pull/4835/files#diff-6de794aa77f377780a91833b78ca288645c68c1443b15c634d60bde6ce68b3e6R35-R37).
Likewise, the xarray that you get out of the 'to_array' method will also have shortcomings in those respects + not match a "normal" load from file (e.g. in terms of time data).

If this "load/save xarray as a netcdf dataset" approach has some **_general_** utility, then we should put it into xarray (not Iris!), 
and work on those problems, which I believe is entirely feasibble.

dcherian:
Thanks @bjlittle 

Is there an issue I can read about the motivation for this?

If it is to take advantage of Xarray's read/write backend options, would it be easier to use the [`to_iris` method](https://docs.xarray.dev/en/stable/generated/xarray.DataArray.to_iris.html)?

pp-mo:
> Thanks @bjlittle
> 
> Is there an issue I can read about the motivation for this?

(https://docs.xarray.dev/en/stable/generated/xarray.DataArray.to_iris.html)?

You're quite right, there should probably be an issue.
I will work towards that ...
For now, I can basically say that the ideal goal is _~lossless conversion between iris, xarray and netcdf files_.

----

> If it is to take advantage of Xarray's read/write backend options, would it be easier to use the [`to_iris` method]

It seems to me that a big "problem" with xarray `to_iris`/`from_iris` is that it requires embedding knowledge of Iris into xarray.  And also, effectively, knowledge of CF, _and_ what parts of CF Iris supports -- all of which is really Iris' business.

Meanwhile, there is a reasonable argument that an `iris.to_array`/`from_xarray` would be a better approach, since what Iris does is mostly "extend" xarray capabilities by adding CF handling.
But that then causes essentially the same problems of _injecting xarray-specifics into Iris_. 

And with both those solutions, there is a **_particular_** problem regarding dependencies and testing, especially for C.I. :
  - in order to test an xarray to/from-iris, the xarray CI needs to test against Iris, which is a costly dependency
  - and exactly the same is true for the "converse solution" (iris to/from_xarray)

**However**, this kind of "dataset adapter" approach _only_ depends on knowledge of xarray and netCDF4.
  * so, as it has no Iris-specific knowledge, it really should **_not_** be here  -- this is only a PoC
     * it could properly belong _in_ xarray, 
     * or in a separate package
  * and it could then be used by **_any other packages_** wanting to establish file-less linkage to xarray data
    * **does that seem feasible ?**

PR #4838, opened 2022-06-28
Remove redundant testing assert methods

rcomer:
### Description

A long time ago Iris implemented some bespoke `assert` methods to help with testing.  For two of these we now have equivalent methods in `unittest`:

* [assertMultiLineEqual](https://docs.python.org/3/library/unittest.html#unittest.TestCase.assertMultiLineEqual) (new in python3.1) appears to do the same as our [assertStringEqual](https://github.com/SciTools/iris/blob/a908a27937e4ca8eb810001ebd64f11ff75dce72/lib/iris/tests/__init__.py#L226).
* [assertWarnsRegex](https://docs.python.org/3/library/unittest.html#unittest.TestCase.assertWarnsRegex) (new in python3.2) does the same job as our [assertWarnsRegexp](https://github.com/SciTools/iris/blob/a908a27937e4ca8eb810001ebd64f11ff75dce72/lib/iris/tests/__init__.py#L564), except it additionally checks the warning type, which seems like a better test to me.  I find these two particularly confusing because my editor suggests both as I'm typing and I can never remember which is which!

---

### DIFF
--- a/docs/src/developers_guide/testing_tools.rst
+++ b/docs/src/developers_guide/testing_tools.rst
@@ -24,8 +24,7 @@ Custom assertions
 =================
 
 :class:`iris.tests.IrisTest` supports a variety of custom unittest-style
-assertions, such as :meth:`~iris.tests.IrisTest_nometa.assertStringEqual`,
-:meth:`~iris.tests.IrisTest_nometa.assertArrayEqual`,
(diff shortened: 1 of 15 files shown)


bjlittle (changes requested):
@rcomer LGTM, thanks :+1:

Did you want to add a `whatsnew` entry in the `Internal` section?

bjlittle (approved):

PR #4856, opened 2022-07-07
Choose prototypes more intelligently

### LINKED ISSUE #4842: Mergenate should always extend coords of flat cubes
## 🐛 Bug Report

Mergenate currently requires `extend_coords=True` if the first cube is flat in the merge dimension and there is a coordinate that lies along that dimension in every cube except the first. If the order of the cubelist is changed then this won't be an issue, as mergenate is checking the first cube in the list only.

## How To Reproduce
This doctest fails without `extend_coords=True`

https://github.com/SciTools/iris/blob/3fe8fd1d60e10195c4eeaeec08b69492607e3719/lib/iris/experimental/mergenate.py#L574-L587

## Expected behaviour

Requiring `extend_coords` should be independent of the cubelist order.
A cube that is flat in the merge dimension should have any required scalar coordinates extended by the mergenate process.

## Other info
This line could be smarter

https://github.com/SciTools/iris/blob/3fe8fd1d60e10195c4eeaeec08b69492607e3719/lib/iris/experimental/mergenate.py#L303

We'll also need changes in this function

https://github.com/SciTools/iris/blob/3fe8fd1d60e10195c4eeaeec08b69492607e3719/lib/iris/experimental/mergenate.py#L112

wjbenfold:
### Description

Closes #4842 by using a more intelligent cube as the source of the prototype information (a cube that isn't flat in the merge dimension, if one is available). Also artificially sets `extend_coords` if all cubes are of height 1 in the merge dimension, as that's also a good reason to extend coords.

---


bjlittle:
@wjbenfold I've unblocked the `SciTools-CLA-checker` for you 😉

bjlittle (approved):
@wjbenfold Let's go 🚀

PR #4857, opened 2022-07-07
min pin for cf-units

bjlittle:
### Description
This PR adds a min pin to `cf-units` for the `v3.2.x` feature branch.

---


wjbenfold (approved):
LGTM

`nc-time-axis` pins `cftime` to `>=1.5.0` so we don't need to keep that pin

requirements/ci/py38.yml
```diff
@@ -11,8 +11,8 @@ dependencies:
 
 # Core dependencies.
   - cartopy >=0.20
-  - cf-units >=3
-  - cftime >=1.5
+  - cf-units ==3.0.1
```
bjlittle, line 14: Bizarrely, although the `cf-units-3.0.1.post0` package is **actually** available on the `conda-forge` anaconda channel [here](https://anaconda.org/conda-forge/cf-units?version=3.1.0.post0&channel=main)... `conda` can't find it to install it! :confounded: 

However, there is no functional difference between `3.0.1` and `3.0.1.post0`, so it is safe to proceed with this verison pin "as-is" to get around this `conda` package version weirdness.

setup.cfg
```diff
@@ -46,8 +46,8 @@ version = attr: iris.__version__
 include_package_data = True
 install_requires =
     cartopy>=0.20
-    cf-units>=3
-    cftime>=1.5.0
+    cf-units==3.0.1.post0
```
bjlittle, line 49: However, this version is available on PyPI and installable using `pip`... so let's keep that honest.

bjlittle: The subtle thing here is that the `3.0.1` and `3.0.1.post0` PyPI wheels were built with different mechanisms... so I'd need to check what the actual differences were to roll it back to `3.0.1`.

Worth understanding... I'll check

bjlittle: Ahhh okay... so the PyPI `3.0.1` only contains the `sdist` with no `wheels`.

Whereas `3.0.1.post0` has both i.e., the `post0` change was the first iteration of infrastructure on `cf-units` to automate and build the `wheels` for `py37`, `py38` and `py39` from the `sdist`.

bjlittle: `pip` can install from `sdist`, although it's less preferred over using a `wheel`.

Let me experiment as see if it's safe to roll back to `3.0.1` to keep things consistent.

bjlittle: Okay, so my recommendation is to keep `site.cfg` with `cf-units==3.0.1.post0`. Why?

Simply because the `3.0.1.post0` version from PyPI will install with a `wheel`, which includes the required `udunits2` shared object that `cf-units` links into at runtime i.e., if you `pip install cf-units==3.0.1.post0` it will "just work" :tm:

If users `pip install scitools-iris==3.2.1.post0` for whatever reason and we pin `cf-units==3.0.1` within `setup.cfg`, then it could hit the problems that `cf-units==3.0.1.post0` solves by providing bdist wheels i.e., installing `cf-units` from a `sdist` relies on the environment already having `udunits2` available, which is not guaranteed to be the case, and if so the install with fail.

What say ye @trexfeathers and @lbdreyer ?

trexfeathers: OK, thanks for explaining 👌

trexfeathers (approved):
I'm 👍 on this PR

I will wait for @lbdreyer

lbdreyer (approved):
Thanks for all your effort on this @bjlittle ! I think we can comfortably say this release is ready to go.
I'm just going to wait for the benchmarks to run, then I will merge this in!

bjlittle:
@wjbenfold @trexfeathers and @lbdreyer 

Awesome, thanks guys for helping get the across the line :beers:

PR #4861, opened 2022-07-12
pytest verbosity

rcomer:
### Description

We are currently both increasing and decreasing the verbosity of pytest in our CI, because we set the `-v` and `-q` flags in different places.  If we remove these flags

* I think the CI output will be the same (can check in a minute!)
* Developers who want more verbose output when running locally may use the `-v` flag.  Currently it doesn't work because the `-q` setting in the `pyproject.toml` is picked up, and the settings cancel each other out.

---

### DIFF
--- a/lib/iris/tests/runner/_runner.py
+++ b/lib/iris/tests/runner/_runner.py
@@ -130,7 +130,6 @@ class TestRunner:
 
         args = [
             None,
-            "-v",
             f"-n={self.num_processors}",
         ]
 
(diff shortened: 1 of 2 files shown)


rcomer:
[Test output from this PR](https://github.com/SciTools/iris/runs/7306811447?check_suite_focus=true#step:12:591)
[Test output from last commit to main](https://github.com/SciTools/iris/runs/7279634980?check_suite_focus=true#step:12:593)

Look the same to me 🙂

bjlittle (approved):
@rcomer LGTM :+1:

rcomer:
Thanks @bjlittle 😊

PR #4885, opened 2022-08-01
Sphinx reset modules

bjlittle:
### Description
This PR resolves the issue of the documentation failing to build on RTD.

This is related to the recent release of `sphinx-gallery` 0.11.0.

Also see:

- https://github.com/sphinx-gallery/sphinx-gallery/pull/890
- https://github.com/matplotlib/matplotlib/pull/23514

Thanks @rcomer :+1: 

TODO:

- [x] ~Resolve the conda lockfiles as part of this PR~

---

### DIFF
--- a/docs/src/developers_guide/contributing_documentation_full.rst
+++ b/docs/src/developers_guide/contributing_documentation_full.rst
@@ -147,7 +147,7 @@ can exclude the module from the API documentation.  Add the entry to the
 Gallery
 ~~~~~~~
 
-The Iris :ref:`sphx_glr_generated_gallery` uses a sphinx extension named
+The Iris :ref:`gallery_index` uses a sphinx extension named
 `sphinx-gallery <https://sphinx-gallery.github.io/stable/>`_
 that auto generates reStructuredText (rst) files based upon a gallery source
 directory that abides directory and filename convention.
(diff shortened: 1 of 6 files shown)


docs/src/developers_guide/contributing_documentation_full.rst
```diff
@@ -147,7 +147,7 @@ can exclude the module from the API documentation.  Add the entry to the
 Gallery
 ~~~~~~~
 
-The Iris :ref:`sphx_glr_generated_gallery` uses a sphinx extension named
+The Iris :ref:`gallery_index` uses a sphinx extension named
```
bjlittle, line 150: `sphinx-gallery` 0.11.x doesn't appear to automagically create the `sphx_glr_generated_gallery` namespace reference, so defaulting back to the static reference of `gallery_index`, which is guaranteed to always be available.

bjlittle:
This is an example of RTD building with this PR, see [here](https://readthedocs.org/projects/bjlittle-iris/builds/17594084/)

bjlittle:
Appear to have developed a single genuine test failure across `py38`, `py39` and `py310` as a result of updating the dependencies...

bjlittle:
@rcomer Not sure... but I noticed that the `pyshp` dependency has changed.

I'll drill down and see if I can isolate the root cause. Joy.

bjlittle:
To separate concerns, I've reverted the `conda` lockfile update, as that's causing a follow-on test failure.

So, let's just bank this PR "as-is" to unblock the documentation building and RTD, and I'll follow-up with another PR to update the lockfiles and address the test failure.

rcomer (approved):
Change is consistent with matplotlib’s approach and I’ve seen that the docs build works with new sphinx-gallery.  LGTM 👍🍻

PR #4889, opened 2022-08-03
Upgrade `iris.util.mask_cube`

### LINKED ISSUE #3717: mask_cube() both acts in-place and returns the Cube
This is following a post I made on the Google Group: https://groups.google.com/forum/#!topic/scitools-iris/eZwzmiZO3UU

There has been some discussion in the past about whether functions acting on Cubes should modify the Cube in-place or return a new Cube (see #3430 and #2615).  I find the behaviour of ```iris.util.mask_cube()``` surprising in that it does both.  Seeing it returning a Cube, I was surprised to find my original Cube also masked:
```
>>> cube = iris.cube.Cube(range(5))
>>> points_to_mask = [False, True, False, False, True]
>>> cube2 = iris.util.mask_cube(cube, points_to_mask)
>>> cube.data
masked_array(data=[0, --, 2, 3, --],
             mask=[False,  True, False, False,  True],
       fill_value=999999)
```

Moreover, the returned Cube is the same object as the one I started with:
```
>>> cube2 is cube
True
```

I suggest ```mask_cube()``` should either modify the Cube in-place or return a new Cube, not both.

### LINKED ISSUE #3936: iris.util.mask_cube wish list
## ✨ Feature Request
It looks like the [`mask_cube`](https://scitools-iris.readthedocs.io/en/latest/generated/api/iris/util.html#iris.util.mask_cube) function was added as a convenience as part of a larger change (#3144).  However, now it's in the public API, users are having a go with it but not finding it ideal.  Here are a few ideas for improvement that I think would be fairly easy to implement:

* The docstring states that both `cube` and `points_to_mask` must be 2D but, looking at the code, I can't see why it wouldn't work as long as they are the same shape.  So at very least, the docstring should be updated.
* A common masking task is to apply a 2D land/ocean mask to data that has higher dimensionality.  So it would be nice to be able to pass a `dim_map`, so the function could broadcast the mask out via `iris.util.broadcast_to_shape`.
* Enforce boolean dtype on the mask.  With Iris3rc we have:
```python
cube = iris.cube.Cube(np.arange(24).reshape(6,4))
mymask = np.array([0, 1, 0, 1])
iris.util.mask_cube(cube, mymask)
cube.data
```
gives
```
masked_array(
  data=[[--, --, --, --],
        [--, --, --, --],
        [8, 9, 10, 11],
        [12, 13, 14, 15],
        [16, 17, 18, 19],
        [20, 21, 22, 23]],
  mask=[[ True,  True,  True,  True],
        [ True,  True,  True,  True],
        [False, False, False, False],
        [False, False, False, False],
        [False, False, False, False],
        [False, False, False, False]],
  fill_value=999999)
```
which is not at all what I would expect (and I get this result even if I broadcast the mask before applying it).

* Handle lazy data.

## Additional context
See also #3717

rcomer:
### Description

Closes #3717 and closes #3936.

This PR re-implements the `mask_cube` function to solve various issues:

:heavy_check_mark: Works in place **or** returns a new cube (user choice).  I made `in_place=False` the default as, based on the existing test, that appears to be the original intention.  It is also consistent with e.g. arithmetic.
:heavy_check_mark: If either of the two inputs are lazy, the result will be lazy.
:heavy_check_mark: The mask array is cast to boolean within the function, so it works in the same way if it's originally supplied as integers (or floats).
:heavy_check_mark: Uses the same infrastructure as arithmetic, so the mask may be supplied as a cube, which will then be broadcast across the appropriate dimensions based on coordinate matching.

Thanks to the magic of `Resolve`, we could even use this to produce a cube with higher dimensionality than the original:
<details><summary>Example creating a series of leave-one-out samples for verification.</summary>

```python
import iris
import numpy as np

# Load some data
cube = iris.load_cube(iris.sample_data_path("E1_north_america.nc"))[:8]
print("Original cube:", cube, sep="\n", end="\n\n")

# Create a mask cube with each time masked (i.e. True) for exactly one sample.
mask = iris.cube.Cube(np.identity(8, dtype=bool), long_name="leave_one_out")
sample_coord = iris.coords.DimCoord(range(1, 9), long_name="sample_number")
for dim, coord in enumerate([sample_coord, cube.coord("time")]):
    mask.add_dim_coord(coord, dim)
    
print("Mask cube:", mask, sep="\n", end="\n\n")

# Add a length one dummy "sample_number" coord to our original cube, so Resolve
# knows where to put the new dimension.

cube.add_aux_coord(sample_coord[0])
cube = iris.util.new_axis(cube, "sample_number")

# Create our leave-one-out samples.
samples = iris.util.mask_cube(cube, mask)

print("Output cube:", samples, sep="\n", end="\n\n")
print("Third sample's data:", samples[2, :, :2, :2].data, sep="\n", end="\n\n")
```

Out:
```
Original cube:
air_temperature / (K)               (time: 8; latitude: 37; longitude: 49)
    Dimension coordinates:
        time                             x            -              -
        latitude                         -            x              -
        longitude                        -            -              x
    Auxiliary coordinates:
        forecast_period                  x            -              -
    Scalar coordinates:
        forecast_reference_time     1859-09-01 06:00:00
        height                      1.5 m
    Cell methods:
        mean                        time (6 hour)
    Attributes:
        Conventions                 'CF-1.5'
        Model scenario              'E1'
        STASH                       m01s03i236
        source                      'Data from Met Office Unified Model 6.05'

Mask cube:
leave_one_out / (unknown)           (sample_number: 8; time: 8)
    Dimension coordinates:
        sample_number                             x        -
        time                                      -        x

Output cube:
air_temperature / (K)               (sample_number: 8; time: 8; latitude: 37; longitude: 49)
    Dimension coordinates:
        sample_number                             x        -            -              -
        time                                      -        x            -              -
        latitude                                  -        -            x              -
        longitude                                 -        -            -              x
    Auxiliary coordinates:
        forecast_period                           -        x            -              -
    Scalar coordinates:
        forecast_reference_time     1859-09-01 06:00:00
        height                      1.5 m
    Cell methods:
        mean                        time (6 hour)
    Attributes:
        Conventions                 'CF-1.5'
        Model scenario              'E1'
        STASH                       m01s03i236
        source                      'Data from Met Office Unified Model 6.05'

Third sample's data:
[[[296.0785827636719 296.1764221191406]
  [295.5174255371094 295.58868408203125]]

 [[296.3978576660156 296.53228759765625]
  [295.7396545410156 295.9183044433594]]

 [[-- --]
  [-- --]]

 [[295.8290710449219 295.8817443847656]
  [295.24163818359375 295.3330078125]]

 [[296.4771423339844 296.5668640136719]
  [295.7626037597656 295.797119140625]]

 [[296.40985107421875 296.557861328125]
  [295.5298156738281 295.6541442871094]]

 [[296.548583984375 296.80865478515625]
  [295.57696533203125 295.83612060546875]]

 [[298.5147399902344 298.51239013671875]
  [297.7998046875 297.839111328125]]]

```
The third sample has all points masked for the third year.

</details>

---


lib/iris/common/metadata.py
```diff
@@ -1413,6 +1411,8 @@ def metadata_filter(
         to only those that matched the given criteria.
 
     """
+    from ..util import guess_coord_axis
```
rcomer, line 1414: Moved this to address a circular import problem.

lib/iris/tests/__init__.py
```diff
@@ -141,6 +141,73 @@ def main():
         unittest.main()
 
 
+def _assert_masked_array(assertion, a, b, strict, **kwargs):
```
rcomer, line 144: Pulled these off the test class so I could use them in `pytest` style tests.  Their implementation is unchanged.

rcomer:
Ping @SimonPeatman, @philip-brohan, @wjbenfold, @Holmgren825 in case you'd like to comment on this.

lib/iris/tests/unit/util/test_mask_cube.py
```diff
@@ -40,9 +63,132 @@ def test_mask_cube_2d(self):
...
+
+    def test_mask_cube_2d_second_dim(self):
+        mask_coord = iris.coords.AuxCoord(
+            [0, 0, 1, 1], long_name="mask", units=1
+        )
+        returned = mask_cube(self.cube, mask_coord, in_place=False, dim=1)
+        self.assertOriginalMetadata(returned, "simple_2d")
+        for subcube in returned.slices("foo"):
+            # Mask should have been broadcast across "bar" dimension.
+            np.testing.assert_array_equal(subcube.data.mask, mask_coord.points)
```
bjlittle, line 128: @rcomer Did you want to add a simple test here for the case where the `points_to_mask` is a simple `numpy` array?

Bonus credits for passing the `ndarray` but with no `dim` specified and catching the traceback (which hopefully should "just happen" :wink:)

That would then cover the three use cases of the mask either being a `Cube`, `Coord` or a vanilla `ndarray` :+1:

bjlittle (changes requested):
@rcomer This is a great enhancement that builds on the capability that's available for free from `iris.common.resolve.Resolve`.

Lovely joining of the dots to make this work in such a minimalise way. I :heart: it!

Just some minor comments to service, plus if you could add a `whatsnew` entry, then I'll happily bank it in time for iris 3.3 ;-)

lib/iris/util.py
```diff
@@ -1754,29 +1756,121 @@ def find_discontiguities(cube, rel_tol=1e-5, abs_tol=1e-8):
...
+    else:
+        al = np
 
-    Returns:
+    points_to_mask = points_to_mask.astype(bool)
 
-    * result (`iris.cube.Cube`):
-        A cube whose data array is masked at points specified by input array.
+    # Treat any masked points on our mask as False.
+    points_to_mask = al.ma.filled(points_to_mask, False)
```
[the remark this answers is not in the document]
rcomer: Yes exactly.  I think most often the mask's mask would match `cube.data`'s mask.  E.g. if you had a NEMO cube (land points masked) and did something like

```python
points_to_mask = cube.data > some_threshold
```
In this case, this line obviously won't make any difference.

I don't know how likely it is that someone has different masks on their `cube.data` and `points_to_mask`, so I thought about catching that case and raising some sort of `NotImplemented`.  But that seemed more complicated than just making a decision and implementing it.

lib/iris/tests/unit/util/test__mask_array.py
```diff
@@ -0,0 +1,171 @@
...
+
+@IN_PLACE_PARAMETRIZE
+def test_trailing_mask(in_place):
+    array = ma.array(array_2by3.copy())
+    mask = np.array([0, 1, 0])
+    expected = ma.array(array_2by3, mask=[[0, 1, 0], [0, 1, 0]])
+    result = _mask_array(array, mask, in_place=in_place)
+    assert_masked_array_equal(result, expected)
+    if in_place:
+        assert_masked_array_equal(array, expected)
```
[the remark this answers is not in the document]
rcomer: Apparently the second `assert` here is a syntax error.  Have tweaked.

lib/iris/tests/unit/util/test_mask_cube.py
rcomer: The `dim` keyword is only used when passing a coordinate (if the coordinate is not attached to the cube).  For arrays, we just rely on numpy broadcasting.  This case covers broadcasting of an array-like:

https://github.com/rcomer/iris/blob/4fb113a3d11dfc66bd5d7bd6dd17c1d5fc99b53d/lib/iris/tests/unit/util/test_mask_cube.py#L87-L99

I've updated the name of it to make it clearer.

Or is there something else I've missed?

rcomer:
Thanks @bjlittle for the thorough review.  I think I've responded to everything.

lib/iris/tests/unit/util/test__mask_array.py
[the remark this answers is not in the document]
bjlittle: My bad. I was winging it :stuck_out_tongue_winking_eye:. I guess `assert` requires to be a stand alone statement :+1:

lib/iris/tests/unit/util/test_mask_cube.py
bjlittle: Nope, actually it's me misremembering. 

Just clarified with @rcomer that the `dim` kwarg is only required for coordinates, and a vanilla `ndarray` simply relies on `numpy` broadcasting :+1:

bjlittle (approved):

rcomer:
Thanks @bjlittle 🎉

Issue #4799, opened 2022-06-14
Review the Iris user experience


trexfeathers:
Raised as a 'wish-list' item in an offline meeting. User experience is usually something to think about while making some other change. Approaching UX as its own 'problem' to 'solve' offers more scope to truly improve the experience. [Vote](https://github.com/SciTools/iris/issues/4631) if you want to prioritise this 🙂

Possibly related:

* #891
* #3429

ukmo-ccbunney:
Some loose thoughts from a Dragon Taming meeting - these need scoping out a bit more:

- Tidying up the API (but minimizing user impact)
- Sharing of data objects in Iris
- Documentation, both RTD and docstring; probably already quite good but worth reviewing
   - Terminology: Eager (Xarray) vs Real (Iris) data terminology. Is in "package phrasebook", but worth highlighting
   - Update Phrasebook and highlight/publicise more widely.
- Make it easier to load complex data:
  - often easier to (at least) load using xarray
  - make Iris more lenient so we can at least load the data in some form and fix later. (Martin's #6317 will help this) 
- Cube copying with data=None is a bit confusing with dataless cubes. Align the API with other packages where possible.
- Interpolation sometimes silently fails.
   - Example for AVD Surgery: interpolation silently failing due not to mismatched coord bounds
   - Not a lot of questions on the Discussion board from external users??? Last one over a year ago. Just an observation

Going forward - we need to cast a wide net here to capture examples of where UX was suboptimal:
 - Trawling viva engage for user issues
 - Going through Iris issues
 - Reach out to known "power users".

trexfeathers:
Interesting point from @pp-mo:

While some of the tasks here are looking at creating a 'space' where UX gets the priority it deserves...

Other tasks here are about improving a specific element of UX. These sorts of thing ought to go on FOREVER - we should always be thinking about our user experience, and the ecosystem evolves too - so is it wise to have these as subtasks? If we are doing our jobs well, this would mean the Dragon would never be finished.

Should we instead be satisfied that these specific actions are getting attention thanks to the systems we are laying down such as labelling and scheduled meetings?

PR #4890, opened 2022-08-04
Improvements to Pandas-to-Iris bridge

### LINKED ISSUE #3324: Pandas bridge to support string dtype
Extend the capability of the `iris.pandas` functions to cope with string data

- `as_cube`
- `as_series`
- `as_data_frame`

trexfeathers:
### Description

Full overhaul of `iris.pandas.as_cube()`. Closes #3324.

* More Pandas-esque handling of `DataFrame`s, with each column assumed to represent a specific variable (i.e. a single `Series`).
  (Current model converts a `DataFrame` into a 2-dimensional `Cube`, treating the columns as the second dimension).
* Can specify columns that correspond to `AuxCoord`s/`CellMeasure`s/`AncillaryVariable`s.
* Can create n-dimensional `Cube`s based on index levels in the Pandas object.
  (The default index is a `RangeIndex`, which will create a 1D `Cube`).
* Can create multiple `Cube`s - one from each column not specified as dimensional metadata.
  All `Cube`s share the same dimensional metadata that is generated from the specified columns.

This is a breaking improvement, but is targeting Iris `v3.3` which cannot include breaking changes since it is a minor release. I am therefore proposing using `iris.FUTURE` to allow users to opt in.

To do:

- [x] Tests
- [x] Alignment with @hsteptoe's work in #4669
- [x] `calendar` handling like the original implementation does
- [x] Docstring updates
      Including example of how to reshape a `DataFrame` to make all data 1-dimensional
- [x] What's New entry
- [x] `Series`/`DataFrame` agnosticism
- [x] Memory efficiency
  - Are we getting NumPy views, or copying the arrays?
  - How does this handle Dask `DataFrame`s?

---


lib/iris/tests/test_pandas.py
```diff
@@ -491,13 +502,433 @@ def test_copy_true(self):
...
+            dim_coords_and_dims=[(expected_coord, 0)],
+        )
+        assert result == [expected_cube]
+
+    def test_1d_with_index(self):
+        df = self._create_pandas(index_levels=1)
+        result = iris.pandas.as_cubes(df)
+
+        expected_coord = DimCoord(df.index.values, long_name=df.index.name)
+        result_cube = result[0]
```
lbdreyer, line 581: You could make this (and the other places you've used this pattern in the tests)
```
result_cube, = result
```
So that if there are more than one cubes it raises an error and thus fails the test?

lib/iris/pandas.py
```diff
@@ -116,6 +205,290 @@ def as_cube(pandas_array, copy=True, calendars=None):
...
+            "sort_index() method before passing in."
+        )
+        raise ValueError(message)
+
+    cube_shape = getattr(pandas_index, "levshape", (pandas_index.nunique(),))
+
+    cube_kwargs = {}
+
+    def format_dimensional_metadata(dm_class_, values_, name_, dimensions_):
+        # Common convenience for to get the right DM in the right format for
```
lbdreyer, line 404: Small typo. This can just be "Common convenience to get the ..."

lib/iris/pandas.py
```diff
@@ -116,6 +205,290 @@ def as_cube(pandas_array, copy=True, calendars=None):
...
+
+    non_data_names = []
+    for dm_class, column_names, kwarg in class_arg_mapping:
+        class_kwarg = []
+        if is_series and column_names:
+            message = (
+                "The input pandas_array is a Series; ignoring "
+                f"{dm_class.__name__} columns: {column_names} ."
+            )
+            warnings.warn(message)
```
lbdreyer, line 444: Trouble with this is that the warning could be displayed up to 3 times. I'm always a bit wary of adding warnings after having heard a fair amount of complaints about them from our users. Could you rather put this if block outside the for loop and instead aggregate the checks into one warning?

lib/iris/pandas.py
```diff
@@ -116,6 +205,290 @@ def as_cube(pandas_array, copy=True, calendars=None):
     return cube
 
 
+def as_cubes(
+    pandas_array,
```
lbdreyer, line 209: I'm not so sure about the name `pandas_array`. There does appear to be a concept of a `pandas array` which is neither a Dataframe or Series. Instead of `pandas_array` how about: `pandas_data`, `pandas_object`, `pandas_data_structure`, `pandas_dataset` or something else like that?

lbdreyer (commented):
I still have some of the tests to review but I thought I would submit the comments I have so far. Nothing too major, just a few small issues

lib/iris/tests/test_pandas.py
trexfeathers: Fantastic idea, thanks!

dfabf35762b4b43fd3eab0cebcdf131619492bbc

lib/iris/pandas.py
trexfeathers: a39581fd2fc0edcc19cf863f581c88dde1349701

lib/iris/pandas.py
trexfeathers: 7e8ff414d81e77d52a12d0ff1de5e997bacb6c7f

lib/iris/pandas.py
trexfeathers: I've gone with `pandas_structure`, based on how the Pandas documentation refers to `Series` and `DataFrame`s collectively ([Intro to data structures](https://pandas.pydata.org/docs/user_guide/dsintro.html)).

5aa5af8883c07720f72cd16fa215bbb044602ab4
e59f86934a5394b2cf9d388c7569d8a95684e038

lib/iris/__init__.py
```diff
@@ -140,7 +140,7 @@ def callback(cube, field, filename):
 class Future(threading.local):
     """Run-time configuration controller."""
 
-    def __init__(self, datum_support=False):
+    def __init__(self, datum_support=False, pandas_ndim=False):
```
trexfeathers, line 143: @lbdreyer should this be documented? If so, how?

trexfeathers:
# Alignment between #4669 and #4890

* #4669
* #4890

Differences we should resolve to ensure consistent UX (UX deserves more importance - #4799).

## My proposal

Since this is quite a large list, and the `v3.3` release is imminent: I propose we get #4890's features into `main` in time for `v3.3`. #4890 was always slated for `v3.3` but we made no such promises about #4669, and once `v3.3` is out we can iterate on the feature branch to address the below points, ready for a future release. Things needed in #4890 for this:

- [ ] Decision about **Copy behaviour** (see heading below).
- [ ] Decision about **Multi-dimensional coordinates** (see heading below).
- [ ] Move the **Future flag** out into a dedicated PR.
- [ ] What's New entry

---

## Future flag

We are introducing breaking changes, but not in a major release. #4890 got lucky here, since the new functionality dictated a differently named function - `as_cubes() and therefore just had to deprecate `as_cube()`.

#4669 is changing the behaviour of `as_dataframe()`, and therefore needs to isolate the new behaviour underneath a future flag - `iris.FUTURE.pandas_ndim`. #4980 introduces said future flag, so my original plan was for #4669 to wait for #4890 to be merged; alternatively the future flag can be placed in an independent PR (we're working on a feature branch).

## Copy behaviour

#4890: `copy=True` ; #4669: `copy=False`

I would argue that `copy=True` should be the default as it is less surprising and will not cause invisible problems (where both a `Cube` and a `DataFrame` change simultaneously). `copy=True` could blow memory, but this is a _visible_ problem, and upon noticing users can consciously opt-in to `copy=False`.

## `CellMeasure`s and `AncillaryVariable`s

Only #4890 supports converting these objects. This wider scope was unfortunately always a requirement as it started with #3324, so the only acceptable solution would be to add this functionality to #4669.

## Multi-dimensional coordinates

Only #4890 supports converting these. The quicker resolution: change both PR's to have multi-dimensional coordinates skipped / throw an error. The richer resolution: add multi-dimensional support to #4669.

## Docstrings

Both should be NumPy format ([see the Iris docs](https://scitools-iris.readthedocs.io/en/latest/developers_guide/documenting/docstrings.html#docstrings)). Currently only #4890 is.

## Global attributes

#4669 has the option to turn global attributes into dedicated `DataFrame` columns. #4890 should have another `global_attribute_cols` argument for conversion in the other direction.

## Testing

#4890 has converted `test_pandas.py` to pytest-style. Should be fairly simple to align #4669, it's only a few lines.

## Example doc tests?

These PR's solve complex problems. #4890 introduces the complexity via several doctests. Consistency merits doing the same in #4669, especially for manually getting a `Series` since that is automation we are proposing to remove.

lib/iris/pandas.py
```diff
@@ -22,16 +24,18 @@
 try:
     from pandas.core.indexes.datetimes import DatetimeIndex  # pandas >=0.20
 except ImportError:
+    pass
```
lbdreyer, line 27: I'm not sure I understand the intention behind the `pass` statement. It seems like this code would work without it?

lib/iris/pandas.py
```diff
@@ -116,6 +196,291 @@ def as_cube(pandas_array, copy=True, calendars=None):
...
+    cube_kwargs = {}
+
+    def format_dimensional_metadata(dm_class_, values_, name_, dimensions_):
+        # Common convenience to get the right DM in the right format for
+        #  Cube creation.
+        calendar = calendars.get(name_)
+        instance = _get_dimensional_metadata(
+            name_, values_, calendar, dm_class_
+        )
+        return (instance, dimensions_)
```
lbdreyer, line 401: It doesn't look like you use `dimensions_` in this function, you just pass it out again, so you could remove it from here

lbdreyer (commented):
A few more comments I forgot to raise yesterday

lib/iris/tests/test_pandas.py
```diff
@@ -491,13 +502,427 @@ def test_copy_true(self):
...
+            **expected_cube_kwargs,
+        )
+        expected_cube_1 = Cube(
+            data=df[new_name].values,
+            long_name=new_name,
+            **expected_cube_kwargs,
+        )
+        assert result == [expected_cube_0, expected_cube_1]
+
+    def test_no_phenom(self):
```
lbdreyer, line 756: Should we also care about an empty dataframe. Currently we get:
```
>>> my_df = pd.DataFrame({})
>>> result = iris.pandas.as_cubes(my_df)
  File ".../code.py", line 90, in runcode
    exec(code, self.locals)
  File "<input>", line 1, in <module>
  File "/net/data/users/ldreyer/dev/iris/lib/iris/pandas.py", line 419, in as_cubes
  File ".../iris/lib/iris/pandas.py", line 407, in format_dimensional_metadata
  File ".../iris/lib/iris/pandas.py", line 62, in _get_dimensional_metadata
    coord = AuxCoord(points, units=units)
  File ".../site-packages/cf_units/__init__.py", line 1970, in date2num
    return cftime.date2num(date, self.cftime_unit, self.calendar)
  File "src/cftime/_cftime.pyx", line 287, in cftime._cftime.date2num
IndexError: list index out of range
````

Comment (trexfeathers) [reply lib/iris/tests/test_pandas.py:756]: Oooh good catch! Do you think we should:

a. Throw an `Error`?
b. Return an empty `CubeList`?

File: lib/iris/pandas.py (hunk 7 above)
Comment (trexfeathers) [reply lib/iris/pandas.py:401]: Part of the common convenience of `format_dimensional_metadata()` is getting the objects out in the correct format, since I do this more than once in the main code. That includes placing the object and its dimensions into a `tuple`.

Is that a bit strange? Should I just be creating the appropriate `tuple` each time, outside this function?

File: lib/iris/pandas.py (hunk 6 above)
Comment (trexfeathers) [reply lib/iris/pandas.py]: Don't worry that was just me being a 🔧

9391f432ec987b3e6f948c1e5cca3ec679777e23

File: lib/iris/pandas.py [hunk 9]
```diff
@@ -116,6 +205,291 @@ def as_cube(pandas_array, copy=True, calendars=None):
...
+
+    >>> my_series = Series([300, np.NaN, 302], name="air_temperature")
+    >>> converted_cube = as_cubes(my_series)[0]
+    >>> print(converted_cube.data)
+    [300.  nan 302.]
+    >>> converted_cube.data = np.ma.masked_invalid(converted_cube.data)
+    >>> print(converted_cube.data)
+    [300.0 -- 302.0]
+
+    If the :class:`~pandas.DataFrame` uses columns as as a second dimension,
```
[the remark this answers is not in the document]
trexfeathers: 5f9531fd52d82e3f5a189f5506180e0f21d6846e

lib/iris/tests/test_pandas.py
trexfeathers: We went for an empty `CubeList` - 9619fd065f381c89e58f1383a2a2ccbced3adfdc

Is the pattern of using an early `return` OK? I know we typically `return result` having conditionally assigned a value to `result`, but I didn't like the idea of almost everything being within an `if` block.

lib/iris/pandas.py
lbdreyer: I see, that makes sense. I think it's fine to keep it as it is. As you say you are getting it in the right format. I had actually briefly wondered how best to write that if it weren't included in this function. I think it makes sense to do it here

lbdreyer (approved):
All the issues I found have been addressed so I would say this is good to go :rocket:

trexfeathers:
Thanks for wading through this @lbdreyer 🎖

PR #4841, opened 2022-06-29
adopt setuptools-scm for auto versioning

bjlittle:
### Description
This PR adopts `setuptools-scm` to automate the versioning of `iris`.

This means that we no longer need to care about bumping the `iris.__version__` manually for a release, or post-release to a dev version.

Requires #4840 to be merged first, in order to include the relevant updates to the requirements `yml` and `lock` files.

Also, once this is merged, then we are in the position to automate the routine creation, testing and publishing of `iris` source and binary PyPI wheels, as a GHA.

Reference: see [setuptools-scm](https://github.com/pypa/setuptools_scm)

TODO:
- [ ] Update the developer documentation
- [ ] Update `py310` requirements (#4840)
- [ ] Include re-rendered `conda-lock` files

---
[Consult Iris pull request check list]( https://scitools-iris.readthedocs.io/en/latest/developers_guide/contributing_pull_request_checklist.html)

PR #4892, opened 2022-08-04
set default 'unknown' version

bjlittle:
### Description
Ever since we committed #4841, as developers we require the [setuptools-scm](https://github.com/pypa/setuptools_scm) generated `iris._version` file to be available to garner the `iris.__version__`.

If this file is **not** available, then the following traceback will be encountered by the developer whenever `import iris` is performed:

```bash
ModuleNotFoundError: No module named 'iris._version'
```

Thanks to our adoption of a [PEP-0518](https://peps.python.org/pep-0518/) build backend, as a developer you do **not** require to have the `setuptools-scm` package installed within your development environment to use `iris` (only do that if you care about getting the correct version). However, to generate the `iris._version` file without explicitly installing `setuptools-scm`, from the root source directory of `iris` simply:

```bash
> pip install --no-deps --editable .
```

It seems a shame to force developers to have to perform this step. Other packages which have adopted `setuptools-scm` default to the pattern of setting the `__version__ = "unknown"` to circumvent this situation as a convenience.

This PR introduces this pattern to `iris` as a similar developer convenience.

Ping @hsteptoe

---

### DIFF
--- a/lib/iris/__init__.py
+++ b/lib/iris/__init__.py
@@ -101,7 +101,12 @@ import iris.config
 import iris.io
 
 from ._deprecation import IrisDeprecation, warn_deprecated
-from ._version import version as __version__  # noqa: F401
+
+try:
+    from ._version import version as __version__  # noqa: F401
+except ModuleNotFoundError:
+    __version__ = "unknown"
+
 
 try:
     import iris_sample_data


lbdreyer (approved):
Seems sensible to me to have iris.__version__ return at least something rather than just error, in these cases. Thanks @bjlittle !

PR #4901, opened 2022-08-11
Lockfile update and skip questionable test

rcomer:
### Description

Closes #4893 

This is the second time recently that this test has failed due to a dependency update.  At #4331, there was a fair bit of effort to understand the cause of the change in output values.  There it was noted that the original values were just what was produced originally, and it was unclear whether the failure indicated something going wrong or something having been fixed.  More recently, the module containing the code in question was deprecated at Iris3.2 ([point 8 here](https://scitools-iris.readthedocs.io/en/stable/whatsnew/3.2.html#deprecations)).

Several developers therefore agreed at https://github.com/SciTools/iris/pull/4893#issuecomment-1210441603 that this test could be skipped until the deprecated API is removed.

---


bjlittle (approved):

rcomer:
Thanks @bjlittle.

PR #4905, opened 2022-08-15
SUM, COUNT, PROPORTION: don't use dask on numpy arrays

rcomer:
### Description

We have a few aggregators that are implemented purely in dask, and the "real" version works by converting the numpy array to a dask array, passing it through the dask functions, and then realising the data.  This appears to have been done at #3168 in response to a review request to reduce code repetition.  However, it is slow compared to just working directly with numpy.  Consider the following:

```python
import timeit

import iris.analysis
import numpy as np

def get_array():
    return np.random.random(10000).reshape(100, 100)

def over_half(a):
    return a > 0.5

setup = """
import iris
from __main__ import get_array, over_half
arr = get_array()
"""

print("COUNT", timeit.timeit(
    'iris.analysis.COUNT.aggregate(arr, 1, function=over_half)',
    setup=setup, number=500))

print("PROPORTION", timeit.timeit(
    'iris.analysis.PROPORTION.aggregate(arr, 1, function=over_half)',
    setup=setup, number=500))  

print("SUM", timeit.timeit(
    'iris.analysis.SUM.aggregate(arr, 1)',
    setup=setup, number=500))
```

On my local machine with `main`, this gives
```
COUNT 1.7456789941061288
PROPORTION 1.7775695940945297
SUM 1.26268538297154
```
With my branch it gives
```
COUNT 0.008400967926718295
PROPORTION 0.014432693016715348
SUM 0.004979556077159941
```

Note that https://github.com/dask/dask/pull/6393 introduced the `__array_function__` and `__array_ufunc__` methods to dask arrays so that, since [it was included at version 2.26](https://docs.dask.org/en/stable/changelog.html#id137), `numpy.sum` will automatically use `dask.array.sum` if it finds a dask array.

```python
import numpy
import dask
import dask.array as da

for package in [numpy, dask]:
    print(package.__name__, package.__version__)
    
array = da.array(range(5))
print(numpy.sum(array))
```
```
numpy 1.19.5
dask 2.26.0
dask.array<sum-aggregate, shape=(), dtype=int64, chunksize=(), chunktype=numpy.ndarray>
```
I have therefore updated the dask minimum pin to version 2.26.

One other aggregator, `MAX_RUN` is also implemented in the dask-only way.  However, since it's rather more complicated than these, I've left it for now.  Could be investigated separately.

---

### DIFF
--- a/docs/src/whatsnew/latest.rst
+++ b/docs/src/whatsnew/latest.rst
@@ -32,14 +32,15 @@ This document explains the changes made to Iris for this release
 ===========
 
 #. `@ESadek-MO`_ edited :func:`~iris.io.expand_filespecs` to allow expansion of
-    non-existing paths, and added expansion functionality to :func:`~iris.io.save`.
-    (:issue:`4772`, :pull:`4913`)
+   non-existing paths, and added expansion functionality to :func:`~iris.io.save`.
+   (:issue:`4772`, :pull:`4913`)
 
 
 🐛 Bugs Fixed
 =============
 
-#. N/A
+#. `@rcomer`_ factored masking into the returned sum-of-weights calculation from
+   :obj:`~iris.analysis.SUM`. (:pull:`4905`)
 
 
 💣 Incompatible Changes
@@ -51,7 +52,9 @@ This document explains the changes made to Iris for this release
 🚀 Performance Enhancements
 ===========================
 
-#. N/A
+#. `@rcomer`_ increased aggregation speed for :obj:`~iris.analysis.SUM`,
+   :obj:`~iris.analysis.COUNT` and :obj:`~iris.analysis.PROPORTION` on real
+   data. (:pull:`4905`)
 
 
 🔥 Deprecations
@@ -63,7 +66,8 @@ This document explains the changes made to Iris for this release
 🔗 Dependencies
 ===============
 
-#. N/A
+#. `@rcomer`_ introduced the ``dask >=2.26`` minimum pin, so that Iris can benefit
+   from Dask's support for `NEP13`_ and `NEP18`_. (:pull:`4905`)
 
 
 📚 Documentation
@@ -89,3 +93,5 @@ This document explains the changes made to Iris for this release
     Whatsnew resources in alphabetical order:
 
 
+.. _NEP13: https://numpy.org/neps/nep-0013-ufunc-overrides.html
+.. _NEP18: https://numpy.org/neps/nep-0018-array-function-protocol.html
--- a/docs/src/whatsnew/latest.rst
+++ b/docs/src/whatsnew/latest.rst
@@ -39,8 +39,8 @@ This document explains the changes made to Iris for this release
 🐛 Bugs Fixed
 =============
 
-#. `@rcomer`_ factored masking into the returned sum-of-weights calculation from
-   :obj:`~iris.analysis.SUM`. (:pull:`4905`)
+#. `@rcomer`_ and `@pp-mo`_ (reviewer) factored masking into the returned
+   sum-of-weights calculation from :obj:`~iris.analysis.SUM`. (:pull:`4905`)
 
 
 💣 Incompatible Changes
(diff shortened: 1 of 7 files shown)


lib/iris/analysis/__init__.py
```diff
@@ -1604,23 +1601,29 @@ def _lazy_rms(array, axis, **kwargs):
...
+                # Dask version of ones_like does not preserve masks. See dask#9301.
+                weights = da.ma.masked_array(
+                    weights, da.ma.getmaskarray(array)
+                )
         else:
-            weights = weights_in
-        rvalue = (wsum, da.sum(weights, axis=axis_in))
+            weights = al.ma.masked_array(
+                weights_in, mask=al.ma.getmaskarray(array)
+            )
```
rcomer, line 1625: Not relevant to the main point of this PR, but ignoring the mask here seemed like a bug to me.  See the new tests.

rcomer: See also equivalent handling for `average` in [numpy.ma](https://github.com/numpy/numpy/blob/0b90540d2086efb90b29af1faf2cbf6387b8ed59/numpy/ma/extras.py#L644-L648) and [dask.array](https://github.com/dask/dask/blob/8b95f983c232c1bd628e9cba0695d3ef229d290b/dask/array/routines.py#L2476-L2480).

lib/iris/analysis/__init__.py
```diff
@@ -1740,9 +1743,9 @@ def interp_order(length):
 #
 COUNT = Aggregator(
     "count",
-    iris._lazy_data.non_lazy(_lazy_count),
+    _count,
     units_func=lambda units: 1,
-    lazy_func=_lazy_count,
+    lazy_func=_build_dask_mdtol_function(_count),
```
pp-mo, line 1748: I think it also needs explaining that `_count` can function as a (lazy-preserving) dask operation as well, 
purely due to _how_ it is implemented.
Or possibly, that it is better said where it is defined ?

lib/iris/analysis/__init__.py
```diff
@@ -2114,8 +2117,8 @@ def interp_order(length):
 
 SUM = WeightedAggregator(
     "sum",
-    iris._lazy_data.non_lazy(_lazy_sum),
-    lazy_func=_build_dask_mdtol_function(_lazy_sum),
+    _sum,
+    lazy_func=_build_dask_mdtol_function(_sum),
```
pp-mo, line 2121: As for `_count`, worth a note explaining that `_sum` must and can act as a lazy operator.

pp-mo (changes requested):

pp-mo:
I haven't asked for any further changes, but I must say I am still disquieted ...
While I don't think these changes _add_ significant complexity, they have been really rather hard to understand.

So I made a related issue about improving this #4915

lib/iris/analysis/__init__.py
```diff
@@ -1499,18 +1499,21 @@ def _weighted_percentile(
...
 
 
-@_build_dask_mdtol_function
-def _lazy_count(array, **kwargs):
-    array = iris._lazy_data.as_lazy_data(array)
+def _count(array, **kwargs):
+    """
+    Counts the number of points along the axis that satisfy the condition
+    specified by ``function``.  Uses Dask's support for NEP13/18 to work as
+    either a lazy or a real function.
```
rcomer, line 1506: Though it occurs to me that this rather depends whether the passed `function` can cope with lazy data.  Probably always did.

rcomer:
Thanks @pp-mo I have added some docstrings.  Should I also add a whatsnew or do we wait until after the release dust has settled?

lib/iris/analysis/__init__.py
pp-mo: Yes I agree. It's possibly a bit of a hole, but it's no worse that it was before
:crossed_fingers:

pp-mo:
> @rcomer Should I also add a whatsnew or do we wait until after the release dust has settled?

Thanks @rcomer.  Please add a whatsnew (I forgot that again !)
But I will anyway hold off merge until after the v3.3 release (which is imminent, hopefully today)

rcomer:
Thanks @pp-mo.  I have now rebased and whatsnewed.

docs/src/whatsnew/latest.rst
```diff
@@ -32,14 +32,15 @@ This document explains the changes made to Iris for this release
 ===========
 
 #. `@ESadek-MO`_ edited :func:`~iris.io.expand_filespecs` to allow expansion of
-    non-existing paths, and added expansion functionality to :func:`~iris.io.save`.
-    (:issue:`4772`, :pull:`4913`)
+   non-existing paths, and added expansion functionality to :func:`~iris.io.save`.
+   (:issue:`4772`, :pull:`4913`)
```
rcomer, line 36: This isn't currently rendering quite right. https://scitools-iris.readthedocs.io/en/latest/whatsnew/latest.html#features

trexfeathers: Oh that's why! @ESadek-MO and I couldn't figure it out!

pp-mo (approved):

PR #4483, opened 2022-01-07
Standard names table update.

pp-mo:
Closes #4479
Updated standard names, in advance of 3.2 release.

Was considering also the metadata translations, and https://github.com/SciTools/iris/issues/3950
But I have been reminded that standard-names is not really a part of that, so I'm going to deal with that [separately](https://github.com/SciTools/iris/pull/4484).

PR #4910, opened 2022-08-17
Standard names table update (v79)

stephenworsley:
As in #4483, updates the standard names in advance of 3.3 release.


lbdreyer (approved):
Looks good!

PR #4913, opened 2022-08-17
Tilde's and wildcard recognition within iris.save.

### LINKED ISSUE #4772: Extend `expanduser` behaviour
## ✨ Feature Request

`iris.save` should expand `~` to e.g. `/home/wjbenfold`

## Motivation

I'm surprised when an iris save operation fails with a `PermissionError` because I've passed in a path starting with `~`, particularly having just successfully loaded a file from a path starting with `~`. Making this consistent would make using Iris easier and less surprising.

ESadek-MO:
## 🚀 Tilde's and wildcard recognition within iris.save.

Closes #4772.

### Description
Saving a file to a directory path that includes `~` or a wildcard will now expand the file path; e.g. `iris.save(cube, ~/file.pp)` will save to `HOME/file.pp`. Provisional testing written.


lib/iris/tests/unit/io/test_save.py
```diff
@@ -18,6 +18,7 @@
 
 class TestSave(tests.IrisTest):
     def test_pathlib_save(self):
+        # DOESN'T CHECK TILDES WORK DUE TO MOCK.
```
trexfeathers, line 21: @pp-mo do you have any opinions on what should happen here? Not testing expansion feels like an unfortunate oversight but it sounds like any alternative would be a big project?

lib/iris/tests/unit/io/test_expand_filespecs.py
```diff
@@ -96,6 +96,12 @@ def test_files_and_none(self):
 
         self.assertMultiLineEqual(str(err.exception), expected)
 
+    def test_false_bool(self):
+        tempdir = self.tmpdir
+        msg = os.path.join(tempdir, "no_exist.txt")
+        result = iio.expand_filespecs([msg], False)[0]
```
trexfeathers, line 102:
```suggestion
        (result,) = iio.expand_filespecs([msg], False)
```

lib/iris/tests/unit/io/test_expand_filespecs.py
```diff
@@ -96,6 +96,12 @@ def test_files_and_none(self):
 
         self.assertMultiLineEqual(str(err.exception), expected)
 
+    def test_false_bool(self):
```
trexfeathers, line 99: Could we split this into something equivalent to `test_absolute_path()` and `test_relative_path()`? Just to confirm that the expansion functionality still works with `files_expected=False`.

trexfeathers (changes requested):
Nice job @ESadek-MO! Some comments for you, and this will need a What's New entry - for which you will need to merge in `main` again(!) to get #4924.

lib/iris/tests/unit/io/test_expand_filespecs.py
```diff
@@ -96,6 +96,12 @@ def test_files_and_none(self):
...
+        tempdir = self.tmpdir
+        msg = os.path.join(tempdir, "no_exist.txt")
+        result = iio.expand_filespecs([msg], False)[0]
+        self.assertEqual(result, msg)
+
```
pp-mo, line 104: I think there should be a testcase here that confirms tilde expansion.
I think you can do that, probably, by patching the value of `os.environ['HOME']` to ensure a consistent answer.

lib/iris/tests/unit/io/test_save.py
pp-mo: I think that limitation **_is_** ok, because here we are not testing that part of the behaviour.  Instead, we are just checking that the different valid types of argument are all processed OK.
The expansion behaviour can be tested elsewhere.

On the other hand,  I'm not convinced it actually is tested.  I think there should really be a testcase  in test_expand_filespecs to do that.  
See [other comment there](https://github.com/SciTools/iris/pull/4913/files#r951477688)

trexfeathers (approved):

PR #4918, opened 2022-08-18
Avoid using mutable default.

pp-mo:
I noticed a tiny problem with the recently-merged extension to new-axis, in that it is using a mutable default argument.

As it is, the routine never modifies the argument, and in fact always replaces it [here](https://github.com/SciTools/iris/blob/505dca8ca5d7fbe5b45d2e24180280cd62ebfab8/lib/iris/util.py#L1180-L1182).
But I think it's accepted that this is not good practice generally.

---

### DIFF
--- a/lib/iris/util.py
+++ b/lib/iris/util.py
@@ -1096,7 +1096,7 @@ def format_array(arr):
     return result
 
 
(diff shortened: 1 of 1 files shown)


lbdreyer (approved):
Nice catch, thanks @pp-mo !

PR #4919, opened 2022-08-18
Remove apparently obsolete code in iris.save

pp-mo:
### Description

---


pp-mo:
Result : it is not actually obsolete !
Though, it seems only used for DOT+PNG at present, which we might well want to drop.

PR #4921, opened 2022-08-18
Update whatsnew for v3.3.0rc0

stephenworsley:
Updates the location of the whatsnew and adds release highlights.

PR #4924, opened 2022-08-22
Update whatsnew

stephenworsley:
Reintroduce the whatsnew latest and template removed in #4921.

### DIFF
--- a/docs/src/whatsnew/latest.rst.template
+++ b/docs/src/whatsnew/latest.rst.template
@@ -1,0 +1,112 @@
+.. include:: ../common_links.inc
+
+|iris_version| |build_date| [unreleased]
+****************************************
+
+This document explains the changes made to Iris for this release
+(:doc:`View all changes <index>`.)
+
+
+.. dropdown:: :opticon:`report` |iris_version| Release Highlights
+   :container: + shadow
+   :title: text-primary text-center font-weight-bold
+   :body: bg-light
+   :animate: fade-in
+   :open:
+
+   The highlights for this major/minor release of Iris include:
+
+   * N/A
+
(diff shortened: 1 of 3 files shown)


docs/src/whatsnew/latest.rst.template
```diff
@@ -0,0 +1,112 @@
...
+
+   * N/A
+
+   And finally, get in touch with us on :issue:`GitHub<new/choose>` if you have
+   any issues or feature requests for improving Iris. Enjoy!
+
+
+NOTE: section below is a template for bugfix patches
+====================================================
+    (Please remove this section when creating an initial 'dev.rst')
```
trexfeathers, line 27:
```suggestion
    (Please remove this section when creating an initial 'latest.rst')
```

I think?

docs/src/whatsnew/latest.rst.template
```diff
@@ -0,0 +1,112 @@
...
+   :body: bg-light
+   :animate: fade-in
+
+   The patches in this release of Iris include:
+
+   #. N/A
+
+NOTE: section above is a template for bugfix patches
+====================================================
+    (Please remove this section when creating an initial 'dev.rst')
```
trexfeathers, line 44:
```suggestion
    (Please remove this section when creating an initial 'latest.rst')
```

I think?

trexfeathers (changes requested):
Thanks @stephenworsley! Couple of points below, and I think you need a reference to `latest.rst` in `index.rst`, too.

docs/src/whatsnew/latest.rst.template
stephenworsley: that seems more accurate now.

trexfeathers (approved):

Issue #4912, opened 2022-08-17
Unpin shapely


stephenworsley:
#4911 has pinned shapely due to an error described here https://github.com/SciTools/cartopy/issues/2067 which caused errors in our documentation. This pin should be removed when this issue has been solved in cartopy or shapely.

trexfeathers:
Closed in #4925

trexfeathers:
> Closed in #4925

No it hasn't, we'll still need to do a merge-back from the `v3.3.x` branch.

pp-mo:
> > Closed in #4925
> 
> No it hasn't, we'll still need to do a merge-back from the `v3.3.x` branch.

Done! #4944

PR #4925, opened 2022-08-23
Unpin shapely

stephenworsley:
Addresses #4912 since the latest version of shapely seems to have fixed this issue.


trexfeathers (approved):
🤞

PR #4926, opened 2022-08-23
Better handle `setuptools_scm` versioning

trexfeathers:
### Description

* Uses the correct `version_scheme` setting for our release branch strategy.
* Properly copes with the conflict between `setuptools_scm` and Read the Docs.
* Changes the RTD version badge to better reflect all fixed-version builds.

[Here's a rendered example for the fictitious `5.6.9` version](https://trexfeathers-iris.readthedocs.io/en/5.6.9/)

Here are the five badges I'm proposing:
- latest development version-5.6.10dev49-gold?style=flat) - whenever RTD labels a build as `latest`.
- stable version-5.6.9-green?style=flat) - whenever RTD labels a build as `stable`.
- archived version (not 'stable' or 'latest')-5.6.8-red?style=?style=flat) - all other builds of tags, which we typically only use for the release tags.
- non-tagged branch commit - everything else built by RTD, which is currently just any branches that we mark for building, and in future could include PR's too.
- non-tagged development commit - all non-RTD builds.

---

### DIFF
--- a/docs/src/_templates/custom_sidebar_logo_version.html
+++ b/docs/src/_templates/custom_sidebar_logo_version.html
@@ -7,15 +7,20 @@
         <span title="stable version">
             <img src="https://img.shields.io/badge/✨️_version_('stable')-{{ version }}-green?style=flat">
         </span>
-    {% else %}
-        {# Covers builds for any specific versions/commits, including RC's. #}
+    {% elif rtd_version_type == 'tag' %}
+        {# Covers builds for specific tags, including RC's. #}
         <span title="archived version (not 'stable' or 'latest')">
             <img src="https://img.shields.io/badge/🔒_version_(archived)-{{ version }}-red?style=?style=flat">
+        </span>
+    {% else %}
+        {# Anything else build by RTD will be the HEAD of an activated branch #}
+        <span title="non-tagged branch commit">
+            <img src="https://img.shields.io/badge/🚧_branch-{{ rtd_version }}_@_{{ commit_sha }}-blue?style=?style=flat">
         </span>
     {% endif %}
 {%- else %}
     {# not on rtd #}
-    <span title="development version">
-        <img src="https://img.shields.io/badge/✏️_version_(dev)-{{ version }}-blue?style=?style=flat">
+    <span title="non-tagged development commit">
+        <img src="https://img.shields.io/badge/🚧_commit_SHA-{{ commit_sha }}-blue?style=?style=flat">
     </span>
 {%- endif %}
--- a/docs/src/_templates/custom_sidebar_logo_version.html
+++ b/docs/src/_templates/custom_sidebar_logo_version.html
@@ -1,20 +1,21 @@
 {% if on_rtd %}
     {% if rtd_version == 'latest' %}
         <span title="latest development version">
-            <img src="https://img.shields.io/badge/latest_version-{{ version }}-orange?style=flat">
+            <img src="https://img.shields.io/badge/⚠️_version='latest'-{{ version }}-gold?style=flat">
         </span>
     {% elif rtd_version == 'stable' %}
         <span title="stable version">
-            <img src="https://img.shields.io/badge/stable_version-{{ version }}-green?style=flat">
+            <img src="https://img.shields.io/badge/✨️_version='stable'-{{ version }}-green?style=flat">
         </span>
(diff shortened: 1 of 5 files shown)


trexfeathers:
My demo isn't working! Will look tomorrow...

trexfeathers:
[Here's a build without a version tag](https://trexfeathers-iris.readthedocs.io/en/better_versioning/). This demonstrates `release-branch-semver` working - builds not on release branches automatically increment the minor version (better than our current implementation that has incremented to `v3.3.0rc1...`).

docs/src/_templates/custom_sidebar_logo_version.html
```diff
@@ -1,20 +1,21 @@
...
+        {# Covers builds for any specific versions/commits, including RC's. #}
+        <span title="archived version (not 'stable' or 'latest')">
+            <img src="https://img.shields.io/badge/🔒_version_(archived)-{{ version }}-red?style=?style=flat">
         </span>
     {% endif %}
 {%- else %}
     {# not on rtd #}
     <span title="development version">
-        <img src="https://img.shields.io/badge/dev_version-{{ version }}-blue?style=?style=flat">
+        <img src="https://img.shields.io/badge/✏️_version_(dev)-{{ version }}-blue?style=?style=flat">
```
pp-mo, line 19: Regarding the sample "build without a version tag"
(from [comment]),
I don't understand why that is giving "version (archived) 5.7.0.dev1"
- I think saying "archived" it makes it look like a definite endpoint (like a post-release) instead of a moving goalpost -- e.g. what would 5.7.0.dev2 then be?

So. I was expecting that to come here + be labelled "version(dev)"
But I expect I misunderstood something ...

pp-mo (changes requested):
Still want to discuss before merging

docs/src/_templates/custom_sidebar_logo_version.html
```diff
@@ -1,20 +1,21 @@
...
     {% else %}
-        <span title="old version (not stable or latest)">
-            <img src="https://img.shields.io/badge/old_version-{{ version }}-red?style=?style=flat">
+        {# Covers builds for any specific versions/commits, including RC's. #}
+        <span title="archived version (not 'stable' or 'latest')">
+            <img src="https://img.shields.io/badge/🔒_version_(archived)-{{ version }}-red?style=?style=flat">
         </span>
     {% endif %}
 {%- else %}
     {# not on rtd #}
```
pp-mo, line 17: What does "not on rtd" really mean here ?

docs/src/_templates/custom_sidebar_logo_version.html
trexfeathers: When building via Read the Docs:

- `latest` = a dynamic label referring to whatever is the latest commit on `main`.
- `stable` = a dynamic label referring to the latest non-RC semantic-versioned tag.
- Any RTD build is static. Not only does this include all semantic-versioned tags (i.e. releases), but also any other tag/branch that has been selected for an RTD build.

Since these static builds never get rebuilt, most of them will be out of date. But this isn't consistently true - the most recent non-RC semantic-versioned build will correspond with `stable`; and the only way to access release candidate documentation is via its static build. So we want to make the reader aware that they COULD be looking at out of date documentation, or that it might be up to date right now but will go out of date in future. `archived` seemed like the best word (I also considered `fixed` but that seemed open to confusion about bug fixes).

docs/src/_templates/custom_sidebar_logo_version.html
trexfeathers: If you build the docs locally. I find the distinction of having a `dev` badge here quite helpful.

docs/src/_templates/custom_sidebar_logo_version.html
trexfeathers: Note that we don't get an opportunity to re-label any static builds when they become out of date in future. We need a badge that will be appropriate now and in the future.

trexfeathers: Here's the outcome of further conversations with @pp-mo: 16c899e2249ab4b492fbb1d12a13467425eb9d23

See the main description for examples of the badges.

trexfeathers:
Thanks @stephenworsley and @pp-mo!

Issue #4898, opened 2022-08-09
Misleading error message from Cube.coord


rcomer:
## 🐛 Bug Report

I just spent some time, with help from colleagues, trying to figure out why my data didn't have a realization coordinate.  It turned out that it did have a realization coordinate, but the error message was misleading.

## How To Reproduce
```python
import iris
import iris.coords
import iris.cube

print(f"{iris.__version__=}")

cube = iris.cube.Cube(1)
coord = iris.coords.DimCoord(1, standard_name="realization", units=1)

cube.add_aux_coord(coord.copy())

coord.var_name = "realization"  # without this line we get no exception.

cube.coord(coord)
```
```
iris.__version__='3.2.1.post0'
Traceback (most recent call last):
  File "coord_matching.py", line 14, in <module>
    cube.coord(coord)
  File "[python-path]/python3.8/site-packages/iris/cube.py", line 1919, in coord
    raise iris.exceptions.CoordinateNotFoundError(emsg)
iris.exceptions.CoordinateNotFoundError: "Expected to find exactly 1 'realization' coordinate, but found none."
```

## Expected behaviour

The error message should be clearer about what it is trying to match.  Or it should make a more lenient comparison between the coordinates.

## Environment 
 - OS & Version: RHEL7
 - Iris Version: 3.2.1.post0

stephenworsley:
I've put up a very simple PR to make it more obvious that an object is being compared against rather than a string. A more verbose error message could potentially give more detail, but I think it would be a bit more complicated to do so there could be diminishing returns going down that route. Let me know if you think these error messages would have been sufficient.

pp-mo:
> I've put up a very simple PR 

I.E. https://github.com/SciTools/iris/pull/4928 ??

PR #4928, opened 2022-08-24
Improve error messages when comparing against objects vs strings

stephenworsley:
A simple change to error messages to address #4898.

### DIFF
--- a/lib/iris/cube.py
+++ b/lib/iris/cube.py
@@ -1991,8 +1991,8 @@ class Cube(CFVariableMixin):
                 if not isinstance(name_or_coord, str):
                     _name = name_or_coord.name()
                     emsg = (
-                        "Expected to find exactly 1 coordinate matching the "
-                        f"{_name!r} coordinate, but found none."
+                        "Expected to find exactly 1 coordinate matching the given "
+                        f"{_name!r} coordinate's metadata, but found none."
                     )
                     raise iris.exceptions.CoordinateNotFoundError(emsg)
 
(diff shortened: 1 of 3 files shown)


lib/iris/cube.py
```diff
@@ -1984,6 +1984,12 @@ def coord(
             if name_or_coord is not None:
                 if not isinstance(name_or_coord, str):
                     _name = name_or_coord.name()
+                    emsg = (
+                        "Expected to find exactly 1 coordinate matching the "
+                        f"{_name!r} coordinate, but found none."
```
rcomer, line 1989:
```suggestion
                        "Expected to find exactly 1 coordinate matching the given"
                        f"{_name!r} coordinate's metadata, but found none."
```
Definitely an improvement but I wonder if we can be even more explicit.  I _think_ it's just the metadata that has to match, so you can have different points and bounds.

rcomer (commented):
Wow, thanks for the fast work @stephenworsley!  I agree that trying to list out exactly what didn't match might be more trouble than it's worth, but I have a suggestion to be a bit more explicit for the coord case.

rcomer (commented):
Hi @stephenworsley, this looks good to me.  Some very minor suggestions  in line.  I also wonder if it is worth updating the relevant tests to use [assertRaisesRegex](https://docs.python.org/3/library/unittest.html#unittest.TestCase.assertRaisesRegex) and check the message:

[Can't seem to find them for `cube.coord` !]

https://github.com/SciTools/iris/blob/d56824cbc892356302c135ee23d704ff01e12d71/lib/iris/tests/unit/cube/test_Cube.py#L2670-L2678

https://github.com/SciTools/iris/blob/d56824cbc892356302c135ee23d704ff01e12d71/lib/iris/tests/unit/cube/test_Cube.py#L2612-L2620

rcomer (approved):
Thanks @stephenworsley.  Just waiting for the CI...

rcomer:
Seems docs are failing because https://docs.scipy.org/ is down, so the intersphinx mapping doesn't work.  I'm inclined to merge anyway: the only doc change here is the whatsnew entry and it was fine before rebase.

PR #4929, opened 2022-08-24
Support dask arrays in binary operations

rcomer:
### Description

Currently, the various arithmetic functions state that their second operand may be a cube, a coordinate, a numpy array or a number.  There is nothing to stop you passing a dask array as the second operand, but it will be silently realised by the `numpy.asanyarray` call, which seems unhelpful to me.  I think it would be good to explicitly support passing a dask array, though there may be a better way to do it than my current change.

~In draft for now because a bunch of docstrings would need modifying and some tests adding, but I wanted to see if there is agreement in principle first.~

---

### DIFF
--- a/docs/src/conf.py
+++ b/docs/src/conf.py
@@ -235,6 +235,7 @@ templates_path = ["_templates"]
 # See https://www.sphinx-doc.org/en/master/usage/extensions/intersphinx.html
 intersphinx_mapping = {
     "cartopy": ("https://scitools.org.uk/cartopy/docs/latest/", None),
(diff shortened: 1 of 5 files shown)


rcomer:
Since nobody has said otherwise, I went ahead and added a test and updated the docstrings.  While I was in the arithmetic docstrings, I updated them to make them more consistent with each other and also applied the numpydoc style.

Ready for review I think.

bjlittle (approved):
@rcomer Lovely addition, thanks! :100:

PR #4930, opened 2022-08-25
Mergeback (v3.3.x)

stephenworsley:
Update main with changes to v3.3.x.

### NOTE: Should be merged without squashing.


trexfeathers (changes requested):
This looks great. I've just asked @stephenworsley to also regenerate the lock-files, since both `v3.3.x` and `main` have had their own lock-file updates and I don't want to risk any confusion.

trexfeathers:
> This looks great. I've just asked @stephenworsley to also regenerate the lock-files, since both `v3.3.x` and `main` have had their own lock-file updates at different times, and I don't want to risk any confusion.

From further discussion: we have worked out that the lock-files on `v3.3x` were updated more recently than `main`, so there is no risk. Will merge as is.

trexfeathers (approved):

PR #4932, opened 2022-08-25
Fix whatsnew

stephenworsley:
Fixes some links in the whatsnew.


trexfeathers (approved):

stephenworsley:
> Guess this will need a merge-back as well?

Eventually, I don't think this is as urgent as the other changes we merged back so I feel like that could wait 'till the 3.3.0 release.

trexfeathers:
> > Guess this will need a merge-back as well?
> 
> Eventually, I don't think this is as urgent as the other changes we merged back so I feel like that could wait 'till the 3.3.0 release.

I ask because without a merge-back this change won't appear in the `latest` documentation (only the hidden `3.3.x` documentation).

stephenworsley:
The release notes state that the merge back should happen after the release is published. I interpret that to mean it's fince to leave relatively minor things like this 'till then.

Issue #3288, opened 2019-03-04
iris.save crashes for NetCDF for cube with no cell_methods


volcan01010:
I have cube where `cube.cell_methods == (None,)`.  When I try to save it to NetCDF, Iris crashes with the following exception:

```
-> 1728             for name in cm.coord_names:
   1729                 coord = cube.coords(name)
   1730 

AttributeError: 'NoneType' object has no attribute 'coord_names'
```

The error occurs within the [NetCDF `Saver._create_cf_cell_methods()` function](https://github.com/SciTools/iris/blob/45a18abc1b87a43a1373212ab4205713fe92277c/lib/iris/fileformats/netcdf.py#L1702).

I think this is a bug.  Unfortunately I don't know enough about NetCDF to suggest a fix.

The following code reproduces the error using the [NAME_No_time_averaging.txt](https://github.com/SciTools/iris/files/2927057/NAME_No_time_averaging.txt) file.

```python
import iris
cubes = iris.load('NAME_No_time_averaging.txt')
iris.save(cubes, 'test.nc')
```

pp-mo:
> I have cube where `cube.cell_methods == (None,)`

That's not a save bug, it's a bad cube.

`cube.cell_methods` should definitely be an iterable, which should only contain `iris.coords.CellMethod`s . 
A "None" element is definitely not a viable option.

The question is, how did you come by such a thing ?

volcan01010:
> That's not a save bug, it's a bad cube.

OK. The `cube.cell_methods` tuple is iterable, but I didn't know enough about cell methods to know that None was not a valid item.

The cube was created from NAME output.  The [Name_No_time_averaging.txt](https://github.com/SciTools/iris/files/2927057/NAME_No_time_averaging.txt) file is a stripped-down example.

The `None` cell method is created by the `_build_cell_methods()` function of `name_loader.py` 

https://github.com/SciTools/iris/blob/45a18abc1b87a43a1373212ab4205713fe92277c/lib/iris/fileformats/name_loaders.py#L524

The data in file matches the `no_avg_pattern` regular expression so the cell method is set to `None`.

pp-mo:
> @volcan01010 The data in file matches the no_avg_pattern regular expression so the cell method is set to None.

Ok, my apologies, that one clearly **_is_** a bug !!

The code seems in fact to have 2 different possible ways of getting `cell_method == None`, 
so I reckon there is a simple "`if cell_method is not None:`" wanting here.
[...]

PR #4933, opened 2022-08-25
Fix name loader problem

pp-mo:
Addresses #3288 

~Draft pending :~ 
  - ~[x] tests for enhanced API~
  - ~[x] whatsnew~

### DIFF
--- a/lib/iris/cube.py
+++ b/lib/iris/cube.py
@@ -2299,10 +2299,23 @@ class Cube(CFVariableMixin):
         return self._metadata_manager.cell_methods
 
     @cell_methods.setter
-    def cell_methods(self, cell_methods):
-        self._metadata_manager.cell_methods = (
-            tuple(cell_methods) if cell_methods else tuple()
-        )
+    def cell_methods(self, cell_methods: Iterable):
+        if not cell_methods:
+            # For backwards compatibility: Empty or null value is equivalent to ().
+            cell_methods = ()
+        else:
+            # Can supply any iterable, which is converted (copied) to a tuple.
+            cell_methods = tuple(cell_methods)
+            for cell_method in cell_methods:
+                # All contents should be CellMethods.  Requiring class membership is
+                # somewhat non-Pythonic, but simple, and not a problem for now.
+                if not isinstance(cell_method, iris.coords.CellMethod):
+                    msg = (
+                        f"Cube.cell_methods assigned value includes {cell_method}, "
+                        "which is not an iris.coords.CellMethod."
+                    )
+                    raise ValueError(msg)
+        self._metadata_manager.cell_methods = cell_methods
 
     def core_data(self):
         """
--- a/lib/iris/fileformats/name_loaders.py
+++ b/lib/iris/fileformats/name_loaders.py
@@ -571,7 +571,9 @@ def _generate_cubes(
                 cube.attributes[key] = value
 
         if cell_methods is not None:
-            cube.add_cell_method(cell_methods[i])
+            cell_method = cell_methods[i]
+            if cell_method is not None:
+                cube.add_cell_method(cell_method)
 
         yield cube
 
@@ -610,7 +612,7 @@ def _build_cell_methods(av_or_ints, coord):
             cell_method = None
             msg = "Unknown {} statistic: {!r}. Unable to create cell method."
             warnings.warn(msg.format(coord, av_or_int))
-        cell_methods.append(cell_method)
+        cell_methods.append(cell_method)  # NOTE: this can be a None
     return cell_methods
 
 
--- a/lib/iris/tests/test_name.py
+++ b/lib/iris/tests/test_name.py
@@ -7,8 +7,11 @@
 
 # import iris tests first so that some things can be initialised before
 # importing anything else
-import iris.tests as tests  # isort:skip
+import tempfile
+
 import iris
+
+import iris.tests as tests  # isort:skip
 
 
 @tests.skip_data
@@ -39,13 +42,39 @@ class TestLoad(tests.IrisTest):
         )
         self.assertCMLApproxData(cubes, ("name", "NAMEIII_version2.cml"))
 
-    def test_NAMEII_trajectory(self):
+    def test_NAMEIII_trajectory(self):
         cubes = iris.load(
             tests.get_data_path(("NAME", "NAMEIII_trajectory.txt"))
         )
         self.assertCML(cubes[0], ("name", "NAMEIII_trajectory0.cml"))
         self.assertCML(
             cubes, ("name", "NAMEIII_trajectory.cml"), checksum=False
+        )
+
+    def test_NAMEII__no_time_averaging(self):
+        cubes = iris.load(
+            tests.get_data_path(("NAME", "NAMEII_no_time_averaging.txt"))
+        )
+
+        # Also check that it saves without error.
+        # This was previously failing, see https://github.com/SciTools/iris/issues/3288
+        with tempfile.TemporaryDirectory() as temp_dirpath:
+            iris.save(cubes, temp_dirpath + "/tmp.nc")
+
+        self.assertCML(
+            cubes[0],
+            (
+                "name",
+                "NAMEII_field__no_time_averaging_0.cml",
+            ),
+        )
+        self.assertCML(
+            cubes,
+            (
+                "name",
+                "NAMEII_field__no_time_averaging.cml",
+            ),
+            checksum=False,
         )
 
 
--- a/lib/iris/tests/test_name.py
+++ b/lib/iris/tests/test_name.py
@@ -7,11 +7,11 @@
 
 # import iris tests first so that some things can be initialised before
 # importing anything else
+import iris.tests as tests  # isort:skip
+
 import tempfile
 
 import iris
-
-import iris.tests as tests  # isort:skip
 
 
 @tests.skip_data
--- a/.github/workflows/ci-tests.yml
+++ b/.github/workflows/ci-tests.yml
@@ -46,7 +46,7 @@ jobs:
             session: "tests"
 
     env:
-      IRIS_TEST_DATA_VERSION: "2.15"
+      IRIS_TEST_DATA_VERSION: "2.17"
       ENV_NAME: "ci-tests"
 
     steps:
--- a/docs/src/whatsnew/3.3.rst
+++ b/docs/src/whatsnew/3.3.rst
@@ -42,8 +42,15 @@ v3.3.1 |build_date| [unreleased]
 
    The patches in this release of Iris include:
 
-#. `@pp-mo`_ fixed the Jupyter notebook display of :class:`~iris.cube.CubeList`.
-   (:issue:`4973`, :pull:`4976`)
+    #. `@pp-mo`_ fixed the Jupyter notebook display of :class:`~iris.cube.CubeList`.
+       (:issue:`4973`, :pull:`4976`)
+
+    #. `@pp-mo`_ fixed a bug in NAME loaders where data with no associated statistic would
+       load as a cube with invalid cell-methods, which cannot be printed or saved to netcdf.
+       (:issue:`3288`, :pull:`4933`)
+
+    #. `@pp-mo`_ ensured that :data:`iris.cube.Cube.cell_methods` must always be an iterable
+       of :class:`iris.coords.CellMethod` objects (:pull:`4933`).
 
 
 📢 Announcements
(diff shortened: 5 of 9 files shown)


pp-mo:
@volcan01010 would you be available to review ?

lbdreyer (commented):
This looks good to me, I had one small comment though.

It is a bit of a shame there is no extra testing for the name loader so it would be nice to get hold of some test data. Is there a way we could generate some?

lib/iris/fileformats/name_loaders.py
```diff
@@ -571,7 +571,9 @@ def _generate_cubes(
                 cube.attributes[key] = value
 
         if cell_methods is not None:
-            cube.add_cell_method(cell_methods[i])
+            cell_method = cell_methods[i]
```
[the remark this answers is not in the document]
pp-mo: I don't think this is correct.  The list `cell_methods` contains a result for each "data array" in the file, which is indexed by `i`.  That there is only one in each case is arranged [elsewhere](https://github.com/SciTools/iris/blob/ae2b59ba3dc68fe54893f4e8f4246e5c19c1a5bf/lib/iris/fileformats/name_loaders.py#L579), and I assume is guaranteed by the actual file format (about which I frankly don't know much).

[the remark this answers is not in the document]
lbdreyer: Ahh yes I see I was reading that `i` as a `1`. That makes more sense now!

pp-mo:
> This looks good to me, I had one small comment though.
> 
> It is a bit of a shame there is no extra testing for the name loader so it would be nice to get hold of some test data. Is there a way we could generate some?

As I explained [above](https://github.com/SciTools/iris/pull/4933#issuecomment-1239675286), we could ask @volcan01010 if he is prepared to release his sample file ??
None of the existing test files seem suitable. 
I did test it on a modified version of one of the existing test files, but the problem is I can only guess as to what is correct within the format.

lib/iris/cube.py
```diff
@@ -2293,10 +2293,23 @@ def cell_methods(self):
...
+            cell_methods = tuple(cell_methods)
+            for cell_method in cell_methods:
+                # All contents should be CellMethods.  Requiring class membership is
+                # somewhat non-Pythonic, but simple, and not a problem for now.
+                if not isinstance(cell_method, iris.coords.CellMethod):
+                    msg = (
+                        f"Cube.cell_methods assigned value includes {cell_method}, "
+                        "which is not an iris.coords.CellMethod."
+                    )
+                    raise ValueError(msg)
```
DPeterK, line 2311: Assumedly we could attempt to handle the incorrectly-typed object here rather than immediately erroring? That way we avoid crashing out of the load at the first sign of failure.

For example, we could instead:
* ignore the not-a-`CellMethod` and not include it in the final list of cell methods. This might mean we don't fully represent all of the averaging metadata of the incoming dataset, of course, which is undesirable - but if the object is already malformed maybe we're better off skipping it anyway?
* try passing the not-a-`CellMethod` back through the `CellMethod` constructor and re-inserting it into the list of cell methods. That way if there is a problem with the averaging metadata we can hopefully fail within the `CellMethod` constructor with a more accurate description of _why_ the object is causing the problem.

DPeterK (commented):
Just one comment from me on handling objects that aren't cell methods that show up in the list of cell methods. To be clear, I don't think this is something that needs to be fixed, but we could consider an alternative approach to erroring fast in the middle of a dataset load.

pp-mo:
> None of the existing test files seem suitable. I did test it on a modified version of one of the existing test files, but the problem is I can only guess as to what is correct within the format.

Well, on a closer look I think we actually did have a suitable testfile all along, which was previously unused.
So I used that -- can someone please confirm that it covers the case ?

Meanwhile, apologies for my moving this + some other tests into slightly a more logical hierarchy, it just seemed time to get that done.

lib/iris/cube.py
pp-mo: > objects that aren't cell methods that show up in the list of cell methods

I don't quite get this.  The change here only operates when you assign something to 'cube.cell_methods', so I can't see why we would want to attempt anything _except_ to say that we only accept CellMethod instances.

My thinking is, I don't think this would ever be triggered within other Iris operations, it would only happen in a user-code assignment.

pp-mo:
OK, I confused myself there.
There is no such pre-existing file in iris-test-data, only in my version !

I will fix this,
 and perhaps now revert all the file moves as it makes the changed-files quite unreasonably messy ...

pp-mo:
This should pass, when https://github.com/SciTools/iris-test-data/pull/82 is merged
(unless some other update is merged there first)

lib/iris/tests/test_name.py
```diff
@@ -48,6 +51,32 @@ def test_NAMEII_trajectory(self):
             cubes, ("name", "NAMEIII_trajectory.cml"), checksum=False
         )
 
+    def test_NAMEIII__no_time_averaging(self):
+        cubes = iris.load(
+            tests.get_data_path(("NAME", "NAMEIII_no_time_averaging.txt"))
+        )
```
bsherratt, line 57: Actually, the example file is NAME II format, not NAME III.

The first line of the file that says "NAME III" is in fact describing the version of NAME that *produced* the file, not the version of the output. It can do both versions for compatibility.

I think the main way to tell the difference in output formats - certainly the one iris already uses - is whether it has a "Title" (NAME II) or a "Run name" (NAME III).

bsherratt (commented):
One minor comment about the naming here, but more importantly:
- Having mentioned both NAME II and NAME III formats, they are both widely used in practice (due to a habit of copying NAME config without necessarily reviewing options like the output format), so I would suggest testing both. I can find/create some example output if needed.
- Can this PR please also be added as patch releases to both 3.2 and 3.3? The refactoring of the cube string representation code in 3.2 combines with this bug to mean we often cannot print a cube. While I believe we could do almost everything else with it (except save as originally raised in the linked issue), the lack of printing does not give anyone the confidence to try, so we are essentially stuck using older versions of iris for NAME.

trexfeathers:
>Can this PR please also be added as patch releases to both 3.2 and 3.3? The refactoring of the cube string representation code in 3.2 combines with this bug to mean we often cannot print a cube

Offline discussion of the specifics: a `v3.3.1` bugfix would serve, so that's what we will aim for.

volcan01010:
I've had a look through the PR and it looks good to me.  I haven't worked on that Iris project for a while, so unfortunately I'm not up to speed to do a more detailed review.  Please feel free to incorporate the test data file into the project.

pp-mo:
Many thanks @bsherratt 
Working slightly in the dark here, as don't actually know much about NAME, so help is welcome !

> Can this PR please also be added as patch releases to both 3.2 and 3.3? The refactoring of the cube string representation code in 3.2 combines with this bug to mean we often cannot print a cube. While I believe we could do almost everything else with it (except save as originally raised in the linked issue), the lack of printing does not give anyone the confidence to try, so we are essentially stuck using older versions of iris for NAME.

From internal discussion (peloton meeting), it seems this really is needed, so yes, we will commit to a 3.3.1 release.
(See #4976 for the beginnings)

----
> Actually, the example file is NAME II format, not NAME III.

Ok I will rename the new file + fix it here and in https://github.com/SciTools/iris-test-data/pull/82

----
> * Having mentioned both NAME II and NAME III formats, they are both widely used in practice (due to a habit of copying NAME config without necessarily reviewing options like the output format), so I would suggest testing both. I can find/create some example output if needed.

This sounds a useful improvement, but we should handle it as a separate issue ?
https://github.com/SciTools/iris/issues/4979
I will contact you offline about getting some suitable testfiles ...

pp-mo:
Yay, GTG  !
with thanks, Merges invited ? ...

lib/iris/cube.py
trexfeathers: I'm happy with this. The main thing is that we're confident the various loaders have already simplified things to either `None` or a valid `CellMethod` - nothing else is expected to get this far unless a user is explicitly using the `setter`.

And in all cases I would be uncomfortable with attempting a conversion to a valid `CellMethod` object, as this is fairly opaque behaviour that separates the user from the true situation (often referred to as "magic").

docs/src/whatsnew/latest.rst
```diff
@@ -57,6 +57,13 @@ This document explains the changes made to Iris for this release
    variables and cell measures that map to a cube dimension of length 1 are now
    included in the respective vector sections. (:pull:`4945`)
 
+#. `@pp-mo`_ fixed a bug in NAME loaders where data with no associated statistic would
+   load as a cube with invalid cell-methods, which cannot be printed or saved to netcdf.
+   (:issue:`3288`, :pull:`4933`)
+
+#. `@pp-mo`_ ensured that :data:`iris.cube.Cube.cell_methods` must always be an iterable
+   of :class:`iris.coords.CellMethod` objects (:pull:`4933`).
+
```
trexfeathers, line 66: If you are re-targeting `v3.3.x`, then this entry should be in `3.3.rst`.

trexfeathers (changes requested):
Thanks @pp-mo! Some tiny testing changes for you, also:

## Don't forget to re-target to `v3.3.x`

pp-mo:
> self.assertEqual and self.assertRaisesRegex can be converted to pytest style

Ok I have done that now + I think it looks pretty neat. https://github.com/SciTools/iris/pull/4933/commits/f6853deb1fa2695975fa049f24b3d40f92ff511e

However, It may be appropriate for @lbdreyer  to also check out + OK my use of the fixture here ...

So... I have defined a fixture as a class-instance-method, and labelled it "autouse=True", and it assigns data to "self" properties.
Which is really close to the unittest "setUp" approach -- so I just hope that this is appropriate !!

It seem to work, and I **_think_** it makes sense :  on debugging, I find that the 'self' instance of the "Test__cell_methods" class which the fixture receives is unique for each test method invocation.  So I assume that this 'self' therefore "*is* the test instance", and can be used to hold per-instance data as in the older style.
However, I'd just like a bit more confirmation that this makes sense to someone else, as I don't see this suggestion in the Pytest docs and basically got here more by trial-and-error than from the docs !!

pp-mo:
Meanwhile ...
>  Don't forget to re-target to `v3.3.x`

Good spot.
However, I think it's best to let #4976 go first, as that establishes the whatsnew section for 3.3.1.
Rather than changing this + making a conflict, let's wait for that, after which I will fix/rebase this one to build on that accordingly.

pp-mo:
Thanks @trexfeathers 
I have now **re-targetted and fixed, and I think will now be OK.**
That includes fixing [the outstanding problem with the post-3.3 whatsnew](https://github.com/SciTools/iris/pull/4976#issuecomment-1257683528) : added an indent to put it inside the v.3.3.1 dropdown.

But I've had to squash commits to sort out the changes :disappointed: 

There was also some suggestion we might want to add the netcdf version pin from https://github.com/SciTools/iris/pull/4968 onto 3.3.x, as otherwise we will probably get occasional docs build failures.
But I'm hoping we can address that separately.

trexfeathers (approved):
So GitHub is telling me you've changed pretty much everything since my review on Friday, but I can only see the changes I was expecting so I'm assuming that's just the force-pushing messing things up.

lbdreyer:
So the docs build is failing again (see https://readthedocs.org/projects/scitools-iris/builds/ for the latests builds) I restarted the job a couple of times but it's getting the same HDF error. Interestingly when @pp-mo [re ran the build ](https://github.com/SciTools/iris/pull/4976#issuecomment-1257683528) it worked the second time

trexfeathers:
> So the docs build is failing again (see https://readthedocs.org/projects/scitools-iris/builds/ for the latests builds) I restarted the job a couple of times but it's getting the same HDF error. Interestingly when @pp-mo [re ran the build ](https://github.com/SciTools/iris/pull/4976#issuecomment-1257683528) it worked the second time

>There was also some suggestion we might want to add the netcdf version pin from https://github.com/SciTools/iris/pull/4968 onto 3.3.x, as otherwise we will probably get occasional docs build failures.
>But I'm hoping we can address that separately.

pp-mo:
Belated thanks @trexfeathers 
This was one of those occasions when I couldn't get a straight rebase to work for me.
It is unavoidably tough when a rebase/squash interrupts the comment/response cycle, but I needed
Thanks for wading through !
⭐ 🥇

trexfeathers:
>This was one of those occasions when I couldn't get a straight rebase to work for me.

@pp-mo next time I'm reviewing feel free to merge in the updates instead of rebasing. It doesn't upset my reviewing workflow and would be easier on you for resolving conflicts.

PR #4945, opened 2022-09-06
Fix cube repr for scalar anc_var and cell measures

lbdreyer:
### Description
I noticed that the cube printout wasn't handling scalar ancillary variables and cell measures correctly.

* There wasn't a section for scalar ancillary variables, so I've added that now
* If a cube had a dimension of length 1, and a cell measure that maps to this dimension, it was adding it to the "Scalar cell measures". 

Example to reproduce the second bullet point above:
```
cube = Cube(np.zeros((2, 1)), long_name="name", units=1)
cube.add_cell_measure(CellMeasure([0], long_name="cm"), 1)
```
print as

```
print(cube)
name / (1)                          (-- : 2; -- : 1)
    Scalar cell measures:
        cm
```

but it should print as

```
print(cube)
name / (1)                          (-- : 2; -- : 1)
    Scalar cell measures:
        cm                                  -        x               
    
```
---

### DIFF
--- a/lib/iris/_representation/cube_printout.py
+++ b/lib/iris/_representation/cube_printout.py
@@ -260,6 +260,10 @@ class CubePrinter:
                 elif title in ("attributes:", "cell methods:", "mesh:"):
                     for title, value in zip(sect.names, sect.values):
                         add_scalar_row(title, value)
+                elif title == "scalar ancillary variables:":
+                    # These are just strings: nothing in the 'value' column.
+                    for name in sect.contents:
+                        add_scalar_row(name)
                 elif title == "scalar cell measures:":
                     # These are just strings: nothing in the 'value' column.
                     for name in sect.contents:
--- a/lib/iris/_representation/cube_summary.py
+++ b/lib/iris/_representation/cube_summary.py
@@ -219,6 +219,12 @@ class ScalarCellMeasureSection(Section):
         self.contents = [cm.name() for cm in cell_measures]
 
 
+class ScalarAncillaryVariableSection(Section):
+    def __init__(self, title, ancillary_variables):
+        self.title = title
+        self.contents = [av.name() for av in ancillary_variables]
+
+
 class AttributeSection(Section):
     def __init__(self, title, attributes):
         self.title = title
@@ -274,7 +280,7 @@ class CubeSummary:
 
     """
 
-    def __init__(self, cube, shorten=False, name_padding=35):
+    def __init__(self, cube, name_padding=35):
         self.header = FullHeader(cube, name_padding)
 
         # Cache the derived coords so we can rely on consistent
@@ -314,13 +320,23 @@ class CubeSummary:
             if id(coord) not in scalar_coord_ids
         ]
 
-        # cell measures
-        vector_cell_measures = [
-            cm for cm in cube.cell_measures() if cm.shape != (1,)
-        ]
-
         # Ancillary Variables
-        vector_ancillary_variables = [av for av in cube.ancillary_variables()]
+        vector_ancillary_variables = []
+        scalar_ancillary_variables = []
+        for av, av_dims in cube._ancillary_variables_and_dims:
+            if av_dims:
+                vector_ancillary_variables.append(av)
+            else:
+                scalar_ancillary_variables.append(av)
+
+        # Cell Measures
+        vector_cell_measures = []
+        scalar_cell_measures = []
+        for cm, cm_dims in cube._cell_measures_and_dims:
+            if cm_dims:
+                vector_cell_measures.append(cm)
+            else:
+                scalar_cell_measures.append(cm)
 
         # Sort scalar coordinates by name.
         scalar_coords.sort(key=lambda coord: coord.name())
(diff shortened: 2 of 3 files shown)


docs/src/whatsnew/latest.rst
```diff
@@ -47,6 +47,13 @@ This document explains the changes made to Iris for this release
    differs from 0 or 2. This enables the use of this method on mesh
    coordinates. (:issue:`4672`, :pull:`4870`)
 
+#. `@lbdreyer`_ fixed the cube print out such that scalar ancillary variables
+   are displayed in a dedicated section rather than being added to section for
+   vector ancillary variables. Further, ancillary variables and cell measures
+   that are scalar as they map to cube dimension of length 1 are included in the
+   respective vector sections, mathcing the handling of coordinates.
```
pp-mo, line 54:
```suggestion
   respective vector sections, matching the handling of coordinates.
```
(typo!)

docs/src/whatsnew/latest.rst
```diff
@@ -47,6 +47,13 @@ This document explains the changes made to Iris for this release
    differs from 0 or 2. This enables the use of this method on mesh
    coordinates. (:issue:`4672`, :pull:`4870`)
 
+#. `@lbdreyer`_ fixed the cube print out such that scalar ancillary variables
+   are displayed in a dedicated section rather than being added to section for
+   vector ancillary variables. Further, ancillary variables and cell measures
```
pp-mo, line 52:
```suggestion
   are displayed in a dedicated section rather than being added to the vector
   ancillary variables section. Further, ancillary variables and cell measures
```

lib/iris/_representation/cube_summary.py
```diff
@@ -364,6 +377,11 @@ def add_scalar_section(section_class, title, *args):
         add_scalar_section(
             ScalarCoordSection, "Scalar coordinates:", cube, scalar_coords
         )
+        add_scalar_section(
```
pp-mo, line 380: I note that the order of all possible sections is now this ...
```
vector
  Dimension coordinates:
  Mesh coordinates:
  Auxiliary coordinates:
  Derived coordinates:
  Cell measures:
  Ancillary variables:
scalar
  Mesh:
  Scalar coordinates:
  Scalar ancillary variables:
  Scalar cell measures:
  Cell methods:
  Attributes:
```
So, I think there _is_ good reason for the location of the new scalar-ancils section : since cell-methods and attributes are rather different in "kind" from all the other scalar components, we want them at the end, after the rest.
However, that means that we now list cell-measures **_before_** ancillaries in the vector part, but **_after_** them in the scalar part.

I think the only sensible alternative would be to swap the order of the _vector_ sections.  However, that changes existing behaviour. 
**? did you consider this + what do you think ?**

pp-mo (changes requested):
Some very small points.
Otherwise all good, I think.
The actual implementation is 
The existing choice of which things needed to be tested is possibly a bit random, but your changes do fit in ok I think.

lib/iris/_representation/cube_printout.py
```diff
@@ -260,6 +260,10 @@ def add_scalar_row(name, value=""):
                 elif title in ("attributes:", "cell methods:", "mesh:"):
                     for title, value in zip(sect.names, sect.values):
                         add_scalar_row(title, value)
+                elif title == "scalar ancillary variables:":
```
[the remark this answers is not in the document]
lbdreyer: Yes that's a good point, I have now addressed this in https://github.com/SciTools/iris/pull/4945/commits/53d15291a197fcfd4270cd388fae8a4855edc0a0

lib/iris/_representation/cube_summary.py
lbdreyer: I added `Scalar ancillary variables:` before `Scalar cell measures:` for alphabetical reasons as I couldn't really think of a better way to order them. But as you point out we already decided on the order in the vector section so I'm happy to just follow that.

lib/iris/_representation/cube_printout.py
```diff
@@ -260,8 +260,11 @@ def add_scalar_row(name, value=""):
                 elif title in ("attributes:", "cell methods:", "mesh:"):
                     for title, value in zip(sect.names, sect.values):
                         add_scalar_row(title, value)
-                elif title == "scalar cell measures:":
-                    # These are just strings: nothing in the 'value' column.
+                elif title in (
+                    "scalar ancillary variables:",
+                    "scalar cell measures:",
+                ):
+                    # These are just strings: nothing in the 'va"scalar cell measures:"lue' column.
```
pp-mo, line 267:
```suggestion
                    # These are just strings: nothing in the "scalar cell measures:"l column.
```
Something weird happened in your edit !

pp-mo (changes requested):
Tiny fix, otherwise all now good

lib/iris/_representation/cube_printout.py
lbdreyer: Not sure what happened here! I have now fixed this in the latest commit: https://github.com/SciTools/iris/pull/4945/commits/7673aa9c1d096fa59c64dab57c4d727ef170ed40

pp-mo: I see my suggestion was confused, anyway.
Sorted now :+1:

pp-mo (approved):

pp-mo:
Thanks @lbdreyer !

lbdreyer:
Thanks for the review @pp-mo !

PR #4947, opened 2022-09-07
fix aggregated_by for derived coords

stephenworsley:
### Description

Addresses the remaining comments for #3637.

Does this by having all derived coordinates now derive from their equivalent coordinates in the aggregated cube.

---


bjlittle (approved):
@stephenworsley Looks good to me.

Awesome, thanks! :+1:

PR #4949, opened 2022-09-08
purge unused setup.py clean_source custom command

bjlittle:
### Description
This PR performs more `setup.py` hygiene by removing the archaic and redundant `clean_source` custom command.

IMHO it's simply not necessary, and to my knowledge has never been used as part of our workflow, ever. Even as a developer convenience I've simply never used it. 

Sounds harsh, but AFAIK this is classic deadwood infrastructure code that can be purged.

---

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -48,20 +48,6 @@ class BaseCommand(Command):
         pass
 
 
-class CleanSource(BaseCommand):
-    description = "clean orphaned pyc/pyo files from the source directory"
-
-    def run(self):
-        for root_path, dir_names, file_names in os.walk("lib"):
-            for file_name in file_names:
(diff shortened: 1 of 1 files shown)


lbdreyer (approved):
I suppose this command was added for when we were creating sdists/wheels to ensure we don't accidentally add something we shouldn't.

Now that we've automated this process and that process doesn't use this command, I think it's safe to just remove.

So this seems sensible to me!

bjlittle:
@lbdreyer I think that it may have come about when we were building `pyke` as part of `iris` and needed an automated way to "tidy-up".

But yeah, now that we've automated the process and building in clean, isolated environments, I really don't think this is now necessary.

Awesome, thanks :partying_face:

PR #4950, opened 2022-09-08
purge unnecessary setup.py copy_copyright function

bjlittle:
### Description
This PR performs yet more `setup.py` hygiene by removing the `copy_copyright` function, which copies the `COPYING` and `COPYING.LESSER` license files down into the `iris` package source root when building the `iris` distribution.

IMHO this is unnecessary, and I don't know of any other open source packages that do this.

The license files are already packaged in our PyPI `sdist` and binary `wheels` metadata, and also as part of our conda-forge `iris` conda package tarball, as per all other open source packages which follow recommended package distribution good practice.

Additionally bundling the license files within the `iris` package root source seems unnecessary.

Reference: This change was introduced in #2891 (with no explanation/reason) and first became part of `iris` 2.0.0. To be fair, I reviewed and merged that PR, and didn't challenge it's introduction... now I am :wink: 

---

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -1,5 +1,4 @@
 import os
-from shutil import copyfile
 import sys
 
 from setuptools import Command, setup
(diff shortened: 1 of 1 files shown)


lbdreyer (approved):
Thanks @bjlittle! This seems sensible to me. As you say we already include the licenses files when we upload to PyPI and conda so this command is probably no longer necessary.

I see that there is a conflict. Do you mind resolving this and then I will merge it in

lbdreyer:
FYI I haven't asked for a whats new in this PR or in #4949 as I thought, rather than adding a whats new for each of these PRs, it may be easier to add it to the what's new at the end.

bjlittle:
@lbdreyer Brilliant, thanks :partying_face:

PR #4955, opened 2022-09-09
Speed up Cube.subset/Coord.intersect

bouweandela:
### Description

This speeds up `Cube.subset` / `Coord.intersect` by using a `dict` lookup instead of `list.index` to find the indices of the coordinate that should be used for subsetting the cube.

Here is some benchmark code:

```python
import iris.cube
import iris.coords
import numpy as np

lon = iris.coords.DimCoord(np.arange(0, 360, 0.5, dtype=np.float64), standard_name='longitude')
cube = iris.cube.Cube(np.arange(720, dtype=np.float32))
cube.add_dim_coord(lon, 0)
sublon = lon[(lon.points > 60) & (lon.points < 300)]

%timeit cube.subset(sublon)
```

The current `main` branch result is:
```
93.9 ms ± 4.51 ms per loop (mean ± std. dev. of 7 runs, 10 loops each)
```
and the result of running that code with this branch is:
```
13.5 ms ± 1.54 ms per loop (mean ± std. dev. of 7 runs, 100 loops each)
```

---

### DIFF
--- a/lib/iris/coords.py
+++ b/lib/iris/coords.py
@@ -1345,7 +1345,14 @@ class Cell(namedtuple("Cell", ["point", "bound"])):
         return Cell(point, bound)
 
     def __hash__(self):
-        return super().__hash__()
+        # See __eq__ for the definition of when two cells are equal.
+        if self.bound is None:
+            return hash(self.point)
+        bound = self.bound
+        rbound = bound[::-1]
+        if rbound < bound:
+            bound = rbound
+        return hash((self.point, bound))
 
     def __eq__(self, other):
         """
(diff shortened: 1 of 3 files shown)


lib/iris/coords.py
```diff
@@ -1345,7 +1345,9 @@ def __add__(self, mod):
         return Cell(point, bound)
 
     def __hash__(self):
-        return super().__hash__()
+        if self.bound is None:
+            return hash(self.point)
+        return hash((self.point, *sorted(self.bound)))
```
bouweandela, line 1350: I'm not too sure about the `sorted` here: I tried to mimic what the equals operator does, but sorting may only work if bounds is just two points.

bouweandela: I got rid of the sorted, I think it should work now.

lbdreyer (changes requested):
This is looking great, thanks @bouweandela ! Do you mind adding a [whatsnew](https://github.com/SciTools/iris/blob/main/docs/src/whatsnew/latest.rst) entry to cover this change?

bouweandela:
Thanks for the review @rcomer! I added the entry.

lbdreyer (approved):
Perfect! Thanks @bouweandela !

PR #4969, opened 2022-09-16
Speed up operations that use the `Coord.cells` method for time coordinates

### LINKED ISSUE #4957: Extracting a time range from a cube is slow
## 📰 Custom Issue

Extracting a time range as [described in the documentation](https://scitools-iris.readthedocs.io/en/latest/userguide/subsetting_a_cube.html#constraining-on-time) is quite slow if you want to do it for many cubes and/or cubes with many time points. For a single cube with 10000 time points it already takes 2 seconds on my computer, so if I want to subset a few hundred cubes that becomes quite slow.

Here is a script that demonstrates this:

```python
import cf_units
import iris.cube
import iris.coords
import iris.time
import numpy as np

time_units = cf_units.Unit('days since 1850-01-01', calendar='standard')
time = iris.coords.DimCoord(np.arange(10000, dtype=np.float64), standard_name='time', units=time_units)
cube = iris.cube.Cube(np.arange(10000, dtype=np.float32))
cube.add_dim_coord(time, 0)
pdt1 = iris.time.PartialDateTime(year=1852)
pdt2 = iris.time.PartialDateTime(year=1854)
constraint = iris.Constraint(time=lambda cell: pdt1 <= cell.point < pdt2)

%timeit cube.extract(constraint)
```
Result:
```
1.83 s ± 28 ms per loop (mean ± std. dev. of 7 runs, 1 loop each)
```

From looking at the code in `iris.coords`, it looks like the slow behaviour is caused by converting all time points to datetimes individually for each cell, instead of converting them once and then generating the cells.

Here is some code with timings:
```python
%timeit time.units.num2date(time.points)
```
```
27.3 ms ± 3.18 ms per loop (mean ± std. dev. of 7 runs, 10 loops each)
```
and
```
%timeit list(time.units.num2date(p) for p in time.points)
```
```
1.53 s ± 29.5 ms per loop (mean ± std. dev. of 7 runs, 1 loop each)
```

If this is an interesting feature, I can make a pull request to change the code so it first converts all the time points and then generates the cells?

bouweandela:
### Description

This speeds up functions that make use of the `Coord.cells` method to generate many cells describing a time coordinate. This affects for example `Cube.extract`, `Cube.subset`, and `Coord.intersection`.

Here is a script that demonstrates this:

```python
import cf_units
import iris.cube
import iris.coords
import iris.time
import numpy as np

time_units = cf_units.Unit('days since 1850-01-01', calendar='standard')
time = iris.coords.DimCoord(np.arange(10000, dtype=np.float64), standard_name='time', units=time_units)
cube = iris.cube.Cube(np.arange(10000, dtype=np.float32))
cube.add_dim_coord(time, 0)
pdt1 = iris.time.PartialDateTime(year=1852)
pdt2 = iris.time.PartialDateTime(year=1854)
constraint = iris.Constraint(time=lambda cell: pdt1 <= cell.point < pdt2)

%timeit cube.extract(constraint)
```

Before:
```
1.4 s ± 15 ms per loop (mean ± std. dev. of 7 runs, 1 loop each)
```
After:
```
36.7 ms ± 1.23 ms per loop (mean ± std. dev. of 7 runs, 10 loops each)
```

Note that the changes in this pull request do slow down the case where only a few cells are actually generated:

Before:
```python
%timeit time.cells()
```
```
1.24 µs ± 20.6 ns per loop (mean ± std. dev. of 7 runs, 1,000,000 loops each)
```
```python
%timeit next(time.cells())
```
```
140 µs ± 1.07 µs per loop (mean ± std. dev. of 7 runs, 10,000 loops each)
```
After:
```python
%timeit time.cells()
```
```
216 ns ± 7.68 ns per loop (mean ± std. dev. of 7 runs, 1,000,000 loops each)
```
```python
%timeit next(time.cells())
```
```
16.8 ms ± 395 µs per loop (mean ± std. dev. of 7 runs, 100 loops each)
```
but you can still use `Coord.cell` if you need just a few cells.

Closes #4957.

---

### DIFF
--- a/docs/src/whatsnew/latest.rst
+++ b/docs/src/whatsnew/latest.rst
@@ -71,7 +71,7 @@ This document explains the changes made to Iris for this release
    :obj:`~iris.analysis.SUM`, :obj:`~iris.analysis.COUNT` and
    :obj:`~iris.analysis.PROPORTION` on real data. (:pull:`4905`)
 
-#. `@bouweandela` made :meth:`iris.coords.Coord.cells` faster for time
+#. `@bouweandela`_ made :meth:`iris.coords.Coord.cells` faster for time
    coordinates. This also affects :meth:`iris.cube.Cube.extract`,
    :meth:`iris.cube.Cube.subset`, and :meth:`iris.coords.Coord.intersect`.
    (:pull:`4969`)
--- a/docs/src/whatsnew/latest.rst
+++ b/docs/src/whatsnew/latest.rst
@@ -50,7 +50,7 @@ This document explains the changes made to Iris for this release
 #. `@bjlittle`_ and `@lbdreyer`_ (reviewer) fixed the building of the CF
    Standard Names module ``iris.std_names`` for the ``setup.py`` commands
    ``develop`` and ``std_names``. (:issue:`4951`, :pull:`4952`)
-   
+
 #. `@lbdreyer`_ and `@pp-mo`_ (reviewer) fixed the cube print out such that
    scalar ancillary variables are displayed in a dedicated section rather than
    being added to the vector ancillary variables section. Further, ancillary
@@ -71,6 +71,10 @@ This document explains the changes made to Iris for this release
    :obj:`~iris.analysis.SUM`, :obj:`~iris.analysis.COUNT` and
    :obj:`~iris.analysis.PROPORTION` on real data. (:pull:`4905`)
 
+#. `@bouweandela` made :meth:`iris.coords.Coord.cells` faster for time
+   coordinates. This also affects :meth:`iris.cube.Cube.extract`,
+   :meth:`iris.cube.Cube.subset`, and :meth:`iris.coords.Coord.intersect`.
+   (:pull:`4969`)
 
 🔥 Deprecations
 ===============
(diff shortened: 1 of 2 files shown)


bouweandela:
I ran a few more experiments, and this implementation is faster if you generate more than roughly 100 cells.

bouweandela:
Thanks! I added a what's new entry.

docs/src/whatsnew/latest.rst
```diff
@@ -71,6 +71,10 @@ This document explains the changes made to Iris for this release
    :obj:`~iris.analysis.SUM`, :obj:`~iris.analysis.COUNT` and
    :obj:`~iris.analysis.PROPORTION` on real data. (:pull:`4905`)
 
+#. `@bouweandela` made :meth:`iris.coords.Coord.cells` faster for time
```
trexfeathers, line 74:
```suggestion
#. `@bouweandela`_ made :meth:`iris.coords.Coord.cells` faster for time
```

bouweandela: Thanks!

trexfeathers (approved):

PR #4970, opened 2022-09-16
Simpler/faster data aggregation code in `aggregated_by`

rcomer:
### Description

The code currently used for the lazy case in `aggregated_by` just needs a few tweaks to also use it for the real case.  So we can have fewer lines, and the McCabe complexity of the method is reduced from 27 to 20.

Edit: it turns out this also makes it faster (https://github.com/SciTools/iris/pull/4970#issuecomment-1263663273).

---

### DIFF
--- a/lib/iris/cube.py
+++ b/lib/iris/cube.py
@@ -4240,6 +4240,11 @@ x            -              -
             aggregateby_weights = None
 
         aggregateby_data = stack(result, axis=dimension_to_groupby)
+        # Ensure plain ndarray is output if plain ndarray was input.
+        if ma.isMaskedArray(aggregateby_data) and not ma.isMaskedArray(
+            input_data
+        ):
+            aggregateby_data = ma.getdata(aggregateby_data)
 
         # Add the aggregation meta data to the aggregate-by cube.
         aggregator.update_metadata(
--- a/lib/iris/cube.py
+++ b/lib/iris/cube.py
@@ -4178,7 +4178,7 @@ x            -              -
         data_shape = list(self.shape + aggregator.aggregate_shape(**kwargs))
         data_shape[dimension_to_groupby] = len(groupby)
 
-        # Choose appropriate data and functions.
+        # Choose appropriate data and functions for data aggregation.
         if aggregator.lazy_func is not None and self.has_lazy_data():
             stack = da.stack
             input_data = self.lazy_data()
--- a/lib/iris/cube.py
+++ b/lib/iris/cube.py
@@ -4178,98 +4178,68 @@ x            -              -
         data_shape = list(self.shape + aggregator.aggregate_shape(**kwargs))
         data_shape[dimension_to_groupby] = len(groupby)
 
-        # Aggregate the group-by data.
+        # Choose appropriate data and functions.
         if aggregator.lazy_func is not None and self.has_lazy_data():
-            front_slice = (slice(None, None),) * dimension_to_groupby
-            back_slice = (slice(None, None),) * (
-                len(data_shape) - dimension_to_groupby - 1
-            )
-
-            # Create cube and weights slices
-            groupby_subcubes = map(
-                lambda groupby_slice: self[
-                    front_slice + (groupby_slice,) + back_slice
-                ].lazy_data(),
-                groupby.group(),
-            )
-            if weights is not None:
-                groupby_subweights = map(
-                    lambda groupby_slice: weights[
-                        front_slice + (groupby_slice,) + back_slice
-                    ],
-                    groupby.group(),
-                )
-            else:
-                groupby_subweights = (None for _ in range(len(groupby)))
-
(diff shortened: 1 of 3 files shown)


lib/iris/tests/unit/cube/test_Cube__aggregated_by.py
```diff
@@ -66,9 +66,7 @@ def setUp(self):
 
         self.mock_agg = mock.Mock(spec=Aggregator)
         self.mock_agg.cell_method = []
-        self.mock_agg.aggregate = mock.Mock(
-            return_value=mock.Mock(dtype="object")
-        )
+        self.mock_agg.aggregate = mock.Mock(return_value=np.arange(4))
```
rcomer, line 69: Now that we use `stack` instead of assigning to array indices, the mocked aggregate method needs to return an array of the expected shape.

lib/iris/cube.py
```diff
@@ -4075,98 +4075,65 @@ def aggregated_by(
...
+        front_slice = (slice(None),) * dimension_to_groupby
+        back_slice = (slice(None),) * (
+            len(data_shape) - dimension_to_groupby - 1
+        )
 
-            # Create cube and weights slices
-            groupby_subcubes = map(
-                lambda groupby_slice: self[
+        groupby_subarrs = map(
+            lambda groupby_slice: iris.util._slice_data_with_keys(
```
rcomer, line 4096: Note that `_slice_data_with_keys` is used by [`Cube.__getitem__`](https://github.com/SciTools/iris/blob/de911ce55da50172b51c662c44a7d1adff148dca/lib/iris/cube.py#L2541), so this is just a shortcut to what was already happening.  Should be a little more efficient.

rcomer:
Ran the "cube" benchmarks locally and got

```
-      1.78±0.06s       1.18±0.02s     0.66  cube.Aggregation.time_aggregated_by
```

lib/iris/cube.py
rcomer: > Should be a little more efficient

Slicing just the array rather than the whole cube seems to account for the entire speedup noted at https://github.com/SciTools/iris/pull/4970#issuecomment-1263663273

lib/iris/cube.py
```diff
@@ -4178,98 +4178,65 @@ def aggregated_by(
...
-                if weights_result is not None:
-                    aggregateby_weights[tuple(cube_slice)] = weights_result
-
-            # Restore original weights.
-            if weights is not None:
-                kwargs["weights"] = weights
+            aggregateby_weights = None
+
+        aggregateby_data = stack(result, axis=dimension_to_groupby)
+        # Ensure plain ndarray is output if plain ndarray was input.
```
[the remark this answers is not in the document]
trexfeathers: ... I can see now that this change went in to get the CML tests passing. So it's important to me to know whether the CML tests are asserting 'correct' behaviour, or whether they're in the way of you improving aggregation behaviour.

When would an aggregator receive a vanilla NumPy array, but return a masked array? Is that something we want to suppress, or retain?

trexfeathers (commented):
Thanks @rcomer! Well done for making the code less scary. I've put a question in the comments, and will see if I can replicate your performance findings tomorrow.

lib/iris/cube.py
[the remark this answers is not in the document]
rcomer: A lot of the aggregators use the `np.ma` version of a function regardless of whether input cube has masked data.  E.g. `MEAN`

https://github.com/SciTools/iris/blob/b03509480c5cae7018990ef5fe2955fd32b5777c/lib/iris/analysis/__init__.py#L1978-L1980

The equivalent lines in `main` to ensure we get a plain numpy array out are

https://github.com/SciTools/iris/blob/b03509480c5cae7018990ef5fe2955fd32b5777c/lib/iris/cube.py#L4251-L4258

so it does seem to have been deliberate, however there does not seem to be any equivalent in `collapsed`:

```python
import iris.cube
import numpy as np


cube = iris.cube.Cube(np.arange(6).reshape(2, 3))
for i, coord_name in enumerate(["foo", "bar"]):
    cube.add_dim_coord(iris.coords.DimCoord(range(i+2), long_name=coord_name), i)
    
res = cube.collapsed("foo", iris.analysis.MEAN)
print(repr(res.data))
```
```
masked_array(data=[1.5, 2.5, 3.5],
             mask=False,
       fill_value=1e+20)
```

So this is inconsistent and I agree it's odd and we can't replicate it for lazy data.  You are right I only did it to get the tests passing.  So I would not be averse to reverting this and changing the tests.  I guess that would technically be a breaking change though?

trexfeathers:
> Ran the "cube" benchmarks locally and got
> 
> ```
> -      1.78±0.06s       1.18±0.02s     0.66  cube.Aggregation.time_aggregated_by
> ```

I have been able to replicate this 👍

Issue #3480, opened 2019-10-22
Proposed fixes to _discontiguity_in_bounds() 


bjlittle:
@stephenworsley Reference #3404

Let's have the discussion here, and see if there is any momentum.

@pp-mo and @alastair-gemmell

cpelley:
See #3345

bjlittle:
Peloton discussion - this passes for meshes that are contiguous, but fails to recognise some meshes that aren't contiguous.

bjlittle:
@stephenworsley Is this important for `iris-esmf-regrid` ?

stephenworsley:
> Peloton discussion - this passes for meshes that are contiguous, but fails to recognise some meshes that aren't contiguous.

I've had another look at my comments for this and I believe there is also another bug which would cause discontiguous coords to test as contiguous due to these lines:
https://github.com/SciTools/iris/blob/ae2b59ba3dc68fe54893f4e8f4246e5c19c1a5bf/lib/iris/coords.py#L1978-L1984
which would fail when `cell_size` was sufficiently large.

PR #3404, opened 2019-09-18
Fixes to _discontiguity_in_bounds()

stephenworsley:
Adresses concerns in #3401.

Issue #3401, opened 2019-09-16
2D coord bounds handling


stephenworsley:
~~Currently, a 2D coordinate point is expected to have 4 bounds. There ought to be 8. Assuming the 2D variable is defined over lattitude and longitude, there should be 4 latbounds and 4 lonbounds. http://cfconventions.org/Data/cf-conventions/cf-conventions-1.7/cf-conventions.html#cell-boundaries~~
The current way these bounds are handled is incomplete. [_discontiguity_in_bounds](https://github.com/SciTools/iris/blob/0831ba234db7035b6e28578fca76d79213d1ce5c/lib/iris/coords.py#L1022) should be making more comparisons. It is currently comparing `bounds[*,*,0]` with `bounds[*,*,1]` and  `bounds[*,*,0]` with  `bounds[*,*,3]`. it should also be comparing  `bounds[*,*,1]` with  `bounds[*,*,2]` and  `bounds[*,*,2]` with  `bounds[*,*,3]`. These comparisons should be done for both latbounds and lonbounds.

~~It's also worth noting that it's possible to end up with a coordinate which has bounds in one dimension but not another. This is in fact the case for the gallery example https://scitools.org.uk/iris/docs/v1.9.2/examples/General/cross_section.html . It's unclear how such cases should be treated (they are currently unable to be handled by pcolor/pcolormesh. That is perhaps the correct approach, but in order for something like that to be possible, it seems like there would need to be some more thought put into the structure of bounds handling).~~

stephenworsley:
My intuition is to have the shape of 2D bounds be `(*,*,2,4)`. I've taken a quick look at where in the code may have to change and I came up with the following list of functions:
coords._get_2d_coord_bound_grid
coords.Cell.contains_point
coords.Coord._bounds_setter
coords.Coord._sanity_check_bounds
coords.Coord._discontiguity_in_bounds
coords.Coord.contiguous_bounds
coords.Coord.nbounds
coords.Coord._guess_bounds
coords.DimCoord._new_bounds_requirements
analysis._grid_angles

And potentially a bunch of places with load/save rules.

stephenworsley:
> 
> 
> My intuition is to have the shape of 2D bounds be `(*,*,2,4)`.

Having re-read the CF Conventions, I've reconsidered my position on this. It seems like CF Conventions is consistent with what we are doing currently. With that said, it still seems like we are not comparing enough points for discontiguity.
[...]

Issue #4473, opened 2021-12-29
Unusual comparison in coord contiguity check


jonseddon:
## 📰 Custom Issue
In iris.coords.Coord._discontiguity_in_bounds() there's an unusual comparison that I'm struggling to understand the logic of. Please have a look at https://github.com/SciTools/iris/blob/main/lib/iris/coords.py#L1686

`diffs_along_axis` will be an array of bools, with a value of True when the difference between cells is greater than the tolerance.

`points_close_enough` is then comparing this array of bools against `atol + rtol * cell_size`, which is an array of floats, and doesn't seem to be a comparison that you would want to do.

I think that the logic's currently returning the correct result because the calculation of `points_close_enough` effectively inverts each bool in `diffs_along_axis`. `np.all()` only returns True (i.e. the coord is contiguous) when all of it's elements are True (so is a logical-and for inputs of bools). If any `diffs_along_axis` elements were greater than the tolerance then you will therefore get a False returned by `np.all()`.

Although this does work I was wondering if this may not have been the original intention of this bit of code. If this was the intention then it's not immediately clear to the reader how it actually works and could perhaps do with being simplified. This code was introduced by #3144, merged by @pp-mo.

I'm not totally sure of the intentions of calculating `points_close_enough` currently and so I don't feel able to create a PR to fix this in case I'm missing anything. A simplification could be to go from:

```
                diffs_along_axis = diffs_between_cells > (
                    atol + rtol * cell_size
                )

                points_close_enough = diffs_along_axis <= (
                    atol + rtol * cell_size
                )
                contiguous_along_axis = np.all(points_close_enough)
                return diffs_along_axis, contiguous_along_axis
```

to:

```
                diff_too_big = diffs_between_cells > (
                    atol + rtol * cell_size
                )

                return diffs_along_axis, not diff_too_big
```

but I'm not sure if this would be chopping out some functionality that `points_close_enough` was intended to add. If you are happy with my simplification then please let me know and I can raise a PR. An alternative would be to revert to the logic that existed before #3144. Otherwise I'd be grateful if you could suggest an alternative implementation for this bit of a head scratcher of a piece of code.

stephenworsley:
I believe I attempted to address this, along with some other issues, in #3404. It seems like this became stale due to a lack of momentum, though it might be worth reopening now.

stephenworsley:
[...]

PR #4975, opened 2022-09-21
Fixes to _discontiguity_in_bounds (attempt 2)

stephenworsley:
### Description

Addresses #3401 and #3480, replaces #3404 which I couldn't reopen since it's targeting a deleted branch. 

This adresses three problems at once:
- 2D discontiguity checking wasn't comparing enough bounds.
- The logic for determining discontiguity was somewhat garbled, I couldn't figure out exactly what it was trying to achieve so I replaced it with `np.isclose` like the `np.allclose` in 1D checking.
- It was unclear how the returned diffs ought to be represented as this used to be a boolean array for 2D coords and floats for 1D coords. Since this function is private, this is effectively a case of checking how these values are used. I believe that in each case they are used, they are being treated as an array of booleans so I have changed the behaviour of 1D coord checking and the description of the function to match this.

---


rcomer:
I haven't looked closely, but I think this would also address #4473?

lib/iris/coords.py
```diff
@@ -1937,11 +1937,12 @@ def _discontiguity_in_bounds(self, rtol=1e-5, atol=1e-8):
...
             True if there are no discontiguities.
         * diffs: (array or tuple of arrays)
-            The diffs along the bounds of the coordinate. If self is a 2D
-            coord of shape (Y, X), a tuple of arrays is returned, where the
-            first is an array of differences along the x-axis, of the shape
-            (Y, X-1) and the second is an array of differences along the
-            y-axis, of the shape (Y-1, X).
+            A boolean array or tuple of boolean arrays which are true where
+            there are discontiguities between neighbouring bounds. If self is
+            a 2D coord of shape (Y, X), a tuple of arrays is returned, where
```
pp-mo, line 1942:
```suggestion
            a 2D coord of shape (Y, X), a pair of arrays is returned, where
```

Not your fault, as inherited, but I think the use of "tuple" here is unhelpful.

pp-mo (changes requested):
After taking simply ages to understand what has changed here and why, I think I finally grasped it !
So, `iris.util.find_discontiguity` now finds an **extra** cell to be invalid in each testcase from `iris.tests.unit.util.test_find_discontiguities`, and the `make_bounds_discontiguous_at_point` test routine now encodes this too : although it makes the **_same_** coord adjustments as before, the resulting points which will be judged 'discontiguous' as a result have changed.

I have confirmed that the changes to the data mask in `make_bounds_discontiguous_at_point` correspond to those locations that now fail according to the utility routine `find_discontiguities` : The`find_discontiguities` code has **_not_** changed, but it relies on `Coord._discontiguity_in_bounds`, which of course has.

Most importantly, I have _also_ confirmed that the new results from `iris.util.find_discontiguities` continue to perform as expected : The points which it marks are exactly those which are required to be masked to make iris.plot.pcolormesh succeed for the cubes with bounds discontiguitites generated by the `iris.tests.unit.util.test_find_discontiguities` tests.

So in effect, the results from `iris.util.find_discontiguities` have changed.  But (a) the test code has been changed to match, and (b) the result still matches the requirement for "making mesh plots work".

I hope you agree with this analysis -- it has taken me a long time to convince myself that this is all correct + OK !

I think we can get away with this change, even though it breaks previous behaviour.
What is **still missing is a whatsnew** -- it should make clear that the results of "Coord.is_contiguous" and "util.find_discontiguity" have changed, but that we consider this to be a bugfix rather than a "breaking change".

I also have a slight qualm that the existing `make_bounds_discontiguous_at_point` routine **_always_** modifies two corners at once.  
I think that may have made sense before, when we only compared bounds 0+1' in X and 3+0' in Y, but now it would make more sense to have independent testcases that check the "new comparisons" -- i.e. 2+3' in X and 2+1' in Y.
So, I suggest you should enhance `iris.tests.stock._stock_2d_latlons.make_bounds_discontiguous_at_point` to do these, and **add suitable testcases** to the existing `iris.tests.unit.util.test_find_discontiguities.Test.test_find_discontiguities` and `iris.tests.unit.util.test_find_discontiguities.Test.test_find_discontiguities_in_y`.

lib/iris/tests/stock/_stock_2d_latlons.py
```diff
@@ -357,5 +357,9 @@ def adjust_coord(coord):
 
         masked_data = ma.masked_array(cube.data)
```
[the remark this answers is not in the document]
stephenworsley: addressed in https://github.com/SciTools/iris/pull/4975/commits/6785b40d03016ec42d2539b42c0b3b9ba545829c

lib/iris/coords.py
stephenworsley: addressed in https://github.com/SciTools/iris/pull/4975/commits/6785b40d03016ec42d2539b42c0b3b9ba545829c

lib/iris/util.py
```diff
@@ -1706,7 +1706,8 @@ def _meshgrid(*xi, **kwargs):
 def find_discontiguities(cube, rel_tol=1e-5, abs_tol=1e-8):
     """
     Searches coord for discontiguities in the bounds array, returned as a
-    boolean array (True where discontiguities are present).
+    boolean array (True for all cells which are discontiguous with the cell
+    immediately above them or to their right).
```
pp-mo, line 1710: I believe this is a more accurate description of what it **_used_** to do, but is no longer actually correct, since behaviour has changed  ??
I.E. result boolean array will now be True for each cell which is discontiguous with neighbours in any one of 8 different ways (4 directions in each of  2 coordinates).

lib/iris/util.py
```diff
@@ -1706,7 +1706,8 @@ def _meshgrid(*xi, **kwargs):
 def find_discontiguities(cube, rel_tol=1e-5, abs_tol=1e-8):
     """
     Searches coord for discontiguities in the bounds array, returned as a
-    boolean array (True where discontiguities are present).
+    boolean array (True for all cells which are discontiguous with the cell
```
pp-mo, line 1709: (sorry, not your fault but ...)
`Searches coord` here is actually wrong
It actually searches the X and Y coords of the passed cube.

We should also be clearly stating that the cube **_must_** have 2d X and Y coordinates.

pp-mo (changes requested):
i think there are still a few more minor changes to make.
the only really "wrong" thing is the docstring, but I think it does need fixing.

lib/iris/util.py
stephenworsley: I believe it's still exhibiting the same behaviour except that some discontiguities were not showing up previously. Each *discontiguity* is picked up but it is only associated with the *cell* that is immediately to its left or bellow it. The cell to its right will not be flagged as discontiguous (if there are no other discontiguities). This is a bit arbitrary, but the function has the property that all neighbouring pairs of cells which aren't flagged as discontiguous will have no discontiguities betwen them. The behaviour you are suggesting would have the additional property that all cells not flagged as discontiguous will have *no* discontiguities with *any* neighbouring cells. This may be a useful property which we may want to implement in future, but for the time being I've opted for keeping close to the original design.

lib/iris/tests/unit/util/test_find_discontiguities.py
```diff
@@ -29,12 +29,25 @@ def setUp(self):
         # Set up a 2d lat-lon cube with 2d coordinates that have been
         # transformed so they are not in a regular lat-lon grid.
         # Then generate a discontiguity at a single lat-lon point.
+        # Discontiguities will be caused at the rightmost bounds.
         self.testcube_discontig = full2d_global()
```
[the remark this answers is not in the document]
stephenworsley: addressed by https://github.com/SciTools/iris/pull/4975/commits/d85ed08c07fe8dd8dce90e610d563f7078920c9d

lib/iris/tests/unit/util/test_find_discontiguities.py
```diff
@@ -29,12 +29,25 @@ def setUp(self):
...
+            self.testcube_discontig_top, 2, 4, in_y=True
+        )
+
+        # Repeat for a discontiguity on the botommost bounds.
+        self.testcube_discontig_along_bottom = full2d_global()
+        make_bounds_discontiguous_at_point(
+            self.testcube_discontig_along_bottom, 2, 4, in_y=True, upper=False
         )
 
     def test_find_discontiguities(self):
```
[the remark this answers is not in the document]
stephenworsley: addressed by https://github.com/SciTools/iris/pull/4975/commits/d85ed08c07fe8dd8dce90e610d563f7078920c9d

lib/iris/util.py
stephenworsley: addressed by https://github.com/SciTools/iris/pull/4975/commits/d85ed08c07fe8dd8dce90e610d563f7078920c9d

lib/iris/util.py
pp-mo: But I still think that the part 
> `(True for all cells which are discontiguous with the cell immediately above them or to their right)`

is actually incorrect, if you read it as implying that it is _not_ True for a cell which is discontiguous with those _below or to the left_.  But in fact, such cells are now _also_ marked as 'True' -- i.e. it should now say something like 
> `(True for all cells which are discontiguous with any of those immediately above, below, to left or right)`

? I am misunderstanding this ?

stephenworsley: I think maybe you are misunderstanding this. If you look at the tests, we add a discontiguity to, for example, the two top most bounds of a cell. This makes it discontiguous with the cells to its left and right but if you look at the expected and returned value, there are only two cells flagged in that row, not three.

lib/iris/tests/stock/_stock_2d_latlons.py
```diff
@@ -355,7 +363,16 @@ def adjust_coord(coord):
             msg = "The coordinate {!r} doesn't span a data dimension."
             raise ValueError(msg.format(coord.name()))
 
-        masked_data = ma.masked_array(cube.data)
-        masked_data[at_iy, at_ix] = ma.masked
+    masked_data = ma.masked_array(cube.data)
+
+    # Mask all points which would be found discontiguous.
+    # Note that find_discontiguities finds all instances where a cell is
+    # discontiguous with a neighbouring cell to its *right* or *above*
```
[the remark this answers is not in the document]
stephenworsley: I believe it is still true. as in the comment above, if this were triggering for instances where there is a discontiguity to the *left* we would be seeing instances in the following code where there are three triggered cells in a row. We instead only see two triggered cells in a manner which is consistent with this comment.

lib/iris/util.py
pp-mo: Ok !
@stephenworsley and I have had a long offline discussion + investigation of the behaviour (old + new) of this.
I'm now comfortable that it is doing something more correct, relative to the existing behaviour, and is now better explained.
I'm not so confident _why_ this is the right thing to do, but IIRC the whole thing was invented to address some imagined possible problems with plotting data with 2D bounds via pcolormesh.  What it now does is slightly more "cautious", masking some extra problem points in some cases.  Unfortunately I think we never had a definitive rationale or even test examples to define exactly what is needed here.

pp-mo (approved):

PR #4998, opened 2022-09-28
Adaptions for Matplotlib 3.6

trexfeathers:
### Description

These changes address a Matplotlib deprecation warning, and should get the overnight benchmarks working again.

---

### DIFF
--- a/benchmarks/benchmarks/import_iris.py
+++ b/benchmarks/benchmarks/import_iris.py
@@ -5,17 +5,25 @@
 # licensing details.
 from importlib import import_module, reload
 
-# For use in reset_colormaps.
-from matplotlib import colormaps
+################
+# Prepare info for reset_colormaps:
 
-colormaps_orig = set(colormaps)
+# Import and capture colormaps.
+from matplotlib import colormaps  # isort:skip
+
+_COLORMAPS_ORIG = set(colormaps)
+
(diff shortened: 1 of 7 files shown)


pp-mo (approved):
This looks cool, though I did have a bit of a wobble about requiring Matplotlib >= 3.5.
That is now ~10months old. But I guess this is anyway not an actual Iris dependency, its only for CI, 
so I think that is fine.

pp-mo:
Nice work @trexfeathers .  We'll see how it goes tomorrow morning !

trexfeathers:
> This looks cool, though I did have a bit of a wobble about requiring Matplotlib >= 3.5. That is now ~10months old. But I guess this is anyway not an actual Iris dependency, its only for CI, so I think that is fine.

This is genuine. When this is released (December), Matplotlib 3.4 will not be usable with Iris

pp-mo:
Ok, hadn't really thought of it that way.
But by then, mpl 3.5 will be  > 1 year old.

trexfeathers:
This raises interesting questions about proactively responding to deprecation warnings. We typically don't spot them until the deprecation fully happens, and therefore don't experience the problem of very 'young' pins. In this case I spotted the warning while investigating and chose to action it, introducing this different problem!

PR #5001, opened 2022-09-29
Fix benchmarks `netcdf` import

trexfeathers:
### Description

* Fixes a 'missed spot' from #4803
* [Hopefully] enables manual running of the overnight benchmark GHA, starting from a custom commit.

---
[Consult Iris pull request check list]( https://scitools-iris.readthedocs.io/en/latest/developers_guide/contributing_pull_request_checklist.html)

PR #5005, opened 2022-09-29
Correct wrong type of equality check in benchmark GHA.

trexfeathers:
# 😳

In #5001 I intended to generate `first_commit` ONLY IF it was not already provided by the `workflow_dispatch`, but I got my equality the wrong way round.


trexfeathers:
Life-saver, thanks again @stephenworsley!

PR #5010, opened 2022-10-03
Iris glossary what's new fix

ESadek-MO:
### Description
Fixed the what's new entry for glossary to link to glossary


pp-mo:
Looks fine but history is a problem -- I think you should rebase this onto current main
( or if you prefer, cherry-pick the last commit onto a fresh branch based on main )

Unfortunately, merging back to your branch is **_not_** the right way to resolve this, since we squashed when merging your original PR back to main :  We generally use such squashed merges, which means the individual commits seen here do not appear on the main history, only the single squashed commit https://github.com/SciTools/iris/commit/96c00b6138e296aa6a56f2f215ec0355b753b6a4.

pp-mo (changes requested):
Functionally looks fine, but the history is a problem -- I think you should rebase this onto current main
( or if you prefer, cherry-pick the last commit onto a fresh branch based on main )

Unfortunately, merging back to your branch is **_not_** the right way to resolve this, since we squashed when merging your original PR back to main :  We generally use such squashed merges for Iris PRs (**), which means the individual commits seen here do not appear on the main history, only the single squashed commit https://github.com/SciTools/iris/commit/96c00b6138e296aa6a56f2f215ec0355b753b6a4.

(**) except for release- and feature-branches, where individual commit history is important, so we **_do_** use merge commits.  But you need a repo owner to enable them, as they are disabled by default.

ESadek-MO:
> 

Thanks for catching this! SHall close it

PR #5013, opened 2022-10-04
Sample code to convert ORCA data into a meshcube.

pp-mo:
Fulfils a long-ago promise to write up + share how to make meshes for ORCA data.

Test docs-build of this branch, showing : [the new section](https://pp-iris.readthedocs.io/en/orca_mesh_example/further_topics/ugrid/other_meshes.html#nemo-data-on-orca-tripolar-grid)


trexfeathers (changes requested):
I haven't read through yet, but the image doesn't fit with the others on the page. The other images prioritise showing the kind of grid/mesh that's being converted, even if the image is not part of the script. Could you find something illustrative of an ORCA grid, either from the Internet or using GeoVista?

pp-mo:
> I haven't read through yet, but the image doesn't fit with the others on the page. 

Okay, that is a good call !
I've now replaced that file now with something hopefully more suitable.

N.B. now also provided a test docs-build [here](https://pp-iris.readthedocs.io/en/orca_mesh_example/further_topics/ugrid/other_meshes.html#nemo-data-on-orca-tripolar-grid)

trexfeathers (changes requested):
Looks great, thanks @pp-mo!

Definitely deserves a What's New entry

trexfeathers (approved):

trexfeathers:
GHA is being weird. I'm merging anyway - the offending job passed for previous near-identical commits AND passes for other Python versions.

PR #5017, opened 2022-10-07
Lazy weighted RMS calculation

rcomer:
### Description

Lazy aggregation with weights was not previously implemented for the RMS aggregator because of a bug in the weighted masked case reported at https://github.com/dask/dask/issues/3846 (see removed code comments).  This bug was fixed at https://github.com/dask/dask/pull/4236, and the fix included in [version 1.1.0](https://docs.dask.org/en/stable/changelog.html#v1-1-0-2019-01-18).  Iris's minimum supported version of Dask is now [2.26](https://github.com/SciTools/iris/blob/d1f3e4564e09de1e269872f5d14d4d3bf3ce5489/setup.cfg#L53).

---

### DIFF
--- a/docs/src/whatsnew/latest.rst
+++ b/docs/src/whatsnew/latest.rst
@@ -36,6 +36,9 @@ This document explains the changes made to Iris for this release
    :ref:`documentation page<community_plugins>` for further information.
    (:pull:`5144`)
 
+#. `@rcomer`_ enabled lazy evaluation of :obj:`~iris.analysis.RMS` calcuations
+   with weights. (:pull:`5016`)
+
 
 🐛 Bugs Fixed
 =============
--- a/docs/src/whatsnew/latest.rst
+++ b/docs/src/whatsnew/latest.rst
@@ -37,7 +37,7 @@ This document explains the changes made to Iris for this release
    (:pull:`5144`)
 
 #. `@rcomer`_ enabled lazy evaluation of :obj:`~iris.analysis.RMS` calcuations
-   with weights. (:pull:`5016`)
+   with weights. (:pull:`5017`)
 
 
 🐛 Bugs Fixed
--- a/lib/iris/analysis/__init__.py
+++ b/lib/iris/analysis/__init__.py
@@ -1583,27 +1583,15 @@ def _lazy_max_run(array, axis=-1, **kwargs):
 
 
 def _rms(array, axis, **kwargs):
-    # XXX due to the current limitations in `da.average` (see below), maintain
-    # an explicit non-lazy aggregation function for now.
-    # Note: retaining this function also means that if weights are passed to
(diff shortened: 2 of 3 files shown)


lib/iris/analysis/__init__.py
```diff
@@ -1584,27 +1584,19 @@ def _lazy_max_run(array, axis=-1, **kwargs):
 
 
 def _rms(array, axis, **kwargs):
-    # XXX due to the current limitations in `da.average` (see below), maintain
-    # an explicit non-lazy aggregation function for now.
```
rcomer, line 1588: I'm guessing this comment about explicit non-lazy function is because previously some aggregators used a dask function for both lazy and non-lazy cases.  We had a rethink on that at #4905.

lib/iris/analysis/__init__.py
```diff
@@ -1584,27 +1584,19 @@ def _lazy_max_run(array, axis=-1, **kwargs):
...
-    # Note: retaining this function also means that if weights are passed to
-    # the lazy aggregator, the aggregation will fall back to using this
-    # non-lazy aggregator.
-    rval = np.sqrt(ma.average(np.square(array), axis=axis, **kwargs))
-    if not ma.isMaskedArray(array):
-        rval = np.asarray(rval)
```
rcomer, line 1594: I've not preserved this handling to revert to a plain numpy array because it's inconsistent with the MEAN aggregator, and I don't think there is any way to make the lazy case equivalent.  Note that `aggregated_by` has its own handling anyway:

https://github.com/SciTools/iris/blob/d1f3e4564e09de1e269872f5d14d4d3bf3ce5489/lib/iris/cube.py#L4167-L4174

though I can't see any equivalent of that in `collapsed` or `rolling_window`....

bjlittle:
@rcomer do you have time to rebase this PR? :smile:

docs/src/whatsnew/latest.rst
```diff
@@ -36,6 +36,9 @@ This document explains the changes made to Iris for this release
    :ref:`documentation page<community_plugins>` for further information.
    (:pull:`5144`)
 
+#. `@rcomer`_ enabled lazy evaluation of :obj:`~iris.analysis.RMS` calcuations
+   with weights. (:pull:`5016`)
```
ESadek-MO, line 40: Should be 5017?

ESadek-MO (changes requested):
Hey @rcomer! Looks great, I think the PR number might be wrong in the What's New, but otherwise looks good to go!

docs/src/whatsnew/latest.rst
rcomer: Ooops, done.  Thanks @ESadek-MO!

ESadek-MO (approved):

Issue #4860, opened 2022-07-12
`MeshCoord` constructor always uses metadata of node coordinate regardless of `location`


schlunma:
The [`MeshCoord` constructor](https://github.com/SciTools/iris/blob/2cbaec747820410cce41db004a41e68d6ff63ee8/lib/iris/experimental/ugrid/mesh.py#L2758) always uses the metadata (`standard_name`, `long_name`, `units`, etc.) of the node coordinate, regardless of the `location` that is specified:

https://github.com/SciTools/iris/blob/2cbaec747820410cce41db004a41e68d6ff63ee8/lib/iris/experimental/ugrid/mesh.py#L2845-L2853

That feels really weird to me and can lead to unexpected behavior. For example, if you compare the original face coordinate that is used to construct a `Mesh` and compare it to the output of `Mesh.to_MeshCoord` you get different results:

```py
# define node coordinates
(node_lat, node_lon) = ...

# define face coordinates with different metadata
(face_lat, face_lon) = ...

# construct mesh with it
mesh = Mesh(                                                                                                                                                                                                                                                           
            topology_dimension=2,                                                                                                                                                                                                                                              
            node_coords_and_axes=[(node_lat, 'y'), (node_lon, 'x')],                                                                                                                                                                                                           
            connectivities=[connectivity],                                                                                                                                                                                                                                     
            face_coords_and_axes=[(face_lat, 'y'), (face_lon, 'x')])

# extract MeshCoord
mesh_face_lat = mesh.to_MeshCoord('face', 'y')
```

Here, the metadata of `face_lat` and `mesh_face_lat` differ since the metadata of `face_lat` and `node_lat` differ.                                                                                                                     

 I am not sure if this is an intended feature or not, so please feel free to just close this if this is intentional.

Cheers :+1: 

***
Iris version: `3.2.1.post0`

bjlittle:
@schlunma Thanks for taking the time to report this, much appreciated!

I'm particularly excited simply because it means that you're using our experimental `ugrid` support, which is brilliant :star_struck: 

We're mostly all busy this week attending SciPy 2022 (13-15 July), but we'll get back to you asap :+1: 

I'll class this issue as a `bug`, until we know otherwise.

schlunma:
Thanks @Bill, that's great!

Yes, I've used it quite extensively in the last couple of weeks (also in combination with iris-esmf-regrid), and I'm really happy with it :+1: Thanks for all the great work on this :rocket:

[...]

PR #5020, opened 2022-10-10
Meshcoord metadata follows face/edge coords where present.

pp-mo:
Addresses #4860


lib/iris/experimental/ugrid/mesh.py
```diff
@@ -2839,17 +2839,77 @@ def __init__(
...
+                    # But.. if we have standard-names to compare, ignore long-names.
+                    node_metadict["long_name"] = None
+                for key in ("standard_name", "long_name", "units"):
+                    # N.B. don't test *attributes* for comparison :  As long as the
+                    # node-coord otherwise matches the bounds-coord, ignore its
+                    # attributes and take those of the bounds-coord instead.
+                    if key == "units":
+                        empty_val = unit_unknown
+                    elif key == "attributes":
+                        empty_val = {}
```
pp-mo, line 2885: Obsolete !

pp-mo, line 2885: This bit now obsolete !

pp-mo:
Ping @schlunma
Hoping you may be able to test against this for usability in light of #4860 ?

This code should give an idea of whether the idea works.
Though it is still very experimental, and I may well make further changes in following days.

**Please note**
I am proposing to "outlaw" clashes of phenomenon identity (including units) between node- and face/edge-coords.
This may go against some of your suggested testcases in #4860, but ...
  * I believe Iris simply _cannot_ tolerate any conflict of  phenomenon or units between a Coord points and its bounds, since we don't treat bounds as a separate object.
  * so likewise, for MeshCoords, that applies also to any disagreements between node- and face/edge-coords metadata

The currently-proposed code **_is_** allowing matches where one of the two coords is missing a standard/long-name or units property.  But it makes **_no_** requirement for attributes to match. 
So ... these are precisely the questions that we need to resolve, for which **your opinions + testing are much wanted !**

schlunma:
Thanks @pp-mo for these changes!

I fully agree with your suggestions here that iris cannot accept node and face/edge coordinate metadata to be different. The main problem I'm having right now and described in #4860 is that currently they **can** differ (which even can lead to wrong results when the units differ!), so not allowing these to differ in first place also solves my problem.

Three comments on this:

1. Is there a particular reason the `var_name` is fixed to `None`? What's the reason this is not treated similarily to `long_name` (i.e., ignored when `standard_name` is not set)? This is actually the largest problem we have in ESMValTool, since we except the input coordinates to have specific names given by the CMIP data request, and this hardcoded `None` causes our checks to fail. For all other metadata we could solve this with a workaround (in theory).
2. I am not entirely sure if matches should be allowed if one of them is empty. Especially for the units this seems dangerous, as you might still end up with `points` or `bounds` not matching their `units`. Thus, I'd rather not allow matches when one of the values is missing. I'm fine with ignoring the long name if the standard name is set and with ignoring the `attributes` altogether.
3. I still believe that it would make sense to compare the `bounds` of the face/edge coordinate (if they are not empty) with the corresponding values calculated from the connectivity and the node coordinate in the `MeshCoord` constructor. Yes, it's just a warning in the UGRID checker, but in my opinion this is an inconsistency in the data that should not be allowed.

Unfortunately I probably won't have the chance to actually test this before the ESMValTool workshop, but the code changes look really good already. I am happy to focus on this during the workshop.

Cheers, and thanks again for all your work on this! 🚀

pp-mo:
Some initial answers on your points ...
Hopefully, we get to talk next week !

> Is there a particular reason the var_name is fixed to None? What's the reason this is not treated similarily to long_name 

The iris ".var_name" carries the original file identity of an Iris object which was loaded from a variable in a netcdf file  -- and it likewise determines variable name on writing out to netcdf.
A MeshCoord, however, is basically a "convenience" object that has no state of its own, does _not_ represent any netcdf variable, and is never written to an output file (instead, the referenced mesh coordinates and connectivities would be).  Since it never was or can be a netcdf variable, a var_name is not appropriate.

----

> I still believe that it would make sense to compare the bounds of the face/edge coordinate (if they are not empty) with the corresponding values calculated from the connectivity and the node coordinate in the MeshCoord constructor. Yes, it's just a warning in the UGRID checker, but in my opinion this is an inconsistency in the data that should not be allowed.

I think we rejected this because of concerns about performance costs with large meshes. The UGRID checker only checks it if the arrays are "small enough", with a user-defined memory size threshold.  Which I think works OK for a command utility, but is not really appropriate for Iris.
The UGRID spec also seems to state that, if they exist, bounds values "should" be redundant.  And in particular, that they cannot be used **_instead of_** a connectivity variable.  So, we felt it was justified to simply ignore them and bypass the problems of checking.  
Iris [explicitly avoids adding bounds to a coordinate in these cases](https://github.com/SciTools/iris/blob/ea67311f24ef0b039976adefb31007b7eece51cb/lib/iris/experimental/ugrid/load.py#L243-L246), so it  treats this information as "irrelevant" :   Since, from Iris PoV, it cannot cause any conflict, I think not checking it is defensible.

Though ... the ugrid-checker position is a bit different, since unlike Iris it is intended to cover _all_ UGRID (**).  
So maybe this **_should_** be an error, but at the time I felt the door was still open to bounds coding something _different_ from the product of the connectivity and node-coordinates.  So as currently proposed it is [a "should" rather than a "must"](https://ugrid-conventions.readthedocs.io/en/conformance/conformance/#A205)
So it might be worth re-visiting this choice.  However, note that **_all_** of this is rather lacking oversight by anyone but me, and not yet [adopted by UGRID](https://github.com/ugrid-conventions/ugrid-conventions/pull/54).

(**) with the rather notable exception of 3d meshes / volumes !

pp-mo:
This one is maybe a bit trickier ...

>  I am not entirely sure if matches should be allowed if one of them is empty. Especially for the units this seems dangerous, as you might still end up with points or bounds not matching their units.

As the meaning of the 'coordinate' metadata in UGRID is nowhere explicitly stated, I'm not confident that there is a solid logic to determine on this.  So I wanted to be _very tolerant_ of what we find in a file, since in the past Iris loading code has often proved too strict, and we have had to install extra workarounds + issue warnings so that it can cope with "incorrect" non-CF input rather than failing to load it altogether.

Details...   in this case, I reckon that we could+should cope if one or other of the "face/edge" and "node" coordinates does not have a usable identity **_at all_** (i.e. a name or units).  (maybe this could be seen like a regular bounds-variable, getting its identity from the parent variable).   I am also allowing the other to be missing a name or units, so that only a _definite conflict_ will cause an error.  
Note, however, that in any case it will not _mix_ information from the two possible metadata-s, but only choose _which_ of them should be used.

However, I now see that when something **_does_** go wrong by this definition, it will cause a load failure.
So perhaps I should  **adapt that strategy to raise a warning instead** --  do you think that would be more useful ?
Certainly, it would then make more sense to complain about _other_ unexpected occurences (like name vs no-name), if they are raised as warnings rather than fatal.

N.B. at present, this is rather undermined by the Mesh class, which currently **_requires_** coordinates to have names that identify an axis.
But, I have argued elsewhere that this is not really in the spirit of UGRID, and I'm hoping we can fix it to make it more tolerant in future - see https://github.com/SciTools/iris/discussions/4438, point (2).

pp-mo:
From offline discussion with @bjlittle @schlunma ..
  * for convenience in cmor checking within ESMValTool, **_really_** would like MeshCoords to have a (non-empty) `var_name`
    Specifically, for compatibility with "ordinary" coords, want to be able to extract a MeshCoord with `cube.coord(var_name=name)`
    * so, this is not really a big deal for Iris, we will just copy `var_name` with the other metadata
  * @schlunma still thinking that accepting mismatch between name or units of face-coord and node-coord is unwise : 
     i.e. should really be an error, not a warning.  But one missing (e.g. no units) probably can be tolerant, as currently proposed
       * ok!
  * explained that Iris is ignoring (not loading) bounds of face/edge/node coordinates when loading mesh info,
    so checking would mean allowing that
    * probably not worth it -- can leave out

@bjlittle standing ready to merge !
But we also have some outstanding test errors to fix (and possibly some missing tests, TBD) ..

pp-mo:
Replaced earlier code with much simpler implementation (and tests likewise), 
since I realised that the current Mesh limitations mean that many of the complex cases it was trying to address can't actually occur (!)

pp-mo:
The new, intended behaviour, is that:
  * a face/edge MeshCoord now copies **_all_** the metadata from the face/edge ("points") coord
    -- i.e., now including `var_name`.  
  * an error is raised if the face/edge and node coords have different `standard_name` or `units`
     -- but different `long_name`/`attributes` are allowed/ignored (also `var_name`, obviously).
     This also now provides that a provided name/unit will _not_ match a missing one.

pp-mo:
I think I've finally got somewhere with this.
@schlunma will this work for you now ?
@bjlittle care to review ?
Thanks

schlunma (commented):
Thanks for all these changes @pp-mo ! I just tested this within ESMValTool and it indeed solves our problems :tada:

I didn't test this in too much detail (e.g., if it successfully fails the metadata of the nodes and faces differs), but the code looks fine to me :+1: Cheers, thanks for all your help on this!!

bjlittle (changes requested):
@pp-mo This looks really good to me 🤩 

Just a minor comment to service, otherwise let's get this banked 👍

bjlittle:
@pp-mo Do you fancy also adding a `whatsnew` entry?

pp-mo:
Thanks @bjlittle hope that does ok

PR #5024, opened 2022-10-12
Load+save things other than filepaths, and specifically support netCDF4.Dataset-s

pp-mo:
Replaces the iris-load changes part of #4835,
So, re-presented as a separate issue which should be much simpler.

TBD: reform + add the other parts of that logic, in subsequent PRs.

NOTE: 
I think that, whilst we are targetting a feature-branch, it _isn't_ generally appropriate to add whatsnew-s.
So, we can sort out how to announce+explain it all when we merge back the FB.
Deferring the changes also probably avoids a whole lot of trivial conflicts whenever we do merge main-to-FB or vice versa.

### DIFF
--- a/lib/iris/fileformats/netcdf/loader.py
+++ b/lib/iris/fileformats/netcdf/loader.py
@@ -13,6 +13,7 @@ and `netCDF4 python module <https://github.com/Unidata/netcdf4-python>`_.
 Also : `CF Conventions <https://cfconventions.org/>`_.
 
 """
+from collections import Iterable
 import warnings
 
 import netCDF4
@@ -529,7 +530,7 @@ def load_cubes(filenames, callback=None, constraints=None):
     # Create an actions engine.
     engine = _actions_engine()
 
-    if isinstance(filenames, str):
+    if isinstance(filenames, str) or not isinstance(filenames, Iterable):
         filenames = [filenames]
 
     for filename in filenames:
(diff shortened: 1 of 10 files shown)


lib/iris/fileformats/netcdf/loader.py
```diff
@@ -587,7 +587,7 @@ def load_cubes(file_sources, callback=None, constraints=None):
                 warnings.warn("{}".format(e))
 
             # Perform any user registered callback function.
-            cube = run_callback(callback, cube, cf_var, cf.filename)
+            cube = run_callback(callback, cube, cf_var, file_source)
```
pp-mo, line 590: I think this makes more sense.
I believe it is consistent with what we do for UM "structured loading"
see "Use of callbacks" in [structured_um_loading api docs](https://scitools-iris.readthedocs.io/en/stable/generated/api/iris/fileformats/um.html?highlight=structured#iris.fileformats.um.structured_um_loading)

pp-mo:
Closing in favour of #5214 

Which now targets main, and required a lot of rebasing !
The F-B rather hid the amount of work that was required to integrate these changes, but I don't see the point in updating the F-B since plans have changed : we now intend to host the "bridge" code outside the Iris repo.
I will also remove the F-B branch.

pp-mo:
Closing in favour of #5214 

Which now targets main, and required a lot of rebasing !
The F-B rather hid the amount of work that was required to integrate these changes, but I don't see the point in updating the F-B since plans have changed : we now intend to host the "bridge" code outside the Iris repo.
I will also remove the F-B branch.

PR #5029, opened 2022-10-18
Prioritise dim coord in `_get_lon_lat_coords()`

### LINKED ISSUE #3916: Can't get area weights of cube with additional auxillary coordinates of lat and lon
## 🐛 Bug Report
Iris is unable to calculate area weights for rotated pole grids if auxiliary coordinates with true latitude and longitude have been added to the cube.

## How To Reproduce

1.  load data on a rotated pole grid
```
# load rotated pole grids
fname = iris.sample_data_path("rotated_pole.nc")
cube = iris.load_cube(fname)
```
2. attach the aux coords thus:
```
# get cube's coordinate system
cs = cube.coord_system()
# get coord names
# Longitude
xcoord = cube.coord(axis="X", dim_coords=True)
# Latitude
ycoord = cube.coord(axis="Y", dim_coords=True)

# read in the grid lat/lon points from the cube
glat = cube.coord(ycoord).points
glon = cube.coord(xcoord).points

# create a rectangular grid out of an array of
# glon and glat values, shape will be len(glat)xlen(glon)
x, y = np.meshgrid(glon, glat)

# get the cube dimensions which corresponds to glon and glat
x_dim = cube.coord_dims(xcoord)[0]
y_dim = cube.coord_dims(ycoord)[0]

# define two new variables to hold the unrotated coordinates
rlongitude, rlatitude = iris.analysis.cartography.unrotate_pole(
    x, y, cs.grid_north_pole_longitude, cs.grid_north_pole_latitude
)

# create two new auxillary coordinates to hold
# the values of the unrotated coordinates
reg_long = iris.coords.AuxCoord(rlongitude, long_name="longitude", units="degrees")
reg_lat = iris.coords.AuxCoord(rlatitude, long_name="latitude", units="degrees")

# add two auxilary coordinates to the cube holding
# regular(unrotated) lat/lon values
cube.add_aux_coord(reg_long, [y_dim, x_dim])
cube.add_aux_coord(reg_lat, [y_dim, x_dim])

```
3. print the cube:

```
air_temperature / (K)               (grid_latitude: 550; grid_longitude: 700)
     Dimension coordinates:
          grid_latitude                           x                    -
          grid_longitude                          -                    x
     Auxiliary coordinates:
          latitude                                x                    x
          longitude                               x                    x

4. try and get area weights:
weights =iris.analysis.cartography.area_weights(cube)
```
  File "area_bug.py", line 49, in <module>
    weights =iris.analysis.cartography.area_weights(cube)
  File ".../lib/python3.6/site-packages/iris/analysis/cartography.py", line 399, in area_weights
    lon, lat = _get_lon_lat_coords(cube)
  File ".../python3.6/site-packages/iris/analysis/cartography.py", line 189, in _get_lon_lat_coords
    "Calling `_get_lon_lat_coords` with multiple lat or lon coords"
ValueError: Calling `_get_lon_lat_coords` with multiple lat or lon coords is currently disallowed
```
5. remove the extra coords - works fine:

cube.remove_coord('longitude')
cube.remove_coord('latitude')
weights =iris.analysis.cartography.area_weights(cube)

```
## Expected behaviour
I expect that a cube like this should return the correct area weights

## Environment 
 - RHEL 7.9 
 - Iris Version: 2.2.0

## Additional context

<details>

```
https://github.com/SciTools/iris/blob/master/lib/iris/analys... 

I would suggest that this: 

coord for coord in cube.coords() if "latitude" in coord.name() 

 

should actually be: 

coord for coord in cube.coords(dim_coords=True) if "latitude" in coord.name()
```
See [here](https://gist.github.com/pierrejoubert73/902cc94d79424356a8d20be2b382e1ab) for further details.
</details>

ESadek-MO:
## 🚀 Pull Request

Closes #3916.

Added dim coord prioritisation to `_get_lon_lat_coords()` in `iris.analysis.cartography`. This allows `iris.analysis.cartography.area_weights()` and `iris.analysis.cartography.project()` to handle cubes which contain both dim and aux coords of the same type e.g. `longitude` and `grid_longitude`.


lib/iris/tests/unit/analysis/cartography/test__get_lon_lat_coords.py
```diff
@@ -0,0 +1,79 @@
...
+
+import pytest
+
+from iris.analysis.cartography import _get_lon_lat_coords as g_lat_lon
+from iris.coords import AuxCoord
+from iris.tests.stock import lat_lon_cube
+
+dim_only_cube = lat_lon_cube()
+lon_dim = dim_only_cube.coord("longitude")
+lat_dim = dim_only_cube.coord("latitude")
```
stephenworsley, line 16: The style of test seems very different to the existing style for most tests. I realise this is probably due to this only now being possible due to using pytest, however there's still some things I'm unused to here which I'd like to make sure are appropriate in the new pytest context.

That is, running code outside of test functions. I'm used to all or most code required for a test being run inside the function for that test. It seems that the pattern of running code outside these functions could possibly yield performance improvements, though perhaps at the cost of robustnesss.

lib/iris/tests/unit/analysis/cartography/test__get_lon_lat_coords.py
```diff
@@ -0,0 +1,79 @@
...
+dim_aux_cube.add_aux_coord(lon_aux, 1)
+
+
+def test_dim_and_aux():
+    lon, lat = g_lat_lon(dim_aux_cube)
+    assert lon == lon_dim
+    assert lat == lat_dim
+
+
+aux_only_cube = dim_aux_cube.copy(dim_aux_cube.data)
```
stephenworsley, line 40: I'm uneasy with the pattern of re-using one object between tests (here, dim_aux_cube.data). I can appreciate that this could reduce memory use, though there still seems to be an inconsistency between copying and calling `lat_lon_cube()`

stephenworsley (changes requested):
This looks good.
My comments here are mostly about the style of the tests. These seem to be more in line with pytest, which we have relatively recently adopted, so I'm not against a stylistic change in principle. If there is an example in our code of tests in this style I would be interested in comparing it. Though if this is setting a precedent then I think it might want a broader discussion about what precedents we want to set before departing from our existing style.

trexfeathers:
@PaytonLeedham is going to give this a once over since we're all still getting used to pytest!

lib/iris/tests/unit/analysis/cartography/test__get_lon_lat_coords.py
```diff
@@ -0,0 +1,114 @@
...
+import pytest
+
+from iris.analysis.cartography import _get_lon_lat_coords as g_lon_lat
+from iris.coords import AuxCoord
+from iris.tests.stock import lat_lon_cube
+
+
+@pytest.fixture
+def dim_only_cube():
+    yield lat_lon_cube()
```
stephenworsley, line 17: I believe this should work as just `return`. From what I can tell of pytest, `yield` tends to be used when there is teardown code to be run afterwards.

lib/iris/tests/unit/analysis/cartography/test__get_lon_lat_coords.py
```diff
@@ -0,0 +1,114 @@
...
+    lat_aux = AuxCoord.from_coord(lat_dim)
+    lat_aux.standard_name = "grid_latitude"
+    lon_aux = AuxCoord.from_coord(lon_dim)
+    lon_aux.standard_name = "grid_longitude"
+
+    dim_aux_cube = dim_only_cube
+    dim_aux_cube.add_aux_coord(lat_aux, 0)
+    dim_aux_cube.add_aux_coord(lon_aux, 1)
+
+    yield dim_aux_cube
```
stephenworsley, line 42: as above, yield vs return.

lib/iris/tests/unit/analysis/cartography/test__get_lon_lat_coords.py
```diff
@@ -0,0 +1,114 @@
...
+
+@pytest.fixture
+def aux_only_cube(dim_aux_cube):
+    lon_dim, lat_dim = dim_aux_cube.dim_coords
+
+    aux_only_cube = dim_aux_cube
+    aux_only_cube.remove_coord(lon_dim)
+    aux_only_cube.remove_coord(lat_dim)
+
+    yield dim_aux_cube
```
stephenworsley, line 62: as above, yield vs return.

lib/iris/tests/unit/analysis/cartography/test__get_lon_lat_coords.py
```diff
@@ -0,0 +1,114 @@
...
+    assert lon == aux_lon
+    assert lat == aux_lat
+
+
+@pytest.fixture
+def double_dim_cube(dim_only_cube):
+    double_dim_cube = dim_only_cube
+    double_dim_cube.coord("latitude").standard_name = "grid_longitude"
+
+    yield double_dim_cube
```
stephenworsley, line 79: as above, yield vs return.

lib/iris/tests/unit/analysis/cartography/test__get_lon_lat_coords.py
```diff
@@ -0,0 +1,114 @@
...
+    with pytest.raises(ValueError, match=t_error_message):
+        g_lon_lat(double_dim_cube)
+
+
+@pytest.fixture
+def double_aux_cube(aux_only_cube):
+    double_aux_cube = aux_only_cube
+    double_aux_cube.coord("grid_latitude").standard_name = "longitude"
+
+    yield double_aux_cube
```
stephenworsley, line 94: as above, yield vs return.

lib/iris/tests/unit/analysis/cartography/test__get_lon_lat_coords.py
```diff
@@ -0,0 +1,114 @@
...
+    with pytest.raises(ValueError, match=t_error_message):
+        g_lon_lat(double_aux_cube)
+
+
+@pytest.fixture
+def missing_lat_cube(dim_only_cube):
+    missing_lat_cube = dim_only_cube
+    missing_lat_cube.remove_coord("latitude")
+
+    yield missing_lat_cube
```
stephenworsley, line 109: as above, yield vs return.

stephenworsley (changes requested):
Looks good, just one minor point on the use of yield vs return. As far as I'm aware, using yield causes pytest to use a little extra machinery to remember and run the rest of the code in these fixtures. I'm not sure of all the details of how pytest deals with yield with no code afterwards vs how it deals with return, but it seems like using return is the simpler case and therefore better practice to use when yield is not necessary. This seems to be the pattern in the [pytest documentation](https://docs.pytest.org/en/6.2.x/fixture.html).

lib/iris/tests/unit/analysis/cartography/test__get_lon_lat_coords.py
ESadek-MO: I was originally under the impression that they were interchangable, in which case the one with the most potential functionality would've made most sense. In hindsight, they wouldn't have both for no reason (I'd hope) so changed it back!

stephenworsley:
One more thought, it might be worth adding a note here to say which type of coordinate you are prioritising:
https://github.com/SciTools/iris/blob/d56824cbc892356302c135ee23d704ff01e12d71/lib/iris/analysis/cartography.py#L581

Something like: 
```
If there are both 1 and 2 dimensional latitude-longitude coordinates, only
the 1 dimensional coordinates will be used.
```

stephenworsley (approved):
Looks good!

Issue #4959, opened 2022-09-12
Add user advice on Dask "best practices" 


pp-mo:
configuration, distributed scheduling etc

Promote dask/distributed
Demo efficient approaches for large-scale processing
```[tasklist]
### Tasks
- [ ] https://github.com/SciTools/iris/pull/5190
- [ ] https://github.com/SciTools/iris/issues/5344
```

Issue #2063, opened 2016-06-28
Segmentation fault when differencing cubes


PAGWatson:
Hello,

I get a seg fault when I take the difference between two cubes and then try to access the data. I've attached the files containing the data at https://groups.google.com/forum/#!topic/scitools-iris/OgFbHKtNGqU .

```
cube1=iris.load_cube('get_autocorr_data_lab0_tp_2.5x2.5_lat_-90.0-90.0_daily_mean_1998-2007_lag1.nc')
cube2=iris.load_cube('get_autocorr_data_gpcp_tp_2.5x2.5_lat_-90.0-90.0_daily_mean_1998-2007_lag1.nc')
diff=cube1-cube2
diff_data=diff.data  #The seg fault happens at this line
```

However, replacing the last two lines with the line below works fine.

```
diff_data=cube1.data - cube2.data
```

(I'm using Iris 1.9.2 in ipython 4.0.3.)

Andrew Dawson reproduced the issue and posted the full traceback at the above link.

shoyer:
As @ajdawson notes "Interestingly, it is sufficient to load only one of the cubes data payloads for this bug to disappear... This should be reported as a bug on the iris issue tracker."

I'm pretty sure the solution is that Iris needs to use a thread lock when accessing data from netCDF4, because the HDF5 library is not thread safe. We use such a thread lock in xarray/dask.

ajdawson:
Thanks @shoyer this is useful information. I did a quick test by adding a global lock to data reads within a `biggus.ProducerNode`, which does seem to prevent the error described by @PAGWatson.

marqh:
https://github.com/SciTools/iris/blob/master/INSTALL#L89

also provides a suggestion that there may be an issue here if the HDF5 build is not thread safe.  I have tripped over this in the past

@PAGWatson is this still an open issue, from your p.o.v.?

PAGWatson:
Well I can work around this, so it's not a huge problem for me for now. I've not checked if the bug is still present in the latest version of Iris, though.

shoyer:
There's no harm in adding a defensive `threading.Lock()` around HDF5/netCDF4 calls, given that HDF5 doesn't support multi-threading anyways.

mheikenfeld:
I've run into the same issue over the last couple of days (iris 1.10.0). Any updates on this? I can work around it as well for now, but seems like something that people are very prone to run into and seems hard to make sense of since all you get is the segmentation fault.

[...]

PR #5031, opened 2022-10-21
Basic functional lazy saving.

pp-mo:
@bjlittle @bouweandela following discussion yesterday, I couldn't resist having a bash at lazy saving.
So here it is!

This is a slightly crude first attempt, but seems to basically work.
In principle, addresses #4190 
@bouweandela **can you see if this actually functions for parallelising + whether it shows a benefit ?**
Sorry again that I can't be there this morning

In yesterday's discussion, I got a much clearer idea of what is needed.
I think also @zklaus explained to me what the xarray implementation actually does.
So I've tried to copy that idea.

<details>
<summary>My simple test code [click to expand] </b></summary>

```
import iris
from iris.tests.stock import realistic_4d
import dask.array as da

# Just avoid a warning
iris.FUTURE.datum_support = True

def make_lazy_datacube():
    cube = realistic_4d()
    # Replace data + aux-coords with lazy content.
    def lazy_like(array):
        if array is not None:
            dmin, dmax =  array.min(), array.max()
            chunks = list(array.shape)
            chunks[0] = 1
            array = da.random.uniform(dmin, dmax, size=array.shape, chunks=chunks)
        return array
    cube.data = lazy_like(cube.data)
    auxcoord = cube.coord('surface_altitude')
    auxcoord.points = lazy_like(auxcoord.points)
    return cube


if __name__ == '__main__':
    cube = make_lazy_datacube()
    print(cube)
    print('')
    print('Cube details ...')
    print('Core data:', cube.core_data())
    print('Alt coord points:', cube.coord('surface_altitude').core_points())
    def surface_sample(cube):
        return cube.coord('surface_altitude').core_points().flatten()[:10].compute()

    samp = surface_sample(cube)
    print('Alt coord points actual sample values', samp)
    print('')
    print('Lazy save..')
    lazy_save = iris.save(cube, 'tmp.nc', compute=False)
    print('.. done.')
    print('result = ', lazy_save)
    readback = iris.load_cube('tmp.nc', 'air_potential_temperature')
    print('Readback altitude:', surface_sample(readback))
    print('')
    print('Now complete the lazy save..')
    lazy_save.compute()
    print('..done.')
    print('Readback altitude:', surface_sample(readback))
```

</details>

<details>
<summary>..and sample resulting output [click to expand] </b></summary>

```
air_potential_temperature / (K)     (time: 6; model_level_number: 70; grid_latitude: 100; grid_longitude: 100)
    Dimension coordinates:
        time                             x                      -                  -                    -
        model_level_number               -                      x                  -                    -
        grid_latitude                    -                      -                  x                    -
        grid_longitude                   -                      -                  -                    x
    Auxiliary coordinates:
        level_height                     -                      x                  -                    -
        sigma                            -                      x                  -                    -
        surface_altitude                 -                      -                  x                    x
    Derived coordinates:
        altitude                         -                      x                  x                    x
    Scalar coordinates:
        forecast_period             0.0 hours
    Attributes:
        source                      'Iris test case'

Cube details ...
Core data: dask.array<uniform, shape=(6, 70, 100, 100), dtype=float64, chunksize=(1, 70, 100, 100), chunktype=numpy.ndarray>
Alt coord points: dask.array<uniform, shape=(100, 100), dtype=float64, chunksize=(1, 100), chunktype=numpy.ndarray>
Alt coord points actual sample values [435.58198889 475.71775563 421.56953978 393.62885393 317.8254836
 464.83843177 317.20213419 246.2240947  367.07881807 224.52397621]

Lazy save..
.. done.
result =  Delayed('combined_delayeds-abbd0f8e-5712-4d49-af4e-6e4d760858f0')
Readback altitude: [-- -- -- -- -- -- -- -- -- --]

Now complete the lazy save..
..done.
Readback altitude: [435.58198889 475.71775563 421.56953978 393.62885393 317.8254836
 464.83843177 317.20213419 246.2240947  367.07881807 224.52397621]
```

</details>

### DIFF
--- a/lib/iris/fileformats/netcdf/saver.py
+++ b/lib/iris/fileformats/netcdf/saver.py
@@ -25,6 +25,8 @@ import warnings
 import cf_units
 import dask
 import dask.array as da
+from dask.utils import SerializableLock
+import filelock
 import netCDF4
 import numpy as np
 import numpy.ma as ma
@@ -499,37 +501,31 @@ class DeferredSaveWrapper:
     TODO: could be improved with a caching scheme, but this just about works.
     """
 
-    def __init__(self, filepath, cf_var):
-        # Grab useful properties of the variable, including the identifying 'name'.
+    def __init__(self, filepath, cf_var, lockfile_path):
         self.path = filepath
-        for key in ("shape", "dtype", "ndim", "name"):
-            setattr(self, key, getattr(cf_var, key))
+        self.varname = cf_var.name
+        self.lockfile_path = lockfile_path
 
     def __setitem__(self, keys, array_data):
         # Write to the variable.
-        # Re-open the file for writing.
-        dataset = netCDF4.Dataset(self.path, "r+")
+        # First acquire a file-specific lock
+        # Importantly, in working via the file-system, this is common to all workers,
+        # even when using processes or distributed.
+        lock = filelock.FileLock(self.lockfile_path)
+        lock.acquire()
+        # Now re-open the file for writing + write to the specific file variable.
+        dataset = None
         try:
-            var = dataset.variables[self.name]
+            dataset = netCDF4.Dataset(self.path, "r+")
+            var = dataset.variables[self.varname]
             var[keys] = array_data
         finally:
-            dataset.close()
+            if dataset:
+                dataset.close()
+            lock.release()
 
     def __repr__(self):
-        fmt = (
-            "<{self.__class__.__name__} shape={self.shape}"
-            " dtype={self.dtype!r} path={self.path!r}"
-            " name={self.name!r}>"
-        )
-        return fmt.format(self=self)
-
-
-@dask.delayed
-def combined_delayeds(*args):
-    """A delayed function which simply computes all its arguments."""
-    # Dask computes the lazy args before passing them here.
-    # So job done -- we don't need to do anything with them.
-    pass
+        return f"<{self.__class__.__name__} path={self.path!r} var={self.varname!r}>"
 
 
 class Saver:
@@ -595,23 +591,25 @@ class Saver:
         self._formula_terms_cache = {}
         #: Whether lazy saving.
         self.lazy_saves = not compute
-        #: A list of deferred writes (if lazy saving)
+        #: A list of deferred writes for lazy saving : each is a (source, target) pair
         self.deferred_writes = []
         #: Target filepath
-        self.filepath = filename
+        self.filepath = os.path.abspath(filename)
+        #: Target lockfile path
+        self._lockfile_path = self.filepath + ".lock"
         #: NetCDF dataset
         self._dataset = None
         try:
             self._dataset = netCDF4.Dataset(
-                filename, mode="w", format=netcdf_format
+                self.filepath, mode="w", format=netcdf_format
             )
         except RuntimeError:
-            dir_name = os.path.dirname(filename)
+            dir_name = os.path.dirname(self.filepath)
             if not os.path.isdir(dir_name):
                 msg = "No such file or directory: {}".format(dir_name)
                 raise IOError(msg)
             if not os.access(dir_name, os.R_OK | os.W_OK):
-                msg = "Permission denied: {}".format(filename)
+                msg = "Permission denied: {}".format(self.filepath)
                 raise IOError(msg)
             else:
                 raise
@@ -2516,14 +2514,10 @@ class Saver:
                     # Create a data-writeable object that we can stream into, which
                     # encapsulates the file to be opened + variable to be written.
                     writeable_var_wrapper = DeferredSaveWrapper(
-                        self.filepath, cf_var
+                        self.filepath, cf_var, self._lockfile_path
                     )
-                    # Add a delayed save to our 'deferred_writes' list.
-                    self.deferred_writes.append(
-                        da.store(
-                            [data], [writeable_var_wrapper], compute=False
-                        )
-                    )
+                    # Add to the list of deferred writes, used in _deferred_save().
+                    self.deferred_writes.append((data, writeable_var_wrapper))
                     # NOTE: in this case, no checking of fill-value violations so just
                     # return dummy values for this.
                     # TODO: just for now -- can probably make this work later
@@ -2588,6 +2582,39 @@ class Saver:
                 "keyword during saving, otherwise use ncedit/equivalent."
             )
             warnings.warn(msg.format(cf_var.name, fill_value))
+
+    def _deferred_save(self):
+        """
+        Create a 'delayed' to trigger file completion for lazy saves.
+
+        This contains all the deferred writes, which complete the file by filling out
+        the data of variables initially created empty.
+
+        """
+        # Create a lock to satisfy the da.store call.
+        # We need a serialisable lock for scheduling with processes or distributed.
+        # See : https://github.com/dask/distributed/issues/780
+        # However, this does *not* imply safe access for file writing in parallel.
+        # For that, DeferredSaveWrapper uses a filelock as well.
+        lock = SerializableLock()
+
+        # Create a single delayed da.store operation to complete the file.
+        sources, targets = zip(*self.deferred_writes)
+        result = da.store(sources, targets, compute=False, lock=lock)
+
+        # Wrap that in an extra operation that follows it by deleting the lockfile.
+        @dask.delayed
+        def postsave_remove_lockfile(store_op, lock_path):
+            if os.path.exists(lock_path):
+                try:
+                    os.unlink(lock_path)
+                except Exception as e:
+                    msg = f'Could not remove lockfile "{lock_path}".  Error:\n{e}'
+                    raise Exception(msg)
+
+        result = postsave_remove_lockfile(result, self._lockfile_path)
+
+        return result
 
 
 def save(
@@ -2886,6 +2913,6 @@ def save(
         if compute:
             result = None
         else:
-            # For lazy save, return a single 'delayed' representing all lazy writes.
-            result = combined_delayeds(sman.deferred_writes)
+            result = sman._deferred_save()
+
         return result
--- a/lib/iris/fileformats/netcdf/saver.py
+++ b/lib/iris/fileformats/netcdf/saver.py
@@ -499,7 +499,7 @@ class DeferredSaveWrapper:
     TODO: could be improved with a caching scheme, but this just about works.
     """
 
-    def __init__(self, filepath: str, cf_var: netCDF4.Variable):
+    def __init__(self, filepath, cf_var):
         # Grab useful properties of the variable, including the identifying 'name'.
         self.path = filepath
         for key in ("shape", "dtype", "ndim", "name"):
--- a/lib/iris/fileformats/netcdf/saver.py
+++ b/lib/iris/fileformats/netcdf/saver.py
@@ -23,6 +23,7 @@ from typing import List
 import warnings
 
 import cf_units
+import dask
 import dask.array as da
 import netCDF4
 import numpy as np
@@ -490,10 +491,51 @@ class _FillValueMaskCheckAndStoreTarget:
 MESH_ELEMENTS = ("node", "edge", "face")
 
 
+class DeferredSaveWrapper:
+    """
+    An object which mimics the data access of a netCDF4.Variable, and can be written to.
+    It encapsulates the netcdf file and variable which are actually to be written to.
+    This opens the file each time, to enable writing the data chunk, then closes it.
+    TODO: could be improved with a caching scheme, but this just about works.
+    """
+
+    def __init__(self, filepath: str, cf_var: netCDF4.Variable):
+        # Grab useful properties of the variable, including the identifying 'name'.
+        self.path = filepath
+        for key in ("shape", "dtype", "ndim", "name"):
+            setattr(self, key, getattr(cf_var, key))
+
+    def __setitem__(self, keys, array_data):
+        # Write to the variable.
+        # Re-open the file for writing.
+        dataset = netCDF4.Dataset(self.path, "r+")
+        try:
+            var = dataset.variables[self.name]
+            var[keys] = array_data
+        finally:
+            dataset.close()
+
+    def __repr__(self):
+        fmt = (
+            "<{self.__class__.__name__} shape={self.shape}"
+            " dtype={self.dtype!r} path={self.path!r}"
+            " name={self.name!r}>"
+        )
+        return fmt.format(self=self)
+
+
+@dask.delayed
+def combined_delayeds(*args):
+    """A delayed function which simply computes all its arguments."""
+    # Dask computes the lazy args before passing them here.
+    # So job done -- we don't need to do anything with them.
+    pass
+
+
 class Saver:
     """A manager for saving netcdf files."""
 
-    def __init__(self, filename, netcdf_format):
+    def __init__(self, filename, netcdf_format, compute=True):
         """
         A manager for saving netcdf files.
 
@@ -505,6 +547,15 @@ class Saver:
         * netcdf_format (string):
             Underlying netCDF file format, one of 'NETCDF4', 'NETCDF4_CLASSIC',
             'NETCDF3_CLASSIC' or 'NETCDF3_64BIT'. Default is 'NETCDF4' format.
+
+        * compute (bool):
+            If True, the Saver performs normal 'synchronous' data writes, where data
+            is streamed directly into file variables during the save operation.
+            If False, the file is created as normal, but computation and streaming of
+            any lazy array content is instead deferred to :class:`dask.delayed` objects,
+            which are held in a list in the saver 'delayed_writes' property.
+            The relavant file variables are created empty, and the write can
+            subsequently be completed by computing the 'save.deferred_writes'.
 
         Returns:
             None.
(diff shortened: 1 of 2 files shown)


bjlittle:
@pp-mo We could also offer a similar pattern for parallel loading, akin to `xarray.open_mfdataset(parallel=True)` :thinking:

bjlittle:
@pp-mo Kicking the tyres with this using `distributed`... and running into issues

lib/iris/fileformats/netcdf/saver.py
```diff
@@ -490,10 +491,51 @@ def __setitem__(self, keys, arr):
 MESH_ELEMENTS = ("node", "edge", "face")
 
 
+class DeferredSaveWrapper:
+    """
+    An object which mimics the data access of a netCDF4.Variable, and can be written to.
+    It encapsulates the netcdf file and variable which are actually to be written to.
+    This opens the file each time, to enable writing the data chunk, then closes it.
+    TODO: could be improved with a caching scheme, but this just about works.
```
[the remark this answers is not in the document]
pp-mo: Dead right.  I had not understood much, which is why it did not work with 'distributed'.
Hopefully that is now fixed @ed654d  ...

pp-mo:
> @pp-mo Kicking the tyres with this using `distributed`... and running into issues

Updated version **_should_** now work with 'distributed', I think.
@bjlittle @bouweandela can you try this again + report back ?

My experience has been that the problems with 'distributed' mostly behaved just the same with a local 'processes' scheduler.
So that could be useful for testing.

This whole idea is still somewhat experimental --  work to do, not least add some kind of testing.
I would like some feedback on whether this approach will work in practice, before going further.

pp-mo:
### Status summary

My experience has been that the problems with 'distributed' mostly behaved just the same with a local 'processes' scheduler.
So that could be useful for testing.

This whole idea is still somewhat experimental --  work to do, not least add some kind of testing.
I would like some feedback on whether this approach will work in practice, before going further.

pp-mo:
> @pp-mo We could also offer a similar pattern for parallel loading, akin to `xarray.open_mfdataset(parallel=True)`, returning delayed futures thinking

BTW this is certainly possible, but it's not nearly so crucial to have specialist support IMHO, because ...
Since loading does _not_ involve dask "compute", loading operations can be parallelised simply by "@delayed" on a loading routine like "def get_data(filepath): ...".
This would certainly fit well into ["Add user advice on Dask "best practices"](https://github.com/SciTools/iris/issues/4959) though.  We already have an example like this in our internal "Dask Best Practices" documentation.

lib/iris/fileformats/netcdf/saver.py
```diff
@@ -506,6 +544,15 @@ def __init__(self, filename, netcdf_format):
             Underlying netCDF file format, one of 'NETCDF4', 'NETCDF4_CLASSIC',
             'NETCDF3_CLASSIC' or 'NETCDF3_64BIT'. Default is 'NETCDF4' format.
 
+        * compute (bool):
+            If True, the Saver performs normal 'synchronous' data writes, where data
+            is streamed directly into file variables during the save operation.
+            If False, the file is created as normal, but computation and streaming of
+            any lazy array content is instead deferred to :class:`dask.delayed` objects,
+            which are held in a list in the saver 'delayed_writes' property.
```
pp-mo, line 552: This bit can probably be treated as internal/private detail now. Maybe just make "._deferred_save" public.

lib/iris/fileformats/netcdf/saver.py
```diff
@@ -2509,6 +2583,39 @@ def store(data, cf_var, fill_value):
             )
             warnings.warn(msg.format(cf_var.name, fill_value))
 
+    def _deferred_save(self):
```
pp-mo, line 2586: Worth making public?
Almost certainly better named 'delayed' than 'deferred' (!)

bouweandela:
> @bjlittle @bouweandela can you try this again + report back ?

Thanks for all the good work @pp-mo! I tried this using a `dask_jobqueue.SLURMCluster` today and it seems to work now. Are the .lock files needed though? I don't see those appearing when doing the same thing using xarray's lazy save function.

pp-mo:
> @bouweandela  I tried this using a dask_jobqueue.SLURMCluster today and it seems to work now.

That is great news, thanks for testing !

----
> @bouweandela  Are the .lock files needed though? I don't see those appearing when doing the same thing using xarray's lazy save function.

Well, it's just **_a_** way of solving the problem.  And I think a simple one :+1:.
I wasn't quite sure how to implement a lock that was genuinely the _same_ for all workers in a cluster, so I knew that doing it this way with the file-system should be a fairly general approach.
I did try to work out what xarray is using, but there are a lot of unfamiliar code layers + it's slightly beyond me (!). 
I **_think_** it is probably using [_different_ approaches for different dask schedulers](https://github.com/pydata/xarray/blob/main/xarray/backends/locks.py#L39-L67), and in this case a `distributed.DistributedLock`, but I might still not be looking in quite the right place for this.

bouweandela:
The .lock files are indeed removing themselves, but I have some doubts about this approach:
- I suspect it can lead to race conditions if not all machines on the cluster have the same view of the (distributed) filesystem at the same time
- What happens if your script is killed halfway through and you try to rerun it with some of those lock files still around from the previous run? Will it hang indefinitely or do you need to clean them up manually before? I'm not sure this will give a good user experience

bouweandela:
> I did try to work out what xarray is using, but there are a lot of unfamiliar code layers + it's slightly beyond me (!).
I think it is probably using [different approaches for different dask schedulers](https://github.com/pydata/xarray/blob/main/xarray/backends/locks.py#L39-L67), and in this case a distributed.DistributedLock, but I might still not be looking in quite the right place for this.

I would be happy to have a look at the code myself and see if we can figure it out together. Would that help? I will probably have time for that next week.

pp-mo:
> The .lock files are indeed removing themselves, but I have some doubts about this approach:
> 
> * I suspect it can lead to race conditions if not all machines on the cluster have the same view of the (distributed) filesystem at the same time
> * What happens if your script is killed halfway through and you try to rerun it with some of those lock files still around from the previous run? Will it hang indefinitely and/or do you need to clean them up manually before? I'm not sure this will give a good user experience

Re: "if your script is killed halfway through"
In Linux terms, if triggered as a normal SIGINT, this ought to exit in a controlled fashion, removing locks via the "finally" actions.
I guess SIGKILL would defeat that, but that is usually reserved as an emergency action.

Re: race conditions
I'm not convinced this is a real problem :  I would hope that using a common file-system to control would be _the_ most reliable cross-platform way to achieve co-ordination between processes on a cluster -- that's the whole point of 'filelock'. 

I'm not saying there is not a better way, but I was more concerned about the performance/efficiency of the solution.
As far as I can see though, if we need dask workers in a cluster to share the **_same_** lock, the **_only_** easy way to do that --apart from the file-system-- is to make the locks used specific to the dask scheduler type, as xarray is doing.  

So from a simplicity + generality PoV, 'filelock' is still winning for me !
The alternative would have to be a non-dask solution to locks using message passing -- since distributed workers are in different memory spaces.
So far, I don't know what we'd use for that, or what other dependencies it might require.

pp-mo:
> An alternative could be, a non-dask solution to locks using message passing -- since distributed workers are in different memory spaces. So far, I don't know what we'd use for that, or what other dependencies it might require.

P.S. a little searching turns up ['sherlock'](https://github.com/py-sherlock/sherlock), which implements a generic multiprocessing-like interface to various solutions, including 'Redis' and 'filelock'.  Well, I have at least I've _heard_ of those ones.

To be specific : I don't really like to include distributed-specific code, but maybe there are benefits?

pp-mo:
> it looks like the code in this branch does not work if the data is not lazy. I'm not sure what it is supposed to do if the data is not lazy, but a clear error message would be nice. Do you have ideas on this?

I think if we are requested to do a deferred save, then we must do one even if the data **_is_** already in-memory.

By my reading, [these 2 lines](https://github.com/SciTools/iris/pull/5031/files#diff-f9ee001b7207df38cebdcaebf980dd17d7424e7f20bf2ef0140f4dcc4480eaf4R2510-R2511) are making the choice in the wrong order.
So, the saving choice _should_ be ..
```
if lazy-save enabled:
  # lazy save is *always* deferred
  store = "lazy-store"
elif data is lazy:
  store = "streamed-store"
else:  # non-lazy, and data is real
  store = "direct-store"
```
[lazy-store](https://github.com/SciTools/iris/pull/5031/files#diff-f9ee001b7207df38cebdcaebf980dd17d7424e7f20bf2ef0140f4dcc4480eaf4R2513-R2525)
[streamed-store](https://github.com/SciTools/iris/pull/5031/files#diff-f9ee001b7207df38cebdcaebf980dd17d7424e7f20bf2ef0140f4dcc4480eaf4R2529-R2536)
[direct-store](https://github.com/SciTools/iris/pull/5031/files#diff-f9ee001b7207df38cebdcaebf980dd17d7424e7f20bf2ef0140f4dcc4480eaf4R2540-R2544)

pp-mo:
WARNING: 
it looks like the solutions being [investigated to fix the netcdf1.6.1 thread safety problems](https://github.com/SciTools/iris/pull/5061) will also be affecting this code, [e.g. here](https://github.com/SciTools/iris/pull/5061/files#diff-f9ee001b7207df38cebdcaebf980dd17d7424e7f20bf2ef0140f4dcc4480eaf4R2539).

So, expect more changes to come here.

pp-mo:
> @bouweandela apologies for being slow to get back to this

P.S. please don't worry about the timeliness.
We are currently busy doing other things, and progress here on longterm dev issues is pretty slow !! 🐌

fnattino:
Hello @pp-mo, as @bouweandela mentioned I have been running some tests with the Iris lazy saving together with ESMValCore. Thanks for all the nice work, it indeed seems to work well with [ESMValCore@dask-distributed](https://github.com/ESMValGroup/ESMValCore/tree/dask-distributed). 

Just a couple of comments:
- The synchronisation of the parallel writing via `filelock` seems to rely on the target .lock file being locked at the file-system level. This feature might not been supported from all file-systems (for instance in Lustre this needs to be [specifically enabled](https://wiki.lustre.org/Mounting_a_Lustre_File_System_on_Client_Nodes) when mounting client nodes). Also considering that `threading` is the default Dask scheduler for Dask arrays, would it make sense to have the `threading.Lock` as default locking scheme here? 
- One could leave the possibility to pass to Iris a different type of lock to customize the lock on the basis e.g. of the scheduler employed. One could do:
  ```python
  lazy_save = iris.save(cube, 'tmp.nc', compute=False)
  lazy_save.compute()  # use the default multithreading scheduler and threading.Lock() 
  ```
  or:
  ```python
  from dask.distributed import Client, Lock
  with Client() as client:
    lazy_save = iris.save(cube, 'tmp.nc', compute=False, lock=Lock())
    lazy_save.compute()  # use the distributed scheduler and the user-provided lock
  ```
  This would be similar to how this is handled in [rioxarray](https://corteva.github.io/rioxarray/stable/examples/dask_read_write.html). ESMValCore could set up the distributed lock outside Iris, so Iris would remain distributed-free. 
- I think that the additional lock passed on to `dask.array.store` is not needed, since the synchronization of the parallel writing is already taken care of by the `DeferredSaveWrapper` (see comment). I think this is also how Xarray handles it, with the lock for parallel writing being set at the datastore level: https://github.com/pydata/xarray/blob/6e77f5e8942206b3e0ab08c3621ade1499d8235b/xarray/backends/netCDF4_.py#L68 
  while the `ArrayWriter`, which ultimately calls `dask.array.store`, does not have a lock set:  
https://github.com/pydata/xarray/blob/6e77f5e8942206b3e0ab08c3621ade1499d8235b/xarray/backends/common.py#L171
- I have bumped into the same issue as @bouweandela with real (not lazy) data. In addition to your fix in the if-statement, one would also need to make sure that the sources are Dask arrays when calling `dask.array.store` (see comment).

lib/iris/fileformats/netcdf/saver.py
```diff
@@ -2509,6 +2583,39 @@ def store(data, cf_var, fill_value):
...
+        This contains all the deferred writes, which complete the file by filling out
+        the data of variables initially created empty.
+
+        """
+        # Create a lock to satisfy the da.store call.
+        # We need a serialisable lock for scheduling with processes or distributed.
+        # See : https://github.com/dask/distributed/issues/780
+        # However, this does *not* imply safe access for file writing in parallel.
+        # For that, DeferredSaveWrapper uses a filelock as well.
+        lock = SerializableLock()
```
fnattino, line 2599: As you mention, the synchronisation of the parallel writing should be taken care of by the DeferredSaveWrapper, so one could set `lock=False` here.

lib/iris/fileformats/netcdf/saver.py
```diff
@@ -2509,6 +2583,39 @@ def store(data, cf_var, fill_value):
...
+
+        # Create a single delayed da.store operation to complete the file.
+        sources, targets = zip(*self.deferred_writes)
+        result = da.store(sources, targets, compute=False, lock=lock)
```
fnattino, line 2603: Some of the sources can be numpy array here, so one could do something like:
```suggestion
        result = da.store([da.asarray(s) for s in sources], targets, compute=False, lock=lock)
```

lib/iris/fileformats/netcdf/saver.py
pp-mo: Thanks for the hint.  The truth is, we are (I am) still learning here.
And also, it looks like the #2063 issue will affect the eventual implementation of this -- this code is definitely only a draft "Proof of Concept" solution.

pp-mo:
> 

> * The synchronisation of the parallel writing via `filelock` seems to rely on the target .lock file being locked at the file-system level.

IIUC,  "filelock" is simply a way of ensuring a mutex that is shared across possibly multiple systems or CPUS.  So it uses the common filesystem as a communication mechanism between them, and connects any that use that same file path.
Which is [exactly what the dask "distributed lock" does](https://github.com/dask/distributed/blob/main/distributed/lock.py#L82-L84).
So IIUC, filelock uses the files to support atomic operations e.g. check whether someone is assigned access /  aquire access for yourself.  And what this requires is simply that only one client can have a file open for writing.  Which I think will be enforced by the standard unix filesystem access.

pp-mo:
> * One could leave the possibility to pass to Iris a different type of lock to customize it on the basis e.g. of the scheduler employed. One could do:

Yes, I think this would be useful.

> * I think that the additional lock passed on to `dask.array.store` is not needed

Yes, I agree.  Still learning here !


> * I have bumped into the same issue as @bouweandela with real (not lazy) data. In addition to your fix in the if-statement, one would also need to make sure that the sources are Dask arrays when calling `dask.array.store` (see comment).

Yes also agreed !
We will also address this when we get to a "real proposed implementation".

lib/iris/fileformats/netcdf/saver.py
pp-mo: See [comment above](https://github.com/SciTools/iris/pull/5031#issuecomment-1322340895)
I think this is the same issue ?

lib/iris/fileformats/netcdf/saver.py
fnattino: Sure! Just thought I would write this down, since it also took me a while to figure out from Xarray..

lib/iris/fileformats/netcdf/saver.py
fnattino: Yes, it's also related to having real data. One needs to make sure these are converted to Dask arrays.

fnattino:
Just wanted to mention that while Xarray seems to support threading, multiprocessing and distributed as schedulers for writing NetCDFs, the multiprocessing implementation based on the `multiprocessing.Lock` seems not to work: https://github.com/pydata/xarray/issues/3781 (it's an old issue, but I still get the same error). Maybe just having the multithreading implementation (default) and distributed supported would already cover many (most?) use cases?

bouweandela:
> Maybe just having the multithreading implementation (default) and distributed supported would already cover many (most?) use cases?

That sounds good to me. If some really wanted to use multiple processes on a single machine, they could just use the distributed `LocalCluster` to set that up.

pp-mo:
> > @fnattino Maybe just having the multithreading implementation (default) and distributed supported would already cover many (most?) use cases?
> 
> @bouweandela That sounds good to me. If someone really wanted to use multiple processes on a single machine, they could just use the distributed `LocalCluster` to set that up.

Just FYI I have looked into this within work for #5191, and it is not that simple.
I could find **_no_** good way of fixing it using Dask features :  It seems not have an equivalent of the distributed.Lock for a local-processes scheduler.  I did I try using filelock for this, but the worker processes hang for reasons I that weren't clear (it _might_ be worth just trying that again when code is more mature).
According to https://stackoverflow.com/a/25558333/2615050, it may be possible to use a [multiprocessing.Manager](https://docs.python.org/3/library/multiprocessing.html#managers) for this.  I think that can also work for processes that were already in existence when the Manager was created, but still a little unsure.

So I'm still considering this an "optional extra" for purposes of #5191

pp-mo:
Closing this now, in favour of #5191 
The other outstanding issues from here I have copied forward to there.
Which were (I think):
  * **support for local-process scheduler**  See [above comment](https://github.com/SciTools/iris/pull/5031#issuecomment-1470487423) : shelved for now, on list of things-to-consider
  * **problems with saving real data** (see [@fnattino  comment](https://github.com/SciTools/iris/pull/5031#issuecomment-1344501872)) :  Already fixed in new implementation, I believe 
  * **option for user to pass in a lock** (also [@fnattino  comment](https://github.com/SciTools/iris/pull/5031#issuecomment-1344501872)) : forwarded to the checklist of work-to-do

PR #5036, opened 2022-10-26
Use reusable workflow for refresh lockfile workflow

lbdreyer:
As there is some duplication in the use of workflows across our repos, and this is likely to increase, I have been looking at implementing [reusable workflows](https://docs.github.com/en/actions/using-workflows/reusing-workflows).  

I propose we have a central repo that will contain the workflows that we can reuse across iris, iris-grib, cf-units, and more if need be.

As a first step, I have chosen to look at the simplest workflow which is the one we use to refresh lockfiles. 

This PR modifies the workflow so it instead calls a reusable workflow which is stored in a `scitools/reusable_workflows` repo which I will create soon!

### DIFF
--- a/.github/workflows/refresh-lockfiles.yml
+++ b/.github/workflows/refresh-lockfiles.yml
@@ -1,13 +1,5 @@
-# This workflow periodically creates new environment lock files based on the newest
-# available packages and dependencies.
-#
-# Environment specifications are given as conda environment.yml files found in
-# `requirements/ci/py**.yml`.  These state the packages required, the conda channels
-# that the packages will be pulled from, and any versions of packages that need to be
-# pinned at specific versions.
-#
-# For environments that have changed, a pull request will be made and submitted
-# to the main branch
+# Updates the environment lock files. See the called workflow in the
+# scitools/reusable_workflows repo for more details.
 
 name: Refresh Lockfiles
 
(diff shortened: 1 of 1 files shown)


.github/workflows/refresh-lockfiles.yml
```diff
@@ -20,98 +12,7 @@ on:
...
-
-      - name: Check Pull Request
-        if: steps.cpr.outputs.pull-request-number != ''
-        run: |
-          echo "pull-request #${{ steps.cpr.outputs.pull-request-number }}"
-          echo "pull-request URL ${{ steps.cpr.outputs.pull-request-url }}"
-          echo "pull-request operation [${{ steps.cpr.outputs.pull-request-operation }}]"
-          echo "pull-request head SHA ${{ steps.cpr.outputs.pull-request-head-sha }}"
+  refresh_lockfiles:
+    uses: scitools/reusable_workflows/.github/workflows/refresh-lockfiles.yml@main
```
lbdreyer, line 17: I use the `main` reference here temporarily. We will likely want to use either a commit SHA or version when we actually come to implement this

.github/workflows/refresh-lockfiles.yml
```diff
@@ -20,98 +12,7 @@ on:
...
+    secrets: inherit
```
lbdreyer, line 18: This is so that it can inherit the secrets we use to [generate the token](https://github.com/SciTools/iris/blob/d56824cbc892356302c135ee23d704ff01e12d71/.github/workflows/refresh-lockfiles.yml#L85-L90)) from the calling repository (e.g. Iris)

.github/workflows/refresh-lockfiles.yml
bjlittle: Yeah, agreed. For maintenance and continuity it would be best practice to adopt specific SHAs or TAGs :+1:

.github/workflows/refresh-lockfiles.yml
```diff
@@ -20,98 +12,7 @@ on:
...
-
-      - name: Check Pull Request
-        if: steps.cpr.outputs.pull-request-number != ''
-        run: |
-          echo "pull-request #${{ steps.cpr.outputs.pull-request-number }}"
-          echo "pull-request URL ${{ steps.cpr.outputs.pull-request-url }}"
-          echo "pull-request operation [${{ steps.cpr.outputs.pull-request-operation }}]"
-          echo "pull-request head SHA ${{ steps.cpr.outputs.pull-request-head-sha }}"
+  refresh_lockfiles:
+    uses: scitools/workflows/.github/workflows/refresh-lockfiles.yml@main
```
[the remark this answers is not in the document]
bjlittle: Agreed :smile:

bjlittle (approved):

Issue #4983, opened 2022-09-23
split-attrs: cube attributes


pp-mo:
**Note:** #5040 now to implement this

----
"cube.attributes" to become a wrapper object, 
providing read+write accesses ~same as existing (combined local+global), _plus_ access to local + global settings specifically, when required. 

At this point, we expect all existing tests to pass **_unmodified_**
  - maybe just a few odd exceptions ?

(N.B. have come code for this prototyped already)

pp-mo:
Done #5040

Issue #4982, opened 2022-09-23
split-attrs: cube metadata


pp-mo:
Cube metadata will contain separate 'local' + 'global' attribute dicts.
Effectively, the 'local' is the existing dict, named "metadata.attributes', while the 'metadata.global_attributes' is a new one.

Nothing special links the two (at least initially).
There may be future scope for additional behaviour that understands the connection between the two (e.g. a lenient type service)

Related task #4983 will specialise access to "cube.attributes" to enable a unified access to both dicts, or to each specifically.

pp-mo:
Update: when attempting this #5038, hit a number of problems...

Basically, existing code had a number of assumed contracts with the relationship between a CFVariableMixin-derived object and its common-metadata class.
Notably:
  1. assumption that each CubeMetadata field --> corresponding kwarg of `Cube.__init__`
      * e.g. [here in merge](https://github.com/SciTools/iris/blob/15bbcc5ac3d539cb6e820148b66e7cf55d91c5d2/lib/iris/_merge.py#L1596-L1612)
      * and [here in experimental/stratify](https://github.com/SciTools/iris/blob/15bbcc5ac3d539cb6e820148b66e7cf55d91c5d2/lib/iris/experimental/stratify.py#L175)
  2. assumption that each CubeMetadata field -->  corresponding (computed) named property of `Cube`
      * [here in CFVariableMixin](https://github.com/SciTools/iris/blob/15bbcc5ac3d539cb6e820148b66e7cf55d91c5d2/lib/iris/common/mixin.py#L237-L245)

This implied sufficiently many knock-on changes that I abandoned the approach.
So this task closes, and #5040 replaces #5038

PR #5038, opened 2022-10-27
Splitattrs cube metadata

pp-mo:
Addresses #4982 

This implements the basics of adding a separate `.global_attributes` field to a `CubeMetadata`,
and tests that it behaves exactly the same as the `.attributes` field.

NB this field is not yet accessible from the cube itself, so no behaviour changes should result.
That awaits #4983, which will replace the implemenation of `cube.attributes` and add `cube.attributes.local` and `cube.attributes.global`.


pp-mo:
Closing in favour of #5040

PR #5040, opened 2022-10-31
Implement split cube attributes.

pp-mo:
Initial code implementation
This is an alternative to the #5038 approach, which has hit some problems (a lot of consequential knock-ons).
See [closing comment of #4982](https://github.com/SciTools/iris/issues/4982#issuecomment-1302823564) for details

This introduces a new "split dictionary" type, `CubeAttrsDict`, and ensures that a `Cube.attributes` is always one of those.
The new type has special behaviours to make it inter-compatible with a 'normal' dictionary -- or more precisely, a `iris.common.mixin.LimitedAttributeDict` which is what all other `.attributes` properties are.

This means that now, a `Cube.attributes` is a rather different kind of a thing from a `Coord.attributes`.
We hope this makes few practical differences, compared to a LimitedAttributeDict, but there are a few, such as :
  * the object type and class inheritance has changed 
     -- it is still a `MutableMapping`, but no longer a subtype of `dict`.
  * ordering of keys is affected
  * 'repr' form has changed
     I thought it was no longer appropriate to make it "look like" a standard dict, as LimitedAttributesDict does, since that style can no longer fully represent the content state.  Unfortunately that complicates some of the existing docs-examples, so I've tried to adjust those to cope.


pp-mo:
Ping @lbdreyer 
I reckon this is now ready to go live.

Since I changed tack (from #5038), there are no longer separate changes to cube metadata #4982 and cube attributes handling #4983 :  This is effectively doing both, and no change to CubeMetadata is required after all.

In itself, this changes **_almost_** nothing in behaviour, since all the existing code (including netcdf load+save) reads+writes the new split-type attributes dictionary as if it were an ordinary dictionary.  So, real behaviour change can only occur when we change the load+save code.
However, there _are_ some odd corner-cases which will not behave quite as before.  Most notably, the ordering of attributes (keys) may be changed -- though NB this does **_not_** affect dictionary comparison/equality.  And of course, any type-dependent usages may fail.

docs/src/further_topics/metadata.rst
```diff
@@ -131,7 +131,7 @@ We can easily get all of the associated metadata of the :class:`~iris.cube.Cube`
 using the ``metadata`` property:
 
     >>> cube.metadata
-    CubeMetadata(standard_name='air_temperature', long_name=None, var_name='air_temperature', units=Unit('K'), attributes={'Conventions': 'CF-1.5', 'STASH': STASH(model=1, section=3, item=236), 'Model scenario': 'A1B', 'source': 'Data from Met Office Unified Model 6.05'}, cell_methods=(CellMethod(method='mean', coord_names=('time',), intervals=('6 hour',), comments=()),))
+    CubeMetadata(standard_name='air_temperature', long_name=None, var_name='air_temperature', units=Unit('K'), attributes=CubeAttrsDict(globals={}, locals={'Conventions': 'CF-1.5', 'STASH': STASH(model=1, section=3, item=236), 'Model scenario': 'A1B', 'source': 'Data from Met Office Unified Model 6.05'}), cell_methods=(CellMethod(method='mean', coord_names=('time',), intervals=('6 hour',), comments=()),))
```
pp-mo, line 134: I think this is showing a problem, also highlighted by the trial code in #5048 :
"Conventions" should really be treated as a global, but our classification currently only recognises "conventions".
I don't quite understand where this confusion originally came from, but that is how it is in the existing code.

lib/iris/tests/unit/cube/test_CubeAttrsDict.py
```diff
@@ -0,0 +1,323 @@
...
+        # .. and check key order of combined attrs
+        assert list(attrs.keys()) == ["a", "m", "f", "z"]
+        # create the "same" thing with locals before globals, *and* different key order
+        attrs = CubeAttrsDict()
+        attrs.locals.update(dict(z=4, f=7))
+        attrs.globals.update(dict(m=3, a=1))
+        # .. this shows that the result is not affected either by alphabetical key
+        # order, or the order of adding locals/globals
+        # I.E. result is globals-in-create-order, then locals-in-create-order
+        assert list(attrs.keys()) == ["m", "a", "z", "f"]
```
lbdreyer, line 231: Could you split this into two tests? I would rather we keep unit tests to each testing a single thing. Otherwise you run into the issue where you won't know whether the second test is failing if the first test fails.

lib/iris/tests/unit/cube/test_CubeAttrsDict.py
```diff
@@ -0,0 +1,323 @@
...
+        attrs = CubeAttrsDict(locals={global_attrname: 1})
+        # The attr *remains* local, even though it would be created global by default
+        attrs[global_attrname] = 2
+        check_content(attrs, locals={global_attrname: 2})
+
+    def test_overwrite_forced_global(self):
+        attrs = CubeAttrsDict(globals={"data": 1})
+        # The attr remains local, even though it would be created local by default
+        attrs["data"] = 2
+        check_content(attrs, globals={"data": 2})
```
lbdreyer, line 270: This seems like a repeat of `test_overwrite_global`?

lib/iris/cube.py
```diff
@@ -758,6 +767,231 @@ def _is_single_item(testee):
...
 
 
+class CubeAttrsDict(MutableMapping):
+    """
+    A dict-like object for "Cube.attributes", which provides unified user access to
+    the combined cube 'local' and 'global' attributes, mimicking the behaviour of a
+    simple dict.
+
+    This supports all the regular methods of a 'dict',
+    However, a few things such as the detailed print (repr) are different.
```
lbdreyer, line 777:
```suggestion
    however, a few things such as the detailed print (repr) are different.
```
You end the above line with a comma

lib/iris/cube.py
```diff
@@ -758,6 +767,231 @@ def _is_single_item(testee):
...
+    both :class:`~iris.common.mixin.LimitedAttributeDict`.  The CubeAttrsDict object
+    contains *no* additional state of its own, but simply acts as a view on these two.
+
+    When reading (__getitem__, pop. popitem, keys, values etc), it contains all the
+    keys + values of both 'locals' and 'globals'.  When a key occurs in *both* 'locals'
+    and 'globals', the result is the local value.
+
+    When writing (__setitem__, setdefault, update, etc) to a key already present, the
+    existing entry in either 'locals' or 'globals' is updated.  If both are present,
+    'locals' is updated.
```
lbdreyer, line 797: Why did you choose not to update both?
I just wonder whether this may catch some users out.e

lbdreyer (changes requested):
This is looking good so far! Most of my comments are minor, although the two main points I am unsure about are the behaviour of __setitem__ and whether this needs to part of the public API

lib/iris/cube.py
lbdreyer: It may be worth drawing more attention to this in the docstring. Currently, you mention it https://github.com/SciTools/iris/blob/7db42ce754e1d6d8b22f69aed2baa3eb4406c250/lib/iris/cube.py#L795-L797 but I wonder if including example, e.g. 
```
>>> a = CubeAttrsDict(locals={"z": "this"}, globals={"z": "that"})
>>> a['z'] = 'other'
>>> a
CubeAttrsDict(globals={'z': 'that'}, locals={'z': 'other'})
>>> a.globals['z'] = 'else'
>>> a
CubeAttrsDict(globals={'z': 'else'}, locals={'z': 'other'})
```
or even just highlighting this sentence may help?

lib/iris/cube.py
```diff
@@ -758,6 +767,231 @@ def _is_single_item(testee):
...
+        >>> print(cube.attributes)
+        {'history': 'fresh', 'x': 3}
+        >>> print(repr(cube.attributes))
+        CubeAttrsDict(globals={'history': 'fresh'}, locals={'x': 3})
+
+        >>> cube.attributes['history'] += '+added'
+        >>> print(repr(cube.attributes))
+        CubeAttrsDict(globals={'history': 'fresh+added'}, locals={'x': 3})
+
+        >>> cube.attributes.locals['history'] = 'per-var'
```
lbdreyer, line 821: Could you make these attribute values more similar to what you would encounter in real life? I think including real-world examples in the docs strings are more helpful for the user to understand what is going on

pp-mo:
Coming up : I'm intending now to repsond to all the above comments.
But I'll also re-base this onto the latest feature-branch version, which will probably introduce some errors/changes in the 'status quo' tests ...

lib/iris/cube.py
```diff
@@ -758,6 +767,231 @@ def _is_single_item(testee):
     return isinstance(testee, str) or not isinstance(testee, Iterable)
 
 
+class CubeAttrsDict(MutableMapping):
```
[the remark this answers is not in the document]
pp-mo: I think so, yes, because pretending that attributes are a simple dictionary only works 'up to a point', and for user-defined attributes we would intend future user code to create attributes of a specific 'type'.
I have also chosen to make it [visible in the object reprs](https://github.com/SciTools/iris/blob/7db42ce754e1d6d8b22f69aed2baa3eb4406c250/lib/iris/cube.py#L814-L815) : we now see `CubeAttrsDict`as a class name.  
Please say, if you think we should instead hide this ...

lib/iris/cube.py
```diff
@@ -758,6 +767,231 @@ def _is_single_item(testee):
...
+        # (the local one having priority).
+        if key in self.locals:
+            store = self.locals
+        elif key in self.globals:
+            store = self.globals
+        else:
+            # If NO existing attribute,  create local unless it is a "known global" one.
+            from iris.fileformats.netcdf.saver import _CF_GLOBAL_ATTRS
+
+            # Not in existing
```
pp-mo, line 965: This comment is unhelpful (!)
(probably left over from an earlier version)

trexfeathers (changes requested):
My review is in progress, still need to look at `test_CubeAttrsDict.py`. Here are some comments in the meantime 😊

docs/src/further_topics/metadata.rst
pp-mo: :+1:
https://github.com/SciTools/iris/issues/5379

Ok ?

trexfeathers (changes requested):
Here's the rest of my review

lib/iris/cube.py
```diff
@@ -758,6 +767,231 @@ def _is_single_item(testee):
...
+    def __iter__(self):
+        # Ordering: all global keys, then all local ones, but omitting duplicates.
+        # NOTE: this means that in the "summary" view, attributes present in both
+        # locals+globals are listed first, amongst the globals, even though they appear
+        # with the *value* from locals.
+        # Otherwise follows order of insertion, as is normal for dicts.
+        return itertools.chain(
+            self.globals.keys(),
+            (x for x in self.locals.keys() if x not in self.globals),
+        )
```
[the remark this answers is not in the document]
pp-mo: I don't think this is a good replacement : it produces the correct set of attribute names, but not in the desired _order_.  
I think it's important that the order is that of the globals, followed by locals-where-different, as stated.  Neither "locals | globals" or "globals | locals" reproduces that correctly.

I also think we don't need a special 'combined' method, since `dict(self)` provides exactly that, based on the iteration order as defined here.  That is actually used in the `__str__` function, but by contrast `len` is based directly on `iter`, since using a combined dictionary there would be making a copy every time.

lib/iris/cube.py
```diff
@@ -758,6 +767,231 @@ def _is_single_item(testee):
...
+        return len(list(iter(self)))
+
+    def __getitem__(self, key):
+        # Fetch an item from the "combined attributes".
+        # If both present, the local setting takes priority.
+        if key in self.locals:
+            store = self.locals
+        else:
+            store = self.globals
+        return store[key]
```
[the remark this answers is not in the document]
pp-mo: As stated above, I think overuse of the 'combine' operation would be making copies too much.
This is designed so we don't do that, but just fetch from the appropriate sub-group.

lib/iris/cube.py
```diff
@@ -758,6 +767,231 @@ def _is_single_item(testee):
...
+        CubeAttrsDict(globals={'history': 'fresh+added'}, locals={'x': 3, 'history': 'per-var'})
+
+    """
+
+    def __init__(
+        self,
+        combined: Optional[Mapping] = "__unset",
+        locals: Optional[Mapping] = None,
+        globals: Optional[Mapping] = None,
+    ):
```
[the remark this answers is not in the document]
pp-mo: Ok, actioning,
But this will also look better after another mergeback-from-main, since I think  recent style changes will present class/init docstrings more clearly.

lib/iris/cube.py
```diff
@@ -758,6 +767,231 @@ def _is_single_item(testee):
...
+    # Provide a serialisation interface
+    #
+    def __getstate__(self):
+        return (self.locals, self.globals)
+
+    def __setstate__(self, state):
+        self.locals, self.globals = state
+
+    #
+    # Support simple comparison, even when contents are arrays.
```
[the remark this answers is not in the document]
pp-mo: Only because it forces both 'locals' and 'globals'  to be LimitedAttributeDict, and _that_ implements a comparison which is array-aware.

lib/iris/tests/unit/cube/test_CubeAttrsDict.py
```diff
@@ -0,0 +1,323 @@
...
+        check_content(attrs, locals={"q": 3}, globals={"history": "something"})
+
+    def test_from_separate_dicts(self):
+        locals = {"q": 3}
+        globals = {"history": "something"}
+        attrs = CubeAttrsDict(locals=locals, globals=globals)
+        check_content(attrs, locals=locals, globals=globals)
+
+    def test_from_cubeattrsdict(self, sample_attrs):
+        result = CubeAttrsDict(sample_attrs)
```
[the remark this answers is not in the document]
pp-mo: We are checking that creating a CAD _from_ a CAD actually _works_ for making a copy, and results in a matching  object  (but a distinct object -- see `check_content`).  
Actually, `matches=sample_attrs` is a better way provided for that -- will do that.

lib/iris/cube.py
```diff
@@ -758,6 +767,231 @@ def _is_single_item(testee):
     return isinstance(testee, str) or not isinstance(testee, Iterable)
 
 
+class CubeAttrsDict(MutableMapping):
+    """
+    A dict-like object for "Cube.attributes", which provides unified user access to
```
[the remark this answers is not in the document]
pp-mo: I have tried to make a few improvements, but now I can't build the docs either locally or in the PR. 
 I'm also broken locally installing the required version of isort for pre-commit.  

Unfortunately, I think we now **_need_** a mergeback to cure some of these problems.
But I will avoid doing that until we're ~happy with all the existing comments + changes here.

lib/iris/tests/unit/cube/test_CubeAttrsDict.py
pp-mo: > a repeat of `test_overwrite_global`

Hmm. Yes I have to agree, because that is _also_ using a "forced" global rather than a "natural" one.
Rather than remove this test, I instead changed `test_overwrite_global` that one to test all recognised global names.
[a1248e859](https://github.com/SciTools/iris/pull/5040/commits/53b05d3b53d5b1644034f6746bd25ec724019ae3)

lib/iris/cube.py
pp-mo: Rather, replaced the comma with period.

lib/iris/cube.py
```diff
@@ -758,6 +767,231 @@ def _is_single_item(testee):
...
+                self.globals.update(deepcopy(combined.globals))
+                self.locals.update(deepcopy(combined.locals))
+            else:
+                # Treat any arbitrary single input value as a mapping (dict), and
+                # update from it.
+                self.update(dict(deepcopy(combined)))
+
+    #
+    # Ensure that the stored local/global dictionaries are "LimitedAttributeDicts".
+    #
```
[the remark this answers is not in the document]
pp-mo: No I made it up.
I don't think there is any 'official' way to group methods into functional categories, as I'm trying to do here.
I just thought it was convenient for clarity.

lib/iris/cube.py
```diff
@@ -758,6 +767,267 @@ def _is_single_item(testee):
...
+            values to init both 'self.globals' and 'self.locals'.  If 'combined' itself
+            has attributes named 'locals' and 'globals', these are used to update the
+            respective content (after initially setting the individual ones).
+            Otherwise, 'combined' is treated as a generic mapping, applied as
+            ``self.update(combined)``,
+            i.e. it will set locals and/or globals with the same logic as '__setitem__'.
+        locals : dict
+            initial content for 'self.locals'
+        globals : dict
+            initial content for 'self.locals'
```
trexfeathers, line 857:
```suggestion
            initial content for 'self.globals'
```

lib/iris/cube.py
```diff
@@ -758,6 +767,267 @@ def _is_single_item(testee):
...
+        --------
+        >>> CubeAttrsDict({'history': 'data-story', 'comment': 'this-cube'})
+        CubeAttrsDict(globals={'history': 'data-story'}, locals={'comment': 'this-cube'})
+        >>> CubeAttrsDict({'history': 'data-story', 'comment': 'this cube'}).globals
+        {'history': 'data-story'}
+        >>> CubeAttrsDict(locals={'history':'local'})
+        CubeAttrsDict(globals={}, locals={'history': 'local'})
+        >>> CubeAttrsDict(globals={'x': 'global'}, locals={'x':'local'})
+        CubeAttrsDict(globals={'x': 'global'}, locals={'x': 'local'})
+        >>>
```
trexfeathers, line 869:
```suggestion
```

lib/iris/cube.py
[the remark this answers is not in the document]
trexfeathers: >But, this will also look better after another mergeback-from-main

Agreed. I'll resolve this and look again once we have the latest docs build.

lib/iris/cube.py
```diff
@@ -758,6 +767,267 @@ def _is_single_item(testee):
...
+    A dict-like object for "Cube.attributes", which provides unified user access to
+    the combined cube 'local' and 'global' attributes, mimicking the behaviour of a
+    simple dict.
+
+    This supports all the regular methods of a 'dict'.
+    However, a few things such as the detailed print (repr) are different.
+
+    In addition, the 'locals' and 'globals' properties provide specific access to local
+    and global attributes, as regular
+    :class:`~iris.common.mixin.LimitedAttributeDict`\\s.
```
trexfeathers, line 781:
```suggestion
    :class:`~iris.common.mixin.LimitedAttributeDict`\\ s.
```

lib/iris/cube.py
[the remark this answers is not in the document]
trexfeathers: OK understood. If you're going to mention this in the comment then I think it is worth elaborating, too.

lib/iris/tests/unit/cube/test_CubeAttrsDict.py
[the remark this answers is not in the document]
trexfeathers: Ah, I should have read the test name more thoroughly!

lib/iris/cube.py
pp-mo: I'm struggling with this suggestion, as I think it would over-emphasize the really _rare_ case when an attribute is found in both locals + globals.  
To me, it's much more natural to start by saying that any _existing_ attribute will be updated.  So, saying "locals is the default" is really unhelpful + will give the wrong impression, in my view.

lib/iris/cube.py
pp-mo: For the **local** test attribute, I'm not aware of a "common" usage.  I changed it to "mycode", but I don't really know how helpful that is.  
For the **global** test attribute, I can only one of the defined names conventions / featureType / history / title, so perhaps 'title' might be better ?  Not sure though.  Can do, if you think that is clearer.

lib/iris/cube.py
pp-mo: Fixed https://github.com/SciTools/iris/commit/9bbaf5559db77e271fad6c817e3129877598e8fe

lib/iris/cube.py
pp-mo: Fixed https://github.com/SciTools/iris/commit/f143de066f538c447779e9dc9ff719d59867bb49

lib/iris/cube.py
pp-mo: Fixed https://github.com/SciTools/iris/commit/ade712e7f2993c5660866e3ddf20b76873eff70c

trexfeathers (changes requested):
Thanks for getting the docs rendering nicely! I have therefore been able to finish reviewing the docstrings.

lib/iris/cube.py
trexfeathers: I have no idea how to improve this so I'm just going to merge

lib/iris/cube.py
```diff
@@ -790,6 +800,285 @@ def _is_single_item(testee):
...
+    contains *no* additional state of its own, but simply acts as a view on these two.
+
+    When reading (__getitem__, pop, popitem, keys, values etc), it contains all the
+    keys + values of both 'locals' and 'globals'.  When a key occurs in *both* 'locals'
+    and 'globals', the result is the local value.
+    See :meth:``~iris.cube.CubeAttrsDict__getitem__``.
+
+    When writing (__setitem__, setdefault, update, etc) to a key already present, the
+    existing entry in either 'locals' or 'globals' is updated.  If both are present,
+    'locals' is updated.  See :meth:``~iris.cube.CubeAttrsDict__setitem__``.
```
trexfeathers, line 832:
```suggestion
    See :meth:`~iris.cube.CubeAttrsDict.__getitem__`.

    When writing (__setitem__, setdefault, update, etc) to a key already present, the
    existing entry in either 'locals' or 'globals' is updated.  If both are present,
    'locals' is updated.  See :meth:`~iris.cube.CubeAttrsDict.__setitem__`.
```

This may be pointless though as I don't think these methods get rendered.

lib/iris/cube.py
```diff
@@ -790,6 +800,285 @@ def _is_single_item(testee):
...
+        return len(list(iter(self)))
+
+    def __getitem__(self, key):
+        """
+        Fetch an item from the "combined attributes".
+
+        If the name is present in *both* ``self.locals`` and ``self.globals``, then
+        the local value is returned.
+
+        """
```
trexfeathers, line 1021: I think this docstring and `__setitem__` are an opportunity to shrink the class and `__init__` docstrings to give this block more structure and make it less intimidating.

I might not make such a fuss but `iris.cube` is one of the most visited API docs pages.

lib/iris/cube.py
```diff
@@ -758,6 +767,231 @@ def _is_single_item(testee):
...
+        Return a copy.
+
+        Implemented with deep copying, consistent with general Iris usage.
+
+        """
+        return CubeAttrsDict(self)
+
+    #
+    # The remaining methods are sufficient to generate a complete standard Mapping
+    # API -- see docs for :class:`collections.abc.MutableMapping`.
```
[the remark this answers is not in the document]
pp-mo: Fixed https://github.com/SciTools/iris/commit/8b070d6fc0d28f2355934398481107b31e9b0796

lib/iris/cube.py
```diff
@@ -790,6 +800,270 @@ def _is_single_item(testee):
...
+    A dict-like object for "Cube.attributes", which provides unified user access to
+    the combined cube 'local' and 'global' attributes, mimicking the behaviour of a
+    simple dict.
+
+    This supports all the regular methods of a 'dict'.
+    However, a few things such as the detailed print (repr) are different.
+
+    In addition, the 'locals' and 'globals' properties provide specific access to local
+    and global attributes, as regular
+    :class:`~iris.common.mixin.LimitedAttributeDict`\\s.
```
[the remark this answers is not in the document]
pp-mo: Ok, I decided it was worth adding it.
https://github.com/SciTools/iris/pull/5040/commits/16594a28947dc25ae80bb5f4d0ebc5c61db71daf

lib/iris/cube.py
```diff
@@ -790,6 +800,285 @@ def _is_single_item(testee):
     return isinstance(testee, str) or not isinstance(testee, Iterable)
 
 
+class CubeAttrsDict(MutableMapping):
+    """
+    A dict-like object for "Cube.attributes", which provides unified user access to
+    the combined cube 'local' and 'global' attributes, mimicking the behaviour of a
+    simple dict.
+
+    This supports all the regular methods of a 'dict'.
```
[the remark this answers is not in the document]
trexfeathers: FYI I finally worked out the intersphinx syntax for Python builtins!

```suggestion
    This supports all the regular methods of a :class:`dict`.
```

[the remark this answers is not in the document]
pp-mo: OK let's try that !

lib/iris/cube.py
```diff
@@ -790,6 +800,285 @@ def _is_single_item(testee):
...
+
+    """
+
+    def __init__(
+        self,
+        combined: Optional[Union[Mapping, str]] = "__unspecified",
+        locals: Optional[Mapping] = None,
+        globals: Optional[Mapping] = None,
+    ):
+        """
```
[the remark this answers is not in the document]
pp-mo: OK I've worked out that I can add docstrings for `__setattr__` and `__getattr__`, and it will show them (which IIRC used to not work without special handling).
So I think I can simplify via that, and also make use of how the class-docs and init-docs look together to shorten + cross-link the info.
Watch this space ...

lib/iris/cube.py
pp-mo: As for the previous - See [above](https://github.com/SciTools/iris/pull/5040#discussion_r1269183952)

pp-mo:
Ok, I've tried to make the docs a bit shorter, and hopefully tidier.
N.B. also merged from truck again to fix the linkcheck breakage.
@trexfeathers  please re-review !

trexfeathers:
Love the updated documentation @pp-mo  💯

Three unresolved conversations that are hidden:

- https://github.com/SciTools/iris/pull/5040#discussion_r1265344230
- https://github.com/SciTools/iris/pull/5040#discussion_r1268028827
- https://github.com/SciTools/iris/pull/5040#discussion_r1265349015

pp-mo:
> Love the updated documentation @pp-mo 💯
> 
> Three unresolved conversations that are hidden:
> 
> * [Implement split cube attributes. #5040 (comment)](https://github.com/SciTools/iris/pull/5040#discussion_r1265344230)
> * [Implement split cube attributes. #5040 (comment)](https://github.com/SciTools/iris/pull/5040#discussion_r1268028827)
> * [Implement split cube attributes. #5040 (comment)](https://github.com/SciTools/iris/pull/5040#discussion_r1265349015)

Oops, yes.
Coming shortly ..

lib/iris/cube.py
```diff
@@ -758,6 +767,231 @@ def _is_single_item(testee):
...
+        self.locals, self.globals = state
+
+    #
+    # Support simple comparison, even when contents are arrays.
+    #
+    def __eq__(self, other):
+        # For equality, require both globals + locals to match
+        other = CubeAttrsDict(other)
+        result = self.locals == other.locals and self.globals == other.globals
+        return result
```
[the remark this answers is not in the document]
pp-mo: Having tried it, no.
It seems the default code for mappings is probably doing something more like 
`all(self[k] == other[k] for k in self)`.  
hence we get `CAD(globals={'a':1}, locals={'a':2}) == CAD(locals={'a':2})`
Which is not good enough in our case.

lib/iris/cube.py
```diff
@@ -758,6 +767,231 @@ def _is_single_item(testee):
...
+
+    def __ne__(self, other):
+        return not self == other
```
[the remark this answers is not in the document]
pp-mo: :+1: yes, good spot.
This one is redundant !

lib/iris/cube.py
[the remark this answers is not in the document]
pp-mo: Ok, I updated the comments a bit.  Does not appear in docs, but it's rather a detail point, so I think that's ok.

pp-mo:
Thanks @trexfeathers hope that fixes the outstanding.

PR #5048, opened 2022-11-04
Splitattrs ncload

pp-mo:
**Update: Monday 3rd July**
still waiting on #5040, but looking generally OK and "close"
Basically, what truly belongs here is the last 2 commits **_only_**
 - note that the actual changes made in the main ["Distinguish local+global attributes in netcdf loads." commit](https://github.com/SciTools/iris/pull/5048/commits/9907ea73f1237daa75b0466259c8e5d25205f1b6) seem really tiny, 
   but that's just because a handy API is already implemented in #5040 : it only remains to assign global and local attrs to the cube in "slightly different" ways.

----
Addresses #4984 
WIP for the present, 
  1. just to see which existing tests may fail.
  2. based on #5040, so wait on that + may need rebasing

It will be a good idea to adopt #4960 here, which should test legacy operation more thoroughly.
But #4960 still has some outstanding work to complete (review asks for more tests).


pp-mo:
Replaced

Issue #5045, opened 2022-11-02
Pressing the "stable"-link in documentation pages doesn't take you to the corresponding doc page for a stable version of Iris


krikru:
## 🐛 Bug Report

On documentation pages for the latest version of Iris, the "stable"-link in the red box in the top of the page takes you to the _main documentation page_ for a stable version of Iris, and not to the doc page for a stable version _corresponding to the page that is currently displayed_.

## How To Reproduce
Steps to reproduce the behaviour:

1. Go to (e.g.) https://scitools-iris.readthedocs.io/en/latest/generated/api/iris/coord_systems.html.
2. Press the link that says "stable" in the top, red box.

This will take you to the page https://scitools-iris.readthedocs.io/en/stable/.

## Expected behaviour

The link should take you to the page https://scitools-iris.readthedocs.io/en/stable/generated/api/iris/coord_systems.html.

krikru:
What's more, if you view an old documentation page (e.g. [this one](https://scitools.org.uk/iris/docs/v1.9.0/html/iris/iris/coord_categorisation.html)), there will be a top banner saying "You are not reading the most recent version of this documentation. Please see the [latest documentation](https://scitools-iris.readthedocs.io/en/latest/) instead.", which will link you to the main documentation page (not the corresponding documentation page) of the latest version of Iris, which will _also_ have a banner on _its_ top because that version of Iris is not considered stable!

Issue #4396, opened 2021-11-03
Misleading sources of documentation, round 2


jamesp:
Ok, where is this version of the iris docs (v1.9!) being published from?

https://scitools.org.uk/iris/docs/v1.9.0/html/index.html

Am I the only person who gets the wrong docs every time from google?

tkknight:
I believe that is a static build of Iris v1.9 that is hosted via github pages.  The project (https://github.com/SciTools/scitools.org.uk) is configured to use scitools.org.uk via the setting page: https://github.com/SciTools/scitools.org.uk/settings/pages (viewable only if you have the permissions)

All Iris documentation before Iris 3.0 is hosted there, Iris 3.0 and later are on https://scitools-iris.readthedocs.io/en/latest/

jamesp:
Thanks @tkknight that makes sense now.  SEO is a hard problem to solve, but should we consider making a plan to eventually deprecate these docs? From what I can deduce, rtd is better at SEO to ensure the latest docs appear first in google even though they host many versions.

Here are a few things that I think are confusing to new users. Google search for:

1. "iris regridding" -> 1st result is to iris 2.0 on scitools.org.uk.  RTD 3.2 is 2nd in list.
2. "iris projection" -> 1st result is to iris 1.9 on scitools.org.uk
3. "iris plotting" -> 1st result is to iris 2.0 on scitools.org.uk. Second result is iris 0.9 on esc24.github.io. docs for 3xx don't even appear on the list.
4. "iris zonal mean" -> 1st result is to iris 1.9 on scitools.org.uk

tkknight:
I would have thought google search would have eventually pointed to the latest version by now - but clearly something is stopping that.

There are no current plans to retire the old ones.  It is not viable to migrate them to rtd's as they would need work to do this.

We could look to archive these docs as a bundle zip file for reference if needed (perhaps in an archive folder here https://github.com/SciTools/scitools.org.uk/tree/master/iris/docs).  We could do this for all v2.4 and all previous versions.

This approach would break all links to older docs.  We could add a link to the legacy docs from the latest docs for reference.

This may need a little more thought before any action is made incase there are other impacts.

trexfeathers:
Team discussion: Iris 2.3 is still currently widely used at the UK Met Office, but will be moved onto 2.4 etc. in due course. When this happens, we think that will be the appropriate time to 'archive'/zip/... the older non-RTD docs.

jamesp:
[...]

PR #5055, opened 2022-11-09
Reference the RTD version switcher in the docs latest warning

trexfeathers:
### Description

Ref: #5045 and https://github.com/SciTools/iris/issues/4396#issuecomment-1273592280

---


tkknight (approved):

tkknight:
Looks good.  There is a chance that a user is viewing a docs page that is in latest and added in latest, so if they view a stable version via the version switch flyout menu, the browser will try to load a page that does not exist for that earlier version.  Read The docs will show a page like this:

image

I think this will not be a frequent occurrence and the error message is informative so lets get this merged.

trexfeathers:
Thanks @tkknight!

Issue #5016, opened 2022-10-07
Investigate segfaults with NetCDF4 `>1.6.0`


trexfeathers:
## 📰 Custom Issue

A bad interaction between Iris, Dask and NetCDF4 is causing segfaults when loading. This is apparently fixed if Dask is set to single-threaded mode - see https://github.com/Unidata/netcdf4-python/issues/1192#issuecomment-1267390240

This workaround removes a large part of the benefit of using Dask in Iris, so we need to find out if we can instead fix via changes to the Iris loading code.

Before the next release ([`v3.4`](https://github.com/orgs/SciTools/projects/3)), 1 of the below needs to happen:

* Permanent fix - change Iris' loading code to avoid the segfaults
* Temporary workaround - pin NetCDF4 `<=1.6.0`
* Temporary workaround, with big performance regressions - set Dask to single-threaded within Iris

trexfeathers:
@bjlittle has shared how xarray handles NetCDF4 not being thread-safe:

https://github.com/pydata/xarray/blob/main/xarray/backends/file_manager.py#L52

bjlittle:
Ping @valeriupredoi

valeriupredoi:
awesome, cheers @bjlittle - also, let me know if I can help :beer:

trexfeathers:
# Current thinking

Fix will be a lot of work. We should get started as soon as we can. But we don't need to coincide with the planned release schedule.

## Rationale

### `single-threaded`? ❌

Have confirmed the suggestion that the `single-threaded` Dask scheduler avoids the problem. However much of our user-base would not cope with the slowdowns this would bring, especially if they are depending on large compute clusters to get fast enough execution times.

### File locking? ⏳

File locking (like Xarray) is not going to be a quick fix. Iris' load and save code uses file paths all over the place, which makes it difficult to protect access to a file. @bjlittle and I made a few unsuccessful attempts.

We probably need the common currency in Iris' internals to be a specialised file-handling object, instead of file paths. This new object could control file locking and also pickling, which is important for Dask. This will be a large piece of work, and difficult to estimate since it touches so much of Iris.

### Pinning? 😕

[...]

PR #5061, opened 2022-11-11
Netcdf thread safety

pp-mo:
Ultimately, to address #5016 

For now, **trial code for adding thread-safety and file-locking**.
Both in saver and loader.

This trial code is rather rough, and not very thought-through.

**At present**, all of the tests are locking up + not completing.
See notes below..

#### netcdf library is not considered thread-safe
Actually, it strictly never was, but it got worse with 1.6.1 which now releases the GIL during file operations
according to at least one [authorative comment elsewhere](https://github.com/Unidata/netcdf4-python/issues/1192#issuecomment-1255347911)

So in addition to any per-file locking that might be considered desirable (e.g. to control access to a file when doing lazy arithmetic between different netcdf variables), we **_do_** need a global lock to control all our access to the netcdf library, if any of it can be multi-threaded.  
... and with Dask, presumably, access **_can_** always be multi-threaded, since data chunks are processed by separate tasks --> different worker threads.

#### process-based scheduling
[like this](https://github.com/SciTools/iris/blob/4c430bb732f9224963e23d40cd9dc0b693175ca1/lib/iris/tests/integration/test_netcdf.py#L45)
Rather to my surprise, lots of our existing code breaks if you simply switch to process-based Dask workers (as are used by default for Bag computations).
I had rather thought that this would "just work"  :tm:  But it's now clear that a simple "da.store" operation is not serialisable, unless you give it a serialisable lock.  Just as we already saw in the [lazy saving investigation](https://github.com/SciTools/iris/blob/ed654db19a32d596dd5996cb7ca7e7e9ba51cf5b/lib/iris/fileformats/netcdf/saver.py#L2594-L2603).

#### single-threaded
[Like this](https://github.com/SciTools/iris/blob/4c430bb732f9224963e23d40cd9dc0b693175ca1/lib/iris/tests/integration/test_netcdf.py#L44-L46)
Solves the above problems, and probably all others, but it's not a practical solution.
Has already proved useful in tracking down where code uses locks wrong ,I.E. failing to release before re-acquiring --> simply hangs when single-threaded.

#### open/closing of netcdf Dataset objects
This is somewhat simpler than I had feared.
For the reading process, a netCDF4.Dataset is opened by creating a CFReader, from which cubes are extracted, and is **_held_** open :  in the new code, that is [_along with_ the netcdf library global access lock](https://github.com/SciTools/iris/blob/4c430bb732f9224963e23d40cd9dc0b693175ca1/lib/iris/fileformats/cf.py#L1075-L1092) 
During the process, the load code and 'actions' refer to to the dataset and its Variables, but once all the cubes have been produced, the CFReader is discarded : Then the dataset is closed, and the only remaining references are in the [NetcdfDataProxy](https://github.com/SciTools/iris/blob/5c9d9b30015003dc76f65c90db8e780c80a33062/lib/iris/fileformats/netcdf/loader.py#L58-L82) objects used to build lazy arrays -- which contain a filepath + variable-name, and thus don't refer to the 'original' Dataset object.

The only tricky part was that the existing code is relying on [`CFReader.__del__` to close the dataset](https://github.com/SciTools/iris/blob/5c9d9b30015003dc76f65c90db8e780c80a33062/lib/iris/fileformats/cf.py#L1297-L1300).
Which is in fact unreliable, and may not always happen when you expected/wanted it.  Effectively it depends on how the garbage collector works, which is not a constant.  I also found that this is not a technical detail, but really happens.
<details>
<summary> really simple example.. click to expand </summary>

```
    >>> class MyC:
    ...   def __init__(self, x):
    ...     self.x = x
    ...   def __str__(self):
    ...     return f"<MyC id={id(self)}(x={self.x})"
    ...   def __del__(self):
    ...     print('delete: ', self)
    ...
    >>> myc1 = MyC(1)
    >>> myc2 = MyC(2)
    >>> myc2 = None
    delete:  <MyC id=140005502283392(x=2)
    >>> myc1
    <__main__.MyC object at 0x7f55923aef70>
    >>> myc1 = 13
    #
    # ??? why no delete message here ???
    #
    >>> del myc1
    >>> exit()
    delete:  <MyC id=140005502283632(x=1)

```

</details>

So, this is actually unnecessary, and I reckon we would be better off using the CFReader [as a context manager, so it can automatically (always) close after operation](https://github.com/SciTools/iris/blob/4c430bb732f9224963e23d40cd9dc0b693175ca1/lib/iris/fileformats/cf.py#L1094-L1106).  So I am now doing that [here](https://github.com/SciTools/iris/blob/4c430bb732f9224963e23d40cd9dc0b693175ca1/lib/iris/fileformats/netcdf/loader.py#L556).
BTW this is exactly how the [netcdf.saver.Saver is already used](https://github.com/SciTools/iris/blob/5c9d9b30015003dc76f65c90db8e780c80a33062/lib/iris/fileformats/netcdf/saver.py#L2751).

#### nature of errors.
So far, various operations which succeed nevertheless produce a stream of HDF error messages.
Not clear how that can be (if it suceeds), and being multi-threased, naturally it's hard to debug.
Maybe a fault in my code design, or for some reason something else needs guarding 

#### lazy streaming netcdf-to-netcdf in save operations
There were some very obvious things going wrong with save-load-save roundtrips, like [this one](https://github.com/SciTools/iris/blob/5c9d9b30015003dc76f65c90db8e780c80a33062/lib/iris/tests/integration/test_netcdf.py#L98-L111).

This highlights that when streaming _from_ lazy netcdf data to save results, it is necessary that the saving process releases the global netcdf-access lock, since otherwise the lazy input compute will lock up.  
Which means both the lazy-write and the lazy-read operations must independently ensure that the lock is used when reading/writing.
N.B. with synthetic data, like in these tests, you _only_ see this if you save-load-save : The problem is unique to cases where the saved source data is lazy and in a netcdf file :  It's a particularly specific problem.

> it is necessary that the saving process releases the global netcdf-access lock

Which makes changes in the Saver code much more tricky + extensive than those in CFReader
The load_cubes/CFReader action is quite simple, since it involves no "compute" at all, only straight-line code to snapshot the file + create data-proxies.  At the end of which, the file is simply closed.

### Scope
Not considering process-parallel execution yet (dask config "scheduler='processes'", or distributed), which still does not work.
However, I have provided [per-file locking on read](https://github.com/SciTools/iris/blob/4c430bb732f9224963e23d40cd9dc0b693175ca1/lib/iris/fileformats/cf.py#L1067-L1070), and an eventual solution **_may_** need to adopt many of the ideas from #5130

### DIFF
--- a/lib/iris/fileformats/cf.py
+++ b/lib/iris/fileformats/cf.py
@@ -15,9 +15,10 @@ References:
 """
 
 from abc import ABCMeta, abstractmethod
-from collections.abc import Iterable, MutableMapping
+from collections.abc import Iterable, Mapping, MutableMapping
 import os
 import re
+import threading
 import warnings
 
 import netCDF4
@@ -1021,6 +1022,21 @@ class CFGroup(MutableMapping):
 
 
 ################################################################################
+_file_locks: Mapping[str, threading.Lock] = {}
+
+
+def get_filepath_lock(path, already_exists=None):
+    if already_exists is not None:
+        assert already_exists == (path in _file_locks)
+    if path not in _file_locks:
+        _file_locks[path] = threading.RLock()
+    result = _file_locks[path]
+    return result
+
+
+GLOBAL_NETCDF_ACCESS_LOCK = threading.Lock()
+
+
 class CFReader:
     """
     This class allows the contents of a netCDF file to be interpreted according
@@ -1045,28 +1061,49 @@ class CFReader:
 
     def __init__(self, filename, warn=False, monotonic=False):
         self._dataset = None
-        self._filename = os.path.expanduser(filename)
+        filename = os.path.expanduser(filename)
+        filename = os.path.abspath(filename)
+        self._filename = filename
+        self._lock = get_filepath_lock(self._filename)
+        # NOTE: we'd really like to defer this to the start of the related context, but
+        # prior usage requires us to do most of the work within the init call.
+        self._lock.acquire()
 
         #: Collection of CF-netCDF variables associated with this netCDF file
         self.cf_group = self.CFGroup()
 
-        self._dataset = netCDF4.Dataset(self._filename, mode="r")
+        with GLOBAL_NETCDF_ACCESS_LOCK:
+            self._dataset = netCDF4.Dataset(self._filename, mode="r")
 
-        # Issue load optimisation warning.
-        if warn and self._dataset.file_format in [
-            "NETCDF3_CLASSIC",
-            "NETCDF3_64BIT",
-        ]:
-            warnings.warn(
-                "Optimise CF-netCDF loading by converting data from NetCDF3 "
-                'to NetCDF4 file format using the "nccopy" command.'
-            )
+            # Issue load optimisation warning.
+            if warn and self._dataset.file_format in [
+                "NETCDF3_CLASSIC",
+                "NETCDF3_64BIT",
+            ]:
+                warnings.warn(
+                    "Optimise CF-netCDF loading by converting data from NetCDF3 "
+                    'to NetCDF4 file format using the "nccopy" command.'
+                )
 
-        self._check_monotonic = monotonic
+            self._check_monotonic = monotonic
 
-        self._translate()
-        self._build_cf_groups()
-        self._reset()
+            self._translate()
+            self._build_cf_groups()
+            self._reset()
+
+    def __enter__(self):
+        # Enable use as a context manager
+        # N.B. this **guarantees* closure of the file, when the context is exited.
+        # Note: ideally, the class would not do so much work in the __init__ call, and
+        # would do all that here, after acquiring necessary permissions/locks.
+        # But for legacy reasons, we can't do that.  So **effectively**, the context
+        # (in terms of access control) alreday started, when we created the object.
+        return self
+
+    def __exit__(self, exc_type, exc_value, traceback):
+        # When used as a context-manager, **always** close the file on exit.
+        self._close()
+        self._lock.release()
 
     @property
     def filename(self):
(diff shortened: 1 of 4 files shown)


trexfeathers:
Tests have been running for 30mins and counting...

pp-mo:
> Tests have been running for 30mins and counting...

It is still deadlocking in some operations, to be investigated + fixed.
That's known about but I [removed the offending one in integration/test_netcdf.py](https://github.com/SciTools/iris/blob/4c430bb732f9224963e23d40cd9dc0b693175ca1/lib/iris/tests/integration/test_netcdf.py#L170-L184).
Other than that, those tests now succeed -- but with the unexplained warning messages, as previously described.

I told you it was WIP !

pp-mo:
### Notes on existing errors
I have been focussing on the set of tests in `iris.tests.integration.test_netcdf.py` as that has proved a useful testbed.
(where errors emerged in the original CI problems).

I've [skipped out](https://github.com/SciTools/iris/blob/4c430bb732f9224963e23d40cd9dc0b693175ca1/lib/iris/tests/integration/test_netcdf.py#L40-L41) all the tests there which now "just work".
The remaining (unskipped) tests are producing "interesting" behaviour -- in nearly all cases the test succeeds but there are a lot of HDF errors appearing.  N.B. though these are _described_ as errors,  something nevertheless "appears to work"? Could be failing to read data, but not checking actual values ?

[This test](https://github.com/SciTools/iris/blob/4c430bb732f9224963e23d40cd9dc0b693175ca1/lib/iris/tests/integration/test_netcdf.py#L170-L184) was commented out because this one (still) actually **_hangs_**  -- a lot of roundtrip tests used to do that, before the locks were added, but this shows that something there has still been missed.

#### Update Mon 2022-11-14
Discussed with @bjlittle .  
Particularly [this test](https://github.com/SciTools/iris/blob/4c430bb732f9224963e23d40cd9dc0b693175ca1/lib/iris/tests/integration/test_netcdf.py#L518-L529), which triggers a stream of HDF errors even though it is "only" loading cubes
(i.e. not fetching or streaming lazy data).
Confirmed that data is fetched from NetcdfDataProxy-s, even when "just" loading like this.
@bjlittle points out that the post-load merge operation is a likely culprit.
@pp-mo also noted that the CFVariable holds "back-door" accesses to netCDF4 Variable objects, potentially even after the file is closed.  Which should be addressed (but might not be "the" problem here).  Then tried constructing a list of cubes, then yielding after the CFReader context exits, but that did _not_ give any more errors as we might have expected from this?

trexfeathers:
We're currently hoping this can be a bugfix for `v3.3` and `v3.4`.

pp-mo:
Close in favour of #5095

PR #5066, opened 2022-11-14
Clarify Laziness in Iris Functions

### LINKED ISSUE #3292: Clarify laziness status of all iris functions
Make clear in all docstrings which functions are lazy and which are not.

Currently only some functions have this description (presumably those which have been updated to be lazy recently).  It would be much more useful to the user if they could immediately see from every single docstring whether the function is lazy or not.

ESadek-MO:
## 🚀 Pull Request
Addresses #3292.

### Description
Add comments in docstrings for each iris function clarifying if it can maintain laziness or not.

https://esadek-mo-irisdocs.readthedocs.io/en/lazydata/


trexfeathers (changes requested):
Here are objects where I would expect to see a note, but didn't find one:

- `iris.analysis.MEDIAN`
- `iris.analysis.PROPORTION`
- `iris.analysis.WPERCENTILE`
- `iris.analysis.cartography.project`
- `iris.analysis.cartography.rotate_grid_vectors`
- `iris.analysis.geometry.geometry_area_weights`
- `iris.analysis.maths.apply_ufunc`
- `iris.cube.Cube.interpolate`

You've covered everything in `iris.cube.Cube` that mentions coords, probably a few more places than you needed to, but we should have consistency with cell measures and ancillary variables, which have analogous behaviour with coords. Same goes for aux_factories even though they are a bit different.

The note isn't applicable for `iris.palette.is_brewer`, since it doesn't touch any array objects.

trexfeathers (approved):
Great! @ESadek-MO and I have agreed that we'll continue with #3292 after `v3.4`, but this is already a big improvement for user clarity 😁

Happy to merge once this comes out of `Draft`

Issue #5053, opened 2022-11-07
What makes the NAME loader faster than the NetCDF loader?


trexfeathers:
## 📰 Custom Issue

@bsherratt has provided an interesting case. A Name III `.txt` file describing:

- 200 x 200 grid
- ~200 phenomena

The file takes `~20s` to load. If Iris is used to convert the file to NetCDF, that NetCDF file takes `~105s` to load. Both loaders are unreasonably slow in this example, since the grid is shared between all phenomena, yet the `Coord` instances are created on a `Cube`-by-`Cube` basis (see #3172 for discussion on sharing). But why is NetCDF _even_ slower?

File available on request.

bsherratt:
Just to update...

We had a theory that "breaking" the coordinate metadata could speed this up, and it really did.

Pre-processing: rename each variable's `coordinates` attribute to something that iris does not recognise. It will therefore not attempt to create any aux coords on load (so they show up as cubes instead).

<details>
<summary>netCDF4 example</summary>

```python
with nc.Dataset(filename, "a") as file:
    for var in file.variables.values():
        if "coordinates" in var.ncattrs():
            var.renameAttribute("coordinates", "fake_coordinates")
    file.sync()
```

</details>

Post-processing: identify the "cubes" that were supposed to be aux coords, and add them to the relevant cubes. This can be done by overriding the usual netcdf loaders in `iris.fileformats.FORMAT_AGENT` with a simple wrapper around `iris.fileformats.netcdf.load_cubes`:

<details>
<summary>Incomplete example</summary>

```python
def load_cubes(filenames, callback=None, constraints=None):
    # Identify coords
    cubes = []
    unknown = []
    coord_names = set()
    for cube in netcdf.load_cubes(filenames, callback, constraints):
        fake_coords = cube.attributes.get("fake_coordinates", "").split()
        if fake_coords:
            # Definitely a cube
            cubes.append(cube)
            coord_names.update(fake_coords)
        else:
            # Might be a cube or a coord
            unknown.append(cube)

    # Create the coords
    coords = {}
    for cube in unknown:
        if cube.name() in coord_names:
            coord = AuxCoord(cube.data)
            coord.metadata = cube.metadata
            coords[coord.var_name] = coord
        else:
            cubes.append(cube)

    # Add coords to cubes
    for cube in cubes:
        fake_coords = cube.attributes.pop("fake_coordinates", "").split()
        for name in fake_coords:
            # TODO: Assumes scalar...
[...]

Issue #4883, opened 2022-07-29
Trial different memory profiler in benchmarks


pp-mo:
From work elsewhere, we find that it might be better to track Python heap usage, instead of Linux RSS space

At least the observation [here ("When initially run, this claimed that no memory at all was used ")](https://github.com/SciTools/iris/issues/4845#issuecomment-1198220407) is relevant :  It show that "RSS" is not working for a simple tests where a setup may allocate space that is then re-used within the tested operation -- but "tracemalloc" does seem to do that right.

pp-mo:
A really simple tracemalloc  solution may look like this (thanks to @jamesp )
```
class MemoryMonitor:
    def __init__(self):
        self.memory_mb = None

    @contextmanager
    def context(self):
        """Measure the time+memory used by a codeblock."""
        tracemalloc.start()
        yield
        _, peak_mem = tracemalloc.get_traced_memory()
        tracemalloc.stop()
        self.memory_mb = peak_mem * 1.0 / 1024 ** 2

memmon = MemoryMonitor()
with memmon.context():
  op()
print(memmon.memory_mb)

```

pp-mo:
TODO : need to compare results with/without offline to assess.
As @trexfeathers points out, you **can't use ASV for this !**

pp-mo:
Having used tracemalloc elsewhere, I'm increasingly convinced that it does offer a better solution.
E.G. See #5753

pp-mo:
> Having used tracemalloc elsewhere, I'm increasingly convinced that it does offer a better solution.
> E.G. See #5753

See now also : #5796

trexfeathers:
Closed by #5981

PR #5069, opened 2022-11-15
Trial solution.

pp-mo:
See #5053

### DIFF
--- a/lib/iris/fileformats/netcdf/loader.py
+++ b/lib/iris/fileformats/netcdf/loader.py
@@ -211,26 +211,34 @@ def _get_actual_dtype(cf_var):
     return dummy_data.dtype
 
 
-def _get_cf_var_data(cf_var, filename):
-    # Get lazy chunked data out of a cf variable.
-    dtype = _get_actual_dtype(cf_var)
+_NONLAZY_SMALL_AUXCOORDS = True
 
-    # Create cube with deferred data, but no metadata
-    fill_value = getattr(
-        cf_var.cf_data,
-        "_FillValue",
-        netCDF4.default_fillvals[cf_var.dtype.str[1:]],
-    )
-    proxy = NetCDFDataProxy(
-        cf_var.shape, dtype, filename, cf_var.cf_name, fill_value
-    )
-    # Get the chunking specified for the variable : this is either a shape, or
-    # maybe the string "contiguous".
-    chunks = cf_var.cf_data.chunking()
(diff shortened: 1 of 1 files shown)


pp-mo:
The test results are encouraging : nothing unexpected.

pp-mo:
At present, a NetcdfDataProxy contains [`__slots__ = ("shape", "dtype", "path", "variable_name", "fill_value")`](https://github.com/SciTools/iris/blob/aa2bfa9f59740027fb4ce4846e419df959f9bc72/lib/iris/fileformats/netcdf/loader.py#L61) : """.
So, if we're interested in the smallest cases (like scalar coords), then using "sizeof" on simple python objects gives, typically ...
  * shape : (1,) ~**32**
  * dtype('f8') :  ~**96**
  * path : str('X' * 40) : ~**89**
  * variable_name : str('X' * 10) : ~**59**
  * fill_value: 1.e30 : ~**24**

Plus "sizeof" a NetCDFDataProxy object itself is ~**56**
So, that's about 350 bytes.

However, the question of how much memory a Python object "costs" has, it seems generally agreed,
[no definite straightforward answer](https://stackoverflow.com/a/35645/2615050).
Not least because, if other objects are referenced, you need to judge whether they are
unique to this structure, or may be shared elsewhere, or should count as separate items in their own right.

Also, look at this ...
```
    >>> miniarr = np.array([], dtype=np.float64)
    >>> miniarr.__sizeof__()
    112
    >>> miniarr.data.__sizeof__()
    168
    >>>
```
So, it looks like anything beyond a Python primitive object is generally rather hard to assess for memory consumption.


### So Instead 
Using very simplistic measures, based on "tracemalloc", as explained in #4883
```
    >>> arrs = []
    >>> with mm.context():
    ...  for n in range(2000):
    ...    mindp = NetCDFDataProxy(shape=(1,), dtype=np.dtype('f8'), path=' ' * 40, variable_name=' ' * 10, fill_value=1.e30)
    ...    minarr = da.from_array(mindp, chunks=-1, meta=np.ndarray)
    ...    arrs.append(minarr)
    ...
    >>> mm.memory_mb * 2.**20 / 2000
    4883.27
```

This example is intending to mimic an _absolutely minimal_  (1-point) lazy array, like that within a **_scalar coordinate_**.

Likewise *100 and *1 measured (average) sizes of     5144.45 and   5496.0
So, **4800 - 5500 bytes seems like a reasonable minimum size ??**

It seems a lot, doesn't it ?!?

pp-mo:
Replaced by #5229

PR #5074, opened 2022-11-16
`iris.pandas.as_data_frame()` n-dimensional `Cube` conversion

trexfeathers:
### Description

This finally merges @hsteptoe's brilliant work from the `pandas_ndim` feature branch, allowing `as_data_frame()` to convert `Cube`s of any dimensionality to pandas `DataFrame`s, complete with optional dimensional metadata.

### Questions

- Can this be merged as a squash? It's a single self-contained feature that we have collaborated on.
- If so, should the What's New entry be updated to use the number of this PR (`#5074`), rather than `#4669`?
  (Guessing that would mean another tiny PR targeting `pandas_ndim`).

---


docs/src/whatsnew/latest.rst
```diff
@@ -25,7 +25,8 @@ This document explains the changes made to Iris for this release
 📢 Announcements
 ================
 
-#. Welcome to `@ESadek-MO`_ and `@TTV-Intrepid`_  who made their first contributions to Iris 🎉
+#. Welcome to `@ESadek-MO`_, `@TTV-Intrepid`_ and @`hsteptoe`_,  who made their
```
trexfeathers, line 28:
```suggestion
#. Welcome to `@ESadek-MO`_, `@TTV-Intrepid`_ and `@hsteptoe`_,  who made their
```

🙄

stephenworsley:
I think merging without squashing is probably the proper way to go here, I think the only practical difference would be that the blame would point to the individual PRs, but there seems to be enough discussion in those extra PRs that this is worthwhile. I think #4669 is still the most important PR to mention in the whatsnew since it was authored by @hsteptoe, but perhaps you could also add a reference to this PR for completeness. With merging, I think the changes to the whatsnew are probably better done as a PR so that we aren't left with direct commits in the history.

stephenworsley (approved):
Good job! I think this is all good to merge in now.

PR #5075, opened 2022-11-17
More accurate netcdf4 pin `<1.6.1`.

trexfeathers:
### Description

See #5016. We're on the way to a proper fix in #5061 but in the meantime we need a netcdf4 pin that can account for future releases of the package, rather than being specific to avoiding one specific release.

---


trexfeathers:
Thanks @bjlittle!

PR #5078, opened 2022-11-17
Documentation updates for `v3.4.0rc0` release.

trexfeathers:
### Description

---


docs/src/whatsnew/3.4.rst
```diff
@@ -1,7 +1,7 @@
 .. include:: ../common_links.inc
 
-|iris_version| |build_date| [unreleased]
```
stephenworsley, line 3: I believe the previous times we've done this we've set the date and version at this point, e.g. https://github.com/SciTools/iris/pull/4921. At the very least I think we should do this for the actual release so that it doesn't change when we build 3.4.1

stephenworsley (changes requested):
Just a small question about setting the date and version number.

docs/src/whatsnew/3.4.rst
trexfeathers: These are placeholders that get automatically replaced:

https://github.com/SciTools/iris/blob/1c05e92dc4b5f1652ec0d2b77055e151b4fbec7a/docs/src/conf.py#L141-L142

Here's how this built locally for me (the pictured version is still `dev` because we haven't yet tagged the release):

image

trexfeathers: Discussion with @stephenworsley: we should be hard-coding for actual releases. The placeholder replacement is only useful for the `latest` builds between releases.

2e866db1abfd6881b6eb77859ad55520fe7821ec

stephenworsley (approved):
Looks good.

PR #5101, opened 2022-12-09
Remove test timings

rcomer:
### Description

Here is some more Iris functionality that I did not know existed, until @pp-mo mentioned it the other day.  Described by the removed comment
https://github.com/SciTools/iris/blob/6eb0401011a8edce51a942a176e2feb06f4b3f09/lib/iris/tests/__init__.py#L940-L946

I tried to run this with `unittest discover` as the comment recommends, but got errors, particularly from [test_pandas.py](https://github.com/SciTools/iris/blob/main/lib/iris/tests/unit/pandas/test_pandas.py).  I found that this functionality can be used with `pytest -s -v`, so if we do want to keep the functionality we should probably document that.  I'm unsure if it can be made to work with `pytest-xdist`.

In any case, pytest provides its own functionality to tell you the timings of the slowest running tests via the `--durations` flag, and of course we now have some `pytest` tests which don't inherit from `IrisTest`, so won't work with the bespoke functionality unless we go through and add the decorator.

More generally, the more we can simplify these test classes, the better chance we have of eventually ditching `unittest` altogether.

This functionality was added at #2372.

---

### DIFF
--- a/docs/src/whatsnew/latest.rst
+++ b/docs/src/whatsnew/latest.rst
@@ -76,6 +76,9 @@ This document explains the changes made to Iris for this release
 #. `@fnattino`_ changed the order of ``ncgen`` arguments in the command to
    create NetCDF files for testing  (caused errors on OS X). (:pull:`5105`)
 
+#. `@rcomer`_ removed some old infrastructure that printed test timings.
+   (:pull:`5101`)
+
 
 .. comment
     Whatsnew author names (@github name) in alphabetical order. Note that,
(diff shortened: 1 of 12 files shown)


lbdreyer (changes requested):
Looking great! Thanks for doing this @rcomer !

I spotted one other mention of IrisTest_nometa here: https://github.com/SciTools/iris/blame/main/lib/iris/tests/graphics/README.md#L27

Do you mind also updating that in this PR?

rcomer:
Thanks @lbdreyer, good spot!  I only grepped the python files.  Now fixed and also rebased to address the whatsnew conflict.  I don't know why the CI isn't running 😕

lbdreyer (approved):

PR #5109, opened 2022-12-14
Fix link checks

trexfeathers:
### Description

Since the offending link is in [`whatsnew/1.0.rst`](https://github.com/SciTools/iris/blob/5555d455df4bfdbaec607e8a8da439ef8663dc1e/docs/src/whatsnew/1.0.rst), and is proving difficult to debug (only a problem on GHA), I'd say it's not worth fixing.

Moreover: it is adjacent to a slightly different WORKING link for ECMWF, so I have decided to remove the broken one rather than just ignore it during link checks - we get very little benefit by keeping it, but risk the link going silently stale in future.

---


trexfeathers:
Interestingly I can't replicate this when I run the same `nox --session linkcheck` locally, so it's something specific to GHA.

trexfeathers:
Since the offending link is in [`whatsnew/1.0.rst`](https://github.com/SciTools/iris/blob/5555d455df4bfdbaec607e8a8da439ef8663dc1e/docs/src/whatsnew/1.0.rst), and is proving difficult to debug (only a problem on GHA), I'd say it's not worth fixing.

Moreover: it is adjacent to a slightly different WORKING link for ECMWF, so I have decided to remove the broken one rather than just ignore it during link checks - we get very little benefit by keeping it, but risk the link going silently stale in future.

docs/src/whatsnew/latest.rst
```diff
@@ -69,6 +69,8 @@ This document explains the changes made to Iris for this release
 
 #. `@rcomer`_ clarified instructions for updating gallery tests. (:pull:`5100`)
 
+#. `@trexfeathers`_ fixed some broken links. (:pull:`5109`)
+
```
ESadek-MO, line 73: Could maybe be slightly more specific?

ESadek-MO (changes requested):
Hey @trexfeathers, this looks good! Would perhaps make the Whatsnew entry more specific, then I think this would be good to go!

docs/src/whatsnew/latest.rst
trexfeathers: 17303d78ba1660eb4f115c5932d355abc191b7f9

ESadek-MO (approved):

Issue #4679, opened 2022-04-04
Make it clearer / easier to find how to cite Iris


wjbenfold:
## 📚 Documentation

It's currently pretty hard to work out how to cite Iris (and not in the docs). I think the only indication is that we have a Zenodo tag on our GitHub README. I'd suggest improving this.

Examples from elsewhere (both what they use and how you find out about it):
- Pandas use Zenodo and a paper: https://pandas.pydata.org/about/citing.html
- Numpy have a paper: https://numpy.org/citing-numpy/
- SciPy have a paper: https://scipy.org/citing-scipy/
- Matplotlib have a paper and a Zenodo DOI (Digital Object Identifier): https://matplotlib.org/stable/users/project/citing.html

tkknight:
We have a docs page for this but I expect it could do with an overhaul.  

https://scitools-iris.readthedocs.io/en/latest/userguide/citation.html#citing-iris

wjbenfold:
Odd - when I search the docs for the word "cite" I don't get that page, only a very old what's new entry (which does link to that page). Maybe it's just a matter of cross-linking a bit better and updating that page

trexfeathers:
@HGWright any thoughts on the status of this? I've lost track if we've already done the "make it clearer / easier" step

HGWright:
@trexfeathers Thanks for the reminder I had forgoten about this. It still needs work to finish #5161.

ESadek-MO:
Completed in #5483

PR #5116, opened 2022-12-20
Updated citation

jonseddon:
### Description
A documentation tweak to update the citation to a more recent version (it's currently using version 1.2) and to better use DOIs.

Implements #4679


trexfeathers:
I like this. Anyone in @SciTools/iris-devs have expertise in citations, and be happy to take a look?

docs/src/userguide/citation.rst
```diff
@@ -15,11 +15,12 @@ For example::
...
- title = {Iris: A Python package for analysing and visualising meteorological and oceanographic data sets},
- edition = {v1.2},
- year = {2010 - 2013},
+ title = {Iris: A powerful, format-agnostic, and community-driven Python package for analysing and visualising Earth science data },
+ edition = {v3.4},
+ year = {2010 - 2022},
  address = {Exeter, Devon },
- url = {http://scitools.org.uk/}
+ url = {http://scitools.org.uk/},
+ doi = {10.5281/zenodo.595182}
```
bouweandela, line 23:
```suggestion
 doi = {10.5281/zenodo.7386117}
```
is the DOI for iris 3.4 (see https://doi.org/10.5281/zenodo.595182)

bouweandela:
It would be nice to make the package easily citable from GitHub by adding a citation file, you can find more information about adding that here: https://docs.github.com/en/repositories/managing-your-repositorys-settings-and-features/customizing-your-repository/about-citation-files

docs/src/userguide/citation.rst
trexfeathers: `595182` is the parent ID for all releases of Iris, and automatically redirects to the latest release. I prefer `595182` as it avoids a manual update for each new release.

jonseddon: I did choose `595182` so that it was valid for all versions. If someone explicitly wants to reference a specific version then they can update the DOI.

jonseddon:
#5118 added to cover @bouweandela's suggestion about the citation file. I won't add that in this PR to keep everything separate.

docs/src/userguide/citation.rst
bouweandela: At the moment, it looks like this BibTex file should be considered an example. It refers to a specific version of Iris on line 19 and a specific year on line 20. Just replacing the version-specific DOI with the Concept DOI won't prevent it from going out of date.

trexfeathers: I'm convinced, @bouweandela's suggestion looks like the way to go!

stephenworsley (changes requested):
This looks great in principle, It would be good to add instructions for how/when to update this in future releases to the developer docs. (somewhere in here, probably as part of the post release steps https://github.com/SciTools/iris/blob/main/docs/src/developers_guide/release.rst)

jonseddon:
@stephenworsley note added to https://github.com/SciTools/iris/blob/main/docs/src/developers_guide/release.rst

I guess that I probably need to rebase as I've done "git pull upstream main" several times. Please let me know if a rebase is required - I'm not used to working on a Git repo that's so active.

pp-mo:
>  I probably need to rebase as I've done "git pull upstream main" several times.

I think on the contrary, if you've been updating from main then a rebase is **_not_** required.
You definitely don't want to mix the two, as when code changes from main are included in the development commits on your branch (i.e. the ones which _aren't_ merges from the target), that _will_ usually cause conflicts when trying to merge back.

The tests panel here shows that it is currently OK to merge -- i.e. has no conflicts with the current main branch.  If there was a problem, it would say there that there are conflicts (and merge button will be disabled).

Rebasing is another way of fixing merges conflicts, whereby git re-bases all your commits onto the latest in the target branch.  
It effectively re-plays your change commits, one at a time, starting from the current target and merging each changes into it. 
At each step, if it fails to automate the change you must manually resolving conflicts, just as if merging.
The benefit of this approach is that the history on the the PR branch is much simpler, not including merges back from the target, and so you can easily see "what I changed between point A and point B of this development process" without changes drawn from main complicating the difference.
It's great when it works!  But when it goes wrong, it's usually easier to squash all your commits into one and resolve all conflicts in one go -- which of course destroys the sequence, and may not be good if that makes the changes really hard to understand.

It's also worth noting that some people seem to hate rebase, because it _automatically_ re-writes your changes so that they are now "magically" ok -- i.e. will merge without conflict.  Yet, you didn't do that, and in principle you should carefully review each new commit + check it makes sense and you are happy to "sign it off".  
However, in **_my_** opinion this is really no different from _any_ automatic merge : whenever it says "merges with no conflicts", that means it **_is_** making automated changes to what you wrote!

Bit of a rant. HTH ?!?

stephenworsley (approved):

PR #5138, opened 2023-01-18
PP Proleptic Gregorian

### LINKED ISSUE #3561: BUG: iris pp treats standard calendar as a proleptic gregorian and ignores real proleptic gregorian calendars
https://github.com/SciTools/iris/blob/75c7570bf4d509295e239aff47b70324cd68c22d/lib/iris/fileformats/pp_save_rules.py#L398-L404

Standard calendar cannot be treated as a Proleptic Gregorian for dates `1582-01-01`
More context here: [ancil/ticket #1159](https://code.metoffice.gov.uk/trac/ancil/ticket/1159)

UM F03 more [recently clarified](https://code.metoffice.gov.uk/trac/um/ticket/5272) that the calendar it expects isn't Standard but a Proleptic Gregorian.  Not only is iris pp save ignoring time coordinates with `proleptic_gregorian` calendars entirely, it also allows saving Standard dates prior to `1582-01-01` (as though they are proleptic gregorian dates when they are not).

rcomer:
### Description

A proposal to resolve #3561, hopefully without creating breaking changes for the user.  Quoting myself from that issue:

> Noting that
>
>* 1st January 1970 is the same date in the Proleptic Gregorian and Standard calendars
>* Any time point expressed as _hours since_ 1st January 1970 is also therefore the same in the Proleptic Gregorian and Standard calendars
>
>So, once we have converted our pp time headers into a coordinate with unit "hours since 1970-01-01 00:00:00", it is somewhat arbitrary whether we label it "proleptic_gregorian" or "standard".  Would it therefore be reasonable to use the Proleptic Gregorian calendar to do the conversion from pp-header to coordinate, but still label the coordinate with the "standard" calendar?  It would be a bit of a fudge, but would mean any change in behaviour would be limited to dates before 15th October 1582, and therefore a bugfix.

Similarly, when saving we can convert a standard calendar unit to Proleptic Gregorian with the `cf_units.Unit.change_calendar` method, and then use that new unit to do the `date2num` conversion.

I've done a fair bit of test wrangling.  I think the last few failures reflect that the years 0 and 1 are genuinely different in the two calendars, so the expected values could just be updated.  Leaving that for now and opening as draft to see what people think.

---

### DIFF
--- a/lib/iris/fileformats/pp.py
+++ b/lib/iris/fileformats/pp.py
@@ -1078,7 +1078,7 @@ class PPField(metaclass=ABCMeta):
     def calendar(self):
         """Return the calendar of the field."""
         # TODO #577 What calendar to return when ibtim.ic in [0, 3]
-        calendar = cf_units.CALENDAR_STANDARD
+        calendar = cf_units.CALENDAR_PROLEPTIC_GREGORIAN
         if self.lbtim.ic == 2:
             calendar = cf_units.CALENDAR_360_DAY
         elif self.lbtim.ic == 4:
--- a/lib/iris/fileformats/pp_load_rules.py
+++ b/lib/iris/fileformats/pp_load_rules.py
@@ -503,6 +503,11 @@ def _new_coord_and_dims(
         a new (coordinate, dims) pair.
 
     """
+    if (
+        units.is_time_reference()
+        and units.calendar == cf_units.CALENDAR_PROLEPTIC_GREGORIAN
+    ):
+        units = units.change_calendar(cf_units.CALENDAR_STANDARD)
     bounds = lower_and_upper_bounds
     if is_vector_operation:
         dims, points, bounds = _reduce_points_and_bounds(points, bounds)
@@ -548,7 +553,7 @@ def _epoch_date_hours_internals(epoch_hours_unit, datetime):
         if m == 0:
             # Add a 'January', by changing month=0 to 1.
             m = 1
-            if calendar == cf_units.CALENDAR_STANDARD:
+            if calendar == cf_units.CALENDAR_PROLEPTIC_GREGORIAN:
                 days_offset += 31
             elif calendar == cf_units.CALENDAR_360_DAY:
                 days_offset += 30
@@ -561,7 +566,7 @@ def _epoch_date_hours_internals(epoch_hours_unit, datetime):
         if y == 0:
             # Add a 'Year 0', by changing year=0 to 1.
             y = 1
-            if calendar == cf_units.CALENDAR_STANDARD:
+            if calendar == cf_units.CALENDAR_PROLEPTIC_GREGORIAN:
                 days_in_year_0 = 366
             elif calendar == cf_units.CALENDAR_360_DAY:
                 days_in_year_0 = 360
--- a/lib/iris/fileformats/pp_load_rules.py
+++ b/lib/iris/fileformats/pp_load_rules.py
@@ -507,7 +507,9 @@ def _new_coord_and_dims(
         units.is_time_reference()
         and units.calendar == cf_units.CALENDAR_PROLEPTIC_GREGORIAN
     ):
-        units = units.change_calendar(cf_units.CALENDAR_STANDARD)
+        units = cf_units.Unit(
+            "hours since epoch", calendar=cf_units.CALENDAR_STANDARD
+        )
     bounds = lower_and_upper_bounds
     if is_vector_operation:
         dims, points, bounds = _reduce_points_and_bounds(points, bounds)
--- a/lib/iris/fileformats/pp_save_rules.py
+++ b/lib/iris/fileformats/pp_save_rules.py
@@ -114,6 +114,11 @@ def _general_time_rules(cube, pp):
     frt_coord = scalar_coord(cube, "forecast_reference_time")
     clim_season_coord = scalar_coord(cube, "clim_season")
 
+    if time_coord is not None:
+        time_unit = time_coord.units
+        if time_unit.calendar == "standard":
+            time_unit = time_unit.change_calendar("proleptic_gregorian")
+
     cm_time_mean = scalar_cell_method(cube, "mean", "time")
     cm_time_min = scalar_cell_method(cube, "minimum", "time")
     cm_time_max = scalar_cell_method(cube, "maximum", "time")
@@ -122,7 +127,7 @@ def _general_time_rules(cube, pp):
     if time_coord is not None and fp_coord is None and frt_coord is None:
         pp.lbtim.ia = 0
         pp.lbtim.ib = 0
-        pp.t1 = time_coord.units.num2date(time_coord.points[0])
+        pp.t1 = time_unit.num2date(time_coord.points[0])
         pp.t2 = cftime.datetime(0, 0, 0, calendar=None, has_year_zero=True)
 
     # Forecast.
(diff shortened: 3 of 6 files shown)


trexfeathers:
## 📢 @SciTools/iris-devs 📢

Need to make sure we get this one right. If you have any experience with calendars / PP-format it would be great to have your informed opinion ❤

rcomer:
Full disclosure: this is my first time doing anything in `fileformats` so, um 😨

trexfeathers:
@nhsavage are you able to test this branch against any of your scripts, to see if it addresses your concerns?

nhsavage:
I am struggling a bit to do this - the problem is that the bug mainfested when using ANTS not directly in Iris. trying to mimic it just in Iris doesn't seem to work and at the moment, ANTS is at Iris 2.3.0

nhsavage:
so I've boiled the Iris part of the problem down to something simple. Take any netCDF file with a proleptic gregorian calendar and run:

cube=iris.load_cube(infile)
cube.attributes["STASH"] = iris.fileformats.pp.STASH.from_msi("m01s00i024")
time_coord = cube.coord('time')
# save to a pp file
iris.save(cube,'sst_out_fix.pp')

if you look at the output of this, then the calendar is set to 0 which is not a valid. Header from pumf:

(13) lbtim   :                 20

it should be 21 as the calendar is proleptic gregorian.

trexfeathers:
> if you look at the output of this, then the calendar is set to 0 which is not a valid. Header from pumf:
> 
> `(13) lbtim : 20`
> 
> it should be 21 as the calendar is proleptic gregorian.

I can't replicate this, I get:

```
  (13) lbtim   :                       0
```

<details><summary>Here's my code</summary>

```python
from cf_units import CALENDAR_PROLEPTIC_GREGORIAN, Unit
import numpy as np
from numpy import random

import iris
from iris.coords import AuxCoord, DimCoord
from iris.cube import Cube

dim_names_lens = [("longitude", 2), ("latitude", 3)]
dim_coords = [DimCoord(np.arange(l), standard_name=n) for n, l in dim_names_lens]

time_unit = Unit("days since 2001-01-01", calendar=CALENDAR_PROLEPTIC_GREGORIAN)
time_coord = AuxCoord(1, standard_name="time", units=time_unit)

prep_cube = Cube(
    random.rand(*[c.shape[0] for c in dim_coords]),
    standard_name="sea_surface_temperature",
    units="K",
    dim_coords_and_dims=[(c, ix) for ix, c in enumerate(dim_coords)],
    # Make this a scalar coord.
    aux_coords_and_dims=[(time_coord, None)],
)

infile = "sst_in.nc"
iris.save(prep_cube, infile)

cube=iris.load_cube(infile)
# save to a pp file
iris.save(cube,'sst_out_fix.pp')

```

</details>

rcomer:
Looks like I missed a bit 👀

nhsavage:
First of all LBTIM is defined thus (in my well thumbed copy of UMDP F03, excellent reading if you can't sleep)
BTIM is coded as (100*IA + 10*IB + IC) where:
for time aggregated fields IA is the time interval in hours between the individual fields from which
the mean was computed [zero in both our cases. yours isn't a time mean, the netCDF I read doesn't include this info

IB = 2 if the field is a time mean between T1 and T2 [my data is time mean yours isn't]

IC (the only relevant part for this PR)

= 0 if ’model time’ is used for T1 and T2 (i.e. only day number, hour and minute are set).
– = 1 if the Proleptic Gregorian calendar is used for T1 and T2. This calendar is equivalent to the
Gregorian calendar for dates after 1582. Applications producing data using the Proleptic Grego-
rian calendar should extend back in time to BC years in a way that make the date continuous:
year 1 is 1 AD, year 0 is 1 BC, year -1 is 2 BC etc. (Note some ancillary files use year 0 in this
calendar for ancillary fields that do not vary with time.)
– = 2 if the ’360-day’ calendar (i.e. 12 30-day months) is used for T1 and T2.
– = 3 if ’model time’ is used for T1 and T2 (i.e. only day number, hour and minute are valid; year,
month and day in month are to be ignored if set).
– = 4 if the ’365-day’ (no leap year) calendar is used for T1 and T2

nhsavage:
this code can replace the save and pumf steps
```
fields = iris.fileformats.pp.as_fields(prep_cube)

for field in fields:
    print(field.lbtim)
```

currently gives 0 (no calendar defined) which is wrong

rcomer:
My latest commit gives lbtim of 1 with @trexfeathers' code.

trexfeathers:
> My latest commit gives lbtim of 1 with @trexfeathers' code.

I can replicate this. Thanks @nhsavage for the help debugging.

I'm concerned about whether I would have spotted this even if I were to carefully read UMDP F03 (and any other applicable resources); it's just not where my talents lie. I can do my best (resource allowing), but it would be great to have more users trying this out too (it would be mean to lean exclusively on Nick).

I will see if I can apply the changes here to an older version of Iris to allow more testing by people close to the ANTS tool.

nhsavage:
I note that #3561 is about to be closed as stale. Is there an update on this?

trexfeathers:
> I note that #3561 is about to be closed as stale. Is there an update on this?

Sorry, no.

Nothing unique to this topic, but it becoming increasingly difficult to deliver improvements to Iris:

- A gradual increase in the complexity of the codebase.
- A marked increase in frequency of upstream breaking changes - unplanned work ™. We have discovered 3 in the last two weeks, and I have personally needed to 'down tools' many times in the last year.

As Iris 'primary lead' I am thinking carefully about changes we might make to our design and our working practices to free up more resource/focus for Iris improvements.

nhsavage:
thanks for clarifying that. I will let #3561 close then. It's unfortunate, but I understand. Is it worth closing this PR?

trexfeathers:
>Is it worth closing this PR?

Good point. Since @rcomer has put in the work I would say it ought to get higher priority than other things (which might still be abstract ideas). I will try to treat it with more importance going forward.

rcomer:
I think the reason this stalled was that we didn't have a reproducing example to demonstrate the problem described in the issue.  So I cannot actually prove that this fixes it.  If someone could provide such an example, we could pick this up again.

arjclark:
From the ants end, I'm not worried about you testing it with a backported version. We'll catch it up (and many other fixes!) next time we update our dependencies. Original reporter in #3561 is long moved on, so @nhsavage is probably best person to know what's needed for testing. Can confirm from that original ticket that climate calendars under the "proper" calendar are starting to appear (the standard vs gregorian naming change since Iris 3.3 caught out the afterburner climate model monitor package recently).

nhsavage:
I'll try and manufacture some data. I am struggling to remember exactly the circumstances it failed in though. I think it was when the base time of the calendar was 1/1/1 AD in some Aussie files. I need to think exactly what we want to capture here.

nhsavage:
making some test data with mule doesn't seem to replicate the issue. Next I am going to go back to the data that gave use the problems in the first place SST and SeaIce cmip5 ACCESS 1.3.

nhsavage:
I have tried and failed to reproduce the problem. It is so long ago now that probably most people with this issue have implemented work arounds, so I am wondering if we should just close this now as won't fix

rcomer:
I  agree there is no sense keeping this open if we can no longer reproduce the problem.

@arjclark do you have an opinion on this?

arjclark:
@hdyson - I think I'm happy to let this lapse and get closed if you are?

hdyson:
We have a longstanding workaround in ANTS for this which we've not needed to update in a while (it does follow our "workaround" working practice so a grep for "3561" is sufficient to find it - I don't want to link to the trac location since we're mid migration to GitHub).  So I'm happy to let this be closed - it's not something that's actively causing us issues.

I will add that we do have unittests for our workaround.  If this does become a priority due to other users, I think we should be able to adapt those unittests to generate an example if that would be helpful?  It looks like the trick is to have a time coordinate with "<units> since 0000-01-01".

rcomer:
OK, I will close this but keep the branch in case we ever want to pick this up again.

PR #5134, opened 2023-01-12
Rewrite release instructions.

trexfeathers:
### Description

There are lots of simple steps to a release, making it easy to forget one or do one wrong. This is an ideal application for a [do-nothing script](https://blog.danslimmon.com/2019/07/15/do-nothing-scripting-the-key-to-gradual-automation), so I've written one.

I've also re-organised the docs page on releases, as that was proving difficult to follow logically. When reviewing it's probably easiest to read it as a new page, rather than trying to track the changes.

---
[Consult Iris pull request check list]( https://scitools-iris.readthedocs.io/en/latest/developers_guide/contributing_pull_request_checklist.html)

PR #5141, opened 2023-01-30
add readme #showyourstripes

bjlittle:
### Description

At the very least, I think we should do our (minimal) bit to spread the good word for [#ShowYourStripes](https://showyourstripes.info/s/globe/)...

This is the suggested [README.md](https://github.com/bjlittle/iris/tree/readme-showyourstripes#-----) change.

Thanks @ed-hawkins et al :100: :rocket: :star_struck: 

---

### DIFF
--- a/README.md
+++ b/README.md
@@ -54,3 +54,23 @@ For documentation see the
 developer version or the most recent released
 <a href="https://scitools-iris.readthedocs.io/en/stable/">stable</a> version.
 </p>
+
+## [#ShowYourStripes](https://showyourstripes.info/s/globe)
+
+<h4 align="center">
+  <a href="https://showyourstripes.info/s/globe">
+    <img src="https://raw.githubusercontent.com/bjlittle/geovista-media/2023.01.1/media/readme/showyourstripes.png"
+         alt="#showyourstripes Global 1850-2021"></a>
+</h4>
+
+**Graphics and Lead Scientist**: [Ed Hawkins](http://www.met.reading.ac.uk/~ed/home/index.php), National Centre for Atmospheric Science, University of Reading.
+
+**Data**: Berkeley Earth, NOAA, UK Met Office, MeteoSwiss, DWD, SMHI, UoR, Meteo France & ZAMG.
+
+<p>
+<a href="https://showyourstripes.info/s/globe">#ShowYourStripes</a> is licensed under a
+<a href="https://creativecommons.org/licenses/by/4.0/">Creative Commons Attribution 4.0 International License</a>
+<a href="https://creativecommons.org/licenses/by/4.0/">
+  <img src="https://i.creativecommons.org/l/by/4.0/80x15.png" alt="creative-commons-by" style="border-width:0"></a>
+</p>
+
--- a/README.md
+++ b/README.md
@@ -69,7 +69,7 @@ developer version or the most recent released
 **Data**: Berkeley Earth, NOAA, UK Met Office, MeteoSwiss, DWD, SMHI, UoR, Meteo France & ZAMG.
 
 <p>
-<a href="https://showyourstripes.info/s/globe">#ShowYourStripes</a> is licensed under a
+<a href="https://showyourstripes.info/s/globe">#ShowYourStripes</a> is distributed under a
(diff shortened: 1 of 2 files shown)


tkknight:
can see the rendered README: https://github.com/bjlittle/iris/blob/readme-showyourstripes/README.md

README.md
```diff
@@ -54,3 +54,23 @@ For documentation see the
 developer version or the most recent released
 <a href="https://scitools-iris.readthedocs.io/en/stable/">stable</a> version.
 </p>
+
+## [#ShowYourStripes](https://showyourstripes.info/s/globe)
+
+<h4 align="center">
+  <a href="https://showyourstripes.info/s/globe">
+    <img src="https://raw.githubusercontent.com/bjlittle/geovista-media/2023.01.1/media/readme/showyourstripes.png"
```
bjlittle, line 62: When `geovista` migrates to the `SciTools` organisation, we can update the URL to this asset.

bjlittle:
@trexfeathers I'll find the fix for `pre-commit.ci`... :disappointed:

bjlittle:
This seems relevant, see https://github.com/PyCQA/isort/issues/2077

README.md
trexfeathers: This must be hosted somewhere already, right?! You've linked to their wonderful website - is there nothing there that we can reference directly?

trexfeathers (changes requested):
Great idea @bjlittle 🙂

As well as my comment below, I think we should use the What's New as an opportunity to make a little more noise.

README.md
trexfeathers: https://showyourstripes.info/stripes/GLOBE---1850-2021-MO.png ?

rcomer: It also gets updated every year

trexfeathers: Ugh, I give up - doesn't look like there's a way of doing this in HTML without blurring.

bjlittle: @trexfeathers Pushed a stable workaround... I think, after a wee bit of detective work :male_detective: 

See https://github.com/SciTools/iris/pull/5141/commits/b0e7e1d99387e67eb4ee671eb943a47c45c77d7d

bjlittle:
@trexfeathers 
> Great idea @bjlittle slightly_smiling_face
> 
> As well as my comment below, I think we should use the What's New as an opportunity to make a little more noise.

Agreed. See https://github.com/SciTools/iris/pull/5141/commits/1593c8feaa2f3b0464bc425f3d84661489d10d76.

README.md
```diff
@@ -54,3 +54,24 @@ For documentation see the
 developer version or the most recent released
 <a href="https://scitools-iris.readthedocs.io/en/stable/">stable</a> version.
 </p>
+
+## [#ShowYourStripes](https://showyourstripes.info/s/globe)
+
+<h4 align="center">
+  <a href="https://showyourstripes.info/s/globe">
+    <img src="https://raw.githubusercontent.com/ed-hawkins/show-your-stripes/master/2021/GLOBE---1850-2021-MO.png"
```
bjlittle, line 62: We can update this as and when new yearly (?) images are published @ https://github.com/ed-hawkins/show-your-stripes

bjlittle:
Also see https://github.com/ed-hawkins/show-your-stripes/issues/1

README.md
trexfeathers: Having recently written out the steps it takes to do a release (#5134), I'd rather leave this out for the sake of brevity!

This still has the appropriate impact even if it isn't the latest, and whenever we have a moment it will be easy to update.

README.md
trexfeathers: OK, if the blurring isn't a problem for you then I'm fine with it too.

(I did some reading about cropping in HTML but it needs CSS and that's not something that works in GitHub Markdown. SVG would also avoid the problem but I don't want to create more work for Prof. Hawkins!)

trexfeathers (approved):

README.md
bjlittle: SVG would be ideal, but i'm not really seeing any blurring that's obviously detrimental or a blocker

ed-hawkins:
Good idea to generate and tag releases on ShowYourStripes. Will do that this year when we update everything, probably in the next few weeks. Thanks.

Issue #4798, opened 2022-06-14
Iris Core + Plugins


trexfeathers:
Separating Iris API into 'core' API and multiple 'plugins' is a popular ambition offline, so deserves to be recorded publicly, especially for voting purposes 😉 .

Benefits:

* Smaller package(s)
* Quicker to install - fewer core dependencies
* Can run on Iris on smaller machines e.g. AWS lambda
* Faster testing - plugins (with accompanying dependencies) get their own separate testing routines
* Clearer development philosophy

Examples of elements that could be plugins:

* Plotting
* Regridding
* File loaders/savers

rcomer:
There are various small packages out there that do specific things that are useful for scientists.  For example, I am currently having a go with [xeofs](https://github.com/nicrie/xeofs) which currently has interfaces for pandas, xarray and numpy types.  If one were to think about adding an Iris interface to such a package, it would be useful (for testing) to have an Iris flavour that knows about the cube class and probably not much else.

pp-mo:
@HGWright pointed out : The big cost of this, not just the effort of doing it but the ongoing extra repo maintenance (e.g. multiple dependabot hits)

trexfeathers:
I am still in favour of establishing a core 'concept' - defining Iris core determines Iris' core dependencies - but I don't think plugins would be worth the hassle. We can instead cut the Conda recipe down to the core dependencies and introduce errors/warnings for when a user needs a core dependency but has not yet installed it.

trexfeathers:
An important consideration:

- #5936

trexfeathers:
Something not well captured above:

Iris-grib and iris-esmf-regrid have benefitted from being separated from the release cadence of Iris itself. This may be an opportunity for other areas of Iris too.

If any part of Iris is unlikely to reap this benefit (e.g. our plotting capabilities do not evolve very frequently) then I am still keen to explore optional Conda dependencies rather than full blown repo separation.

PR #5144, opened 2023-02-01
Plugin support

bsherratt:
## 🚀 Pull Request

Plugins! cf #4798 and the older proof of concept #3657.

It uses the same approach of namespace packages, but as `iris.plugins` instead of limiting ourselves to `iris.io.plugins`, and does not try to discover which plugins are available (should it..?)

This has been successfully used to separate out the NAME loader, although that will be a separate PR once we're happy with the approach to plugins.

---

### DIFF
--- a/docs/src/community/index.rst
+++ b/docs/src/community/index.rst
@@ -1,4 +1,5 @@
 .. include:: ../common_links.inc
+.. _namespace package: https://packaging.python.org/en/latest/guides/packaging-namespace-packages/
 
 .. todo:
     consider scientific-python.org
@@ -46,3 +47,23 @@ smoother interoperability:
    :hidden:
 
    iris_xarray
+
+Plugins
+-------
+
+Iris supports *plugins* under the `iris.plugins` `namespace package`_.  This
+allows packages that extend Iris' functionality to be developed and maintained
+independently, while still being installed into `iris.plugins` instead of a
+separate package.  For example, a plugin may provide loaders or savers for
+additional file formats, or alternative visualisation methods.
+
+Once a plugin is installed, it can be used either via the :func:`iris.use_plugin`
+function, or by importing it directly:
+
+.. code-block:: python
+
+    import iris
+
+    iris.use_plugin("my_plugin")
+    # OR
+    import iris.plugins.my_plugin
--- a/docs/src/community/index.rst
+++ b/docs/src/community/index.rst
@@ -1,5 +1,4 @@
 .. include:: ../common_links.inc
-.. _namespace package: https://packaging.python.org/en/latest/guides/packaging-namespace-packages/
 
 .. todo:
     consider scientific-python.org
@@ -51,19 +50,9 @@ smoother interoperability:
 Plugins
 -------
 
-Iris supports *plugins* under the `iris.plugins` `namespace package`_.  This
-allows packages that extend Iris' functionality to be developed and maintained
-independently, while still being installed into `iris.plugins` instead of a
-separate package.  For example, a plugin may provide loaders or savers for
-additional file formats, or alternative visualisation methods.
+Iris can be extended with **plugins**!  See below for further information:
 
-Once a plugin is installed, it can be used either via the :func:`iris.use_plugin`
-function, or by importing it directly:
+.. toctree::
+   :maxdepth: 2
 
-.. code-block:: python
-
-    import iris
-
-    iris.use_plugin("my_plugin")
-    # OR
-    import iris.plugins.my_plugin
+   plugins
--- a/lib/iris/__init__.py
+++ b/lib/iris/__init__.py
@@ -478,6 +478,14 @@ def use_plugin(plugin_name):
     """
     Convenience function to import a plugin
 
+    For example::
+
+        use_plugin("my_plugin")
+
+    is equivalent to::
+
+        import iris.plugins.my_plugin
+
     This is useful for plugins that are not used directly, but instead do all
     their setup on import.  In this case, style checkers would not know the
     significance of the import statement and warn that it is an unused import.
--- a/lib/iris/__init__.py
+++ b/lib/iris/__init__.py
@@ -91,6 +91,7 @@ All the load functions share very similar arguments:
 
 import contextlib
 import glob
+import importlib
 import itertools
 import os.path
 import pathlib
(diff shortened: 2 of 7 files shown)


lib/iris/plugins/README.md
```diff
@@ -0,0 +1,4 @@
+# Iris plugins
```
trexfeathers, line 1: Could this also include a link to the Iris documentation page/section on plugins?

lib/iris/__init__.py
```diff
@@ -471,3 +472,14 @@ def sample_data_path(*path_to_join):
             "appropriate for general file access.".format(target)
         )
     return target
+
+
+def use_plugin(plugin_name):
+    """
+    Convenience function to import a plugin
+
+    This is useful for plugins that are not used directly, but instead do all
```
trexfeathers, line 481:
```suggestion

    e.g. ``use_plugin("my_plugin")`` - equivalent to 
    ``from iris.plugins import my_plugin``
    
    This is useful for plugins that are not used directly, but instead do all
```

lib/iris/__init__.py
```diff
@@ -471,3 +472,14 @@ def sample_data_path(*path_to_join):
             "appropriate for general file access.".format(target)
         )
     return target
+
+
+def use_plugin(plugin_name):
```
trexfeathers, line 477: This needs including in [`__all__`](https://github.com/SciTools/iris/blob/b08cfa63bb435566546d9deb52985ea4561cf391/lib/iris/__init__.py#L118-L132) before it will be listed in the documentation.

docs/src/community/index.rst
```diff
@@ -46,3 +47,23 @@ smoother interoperability:
...
+Once a plugin is installed, it can be used either via the :func:`iris.use_plugin`
+function, or by importing it directly:
+
+.. code-block:: python
+
+    import iris
+
+    iris.use_plugin("my_plugin")
+    # OR
+    import iris.plugins.my_plugin
```
trexfeathers, line 69: Forgive my ignorance: will the current syntax definitely work? I'm struggling to map the namespace case onto my existing knowledge about which import syntax will work.

```suggestion
    from iris.plugins import my_plugin
```

docs/src/community/index.rst
```diff
@@ -46,3 +47,23 @@ smoother interoperability:
    :hidden:
 
    iris_xarray
+
+Plugins
+-------
+
+Iris supports *plugins* under the `iris.plugins` `namespace package`_.  This
```
trexfeathers, line 54:
```suggestion
Iris supports **plugins** under the `iris.plugins` `namespace package`_.  This
```

docs/src/community/index.rst
```diff
@@ -46,3 +47,23 @@ smoother interoperability:
...
+allows packages that extend Iris' functionality to be developed and maintained
+independently, while still being installed into `iris.plugins` instead of a
```
trexfeathers, line 56:
```suggestion

Iris supports *plugins* under the ``iris.plugins`` `namespace package`_.  This
allows packages that extend Iris' functionality to be developed and maintained
independently, while still being installed into ``iris.plugins`` instead of a
```

Needs double-ticks to appear as monospace.

---

I'm assuming it would be an anti-pattern to create `iris/plugins/__init__.py`, because that would allow:

```suggestion

Iris supports *plugins* under the :mod:`iris.plugins` `namespace package`_.  This
allows packages that extend Iris' functionality to be developed and maintained
independently, while still being installed into :mod:`iris.plugins` instead of a
```

docs/src/community/index.rst
```diff
@@ -46,3 +47,23 @@ smoother interoperability:
    :hidden:
 
    iris_xarray
+
+Plugins
+-------
```
trexfeathers, line 52: Please could this be a new `docs/src/community/plugins.rst`? I can picture the page expanding to list plugins that we know about, and provide instruction on how to create a plugin.

Would still be good to have the heading in this file (`community/index.rst`), but just with a single sentence and a link.

trexfeathers (changes requested):
Thanks so much for this @bsherratt, very exciting!

A few documentation suggestions for you.

lib/iris/__init__.py
trexfeathers: Or ``import iris.plugins.my_plugin`` if you think more appropriate.

docs/src/community/index.rst
bsherratt: I always forget that rst wants double backticks...

Adding `__init__.py` is not just an anti-pattern, it would mean it's no longer a namespace package.  Perhaps there is a way to have a page for the module in the API documentation anyway?

lib/iris/plugins/README.md
trexfeathers: I think it's appropriate to link to `latest`, since this README is for developers and we could be actively developing this section.

Appreciate that the link will be broken until merged, but I'm comfortable with that.

docs/src/community/index.rst
trexfeathers: Would probably involve modifying [`generate_package_rst.py`](https://github.com/SciTools/iris/blob/b08cfa63bb435566546d9deb52985ea4561cf391/docs/src/sphinxext/generate_package_rst.py), since we haven't yet migrated to `sphinx-apidoc`. Not worth it IMO - stick with the first option.

docs/src/community/index.rst
bsherratt: Went ahead and included a section on creating a plugin straight away, because it seemed a bit bare without.  Granted there may be more to add once I've gone through the whole process, but I hope it's a good start.

trexfeathers: It's looking great, thanks 🙂

trexfeathers:
I expect this will need #5153 before the CI will pass.

trexfeathers (approved):

trexfeathers:
I'm good for this to me merged, but I know @pp-mo wanted a look first

pp-mo:
> I'm good for this to be merged, but I know @pp-mo wanted a look first

Thanks for the reminder.
I've now looked it over and I must say :+1: ❤️ 😁 all round 
Please merge !!

pp-mo:
> Please merge !!
( Once there is a whatsnew item ! )

bsherratt:
> Once there is a whatsnew item

Now added, including an entry in `common_links`, because actually I was convinced that I was already there from previous contributions but apparently not :shrug:

PR #5159, opened 2023-02-15
Expand scope of common contributor links

rcomer:
### Description

This list is not restricted to core devs, so update the comment to reflect that.

---


pp-mo (approved):
Looks nice, but NB this doesn't actually get rendered anywhere, so visible for devs only !

PR #5170, opened 2023-02-21
`v3.4.1` 'merge-back' - What's New changes only

trexfeathers:
# Use a merge commit (not a squash)

The only _material_ difference between `v3.4.x` and `main` is the `v3.4.1` patch section on the What's New. Everything else is just what it took to port #5095 across, get the CI passing, and resolve conflicts, so I haven't included any of that complication here.


stephenworsley (approved):
Looks good.

PR #5178, opened 2023-02-27
Decorator to speed up operations on all-`False` masked arrays

### LINKED ISSUE #3470: Masked array performance
Given https://github.com/SciTools/iris/pull/3301 and https://github.com/SciTools-incubator/iris-agg-regrid/pull/37 we should consider whether masked arrays with no masked points should be automatically converted to ordinary `ndarray`s for obvious performance benefits.

Note that `python-netcdf4` now automatically always returns masked arrays.

This could be considered as a breaking change in `iris` behaviour, and so puts it plum in `3.0.0` territory.

----
As a user of iris, I want comparable regridding performance when using masked array data where no points are masked (i.e. mask=False) and non-masked arrays.  I would also like to have the ability to apply this functionality to other areas of my code if I deem it necessary.

I would like the capacility to be avilable to users to apply else where in their code.

**Acceptance Criteria**
- [ ] Create a function/decorator that can be applied to my iris function to give equivalent performance if my cube data is masked, but the mask is False.
- [ ] The output cube data should remain masked if it was originally masked.  Remain unmasked if originally unmasked.
- [ ] The function/decorator is applied to appropriate iris regridding methods.
- [ ] Have an ASV benchmark to prove the benefits of the solution.
- [ ] The new functionality should be in the what's new.
- [ ] The new functionality is documented in the user guide.

trexfeathers:
### Description

Closes #3470 - an attempt to provide a generic decorator that can use non-masked arrays in place of any inputs that have all-`False` masked arrays.

I have tested this against the case detailed here: https://github.com/SciTools/iris/issues/3470#issuecomment-1419408329, and it delivers the desired 1 second performance in all cases.

### To do

- [x] More manual experimenting
- [ ] Apply `@lift_empty_masks` to any applicable callables in Iris
- [ ] Finalise code (several `# TODO: ` comments)
- [ ] Tests
- [ ] Benchmarks
- [ ] What's New

---


lib/iris/analysis/__init__.py
```diff
@@ -3133,3 +3136,192 @@ def regridder(self, src_grid, target_grid):
...
+                    if mask_types_args[ix] == uniform_false_mtype:
+                        args[ix] = lift_mask(arg, cube_store)
+                for key, kwarg in kwargs.items():
+                    if mask_types_kwargs[key] == uniform_false_mtype:
+                        kwargs[key] = lift_mask(kwarg, cube_store)
+
+        result = decorated_func(*args, **kwargs)
+
+        if masks_lifted:
+            for cube in cube_store:
```
stephenworsley, line 3309: Is there a way to format this like a context manager or a try, except, finally statement so that cubes remain the same even if this fails partway through? I think this would only affect interactive sessions but it's probably still worth doing.

trexfeathers: Great idea, I have done that in the refactored version.

lib/iris/util.py
```diff
@@ -2025,3 +2052,297 @@ def _strip_metadata_from_dims(cube, dims):
...
+                    mask_type = cls.MaskTypes.ARRAY_FALSE
+        else:
+            mask_type = cls.MaskTypes.NOT_APPLICABLE
+
+        fill_value = getattr(sample, "fill_value", None)
+
+        return mask_type, fill_value
+
+    def __init__(self, func_name: str, *func_args, **func_kwargs):
+        from iris.cube import CubeList
```
stephenworsley, line 2133: I'm used to seeing the `__init__` at the top of a class, I don't believe there is a PEP prescribing method order, but this stackoverflow answer gives a good rationale for this ordering https://stackoverflow.com/a/36967648.

trexfeathers: Sounds good, looks like I got my `classmethod` and `staticmethod` ordering the wrong way round

lib/iris/util.py
```diff
@@ -2025,3 +2052,297 @@ def _strip_metadata_from_dims(cube, dims):
...
+        can significantly improve performance. Any ``args`` or ``kwargs`` of
+        `decorated_func` that are :obj:`~numpy.typing.ArrayLike`\\s or
+        :class:`~iris.cube.Cube`\\s, and have all-``False`` masks,
+        are converted;
+        :class:`~iris.cube.Cube`\\s are modified in-place. After
+        `decorated_func`
+        has run, all returned :obj:`~numpy.typing.ArrayLike`\\s and
+        :class:`~iris.cube.Cube`\\s, together any modified input
+        :class:`~iris.cube.Cube`\\s (see above), are converted to use masked
+        arrays with the original ``False`` mask re-applied.
```
stephenworsley, line 2280: replace "``False``" with "all-``False``"

lib/iris/util.py
```diff
@@ -2025,3 +2052,297 @@ def _strip_metadata_from_dims(cube, dims):
...
+        -----
+        If inconsistent of False-mask-types or fill_values are present (see Notes).
+
+        Warnings
+        --------
+        Be careful what callables this is applied to. If any input objects
+        have an
+        all-``False`` mask, **all** returned non-masked objects will be
+        converted
+        to have an all-``False`` mask. And if the callable assigns
```
stephenworsley, line 2301: Is this true? My understanding was that returned objects will become masked only if *all* array/cube-like objects have a mask (of the same type and fill value).

lib/iris/util.py
```diff
@@ -2025,3 +2052,297 @@ def _strip_metadata_from_dims(cube, dims):
...
+        :obj:`~numpy.typing.ArrayLike`\\s/:class:`~iris.cube.Cube`\\s
+        in a way that can't be detected (e.g. modifying a global object, or
```
stephenworsley, line 2303: It might be worth clarifying precisely what *can* be detected. I believe this is anything returned directly or returned as part of a tuple (correct me if I'm mistaken).

lib/iris/util.py
```diff
@@ -2025,3 +2052,297 @@ def _strip_metadata_from_dims(cube, dims):
...
+        -----
+        It is not possible to map specific properties from the inputs of
+        `decorated_func` to its outputs - could be one-to-many,
+        many-to-one, or many-to-many. Therefore, mask-lifting will only
+        take place if all the :obj:`~numpy.typing.ArrayLike`\\s and
+        :class:`~iris.cube.Cube`\\s have uniform ``fill_values`` and
+        ``False``-mask-types (2 types may exist - a scalar
+        :obj:`numpy.ma.nomask`, or a full :obj:`~numpy.typing.ArrayLike` of
+        ``False`` values). The uniform properties detected in ``args`` and
+        ``kwargs`` are those that are applied after `decorated_func` has run.
```
stephenworsley, line 2316: applied to what?

stephenworsley (changes requested):
Looking good! A lot of this seems quite fiddly and I spent some time thinking if there were any way to simplify it but I think the complexity is justified here. Just a couple suggestions here which ought to improve the readability of the code.

lib/iris/util.py
trexfeathers: c3528b560

lib/iris/util.py
```diff
@@ -2025,3 +2052,297 @@ def _strip_metadata_from_dims(cube, dims):
...
+            )
+            warn(warn_message_template.format(warn_phrase))
+
+        for address in all_addresses:
+            func_input = self.inputs_get_set(address)
+            mask_type, fill_value = self.get_mask_info(func_input)
+            if mask_type in self.false_mask_types:
+                if len(self.lift_addresses) == 0:
+                    self.uniform_false_mtype = mask_type
+                    self.uniform_fill_value = fill_value
```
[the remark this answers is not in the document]
trexfeathers: 3ae973ab4

lib/iris/analysis/_regrid.py
```diff
@@ -909,6 +912,7 @@ def __call__(self, src):
             in the horizontal dimensions will be combined before regridding.
 
         """
+        # TODO: consider refactoring to allow use of @iris.util.lift_empty_masks.
```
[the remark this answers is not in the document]
trexfeathers: I did actually want to decorate this method, since it's doing a lot of heavy lifting.

I've added some more explanation to the `# TODO`: 76f7b1400

It would perhaps be even better to refactor to be more like the curvilinear regridder, but that isn't what I was suggesting in this case.

lib/iris/util.py
```diff
@@ -2025,3 +2052,297 @@ def _strip_metadata_from_dims(cube, dims):
...
+        :obj:`~numpy.typing.ArrayLike`\\s
+        (e.g. :class:`numpy.ndarray` / :class:`dask.array.Array`)
+        can significantly improve performance. Any ``args`` or ``kwargs`` of
+        `decorated_func` that are :obj:`~numpy.typing.ArrayLike`\\s or
+        :class:`~iris.cube.Cube`\\s, and have all-``False`` masks,
+        are converted;
+        :class:`~iris.cube.Cube`\\s are modified in-place. After
+        `decorated_func`
+        has run, all returned :obj:`~numpy.typing.ArrayLike`\\s and
+        :class:`~iris.cube.Cube`\\s, together any modified input
```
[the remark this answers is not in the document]
trexfeathers: 1f99be5b8

lib/iris/util.py
trexfeathers: 1f99be5b8

lib/iris/util.py
trexfeathers: 1f99be5b8

lib/iris/util.py
```diff
@@ -2025,3 +2052,297 @@ def _strip_metadata_from_dims(cube, dims):
...
+                decorated_func.__name__, *args, **kwargs
+            )
+            mask_lifter.lift_masks()
+            try:
+                result = decorated_func(
+                    *mask_lifter.func_args, **mask_lifter.func_kwargs
+                )
+            finally:
+                mask_lifter.restore_cube_masks()
+            return mask_lifter.re_apply_masks(result)
```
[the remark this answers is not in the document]
trexfeathers: 66a6dbbdb

lib/iris/util.py
trexfeathers: 6e0e7317e

stephenworsley (changes requested):
Looking good, just a couple suggestions for the tests.

lib/iris/util.py
stephenworsley: Okay, I've looked through the code again and I'm convinced the docstring is correct here.

trexfeathers:
Brief note from conversation with @stephenworsley:

Benchmarks prove that it can be detrimental to use this in the wrong places. Lack of benchmark coverage in other areas means we haven't proved its value outside of the agg-regrid toy example (https://github.com/SciTools/iris/issues/3470#issuecomment-1419408329). So the decorator is pretty much ready to go, but it should not be used without a benchmark proving it improves performance, so that will need subsequent PR's.

This PR will provide the decorator, and some sort of proof that it is worth using in the right cases. But it will not apply the decorator anywhere.

lib/iris/tests/unit/util/test_lift_empty_masks.py
```diff
@@ -0,0 +1,357 @@
...
+            func = da.ma.masked_array
+        else:
+            func = np.ma.masked_array
+        array = func([1, 2], mask=mask, fill_value=fill_value)
+    else:
+        if dask:
+            func = da.array
+        else:
+            func = np.array
+        array = func([1, 2])
```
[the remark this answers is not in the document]
trexfeathers: ade5981fdc1dba1d7a624c2174cab2f5b738f6d5

lib/iris/tests/unit/util/test_lift_empty_masks.py
```diff
@@ -0,0 +1,357 @@
...
+    @staticmethod
+    @lift_empty_masks
+    def input_checker(*args, **kwargs):
+        def get_has_mask(input_):
+            if isinstance(input_, Cube):
+                input_ = input_.core_data()
+            return has_mask(input_)
+
+        result_args = [get_has_mask(a) for a in args]
+        result_kwargs = [get_has_mask(v) for v in kwargs.values()]
```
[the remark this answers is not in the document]
trexfeathers: 31dd47d6dd5fdb9207df2ae51403f3e8076870dc

lib/iris/util.py
```diff
@@ -7,28 +7,353 @@
...
+        fill_value = getattr(sample, "fill_value", None)
+
+        return mask_type, fill_value
+
+    @staticmethod
+    def is_cube(array_or_cube):
+        """Duck-typing-style check."""
+        # Can't use hasattr as that runs Cube.data .
+        dir_result = dir(array_or_cube)
+        return "core_data" in dir_result and "data" in dir_result
```
trexfeathers, line 239: Changed since last review

lib/iris/util.py
```diff
@@ -7,28 +7,353 @@
...
+
+    @classmethod
+    def get_data(cls, array_or_cube):
+        if cls.is_cube(array_or_cube):
+            result = array_or_cube.core_data()
+        elif (
+            hasattr(array_or_cube, "__array__")
+            and array_or_cube.shape != tuple()
+        ):
+            # Generic check for NumPy, Dask and masked arrays, excluding scalars.
```
trexfeathers, line 202: Changed since last review

trexfeathers:
Results from the new benchmarks:

```
ON_DEMAND_BENCHMARKS=True python bm_runner.py custom run HEAD^! --bench LiftEmptyMasks --show-stderr

[ 75.00%] ··· util.LiftEmptyMasks.time_many_access                                                                                                                                                                                    ok
[ 75.00%] ··· ================= ============ =============
              --                    scalar false mask     
              ----------------- --------------------------
               lifting enabled     False          True    
              ================= ============ =============
                    False        6.20±0.1ms   5.68±0.05ms 
                     True         584±50μs      508±50μs  
              ================= ============ =============

[100.00%] ··· util.LiftEmptyMasks.time_mean                                                                                                                                                                                           ok
[100.00%] ··· ================= ============= =============
              --                     scalar false mask     
              ----------------- ---------------------------
               lifting enabled      False          True    
              ================= ============= =============
                    False        2.18±0.08ms     707±30μs  
                     True        1.50±0.05ms   1.28±0.07ms 
              ================= ============= =============

```

trexfeathers:
Masking clearly slows things down, however a magic decorator solution is looking increasingly questionable. It sometimes is faster, and sometimes slower. And @pp-mo believes it is only fair to inspect the contents of a 'lazy file' if it can be done in a `map_blocks()` fashion - since otherwise we have to access the whole file.

It may be that each of the specific cases needs specific mask intelligence (like some of our existing regridders), rather than a global easy solution.

Not sure what everyone else thinks, but it wouldn't hurt my feelings to call this a thorough investigation that ends in no action.

pp-mo (changes requested):
As comments -- not sure there isn't a big problem with this.

But I need to review -- possibly may have misunderstood

lib/iris/_lazy_data.py
```diff
@@ -419,3 +419,21 @@ def map_complete_blocks(src, func, dims, out_sizes):
...
+    Intended for checking masks and fill values - dask.array.ma arrays have no
+    distinguishing properties from normal dask arrays, so must unfortunately
+    realise a small section first.
+    """
+    assert is_lazy_data(array)
+
+    if array.shape == tuple():
+        to_concrete = array
+    else:
+        to_concrete = array.blocks[0]
```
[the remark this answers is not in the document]
pp-mo: Just for what it's worth, I checked out how to demonstrate this idea, here :
```
>>> import numpy as np
>>> import dask.array as da
>>> class MyArraylike:
...     def __init__(self, content):
...         self._content = content
...         # Provide minimum array API required for dask.from_array
...         self.ndim = self._content.ndim
...         self.shape = self._content.shape
...         self.dtype = self._content.dtype
...     
...     def __getitem__(self, keys):
...         print('accessed: ', keys)
...         return self._content[keys]
... 
>>> mya = MyArraylike(np.arange(100))
>>> lazy = da.from_array(mya, chunks=10)
accessed:  (slice(0, 0, None),)
>>> print(lazy)
dask.array<array, shape=(100,), dtype=int64, chunksize=(10,), chunktype=numpy.ndarray>
>>> 
>>> chunk0 = lazy[:10]
>>> print(chunk0.compute())
accessed:  (slice(0, 10, None),)
[0 1 2 3 4 5 6 7 8 9]
>>> 
>>> chunk_part = lazy[22:24]
>>> print(chunk_part.compute())
accessed:  (slice(22, 24, None),)
[22 23]
>>> 
>>> cross_chunk = lazy[8:14]
>>> print(cross_chunk.compute())
accessed:  (slice(8, 10, None),)
accessed:  (slice(10, 14, None),)
[ 8  9 10 11 12 13]
>>> 

This shows how dask can+does fetch sub-sections of chunks.
```

lib/iris/_lazy_data.py
```diff
@@ -419,3 +419,21 @@ def map_complete_blocks(src, func, dims, out_sizes):
...
 
     return data.map_blocks(func, chunks=out_chunks, dtype=src.dtype)
+
+
+def sample(array):
+    """
+    Realise and return the smallest possible section of a lazy array.
+
+    Intended for checking masks and fill values - dask.array.ma arrays have no
+    distinguishing properties from normal dask arrays, so must unfortunately
```
bouweandela, line 429: Note that dask arrays seem to keep track of the array type of their chunks, I think there may be no need to sample. This code will give you the array type:
```python
type(dask.array.utils.meta_from_array(array))
```

trexfeathers:
I don't think there is merit in keeping this open. It was an interesting proof AGAINST the concept!

PR #5181, opened 2023-02-27
removed cell measure mask check and error

### LINKED ISSUE #5147: saving cubes with masked cell measures
Hey guys, how's it going? We're hitting an issue with saving cubes that have masked cell measures attached to the main variable, see https://github.com/ESMValGroup/ESMValCore/issues/1917 - I don't really understand the issue with this, so long as one supplies the saver with a valid `fill_value` (like we do, the infamous 1e20) - is this something that's a bit legacy or you're still afraid of it? Cheers muchly :beer:

ESadek-MO:
## 🚀 Pull Request
Closes #5147.

### Description
Removed check and error raise for saving cubes with masked cell measures.

### DIFF
--- a/lib/iris/tests/unit/fileformats/netcdf/test_Saver.py
+++ b/lib/iris/tests/unit/fileformats/netcdf/test_Saver.py
@@ -1051,30 +1051,6 @@ class Test__create_cf_grid_mapping(tests.IrisTest):
         self._test(coord_system, expected)
 
 
-class Test__create_cf_cell_measure_variable(tests.IrisTest):
-    # Saving of masked data is disallowed.
-
-    # Attribute is substituted in test_Saver__lazy.
-    array_lib = np
-
-    def setUp(self):
-        self.cube = stock.lat_lon_cube()
-        self.names_map = ["latitude", "longitude"]
-        masked_array = self.array_lib.ma.masked_array(
-            [0, 1, 2], mask=[True, False, True]
-        )
-        self.cm = iris.coords.CellMeasure(masked_array, var_name="cell_area")
-        self.cube.add_cell_measure(self.cm, data_dims=0)
-        self.exp_emsg = "Cell measures with missing data are not supported."
-
-    def test_masked_data__insitu(self):
-        # Test that the error is not being raised.
-        with self.temp_filename(".nc") as nc_path:
-            saver = Saver(nc_path, "NETCDF4")
-            saver._create_generic_cf_array_var(
-                self.cube, self.names_map, self.cm
-                )
-
 
 if __name__ == "__main__":
     tests.main()
--- a/lib/iris/tests/unit/fileformats/netcdf/test_Saver.py
+++ b/lib/iris/tests/unit/fileformats/netcdf/test_Saver.py
@@ -1050,7 +1050,5 @@ class Test__create_cf_grid_mapping(tests.IrisTest):
         }
         self._test(coord_system, expected)
 
-
-
 if __name__ == "__main__":
     tests.main()
--- a/lib/iris/tests/unit/fileformats/netcdf/test_Saver.py
+++ b/lib/iris/tests/unit/fileformats/netcdf/test_Saver.py
@@ -1050,5 +1050,6 @@ class Test__create_cf_grid_mapping(tests.IrisTest):
         }
         self._test(coord_system, expected)
 
+
 if __name__ == "__main__":
     tests.main()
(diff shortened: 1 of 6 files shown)


pp-mo:
Hi @ESadek-MO 
Note that, if you have git pre-commit enabled on your local repo clone, then style errors like https://github.com/SciTools/iris/pull/5181/commits/2498f9b1664ae964b31c90f8b2ade32081b2bd08 should not occur, because it will be fixed whenever you make a commit.

For this, you need to run from an env containing 'pre-commit', and you need (just once) to run "pre-commit install" within you local repo clone (i.e. the checkout on disk).  Then it runs black via the [`.pre-commit-config.yaml`](https://github.com/SciTools/iris/blob/v3.4.1/.pre-commit-config.yaml)

There is some written explanation here : https://scitools-iris.readthedocs.io/en/stable/developers_guide/contributing_code_formatting.html?highlight=pre-commit

Sorry if you already know all this !!

lib/iris/tests/unit/fileformats/netcdf/test_Saver.py
```diff
@@ -1068,21 +1068,13 @@ def setUp(self):
...
         with self.temp_filename(".nc") as nc_path:
             saver = Saver(nc_path, "NETCDF4")
-            with self.assertRaisesRegex(ValueError, self.exp_emsg):
-                saver._create_generic_cf_array_var(
-                    self.cube, self.names_map, self.cm
+            saver._create_generic_cf_array_var(
+                self.cube, self.names_map, self.cm
                 )
 
-    def test_masked_data__save_pipeline(self):
```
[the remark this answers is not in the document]
pp-mo: :+1:

lib/iris/tests/unit/fileformats/netcdf/test_Saver.py
```diff
@@ -1068,20 +1068,12 @@ def setUp(self):
         self.exp_emsg = "Cell measures with missing data are not supported."
```
pp-mo, line 1068: `self.exp_emsg` is no longer needed

lib/iris/tests/unit/fileformats/netcdf/test_Saver.py
```diff
@@ -1068,20 +1068,12 @@ def setUp(self):
...
-                    self.cube, self.names_map, self.cm
-                )
-
-    def test_masked_data__save_pipeline(self):
-        # Test that the right error is raised by the saver pipeline.
-        with self.temp_filename(".nc") as nc_path:
-            with Saver(nc_path, "NETCDF4") as saver:
-                with self.assertRaisesRegex(ValueError, self.exp_emsg):
-                    saver.write(self.cube)
+            saver._create_generic_cf_array_var(
```
pp-mo, line 1074: There is a problem with this, because the previous idea was to do the minimum setup to enable us to call this routine + have the check fail.
Unfortunately, to get the routine to **_succeed_** we must do rather more setup on the file first.
So I added these lines, which do at least make the call succeed ...
```
    # Do minimum of saver.write to let us to create the var successfully
    _, cube_dimensions = saver._get_dim_names(self.cube)
    saver._create_cf_dimensions(self.cube, cube_dimensions)
```

But then I think it's not totally clear what this should be testing **_for_**.  
My instinct was to add more, so I initially added some rather overkill asserts to check that the array is encoded as expected, like this ...
```
    # Check that data is written with masked points + default fill-value
    file_var = saver._dataset.variables['cell_area']
    assert '_FillValue' not in file_var.ncattrs()
    fill_value = nc.default_fillvals['i8']
    file_array = file_var[:]
    assert np.all(file_array.data == [fill_value, 1, fill_value])
    assert np.all(file_array.mask == [True, False, True])
```
But on closer inspection, this is actually what **any** masked save should be doing, and there is now nothing in it which is specific to cell-measures !!

So, it might seem more logical to test the operation of `saver._add_cell_measures`.  
And maybe all we really need to do is check that `_add_inner_related_vars` or `_create_generic_cf_array_var` is called by `_add_cell_measures`.
But unfortunately that requires even _more_ prior setup of the file before it can function ... and it still isn't actually testing anything specific to cell-measures.

Fundamentally, I think the 'setup problem'  is because the code itself is not structured to _enable_ testing isolated routines in this way.

Since cell-measures are basically now no different to anything else which gets passed into `_create_generic_cf_array_var`,
after much thought, I think I would just 
**remove this whole test !** -- does that sound crazy, or what ?

pp-mo (approved):
Ok, I think this is OK now, with nothing in place of the old test.

pp-mo:
Nice job @ESadek-MO . Sorry for all my fretting over the broken test !

PR #5184, opened 2023-03-01
Switch to tox

lbdreyer:
~WIP~ This is now ready for review

This switches our use of nox to tox, re #4806


.github/workflows/ci-tests.yml
```diff
@@ -91,10 +100,10 @@ jobs:
       - name: "conda environment cache"
         uses: ./.github/workflows/composite/conda-env-cache
         with:
-          cache_build: 0
+          cache_build: 1
           cache_period: ${{ env.CACHE_PERIOD }}
           env_name: ${{ env.ENV_NAME }}
-          install_packages: "cartopy nox pip"
+          install_packages: "cartopy tox pip"
```
bjlittle, line 106: @lbdreyer I think this should be `tox<4` atm as `tox-conda` isn't 4.x safe. Also `tox-wheel` hasn't been released since Oct 2022, so it's probably not 4.x compliant either, you'd have to check.

Note that `tox` 4.x was released in Dec 2022.

tox.ini
```diff
@@ -0,0 +1,37 @@
...
+    py38: {toxinidir}{/}requirements{/}locks{/}py38-linux-64.lock
+    py39: {toxinidir}{/}requirements{/}locks{/}py39-linux-64.lock
+    py310: {toxinidir}{/}requirements{/}locks{/}py310-linux-64.lock
+
+[testenv:py{38,39,310}-tests]
+description = Perform Iris unit, integration and system tests.
+usedevelop =
+    true
+commands =
+    pytest -n auto {posargs} {toxinidir}{/}lib{/}iris{/}tests
```
trexfeathers, line 16: I'm not clear from the documentation whether `{posargs}` is a special value to `tox`?

lbdreyer: See [here](https://tox.wiki/en/3.27.1/example/general.html#interactively-passing-positional-arguments)

so if you were to run:
```tox -e py310-tests --cov``

it will create the py3.10 environment, install a dev version of iris, then run:
```pytest -n -auo --cov ./lib/iris/tests```

File: tox.ini [hunk 3]
```diff
@@ -0,0 +1,35 @@
...
+requires =
+    tox-conda
+
+[testenv]
+conda_spec =
+    py38: {toxinidir}{/}requirements{/}locks{/}py38-linux-64.lock
+    py39: {toxinidir}{/}requirements{/}locks{/}py39-linux-64.lock
+    py310: {toxinidir}{/}requirements{/}locks{/}py310-linux-64.lock
+usedevelop =
+    true
```
lbdreyer, line 11: These are all inherited by the below test environments, although note that with the wheel job `skip_install` takes precedence over `usedevelop`.

tox.ini
```diff
@@ -0,0 +1,35 @@
...
+commands =
+    make clean html
+    linkcheck: make linkcheck
+    tests: make doctest
+
+[testenv:py{38,39,310}-wheel]
+description = Install wheel and test the Iris import
+skip_install = true
+deps =
+    {toxinidir}{/}dist{/}scitools_iris*.whl
```
lbdreyer, line 33: Note that with [nox](https://github.com/SciTools/iris/blob/main/noxfile.py#L288-L294) we caught the scenarios where there were no wheels and where there were more than one wheel. I could have added this checking for tox, I would have had to remove this `deps` section and instead set the commands to ensure it gets one wheel , then to manually pip install it.

But I felt this was more simple and the above was unnecessary complication.
I tested to see how tox would handle the wrong number of wheels:

1) Two wheels available (iris3.5 and iris3.7):
```
$ tox -e py310-wheel
py310-wheel create: .../iris/.tox/py310-wheel
py310-wheel installcondadeps: --file=.../iris/requirements/locks/py310-linux-64.lock
WARNING: could not determine version of: .../iris/dist/scitools_iris-3.5.0.dev51+dirty-py3-none-any.whl
WARNING: could not determine version of: .../iris/dist/scitools_iris-3.7.0.dev51+dirty-py3-none-any.whl
_______________________________________________________________ summary ________________________________________________________________
  py310-wheel: commands succeeded
  congratulations :)
...
tox.exception.MissingDependency: MissingDependency:.../iris/dist/scitools_iris*.whl
```

2) no wheels available
```
$ tox -e py310-wheel
py310-wheel create: .../iris/.tox/py310-wheel
py310-wheel installcondadeps: --file=.../iris/requirements/locks/py310-linux-64.lock
_______________________________________________________________ summary ________________________________________________________________
  py310-wheel: commands succeeded
  congratulations :)
...
tox.exception.MissingDependency: MissingDependency: .../iris/dist/scitools_iris*.whl
```

So my preferrence is to just rely on tox's error message

lbdreyer: Regarding the `WARNING: could not determine version of` in the scenario of two wheels present, those warnings don't appear when there is only one wheel:
```
]$ tox -e py310-wheel
py310-wheel create: .../iris/.tox/py310-wheel
py310-wheel installcondadeps: --file=.../iris/requirements/locks/py310-linux-64.lock
py310-wheel installdeps: .../iris/dist/scitools_iris-3.5.0.dev51+dirty-py3-none-any.whl
...
py310-wheel run-test-pre: PYTHONHASHSEED='1378407663'
py310-wheel run-test: commands[0] | python -c 'import iris; print(f'"'"'{iris.__version__=}'"'"')'
iris.__version__='3.5.0.dev51+dirty'
_______________________________________________________________ summary ________________________________________________________________
  py310-wheel: commands succeeded
  congratulations :)

```

tox.ini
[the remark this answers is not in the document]
lbdreyer: As the py*-tests job just pass the posargs to pytest you can run: `tox -e py310-tests -- --cov` and that will run `pytest -n -auto --cov ./lib/iris/tests` 

I prefer this as then the dev has control over coverage testing configuration, particularly you aren't forced to use xml, you could use html if you prefer

benchmarks/bm_runner.py
```diff
@@ -48,27 +48,27 @@ def _prep_data_gen_env() -> None:
...
         subprocess.run(
             [
-                "nox",
-                f"--noxfile={root_dir / 'noxfile.py'}",
-                "--session=tests",
-                "--install-only",
-                f"--python={python_version}",
+                "tox",
+                f"-c {root_dir / 'tox.ini'}",
+                f"-e py{python_version}",
```
lbdreyer, line 63: Nice thing about tox is that we get an environment, with a dev installation of iris, for free, because of the definition of:
https://github.com/SciTools/iris/blob/62bcb99d4aff5769f293413a3952ad7b8e580ac2/tox.ini#L5-L11
This allows you to run `tox -e py310` 


Alternatively tox has a `--notests` flag which will create the environment, but not run the commands, aka what nox's `--install-only` was doing

lib/iris/tests/test_coding_standards.py
```diff
@@ -71,8 +71,8 @@ def test_python_versions():
     This test is designed to fail whenever Iris' supported Python versions are
     updated, insisting that versions are updated EVERYWHERE in-sync.
     """
-    latest_supported = "3.10"
-    all_supported = ["3.8", "3.9", latest_supported]
+    latest_supported = "310"
+    all_supported = ["38", "39", latest_supported]
```
lbdreyer, line 75: We now have more usages of non-dotted versions (the requirements files, tox likes no dot) than dotted versions so it's more easy to have non-dotted here

tox.ini
```diff
@@ -0,0 +1,35 @@
...
+[testenv:py{38,39,310}-docs{,-linkcheck,-tests}]
+description = Build, and optionally linkcheck or test, the Iris documentation
+allowlist_externals =
+    make
+changedir =
+    {toxinidir}{/}docs
+commands =
+    make clean html
+    linkcheck: make linkcheck
+    tests: make doctest
```
lbdreyer, line 27: This results in the following py3.10 jobs:
py310-docs: simply build the documentation
py310-linkcheck: build the docs and run the link check
py310-tests: build the docs and run the tests

lbdreyer:
So this _should_ be all good to go, but there are now three test failures: 
```
FAILED lib/iris/tests/unit/analysis/stats/test_pearsonr.py::Test::test_perfect_corr
FAILED lib/iris/tests/integration/plot/test_vector_plots.py::TestBarbs::test_2d_plain_latlon
FAILED lib/iris/tests/integration/plot/test_vector_plots.py::TestBarbs::test_2d_plain_latlon_on_polar_map
==== 3 failed, 6767 passed, 36 skipped, 15220 warnings in 178.70s (0:02:58) ====
```

Which is interesting since I've not touched any test files. Maybe something to do with the environment creation!

tox.ini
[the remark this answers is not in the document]
trexfeathers: Fair enough, I can see the argument for that.

tox.ini
trexfeathers: I'm not especially worried about this. I expect Bill only originally included the defensive code because things like that are easier with Nox. So I'm comfortable with your proposal.

trexfeathers (changes requested):
This is coming together nicely! Couple of final observations.

lbdreyer:
This is getting closer. I just have to make change to the benchmarks and test them locally.

lbdreyer:
Benchmarks ran, so this should now be ready!

trexfeathers:
Since everything is in a single commit I think I'll need to go back through all the changes - I can't remember which bits I did and didn't review back in March.

So this will need to wait until after the sprint.

lbdreyer:
Thanks! no rush at all!

.github/workflows/ci-tests.yml
```diff
@@ -91,10 +91,10 @@ jobs:
       - name: "conda environment cache"
         uses: ./.github/workflows/composite/conda-env-cache
         with:
-          cache_build: 0
+          cache_build: 1
           cache_period: ${{ env.CACHE_PERIOD }}
           env_name: ${{ env.ENV_NAME }}
-          install_packages: "cartopy nox pip"
+          install_packages: "cartopy tox'<4'"
```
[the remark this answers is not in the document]
trexfeathers: [Do we need Nox OR Tox anymore, now that we use lock files?!](https://github.com/catalyst-cooperative/pudl/pull/2968)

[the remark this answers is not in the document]
trexfeathers: > [Do we need Nox OR Tox anymore, now that we use lock files?!](https://github.com/catalyst-cooperative/pudl/pull/2968)

Having thought more, that was ill-informed, since Tox still offers neat inheritance tricks. But if we stick with tox-conda, we need to be comfortable sticking with the pin as well. Of course being pinned kinda undermines the idea of adopting something more popular.

Thanks @ESadek-MO and @HGWright for conversation on this topic.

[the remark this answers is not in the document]
trexfeathers: Also leaves us to potential security vulnerabilities:

https://github.com/tox-dev/tox/issues/2524#issuecomment-1298934474

trexfeathers:
Looks like this didn't get recorded, but discussions between @SciTools/peloton in December concluded that we should investigate how easy it would be to maintain our own tox-conda plugin, with just the features that we need. @pp-mo expressed an interest in this, I don't know how it has progressed.

We accepted that this is less good than maintaining the real, public plugin. But we also agreed that it could be the pragmatic approach since a private plugin would take less maintenance and make us less vulnerable to outside forces.

@HGWright highlighted that if we got a proof-of-concept together then we could potentially start a conversation on the tox-conda repo to see how many others are maintaining their own plugin and discuss pooling resources.

This matters for this PR, as I am increasingly reluctant to approve tox-conda becoming a dependency.

PR #5186, opened 2023-03-06
Adopt codespell

bjlittle:
### Description

This PR automates the detection of ~speeling~ spelling mistakes within the code base, as part of our `pre-commit.ci` workflow using [codespell](https://github.com/codespell-project/codespell).

Also adds a `.git-blame-ignore-revs` to ensure that I don't get blamed for the original authors change, see [here](https://github.blog/changelog/2022-03-24-ignore-commits-in-the-blame-view-beta/) for further details.

---


.pre-commit-config.yaml
```diff
@@ -28,6 +28,13 @@ repos:
         # Don't commit to main branch.
     -   id: no-commit-to-branch
 
+-   repo: https://github.com/codespell-project/codespell
+    rev: "v2.2.2"
+    hooks:
+    -   id: codespell
+        types_or: [asciidoc, python, markdown, rst]
```
bjlittle, line 35: Determining the `types` can be de-risked by using the [identity](https://pre-commit.com/#filtering-files-with-types) CLI tool.

.git-blame-ignore-revs
```diff
@@ -0,0 +1 @@
+c178e0b482d65919a4c9b63e7b0e03974ec9b9f1
```
rcomer, line 1: Will this still work for `main` if the PR is squash-merged?

Tangent for a possible separate PR: can commits be added retrospectively to this file?  I’m thinking of the various `black` updates.

docs/src/whatsnew/latest.rst
```diff
@@ -36,7 +36,7 @@ This document explains the changes made to Iris for this release
    :ref:`documentation page<community_plugins>` for further information.
    (:pull:`5144`)
 
-#. `@rcomer`_ enabled lazy evaluation of :obj:`~iris.analysis.RMS` calcuations
+#. `@rcomer`_ enabled lazy evaluation of :obj:`~iris.analysis.RMS` calculations
```
rcomer, line 39: For shame! 🤦‍♀️

lib/iris/tests/test_basic_maths.py
```diff
@@ -589,9 +589,9 @@ def test_apply_ufunc(self):
             in_place=False,
         )
 
-        ans = a.data**2
+        answer = a.data**2
```
rcomer, line 592: Wait - can we only use proper words as variable names now?

lib/iris/tests/unit/coords/__init__.py
```diff
@@ -79,7 +79,7 @@ def arrays_share_data(a1, a2):
 
 
 def lazyness_string(data):
-    # Represent the lazyness of an array as a string.
+    # Represent the laziness of an array as a string.
```
rcomer, line 82: I’m not sure about this one: the property of data being lazy is arguably a different concept from just, y’know, idleness.  So maybe we are OK to create our own word?

lib/iris/tests/test_basic_maths.py
bjlittle: Nope, we can easily add exceptions... the spell-checker obviously tried to correct this to something like `ans->and` (and a bunch of other suggestions).

To be fair `ans` is kinda weak as a variable name. Something like `expected` might be better? Dunno.

Anyways, we could add `ans` as a legitimate word to ignore, but then we run the risk of missing a genuine typo when the author intended `and` and put `ans` instead.

I'm not fussed really, happy to go with whatever you suggest/recommend.

.git-blame-ignore-revs
bjlittle: Good point, in this case we could choose to 

1. not squash this PR
2. squash it and follow-up with a single commit pull-request with the sha of the previous squashed pull-request commit
3. ditch the use of `.git-blame-ignore-revs` as a concept

lib/iris/tests/unit/coords/__init__.py
bjlittle: Personally I think it's `laziness` or I guess we could:

1. invent a new word i.e., `lazyness`
2. keep the usage of `lazyness` as a word, but skip it for correction
3. opt for a compromise, such as `lazy-ness` (🤢)
4. change the language e.g., `Represent the lazy state of the array as a string.` i.e., always use `lazy` not `lazyness` nor `laziness`.

rcomer: I don’t have a strong opinion.  Though I note that `lazyness` is also used in the function name.  The function is public but undocumented…

.git-blame-ignore-revs
bjlittle: @rcomer Yup, okay, I'll hunt those commits down and look to add them.

I'm in total agreement with you, I'm interested in the code changes, not so much style.

lib/iris/tests/unit/coords/__init__.py
bjlittle: Seems like that horse has bolted... I'm just gonna back-out that change, and make `lazyness` a word to be ignored :+1:

bjlittle: Although the code is peppered with usage of `laziness`... TBH I'm past the point of caring about this one

rcomer: 🤷‍♀️

.git-blame-ignore-revs
rcomer: Might also be good to add comment so future us knows why each commit was skipped.
https://github.com/matplotlib/matplotlib/blob/main/.git-blame-ignore-revs

lib/iris/tests/unit/coords/__init__.py
bjlittle: See https://github.com/SciTools/iris/pull/5186/commits/25b8463ea62ee640333713a04d7511f2d1bf1b44

.git-blame-ignore-revs
bjlittle: Done. See [08cb0a5](https://github.com/SciTools/iris/pull/5186/commits/08cb0a5d4931026385579d889b1283d1c1a633df)

bjlittle:
This PR, by it's nature, is prone to go stale very quickly, which is fine.

I'm only going to resolve conflicts once we're happy to merge.

bjlittle:
Conflicts resolved 👍

tkknight (approved):

PR #5190, opened 2023-03-10
Adding a Dask best practices section to the user guide

HGWright:
### Description

I have brought the internal Dask best practice advice and examples into the Dask documentation. I have updated to change specific internal information to more generic language, more relevant in the documentation.

This should be linked to #4959. But does not fully close the issue.

---


lbdreyer (changes requested):
Looking good, thanks @HGWright !

A few general comments: 
* There are a few terms that need tidying up. I think we were quite lazy in the draft Dask best practices docs, but we need to be more careful here if this is to be included in the Iris docs, particularly, the following need to be update:
`numpy` -> `NumPy`
`netcdf` -> `netCDF` (I don't _believe_ this needs to be capitalised)
Also `dask` should always be capitalised `Dask`

* We have used `CPU's` in quite a few places, but the apostrophe is incorrect, so that should just be `CPUs`.

* You have used the term "multiprocessing system", but I think a term like "computing cluster" would be more appropriate.

* There are a few examples of MO specific sections that need generalising a bit more, I have added specific comments where this is required.

docs/src/further_topics/dask_best_practices/dask_bags_and_greed.rst
```diff
@@ -0,0 +1,235 @@
...
+    with dask.config.set(num_workers=num_workers):
+        bag = db.from_sequence(fpaths).map(func)
+        cubes = iris.cube.CubeList(bag.compute())
+
+This achieves a significant performance improvement - more than twice as
+fast as the previous benchmarks:
+
++---------------+-----------------------+---------------+---------------+-----------+
+| Machine       | CPUs Available       | CPUs Used    | Previous Time | New Time  |
++===============+=======================+===============+===============+===========+
```
lbdreyer, line 185:
```suggestion
+---------------+-----------------------+---------------+---------------+-----------+
| Machine       | CPUs Available        | CPUs Used     | Previous Time | New Time  |
+===============+=======================+===============+===============+===========+
```

pp-mo:
> * "This branch is out-of-date with the base branch"

This is not an error.
I think it is just something that GitHub has started offering as an option -- basically, an automatic merge-back from the target branch, or rebase onto it.
I don't really understand the benefit of these, if there are no conflicts, but possibly it enables you to see more exactly what the result will be when merged back (e.g. docs builds).

pp-mo:
How open are we to further modifying this now?
I'm just re-reading some of the content and clearly some things could be improved.  
Obviously this content is quite old now, and besides what may have changed, I think in some places our understanding may also have improved since it was written.

pp-mo:
> Are we intending this to go in Iris 3.5?

I had thought so, but I see it's not actually on the board.
@ESadek-MO can you explain ?

HGWright:
@pp-mo & @lbdreyer. Given that I have lost some momentum with this, my preference would be to bank this and then I will open a new issue to make improvements as this is already quite a big PR. Then at least the information is out there.

I think this should be good to go if that's what we are doing.

docs/src/whatsnew/latest.rst
```diff
@@ -71,6 +71,10 @@ This document explains the changes made to Iris for this release
    to use it.  By default the theme will be based on the users system settings,
    defaulting to ``light`` if no system setting is found.  (:pull:`5299`)
 
+#. `@HGWright`_ added a :doc:`/further_topics/dask_best_practices/index`
+   section into the user guide, containing advice and use cases to help users
+   get the best out of Dask with IRIS.
```
lbdreyer, line 76:
```suggestion
   get the best out of Dask with Iris.
```

lbdreyer (changes requested):
Looks good to me, just one final small change and then this should be ready to merge. Could you also create an issue to capture the extra work that @pp-mo suggests?

lbdreyer (approved):

lbdreyer:
Great work @HGWright ! Please remember to create a new issue to address [these points](https://github.com/SciTools/iris/pull/5190#issuecomment-1490466234)

HGWright:
Thanks @lbdreyer great to finally get this across the line. For the new issue please see #5344

PR #5194, opened 2023-03-13
iris.util.new_axis anonymous new dimension fix

### LINKED ISSUE #4415: `iris.util.new_axis` gives odd result if existing dim coord is passed
## 🐛 Bug Report

If the scalar coord passed to `iris.util.new_axis` is already a dimension coordinate, an anonymous new dimension is created.

## How To Reproduce
```python
import iris
import iris.cube

print(iris.__version__)

cube = iris.cube.Cube(1)
coord = iris.coords.DimCoord(1, long_name="spam")
cube.add_aux_coord(coord)

new_cube = iris.util.new_axis(cube, coord)
new_new_cube = iris.util.new_axis(new_cube, coord)

for c in [cube, new_cube, new_new_cube]:
    print(c)
```
output:
```
3.2.dev0
unknown / (unknown)                 (scalar cube)
    Scalar coordinates:
        spam                        1
unknown / (unknown)                 (spam: 1)
    Dimension coordinates:
        spam                             x
unknown / (unknown)                 (-- : 1; spam: 1)
    Dimension coordinates:
        spam                            -        x
```

## Expected behaviour

I think we should probably get an exception if our "spam" coord is already a dim coord.  It might be as simple as adding `dim_coords=False` in the call to `src_cube.coord` here:
https://github.com/SciTools/iris/blob/ab1e699bcb9d05eb1dd212b5e159ca01c961931b/lib/iris/util.py#L1116-L1117

## Environment 
 - OS & Version: RHEL 7.9
 - Iris Version: 3.2.dev0

scottrobinson02:
## 🚀 Pull Request
Closes #4415 
### Description
Passing an existing dim coord to iris.util.new_axis now raises an exception

### DIFF
--- a/docs/src/whatsnew/latest.rst
+++ b/docs/src/whatsnew/latest.rst
@@ -140,6 +140,7 @@ This document explains the changes made to Iris for this release
 
 .. _@fnattino: https://github.com/fnattino
 .. _@ed-hawkins: https://github.com/ed-hawkins
+.. _@scottrobinson02: https://github.com/scottrobinson02
 
 .. comment
     Whatsnew resources in alphabetical order:


lib/iris/util.py
```diff
@@ -1171,7 +1171,7 @@ def _handle_dimensional_metadata(
         cube_add_method(new_dm_item, new_dims)
 
     if scalar_coord is not None:
-        scalar_coord = src_cube.coord(scalar_coord)
+        scalar_coord = src_cube.coord(scalar_coord, dim_coords=False)
```
bjlittle, line 1174: @scottrobinson02 Neat use of `dim_coords=False` :+1: 

Note that, the traceback generated here when the `scalar_coord` already exists on the `src_cube` **and** is associated with a cube dimension, is the same traceback when the `src_cube` doesn't contain the provided `scalar_coord`.

This might be a source of confusion to the user. Perhaps it might be clearer to generate different tracebacks for each of these cases i.e., 

1. if the `scalar_coord` already exists on the `src_cube` and is a dimension coordinate, then generate a new exception message to say that the `scalar_coord` cannot be promoted to the new axis as it is already a dimension coordinate of another existing dimension (perhaps even give the cube dimension offset that it covers)
2. otherwise, call `src_cube.coord(scalar_coord, dim_coords=False)` which will either succeed, or generate an exception because the `scalar_coord` genuinely doesn't exist on the `src_cube` i.e., `iris.exceptions.CoordinateNotFoundError: "Expected to find exactly 1 coordinate matching the given 'spam' coordinate's metadata, but found none."`

My point here is that the **same** traceback is generated in each of the following two **different** scenarios:

```python
import iris
from iris.cube import Cube
from iris.util import new_axis

cube = Cube(1)
spam = DimCoord(1, long_name="spam")
parrot = DimCoord(1, long_name="parrot")

cube.add_aux_coord(spam)

new_cube = new_axis(cube, spam)
new_new_cube = new_axis(new_cube, spam)
```

and

```python
import iris
from iris.cube import Cube
from iris.util import new_axis

cube = Cube(1)
spam = DimCoord(1, long_name="spam")
parrot = DimCoord(1, long_name="parrot")

cube.add_aux_coord(spam)

new_cube = new_axis(cube, spam)
new_new_cube = new_axis(new_cube, parrot)
```

bjlittle (changes requested):
@scottrobinson02 This is a lovely contribution, thanks so much :100: 

If you could discriminate between the two **different** cases that generate the **same** exception, then that would be awesome :star_struck: 

You'll probably have to bolster the units tests to cover this scenario.

Also, please make sure to add an entry in the documentation `whatsnew` - you totally deserve the recognition and credit for your contribution.

See the [latest.rst](https://github.com/SciTools/iris/blob/main/docs/src/whatsnew/latest.rst?plain=1#L49) and the `🐛 Bugs Fixed` section :+1: 

If you need any help or advice, don't hesitate to ask :smile:

docs/src/whatsnew/latest.rst
```diff
@@ -56,6 +56,11 @@ This document explains the changes made to Iris for this release
    cubes with masked :class:`iris.coords.CellMeasure`.
    (:issue:`5147`, :pull:`5181`)
 
+#. `@scottrobinson02`_ fixed :class:`iris.util.new_axis` creating an anonymous new
```
bjlittle, line 59: @scottrobinson02 As you're not a core developer (just yet :wink:) you need to add a link to your user at the bottom of the [latest.rst here](https://github.com/SciTools/iris/blob/main/docs/src/whatsnew/latest.rst?plain=1#L136-L137)

bjlittle (approved):
@scottrobinson02 Awesome. Thanks for your first contribution to `iris` :partying_face: :tada: :100:

PR #5201, opened 2023-03-20
enforce modern unittest usage

bjlittle:
### Description
This PR uses the [teyit](https://github.com/isidentical/teyit) tool to enforce modern `unittest` patterns.

Ideally, we would drop our use of `unittest` in preference for `pytest`, but that's dealing with a huge amount of tech debt... the next best thing is to ensure that we're not using deprecated parts of `unittest`, which is happening in the forthcoming [python 3.12](https://docs.python.org/dev/whatsnew/3.12.html#removed) release.

`teyit` is written and maintained by [Batuhan Taskaya](https://github.com/isidentical), who is a [Python Software Foundation](https://www.python.org/psf-landing/) fellow and a `python` core developer.

---

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -60,3 +60,9 @@ repos:
     hooks:
     -   id: sort-all
         types: [file, python]
+
+-   repo: https://github.com/isidentical/teyit
+    rev: 0.4.3
+    hooks:
+    -   id: teyit
+        types: [file, python]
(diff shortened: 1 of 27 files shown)


trexfeathers (changes requested):
Thanks @bjlittle, looks good.

I think it would be helpful to advertise this to developers in a What's New entry.

rcomer (commented):
I didn't look through everything, but I noticed this has caused some duplicated tests.  Here are some specific ones I found before I ran out of enthusiasm...

bjlittle:
> I didn't look through everything, but I noticed this has caused some duplicated tests.  Here are some specific ones I found before I ran out of enthusiasm...

I'll step through and check them 👍

PR #5205, opened 2023-03-22
Remove Iris' TestRunner

### LINKED ISSUE #5173: Retire Iris' TestRunner
I propose we retire the TestRunner and instead use pytest (or nox*) directly.

As far as I can tell, originally the TestRunner was implemented such that it was called when you run `python setup.py test`.  It was then moved to a new command:  `python -m iris.tests.runner` and we then used that to run the tests in the CI or locally.

Since adopting nox and pytest, we now have multiple layers in how we test Iris: 
**nox** calls **the test runner** which calls **pytest**
This seems unnecessary and potentially confusing to a new developer; it's hard to tell where to look for the testing logic. Interestingly [our docs ](https://scitools-iris.readthedocs.io/en/latest/developers_guide/contributing_running_tests.html) only mention running the tests with pytest and nox. I can see the benefit of the test runner as it provides conveniences for running certain tests, but I suspect it may make more sense for that logic to live in pytest (in pytest.ini) or perhaps in nox instead.

So unless there are any objections, I propose we remove the TestRunner.

*Note I have mentioned nox in this issue but[ considering the move to tox](https://github.com/SciTools/iris/discussions/4806) it might be that we use that instead.

lbdreyer:
Closes #5173

The iris.tests.runner had various functionality. If we remove it we can replace it with the alternative listed below

|iris.tests.runner option|nox alternative|pytest alternative|
|--|--|--|
|--default-tests|`nox --session tests`|`pytest lib/iris/tests`|
|--gallery-tests|`nox --session gallery`| `pytest docs/gallery_tests`|
|--system-test|-|`pytest lib/iris/tests/system_test.py`|
|--coverage|`nox --session tests --coverage`|`pytest lib/iris/tests --cov=lib/iris`|
|--create-missing|`export IRIS_TEST_CREATE_MISSING=True nox --session tests`|`export IRIS_TEST_CREATE_MISSING=True pytest lib/iris/tests`|
|--no-data| -| -|
|--num-processors| (nox will use `auto`)|`pytest -n 2 lib/iris/tests`|
| --stop | - |`pytest -x lib/iris/tests`|

Before doing this, I didn't even realise the test runner had all these options, e.g. I have always manually set `IRIS_TEST_CREATE_MISSING`
There is not a direct alternative for every option. For example --no-data doesn't really have an alternative, but I can't really see the use case for this. At a later stage we could consider controlling this with pytest markers.


trexfeathers (approved):
Let's do it!

Issue #4994, opened 2022-09-27
Xarray bridge


pp-mo:
## Provide a bridge to convert data to+from Xarray
Enable use of Iris+Xarray together, with fast+efficient exchange between the two.

Primary Goals:
  * lossless, file-less exchange of information between Iris and Xarray
    * (i.e. cubes <--> xr.Dataset)
  * should be ~equivalent to file exchange : iris-save/xr-load, or the converse
  * but...
    * fast and easy !
    * no-copy and lazy-preserving transfer of (large) data arrays

Roughly following the ideas (~#4835~ now F-B https://github.com/SciTools/iris/tree/FEATURE_xarray_readwrite ) 
And especially to enable possible solutions [listed in this comment](https://github.com/SciTools/iris/pull/4835#issuecomment-1172198937)

Also relevant (to lazy data) : ideas in [pp-mo/iris#73](https://github.com/pp-mo/iris/pull/73)
But NB while functional, this approach is very clunky : see [generally simpler scheme suggested here](https://github.com/pp-mo/iris/pull/73#discussion_r914639958)

```[tasklist]
### Tasks
- [ ] https://github.com/SciTools/iris/issues/5724
```

pp-mo:
#5024 is the beginnings of a new approach to this.

dcherian:
> lossless, file-less exchange of information between Iris and Xarray

Is there something missing in Xarray's `to_iris` and `from_iris` methods? See https://github.com/pydata/xarray/blob/cdab326bab0cf86f96bcce4292f4fae24bddc7b6/xarray/convert.py

pp-mo:
> > lossless, file-less exchange of information between Iris and Xarray
> 
> Is there something missing in Xarray's `to_iris` and `from_iris` methods? See https://github.com/pydata/xarray/blob/cdab326bab0cf86f96bcce4292f4fae24bddc7b6/xarray/convert.py

Not really as such, but there is quite a list of features that we can't get from that which we'd like.
I haven't really looked into the differences between `iris.save(cubes)` and `xr.convert.from_iris(cubes).to_netcdf()`.
But I **_have_** looked extensively into differences between `iris.load('var')` and `xr.convert.to_iris(xr.open_dataset()['var']`
Apart from some corner cases and problems which could be addressed, there is a fairly long list of things which Iris recognises, which are lost in loading 'via' xarray in this way.
Here's a summary ...
  * global attributes
  * coordinate systems, attached via a 'grid_mapping' link
[...]

PR #5212, opened 2023-03-27
Support netcdf variable emulation

pp-mo:
The basic idea : when loading from / saving to a dataset-like object (as provided by #5214), Iris can spot objects which "emulate" a netCDF4.Variable and provide direct access to a content array (i.e. not via file access).  
This allows us to directly transfer array data (real or lazy) without copying.

**TODO**: in retrospect, `_in_memory_array` is really _not_ a good choice, as it seems to imply numpy data, whereas crucially it will often be dask.  We should change the name to something better.
Maybe "_data_array".

This, along with #5214 (formerly #5024), should provide all we need to support the intended "Xarray bridge" - see #4994 .
Scheduled for  Iris 3.6

Currently ***draft/WIP***, because this will need adjustment after #5191, and it's far easier to resolve if that goes in first !

Also note [comment here](https://github.com/SciTools/iris/pull/5216#issue-1642432269) 
> we may have a bit more work to successfully integrate these changes with those from https://github.com/SciTools/iris/pull/5095

### DIFF
--- a/lib/iris/fileformats/netcdf/loader.py
+++ b/lib/iris/fileformats/netcdf/loader.py
@@ -190,34 +190,42 @@ def _get_cf_var_data(cf_var, filename):
     unnecessarily slow + wasteful of memory.
 
     """
-    total_bytes = cf_var.size * cf_var.dtype.itemsize
-    if total_bytes < _LAZYVAR_MIN_BYTES:
-        # Don't make a lazy array, as it will cost more memory AND more time to access.
-        # Instead fetch the data immediately, as a real array, and return that.
-        result = cf_var[:]
-
+    if hasattr(cf_var, "_data_array"):
+        # The variable is not an actual netCDF4 file variable, but an emulating
+        # object with an attached data array (either numpy or dask), which can be
+        # returned immediately as-is.  This is used as a hook to translate data to/from
+        # netcdf data container objects in other packages, such as xarray.
+        # See https://github.com/SciTools/iris/issues/4994 "Xarray bridge".
+        result = cf_var._data_array
     else:
-        # Get lazy chunked data out of a cf variable.
-        dtype = _get_actual_dtype(cf_var)
+        total_bytes = cf_var.size * cf_var.dtype.itemsize
+        if total_bytes < _LAZYVAR_MIN_BYTES:
+            # Don't make a lazy array, as it will cost more memory AND more time to access.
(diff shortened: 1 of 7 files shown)


pp-mo:
Ping @ESadek-MO @trexfeathers.  Ideally I'd like this to squeeze into [Iris 3.5](https://github.com/orgs/SciTools/projects/4).

Although we've now bumped #4994 to Iris 3.6, the remainder of that work will happen in a separate "ncdata" package.  
So if we can get these preliminaries into Iris itself, we can go live with that,  **_without_** requiring a further Iris release to support it  (I think :crossed_fingers: !)

trexfeathers:
>Ping @ESadek-MO @trexfeathers. Ideally I'd like this to squeeze into [Iris 3.5](https://github.com/orgs/SciTools/projects/4).

IMHO this should be a stretch goal, since we've already committed to the work that we will do for 3.5.

pp-mo:
> > Ping @ESadek-MO @trexfeathers. Ideally I'd like this to squeeze into [Iris 3.5](https://github.com/orgs/SciTools/projects/4).
> 
> IMHO this should be a stretch goal, since we've already committed to the work that we will do for 3.5.

It was a mistake !  now in Iris 3.6

pp-mo:
### Update : rebased
Actually more re-written, but it parallels what went before.
Been out rather too long :grimacing: :clock11:

pp-mo:
### Update : Ready for Review
Existing tests now passing.
Tested against a revised WIP 'ncdata' + seems to function.

Ping @ESadek-MO is this for you to review also ?
( when done with #5214 !! )

pp-mo:
> Looks good to me in it's current state, is this waiting on threadsafety as in #5214?

These changes are supposedly ~independent, but yes I think it's probably best to get #5214 in first.
I'm just making a merge of the 2 for re-testing against ncdata, and I'm finding conflicts.

Thanks for looking!

pp-mo:
I think I was wrong about the conflicts -- using an older version.
I am just re-spinning the tests to be sure it's good ...

pp-mo:
Ok thanks @ESadek-MO , I'm happy this is good to go !

( btw I tested a combination of this+#5414 against prototype ncdata, and it seems good 👍  )

ESadek-MO (approved):
Thanks @pp-mo, looks good to me!

Issue #4523, opened 2022-01-22
Fully adopt pathlib.PurePath for file paths


SimonPeatman:
## ✨ Feature Request
Iris uses `os.path` to handle file paths, but the standard in Python 3 is to use `pathlib.PurePath` objects.  #3411 proposed adopting `pathlib.PurePath` and this is partially implemented by #3917.  The present issue proposes replacing `os.path` with the use of `pathlib` across the whole of Iris.

## Details
#3917 implements `pathlib.PurePath` for the main public functions only (`load`, `load_cube`, `load_cubes`, `load_raw` and `save`).  See https://github.com/SciTools/iris/pull/3917#pullrequestreview-846736928 for full details of what has been done and what is still to do.

pp-mo:
Is this actually closable now  @SimonPeatman @wjbenfold ?
Excuse my ignorance, I haven't been following detail here -- just tidying issues

wjbenfold:
No, @pp-mo, this is the issue that has come out of the recent work on pathlib. We've so far introduced it "superficially" - i.e. the most frequently used bits of the public api will now accept a pathlib path but it's almost immediately converted to a string. This ticket is for the further work that would use paths all the way down (maybe converting input strings to paths) and leverage the features of paths that we currently use `os` for.

trexfeathers:
Suggest scanning the entire repo for any uses of `os.path` as an easy way to root things out

PR #5214, opened 2023-03-27
Support netCDF load+save on dataset-like objects as well as filepaths.

pp-mo:
Rebased version of #5024

---


lib/iris/fileformats/cf.py
```diff
@@ -1043,17 +1043,25 @@ class CFReader:
     # TODO: remove once iris.experimental.ugrid.CFUGridReader is folded in.
     CFGroup = CFGroup
 
-    def __init__(self, filename, warn=False, monotonic=False):
-        self._dataset = None
-        self._filename = os.path.expanduser(filename)
+    def __init__(self, file_source, warn=False, monotonic=False):
+        # Ensure safe operation for destructor, should init fail.
+        self._own_file = False
+        if isinstance(file_source, str):
```
pp-mo, line 1049: I think this needs updating to take account of a *Path* argument,
since https://github.com/SciTools/iris/pull/3917 ??

pp-mo:
### Update
Rebased onto latest main
 **still WIP**

lib/iris/fileformats/netcdf/loader.py
```diff
@@ -483,14 +484,15 @@ def inner(cf_datavar):
     return result
 
 
-def load_cubes(filenames, callback=None, constraints=None):
+def load_cubes(file_sources, callback=None, constraints=None):
     """
-    Loads cubes from a list of NetCDF filenames/OPeNDAP URLs.
+    Loads cubes from a list of NetCDF file_sources/OPeNDAP URLs.
```
pp-mo, line 489: TODO: fix text (autoreplace?!?)

lib/iris/fileformats/netcdf/loader.py
```diff
@@ -483,14 +484,15 @@ def inner(cf_datavar):
...
 
     Args:
 
-    * filenames (string/list):
-        One or more NetCDF filenames/OPeNDAP URLs to load from.
+    * file_sources (string/list):
+        One or more NetCDF file_sources/OPeNDAP URLs to load from.
```
pp-mo, line 494: TODO: fix text

lib/iris/fileformats/netcdf/loader.py
```diff
@@ -518,18 +520,18 @@ def load_cubes(filenames, callback=None, constraints=None):
     # Create an actions engine.
     engine = _actions_engine()
 
-    if isinstance(filenames, str):
-        filenames = [filenames]
+    if isinstance(file_sources, str) or not isinstance(file_sources, Iterable):
```
pp-mo, line 523: What about paths?

pp-mo: Update : **Works ok**
for Path / str / Dataset or any iterable of same.
 -- only "str" requires specially treatment "as if not an iterable" : Path is not iterable anyway.

lib/iris/fileformats/cf.py
pp-mo: Loader sources are converted from Path to str by `iris.io`, [here](https://github.com/SciTools/iris/blob/0d0116f9de420530a10fe359a267fb74e8187d17/lib/iris/io/__init__.py#L128-L129).

There might still be some unfinished business about support for Path internally (cf #4523), but it won't affect this work.

pp-mo:
### Update
I discovered that this feature now has a problem with output to a already-open dataset (one owned by the caller) **when`compute=True`** (which is the default!) :  
  * since we added support for distributed schedulers in #5191, lazy streaming requires the original file to be closed -- because the dask workers doing the streaming may now be separate processes, not always threads, which means they must each (re-)open the dataset _for writing_ -- which they can't if the original caller still has it open
  * this simply isn't compatible with saving to an open dataset, because iris.save doesn't have the right to **_close_** the dataset (since I don't think netCDF4 supports "re-opening" a dataset, and closing will invalidate any netCDF4 objects held by user).

So that means that, in most cases,  you can't just `iris.save(cubes, my_dataset)`.
You must instead `saves = iris.save(cubes, my_dataset, compute=False); my_dataset.close(); saves.compute()`

Following offline discussion with @bjlittle @trexfeathers 
  * We can either 
    * forbid this usage, or 
    * error out **_except_** in those specific cases which happen to be OK -- e.g. when dataset is not a real file-based dataset (as intended for 4994 ), or if all data is real.

  * **we decided that** the usecase is so specialist that it will probably only be used for #4994 support.  Hence we should **ban `compute=True` operation saving to a provided dataset**, since the result is still useable (just less convenient in a rare usecase) and it is the simplest solution, _both to implement and document_.

pp-mo:
Ping @ESadek-MO 
Where you going to review this (:crossed_fingers: :pray: ) ?
I think it is at last ready to go !

pp-mo:
> Ping @ESadek-MO Where you going to review this (crossed_fingers pray ) ? I think it is at last ready to go !

P.S. #5212 will be a follow-up, which should complete essential support for #4994.  
But that should be **_a lot_** simpler than this.

lib/iris/tests/integration/netcdf/test_general.py
```diff
@@ -361,5 +362,125 @@ def test_lat_not_loaded(self):
...
+            ["NetCDF", "global", "xyz_t", "GEMS_CO2_Apr2006.nc"]
+        )
+        cls.phenom_id = "Carbon Dioxide"
+        cls.expected = iris.load_cube(cls.filepath, cls.phenom_id)
+
+    def test_basic_load(self):
+        # test loading from an open Dataset, in place of a filepath spec.
+        ds = threadsafe_nc.DatasetWrapper(self.filepath)
+        result = iris.load_cube(ds, self.phenom_id)
+        ds.close()
```
ESadek-MO, line 379: Is it worth ensuring that the dataset isn't already closed? Or does this line check this?

ESadek-MO (commented):
Hey @pp-mo, looks good, just a couple soft suggestions!

lib/iris/io/__init__.py
```diff
@@ -255,6 +268,25 @@ def load_http(urls, callback):
...
+    recognised by the handlers :  This is done in the usual way by passing the
+    context to the format picker to get a handler for each.
+
+    .. note::
+
+        Typically, this function should not be called directly; instead, the
+        intended interface for loading is :func:`iris.load`.
+
+    """
+    # NOTE: this operation is currently *identical* to the http one.
```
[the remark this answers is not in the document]
pp-mo: None really.  I think it's slightly _more_ likely that the http one might change -- though really _not_ very likely ! -- and then we might  need to provide an independent function here.
Nevertheless I think it _is_ best to have a separate named routine, specific to this 'scheme', so I'll add  a note to that effect.

lib/iris/tests/integration/netcdf/test_general.py
```diff
@@ -361,5 +362,125 @@ def test_lat_not_loaded(self):
...
+        # Destroy the temp directory.
+        shutil.rmtree(cls.temp_dir)
+
+    def test_basic_save(self):
+        # test saving to a Dataset, in place of a filepath spec.
+        # NOTE that this requires 'compute=False', as delayed saves can only operate on
+        # a closed file.
+
+        # Give the test cubes a definite order, since this is not stable !
+        testdata = sorted(self.testdata, key=lambda cube: cube.name())
```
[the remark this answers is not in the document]
pp-mo: Whoops, good spot !!

lib/iris/fileformats/netcdf/_thread_safe_nc.py
```diff
@@ -35,22 +35,42 @@ class _ThreadSafeWrapper(ABC):
...
     def _from_existing(cls, instance):
         """Pass an existing instance to __init__, where it is contained."""
-        assert isinstance(instance, cls.CONTAINED_CLASS)
+        assert cls.is_contained_type(instance)
         return cls(instance)
 
     def __init__(self, *args, **kwargs):
         """Contain an existing instance, or generate a new one from arguments."""
-        if isinstance(args[0], self.CONTAINED_CLASS):
+        if len(args) > 0 and self.is_contained_type(args[0]):
```
trexfeathers, line 61:
```suggestion
        if len(args) == 1 and self.is_contained_type(args[0]):
```

This is a very deliberately designed use case. If this receives 2 or more arguments then the developer is going very much off-piste and not using as intended!

lib/iris/fileformats/netcdf/_thread_safe_nc.py
```diff
@@ -35,22 +35,42 @@ class _ThreadSafeWrapper(ABC):
...
+            # Passed a contained-type object : Wrap ourself around that.
             instance = args[0]
+            # Explicitly ban "wrapping a wrapper".
+            if hasattr(instance, "THREAD_SAFE_FLAG"):
+                msg = (
+                    "Cannot create {cls} containing an existing {cls}.".format(
+                        cls=self.__class__.__name__
+                    )
+                )
+                raise ValueError(msg)
```
trexfeathers, line 71: IMO an `assert` probably would have been enough for this sort of developer-level stuff, and be a bit more readable. I could be persuaded otherwise.

trexfeathers (commented):
My thoughts on the changes to `_thread_safe_nc.py`:

lib/iris/fileformats/netcdf/_thread_safe_nc.py
```diff
@@ -35,22 +35,42 @@ class _ThreadSafeWrapper(ABC):
...
     THREAD_SAFE_FLAG = True
 
+    @classmethod
+    def is_contained_type(cls, instance):
+        return all(
+            hasattr(instance, attr) for attr in cls._DUCKTYPE_CHECK_PROPERTIES
+        )
+
     @classmethod
     def _from_existing(cls, instance):
```
trexfeathers, line 54: Do you think we should make this public since it's being used elsewhere now?

Alternatively: if we want to distinguish our dedicated class methods from the original methods of the NetCDF4 classes, then we should make `is_contained_type()` private.

I'm happy either way, but the same should definitely apply to both methods.

lib/iris/fileformats/netcdf/_thread_safe_nc.py
pp-mo: My only concern was to avoid this causing an error with **_no_** arguments.  I think any of the actual intended contained classes will give an error then anyway, but at least it's then raised by the appropriate thing.

pp-mo: Sorry got confused about this.  Yes, just one arg of an expected type is the thing.

lib/iris/fileformats/netcdf/_thread_safe_nc.py
pp-mo: This is good.  I was going to add a test -- but that's definitely overkill !!

pp-mo:
Thanks @trexfeathers 
I think I've responded to all of those, now.
Please re-review !

ESadek-MO (approved):
Hey @pp-mo, all looks good now!
Thanks for the review assist @trexfeathers!

pp-mo:
Thanks so much, @ESadek-MO @trexfeathers !

PR #5204, opened 2023-03-22
standardize requirements structure

bjlittle:
### Description

This PR rationalizes and standardizes the `iris` requirements structure i.e.,

- remove the `requirements/ci` directory
- move the `requirements/ci/nox.lock` directory to `requirements/locks`
- update connective tissue from these changes within `iris` codebase

A subsequent PR will target `SciTools/workflows` to align the `refresh-lockfiles.yml` GHA with this structure.

Reference: 
- https://github.com/SciTools/workflows/pull/8
- https://github.com/SciTools-incubator/iris-esmf-regrid/pull/244
- https://github.com/SciTools/iris-grib/pull/327

---
[Consult Iris pull request check list]( https://scitools-iris.readthedocs.io/en/latest/developers_guide/contributing_pull_request_checklist.html)

PR #5217, opened 2023-03-27
Fix Makefile for consequences of #5204.

pp-mo:
I just bumped into this problem.

FYI @trexfeathers 
I'm not clear if `Makefile` is still a worthwhile thing  (though I just used it!)
We could instead just rely on the command documented [here](https://scitools-iris.readthedocs.io/en/latest/developers_guide/contributing_ci_tests.html?highlight=lockfile#github-actions-test-environment), and not bother with make since it only provides that one command.


bjlittle:
@pp-mo My expectation is that `make` in this context will be replaced with `tox`.

bjlittle (approved):
@pp-mo Nice one, thanks for mopping up 🍻

PR #4279, opened 2021-08-11
CellMethods now printed in CF-compliant manner

MoseleyS:
### Description

Fixes #3592 

This simple PR rearranges the order in which CellMethod information is formatted by the __str__ method. The grouping of this information (where multiple coords refer to the same method) has therefore been removed.
---
[Consult Iris pull request check list]( https://scitools-iris.readthedocs.io/en/latest/developers_guide/contributing_pull_request_checklist.html)

PR #5224, opened 2023-03-31
Cf cell method

HGWright:
### Description

*This is a draft while making sure doctests pass, Also missing a whatsnew entry that will be added when moving out of draft*

Finishes #4279.

The cell methods section of the cube printout was not printing in a CF-compliant manner and so this PR goes towards changing that.

---


lib/iris/coords.py
```diff
@@ -3078,23 +3078,24 @@ def __init__(self, method, coords=None, intervals=None, comments=None):
     def __str__(self):
         """Return a custom string representation of CellMethod"""
         # Group related coord names intervals and comments together
-        cell_components = zip_longest(
-            self.coord_names, self.intervals, self.comments, fillvalue=""
+        coord_string = " ".join([f"{coord}:" for coord in self.coord_names])
+        method_string = f"{str(self.method)}"
```
trexfeathers, line 3082:
```suggestion
        method_string = str(self.method)
```

lib/iris/_representation/cube_summary.py
```diff
@@ -265,12 +265,11 @@ def __init__(self, title, cell_methods):
         self.values = []
         self.contents = []
         for method in cell_methods:
-            name = method.method
-            # Remove "method: " from the front of the string, leaving the value.
-            value = str(method)[len(name + ": ") :]
-            self.names.append(name)
+            index = str(cell_methods.index(method))
```
trexfeathers, line 268: In the line above (it won't let me make a direct suggestion!), use:

```python
for index, method in enumerate(cell_methods):
```

Then you won't need to derive `index` after the fact.

lib/iris/coords.py
```diff
@@ -3078,23 +3078,24 @@ def __init__(self, method, coords=None, intervals=None, comments=None):
...
+        interval_string = " ".join(
+            [f"interval: {interval}" for interval in self.intervals]
         )
```
trexfeathers, line 3086: The below could be notably simpler, if you are up for it:

```python
        comment_string = " ".join([comment for comment in self.comments])

        if interval_string and comment_string:
            comment_string = "".join(
                [f" comment: {comment}" for comment in self.comments]
            )

        cm_summary = f"{coord_string} {method_string}"
        if interval_string or comment_string:
            cm_summary += f" ({interval_string}{comment_string})"

        return cm_summary
```

lib/iris/tests/unit/representation/cube_printout/test_CubePrintout.py
```diff
@@ -537,10 +537,11 @@ def test_section_cell_methods(self):
...
             "name / (1)                          (-- : 1)",
             "    Cell methods:",
-            "        stdev                       area",
-            "        mean                        y (10m, vertical), time (3min, =duration)",
+            "        0                           area: stdev",
+            "        1                           y: time: mean: (interval: 10m"
+            " interval: 3min comment: vertical comment: =duration)",
         ]
-        self.assertEqual(rep, expected)
+        self.assertEqual(expected, rep)
```
trexfeathers, line 544: Why did you swap these?

trexfeathers (changes requested):
Thanks @HGWright, awesome work. Some comments for you...

trexfeathers:
Also this pre-commit failure is an easy fix - just remove the import.

image

pp-mo:
> Also this pre-commit failure is an easy fix - just remove the import.
> 
> image

P.S. you should be forewarned of this when you commit locally, so I guess you are have not installed the pre-commit tool.
There is advice here : https://scitools-iris.readthedocs.io/en/stable/developers_guide/contributing_code_formatting.html?highlight=pre-commit#code-formatting
For extra points, it would be great to know, if that does not work!

lib/iris/common/resolve.py
```diff
@@ -144,7 +144,7 @@ class Resolve:
                 forecast_reference_time     1859-09-01 06:00:00
                 height                      1.5 m
             Cell methods:
-                mean                        time (6 hour)
+                0                           time: mean: (interval: 6 hour)
```
pp-mo, line 147: I reckon that the extra colon here is not correct.
It should be just `"time: mean (interval: 6 hour)"`

lib/iris/tests/unit/representation/cube_printout/test_CubePrintout.py
pp-mo: > Why did you swap these?

I believe this **_is_** better, but I only learnt it lately :  Because of how assertion routines are defined, and how it reports a failure, it should really be "expected-value, test-value".
Thus if x is "wrong-result", then assertEqual(x, 2) will report "2 != 'wrong' ", whereas assertEqual(2, x)  gives "'wrong-result' != 2".

docs/src/further_topics/ugrid/operations.rst
```diff
@@ -566,8 +566,8 @@ Here's another example using a global cubed-sphere data set:
             Auxiliary coordinates:
                 time                            x       -
             Cell methods:
-                mean                        time (300 s)
-                mean                        time_counter
+                0                           time: mean: (interval: 300 s)
+                1                           time_counter: mean
```
pp-mo, line 570: I don't want to totally upset the apple-cart, but I don't really see the value of having the index numbers here -- even though I believe it **_is_** true that order is significant, 
e.g. you can have "time: mean over years time: max within years" 
(and the reverse order means something _different_).

To me, it would make sense for the index column to list the coord (or equivalent measure), i.e. the bit before the colon.  Something like this:
```
    Cell methods:
        time:             mean (interval: 300 s)
        time_counter:     mean
```

And now, looking back to the original proposal in #4729, I see that is what was originally proposed.
What do you think about this, @trexfeathers ?

lib/iris/common/resolve.py
pp-mo: Oh, I think this was maybe already pointed out in [this comment](https://github.com/SciTools/iris/pull/5224/files#r1154437823)

lib/iris/tests/results/cdm/str_repr/cell_methods.__str__.txt
```diff
@@ -8,10 +8,10 @@ air_temperature / (K)               (latitude: 73; longitude: 96)
         pressure                    1000.0 hPa
         time                        1998-12-01 00:00:00
     Cell methods:
-        mean                        longitude (6 minutes, This is a test comment), latitude (12 minutes)
-        average                     longitude (6 minutes, This is another test comment), latitude (15 minutes, This is another comment)
-        average                     longitude, latitude
-        percentile                  longitude (6 minutes, This is another test comment)
+        0                           longitude: latitude: mean: (interval: 6 minutes interval: 12 minutes comment: This is a test comment)
+        1                           longitude: latitude: average: (interval: 6 minutes interval: 15 minutes comment: This is another test comment comment: This is another comment)
```
pp-mo, line 12: I thought this bit was not correct (though it's hard to read in GitHub), 
but I located the test and saved the cube, and I now see it **_is_** correct !
(it was the old form that was seriously wrong)
What it really means, I'm still not quite sure TBH

But it made me wonder... I know these are really old tests (in tests/test_cdm.py), but I wonder if it's worth adding a cross-check into all these original cases, between what you get when you save the cube + how it prints out, .  
I think you could automate that ?

trexfeathers: Regretfully shelved due to time constraints.

trexfeathers (approved):

trexfeathers:
Great! I'll merge once this is out of draft

PR #5226, opened 2023-04-01
Support Python 3.11

rcomer:
### Description

Python 3.11 has been out for [nearly 6 months now](https://www.python.org/downloads/release/python-3110/) so I think we should officially support for the upcoming release.

~I note @trexfeathers' comment at https://github.com/SciTools/iris/pull/5211#issuecomment-1484725649, so appreciate you won't want to merge this immediately. :eyes:~

Needs #5232 

---


rcomer:
Wow I managed to break **_all_** the tests.  Seems I need some help!

I can report that the main tests, gallery tests all passed locally under 3.11.

rcomer:
It's amazing the variety of errors you can generate by just not making sure your branch is up-to-date with upstream 🤦‍♀️

The doctest and linkcheck fails are probably due to #5232, though I can't see the specific errors in the logs.  I am somewhat suspicious that readthedocs passed.

rcomer:
> I am somewhat suspicious that readthedocs passed.

It passed, but the rendered docs are not showing the calling sequence for [e.g. cube instantiation](https://scitools-iris--5226.org.readthedocs.build/en/5226/generated/api/iris/cube.html#iris.cube.Cube).  So I guess the failure is there but RTD is not registering the non-zero exit code?

rcomer:
Behold my full set of green ticks ✅😎

lib/iris/tests/unit/experimental/ugrid/mesh/test_Connectivity.py
```diff
@@ -61,10 +62,14 @@ def test_indices(self):
 
     def test_read_only(self):
         attributes = ("indices", "cf_role", "start_index", "location_axis")
+        if sys.version_info.minor >= 11:
+            msg = "object has no setter"
+        else:
+            msg = "can't set attribute"
```
bjlittle, line 68: @rcomer I'd recommend using [pkg_resources.parse_version](https://setuptools.pypa.io/en/latest/pkg_resources.html#utility-functions) from `setuptools` which is [pep440](https://peps.python.org/pep-0440/) compliant :+1: 

So something along the lines of:
```python
from pkg_resources import parse_version
from platform import python_version

msg = "object has no setter" if parse_version(python_version()) >= parse_version("3.11") else "can't set attribute"
```

.github/workflows/ci-tests.yml
```diff
@@ -35,9 +35,13 @@ jobs:
...
+        python-version: ["3.11"]
         session: ["doctest", "gallery", "linkcheck"]
         include:
+          - os: "ubuntu-latest"
+            python-version: "3.11"
+            session: "tests"
+            coverage: "--coverage"
           - os: "ubuntu-latest"
             python-version: "3.10"
             session: "tests"
```
bjlittle, line 47: @rcomer We should only really be submitting test coverage for one version of `python`, so remove the `coverage: "--coverage"` for `python-version: "3.10"` on `line +48` in the `matrix` :+1:

bjlittle (changes requested):
@rcomer Thanks for doing this :beers: 

Just some minor comments to service, then I think we're good to go :+1:

bjlittle:
@rcomer I'd say that a `whatsnew` entry is defo also needed :100:

lib/iris/tests/unit/experimental/ugrid/mesh/test_Connectivity.py
rcomer: Ah thanks, I wondered if there was a better way to do that 👍

rcomer: `black` turned your one-liner into 5 lines, so I prefer the 4-line loop!

bjlittle (approved):

PR #5229, opened 2023-04-03
Use real array for data of of small netCDF variables.

pp-mo:
Successor to the experimental #5069.
Addresses #5053 

**Draft** while I see what tests it may break.


pp-mo:
@ESadek-MO I think this is good to go now.
Please take a look + see what you think.

N.B. there is basically no API and no "public view" of this, it's just a performance tweak.

On the other hand, it does affect some loads in a backwards-incompatible way.
@trexfeathers can we accept that the real/lazy status of data arrays is not something that we have to absolutely preserve,
or should this count as a breaking change ?

trexfeathers:
>On the other hand, it does affect some loads in a backwards-incompatible way.
>@trexfeathers can we accept that the real/lazy status of data arrays is not something that we have to absolutely preserve,
>or should this count as a breaking change ?

I'm comfortable with this change. We encourage people to CHECK if their data is lazy or not, rather than to assume it is one way or the other.

lib/iris/tests/unit/fileformats/netcdf/loader/test__get_cf_var_data.py
```diff
@@ -25,16 +24,24 @@ def setUp(self):
         self.shape = (300000, 240, 200)
         self.expected_chunks = _optimum_chunksize(self.shape, self.shape)
 
-    def _make(self, chunksizes):
-        cf_data = mock.Mock(_FillValue=None)
+    def _make(self, chunksizes=None, shape=None, dtype="i4"):
```
ESadek-MO, line 27: Perhaps specify dtype= np.dtype("i4")

ESadek-MO (changes requested):
Looks god on the whole, only one thing to address then I think you're good!

lib/iris/tests/unit/fileformats/netcdf/loader/test__get_cf_var_data.py
pp-mo: Actually, I now see that this is OK.
we have a line below which "regularises" this, i.e. [line 35 `dtype = np.dtype(dtype)`](https://github.com/SciTools/iris/blob/3dabecae5f10874fc7b460a2b4e290f0155e0a36/lib/iris/tests/unit/fileformats/netcdf/loader/test__get_cf_var_data.py#L35)
So that makes it OK :  `_make(...).dtype` is always a proper np.dtype object, and the default is `np.dtype('i4')`.

ESadek-MO (approved):

ESadek-MO:
Looks good, thanks @pp-mo!

PR #5235, opened 2023-04-04
Replacing numpy legacy printing with array2string and remaking result…

### LINKED ISSUE #3048: unpin numpy legacy printing

HGWright:
…s for dependent tests

### Description

Replaced the use of a legacy pin for numpy array printing with array2string. This initially caused just under 500 tests to fail and so the files containing the expected results for these tests have been replaced. 

closes #3048 

missing whatsnew entry will remove from draft once added

---


docs/src/whatsnew/latest.rst
```diff
@@ -155,6 +155,10 @@ This document explains the changes made to Iris for this release
    pin for first use of `dask.array.ma.empty_like`_ and replaced `@tinyendian`_
    workaround. (:pull:`5225`)
 
+#. `@HGWright`_, `@bjlittle`_ and `@trexfeathers`_ removed the legacy pin for
+   numpy array printing and replaced the test results files to match the not
```
bjlittle, line 159: @HGWright The wording doesn't seem quite right i.e., `to match the not` ?

bjlittle (changes requested):
@HGWright This is great, thanks for making this happen!

The format of the CML isn't user facing, and dev's hardly ever look at the CML, so I'm not precious about tabbing indents as previously (simply because I know you'd have to go off-piste to make that happen) and I'm keen that we adopt the sane defaults of `numpy` with minimal custiomisation i.e., only do that for a good reason and by exception.

So this all good good to me. Just a minor comment to service on the `whatsnew` entry, and I know the `codecov.yml` and `.git-blame-ignore-revs` are in-bound.

codecov.yml
```diff
@@ -0,0 +1,9 @@
+coverage:
+  # see https://docs.codecov.com/docs/commit-status
+  status:
+    project:
+      default:
+        target: auto
+        # coverage can drop by up to <threshold>% while still posting success
+        threshold: 3%
```
bjlittle, line 8: @HGWright Seems like a sane default initial default to me :+1: 

We can iterate on this `threshold` as time evolves.

bjlittle (approved):

PR #5252, opened 2023-04-17
refresh manual pypi publish instructions

bjlittle:
### Description

This PR refreshes the instructions to manually build and publish the `scitools-iris` sdist and binary wheels to PyPI.

image

---


bjlittle:
Note that, we can also extend this process by adding a `check-manifest` step to ensure that the PyPI assets contain the payload that we expect and require. We can also automate this process as part of the GHA workflow... PRs inbound for that later.

trexfeathers (approved):
I've tested these changes and they work.

PR #5260, opened 2023-04-18
add missing test result data

bjlittle:
### Description

Discovered during #5259

---


trexfeathers (approved):
Looks good. I've checked the test and the file will get overwritten each time so there's no problem storing the result if it makes things easier.

PR #5262, opened 2023-04-18
move setup.cfg to pyproject.toml

bjlittle:
### Description

This PR migrates the `setup.cfg` to the `pyproject.toml`.

For the motivation, see [pep621](https://peps.python.org/pep-0621/).

Note that,
- the `flake8` configuration now lives in the dedicated `.flake8` file. When we migrate to `tox`, then the `.flake8` can then be migrated to the `tox.ini` - see [here](https://flake8.pycqa.org/en/latest/user/configuration.html). Additionally, the `flake8` configuration may be further consolidated if we choose to adopt [ruff](https://github.com/SciTools/iris/discussions/5254) :wink: 
- the PyPI dependencies have not been indirectly migrated into the `pyproject.toml`. Rather, they are now explicitly defined under the `requirements` directory in `*.txt` files. Most devs have fallen foul of updating `conda` dependencies and forgotten about `pypi` dependencies, as they were obfuscated within the `setup.cfg`. Hopefully, banking all our dependencies within the same directory will alleviate this issue
- this pr is related to #5259 i.e., the `MANIFEST.in` requires to be updated to include the `requirements/*.txt` PyPI dependency files
- by doing this work, I since discovered `mo_pack` is not available on PyPI, so it has been removed
- the optional dependency groups are now `dev`, `docs` and `test`, instead of `docs`, `test` and `all` (which included `dev`, `docs` and `test` via using convenience syntax within the `setup.cfg`)

Reference:
- https://setuptools.pypa.io/en/latest/userguide/pyproject_config.html
- #5184

---


pp-mo:
> by doing this work, I since discovered mo_pack is not available on PyPI, so it has been removed

Possibly a bit worrying, as it is so key to UM files support.
We need this to be available in the internal SSS environments, so at least requires adjustment of those definitions.

trexfeathers:
> > by doing this work, I since discovered mo_pack is not available on PyPI, so it has been removed
> 
> Possibly a bit worrying, as it is so essential for UM files support. We need this to be available in the internal SSS environments, so at least requires adjustment of those definitions.

It is available - via Conda

bjlittle:
Thanks to @trexfeathers, #5259 has landed, so I'll rebase and refactor the `MANIFEST.in` :+1:

bjlittle:
I'll also check to see if any changes I've made here make any of the dev docs stale wrt updating PyPI deps (which now live under the `requirements` directory...

bjlittle:
> Thanks to @trexfeathers, #5259 has landed, so I'll rebase and refactor the `MANIFEST.in` +1

Done, see a1bf9e4 and 96b89e1

bjlittle:
> I'll also check to see if any changes I've made here make any of the dev docs stale wrt updating PyPI deps (which now live under the `requirements` directory) ...

Done i.e., nothing to do in the docs (that I can see)

requirements/optional-test.txt
trexfeathers, line 1: Missing some of the things that are in the Conda dependencies.

https://github.com/SciTools/iris/blob/c483352cc677faed02ce5dea58c704ed4eb7898d/requirements/py310.yml#L37-L45

Possible explanations:

- Not available on PyPI
  At least some of the missing dependencies ARE available
- We don't think they need including
  In which case they should also be removed from the Conda YAML files
- They got missed!
  >Most devs have fallen foul of updating `conda` dependencies and forgotten about `pypi` dependencies

requirements/core.txt
trexfeathers, line 1: `pyproj` is in the Conda requirements but missing here.

https://github.com/SciTools/iris/blob/c483352cc677faed02ce5dea58c704ed4eb7898d/requirements/py310.yml#L23

requirements/optional-dev.txt
trexfeathers, line 1: This doesn't align with the same section in the Conda dependencies.

https://github.com/SciTools/iris/blob/c483352cc677faed02ce5dea58c704ed4eb7898d/requirements/py310.yml#L27-L35

### Differences

- ~~`esmpy`~~ : not available on PyPI
- `graphviz`
- `iris-sample-data`
- ~~`mo_pack`~~ : not available on PyPI
- ~~`pip`~~ : can't be included in PyPI dependencies I assume
- `pre-commit` : should be in `optional-test.txt`

trexfeathers: I'm not keen on the `-dev` suffix, as it implies only developers should be interested.

.flake8
trexfeathers, line 1: It's a real shame that a dedicated file is necessary, as it is inconsistent with all the other config we have.

Hopefully PyCQA/flake8#23 will be resolved in the near future, in light of [`PEP 621`](https://peps.python.org/pep-0621/).

pyproject.toml
```diff
@@ -8,6 +8,74 @@ requires = [
...
+    "Programming Language :: Python :: Implementation :: CPython",
+    "Topic :: Scientific/Engineering",
+    "Topic :: Scientific/Engineering :: Atmospheric Science",
+    "Topic :: Scientific/Engineering :: Visualization",
+]
+dynamic = [
+    "dependencies",
+    "optional-dependencies",
+    "readme",
+    "version",
```
trexfeathers, line 37: https://setuptools.pypa.io/en/latest/userguide/pyproject_config.html#dynamic-metadata

>When these fields are expected to be provided by setuptools a corresponding entry is required in the `tool.setuptools.dynamic` table
>...
>```
>...
>[tool.setuptools.dynamic]
>version = {attr = "my_package.VERSION"}
>```

Do we therefore need to include a `version` further down, too? If setuptools_scm is doing something magic here it isn't clear to me.

pyproject.toml
trexfeathers, line 1: 🥵 that was hard work! So many subtle differences

trexfeathers (changes requested):

pp-mo:
> > @bjlittle  by doing this work, I since discovered mo_pack is not available on PyPI, so it has been removed
> > @pp-mo  Possibly a bit worrying, as it is so essential for UM files support. We need this to be available in the internal SSS environments, so at least requires adjustment of those definitions.

> @trexfeathers It is available - via Conda

My point was, if we now have "one true list" of dependencies (for which 👍 🥇!) , will this also remove it from the conda deps ?

pyproject.toml
bjlittle: Yeah, I've hummed and hawed about this one... it's not entirely clear.

I've intentionally followed the pattern on `xesmf`, see https://github.com/pangeo-data/xESMF/pull/248

My understanding is that the `setuptools-scm` plugin can do it under the hood without explicitly specifying it, and it certainly seem to work.

I'll retry with the explicit `dynamic.version` metadata and see how it rolls...

bjlittle:
> My point was, if we now have "one true list" of dependencies (for which +1 1st_place_medal!) , will this also remove it from the conda deps ?

Yeah, from what I can see the ecosystem isn't quite there yet... well, not that I know of, and I'm certainly not an authority on this stuff.

I may ask around at scipy 2023 and see what the word is :+1:

requirements/optional-dev.txt
bjlittle: Summary of changes are to:
- rename `core.txt` to `pypi-core.txt` to reflect these are for `pip` and PyPI
- add a `requirements/README.md` akin to `requirements/locks/README.md`
- remove to `optional-*.txt` dependencies from the `requirements` directory and `pyproject.toml`, it's simply not a pattern that we use

trexfeathers: >Summary of changes are to:

(we agreed this offline)

pyproject.toml
trexfeathers: I also get problems, although they appear to be because it's trying to run `iris.__init__` in an environment without all the dependencies. Anyway I can confirm that build works fine without my suggestion.

pyproject.toml
```diff
@@ -8,6 +8,74 @@ requires = [
...
+]
+license = {text = "LGPL-3.0-or-later"}
+name = "scitools-iris"
+requires-python = ">=3.8"
+
+[project.urls]
+Code = "https://github.com/SciTools/iris"
+Discussions = "https://github.com/SciTools/iris/discussions"
+Documentation = "https://scitools-iris.readthedocs.io/en/stable/"
+Issues = "https://github.com/SciTools/iris/issues"
```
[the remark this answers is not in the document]
trexfeathers: We unfortunately couldn't find an implementation of this that would be sensitive to different versions (since this URL appears on the PyPI page). Should be OK without.

trexfeathers (approved):
OK this is ready to go, but needs a conflict resolved first.

bjlittle:
> OK this is ready to go, but needs a conflict resolved first.

On it now...

bjlittle:
@trexfeathers Hmmmm if only we had `auto-merge` enabled :wink: 

Ping @ESadek-MO

bjlittle:
@trexfeathers Awesome, cheers :beers:

PR #5274, opened 2023-04-24
Two negating pins for dependencies - libnetcdf and numpy

### LINKED ISSUE #5187: `HDF5-DIAG` warnings from libnetcdf `v4.9.1`
## 📰 Custom Issue

Since #5177, NetCDF loading is prompting verbose `HDF5-DIAG` warnings in some/all cases; [here is an example](https://github.com/SciTools/iris/actions/runs/4348704062/jobs/7597539846#step:12:618).

@pp-mo and I have done some testing and isolated this to the advancement of libnetcdf from `v4.8.1` to `v4.9.1`.

As far as we can tell this is probably innocuous, a view tentatively shared in pydata/xarray#7549. However it is going to start alarming developers and users alike, and will be a source of frustration given how verbose Iris' own warnings are as well. We've already had a case where it gets in the way of other print statements and thus makes ordinary work that much harder.

So far not sure what the next step is, but wanted to post this to forewarn everyone else.

### LINKED ISSUE #5245: Unpin libnetcdf
libnetcdf is currently pinned `<4.9` due to multiple issues introduced by `v4.9.1`. We should look to remove this pin at the earliest opportunity.

Related:

- #5182
- #5187
- #5213
- #5242

trexfeathers:
### `libnetcdf !=4.9.1`

Closes #5245, closes #5187. Thanks very much @zklaus for sorting this out in conda-forge/libnetcdf-feedstock#175 🌟. I've tested locally and `stdout` is no longer polluted with numerous HDF5 errors.

### `numpy !=1.24.3`

Updating the lockfile introduced errors in `iris.tests.test_analysis.TestLatitudeWeightGeneration`:

- `test_cosine_latitude_weights_2d`
- `test_cosine_latitude_weights_2d_latitude_first`
- `test_cosine_latitude_weights_2d_latitude_last`

Breadcrumb trail starts with numpy/numpy#23651. This relates to transposition order of masked arrays, and having looked at the way the tests are constructed, I don't believe they are asserting for the wrong answer.

---


trexfeathers:
>### `numpy !=1.24.3`

I can't be 100% sure if the below was originally written accounting for broken NumPy behaviour that has now been fixed. Anyone care to weigh in?

https://github.com/SciTools/iris/blob/1afae0d2e24b503c92a9c1b66ef311dca6cc2e28/lib/iris/analysis/cartography.py#L559-L569

requirements/py310.yml
```diff
@@ -15,10 +15,10 @@ dependencies:
   - cf-units >=3.1
   - cftime >=1.5
   - dask-core >=2022.9.0
-  - libnetcdf <4.9
+  - libnetcdf !=4.9.1
   - matplotlib >=3.5
   - netcdf4
-  - numpy >=1.19
```
bjlittle, line 21: @trexfeathers Whilst you're here could you bump this from `numpy >=1.19` to `numpy >=1.21`, which we're due to due according to the [NEP29 Drop Schedule](https://numpy.org/neps/nep-0029-deprecation_policy.html#drop-schedule)

bjlittle, line 21: @trexfeathers Whilst you're here could you bump this from `numpy >=1.19` to `numpy >=1.21`, which we're due to due according to the [NEP29 Drop Schedule](https://numpy.org/neps/nep-0029-deprecation_policy.html#drop-schedule)

bjlittle (changes requested):
@trexfeathers Just to confirm, from the PR description, are you saying that the lockfiles are generating issues with the new dependencies?

requirements/py310.yml
trexfeathers: c03fde8

bjlittle (approved):

bjlittle:
@trexfeathers We can watch `numpy` and see whether the `1.24.3` avoidance resurfaces in their next minor/patch release 👍

Issue #4795, opened 2022-06-13
Iris Documentation: Needs to be dark mode friendly 


tkknight:
## 📚 Documentation

`pydata-sphinx-theme` has been updated from `0.8.1` to `0.9.0`. The later version supports dark mode which is great, however the iris docs will need a few changes like image transparency, dark friendly logo, etc.

![image](https://user-images.githubusercontent.com/2108488/173373911-0a903d8c-00bf-4c5e-b977-9861e3206180.png)

The biggest item to fix is the **expandable panels** colours are not associated with the theme (always white).

![image](https://user-images.githubusercontent.com/2108488/173373741-9d4bfa07-5178-4f54-a0fa-407fdf203ff6.png)

Suggest we:

- [x] Pin to version 0.8.1 whilst we update the docs (https://github.com/SciTools/iris/pull/4796)
- [ ] Fix the mentioned issues
- [ ] Unpin

trexfeathers:
>dark friendly logo

#3935

rcomer:
Dark mode is already showing in the latest docs 😕 

![ECCD371D-3A6A-44AA-B6E9-91023FCC1501](https://user-images.githubusercontent.com/10599679/174013135-3ea1ec08-b7dd-4468-b31a-48ae380723b5.png)

tkknight:
I believe that is due to readthedocs (**latest**) using the **main** branch for the requirements, that use use the env definition https://github.com/SciTools/iris/blob/main/requirements/ci/readthedocs.yml (symlink to iris.yml -> [py38.yml](https://github.com/SciTools/iris/blob/main/requirements/ci/py38.yml#L50)).

PR https://github.com/SciTools/iris/pull/4796 to pin to the non dark mode version has not been merged just yet.  We have some CI issues atm so the PR will not pass - feel free to merge though, it is a minor change.

trexfeathers:
- SciTools/marketing#2

PR #5278, opened 2023-04-26
Dark logo

trexfeathers:
### Description

Ref: #4795 

---

### DIFF
--- a/docs/src/_static/README.md
+++ b/docs/src/_static/README.md
@@ -1,4 +1,4 @@
-# Iris  logos
+# Iris logos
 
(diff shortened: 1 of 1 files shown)


docs/src/_static/iris-logo.svg
trexfeathers, line 1: Text diffs on the existing logos suggest minor updates to how Python's [`ElementTree`](https://docs.python.org/3/library/xml.etree.elementtree.html) represents XML elements.

trexfeathers:
I can build the documentation locally, so I'm not sure what's going on with RTD taking forever at the build step.

PR #5289, opened 2023-04-28
Benchmark improvements, including passing exceptions upwards

trexfeathers:
### Description

WIP

---

### DIFF
--- a/benchmarks/bm_runner.py
+++ b/benchmarks/bm_runner.py
@@ -36,6 +36,7 @@ def _subprocess_run_print(args, **kwargs):
     # Use subprocess for printing to reduce chance of printing out of sequence
     #  with the subsequent calls.
     subprocess.run(["echo", f"BM_RUNNER DEBUG: {' '.join(args)}"])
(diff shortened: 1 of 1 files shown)


.github/workflows/benchmark.yml
```diff
@@ -74,11 +74,14 @@ jobs:
 
       - name: Benchmark this pull request
         if: ${{ github.event.label.name == 'benchmark_this' }}
+        # HEAD^2 gives us the HEAD of the PR's branch, instead of GitHub's
+        #  simulated merge commit.
         run: |
-          git checkout ${{ github.head_ref }}
+          git checkout HEAD^2
```
stephenworsley, line 80: This ends up comparing the head of the PR with the state of main when the PR branch was made. This does not capture any effects on performance caused by subsequent changes to main. In particular, this would fail to catch changes in performance caused by the *interaction* of different changes. Comparing the commit of the simulated merge with the previous commit in main would give a more accurate measurement of the performance changes you could expect in main due to merging the PR.

Issue #5255, opened 2023-04-18
CF global attribute "Conventions": expose to users, and include CF standard name table version


larsbarring:
## ✨ Feature Request
Currently, the global attribute `Conventions` is difficult to change. The default is hardcoded deep in the code:
https://github.com/SciTools/iris/blob/021d6ff7f719c600a189cfa5671bf32d20f66853/lib/iris/fileformats/netcdf/saver.py#L104 and changing the default is not as easy as one (at least I) would have hoped for, because it is "hidden" as `iris.config.netcdf.conventions_override = True`, which is difficult to find if I search for "global attribute Conventions" (or similar) in the documentation.

For our own use cases we would like to have better control of which CF version is actually written into the files, and moreover we would like to include additional conventions, like ACDD, Clix-meta version, as well as the CF Standard name Table version. Hence, this feature request has three main components:

* In `iris.save` allow users to pass the *kwarg* `Conventions='....'`.
* Change the `tools/generate_std_names.py` so that the `lib/std_names.py` includes the version number (and possibly related information such as creation date). 
* New functionality (in `iris.util`?) to get the default Conventions string, and to get the Standard name table version number from `lib/st_names.py`.

## Motivation
A more general motivation for giving the user (easier) access to setting the CF version is given by CF itself in [Section 2.6.1](http://cfconventions.org/Data/cf-conventions/cf-conventions-1.10/cf-conventions.html#identification-of-conventions):
> When CF is listed with other conventions, this asserts the same full compliance with CF requirements and interpretations as if CF was the sole convention. It is the responsibility of the data-writer to ensure that all common metadata is used with consistent meaning between conventions.

pp-mo:
Not to speak against the proposal, but I just thought it would be worth explaining here why we control this at all, and what it really means -- which I don't think is written down anywhere in user docs ...

In my view, at least, the only meaning that you can attach to the CF version which Iris writes in output files is "whatever we put in this file should be CF-compliant up to this CF version" :  
Effectively, that means **"the file does not include any CF encoding from CF versions later than this"**  -- and that's really all.

You can probably infer that an application interpreting all aspects of that CF version can understand everything that we have tried to state about the data.
But it **_doesn't_** (even) mean that the output is definitely valid under later CF versions -- though to date I think CF have totally avoided any backwards-invalidating changes.
Only that it may contain information which is not understood by _earlier_ versions.

pp-mo:
[...]

PR #5313, opened 2023-05-11
POC: Handle std names and aliases (#5257)

### LINKED ISSUE #5257: Handling of Standard name aliases
## ✨ Feature Request
When Iris is installed the file `lib/std_names.py` is generated from the xml version of the standard name table. In this process all *aliased* standard names are "promoted" to become on par with the all other standard names. 

Wouldn't it be more appropriate to have them replaced by the corresponding new standard names? I.e. that Iris can read/understand the aliased standard names, but translating them to the new ones which then are used in further processing and when writing data. 

## Motivation
To my understanding the idea behind aliasing a standard name is that the new standard name better/more precisely describes the quantity at hand, perhaps even taking new research results into account.

larsbarring:
### Description
This is a "proof-of-concept" PR to address how to better handle standard name aliases. It consists of the following elements:
*   Reorganising the iris.std_names.py a bit to have a separate dict for aliases (via updated `tools/generate_std_names.py`). 
It now includes some table version information (mentioned in  #5255), and a separate dict for the standard name descriptions (optional when generated).
*   Adding a new std_name_table.py containing the following functions:
    -  `get_convention` -- return a *tentative* Conventions string
    -  `set_alias_processing` -- define how to handle aliases: 
          "keep" - current behaviour, treat aliases in the same way as currently valid standard names,
          "warn" - issue a warning (default), otherwise as "keep", 
          "replace" - silently update aliases to current standard names.
    - `get_description` -- return the standard name description if available 
    - `check_valid_standard_name` -- check if a name is a standard name or an alias, and do the translation if requested as defined by `set_alias_processing`
* std_name_table is [naively] imported in `iris.__init__`
* `common/mixin._get_valid_standard_name` is modified to use `check_valid_std_name`

No units test have been added (would be good to first get some feedback whether this POC is a reasonable approach .... )

---


pp-mo:
@SciTools/peloton is this still alive issue, and if so could you sign the CLA @larsbarring ?

larsbarring:
I thought I had already signed the CLA because of one or two previous minor contributions. Anyway, now done. Whether it is alive or not I am not sure. I have not made any further effort since back then as I am not sure whether it is a reasonable approach in the context of Iris.

pp-mo:
> I thought I had already signed the CLA because of one or two previous minor contributions. Anyway, now done. Whether it is alive or not I am not sure. I have not made any further effort since back then ...

Ok thanks!

> I am not sure whether it is a reasonable approach in the context of Iris.

Well, I guess we'll take a look + see about it 
@larsbarring are you at least clear that _something_ like this would still be useful ?

larsbarring:
Yes, I think that this would still be useful. In the context of CF an aliased standard name is typically regarded as deprecated. Software should be able to read data having an aliased standard name, but new data should use the replacement name. Obviously, there are judgements to be made here regarding how to deal with this in practice. But and aliased standard name should not be considered just as an alternative at the same level as standard name.

Also, note that there will likely be some minor changes to the standard name xml file format as of next version, possibly also backported to all previous versions.

schlunma:
A proper way to handle standard name aliases would also be useful for ESMValTool, see https://github.com/ESMValGroup/ESMValCore/issues/1985.

bouweandela:
@LisaBock this may be of interest to you.

HGWright:
@SciTools/peloton  We are planning to talk about this on the 3rd July.

tools/generate_std_names.py
```diff
@@ -62,32 +77,72 @@ def process_name_table(tree, element_name, *child_elements):
...
+
+def write_useful_variables(outfile):
+    outfile.write(
+        '\n# The following variables are used for processing the standard names information below\n'
+        '_ACCEPT = "accept"\n'
+        '_WARN = "warn"\n'
+        '_REPLACE ="replace"\n'
+        '_ALTERNATIVE_MODES = [_ACCEPT, _WARN, _REPLACE]\n'
+        '_DEFAULT = "warn"\n'
+        '_MODE = _DEFAULT\n\n'
```
trexfeathers, line 117: We can see what you are trying to achieve here, but we would like to achieve it using Iris' existing structure instead:

### [`FUTURE` flag](https://github.com/SciTools/iris/blob/8604bb80977cb92bfe87d2486315311094219a31/lib/iris/__init__.py#L141)

Replacement is introduced as an opt-in setting. In future - typically the next major release - it will switch to being opt-out.

### Always warn if replacing

Since you made the proposal, we have agreed a global approach to Iris warnings: [Filtering Warnings — Iris 3.9.0 documentation (scitools-iris.readthedocs.io)](https://scitools-iris.readthedocs.io/en/stable/further_topics/filtering_warnings.html). The idea is that Iris should warn for EVERYTHING that might be a problem for a user, and the user applies warning filters if they don't wish to see a specific type of warning.

lib/iris/common/mixin.py
trexfeathers, line 1: We should consider whether the `rename()` method needs a switch to accept aliases, or some other equivalent granular user control.

@pp-mo I can't recall why using a `FUTURE` flag would not be enough user control on its own?

tools/generate_std_names.py
```diff
@@ -62,32 +77,72 @@ def process_name_table(tree, element_name, *child_elements):
...
+        version[elem] = found_or_none(tree.find(elem))
+    if version["table_name"] is None:
+        if (version["institution"] == "Centre for Environmental Data Analysis"
+                and version["contact"] == "support@ceda.ac.uk"):
+            version["table_name"] = "CF-StdNameTable"
+        else:
+            version["table_name"] = "USER-StdNameTable"
+    prettydict(outfile, "VERSION", version)
+    version_string = "-".join(version[k] for k in ["table_name", "version_number"])
+    outfile.write(f'CONVENTIONS_STRING = "{version_string}"\n\n')
```
trexfeathers, line 106: Note that this was sorted in #5423, albeit in a different way.

Presumably no longer necessary in this PR?

lib/iris/common/mixin.py
trexfeathers, line 1: For consistency with existing Iris code, we believe a new attribute should be introduced: `alias_standard_name`, used in a similar way to `invalid_standard_name`.

This gives users the level of information and control that they are used to.

#### `invalid_standard_name` example

https://github.com/SciTools/iris/blob/8604bb80977cb92bfe87d2486315311094219a31/lib/iris/fileformats/_nc_load_rules/helpers.py#L463-L470

trexfeathers (changes requested):
> @SciTools/peloton We are planning to talk about this on the 3rd July.

@pp-mo and @trexfeathers met on 3rd July. This review is the summary of that meeting.

@larsbarring we really appreciate you taking the time to make this suggestion. It's obviously bad that it's taken us this long to give it any attention. If your priorities have shifted and you are unable to action our comments then please let us know and we'll pick it up.

To prompt more attention from the team, we're adding this to the [3.11 milestone](https://github.com/SciTools/iris/milestone/51).

PR #5322, opened 2023-05-17
Added whatsnew notes on netcdf delayed saving and Distributed support.

pp-mo:
Here's my suggestion explaining the benefits of delayed saves, and Distributed.

We _could_ consider a new UserGuide section for this, but I think we can still argue that this is a special usage + doesn't belong there.

### DIFF
--- a/docs/src/whatsnew/3.6.rst
+++ b/docs/src/whatsnew/3.6.rst
@@ -13,14 +13,37 @@ This document explains the changes made to Iris for this release
    :animate: fade-in
    :open:
 
-   We're so excited about our recent support for **delayed saving of lazy data
-   to netCDF** (:pull:`5191`) that we're celebrating this important step change
-   in behavour with its very own dedicated release 🥳
+We're so excited about our recent support for **delayed saving of lazy data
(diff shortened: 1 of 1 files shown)


pp-mo:
@bjlittle think you suggested this ?

docs/src/whatsnew/3.6.rst
```diff
@@ -13,14 +13,37 @@ This document explains the changes made to Iris for this release
...
+
+Usage example::
+
+   # Create output files with delayed data saving.
+   delayeds = [
+       iris.save(cubes, filepath, compute=False)
+       for cubes, filepath in zip(output_cubesets, output_filepaths)
+   ]
+   # Complete saves in parallel.
+   dask.compute(delayeds)
```
bjlittle, line 34: `dask.compute(*delayeds)`, no?

See https://docs.dask.org/en/stable/api.html?highlight=dask.compute#dask.compute

bjlittle (approved):
@pp-mo Awesome, this is much better, with enough detail to sell the feature.

Thanks :100:

PR #5161, opened 2023-02-15
Adding a CITATION.cff file

HGWright:
### Description

I have written the CITATION.cff file for Iris using the information from the [citation page](https://scitools-iris.readthedocs.io/en/latest/userguide/citation.html) of the documentation.

This pull request is a WIP and will require changes based on the answers to some questions that I have below.

resolves #5118

### Questions for @SciTools/iris-devs

1. In the documentation, currently the BibTeX entry begins with @manual tag. When looking through the old PR's #389 and #5116, I cannot find any consideration being given to the @software tag even though to me this seems like the most natural fit. Could you let me know the reasons behind this choice, and I can update the CITATION.cff to reflect @manual if you feel that is still the best fit.

2. With the existence of this file, what should become of the existing citation page of the documentation?
- My thought is that the documentation should in some way link to or reference the github to obtain the references rather than also manually listing them out in the documentation as currently it would be a duplication and so both would need to be maintained.

3. Where would be the best place for this in the 'What's New' section of the documentation?
- 3.5 Given the potential overhaul of the citation page in the documentation what should happen to the existing entry in the documentation section of the 'What's New' "3. [@jonseddon](https://github.com/jonseddon) updated the citation to a more recent version of Iris. ([PR #5116](https://github.com/SciTools/iris/pull/5116))".

---
[Consult Iris pull request check list]( https://scitools-iris.readthedocs.io/en/latest/developers_guide/contributing_pull_request_checklist.html)

PR #5324, opened 2023-05-18
update bibtex citation for v3.6.0 release

bjlittle:
### Description

This PR updates the documentation BibTex citation post release v3.6.0 and prior merge-back to `main`.

See [DOI](https://doi.org/10.5281/zenodo.7948293)

---

### DIFF
--- a/docs/src/userguide/citation.rst
+++ b/docs/src/userguide/citation.rst
@@ -15,12 +15,12 @@ For example::
 
  @manual{Iris,
  author = {{Met Office}},
- title = {Iris: A powerful, format-agnostic, and community-driven Python package for analysing and visualising Earth science data },
- edition = {v3.5},
+ title = {Iris: A powerful, format-agnostic, and community-driven Python package for analysing and visualising Earth science data},
+ edition = {v3.6},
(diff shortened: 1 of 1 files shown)


bjlittle:
I must confess, I do consider the current BibTex citation workflow within the documentation slightly broken as it stands.

The `stable` version of the documentation will always cite the **previous** release of `iris`. I don't think that's ever so useful.

Meh. Maybe that's just me.

trexfeathers:
> I must confess, I do consider the current BibTex citation workflow within the documentation slightly broken as it stands.
> 
> The `stable` version of the documentation will always cite the **previous** release of `iris`. I don't think that's ever so useful.
> 
> Meh. But maybe that's just me 😐

Perhaps we could roll that problem into #5161? Maybe all we need these days is the `.cff`...

bjlittle:
> Perhaps we could roll that problem into #5161? Maybe all we need these days is the `.cff`...

Yup. Totally.

👍

ESadek-MO (approved):
Thanks @bjlittle, looks good to me!

PR #5335, opened 2023-05-31
DOC: remove advice to delete main branch

rcomer:
### Description

Our dev guide advocates deleting your main branch, with a link to this [pretty old page](https://matthew-brett.github.io/pydagogue/gh_delete_master.html).  I note that

* A quick browse of Iris core devs' forks suggests that not many of us ever followed this advice ourselves.
* The linked advice refers to `master` rather than `main`.  It may not be obvious for someone new to OSS that this is just an old name for the same thing.  Also there were good EDI reasons for dropping "master".
* The described procedure is fairly involved.  These days if you really wanted your default branch to be called something different, you could just rename it within GitHub.

So I think this advice is no longer useful.  Can we remove it?

---

### DIFF
--- a/docs/src/developers_guide/gitwash/development_workflow.rst
+++ b/docs/src/developers_guide/gitwash/development_workflow.rst
@@ -14,8 +14,7 @@ Workflow Summary
 In what follows we'll refer to the upstream iris ``main`` branch, as
 "trunk".
 
-* Don't use your ``main`` (that is on your fork) branch for anything.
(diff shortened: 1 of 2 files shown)


pp-mo (approved):
OK, seems reasonable !

I personally still think it's a thing worth using to avoid confusion, but it may not help the discussion here. 
Interestingly all this was originally copied from "gitwash".  Which does seem to be getting rather old.

Issue #5338, opened 2023-06-06
`weigths` are always realized by the `iris.analysis` module


bouweandela:
## How To Reproduce
Steps to reproduce the behaviour:

Recent versions of iris realize the weights arrays. It looks like the issue was introduced in #5084, so iris versions since 3.5 are affected.

Example:

Use `cube.collapsed(aggregator=iris.analysis.MEAN, weights=weights, coords=['latitude', 'longitude'])` where `cube` is an `iris.cube.Cube` and `weights` is a `dask.array.Array`. This will issue a warning like
```
/home/bandela/mambaforge/envs/test-iris-3.6/lib/python3.11/site-packages/distributed/client.py:3109: UserWarning: Sending large graph of size 386.72 MiB.
This may cause some slowdown.
Consider scattering data ahead of time and using futures.
  warnings.warn(
```
because the weights are realized in this code:
https://github.com/SciTools/iris/blob/13999940a045dc3a4496aa12e8867142d2377719/lib/iris/analysis/__init__.py#L1190-L1291

## Expected behaviour

The laziness of the weights array should be preserved. Because the weights array must be the same size as the data (on a side note: why does it need to be the same size? is this a limitation of numpy?), this makes it impossible to use this feature on large datasets.

## Environment 
 - OS & Version: Ubuntu 23.04
 - Iris Version: 3.5, 3.6

schlunma:
I thought about this for a while now. I don't think it's feasible to subclass `dask.core.Array` consistently similar to what is done with `np.ndarray` here (it would be possible to simply set the `units` attribute for Dask arrays; however, this is lost after operations like indexing or calling `compute`).

The easiest solution to this is probably to not subclass `np.ndarray` anymore but rather use to two attributes `array` and `units` for `_Weights`. `_Weights.array` contains either the numpy or dask array. This requires some changes in the code (we need to replace `weights` with `weights.array` everywhere), but this ensures that the weights are not realized if not necessary.

I'll try to come up with a PR for this.

ESadek-MO:
This was fixed in v3.6.1 release.

PR #5307, opened 2023-05-05
Simplify and lazify broadcast_to_shape

rcomer:
### Description

Attempt to resolve #5295, following my own suggestion at https://github.com/SciTools/iris/issues/5295#issuecomment-1532791248.  It turned out I didn't need to use `da.broadcast_to` directly - this is automatic due to dask's support for NEP13/18.  We *do* still need to handle masked arrays separately because they don't (yet) support NEP13/18.

This is still not working for the case where the array is both lazy and masked (see failing new test) because the mask gets ignored by `da.broadcast_to`.  We could apply a similar approach to that for the numpy masked arrays, but you can't tell whether a dask array is masked until you realise it.  This seems like it must be a common problem.  Anyone know a solution before I attempt to re-invent the wheel?

#### Performance Note

For a simple array I found this new implementation takes nearly twice as long as `main`.  However we are only talking about ~4 microseconds, and it does not appear to depend on array size, so maybe it's OK?

---
[Consult Iris pull request check list]( https://scitools-iris.readthedocs.io/en/latest/developers_guide/contributing_pull_request_checklist.html)

Issue #5295, opened 2023-05-02
`iris.util.broadcast_to_shape` does not work with lazy data


bouweandela:
## How To Reproduce
Steps to reproduce the behaviour:

Run the following code:
```python
>>> import dask.array as da
>>> from iris.util import broadcast_to_shape
>>> a = da.array([[1, 2, 3], [4, 5, 6]])
>>> b = broadcast_to_shape(a, (5, 2, 6, 3), (1, 3))
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "/home/bandela/mambaforge/envs/test/lib/python3.11/site-packages/iris/util.py", line 98, in broadcast_to_shape
    array_view = _broadcast_helper(array)
                 ^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/bandela/mambaforge/envs/test/lib/python3.11/site-packages/iris/util.py", line 95, in _broadcast_helper
    strides[dim] = a.strides[idim]
                   ^^^^^^^^^
AttributeError: 'Array' object has no attribute 'strides'
```

## Expected behaviour

The [function documentation](https://scitools-iris.readthedocs.io/en/v3.5.0/generated/api/iris/util.html#iris.util.broadcast_to_shape) mentions that "This function maintains laziness when called; it does not realise data.", so I would expect the function to return a Dask array instead of crash when following the example from the docstring with a dask array as input.

## Environment 
 - OS & Version: Ubuntu 23.04
 - Iris Version: 3.5

```
$ mamba env export 
```
gives
```
name: test
channels:
  - conda-forge
dependencies:
  - _libgcc_mutex=0.1=conda_forge
  - _openmp_mutex=4.5=2_gnu
  - antlr-python-runtime=4.7.2=py311h38be061_1003
  - brotli=1.0.9=h166bdaf_8
  - brotli-bin=1.0.9=h166bdaf_8
  - brotlipy=0.7.0=py311hd4cff14_1005
  - bzip2=1.0.8=h7f98852_4
  - c-ares=1.18.1=h7f98852_0
  - ca-certificates=2022.12.7=ha878542_0
  - cartopy=0.21.1=py311hd88b842_1
  - certifi=2022.12.7=pyhd8ed1ab_0
  - cf-units=3.2.0=py311hcb2cf0a_0
  - cffi=1.15.1=py311h409f033_3
  - cftime=1.6.2=py311h4c7f6c3_1
[...]

PR #5341, opened 2023-06-07
Make weighted aggregation lazy again

schlunma:
### Description

https://github.com/SciTools/iris/pull/5084 broke lazy weighted aggregation by realizing all input arrays. This PR fixes that by

- refactoring `Weights` so it does **not** subclass `np.ndarray` anymore. Instead, it now provides the attributes `array` and `units`. `array` (which is either lazy or non-lazy depending on the input) is then provided as `weights` keyword argument.
- explicitly using the private keyword argument `_weights_units` when calling `Aggregator.update_metadata` (which in turn calls `units_func`). Apart from that, the weights units are **not** used/provide anywhere else.

Closes https://github.com/SciTools/iris/issues/5338

---


schlunma:
@bouweandela could you test if this solves https://github.com/SciTools/iris/issues/5338 for you?

schlunma:
> Thanks for fixing this so quickly @schlunma! Indeed it solves my issue.

That's great to hear @bouweandela, thanks for testing!

> Regarding the implementation: instead of implementing a custom `_Weights` class, it might be nicer to just pass the weights through to the analysis operator if they are an `iris.coords.AncillaryVariable` or a `iris.coords.CellMeasure` already, or convert them to an `iris.coords.AncillaryVariable` if they are not. But I'll leave that to other @SciTools/iris-devs to comment on.

I am not quite sure I understand what you mean here. The aggregators themselves only accept arrays as input, so I am not quite sure why I would convert them to `iris.coords.AncillaryVariable`. Of course the `_Weights` class is not strictly necessary anymore with these changes, but I think it's a nice way of converting the weights keyword argument (which can be array, cube, str, cell measure, etc.) to an array and units.

bouweandela:
> I am not quite sure I understand what you mean here. The aggregators themselves only accept arrays as input, so I am not quite sure why I would convert them to iris.coords.AncillaryVariable. Of course the _Weights class is not strictly necessary anymore with these changes, but I think it's a nice way of converting the weights keyword argument (which can be array, cube, str, cell measure, etc.) to an array and units.

Thanks for explaining. I was confused because a _Weights instance is created but then it is not used anymore. For me, it would be easier to understand the code if the class was replaced by a single function.

lib/iris/tests/test_analysis.py
```diff
@@ -1926,18 +1885,21 @@ def test__Groupby_repr():
...
-        ({"weights": iris.analysis._Weights([1], CUBE)}, "s"),
-        ({"weights": iris.analysis._Weights([1], CUBE, units="kg")}, "s kg"),
+        ({"_weights_units": None}, "s"),
+        ({"test": "m", "_weights_units": None}, "s"),
+        ({"weights": None, "_weights_units": None}, "s"),
+        ({"weights": [1, 2, 3], "_weights_units": None}, "s"),
+        ({"_weights_units": "kg"}, "s"),
+        ({"test": "m", "_weights_units": "kg"}, "s"),
+        ({"weights": None, "_weights_units": "kg"}, "s"),
+        ({"weights": [1, 2, 3], "_weights_units": "kg"}, "s kg"),
```
trexfeathers, line 1902: FYI `parametrize` can do a matrix, if you're looking to do an exhaustive test of all combinations. That wouldn't work with the `expected` element, but since only one of these combinations is expected to return `"s kg"` you could probably put a conditional within the test.

Not a deal breaker for me.

lib/iris/tests/test_analysis.py
```diff
@@ -1731,188 +1733,145 @@ def setup_test_data(self):
...
+            kwargs, self.cube
         )
-        assert isinstance(weights, iris.analysis._Weights)
-        np.testing.assert_array_equal(weights, np.zeros((5, 5)))
-        assert weights.units == "J"
+        assert new_kwargs is not kwargs
+        assert new_kwargs == {"test": [1, 2, 3], "weights": None}
+        assert weights_units is None
 
-    def test_init_with_invalid_obj(self):
```
trexfeathers, line 1890: I know the circumstances have changed a lot, but shouldn't we still have a test for an object that is of no use? E.g. a `str` or a `list`?

trexfeathers (changes requested):
Sorry this took so long @schlunma. The underlying code-base made it quite confusing to see what is going on, and even once I had, I found it difficult to come up with suggestions to make it less confusing!

As well as my comments below, this needs a What's New entry.

schlunma:
Thanks for your feedback @trexfeathers, I am currently implementing you suggestions!

I noticed that my tests are failing locally now because of https://github.com/SciTools/iris/issues/5295. Turns out the fix for that (https://github.com/SciTools/iris/pull/5307) is not included in `v3.6.x`, but only in `main`. No idea why the CI machine is not picking that up though.

Would it be possible to include https://github.com/SciTools/iris/pull/5307 to `v3.6.x`? I think it essential to broadcast lazy arrays for this PR.

lib/iris/tests/test_analysis.py
schlunma: True! However, including all test cases (e.g., also the one with an empty dict) would add quite some additional code, so I would prefer to keep it like that in terms of readability.

lib/iris/tests/test_analysis.py
```diff
@@ -1731,188 +1733,145 @@ def setup_test_data(self):
...
 
-    def test_update_kwargs_weights_none(self):
-        kwargs = {"test": [1, 2, 3], "weights": None}
-        iris.analysis._Weights.update_kwargs(kwargs, self.cube)
-        assert kwargs == {"test": [1, 2, 3], "weights": None}
+    @pytest.fixture(autouse=True)
+    def setup_test_data(self):
+        self.data = da.arange(6).reshape(2, 3)
+        self.lat = iris.coords.DimCoord(
+            da.from_array([0, 1]), standard_name="latitude", units="degrees"
```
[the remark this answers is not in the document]
schlunma: Done in https://github.com/SciTools/iris/pull/5341/commits/0e5b625010eb9544e97572b46919df96e801b4f3

lib/iris/tests/test_analysis.py
schlunma: Added in https://github.com/SciTools/iris/pull/5341/commits/fb466ba6c1804e79dad81a06f630ddf17494ec79

trexfeathers:
>Would it be possible to include https://github.com/SciTools/iris/pull/5307 to `v3.6.x`? I think it essential to broadcast lazy arrays for this PR.

@bjlittle thoughts?

bjlittle:
> @bjlittle thoughts?

@trexfeathers If it helps this use-case cross the line, then it makes sense to me :+1: 

Technically, #5307 is adding a new feature rather than patching, so it's inclusion is somewhat blurring the lines of the patch/minor release... however, I think that we can defend it's inclusion here.

@SciTools/iris-devs Any other core dev opinions on this proposal?

rcomer:
If you wanted to be strict about the versioning, the other option would be to cut another bonus minor release.

The question for me is whether including #5307 is likely to break anyone’s workflow.  Since #5295 shows that old version of `broadcast_to_shape` barfs when it gets a dask array, I think it’s unlikely that anyone is relying on that behaviour.

trexfeathers:
> Thanks for your feedback @trexfeathers, I am currently implementing you suggestions!
> 
> I noticed that my tests are failing locally now because of #5295. Turns out the fix for that (#5307) is not included in `v3.6.x`, but only in `main`. ~No idea why the CI machine is not picking that up though.~ _EDIT: Tests are failing now as expected._
> 
> Would it be possible to include #5307 to `v3.6.x`? I think it essential to broadcast lazy arrays for this PR.

As part of my review I have replicated @schlunma's findings about #5307 on my own machine.

trexfeathers:
> One potential problem with the current implementation is that the multiplication by `Unit('1')` that is used when a regular array is used as weights might change the units in an undesired ways, e.g.,
> 
> ```python
> >>> from cf_units import Unit
> >>> Unit('kg m-2 s-1') * Unit('1')
> Unit('m-2.kg.s-1')
> ```
> 
> Would it make sense to not multiply the units in this case to be strictly backwards-compatible?

So these two forms are functionally identical, but the second is in the UDUNITS-compliant form. We are hoping to move away from Iris modifying files to make them compliant, so I'm in favour of not multiplying the units if you can make that change. Thanks for spotting

rcomer:
I agree it would be confusing for users if just summing along a dimension changed their units (particular if an unweighted sum does not change the units, but I have not looked in detail at this change).

trexfeathers:
@schlunma: [`v3.6.x`](https://github.com/SciTools/iris/tree/v3.6.x) has now been updated (#5359), so ready for you to update this PR's `HEAD`.

trexfeathers (approved):

trexfeathers:
Thanks everyone for your hard work on this! @schlunma @rcomer @bouweandela @bjlittle

PR #5367, opened 2023-07-03
Move Iris test data version into code

lbdreyer:
Currently the Iris test data version that is used is defined at CI level. I think it would make more sense to be defined within the Iris code, which we could then read.

This came up when I was creating an environment with Iris in it that I then wanted to run the tests in. My workflow was to find out what Iris version was installed, go to that tag on Github, to see what version of the iris test data was used. Now I want to automate it, it would be easier to simply get the value of a variable from my installation of Iris.

It would also be beneficial for the CI of other repos that also install the test data repo, where we currently have to hard code the version of iris test data that is required, everywhere it is used (similarly to how we have it in both the `ci-tests` and `benchmarks` in this repo

This PR moves the test data version into `iris.tests`, and also updates the benchmarks CI to use the iris test data cache composite action


lbdreyer:
It occurred to me that we can't get the data version out using python (i.e.iris.tests.TEST_DATA_VERSION) because Iris isn't actually installed yet. Nox handles the installation of Iris when it runs the test. We could use it by parsing the file but this is starting to feel a bit hacky, so I'm going to close and reconsider

trexfeathers:
I tried and failed to do something similar with the various Python versions that are set all over the codebase. In the end I relied on 'automating humans', by introducing [`test_python_versions`](https://github.com/SciTools/iris/blob/dab88e8bd7f76f9bb82ee07b12b1264e56c5bc5e/lib/iris/tests/test_coding_standards.py#L69-L133). Perhaps something analogous is possible in this space.

PR #5381, opened 2023-07-19
Ignore sticklerci

lbdreyer:
Linkcheck is now broken because https://stickler-ci.com/ is down.

It looks like that website has now been shutdown:
image


tkknight (approved):

lbdreyer:
Thanks @tkknight !

Issue #4986, opened 2022-09-23
split-attrs: netcdf saver


pp-mo:
* changes to netcdf Saver code
    - when saving to nc files, make it respect the local/global status of attributes where possible
      (i.e. typically, if loaded from an original nc file, the local/global status as it was there -- cf https://github.com/SciTools/iris/issues/4984 )
* (plus) : a new FUTURE flag implementation to switch between new/old handling behaviour
   - to match existing behaviour AFAP -- cf status-quo tests https://github.com/SciTools/iris/issues/4981

pp-mo:
In this case, we should get different behaviour --at last !
And we will want to expand the tests.

First, I'm expect to add a new Pytest parameter over "FUTURE.nc_save_splitattrs = (False, True)" to existing tests.
I'm hoping that the differences between old+new results can be captured in the existing testcases (cf #4981)
  * duplicate old+new tests with parameterisation
  * testcode switches on parameter to capture differences between old+new

Also, some specific new behaviour might require additional tests, to run only with "new-style" saving.

pp-mo:
NOTE: as noted [here](https://github.com/SciTools/iris/issues/5255#issuecomment-1513395166), I think the special status of attributes called "conventions" (small-c) is probably a mistake, and can be remove.

It was introduced in #711 -- before that, there was already code which would handle "conventions" or "Conventions" specially, using "str.lower()" in several places.  
I don't believe lower-case-"conventions" itself was ever a CF term, and I suspect this was probably just defensive programming.
Goes all the way back to [the initial commit](https://github.com/SciTools/iris/blob/61ac271d425eb6037a8e9d3be2458f02f7de111d/lib/iris/fileformats/netcdf.py#L619) !

pp-mo:
#5410 is now the PR for this.

pp-mo:
#5410 replaces

pp-mo:
closed by #5410

PR #5384, opened 2023-07-20
Splitattrs ncload

pp-mo:
Changes netcdf loading code to specifically load into `cube.attributes.globals` and `cube.attributes.locals`
Since the same attributes are then appearing in 'cube.attributes', this in itself (i.e. without changes to saving) should have no effect.
-- **_Except_** in some minor corner cases, where an attribute is present as _both_ a local (variable-attribute)) + global (file-attribute).

So, I think that this is (almost) fully backwards-compatible, and **_does not require a control switch like a FUTURE flag_** (but reviewer may think differently).
( N.B. I **_am_** intending to create a FUTURE flag for the new split-attributes saving behaviour, since that will definitely change behaviour quite significantly. Covered in #4986 )

The actual changes in loading code are pretty small (https://github.com/SciTools/iris/commit/f736c4baa56bbf14a6a90d1e21175ada7c008ac7)

I have had to modify change some of the existing tests from #4960, by replacing `cube.attributes == <expected-result-dictionary>` with `dict(cube.attributes) == <expected-result-dictionary>`.  In principle I could have done that first, to prove pre-existing behaviour, but I think the argument is pretty solid as it is.

( Note : replaces older version  #5048 )

### DIFF
--- a/docs/src/further_topics/metadata.rst
+++ b/docs/src/further_topics/metadata.rst
@@ -131,7 +131,7 @@ We can easily get all of the associated metadata of the :class:`~iris.cube.Cube`
 using the ``metadata`` property:
 
     >>> cube.metadata
-    CubeMetadata(standard_name='air_temperature', long_name=None, var_name='air_temperature', units=Unit('K'), attributes=CubeAttrsDict(globals={}, locals={'Conventions': 'CF-1.5', 'STASH': STASH(model=1, section=3, item=236), 'Model scenario': 'A1B', 'source': 'Data from Met Office Unified Model 6.05'}), cell_methods=(CellMethod(method='mean', coord_names=('time',), intervals=('6 hour',), comments=()),))
+    CubeMetadata(standard_name='air_temperature', long_name=None, var_name='air_temperature', units=Unit('K'), attributes=CubeAttrsDict(globals={'Conventions': 'CF-1.5'}, locals={'STASH': STASH(model=1, section=3, item=236), 'Model scenario': 'A1B', 'source': 'Data from Met Office Unified Model 6.05'}), cell_methods=(CellMethod(method='mean', coord_names=('time',), intervals=('6 hour',), comments=()),))
 
 We can also inspect the ``metadata`` of the ``longitude``
 :class:`~iris.coords.DimCoord` attached to the :class:`~iris.cube.Cube` in the same way:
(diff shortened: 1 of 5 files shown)


trexfeathers:
@pp-mo can you clarify for me: [`cf_attrs_unused()`](https://github.com/SciTools/iris/blob/2a9e131875be42c39ab4b8fc9956e4a4d59f83a2/lib/iris/fileformats/cf.py#L215-L220) is everything that doesn't get interpreted during NetCDF load? I.e. excluding `coordinates`, `standard_name`, etcetera?

pp-mo:
> I also _think_ the list of used attributes should be reset before _each_ cube is processed.

From experiment, no it does not do that, but I don't think it is a mistake.
It is the list of attributes which are not referenced during the initial "categorisation" of variables by 'fileformats/cf.py', and is **_not_** related to what happens in rules processing.  It is static after that, **not** reset for each cube (before running rules).

The "unused_attributes" is only called [from "loader._add_unused_attributes", and there it excludes any cf-specific attributes according to a list of 'recognised names'](https://github.com/SciTools/iris/blob/v3.6.1/lib/iris/fileformats/netcdf/loader.py#L149-L162).
"_add_unused_attributes" is [called once per built data object (cubes, coords, etc)  during cube loading](https://github.com/SciTools/iris/blob/v3.6.1/lib/iris/fileformats/netcdf/loader.py#L279-L291) 

So the list can't change, because it is used repeatedly (e.g. to build a new replica dimcoord for each data cube)

pp-mo:
> So this list _mustn't_ change, because it is used repeatedly (e.g. to build a new replica dimcoord for each data cube). I guess the point is that these "unused" attributes are only fetched _by_ calling `cf_var.cf_attrs_unused()`, and **_not_** through `cf_var.__getattr__`, so they don't get "crossed off" the list.

FTR  I'm not a great fan of this code, due to the slightly 'hidden' way it does some things!  
I don't like that the CFVariable classes acts as "wrappers" to the underlying netcdf variables, and I don't think providing access to the 'real' info via CFVariable.__getattr__ is a terribly good idea -- I think it would be altogether clearer if all such accesses were made explicitly with "cf_var.cf_data.getncattr()".

I also find the naming of the "cf_var" and "cf_data" arguments in this code constantly confusing -- effectively "sometimes" cf means the 'fileformats/cf' module, and sometimes it means netCDF4 !

trexfeathers (approved):
Thanks for investigating the unused attributes @pp-mo. I agree it's confusing!

I remain surprised that it has taken so few changes to get the loading code working, however the tests are proof that you haven't missed anything.

I don't think there's any need to cover these changes with a `FUTURE` flag, since they only affect edge cases and not in a catastrophic way.

pp-mo:
Brilliant, thanks for your careful approach @trexfeathers !

PR #5388, opened 2023-07-25
Added link to UDUNITS-2 database to convert_units docstrings

### LINKED ISSUE #3945: No documentation on what units cube.convert_units actually converts?
## 📚 Documentation

While the documentation points at the convert_units() function for cube unit conversion, it never actually tells you what can be converted. There's a few examples scattered over the documentation, but you can't know if convert_units("rankine") will work without testing, and you don't know if the error is because it should be convert_units("Rankine") or iris just doesn't support terrible units - there should be a list (or link to another list somewhere) of what type of units exist under cf_units and can be converted between. I found this quite confusing when first starting to use iris this year

acchamber:
## 🚀 Pull Request
To close #3945 
### Description

Adds a external hyperlink to convert_units methods for cube and coord to UDUNITS-2 documentation containing the information on what units and aliases for them are supported. 

The page in question has the information spread over 5 .xml files, but unless we want to recreate them in the cf-units documentation it's the closest we have to a list in the browser of supported units.

### DIFF
--- a/docs/src/whatsnew/latest.rst
+++ b/docs/src/whatsnew/latest.rst
@@ -75,6 +75,9 @@ This document explains the changes made to Iris for this release
    section into the user guide, containing advice and use cases to help users
    get the best out of Dask with Iris.
 
+#. `@acchamber`_ improved documentation for :meth:`~iris.cube.Cube.convert_units`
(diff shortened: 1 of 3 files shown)


stephenworsley (approved):
Looks good.

PR #5389, opened 2023-07-26
Include location coordinates in the data variable coordinates attribute

### LINKED ISSUE #5206: Saving a cube with a mesh does not include coordinate attribute
## 🐛 Bug Report
This is the distilled version of #5202.  

For consistency with current CF/UGrid conventions, UGrid data variables should include a `coordinates` attribute referencing the mesh face coordinates.

## How To Reproduce

This is very slightly edited from #5202, but note that the loading part of this example is solely to generate a cube with a mesh.  It's the save part that is demonstrating the actual issue: i.e. this is not about preserving data by roundtripping.

    In [1]: import iris
    
    In [2]: # Using scitools default version – same behaviour is seen with default-next too.
    
    In [3]: iris.__file__
    Out[3]: '.../default-2022_11_22/lib/python3.9/site-packages/iris/__init__.py'
      
    In [7]: from iris.experimental.ugrid import PARSE_UGRID_ON_LOAD
    
    In [8]: with PARSE_UGRID_ON_LOAD.context():
       ...:     cubes = iris.load('./lib/ants/tests/resources/stock/data_C4.nc')
       ...: 
    
    In [9]: iris.save(cubes, 'result.nc')
    
    In [10]: # Notice that the coordinate attributes are not included on the data variables, but only on the topology variable.
    
    In [11]: !ncdump -h result.nc | grep 'coordinates'
               example_C4:node_coordinates = "example_C4_node_x example_C4_node_y" ;
               example_C4:face_coordinates = "example_C4_face_x example_C4_face_y" ;
    
    In [12]: # Full ncdump for completeness
    
    In [13]: !ncdump -h result.nc
    netcdf result {
    dimensions:
         num_node = 98 ;
         dim0 = 96 ;
         example_C4_face_N_nodes = 4 ;
    variables:
         int example_C4 ;
               example_C4:cf_role = "mesh_topology" ;
               example_C4:topology_dimension = 2 ;
               example_C4:long_name = "Topology data of 2D unstructured mesh" ;
               example_C4:node_coordinates = "example_C4_node_x example_C4_node_y" ;
               example_C4:face_coordinates = "example_C4_face_x example_C4_face_y" ;
               example_C4:face_node_connectivity = "face_nodes" ;
         double example_C4_node_x(num_node) ;
               example_C4_node_x:units = "degrees_east" ;
               example_C4_node_x:standard_name = "longitude" ;
               example_C4_node_x:long_name = "longitude of 2D mesh nodes." ;
         double example_C4_node_y(num_node) ;
               example_C4_node_y:units = "degrees_north" ;
               example_C4_node_y:standard_name = "latitude" ;
               example_C4_node_y:long_name = "latitude of 2D mesh nodes." ;
         double example_C4_face_x(dim0) ;
               example_C4_face_x:units = "degrees_east" ;
               example_C4_face_x:standard_name = "longitude" ;
               example_C4_face_x:long_name = "longitude of 2D face centres" ;
         double example_C4_face_y(dim0) ;
               example_C4_face_y:units = "degrees_north" ;
               example_C4_face_y:standard_name = "latitude" ;
               example_C4_face_y:long_name = "latitude of 2D face centres" ;
         int face_nodes(dim0, example_C4_face_N_nodes) ;
               face_nodes:long_name = "Maps every quadrilateral face to its four corner nodes." ;
               face_nodes:Conventions = "UGRID-1.0" ;
               face_nodes:cf_role = "face_node_connectivity" ;
               face_nodes:start_index = 1 ;
         double sample_data(dim0) ;
               sample_data:long_name = "sample_data" ;
               sample_data:units = "1" ;
               sample_data:mesh = "example_C4" ;
               sample_data:location = "face" ;
         double additional_sample_data(dim0) ;
               additional_sample_data:long_name = "additional_sample_data" ;
               additional_sample_data:units = "1" ;
               additional_sample_data:mesh = "example_C4" ;
               additional_sample_data:location = "face" ;
    
    // global attributes:
               :online_operation = "once" ;
               :Conventions = "CF-1.7" ;
    }

## Expected behaviour

I would expect iris to include the `coordinates` attribute with a value similar to `"example_C4_face_x example_C4_face_y"` on both of the data variables.  This would be consistent with netCDF behaviour for other coordinates (e.g. time).  The real use case where this was discovered was a file with pseudo levels, where an output file with a data variable having a `coordinates` attribute of just `pseudo_level` was unexpected.

## Environment 
 - OS & Version: RHEL 7
 - Iris Version: Iris Version: 3.3.1 and 3.4.0

## Additional context

There is some scope for interpretation in the UGrid and CF conventions - see the conversation in #5202 for links to related discussions.  Current consensus is that the coordinates attribute is required for CF compatibility, and that iris save should be compatible with both CF and UGrid conventions.

</details>

trexfeathers:
### Description

Closes #5206

For all its complexity, the NetCDF saver code is nicely put together - took a lot of reading but very little change to get this to work.

---

### DIFF
--- a/lib/iris/fileformats/netcdf/saver.py
+++ b/lib/iris/fileformats/netcdf/saver.py
@@ -895,6 +895,9 @@ class Saver:
                                 coord,
                                 element_dims=(mesh_dims[location],),
                             )
+                            # Only created once per file, but need to fetch the
+                            #  name later in _add_inner_related_vars().
+                            self._name_coord_map.append(coord_name, coord)
                             coord_names.append(coord_name)
                         # Record the coordinates (if any) on the mesh variable.
                         if coord_names:
@@ -1024,15 +1027,32 @@ class Saver:
             Names associated with the dimensions of the cube.
 
         """
+        from iris.experimental.ugrid.mesh import (
+            Mesh,
+            MeshEdgeCoords,
+            MeshFaceCoords,
+            MeshNodeCoords,
+        )
+
         # Exclude any mesh coords, which are bundled in with the aux-coords.
-        aux_coords_no_mesh = [
+        coords_to_add = [
             coord for coord in cube.aux_coords if not hasattr(coord, "mesh")
         ]
+
+        # Include any relevant mesh location coordinates.
+        mesh: Mesh = getattr(cube, "mesh")
+        mesh_location: str = getattr(cube, "location")
+        if mesh and mesh_location:
+            location_coords: MeshNodeCoords | MeshEdgeCoords | MeshFaceCoords = getattr(
+                mesh, f"{mesh_location}_coords"
+            )
+            coords_to_add.extend(list(location_coords))
+
         return self._add_inner_related_vars(
             cube,
             cf_var_cube,
             dimension_names,
-            aux_coords_no_mesh,
+            coords_to_add,
         )
 
     def _add_cell_measures(self, cube, cf_var_cube, dimension_names):
--- a/lib/iris/tests/unit/fileformats/netcdf/saver/test_Saver__ugrid.py
+++ b/lib/iris/tests/unit/fileformats/netcdf/saver/test_Saver__ugrid.py
@@ -403,6 +403,10 @@ class TestSaveUgrid__cube(tests.IrisTest):
             all(vars[co][_VAR_DIMS] == [face_dim] for co in face_coords)
         )
 
+        # The face coordinates should be referenced by the data variable.
+        for coord in face_coords:
+            self.assertIn(coord, a_props["coordinates"])
+
         # The dims of the datavar also == [<faces-dim>]
         self.assertEqual(a_props[_VAR_DIMS], [face_dim])
 
@@ -460,8 +464,10 @@ class TestSaveUgrid__cube(tests.IrisTest):
         v_a, v_b = vars["a"], vars["b"]
         self.assertEqual(v_a["mesh"], mesh_name)
         self.assertEqual(v_a["location"], "face")
+        self.assertEqual(v_a["coordinates"], "face_x face_y")
         self.assertEqual(v_b["mesh"], mesh_name)
         self.assertEqual(v_b["location"], "face")
+        self.assertEqual(v_b["coordinates"], "face_x face_y")
 
     def test_multi_cubes_different_locations(self):
         cube1 = make_cube(var_name="a", location="face")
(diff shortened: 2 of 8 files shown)


lib/iris/fileformats/netcdf/saver.py
```diff
@@ -1024,15 +1027,32 @@ def _add_aux_coords(self, cube, cf_var_cube, dimension_names):
...
+        )
+
         # Exclude any mesh coords, which are bundled in with the aux-coords.
-        aux_coords_no_mesh = [
+        coords_to_add = [
             coord for coord in cube.aux_coords if not hasattr(coord, "mesh")
         ]
+
+        # Include any relevant mesh location coordinates.
+        mesh: Mesh = getattr(cube, "mesh")
```
pp-mo, line 1043: I'm not really clear why you are including type hints here.  Are you sure it is a good idea?
I'm a bit ambivalent -- not sure that it adds anything much in this case.

Actually I have generally only seen them used  around object properties and interfaces, not in general code.

Also BTW the `location_coords: MeshNodeCoords | MeshEdgeCoords | MeshFaceCoords = getattr( ...` really surprised me - I didn't even know you can use "|" like this, up to now I always used "typing.Union".

lib/iris/tests/unit/fileformats/netcdf/saver/test_Saver__ugrid.py
```diff
@@ -403,6 +403,10 @@ def test_basic_mesh(self):
             all(vars[co][_VAR_DIMS] == [face_dim] for co in face_coords)
         )
 
+        # The face coordinates should be referenced by the data variable.
```
pp-mo, line 406: All the changes here are good, but I have one nagging doubt..
I wonder if we should definitely check a case where the "coordinates" attribute contains a **_mixture_** of regular aux-coords and ones from the mesh, just to prove that they get combined.

There is actually a come-by-chance example of that in one of the modified CDLs, where in  "ugrid_ex4_3d_layered.cdl" we have `datavar:coordinates = "Mesh2_depth Mesh2_face_x Mesh2_face_y Mesh2_surface" ;`
But that seems a bit fortuitous.

Maybe you  could add an extra, specific  test for that here ?

pp-mo (changes requested):
A couple of very minor queries, otherwise all looking great !

lib/iris/fileformats/netcdf/saver.py
trexfeathers: I wondered if this might be controversial. I agree it offers minimal value to humans reading the code, but that wasn't my motivation:

I'm a big fan of IDE code completion to speed up development, versus manually looking up the properties of any given object. The results of `getattr()` do not get this code completion, because the IDE can't be certain what is being returned, so I'm in the habit of using the hints to access this feature.

I don't know if you consider this a valid reason. I also don't know what [Mypy](https://mypy.readthedocs.io/en/stable/index.html) thinks, as that's often the reasoning others use to decide when to include type hints.

lib/iris/tests/unit/fileformats/netcdf/saver/test_Saver__ugrid.py
trexfeathers: Yes I was glad to see that test case in there, and I should have realised that this was therefore crying out for a targetted test. Will add one

trexfeathers: dbb41d69f4a44fec000de19709975cf212fdc374

pp-mo:
@hdyson 👀 !!

Issue #5175, opened 2023-02-23
Allow extraction with bounded time-coordinates.


pp-mo:
A user enquiry revealed that at least some extractions from bounded time coordinates are "banned".

There is a specific test+fail in place for these cases.
**We think that this restriction can now be removed, as the various time objects all now provide reliable comparison operations.**  (see below)

This was revealed when trying to extract a timepoint matching one from another cube, like this ...
```
>>> import iris
>>> from iris.cube import Cube
>>> from iris.coords import DimCoord
>>> from cf_units import Unit
>>> 
>>> co_time = DimCoord(
...     [0., 1],
...     standard_name='time',
...     units=Unit('hours since 1970-01-01', calendar='standard')
... )
>>> 
>>> times_cube = Cube([0, 1])
>>> times_cube.add_dim_coord(co_time, (0,))
>>> print(times_cube)
unknown / (unknown)                 (time: 2)
    Dimension coordinates:
        time                             x
>>> 
>>> time_point = co_time.units.num2date(co_time.points[0])
>>> constraint = iris.Constraint(time=time_point)
>>> print(time_point)
1970-01-01 00:00:00
>>> 
>>> extract = times_cube.extract(constraint)
>>> print(extract)
unknown / (unknown)                 (scalar cube)
    Scalar coordinates:
        time                        1970-01-01 00:00:00
>>> 
>>> times_cube.coord('time').bounds = [[0., 1], [1, 2]]
>>> extract = times_cube.extract(constraint)
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File ".../iris/cube.py", line 2615, in extract
    return constraint.extract(self)
  File ".../iris/_constraints.py", line 165, in extract
    resultant_CIM = self._CIM_extract(cube)
  File ".../iris/_constraints.py", line 189, in _CIM_extract
    resultant_CIM = resultant_CIM & coord_constraint.extract(cube)
  File ".../iris/_constraints.py", line 323, in extract
    r = np.array([call_func(cell) for cell in coord.cells()])
  File ".../iris/_constraints.py", line 323, in <listcomp>
    r = np.array([call_func(cell) for cell in coord.cells()])
  File ".../iris/_constraints.py", line 306, in call_func
    return c == self._coord_thing
  File ".../iris/coords.py", line 1359, in __eq__
    return self.contains_point(other)
  File ".../iris/coords.py", line 1533, in contains_point
    raise TypeError(
[...]

PR #5396, opened 2023-07-27
removed redundant timeattr check and added tests to ensure datetime/c…

acchamber:
### Description

Solving #5175, this removes the restriction against comparing datetime objects, as the only ones Iris should deal with are datetime.datetime, cftime.datetime, and PartialDateTime. Adds various new tests for checking combinations of those act as they should.

### DIFF
--- a/lib/iris/coords.py
+++ b/lib/iris/coords.py
@@ -1510,14 +1510,6 @@ class Cell(namedtuple("Cell", ["point", "bound"])):
         """
         if self.bound is None:
             raise ValueError("Point cannot exist inside an unbounded cell.")
-        if hasattr(point, "timetuple") or np.any(
-            [hasattr(val, "timetuple") for val in self.bound]
-        ):
-            raise TypeError(
-                "Cannot determine whether a point lies within "
-                "a bounded region for datetime-like objects."
-            )
-
         return np.min(self.bound) <= point <= np.max(self.bound)
 
 
--- a/lib/iris/tests/unit/coords/test_Cell.py
+++ b/lib/iris/tests/unit/coords/test_Cell.py
@@ -10,7 +10,6 @@
 import iris.tests as tests  # isort:skip
 
 import datetime
-from unittest import mock
 
 import cftime
 import numpy as np
@@ -87,29 +86,31 @@ class Test___common_cmp__(tests.IrisTest):
 
 class Test___eq__(tests.IrisTest):
     def test_datetimelike(self):
-        # Check that cell equality works with objects with a "timetuple".
-        dt = mock.Mock(timetuple=mock.Mock())
-        cell = mock.MagicMock(
-            spec=Cell, point=datetime.datetime(2010, 3, 21), bound=None
-        )
-        _ = cell == dt
-        cell.__eq__.assert_called_once_with(dt)
-
-    def test_datetimelike_bounded_cell(self):
-        # Check that equality with a datetime-like bounded cell
-        # raises an error. This is not supported as it
-        # depends on the calendar which is not always known from
-        # the datetime-like bound objects.
-        other = mock.Mock(timetuple=mock.Mock())
+        # Check that cell equality works with different datetime objects
+        # using the same calendar
+        point = cftime.datetime(2010, 1, 1, calendar="gregorian")
         cell = Cell(
-            point=object(),
+            datetime.datetime(2010, 1, 1),
             bound=[
-                mock.Mock(timetuple=mock.Mock()),
-                mock.Mock(timetuple=mock.Mock()),
+                datetime.datetime(2010, 1, 1),
+                datetime.datetime(2011, 1, 1),
             ],
         )
-        with self.assertRaisesRegex(TypeError, "bounded region for datetime"):
-            cell == other
+        self.assertEqual(cell, point)
+
+    def test_datetimelike_bounded_cell(self):
+        # Check that equality with a cell with a different calendar
+        # raises an error. This is not supported
+        point = cftime.datetime(2010, 1, 1, calendar="360_day")
+        cell = Cell(
+            datetime.datetime(2010, 1, 1),
+            bound=[
+                datetime.datetime(2010, 1, 1),
+                datetime.datetime(2011, 1, 1),
+            ],
+        )
+        with self.assertRaisesRegex(TypeError, "different calendars"):
+            cell == point
 
     def test_PartialDateTime_other(self):
         cell = Cell(datetime.datetime(2010, 3, 2))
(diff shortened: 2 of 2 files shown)


lib/iris/coords.py
```diff
@@ -1510,14 +1510,6 @@ def contains_point(self, point):
...
         if self.bound is None:
             raise ValueError("Point cannot exist inside an unbounded cell.")
-        if hasattr(point, "timetuple") or np.any(
-            [hasattr(val, "timetuple") for val in self.bound]
-        ):
-            raise TypeError(
-                "Cannot determine whether a point lies within "
-                "a bounded region for datetime-like objects."
-            )
-
```
trexfeathers, line 1520: There is a remaining similar block, which I assume needs similar changes?

https://github.com/SciTools/iris/blob/c48829aab5978ef9c60259d69b1b95cddfed3c68/lib/iris/coords.py#L1459-L1472

lib/iris/tests/unit/coords/test_Cell.py
```diff
@@ -87,29 +86,31 @@ def test_len_1_numpy_array(self):
...
-            spec=Cell, point=datetime.datetime(2010, 3, 21), bound=None
+        # Check that cell equality works with different datetime objects
+        # using the same calendar
+        point = cftime.datetime(2010, 1, 1, calendar="gregorian")
+        cell = Cell(
+            datetime.datetime(2010, 1, 1),
+            bound=[
+                datetime.datetime(2010, 1, 1),
+                datetime.datetime(2011, 1, 1),
+            ],
```
trexfeathers, line 97:
```suggestion
            bound=None,
```

Otherwise this changes what is under test.

lib/iris/tests/unit/coords/test_Cell.py
```diff
@@ -87,29 +86,31 @@ def test_len_1_numpy_array(self):
...
-                mock.Mock(timetuple=mock.Mock()),
-                mock.Mock(timetuple=mock.Mock()),
+                datetime.datetime(2010, 1, 1),
+                datetime.datetime(2011, 1, 1),
             ],
         )
-        with self.assertRaisesRegex(TypeError, "bounded region for datetime"):
-            cell == other
+        with self.assertRaisesRegex(TypeError, "different calendars"):
+            cell == point
```
trexfeathers, line 113: This is testing for something new - it deserves to be in a new test method.

`test_datetimelike_bounded_cell()` should instead be [how you originally wrote `test_datetimelike()`](https://github.com/SciTools/iris/blob/c48829aab5978ef9c60259d69b1b95cddfed3c68/lib/iris/tests/unit/coords/test_Cell.py#L91C17-L99).

lib/iris/tests/unit/coords/test_Cell.py
```diff
@@ -120,22 +121,87 @@ def test_PartialDateTime_other(self):
 
 
 class Test_contains_point(tests.IrisTest):
```
trexfeathers, line 123: No-one has told you this yet, but we are gradually migrating over to PyTest-style tests. Iris already fully supports these styles of test, so this class is a great opportunity for using [`parametrize`](https://docs.pytest.org/en/stable/how-to/parametrize.html#pytest-mark-parametrize-parametrizing-test-functions).

lib/iris/tests/unit/coords/test_Cell.py
trexfeathers, line 1: 👍 I'm personally a fan of only using `Mock` when absolutely necessary, since it can make test writing quite awkward, so I'm all for this part of the change. The caveat is that we need to be aware these breaking something in `Cell` will cause these tests to fail, but that wouldn't mean we had actually broken the date-time comparison.

lib/iris/coords.py
trexfeathers: Seems this would need some analogous testing as well?

trexfeathers (changes requested):
Thanks for looking into this @acchamber.

Some comments below, plus **this needs a What's New entry**.

lib/iris/coords.py
acchamber: added tests to _common_cmp_ to match this

lib/iris/tests/unit/coords/test_Cell.py
acchamber: Done

lib/iris/tests/unit/coords/test_Cell.py
acchamber: Done

lib/iris/tests/unit/coords/test_Cell.py
trexfeathers: Turns out sticking with `unittest`-style this was a conscious decision with @stephenworsley for keeping the file consistent, which I can agree with.

lib/iris/coords.py
```diff
@@ -1464,12 +1464,12 @@ def __common_cmp__(self, other, operator_method):
...
-                        "Cannot determine whether a point lies "
-                        "within a bounded region for "
-                        "datetime-like objects."
-                    )
+                # if hasattr(other, "timetuple"):
+                #     raise TypeError(
+                #         "Cannot determine whether a point lies "
+                #         "within a bounded region for "
+                #         "datetime-like objects."
+                #     )
```
trexfeathers, line 1472: Please could this be removed?

We always remove commented code, as it often gets forgotten. If we need to get something back we rely on Git.

lib/iris/tests/unit/coords/test_Cell.py
```diff
@@ -120,22 +143,87 @@ def test_PartialDateTime_other(self):
...
```
trexfeathers, line 145: OK in the agreed absence of parametrization, I have made my own matrix and I think a few test cases are missing?

| `Cell` ⬇ &nbsp;&nbsp;&nbsp;&nbsp; Point ➡  | `datetime` | `PartialDateTime` | `cftime.datetime` | `cftime.datetime(calendar="360_day")`  |
|---|---|---|---|---|
| `datetime` |   | ✔ | ✔ | ✔ |
| `cftime.datetime` | ✔ | ✔ |   |   |
| `cftime.datetime(calendar="360_day")` | ✔ | ✔ |   |   |

Is this accurate?

trexfeathers (changes requested):
Definitely getting there, thanks for actioning my last review. Three things left:

- https://github.com/SciTools/iris/pull/5396#discussion_r1279159868
- https://github.com/SciTools/iris/pull/5396#discussion_r1279161588
- What's New entry

lib/iris/tests/unit/coords/test_Cell.py
acchamber: They are, but I didn't write cases for matching cases (datetime/datetime, cftime/cftime, 360day/360 day) on purpose, because the intent of the test case is to check operations between different object classes, not within the same class (on the assumption that is tested by those libaries/classes themselves). Same as cftime.datetime/cftime.datetime(calendar="360_day"), assuming that within a object type calendar operations can be assumed and we just wanted to check across object types. 

I've put in a test for the 360day/360day case, because while it should be fine it is a check that iris handles non-standard calendars if they do match, which is probably worthwhile to include.

lib/iris/coords.py
acchamber: Sorry, left that in from testing, you're entirely right.

trexfeathers (approved):

PR #5402, opened 2023-07-30
Make roundtrip checking more precise + improve some tests accordingly.

pp-mo:
This provides an important tweak to the split-attrs feature branch, which really needs to be shown working with the main-branch code **_prior_** to the adoption of split attributes in cube content and loading (as already merged)

### DIFF
--- a/lib/iris/tests/integration/test_netcdf__loadsaveattrs.py
+++ b/lib/iris/tests/integration/test_netcdf__loadsaveattrs.py
@@ -39,6 +39,7 @@ import iris.fileformats.netcdf
 
 # A list of "global-style" attribute names : those which should be global attributes by
 # default (i.e. file- or group-level, *not* attached to a variable).
+
 _GLOBAL_TEST_ATTRS = set(iris.fileformats.netcdf.saver._CF_GLOBAL_ATTRS)
 # Remove this one, which has peculiar behaviour + is tested separately
 # N.B. this is not the same as 'Conventions', but is caught in the crossfire when that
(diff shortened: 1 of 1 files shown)


pp-mo:
Closing pending a re-submission.

I'm not happy with the apparently spurious commit from pre-commit-ci aiming to fix what is already fixed elsewhere (i.e. "later") on the feature-branch.  
And similarly the test failures from `lib/iris/tests/test_coding_standards.py::test_netcdf4_import` due to direct use of the netCDF4 module -- also a problem fixed "later" in the feature-branch.
So we really **don't** want to fix those problems here since it will probably create conflicts.

Following discussion with @trexfeathers , I think we can simply logically accept that, from results in #5403, these proposed changes to the "status-quo" tests are valid.  
So I will rebase this change onto the current head of the target branch, and **include it in a forthcoming "netcdf split attributes saving PR"** , as-per #4986

pp-mo:
Closing pending a re-submission.

I'm not happy with the apparently spurious commit from pre-commit-ci aiming to fix what is already fixed elsewhere (i.e. "later") on the feature-branch.  
And similarly the test failures from `lib/iris/tests/test_coding_standards.py::test_netcdf4_import` due to direct use of the netCDF4 module -- also a problem fixed "later" in the feature-branch.
So we really **don't** want to fix those problems here since it will probably create conflicts.

Following discussion with @trexfeathers , I think we can simply logically accept that, from results in #5403, these proposed changes to the "status-quo" tests are valid.  
So I will rebase this change onto the current head of the target branch, and **include it in a forthcoming "netcdf split attributes saving PR"** , as-per #4986

PR #5403, opened 2023-07-30
Splitattrs basisfix

pp-mo:
POC for changes "prior" to existing work on the FEATURE_split_attrs branch.
This is just to demonstrate that those changes capture the correct results for "old behaviour" (as still on main).
See #5402

**OUTPUTS**
Only the "ci-tests / tests" runs are relevant.
Also, it is appropriate to ignore the error `FAILED lib/iris/tests/test_coding_standards.py::test_netcdf4_import`,
as this was fixed subsequently in the feature-branch, and fixing it here may cause awkward conficts when merging this in.


pp-mo:
Has fulfilled its function !

Issue #5372, opened 2023-07-07
`unify_time_units()` prevents concatenation due to inconsistent `dtype`


trexfeathers:
## 🐛 Bug Report

Thanks to @HelenRJohnson for spotting this.

Using `iris.util.unify_time_units()` on a `CubeList` where the time units have an integer `dtype` prevents the `CubeList` from being concatenated:

```
iris.exceptions.ConcatenateError: failed to concatenate into a single cube.
  Dimension coordinates metadata differ: time != time
```

## Explanation

This is because `unify_time_units()` converts all time units to match the first one. cf-units will not convert something if it is already equal, so the first time unit does not undergo conversion. Meanwhile all the other time units are coerced to `dtype="float64"`, to avoid problems with partial days etc. So the end result is that the first time unit remains an integer `dtype` while all the others are floats, and this is something that is checked when preparing for concatenation.

## Proposed solution

Force all time coordinates to be `dtype=float64` at the start of the `unify_time_units()` process, and advertise this behaviour in the docstring. We should include tests for the new lines, too. I tested this with the below and it fixed my example.

```diff
    for cube in cubes:
        for time_coord in cube.coords():
            if time_coord.units.is_time_reference():
+               time_coord.points = time_coord.core_points().astype("float64")
+               time_coord.bounds = time_coord.core_bounds().astype("float64")
                epoch = epochs.setdefault(
                    time_coord.units.calendar, time_coord.units.origin
                )
```

## How To Reproduce
Steps to reproduce the behaviour:

1. Define a `CubeList` with multiple `Cube`s, which can be concatenated together.
   Each `Cube` should include a time coordinate, using `dtype=int` for the points.
   Make sure the final `Cube`'s time coordinate has a different epoch to the others (prevents successful concatenation).
2. Apply `iris.util.unify_time_units()` to the `CubeList`.
3. Attempt `CubeList.concatenate_cube()` on this `CubeList`.

## Expected behaviour

Should produce a single concatenated `Cube`.

## Environment 
 - OS & Version: Red Hat 7.9
 - Iris Version: `v3.6.1`

rcomer:
This seems familiar… #3213

trexfeathers:
Closed by #5412

am-thyst:
Thanks for the fix. It's unfortunate that this issue has been reported for over 10 years.

trexfeathers:
> Thanks for the fix. It's unfortunate that this issue has been reported for over 10 years.

[...]

PR #5412, opened 2023-08-03
Fix for #5372

acchamber:
### Description

Bug fix for issue #5372  

Solution is pretty much the same as Martin's proposed solution, with the exception of adding a check a time coord has bounds before casting them to a new type. 

Tests have also been added, checking the cases of different dtypes of times and of times having bounds.

---

### DIFF
--- a/lib/iris/util.py
+++ b/lib/iris/util.py
@@ -1492,6 +1492,11 @@ def unify_time_units(cubes):
     for cube in cubes:
         for time_coord in cube.coords():
             if time_coord.units.is_time_reference():
+                time_coord.points = time_coord.core_points().astype("float64")
+                if time_coord.bounds is not None:
+                    time_coord.bounds = time_coord.core_bounds().astype(
+                        "float64"
+                    )
                 epoch = epochs.setdefault(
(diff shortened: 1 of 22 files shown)


lib/iris/util.py
```diff
@@ -1471,7 +1471,8 @@ def unify_time_units(cubes):
     Performs an in-place conversion of the time units of all time coords in the
     cubes in a given iterable. One common epoch is defined for each calendar
     found in the cubes to prevent units being defined with inconsistencies
-    between epoch and calendar.
+    between epoch and calendar. During this process, all time coordinates have
+    their data type converted to 64-bit floats.
```
ESadek-MO, line 1475: Very much preference, might be worth explaining why they need to be the same type?

ESadek-MO (approved):
Looks good, thanks @acchamber! I've got one light suggestion, but I'd be fine if you disagreed and wanted it merged as is.

lib/iris/util.py
acchamber: No harm in it - most users will be using this function to prepare their cubes for concatenation anyways. pointing out thats the reason doesnt hurt.

PR #5416, opened 2023-08-05
Update Installation Guide

rcomer:
### Description

The installation guide is now out of date in a couple of places:

* Since v3.1 we do not provide platform-specific Conda downloads, so I do not think it is necessary to still list those platforms in the guide.
* Thanks to some outstanding work by @greglucas, Cartopy now has wheels :partying_face:, so users do not need to install any dependencies separately.  I tested the `pip` command in a new venv on my Ubuntu laptop and it Just Worked™.

I've also changed the language we use to distinguish things.

* "Users" vs "Developers" - I think we discussed previously somewhere that this distinction is a little ambiguous because users of Iris are going to be developing _something_.  In any case I might have reason to install a development version to do science and I might have reason to install a released version while contributing to Iris, so I prefer to distinguish by the Iris version type.
* I'm also not sure that "from source" is a helpful distinction for a pure python project.  So I went for "from a git checkout" instead.

I took the liberty of adding this to the v3.7.0 project as I think it would be nice to have this in the upcoming release's docs.  Obviously boot it out again if you disagree.

---

### DIFF
--- a/docs/src/installing.rst
+++ b/docs/src/installing.rst
@@ -3,30 +3,26 @@
 Installing
 ==========
 
-Iris is available using conda for the following platforms:
-
-* Linux 64-bit,
-* Mac OSX 64-bit, and
-* Windows 64-bit.
-
-Windows 10 now has support for Linux distributions via WSL_ (Windows
-Subsystem for Linux).  This is a great option to get started with Iris
-for users and developers.  Be aware that we do not currently test against
-any WSL_ distributions.
-
-.. _WSL: https://learn.microsoft.com/en-us/windows/wsl/install
+Iris can be installed using conda or pip.
 
 .. note:: Iris is currently supported and tested against |python_support|
           running on Linux.  We do not currently actively test on other
           platforms such as Windows or macOS.
+
+          Windows 10 now has support for Linux distributions via WSL_ (Windows
+          Subsystem for Linux).  This is a great option to get started with
+          Iris for users and contributors.  Be aware that we do not currently
+          test against any WSL_ distributions.
+
+.. _WSL: https://learn.microsoft.com/en-us/windows/wsl/install
 
 .. note:: This documentation was built using Python |python_version|.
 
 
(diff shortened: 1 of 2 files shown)


docs/src/installing.rst
```diff
@@ -3,30 +3,26 @@
...
+Iris can be installed using conda or pip.
 
 .. note:: Iris is currently supported and tested against |python_support|
           running on Linux.  We do not currently actively test on other
           platforms such as Windows or macOS.
 
+          Windows 10 now has support for Linux distributions via WSL_ (Windows
+          Subsystem for Linux).  This is a great option to get started with
+          Iris for users and contributors.  Be aware that we do not currently
+          test against any WSL_ distributions.
```
rcomer, line 15: Moved this down here as I think the fact that Iris is only tested on Linux motivates use of WSL, but I may have got the wrong end of the stick.  I've never attempted to do anything with python on Windows...

rcomer:
[Renders like this](https://scitools-iris--5416.org.readthedocs.build/en/5416/installing.html)

pp-mo (approved):
Yes absolutely, I think this is a worthy improvement !!
Thanks @rcomer

pp-mo:
I'm eager to fix this, but it needs a rebase/merge to resolve conflicts.
Could we possibly get that done today @rcomer ???
I'm hoping to cut an `iris3.7.0rc0` shortly

rcomer:
OK I rebased, but `git` didn't tell me I had any conflicts 😕

pp-mo:
> OK I rebased, but `git` didn't tell me I had any conflicts 😕

OK I might have misunderstood.  Can't remember what it looked like, now !

It does appear that you need to resolve the conversation you opened - I can't do it.
(To be clear, I agree -- proposal is fine)

pp-mo:
Thanks @rcomer, it's in !
Iris 3.7 coming soon, I can feel it ..

PR #5418, opened 2023-08-08
Split attrs docs

### LINKED ISSUE #4989: split-attrs: UserGuide
Needs to be explained "somewhere".

Possibly, we are _already_ overdue needing a specific section on netcdf save+load,
analagous to https://scitools-iris.readthedocs.io/en/latest/techpapers/um_files_loading.html

NB this info is currently hidden away in the "developer docs", which was not the original intention and might need changing : see #4990

List of documentation areas that may need addressing
  - [ ] "extra" specific section, _as noted above_,  on attribute handling in netcdf save/load
  - [ ] [Further topics - Metadata](https://scitools-iris.readthedocs.io/en/latest/further_topics/metadata.html)
        details exact metadata members - so 'cube' details need fixing
        It looks so far that nothing else needs enhancing, so long as it's obvious that local+global attrs will now operate independently.

trexfeathers:
## 🚀 Pull Request

Resolves #4989 

### Description

This is very bare, but as @pp-mo has pointed out there just isn't currently an appropriate place to discuss NetCDF I/O. Raising the profile any further would disproportionately focus on attribute handling and read very strangely. I've made a `# TODO:` as a placeholder and should probably create an issue tomorrow.

---

### DIFF
--- a/docs/src/further_topics/metadata.rst
+++ b/docs/src/further_topics/metadata.rst
@@ -91,6 +91,16 @@ actual `data attribute`_ names of the metadata members on the Iris class.
    metadata members are Iris specific terms, rather than recognised `CF Conventions`_
    terms.
 
+.. note::
+
+    :class:`~iris.cube.Cube` :attr:`~iris.cube.Cube.attributes` implement the
+    concept of dataset-level and variable-level attributes, to enable correct
+    NetCDF loading and saving (see :class:`~iris.cube.CubeAttrsDict` and NetCDF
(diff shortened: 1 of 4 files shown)


pp-mo:
Closes #4989

pp-mo (approved):
I agree that this is sufficient, and that we lack a suitable context to give a deeper discussion, **_and_** that this is not the time to write one !

I'm happy to merge this as-is now, but we should  do a final review of how it all looks before merging back to main.

Issue #82, opened 2012-09-11
setup.py fails on Windows due to incorrect handling of sub-directories for packages


munslowa:
Install on Windows fails with package directory 'lib\iris_pyke_rules' does not exist. This is due to find_package_tree not handling the fact that \lib\iris\fileformats directory contains a sub-directory. 

Tests carried out on Windows 7.

PR #5423, opened 2023-08-15
Update standard-names to #82 + add table version in std_names.py

pp-mo:
Updating the standard-names table to latest version 82

Also, since it was one of the requests in https://github.com/SciTools/iris/issues/5255, I have added a little piece of code to include the table version in the file header + provide it as a public global variable.
(and a silly little test for that)


pp-mo:
Ping @HGWright can you take this on ??
This is a standard thing in progress toward a new release (3.7)

HGWright (approved):
We believe the RTD failiure is a one off and we can't run again off of this PR. We are going to merge and make sure this it works on the next PR before the 3.7 release.

HGWright (approved):

PR #5427, opened 2023-08-17
Tweak indents to fix list numbering.

pp-mo:
This should fix a problem with the 3.7.0rc0 docs that was pointed out : the numbering of the lists of items in the whatsnew section was messed up by the insertion of anchor points for the 'feature items' para.

I think this fixes that.
We will include it in the main release, and in any subsequent bugfixes, but we won't make a bugfix release for this alone.
So for now, I'll include it here and, when this is agreed, it will also go into the v3.7.x mergeback to main.


pp-mo:
Ping @HGWright see what you think of this.

HGWright (approved):
After checking out the RTD, the changes seem to have the desired effect so I'm happy to merge this.

PR #5433, opened 2023-08-18
Minor benchmark improvements

trexfeathers:
### Requires

- #5429
- #5430 
- #5431 
- #5432

Will need `git rebase upstream/FEATURE_benchmarks` once those are merged.

### Description

- e2347c702146988c153349cd1a956dc03867efbb : Use `shlex.split()` for readable command line entries, rather than my own home-brew version
- fc12dd3ffdd01b98f3e0bf5b43d1f9c643a5b749 : Some minor documentation clarifications

No What's New entry since this is so minor and internal.

---
[Consult Iris pull request check list]( https://scitools-iris.readthedocs.io/en/latest/developers_guide/contributing_pull_request_checklist.html)

PR #5432, opened 2023-08-18
Warn via issue if overnight benchmarks fail

trexfeathers:
### Requires

- #5429 
- #5430 
- #5431

Will need `git rebase upstream/FEATURE_benchmarks` once those are merged.

### Description

Raise an issue if the overnight benchmark workflow gets a non-zero exit code. Up to now we have not been notified of these problems.

### Demonstrations

I have temporarily created a different default branch on my fork: [`trexfeathers:demo_20230817_main`](https://github.com/trexfeathers/iris/tree/demo_20230817_main), which can simulate what would happen on `main`.

- Issue from failed workflow: https://github.com/trexfeathers/iris/issues/71
- [Past failures on `SciTools:main` that we were never notified of](https://github.com/SciTools/iris/actions/workflows/benchmark.yml?query=branch%3Amain+is%3Afailure).

---
[Consult Iris pull request check list]( https://scitools-iris.readthedocs.io/en/latest/developers_guide/contributing_pull_request_checklist.html)

PR #5431, opened 2023-08-18
Better comparison commits for PR benchmarking

trexfeathers:
### Requires

- #5429 
- #5430

### Description

The benchmark GHA no longer checks out the `HEAD` of the pull request branch. By using the default merge-commit, like all other CI, this will now show what the performance of the  merge target (e.g. `main`) would be IF the PR were merged, and compare that to the performance of the merge target without the PR.

---
[Consult Iris pull request check list]( https://scitools-iris.readthedocs.io/en/latest/developers_guide/contributing_pull_request_checklist.html)

PR #5430, opened 2023-08-18
Post on demand benchmark results as comment (big refactor)

trexfeathers:
### Requires

- #5429 

### Description

- Uses Python for the complexities of producing a benchmark report. Was previously using BASH.
- The report is posted by a **separate** GHA workflow. This is required because allowing GHA to comment on pull requests is only safe when running on a 'safe' branch - i.e. `main` instead of the author's PR branch.

### Demonstrations

I have temporarily created a different default branch on my fork: [`trexfeathers:demo_20230817_main`](https://github.com/trexfeathers/iris/tree/demo_20230817_main), which contains `benchmark_reports.yml` and is therefore able to listen for performance reports and post them appropriately.

- Comment on a pull request: https://github.com/trexfeathers/iris/pull/66#issuecomment-1682548312
- Overnight issue: https://github.com/trexfeathers/iris/issues/76
- Multiple issues from same overnight run (1 per commit):
  - https://github.com/trexfeathers/iris/issues/76
  - https://github.com/trexfeathers/iris/issues/77
- Overnight run with no regressions to report: [`5893483565`](https://github.com/trexfeathers/iris/actions/runs/5893483565)

---
[Consult Iris pull request check list]( https://scitools-iris.readthedocs.io/en/latest/developers_guide/contributing_pull_request_checklist.html)

PR #5434, opened 2023-08-18
Set benchmark runs to error if the subprocess errors

trexfeathers:
### Requires

- #5429 
- #5430 
- #5431 
- #5432 
- #5433 

Will need `git rebase upstream/FEATURE_benchmarks` once those are merged.

### Description

Fixes an oversight introduced when we switched away from Nox (#5215). Called process errors in `bm_runner.py` should cause an error upstream. Without this: we have had some benchmarks erroring without us knowing about it.

Also included handling to allow the overnight run to report whatever results it has, before erroring.

### Demonstrations

I have temporarily created a different default branch on my fork: [`trexfeathers:demo_20230817_main`](https://github.com/trexfeathers/iris/tree/demo_20230817_main), which can simulate what would happen on `main`.

- [Previously missed benchmark errors](https://github.com/SciTools/iris/actions/runs/5884680662/job/15959841341#step:9:116) on `SciTools:main`
- Errors from benchmarks now reported: https://github.com/trexfeathers/iris/issues/80
- Benchmark results still reported even from a failed run: https://github.com/trexfeathers/iris/issues/81

---

### DIFF
--- a/.github/workflows/benchmarks_run.yml
+++ b/.github/workflows/benchmarks_run.yml
@@ -112,6 +112,7 @@ jobs:
           gh issue create --title "$title" --body "$body" --label "Bot" --label "Type: Performance" --repo $GITHUB_REPOSITORY
 
       - name: Upload any benchmark reports
+        if: success() || steps.overnight.outcome == 'failure'
         uses: actions/upload-artifact@v3
         with:
           name: benchmark_reports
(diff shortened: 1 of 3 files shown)


ESadek-MO (approved):
Happy with this once rebased and tests run, thanks @trexfeathers

Issue #5461, opened 2023-09-05
More tolerant AreaWeighted regridding behaviour for target cells stradling the source extents


stephenworsley:
## ✨ Feature Request

Current behaviour for `AreaWeighted` is that a target cell which straddles the extents of the source grid will always be masked _even if the `mdtol` value would otherwise be large enough_. The alternate approach would be to treat out of bounds areas the same way as masked areas, i.e. as _missing data_ for which the tolerance  is controlled by the `mdtol` keyword. If we want to keep behaviour consistent, this could be controlled by another keyword in `AreaWeighted()`, e.g. `out_of_bounds_tolerance`.

PR #5462, opened 2023-09-05
Increase AreaWeighted extent tolerance

stephenworsley:
Addresses #5461


stephenworsley:
It looks like the `mdtol` behaviour relies on the missing data being explicitly from a masked source. While I believe the new behaviour in `_within_bounds` would adress #5461, it remains to change `mdtol` behaviour so that out of bounds and masked areas are equivalent. There should also be additional tests that this equivalence holds.

PR #5482, opened 2023-09-07
idiff acceptance fix

ESadek-MO:
## 🚀 Pull Request

Closes #5098.

### Description
idiff was failing when "accept" was chosen (often with the user unaware), as there was a conditional that checked for file name duplication. Since the idea is to replace the new file, I've removed this conditional entirely. 

Draft because I have yet to test this works/doesn't break anything, and needs a whats new.

Issue #5426, opened 2023-08-17
Quickplot Automatic Time Axis Label Completely Incorrect


rebeccaherman1:
## 🐛 Bug Report
The automatic quickplot time label is unrelated to the printed time ticks and even to native cube units

## Minimum Reproducible Code
Any spatio-temporal cube will do; I've saved an example of Pacific SST from a pre-Industrial control run using years 1191-1200, attached it below (be sure to remove `.txt` from the name):
[sst_example.nc.txt](https://github.com/SciTools/iris/files/12369965/sst_example.nc.txt)

```python
import iris
import iris.quickplot as qplt
import esmvalcore.preprocessor as pp
ex = iris.load_cube('sst_example.nc')  
m_ex = pp.meridional_statistics(ex, operator='mean')
qplt.pcolormesh(m_ex)
#prints native units of cube
Time = m_ex.coord('time')
time_meta_data = Time.metadata.units
Calendar = time_meta_data.calendar
U = time_meta_data.cftime_unit
print(U)
print(max(Time.points))
```
## Expected behaviour
The native units of the cube `m_ex` are `days since 1850-1-1 00:00:00` (see printed output in the screenshot below), and the values are incredibly large negative numbers (the largest time value is `-237266.5`). However, I do not expect to see tick labels in days relative to some offset, but in absolute time presented in years, ranging from 1191 to 1200. I thus expect the time axis to be labelled `Year`, or `Time / Year`

## Observed behavior
The time ticks appear as expected, as all are between 1191 and 1200. However, the automatic label says `Time / days since 1970-01-01`. This is clearly not consistent with the time ticks because the presented numbers of in units of years with no time offset. Furthermore, t's completely unclear where the offset in the label came from, since the native offset of the cube being plotted (`m_ex`) is `1850`. I wonder if these units have been hard-coded somewhere?

![Screenshot from 2023-08-17 15-44-11](https://github.com/SciTools/iris/assets/12139022/17c08733-2e3f-4502-a764-835463a55373)

rcomer:
You are right.  It is hard-coded here:

https://github.com/SciTools/iris/blob/8361532d141a16ec44ecf65e85162c583c7c6d06/lib/iris/quickplot.py#L50-L56

[...]

PR #5484, opened 2023-09-07
removed units for time coords due to not matching autoticks

acchamber:
### Description
Draft pull request fixing #5426 
Implemented Ruth's suggestion and removed automatic units in axis titles for Time reference units entirely - can't ensure they line up with the autoticks. 

Draft because waiting on #5482 to change the image tests to match before pushing and asking for review

---


rcomer:
Does this make the change from #616 redundant?  If so it could be reverted.

acchamber:
I think it does - tests should still pass. Once #5482 fixes the idiff tests I'll check if it can be removed safely.

trexfeathers:
@acchamber now you have less time for SciTools things, what do you think about the status of this PR?

acchamber:
Trying to rebuild my environment has resulted in environment issues when trying to run idiff.py to update the test results - it can't find iris on import, and I'm not sure why. The code should be fine, but until the test results can be updated it's not ready.

rcomer:
> I will not be doing any further work on this PR

I have pulled this change into #5561

PR #5496, opened 2023-09-13
No longer use ASV ``--strict``

### LINKED ISSUE #5421: Stop using `--strict` with ASV benchmarks
The `--strict` parameter is being removed from ASV in the imminent `v0.6` release (airspeed-velocity/asv#1205). We use this as parameter in our benchmark runs:

https://github.com/SciTools/iris/blob/3531f7b0a27d25d50fe20f17c53e06e5ebfc2dc0/benchmarks/bm_runner.py#L29-L32

The good news: the parameter is being removed because it is now the default behaviour

The bad news: we need to stop making calls to strict, and also work out how to modify the times when we _don't_ call strict:

https://github.com/SciTools/iris/blob/3531f7b0a27d25d50fe20f17c53e06e5ebfc2dc0/benchmarks/bm_runner.py#L324-L326

trexfeathers:
### Description

Closes #5421 

---

### DIFF
--- a/benchmarks/bm_runner.py
+++ b/benchmarks/bm_runner.py
@@ -32,8 +32,7 @@ GH_REPORT_DIR = ROOT_DIR.joinpath(".github", "workflows", "benchmark_reports")
 
 # Common ASV arguments for all run_types except `custom`.
 ASV_HARNESS = (
-    "run {posargs} --attribute rounds=4 --interleave-rounds --strict "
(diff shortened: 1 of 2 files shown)


ESadek-MO (approved):
This all looks good to me! Happy to merge this in.

trexfeathers:
Thanks for the emergency review @ESadek-MO

Issue #5500, opened 2023-09-15
Sample mesh plotting code has broken


pp-mo:
## How To Reproduce
Steps to reproduce the behaviour:

1. copy  the sample code from the UserGuide examples in ["further_topics/ugrid/operations#load"](https://scitools-iris.readthedocs.io/en/v3.7.0/further_topics/ugrid/operations.html#load) and ["further_topics/ugrid/operations#plotting"](https://scitools-iris.readthedocs.io/en/v3.7.0/further_topics/ugrid/operations.html#plotting)
2. insert an unstructured netCDF/UGRID test file, such as ["iris-sample-data:mesh_C4_synthetic_float.nc"](/https://github.com/SciTools/iris-sample-data/blob/v2.4.0/iris_sample_data/sample_data/mesh_C4_synthetic_float.nc)
3. run the code to produce a plot

## Expected behaviour
Code should be self-contained, and it should run

## Problems
  * Primarily, [the `from geovista.common import to_xyz` line](https://github.com/SciTools/iris/blob/v3.7.0/docs/src/further_topics/ugrid/operations.rst?plain=1#L487) is now failing, since this geovista routine now got renamed to "geovista.common.to_cartesian"
  * Secondarily, the resulting plot is now near-empty because the coarse data is obscured by the base_layer (e.g. a radius change is needed to avoid one obscuring the other).  It's not clear whether this changed, or maybe did not occur with the data used in the original (we don't know what that was)

There are obviously also some related strategic problems which this highlights:
  * the example is not run, so no way to spot when it goes stale (and in this case, now hits an exception)
  * the example is not self-contained, so you can't simply "run it" anyway -- *and neither can a user*

trexfeathers:
I think #5740 took care of this. @pp-mo please re-open if you disagree.

PR #5501, opened 2023-09-15
Self-contained runnable sample code for basic mesh plotting.

pp-mo:
Attempts to at least partially address #5500 

The new sample code I believe works, and produces the provided plot.

The issue of how we stop this untested code going stale is still unresolved.


bjlittle:
@pp-mo Adding `import geovista.theme` gives you better default colours IMHO.

Although, if you do this here, you'd probably want to be consistent and also do this throughout all the `geovista` examples

docs/src/further_topics/ugrid/operations.rst
```diff
@@ -502,50 +518,29 @@ earlier:
...
-            Mesh coordinates:
-                latitude                        x          -
-                longitude                       x          -
-            Attributes:
-                Conventions                 'CF-1.7'
-
-        # Convert our mesh+data to a PolyData object.
-        # Just plotting a single height level.
-        >>> face_polydata = cube_faces_to_polydata(face_cube[:, 0])
+        ...
```
bjlittle, line 521: @pp-mo You can now simplify the call to `Transform.from_unstructured` as follows:

```python
mesh = Transform.from_unstructured(
    lons.points,
    lats.points,
    connectivity=indices,
    data=cube.data,
    name=f"{cube.name()} / {cube.units}",
)
```

There is now no need to provide the `start_index` as `geovista` will detect this, and it's better to name the `indices` with the `connectivity` kwarg for clarity.

docs/src/further_topics/ugrid/operations.rst
```diff
@@ -502,50 +518,29 @@ earlier:
...
-        # Centre the camera on the data.
-        >>> camera_region = to_xyz(
-        ...     face_cube.coord("longitude").points,
-        ...     face_cube.coord("latitude").points,
-        ...     radius=3,
-        ... )
-        >>> camera_pos = camera_region.mean(axis=0)
-        >>> my_plotter.camera.position = camera_pos
-
+        >>> my_plotter.background_color = "#502020ff"
```
bjlittle, line 534: If you use `import geovista.theme` there's no need to set the `background_color`

docs/src/further_topics/ugrid/operations.rst
```diff
@@ -502,50 +518,29 @@ earlier:
...
+        >>> _ = my_plotter.add_coastlines(color="white", radius=1.01)
+        >>> _ = my_plotter.add_mesh(face_polydata, radius=1.0)
```
bjlittle, line 536: There's no real need here to set the `radius`, it's not adding anything of great value to "basic" plotting... it's just a distraction.

If you're concern is that the `mesh` is so low resolution that it's markedly different from the spherical coastlines, then only specify the `radius` for the coastlines. Setting the `radius=1.0` for the `face_polydata` is a nop.

bjlittle (changes requested):
@pp-mo I've raised some questions for you to consider

ESadek-MO:
From @SciTools/peloton: This was closed by mistake, we think this is still worth doing!

pp-mo:
I revived the branch, which I deleted by mistake !

pp-mo:
Well, as @trexfeathers pointed out, we do now have [this mini geovista example](https://scitools-iris.readthedocs.io/en/v3.14.0/further_topics/ugrid/operations.html#plotting) in Iris docs.

It's not in the gallery, but maybe it doesn't need to be?
I guess people with this type of data will know that it is geovista they need.

**If that's enough, we can close this?**

trexfeathers:
> Well, as @trexfeathers pointed out, we do now have [this mini geovista example](https://scitools-iris.readthedocs.io/en/v3.14.0/further_topics/ugrid/operations.html#plotting) in Iris docs.
> 
> It's not in the gallery, but maybe it doesn't need to be? I guess people with this type of data will know that it is geovista they need.
> 
> **If that's enough, we can close this?**

I would support a full-on gallery entry, but the existing documentation means I'd be comfortable leaving this too, if we'd rather focus on other priorities.

ESadek-MO:
From @SciTools/peloton 

This has alreay been completed; we need an issue for copying this example over to the gallery!

Issue #5499, opened 2023-09-15
Experiment with community workarounds to duplicated warnings


trexfeathers:

stephenworsley:
A solution is proposed by #5506, this implements one of two potential replacements for the `warnings.warn` function. In particular, this mimics the behaviour you would get by calling `warnings.simplefilter("default")` (see https://docs.python.org/3/library/warnings.html#the-warnings-filter). Similarly, one could modify this function slightly to mimic the behaviour of `warnings.simplefilter("once")`. This would make the behaviour of each of these warnings more consistent within a context where they behaved as expected, however it comes at the cost of reduced flexibility when you want to apply a different context like `warnings.simplefilter("always")`. The table below descibes how each proposed function would behave in each context. Note, "sometimes" here means that the warnings do not behave as expected when this bus is present https://github.com/python/cpython/issues/73858.

|Method|Handles "once"|Handles "default"|Handles "always"
|:---|:---|:---|:---|
|warnings.warn|sometimes|sometimes|sometimes|
|Forced default|sometimes|always|never|
|Forced once|always|never|never|

---
An alternate approach is to rewrite the code in iris to avoid the bug described here https://github.com/python/cpython/issues/73858. In particular, this patch of code appears to be resposible for the bug appearing during loading: https://github.com/SciTools/iris/blob/b6e39d758fc17e167b9323b25484965453846dbf/lib/iris/fileformats/_nc_load_rules/helpers.py#L417-L431
It is not guaranteed that all duplications would be a result of code within iris, since some calls to other libraries like `import numpy` or `dask.array.map_blocks` also introduce this bug. It may therefore be worthwhile taking both approaches so that:
- warnings within iris are not duplicated due to code outsied of iris.
- warnings outside of iris are not duplicated due to code within iris.

Finally, it is worth noting that this may end up with us having code that needs to be reverted once this bug is fixed within python, so this ought to be noted by a TODO alongside the code.

stephenworsley:
See this gist for a more in depth demonstration of warning duplication https://gist.github.com/stephenworsley/7716fdba4cc634e73454462ad60b3f67

stephenworsley:
[...]

Issue #5464, opened 2023-09-05
Fix duplicated warnings


trexfeathers:
# Story

- As a
  Iris user
- I want
  To only receive each warning once - the Python default behaviour
- So that
  I do not need to develop specific strategies for working with Iris versus other Python libraries, and so my `stdout`/`stderr` is readable, allowing effective development
  

# Notes

Reports of some warnings being raised once per file. I suspect something related to parallelism - warnings are not thread safe by default, but can be made so (from reading Stack Overflow).

# From refinement discussion

- Initial work must be to diagnose when 'duplicates' are issued. Is it just once per file? Once per `Cube`? Is there some thread safety involved?
- Further steps will depend on the results of the diagnosis.
```[tasklist]
### Tasks
- [ ] https://github.com/SciTools/iris/issues/5463
- [ ] https://github.com/SciTools/iris/issues/5499
- [ ] https://github.com/SciTools/iris/issues/5536
```

Issue #5466, opened 2023-09-05
Improve utility of warnings in Iris tests


trexfeathers:
# Story

- As a
  Iris developer
- I want
  Warnings that are a problem for Iris to result in CI failure (e.g. a `DeprecationWarning`)
- So that
  I can address problems introduced by new commits **proactively** and **immediately** (rather than reactively when issues are discovered downstream)

# Notes

SciTools/iris#4384

Iris Gallery includes (or has included!) code to convert warnings to failures/errors.

# Ideas

- Suppress warnings we know are not a problem, so only those that matter are left. Then set a test run to exit non-`0` if any warnings are raised. Would likely need to address quite a few warnings before this could start working!
- A custom routine that causes a failure if it detects warnings of known categories (might only be `DeprecationWarning`?).

# From discussion

- Cost-benefit is difficult. Focussing on `DeprecationWarning` would be fairly easy, but is it offering much. Covering all warnings may need a lot of leg work to adapt the tests.
- Should start by seeing how many tests fail if warnings are 'promoted', and see how this informs further discussions.
```[tasklist]
### Tasks
- [ ] https://github.com/SciTools/iris/issues/5465
- [ ] https://github.com/SciTools/iris/issues/5491
- [ ] https://github.com/SciTools/iris/issues/5492
- [ ] https://github.com/SciTools/iris/pull/5493
- [ ] https://github.com/SciTools/iris/issues/5535
- [ ] https://github.com/SciTools/iris/issues/5534
- [ ] https://github.com/SciTools/iris/issues/5533
- [ ] https://github.com/SciTools/iris/issues/5530
- [ ] https://github.com/SciTools/iris/issues/5531
- [ ] https://github.com/SciTools/iris/issues/5532
- [ ] https://github.com/SciTools/iris/issues/5889
```

ESadek-MO:
Note: have converted write up from #5465 into issues and added as subtasks on this issue.

HGWright:
From @SciTools/peloton It would be a good idea to have the tests fail on DepricationWarnings, so that we are forced to address them. We have missed these in the past because the tests raise lots of warnings. https://docs.pytest.org/en/stable/reference/reference.html#confval-filterwarnings

PR #5506, opened 2023-09-20
Prevent warning repeats

stephenworsley:
## 🚀 Pull Request

Attempts to address #5464, the result of investigations done for #5499.

This approach attempts to align the behaviour of warnings more closely with what ought to be the _default_ behaviour of warnings. This currently comes at the expense of "hard coding" some of this behaviour to match what you would get when calling `warnings.simplefilter("default")`. When setting `warnings.simplefilter("always")` the expected behaviour would be for all warnings to be raised, however some will be caught, as per the behaviour of `warnings.simplefilter("default")`. Similarly, this does not go as far as to guarantee that `warnings.simplefilter("once")` will behave as expected, though it should not introduce any duplications that weren't already present.

The above problems could potentially be solved with some additional logic if I could figure out a way to determine which of these behaviours `warnings` expected at any given point in the code. Otherwise, it may be an acceptable payoff to hardcode this behaviour if it would reduce unwanted warnings in most cases.

### DIFF
--- a/lib/iris/exceptions.py
+++ b/lib/iris/exceptions.py
@@ -350,16 +350,16 @@ class IrisSaverFillValueWarning(IrisMaskValueMatchWarning, IrisSaveWarning):
 
 
 @lru_cache(None)
-def warn_once(msg, type, stacklevel, frame, **kwargs):
+def warn_once(msg, stacklevel, frame, **kwargs):
     """Raise a warning only if a similar one has not been raised before."""
-    warnings.warn(msg, type, stacklevel=stacklevel, **kwargs)
+    warnings.warn(msg, stacklevel=stacklevel, **kwargs)
 
 
-def warn_once_at_level(msg, type=None, stacklevel=0, **kwargs):
+def warn_once_at_level(msg, stacklevel=0, **kwargs):
     """
     Raise a warning only if a similar one hasn't been raised from the same line
     for a given stack level.
     """
     stacklevel += 1
     frame = traceback.format_stack()[-stacklevel]
-    warn_once(msg, type, stacklevel + 1, frame, **kwargs)
+    warn_once(msg, stacklevel + 1, frame, **kwargs)
--- a/lib/iris/exceptions.py
+++ b/lib/iris/exceptions.py
@@ -350,16 +350,16 @@ class IrisSaverFillValueWarning(IrisMaskValueMatchWarning, IrisSaveWarning):
 
 
 @lru_cache(None)
-def warn_once(msg, type, stacklevel, frame):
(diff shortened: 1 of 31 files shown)


trexfeathers:
## Before

image

## After

image

(forgive the difference in test numbers - different PR's)

trexfeathers:
Thanks @ESadek-MO and @stephenworsley for exploring alternative solutions (#5536). I'm still unhappy that this PR effectively disables some of the `action`s in the [Python warnings filter](https://docs.python.org/3/library/warnings.html#the-warnings-filter), but we've effectively been backed into that corner by forces outside our control.

Thankfully, we now have documentation about Iris warnings: [`filtering_warnings.rst`](https://github.com/SciTools/iris/blob/1d71bd4d0f445548235ecd3a1671982ba81302f7/docs/src/further_topics/filtering_warnings.rst). I would be worth:

1. Adding a heading/[admonitions](https://docutils.sourceforge.io/docs/ref/rst/directives.html#admonitions) to `filtering_warnings.rst` to clearly document Iris' deviation from standard Python behaviour.
2. Reviewing the existing text of `filtering_warnings.rst` to check it is not misleading in light of the changes in this PR.

trexfeathers (changes requested):
See my previous comment: https://github.com/SciTools/iris/pull/5506#issuecomment-1889016831

lib/iris/exceptions.py
```diff
@@ -180,3 +183,19 @@ class CannotAddError(ValueError):
...
+
+
+def warn_once_at_level(msg, type, stacklevel):
+    """
+    Raise a warning only if a similar one hasn't been raised from the same line
+    for a given stack level.
+    """
+    stacklevel += 1
+    frame = traceback.format_stack()[-stacklevel]
+    warn_once(msg, type, stacklevel + 1, frame)
```
bjlittle, line 201: @stephenworsley Would you consider namespacing this in `iris.warnings` instead?

Seems a tad odd for `iris.exceptions` to raise a warning ... initially I'd not expect that behaviour to live there.

trexfeathers: I'm not against this; just wanted to point out there's a lot of other stuff we should move as well, if we do this!

bjlittle: Seems like we should consider doing that then if we're in this space, otherwise it won't happen ... typically, that's the case.

trexfeathers: We did this! #5760

HGWright:
Going to close this, but investigate the warnings raised in the tests, and not rule out coming back to this. #5466

PR #5508, opened 2023-09-20
moved latest warning banner logic to conf.py

tkknight:
### Description
Moved the latest **warning** **banner** logic from a **template** into `conf.py`.  A recent feature added to the [pydata-sphinx-theme](https://pydata-sphinx-theme.readthedocs.io/en/latest/index.html) allows this, see https://pydata-sphinx-theme.readthedocs.io/en/latest/user_guide/announcements.html#provide-local-html-in-your-theme

The theme does have a version switcher banner (see https://pydata-sphinx-theme.readthedocs.io/en/latest/user_guide/version-dropdown.html) but this uses a **json** file to list all versions.  We automate this via Read The Docs so was not a suitable option, just using the **announcement** feature works.

### Current banner
image

### New banner (this PR)
image

The text is also shortened which reads better imho.  Note the "dirty" release version, this is due to the docs being built locally and not on Read the Docs.  This new banner is only visible once build on Read The Docs **AND** it is the latest docs, so this PR docs build should work but not show the banner.  This was tested by overriding some variables, see https://github.com/SciTools/iris/blob/main/docs/src/conf.py#L57

A whatsnew will be done in a separate PR.

---


bjlittle (approved):
@tkknight This is a lovely change. Great to see this pulled into the `conf.py`, where it's more visible.

Thanks 🍻 

Please add your whatsnew comment in a follow-on PR 👍  (as agreed)

PR #5515, opened 2023-09-25
Feedstock rc branch management in do-nothing script

trexfeathers:
### Description

---


tools/release_do_nothing.py
```diff
@@ -526,7 +526,33 @@ def update_conda_forge(
     _wait_for_done(message)
 
     if is_release_candidate:
-        upstream_branch = "release-candidate"
+        # We chose this odd handling of release candidate branches because
```
pp-mo, line 529: I'm probably missing something here, but it seems now to say this new part about "renaming the old branch" **_every_** time we create a release candidate.
Can it not be made to happen only when there _is_ a pre-existing one ?

pp-mo (changes requested):
Generally looks reasonable, no other problems I can seee

bjlittle (approved):
@trexfeathers LGTM 👍

tools/release_do_nothing.py
trexfeathers: I agree this needed more nuance, so I've had another go

cbee22fd09a15e7a6b11dbc3ea7f6769b6b98497

pp-mo (approved):

trexfeathers:
Many thanks @bjlittle and @pp-mo 🥂

PR #5537, opened 2023-10-09
Gallery: show colour bar stealing space from multiple axes

rcomer:
### Description

It was embarrassingly recent that I realised you can get Matplotlib to automatically place a colour bar across multiple axes.  Probably because I originally came to the Iris gallery to learn this sort of thing.  This PR modifies one example to show such placement rather than manually placing a colour bar axes.  Note that we also have the [GloSea example](https://scitools-iris.readthedocs.io/en/latest/generated/gallery/meteorology/plot_lagged_ensemble.html#sphx-glr-generated-gallery-meteorology-plot-lagged-ensemble-py) showing manual colour bar placement.

I have two versions of this in the two commits.  In the first commit, I created Cartopy GeoAxes up front so that I could pass the original array of axes to `colorbar`.  I the second commit, I reverted that and found a different way to get hold of the GeoAxes that Iris auto-creates.  I am in two minds which is the right version to keep:  I think the first version is Better Object Oriented Code, but the second version better illustrates Iris' capabilities (and gotchas).

---

### DIFF
--- a/.github/workflows/benchmarks_run.yml
+++ b/.github/workflows/benchmarks_run.yml
@@ -29,7 +29,7 @@ jobs:
     env:
       IRIS_TEST_DATA_LOC_PATH: benchmarks
       IRIS_TEST_DATA_PATH: benchmarks/iris-test-data
-      IRIS_TEST_DATA_VERSION: "2.19"
+      IRIS_TEST_DATA_VERSION: "2.21"
       # Lets us manually bump the cache to rebuild
       ENV_CACHE_BUILD: "0"
       TEST_DATA_CACHE_BUILD: "2"
--- a/docs/src/whatsnew/latest.rst
+++ b/docs/src/whatsnew/latest.rst
@@ -76,7 +76,7 @@ This document explains the changes made to Iris for this release
    Iris. See :ref:`filtering-warnings`. (:pull:`5509`)
 
 #. `@rcomer`_ updated the "Global average annual temperature maps" to show how
-   a colourbar may steal space from multiple axes. (:pull:`5093`)
+   a colourbar may steal space from multiple axes. (:pull:`5537`)
 
 
 💼 Internal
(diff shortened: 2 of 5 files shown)


rcomer:
Needs https://github.com/SciTools/iris-test-data/pull/86

trexfeathers:
> I was totally expecting a failure because the reference image needs updating 😕

https://github.com/SciTools/iris/issues/5539

trexfeathers (changes requested):
Looks good! I'm happy that this works, and it's definitely more elegant / more idiomatic Matplotlib. I prefer the second approach - the one currently in the PR - as it is more typical Iris use with less to explain.

Got one thing I'd be keen for you to change in the What's New.

docs/src/whatsnew/latest.rst
```diff
@@ -75,6 +75,9 @@ This document explains the changes made to Iris for this release
 #. `@trexfeathers`_ documented the intended use of warnings filtering with
    Iris. See :ref:`filtering-warnings`. (:pull:`5509`)
 
+#. `@rcomer`_ updated the "Global average annual temperature maps" to show how
+   a colourbar may steal space from multiple axes. (:pull:`5537`)
```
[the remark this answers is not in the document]
rcomer: Good call 👍

rcomer:
I think I also need to update the test data version?

trexfeathers:
> I think I also need to update the test data version?

Yes please, thanks for the reminder

.github/workflows/benchmarks_run.yml
```diff
@@ -29,7 +29,7 @@ jobs:
     env:
       IRIS_TEST_DATA_LOC_PATH: benchmarks
       IRIS_TEST_DATA_PATH: benchmarks/iris-test-data
-      IRIS_TEST_DATA_VERSION: "2.19"
+      IRIS_TEST_DATA_VERSION: "2.21"
```
rcomer, line 32: Not sure why it's 2 versions, but we're definitely now on 2.21
https://github.com/SciTools/iris-test-data/commit/a2e2bb7eacadf9ba77ca32e25a4a125f022b85ee

trexfeathers: Looks like `v2.20` was to help [ncdata](https://github.com/pp-mo/ncdata) - SciTools/iris-test-data#85 - rather than it being needed for any changes within Iris.

trexfeathers (approved):

rcomer:
Thanks @trexfeathers 😀

PR #5538, opened 2023-10-10
Splitattrs ncsave redo commonmeta

### LINKED ISSUE #5444: Common metadata operations on split-attributes
As mentioned in #5410 

This is about common-metadata operations : **equal(compare) / difference / combine**, as they apply to CubeMetadata, where the attributes properties are new-style "split" attribute dictionaries.

The #5410 PR provides that the new cube.attributes "CubeAttrsDict" can stand in for the legacy LimitedAttributeDicts where required, with a particular equivalence to the single dict form. Thus, everything using common metadata will still "work",i.e. not fail.

But @trexfeathers pointed out that we might want different behaviours, **_specific_** to split dictionaries.  
For example, do we consider cube metadata equal if their atributes are `CubeAttrsDict(globals={'x':1})` and `CubeAttrsDict(locals={'x':1})`?  These _are_ equivalent, when treated as single "combined" dictionaries, but that is probably not how  we want the operation to work.

The basis of a possible solution is suggested [here](https://github.com/pp-mo/iris/pull/77).

pp-mo:
Closes #5444

Specific handling of common-metadata operations on 'split' attribute dictionaries 
(i.e. for `CubeMetadata`)

---


pp-mo:
Ping @trexfeathers 
I hope this is comprehensible, following some further efforts !
This should be the last piece of the puzzle which completes https://github.com/orgs/SciTools/projects/5
 - except resolving the merge-back

lib/iris/common/metadata.py
```diff
@@ -140,6 +140,88 @@ def __new__(mcs, name, bases, namespace):
...
+# And to convert back again after the operation, if the result is a dictionary.
+#
+# For "strict" operations this clearly does all that is needed.  For lenient ones,
+# we _might_ want for local+global attributes of the same name to interact.
+# However, on careful consideration, it seems that this is not actually desirable for
+# any of the common-metadata operations.
+# So, we simply treat "global" and "local" attributes of the same name as entirely
+# independent. Which happily is also the easiest to code, and to explain.
+#
+def xd_is_split(dic):
```
trexfeathers, line 158: Considerations for 'packaging' of these convenience functions:

- `iris.common.metadata` is a public module. Do we definitely want these functions to be public? They have a very specific purpose.
- Would they be better in their own module since they are all related and distinct from everything that follows?
  e.g. `_cube_attr_handling.py`
- Can you explain the `xd_` naming convention? I don't get it!

lib/iris/common/metadata.py
```diff
@@ -140,6 +140,88 @@ def __new__(mcs, name, bases, namespace):
...
+    Convert the input to a 'normal' dict with paired keys, if it is split-attrs type
+    """
+
+    def _global_then_local_items(dic):
+        # Routine to produce global, then local 'items' in order, and with all keys
+        # "labelled" as local or global type, to ensure they are all unique.
+        for key, value in dic.globals.items():
+            yield ("global", key), value
+        for key, value in dic.locals.items():
+            yield ("local", key), value
```
trexfeathers, line 174: `"global"` vs `"local"` seems an intended use case for [enum](https://docs.python.org/3/library/enum.html), instead of working with straight strings. But I'll let you decide if that's overkill.

trexfeathers (changes requested):
I still need to review the tests but I wanted to give you something to start on.

lib/iris/tests/unit/common/metadata/test_CubeMetadata.py
```diff
@@ -247,6 +426,28 @@ def test_op_different__attribute_value(self, op_leniency):
...
+            cases={
+                "same": ("GaLb:GaLb", [True]),
+                "extra_global": ("GaL-:G-L-", [False, True]),
+                "extra_local": ("G-La:G-L-", [False, True]),
+                "same_global_local": ("GaL-:G-La", [False, True]),
+                "diff_global_local": ("GaL-:G-Lb", [False, True]),
+                "diffglobal_nolocal": ("GaL-:GbL-", [False]),
+                "diffglobal_samelocal": ("GaLc:GbLc", [False]),
+                "difflocal_noglobal": ("G-La:G-Lb", [False]),
+                "difflocal_sameglobal": ("GaLc:GaLd", [False]),
```
trexfeathers, line 446: How did you choose which circumstances to cover?

For instance there is nothing for same-global, one-local (with one local being absent).

I can't tell if these are oversights, or deliberate decisions based on the mechanics.

trexfeathers (changes requested):
Here's the rest of the review.

I ran `test_CubeMetadata.py` against the code in `FEATURE_split_attrs` and got 45 failures, so this clearly demonstrates we were missing something and you have now fixed it 👍

lib/iris/tests/unit/common/metadata/test_CubeMetadata.py
```diff
@@ -103,6 +108,180 @@ def op_leniency(request):
...
+            # Use a unique string to indicate a null result
+            result = "-"
+        else:
+            # result is a CubeMetadata whose .attributes is a pair of CubeAttrsDict
+            assert isinstance(result, CubeMetadata)
+            assert isinstance(result.attributes, tuple)
+            assert len(result.attributes) == 2
+            assert all(
+                isinstance(dic, CubeAttrsDict) for dic in result.attributes
+            )
```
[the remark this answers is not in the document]
pp-mo: I think that I have only used this in _test_ code, where being more explicit is useful,
but duck-typing is used in the functional code.

pp-mo:
## Additional note:
Spotted an outstanding bug [here in the tests](https://github.com/SciTools/iris/blob/28ebcbb4ffb922ef45a59cddfde425119d335ff9/lib/iris/tests/unit/common/metadata/test_CubeMetadata.py#L637-L649)

It seems clear that the second assignment to `self.lvalues` should really be to **`self.rvalues`**
That would also explain the final comment "Result has entirely EMPTY attributes ... is this maybe a mistake of the existing implementation ?"

lib/iris/tests/unit/common/metadata/test_CubeMetadata.py
```diff
@@ -456,6 +657,119 @@ def test_op_different__attribute_value(self, op_leniency):
...
+        else:
+            # strict operation : neither of the "extras" appears
+            pass
+
+        with mock.patch(
+            "iris.common.metadata._LENIENT", return_value=is_lenient
+        ):
+            # Check both l+r and r+l
+            assert lmetadata.combine(rmetadata)._asdict() == expected
+            assert rmetadata.combine(lmetadata)._asdict() == expected
```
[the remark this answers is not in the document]
pp-mo: I'm frankly not sure about these aspects any more, and I agree this needs clarifying + possibly simplifying.

I think I wrote these specific-testcase routines first to confirm some specific behaviours, before adding the "matrix" testing.  They certainly remain easier to understand than the 
There are at least a couple of additional thing which these specific test routines do :
  * they operate with standard fill-in items for all the metadata fields, not just testing 'attributes' alone
  * they all test that "l op r" has same (or equivalent) results as "r op l"

FYI I am still working on this...

trexfeathers:
### An update

@pp-mo and I realised that containing the specialised split-dict behaviour in method overrides in `CubeMetadata` would address most of my discomfort by giving the code an appropriate home and removing the need for conditional behaviour in the parent method.

lib/iris/tests/unit/common/metadata/test_CubeMetadata.py
```diff
@@ -247,6 +426,28 @@ def test_op_different__attribute_value(self, op_leniency):
...
+                "same": ("GaLb:GaLb", [True]),
+                "extra_global": ("GaL-:G-L-", [False, True]),
+                "extra_local": ("G-La:G-L-", [False, True]),
+                "same_global_local": ("GaL-:G-La", [False, True]),
+                "diff_global_local": ("GaL-:G-Lb", [False, True]),
+                "diffglobal_nolocal": ("GaL-:GbL-", [False]),
+                "diffglobal_samelocal": ("GaLc:GbLc", [False]),
+                "difflocal_noglobal": ("G-La:G-Lb", [False]),
+                "difflocal_sameglobal": ("GaLc:GaLd", [False]),
+                "diff_local_and_global": ("GaLc:GbLd", [False]),
```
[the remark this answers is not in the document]
pp-mo: I have completely now reviewed this, and am going to change it.

lib/iris/tests/unit/common/metadata/test_CubeMetadata.py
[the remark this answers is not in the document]
pp-mo: I am now confident that these **are** obsolete + I've removed them. https://github.com/SciTools/iris/pull/5538/commits/fb526243e84883c709d0671c8f31a42ea19016d9

In future, I'm also going to specifically test left-right vs right-left calculation order, which will replace the only extra benefit from these.

lib/iris/tests/unit/common/metadata/test_CubeMetadata.py
pp-mo: You're right, but it's triggered me to review all of this.
I'm now confident that more cases are needed, but also there are more factors I haven't accounted for.

Having prototyped a more comprehensive approach to that, I came up with a full matrix with 1152 tests (!!)
So I realised that I need to test AFAP for the expected 'regularities' in behaviour. Thus :
  * local and global attributes of a given attribute(name) should be calculated exactly the same way
  * local and global values (for same attribute) should be calculated independently of each other
  * each result for "left OP right" should be identical to "right OP left"

However, the basic calculation (combining a right + left value), must still be tested for each of the 3 operations, both strict+lenient, so there's still a fair number of tests.

I think I'm nearly on top of this now,
but I'm building a new testing mechanism, so don't fret about that code for now.

lib/iris/common/metadata.py
pp-mo: OK I have now re-worked all this to its own module, simplified + (hopefully) clarified.
https://github.com/SciTools/iris/pull/5538/commits/8d61d9b42c208c7a81b8332f961bb9b3fff1a1b7

lib/iris/tests/unit/common/metadata/test_CubeMetadata.py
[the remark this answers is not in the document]
pp-mo: Still looking into this.
See [comment below](https://github.com/SciTools/iris/pull/5538#discussion_r1391037920) about plans to reorganise the "matrix" testing

pp-mo:
OK @trexfeathers 
thanks for your patience since, following the previous review, I re-wrote most of the new testing framework !

I have just re-scanned + tweaked the comments+docstrings, and I have now definitely stopped fiddling with this.
I think I've basically now addressed the previous review comments (though you may disagree).
I'll re-scan the above outstanding comments + flag the changes made, where appropriate.

lib/iris/tests/unit/common/metadata/test_CubeMetadata.py
```diff
@@ -103,6 +108,180 @@ def op_leniency(request):
...
+    assert len(input_spec) == 2
+    # convert to a list of (global, *locals) value sets
+    input_values = decode_matrix_input(input_spec)
+
+    # form 2 inputs to the operation
+    def attrsdict(value):
+        if value is None:
+            result = {}
+        else:
+            result = {"_testattr_": value}
```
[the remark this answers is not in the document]
pp-mo: Ok sort of did this now : see [here](https://github.com/SciTools/iris/pull/5538/files#diff-cac8162ceb18e4c84f7a5fcd4b382197c2b42841aa402fd4bc0e545925b6b218R165-R166)

lib/iris/tests/unit/common/metadata/test_CubeMetadata.py
pp-mo: I have now removed the idea of "targetted" testcases with selected combinations of factors.
The new approach is , broadly :  [more matrixing](https://github.com/SciTools/iris/pull/5538/files#diff-cac8162ceb18e4c84f7a5fcd4b382197c2b42841aa402fd4bc0e545925b6b218R165-R166), leverage the expected ['same' results over different factors](https://github.com/SciTools/iris/pull/5538/files#diff-cac8162ceb18e4c84f7a5fcd4b382197c2b42841aa402fd4bc0e545925b6b218R145-R147), and [test the "local+global independence" separately from the other factors](https://github.com/SciTools/iris/pull/5538/files#diff-cac8162ceb18e4c84f7a5fcd4b382197c2b42841aa402fd4bc0e545925b6b218R403-R425), to keep the number of tests under control.

pp-mo:
Note: the linkcheck fail is fixed by #5556

lib/iris/tests/unit/common/metadata/test_CubeMetadata.py
```diff
@@ -95,15 +96,349 @@ def test_bases(self):
...
+    expected = _ALL_RESULTS[operation_name][primary_key][which]
+    if operation_name == "equal" and expected:
+        # Account for the equality cases made `False` by mismatched secondary values.
+        secondary_1, secondary_2 = secondary_inputs
+        if (
+            secondary_1 != "X"
+            and secondary_2 != "X"
+            and secondary_1 != secondary_2
+        ):
+            expected = False
```
trexfeathers, line 355: Doesn't this need some nuance depending on whether we are testing lenient or strict?

trexfeathers (changes requested):
Very elegant, and I think it all makes sense to me. Having stepped through locally.

I only noticed a couple of things that need some attention.

lib/iris/tests/unit/common/metadata/test_CubeMetadata.py
pp-mo: Hmm that's interesting...  Really good spot !
So, "primaryAA" plus "secondaryCX" would show this, being True if lenient and False if strict.
But I think this is missed because I am only doing the secondary options XX / CC / CD

So, if I extend this by adding a "secondaryCX" option, it does indeed fail, 
In `Test___eq__.test_splitattrs_global_local_independence[strict-primaryAA-secondaryCX]`,
the actual results are lenient=True, strict=False, but the "expected" was True in both cases.

I think that means that the above test needs more "nuance", depending on strict/lenient as you suggest.
I will look into it ...

So I do also think this makes a case for adding more 'secondary' options.  I'll think on.

pp-mo: Ok here's my fix : https://github.com/SciTools/iris/pull/5538/commits/dca5dce1d0e3d795329b819fb3d40d965752bb66

I also expanded the "secondary" test options, so we are now checking cases where a missing secondary value does or does not align with a missing primary value. 
I can confirm that some of those new cases _fail_ without the logic fix.

lib/iris/common/_split_attribute_dicts.py
```diff
@@ -0,0 +1,123 @@
...
+        # Convert all inputs into 'pairedkeys' type dicts
+        args = [_convert_splitattrs_to_pairedkeys_dict(arg) for arg in args]
+
+        result = operation(*args, **kwargs)
+
+        # Convert 'pairedkeys' dicts in the result back to split-attributes form.
+        if isinstance(result, Mapping):
+            # Fix a result which is a single dictionary -- for "combine"
+            result = _convert_pairedkeys_dict_to_splitattrs(result)
+        elif isinstance(result, Sequence) and len(result) == 2:
```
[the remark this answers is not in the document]
pp-mo: Ok, I hope that https://github.com/SciTools/iris/pull/5538/commits/dae3a28b74776662a54c3e6f8a91ef7939e20f95 addresses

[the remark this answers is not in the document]
pp-mo: After trying "else .. raise", I realise that we do **_want_** other types of result to pass unchanged.
Clarified in comments -- see : https://github.com/SciTools/iris/pull/5538/commits/b00b37fb0c70a3761ac39c3459876c1d37ed6569

trexfeathers (approved):

PR #5551, opened 2023-10-25
Allowing exemption to axis guessing on coords

### LINKED ISSUE #5003: Allow exemptions to coordinate axis guessing on save
Within the iris NetCDF saver [there is a call](https://github.com/SciTools/iris/blob/main/lib/iris/fileformats/netcdf/saver.py#L1956) to `guess_coord_axis`. This simple utility relies on coordinate names, units, or other attributes to ascribe an axis identifier to each dimension coordinate where possible. One of these guesses [checks whether the units of the coordinate are convertible](https://github.com/SciTools/iris/blob/04e757fe9e31f1fa7918ba12c760babdf61e6485/lib/iris/util.py#L262) to hectopascals, and if so presumes the dimension is a pressure coordinate corresponding to vertical levels; an axis identifier of Z is added.

Working with probabilistic data in which a dimension coordinate may describe different pressure thresholds, i.e. probability of pressure at mean sea level exceeding 1000hPa, 1010hPa, 1020hPa etc. this results in the threshold coordinate being identified as a vertical spatial dimension with an axis identifier of Z. This is undesirable and causes issues downstream of where the data is created (e.g. in Visual Weather). As this is added on save it is not possible to remove the attribute without using an alternative tool to Iris.

This issue requests that there be a means within Iris of preventing the axis identifiers being added to specific named (or var named) coordinates on save.

### Example cube
Below is shown a simple print out of a probabilistic PMSL forecast with a threshold coordinate named `air_pressure_at_sea_level` that has units of Pa, which is not a height coordinate, but which is assigned a Z axis identifier.
```
probability_of_air_pressure_at_sea_level_above_threshold / (1) (air_pressure_at_sea_level: 85; projection_y_coordinate: 970; projection_x_coordinate: 1042)
     Dimension coordinates:
          air_pressure_at_sea_level                                                      x                            -                             -
          projection_y_coordinate                                                        -                            x                             -
          projection_x_coordinate                                                        -                            -                             x
     Scalar coordinates:
          ...
```
ncdump header information corresponding to this coordinate:
```
float threshold(threshold) ;
	threshold:axis = "Z" ;
	threshold:units = "Pa" ;
	threshold:standard_name = "air_pressure_at_sea_level" ;
```

HGWright:
### Description

This PR adds a coordinate property `guess_coord` that forces `guess_coord_axis` to return `None`.

Tests for `guess_coord` and `guess_coord_axis` added.

Closes #5003

---

### DIFF
--- a/docs/src/whatsnew/latest.rst
+++ b/docs/src/whatsnew/latest.rst
@@ -29,9 +29,6 @@ This document explains the changes made to Iris for this release
 
 ✨ Features
 ===========
-
-#. `@rcomer`_ rewrote :func:`~iris.util.broadcast_to_shape` so it now handles
-   lazy data. (:pull:`5307`)
    
 #. `@trexfeathers`_ and `@HGWright`_ (reviewer) sub-categorised all Iris'
    :class:`UserWarning`\s for richer filtering. The full index of
@@ -48,7 +45,7 @@ This document explains the changes made to Iris for this release
    the year of December) instead of the following year (the default behaviour).
    (:pull:`5573`)
    
-   #. `@HGWright`_ added :func:`~iris.coords.Coord.ignore_axis` to allow manual
+   #. `@HGWright`_ added :attr:`~iris.coords.Coord.ignore_axis` to allow manual
    intervention preventing :func:`~iris.util.guess_coord_axis` from acting on a
    coordinate. (:pull:`5551`)
 


lib/iris/coords.py
```diff
@@ -1541,6 +1542,7 @@ def __init__(
         attributes=None,
         coord_system=None,
         climatological=False,
+        guess_coord=True,
```
trexfeathers, line 1545: This is the thing that everyone was against in our first meeting, after which we reduced the proposal to a coordinate property that could be set, but not during instantiation.

lib/iris/util.py
```diff
@@ -258,10 +258,17 @@ def guess_coord_axis(coord):
     This function maintains laziness when called; it does not realise data.
     See more at :doc:`/userguide/real_and_lazy_data`.
 
+    Can be skipped by setting the coordinate property of a coord guess_coord
+    to False.
```
trexfeathers, line 262:
```suggestion
    Can be skipped by setting the :class:`~iris.coords.Coord` 
    :attr:`~iris.coords.Coord.guess_coord` attribute
    to ``False``.
```

lib/iris/coords.py
```diff
@@ -2697,6 +2732,8 @@ def __init__(
             Will set to True when a climatological time axis is loaded
             from NetCDF.
             Always False if no bounds exist.
+        * guess_coord (bool):
+            When True: guess_coord will guess the coord, when False it won't.
```
bjlittle, line 2736: @HGWright This comment, no longer needed.

lib/iris/coords.py
```diff
@@ -1751,6 +1768,24 @@ def climatological(self, value):
...
+
+        Defaults to False, and when set to True it will be skipped by
+        guess_coord_axis.
+        """
+        return self._ignore_axis
+
+    @ignore_axis.setter
+    def ignore_axis(self, value):
+        if not isinstance(value, bool):
+            emsg = "Ignore_axis can only be set to True or False"
```
bjlittle, line 1785:
```suggestion
            emsg = "'ignore_axis' can only be set to 'True' or 'False'."
```

lib/iris/coords.py
```diff
@@ -861,7 +864,8 @@ def xml_element(self, doc):
                 element.setAttribute(
                     "climatological", str(self.climatological)
                 )
-
+        if self.guess_coord:
+            element.setAttribute("guess_coord", str(self.guess_coord))
```
bjlittle, line 868:
```suggestion
        if self.ignore_axis:
            element.setAttribute("ignore_axis", str(self.ignore_axis))
```

lib/iris/util.py
```diff
@@ -258,10 +258,17 @@ def guess_coord_axis(coord):
...
 
+    ``guess_coord_axis`` can be skipped by setting the coordinate property of a coord ignore_axis
+    to False.
+
     """
+
     axis = None
 
-    if coord.standard_name in (
+    if coord.ignore_axis is True:
```
bjlittle, line 268:
```suggestion
    if hasattr(coord, "ignore_axis") and coord.ignore_axis is True:
```

bjlittle (changes requested):
@HGWright It's taking shape, so thanks for persevering 💯 

He's some initial suggestions 👍

lib/iris/coords.py
```diff
@@ -861,7 +864,8 @@ def xml_element(self, doc):
                 element.setAttribute(
                     "climatological", str(self.climatological)
                 )
-
+        if self.ignore_axis:
+            element.setAttribute("ignore_axis", str(self.ignore_axis))
```
bjlittle, line 868:
```suggestion
        if hasattr(self, "ignore_axis") and self.ignore_axis:
            element.setAttribute("ignore_axis", str(self.ignore_axis))
```

bjlittle: Actually ... thinking on this, I don't believe that we should capture the state of `ignore_axis` within the CML. We're not interested in stateful equality here, and we've got separate unit test coverage to give us confidence that the new behaviour is working as expected.

So yeah, I vote from removing this. I genuinely don't see the benefit from the effort of extending the CML here.

@HGWright Thoughts?

lib/iris/tests/unit/util/test_guess_coord_axis.py
```diff
@@ -0,0 +1,51 @@
...
+class TestCoords:
+    def test_pressure_units(self, coord):
+        coord.units = "hPa"
+
+        assert guess_coord_axis(coord) == "Z"
+
+    def test_time_units(self, coord):
+        coord.units = "days since 1970-01-01 00:00:00"
+
+        assert guess_coord_axis(coord) == "T"
```
bjlittle, line 45: @HGWright You could parameterize these two tests into one ...

lib/iris/tests/unit/util/test_guess_coord_axis.py
```diff
@@ -0,0 +1,51 @@
...
+
+    def test_ignore_axis(self, coord):
+        coord.standard_name = "longitude"
+        coord.ignore_axis = True
+
+        assert guess_coord_axis(coord) is None
```
bjlittle, line 51: @HGWright Perhaps parameterize and explicitly test with `ignore_axis` being set to `True`  and then `False`.

lib/iris/util.py
```diff
@@ -258,10 +258,17 @@ def guess_coord_axis(coord):
     This function maintains laziness when called; it does not realise data.
     See more at :doc:`/userguide/real_and_lazy_data`.
 
+    ``guess_coord_axis`` can be skipped by setting the coordinate property of a coord ignore_axis
+    to False.
```
bjlittle, line 262:
```suggestion
    The ``guess_coord_axis`` behaviour can be skipped by setting the coordinate property ``ignore_axis`` to ``False``.
```

bjlittle (changes requested):
@HGWright This is looking really good.

Just a few minor actions to service, then we can bank this 👍

bjlittle:
@HGWright You also need to add a `whatsnew` entry 👍

bjlittle:
@HGWright Could you also raise an issue to capture the need to document these changes.

We can action that separately, rather than bloat the scope of this pull-request.

docs/src/whatsnew/latest.rst
```diff
@@ -30,6 +30,9 @@ This document explains the changes made to Iris for this release
 ✨ Features
 ===========
 
+#. `@rcomer`_ rewrote :func:`~iris.util.broadcast_to_shape` so it now handles
+   lazy data. (:pull:`5307`)
+
```
bjlittle, line 35: ?

docs/src/whatsnew/latest.rst
```diff
@@ -44,6 +47,10 @@ This document explains the changes made to Iris for this release
    Winter - December to February) will be assigned to the preceding year (e.g.
    the year of December) instead of the following year (the default behaviour).
    (:pull:`5573`)
+   
+   #. `@HGWright`_ added :func:`~iris.coords.Coord.ignore_axis` to allow manual
```
bjlittle, line 51:
```suggestion
   #. `@HGWright`_ added :attr:`~iris.coords.Coord.ignore_axis` to allow manual
```

bjlittle: For further details, see [here](https://www.sphinx-doc.org/en/master/usage/domains/python.html#role-py-attr).

bjlittle (approved):

PR #5561, opened 2023-10-31
More sensible time axis and tick labels for 2D plots

### LINKED ISSUE #5426: Quickplot Automatic Time Axis Label Completely Incorrect

rcomer:
### Description

1. Removed the `date2num` conversions from datetimes to floats in `plot.py`.  I'm not clear why these were there, but I assume were an ancient workaround for something.  They had been there since the initial commit to github.  Without them, the time axis ticks get labelled with datetime information instead of floats (see the updated images in the test data PR).
2. Pulled in the axis label change from #5484.  As discussed at #5426, we only want to label these axes as "time".
3. Reverted the quickplot change from #616, as [2] is a more general version of this.

Closes #5484 and closes #5426

---


rcomer:
Test data PR: https://github.com/SciTools/iris-test-data/pull/87

rcomer:
Note that users who specified a DateFormatter and DateLocator should not see any difference.  E.g. this plot is the same with `main` and with my branch:

```python
import matplotlib.dates as mdates

import iris
import iris.plot as iplt
import iris.tests.stock

cube = iris.tests.stock.realistic_4d()
cube = cube[:, 0, 0, :]

fig, _ = plt.subplots(layout="constrained")

iplt.contourf(cube)
ax = plt.gca()
ax.yaxis.set_major_locator(mdates.MinuteLocator(interval=10))
ax.yaxis.set_major_formatter(mdates.DateFormatter("%M minutes after %H"))

plt.savefig("test.png")
```

test

lib/iris/plot.py
```diff
@@ -454,10 +453,6 @@ def _draw_2d_from_bounds(draw_method_name, cube, *args, **kwargs):
                 else:
                     values = coord.contiguous_bounds()
                     values = _fixup_dates(coord, values)
-                    if values.dtype == np.dtype(object) and isinstance(
-                        values[0], datetime.datetime
-                    ):
-                        values = mpl_dates.date2num(values)
```
trexfeathers, line 460: Been staring at these blocks for the past hour trying to work out WTF they are doing here. Nothing stops working without them - I ran all the tests locally - but as you say  we could be looking at some outdated defensive code. What is more confusing to me is that they completely undermine the upstream use of `_fixup_dates()` - why have both?!

I even tried introducing a non-standard calendar but still everything works without these blocks being here.

lib/iris/quickplot.py
```diff
@@ -117,10 +108,8 @@ def _label_with_points(cube, result=None, ndims=2, coords=None, axes=None):
 def _get_titles(u_object, v_object):
     if u_object is None:
         u_object = iplt._u_object_from_v_object(v_object)
-    xunits = u_object is not None and not u_object.units.is_time_reference()
```
trexfeathers, line 120: Are we losing anything without `u_object is not None`, or was that just 'asking permission' before running the rest of the line?

lib/iris/tests/results/imagerepo.json
trexfeathers, line 1: Can confirm that these correspond with the failures I get locally, and with SciTools/iris-test-data#87 👍

lib/iris/plot.py
```diff
@@ -557,10 +552,6 @@ def _draw_2d_from_points(draw_method_name, arg_func, cube, *args, **kwargs):
                     )
                 plot_arrays.append(np.arange(values.size))
                 string_axes[axis_name] = values
-            elif values.dtype == np.dtype(object) and isinstance(
-                values[0], datetime.datetime
-            ):
-                plot_arrays.append(mpl_dates.date2num(values))
```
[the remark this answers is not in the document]
trexfeathers: Thanks for the reasoned argument. Makes sense to me

lib/iris/plot.py
trexfeathers: Good point and 2199 is a nice find, too

trexfeathers (approved):

PR #5564, opened 2023-11-02
added phrasebook

### LINKED ISSUE #5091: Create phrase book for xarray to/from iris
## 📚 Documentation

With the progress of #4994, perhaps a phrase book of sorts could be added to the documentation, to translate terminology between xarray, iris, and other similar libraries. 
e.g. Iris might use non-lazy, whereas xarray uses eager.

ESadek-MO:
Closes #5091.
 
Created a phrasebook for comparing similar packages. 

This is intended as a starting basis, with space for more to be added at a later date.

### DIFF
--- a/docs/src/community/index.rst
+++ b/docs/src/community/index.rst
@@ -40,12 +40,14 @@ smoother interoperability:
 
 * The :mod:`iris.pandas` module
 * :doc:`iris_xarray`
+* :doc:`phrasebook`
 
 .. toctree::
    :maxdepth: 1
    :hidden:
 
    iris_xarray
+   phrasebook
 
 Plugins
 -------
--- a/docs/src/community/phrasebook.rst
+++ b/docs/src/community/phrasebook.rst
@@ -1,0 +1,66 @@
+.. include:: ../common_links.inc
+
+.. _phrasebook:
+
+Package Phrasebook
+===================
+
+There are a number of similar packages to Iris, and a lot of these have their own
+terminology for similar things. Whether you're coming or going, we hope this might
+be a helpful guide to these differences!
+Definitions for each can be found in :ref:`glossary`. See also
+`Xarray terminology <https://docs.xarray.dev/en/stable/user-guide/terminology.html>`_.
+
+.. list-table:: Phrasebook
+   :widths: 25 25 25 50
+   :header-rows: 1
+
+   * - Iris
+     - Xarray
+     - Example
+     - Notes
+   * - Non-Lazy
+     - Eager
+     -
+     - Used to relate to functions, rather than the data.
+   * - Cube
+     - DataArray
+     -
+     -
+   * - CubeList
+     - Dataset
+     -
+     - Though similar, a CubeList is a simpler object, and is
+       not a perfect comparison to a Dataset
+   * - Merge/ Concatenate
+     - Concatenate
+     - `Xarray concatenate <https://docs.xarray.dev/en/stable/user-guide/combining.html#concatenate>`_
+     - Xarray's concatenate has the capability to largely do what both
+       Iris merge and Iris concatenate do. However, this is not a perfect comparison,
+       please see the link for more information.
+   * -
+     - Merge
+     - `Xarray merge <https://docs.xarray.dev/en/stable/user-guide/combining.html#merge>`_
+     - Xarray's Merge function doesn't map neatly map to any Iris feature.
+       Please see the link for more information.
+   * - Scalar Coordinate
+     -
+     -
+     - Iris makes a distinction between scalar coordinates and non-scalar coordinates,
+       whereas xarray documentation makes a distinction between scalar and non-scalar *data*.
+       It is possible to make coordinates with scalar data in both Iris and xarray
+       but only Iris will label such coordinates.
+   * - AuxCoord
+     - Non-Dimensional Coordinate
+     -
+     - Coordinates in Iris and xarray are categorised using different rules,
+       and so are not a one-to-one match.
+   * - DimCoord
+     - Dimension Coordinate
+     -
+     - Coordinates in Iris and xarray are categorised using different rules,
+       and so are not a one-to-one match.
+
+----
+
+`To top <phrasebook_>`_
--- a/docs/src/community/iris_xarray.rst
+++ b/docs/src/community/iris_xarray.rst
@@ -7,6 +7,7 @@ Iris ❤️ :term:`Xarray`
 There is a lot of overlap between Iris and :term:`Xarray`, but some important
 differences too. Below is a summary of the most important differences, so that
 you can be prepared, and to help you choose the best package for your use case.
+See :doc:`phrasebook` for a broad comparison of terminology.
 
 Overall Experience
 ------------------
--- a/docs/src/whatsnew/latest.rst
+++ b/docs/src/whatsnew/latest.rst
@@ -93,6 +93,9 @@ This document explains the changes made to Iris for this release
 
 #. `@tkknight`_  removed broken git links. (:pull:`5569`)
 
+#. `@ESadek-MO`_ added a phrasebook for synonymous terms used in similar
+   packages. (:pull:`5564`)
+
 
 💼 Internal
 ===========


docs/src/community/phrasebook.rst
```diff
@@ -0,0 +1,63 @@
...
+     -
+     -
+   * - Merge
+     - Concatenate
+     -
+     - Functionality is essentially swapped.
+   * - Concatenate
+     - Merge
+     -
+     - Functionality is essentially swapped.
```
stephenworsley, line 41: From what I've read of the xarray documentation, I'm not sure it's quite as simple as marge and concatenate being swapped versions of what they are in iris. The way I read it,  xarray concatenate can act like iris merge but it can also act like iris concatenate. It looks like xarray marge is not quite equivalent to either merge or concatenate since its output is a Dataset, and while Iris has CubeList, this lacks some of the features of a Dataset so I don't think it would be possible to have something that functions quite like xarray merge in Iris.

docs/src/community/phrasebook.rst
```diff
@@ -0,0 +1,63 @@
...
+     -
+     - Used to relate to functions, rather than the data.
+   * - Cube
+     - DataArray
+     -
+     -
+   * - Cubelist
+     - Dataset
+     -
+     -
```
stephenworsley, line 33: It's worth noting that `Datasets` are quite a a bit richer in their data model than `CubeList`s, which are effectively just a collection of objects with some methods attached. I think the major difference from what I can gather is that coordinates are a proper member of a `Dataset` while for a `CubeList`, a coordinate can only be accessed via one of the `Cubes` contained within it. I think in this table it's worth just noting that a `CubeList` is a much simpler object and the correspondance is not exact.

docs/src/community/phrasebook.rst
```diff
@@ -0,0 +1,63 @@
...
+     - Notes
+   * - Non-Lazy
+     - Eager
+     -
+     - Used to relate to functions, rather than the data.
+   * - Cube
+     - DataArray
+     -
+     -
+   * - Cubelist
```
stephenworsley, line 30: The L should be capitalised in `CubeList`.

docs/src/community/phrasebook.rst
```diff
@@ -0,0 +1,63 @@
...
+     -
+     - Functionality is essentially swapped.
+   * - Concatenate
+     - Merge
+     -
+     - Functionality is essentially swapped.
+   * - Scalar Coordinate
+     - Scalar
+     -
+     - Coordinates in Iris and Xarray are split via rules, and so are not a
```
stephenworsley, line 45: I'm not entirely clear what "split via rules" means in this context, could this be clarified?

stephenworsley (changes requested):
Looking good, Just a couple notes here, mostly I think it's worth being a bit more precise about the differences between Iris and xarray.

Also, there ought to be a link to this documentation page somewhere here https://github.com/SciTools/iris/blob/main/docs/src/community/index.rst and also perhaps here https://github.com/SciTools/iris/blob/main/docs/src/community/iris_xarray.rst .

docs/src/community/phrasebook.rst
```diff
@@ -0,0 +1,67 @@
...
+     - Merge
+     - `Xarray merge <https://docs.xarray.dev/en/stable/user-guide/combining.html#merge>`_
+     - Xarray's Merge function doesn't map neatly map to any iris feature.
+       Please see the link for more information.
+   * -
+     - Scalar
+     -
+     - Iris makes a distinction between scalar coordinates and non-scalar
+       coordinates, xarray does not. In xarray, a scalar is any non-array or
+       array-like object e.g. int, float or str.
```
stephenworsley, line 51: I'm still not entirely sure about the escond sentance here. I feel like the concept of a "scalar" described here is something that is implicit in Iris as well so this isn't really a unique xarray term. I feel like the real difference worth highlighting is the fact that while Iris documentation attaches the scalar label to the coordinate, xarray documentation applies the scalar label to the data. I also think it's worth making it clear that it is possible to make coordinates with scalar data in both iris and xarray which are effectively equivalent. This may be getting a bit wordy, but maybe something like:

Iris makes a distinction between scalar coordinates and non-scalar coordinates, xarray documentation makes a distinction between scalar and non-scalar __data__. It is possible to make coordinates with scalar data in both iris and xarray but only iris will label such coordinates.

docs/src/community/phrasebook.rst
```diff
@@ -0,0 +1,67 @@
...
+   * - AuxCoord
+     - Non-Dimensional Coordinate
+     -
+     - Coordinates in Iris and Xarray are categorised using rules, and so are not a
```
stephenworsley, line 55: do you mean to say "categorised using different rules"?

docs/src/community/phrasebook.rst
```diff
@@ -0,0 +1,67 @@
...
+       one-to-one match. This comparison is most accurate, but others could
+       also work.
+   * - DimCoord
+     - Dimension Coordinate
+     -
+     - Coordinates in Iris and Xarray are categorised using rules, and so are not a
```
stephenworsley, line 61: do you mean to say "categorised using different rules"?

docs/src/community/phrasebook.rst
```diff
@@ -0,0 +1,67 @@
...
+       one-to-one match. This comparison is most accurate, but others could
+       also work.
```
stephenworsley, line 63: I don't think this last sentence is necessary and I think this would be clearer without it.

docs/src/community/phrasebook.rst
```diff
@@ -0,0 +1,67 @@
...
+     -
+     - Iris makes a distinction between scalar coordinates and non-scalar
+       coordinates, xarray does not. In xarray, a scalar is any non-array or
+       array-like object e.g. int, float or str.
+   * - AuxCoord
+     - Non-Dimensional Coordinate
+     -
+     - Coordinates in Iris and Xarray are categorised using rules, and so are not a
+       one-to-one match. This comparison is most accurate, but others could
+       also work.
```
stephenworsley, line 57: I don't think this last sentence is necessary and I think this would be clearer without it.

docs/src/community/index.rst
```diff
@@ -40,6 +40,7 @@ smoother interoperability:
 
 * The :mod:`iris.pandas` module
 * :doc:`iris_xarray`
+* :doc:`phrasebook`
```
stephenworsley, line 43: I think this could be added to the toctree below too.

stephenworsley (changes requested):

docs/src/community/phrasebook.rst
```diff
@@ -0,0 +1,66 @@
...
+     - Merge
+     - `Xarray merge <https://docs.xarray.dev/en/stable/user-guide/combining.html#merge>`_
+     - Xarray's Merge function doesn't map neatly map to any iris feature.
+       Please see the link for more information.
+   * - Scalar Coordinate
+     -
+     -
+     - Iris makes a distinction between scalar coordinates and non-scalar coordinates,
+       whereas xarray documentation makes a distinction between scalar and non-scalar *data*.
+       It is possible to make coordinates with scalar data in both iris and xarray
```
stephenworsley, line 51: Iris should have a capital I in keeping with the rest of the documentation.

docs/src/community/phrasebook.rst
```diff
@@ -0,0 +1,66 @@
...
+       but only iris will label such coordinates.
+   * - AuxCoord
+     - Non-Dimensional Coordinate
+     -
+     - Coordinates in Iris and Xarray are categorised using different rules,
```
stephenworsley, line 56: xarray should have a lower case x in keeping with xarray documentation.

docs/src/community/phrasebook.rst
```diff
@@ -0,0 +1,66 @@
...
+       and so are not a one-to-one match.
+   * - DimCoord
+     - Dimension Coordinate
+     -
+     - Coordinates in Iris and Xarray are categorised using different rules,
```
stephenworsley, line 61: xarray should have a lower case x in keeping with xarray documentation.

docs/src/community/phrasebook.rst
```diff
@@ -0,0 +1,66 @@
...
+   * - Merge/ Concatenate
+     - Concatenate
+     - `Xarray concatenate <https://docs.xarray.dev/en/stable/user-guide/combining.html#concatenate>`_
+     - Xarray's concatenate has the capability to largely do what both
+       iris merge and iris concatenate do. However, this is not a perfect comparison,
+       please see the link for more information.
+   * -
+     - Merge
+     - `Xarray merge <https://docs.xarray.dev/en/stable/user-guide/combining.html#merge>`_
+     - Xarray's Merge function doesn't map neatly map to any iris feature.
```
stephenworsley, line 44: Iris should have a capital I in keeping with the rest of the documentation.

stephenworsley (changes requested):
Just a couple minor capitalisation changes and this is looking good.

stephenworsley (approved):
Looks good to me. Good job!

PR #5566, opened 2023-11-07
TESTING - DO NOT MERGE.  Added sphinx_search.extension

tkknight:
### Description
Added support for search as you type via the [readthedocs-sphinx-search](https://readthedocs-sphinx-search.readthedocs.io/en/latest/index.html) package.

The change only includes the extension if it is running on the read the docs servers, it will not run in a local dev environment (but could if installed via pip).

This change may help users find what they want as the results are more interactive.

**TESTING - DO NOT MERGE.**

---

### DIFF
--- a/.readthedocs.yml
+++ b/.readthedocs.yml
@@ -29,3 +29,5 @@ python:
   install:
     - method: pip
       path: .
+      extra_requirements:
+        - docs
--- a/docs/src/conf.py
+++ b/docs/src/conf.py
@@ -164,6 +164,7 @@ extensions = [
 ]
 
 if on_rtd:
+    # search as you type on readthedocs only.
     extensions.extend(["sphinx_search.extension"])
 
 if skip_api == "1":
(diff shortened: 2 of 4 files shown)


tkknight:
The docs buils for this PR does not seem to work properly, however is build against the latest is does in my own RTD account, see https://tkknight-iris-test-doc.readthedocs.io/en/latest/#

This works but not very well:

- it is **not** hooked into the search bar.  I think this is because the theme has change the role to not be "search", see https://readthedocs-sphinx-search.readthedocs.io/en/latest/customization.html#custom-search-input.  You can bring up the search as you type window by pressing the backslash "/"
- when using dark mode the text is white on a white background so cannot be seen
- the searching seems less effective!

bjlittle:
@tkknight That's disappointing ... initially this seemed like a neat, easy win.

So is this pretty much dead in the water as far as you're concerned?

I'm not sure what to suggest ...

PR #5565, opened 2023-11-06
Merge chunk control code into latest iris

ESadek-MO:
## 🚀 Pull Request

Addresses step one of #5399.

Takes code from #4572 and solves conflicts with latest changes in Iris.

PR #5575, opened 2023-11-10
Chunk control modes

### LINKED ISSUE #5399: Finish NetCDF loading chunk control code
## (Complete the below work on a feature branch)

Documentation has been separated into another issue to make it easier to find time for the work. So chunk control will not be 'releaseable' when this issue is complete.

- [x] Create a new PR applying #4572 to the latest Iris code
  (this code has developed a lot so the original PR is too stale for updating) (#5565) 
- [x] Action original conversations
  https://github.com/SciTools/iris/pull/4572#discussion_r810056992 may need more discussion; suggest that mirroring Dask 'modes' should only happen if the behaviour is identical, otherwise the modes deserve their own names (`"default"` or `"no-chunk"` or whatever) to avoid possible confusion. (#5575)
- [x] Flesh out the tests #5583

ESadek-MO:
## 🚀 Pull Request

Follows #5565, closes #5399 .
Implemented new modes, as discussed in a comment in aforementioned issue.

- [x] Default
- [x] File
- [x] ~Guess~
- [x] -1
- [ ] Tests
- [x] Doc strings

**WILL NEED REBASING ONCE PRECEDING PR IS MERGED.**

### DIFF
--- a/lib/iris/_lazy_data.py
+++ b/lib/iris/_lazy_data.py
@@ -225,7 +225,9 @@ def _optimum_chunksize(
     )
 
 
-def as_lazy_data(data, chunks=None, asarray=False, dims_fixed=None, dask_chunking=False):
+def as_lazy_data(
+    data, chunks=None, asarray=False, dims_fixed=None, dask_chunking=False
+):
     """
     Convert the input array `data` to a :class:`dask.array.Array`.
 
@@ -267,7 +269,9 @@ def as_lazy_data(data, chunks=None, asarray=False, dims_fixed=None, dask_chunkin
     """
     if dask_chunking:
         if chunks is not None:
-            raise ValueError(f"Dask chunking chosen, but chunks already assigned value {chunks}")
+            raise ValueError(
+                f"Dask chunking chosen, but chunks already assigned value {chunks}"
+            )
         lazy_params = {"asarray": asarray, "meta": np.ndarray}
     else:
         if chunks is None:
@@ -281,15 +285,20 @@ def as_lazy_data(data, chunks=None, asarray=False, dims_fixed=None, dask_chunkin
         if all(elem > 0 for elem in data.shape):
             # Expand or reduce the basic chunk shape to an optimum size.
             chunks = _optimum_chunksize(
-                chunks, shape=data.shape, dtype=data.dtype, dims_fixed=dims_fixed
+                chunks,
+                shape=data.shape,
+                dtype=data.dtype,
+                dims_fixed=dims_fixed,
             )
-            lazy_params = {"chunks": chunks, "asarray": asarray, "meta": np.ndarray}
+            lazy_params = {
+                "chunks": chunks,
+                "asarray": asarray,
+                "meta": np.ndarray,
+            }
     if isinstance(data, ma.core.MaskedConstant):
         data = ma.masked_array(data.data, mask=data.mask)
     if not is_lazy_data(data):
-        data = da.from_array(
-            data, **lazy_params
-        )
+        data = da.from_array(data, **lazy_params)
     return data
 
 
--- a/lib/iris/_lazy_data.py
+++ b/lib/iris/_lazy_data.py
@@ -290,11 +290,11 @@ def as_lazy_data(
                 dtype=data.dtype,
                 dims_fixed=dims_fixed,
             )
-            lazy_params = {
-                "chunks": chunks,
-                "asarray": asarray,
-                "meta": np.ndarray,
-            }
+        lazy_params = {
+            "chunks": chunks,
+            "asarray": asarray,
+            "meta": np.ndarray,
+        }
     if isinstance(data, ma.core.MaskedConstant):
         data = ma.masked_array(data.data, mask=data.mask)
     if not is_lazy_data(data):
--- a/lib/iris/_lazy_data.py
+++ b/lib/iris/_lazy_data.py
@@ -225,7 +225,7 @@ def _optimum_chunksize(
     )
 
 
-def as_lazy_data(data, chunks=None, asarray=False, dims_fixed=None):
+def as_lazy_data(data, chunks=None, asarray=False, dims_fixed=None, as_dask=False):
     """
     Convert the input array `data` to a :class:`dask.array.Array`.
 
@@ -260,26 +260,34 @@ def as_lazy_data(data, chunks=None, asarray=False, dims_fixed=None):
         but reduced by a factor if that exceeds the dask default chunksize.
 
     """
-    if chunks is None:
-        # No existing chunks : Make a chunk the shape of the entire input array
-        # (but we will subdivide it if too big).
-        chunks = list(data.shape)
+    if as_dask:
+        if chunks is not None:
+            raise ValueError(f"Dask chunking chosen, but chunks already assigned value {chunks}")
+    else:
+        if chunks is None:
+            # No existing chunks : Make a chunk the shape of the entire input array
+            # (but we will subdivide it if too big).
+            chunks = list(data.shape)
 
-    # Adjust chunk size for better dask performance,
-    # NOTE: but only if no shape dimension is zero, so that we can handle the
-    # PPDataProxy of "raw" landsea-masked fields, which have a shape of (0, 0).
-    if all(elem > 0 for elem in data.shape):
-        # Expand or reduce the basic chunk shape to an optimum size.
-        chunks = _optimum_chunksize(
-            chunks, shape=data.shape, dtype=data.dtype, dims_fixed=dims_fixed
-        )
-
+        # Adjust chunk size for better dask performance,
+        # NOTE: but only if no shape dimension is zero, so that we can handle the
+        # PPDataProxy of "raw" landsea-masked fields, which have a shape of (0, 0).
+        if all(elem > 0 for elem in data.shape):
+            # Expand or reduce the basic chunk shape to an optimum size.
+            chunks = _optimum_chunksize(
+                chunks, shape=data.shape, dtype=data.dtype, dims_fixed=dims_fixed
+            )
     if isinstance(data, ma.core.MaskedConstant):
         data = ma.masked_array(data.data, mask=data.mask)
     if not is_lazy_data(data):
-        data = da.from_array(
-            data, chunks=chunks, asarray=asarray, meta=np.ndarray
-        )
+        if as_dask:
+            data = da.from_array(
+                data, asarray=asarray, meta=np.ndarray
+            )
+        else:
+            data = da.from_array(
+                data, chunks=chunks, asarray=asarray, meta=np.ndarray
+            )
     return data
 
 
--- a/lib/iris/_lazy_data.py
+++ b/lib/iris/_lazy_data.py
@@ -225,7 +225,7 @@ def _optimum_chunksize(
     )
 
 
-def as_lazy_data(data, chunks=None, asarray=False, dims_fixed=None, as_dask=False):
+def as_lazy_data(data, chunks=None, asarray=False, dims_fixed=None, dask_chunking=False):
     """
     Convert the input array `data` to a :class:`dask.array.Array`.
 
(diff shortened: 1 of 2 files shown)


lib/iris/fileformats/netcdf/loader.py
```diff
@@ -610,3 +657,132 @@ def load_cubes(file_sources, callback=None, constraints=None):
...
+
+        Notes
+        -----
+        This function acts as a contextmanager, for use in a 'with' block.
+        """
+        try:
+            self.mode = "file"
+            yield
+        finally:
+            self.mode = None
```
trexfeathers, line 777: Having read through everything, I think we might expected 'stacked' uses of this context manager in some cases. Would therefore be safer to:

- Store the previous state of `self.mode` and restore it afterwards
- Store the previous state of `var_dim_chunksizes` and restore it afterwards

lib/iris/fileformats/netcdf/loader.py
```diff
@@ -610,3 +657,132 @@ def load_cubes(file_sources, callback=None, constraints=None):
...
+    def from_file(self) -> None:
+        """
+        Ensures the chunks are loaded in from file variables, else will throw an error.
+
+        Notes
+        -----
+        This function acts as a contextmanager, for use in a 'with' block.
+        """
+        try:
+            self.mode = "file"
```
trexfeathers, line 774: Since there is now only 1 `mode`, perhaps this could be a boolean instead?

lib/iris/fileformats/netcdf/loader.py
```diff
@@ -610,3 +657,132 @@ def load_cubes(file_sources, callback=None, constraints=None):
...
+
+        Parameters
+        ----------
+        var_names : str or list of str
+            apply the ``dimension_chunksizes`` controls only to these variables,
+            or when building cubes from these data variables.
+            If None (the default), settings apply to all loaded variables.
+        dimension_chunksizes : dict: str --> int
+            Kwargs specifying chunksizes for dimensions of file variables.
+            Each key-vchunkalue pair defines a chunksize for a named file
```
trexfeathers, line 704:
```suggestion
            Each key-value pair defines a chunksize for a named file
```

lib/iris/fileformats/netcdf/loader.py
```diff
@@ -230,20 +237,48 @@ def _get_cf_var_data(cf_var, filename):
...
+                chunks = list(cf_var.shape)
+
+            # Modify the chunking in the context of an active chunking control.
+            # N.B. settings specific to this named var override global ('*') ones.
+            dim_chunks = CHUNK_CONTROL.var_dim_chunksizes.get(
+                cf_var.cf_name
+            ) or CHUNK_CONTROL.var_dim_chunksizes.get("*")
+            if not dim_chunks:
+                dims_fixed = None
+            else:
```
trexfeathers, line 258: Don't forget to set `dims_fixed` to all `True` if we're in `from_file` mode.

trexfeathers (changes requested):

lib/iris/fileformats/netcdf/loader.py
```diff
@@ -751,6 +763,24 @@ def set(
...
+        -----
+        This function acts as a contextmanager, for use in a 'with' block.
+        """
+        try:
+            old_file_mode = self.file_mode
+            old_var_dim_chunksizes = self.var_dim_chunksizes
+            self.file_mode = True
+            yield
+        finally:
+            self.file_mode = old_file_mode
```
trexfeathers, line 781: We should be doing the mirror of this in `self.set()`, too (sorry I missed that before).

trexfeathers (changes requested):
Looking good. Couple of things I missed originally...

lib/iris/fileformats/netcdf/loader.py
```diff
@@ -610,3 +659,135 @@ def load_cubes(file_sources, callback=None, constraints=None):
...
+        """
+        Ensures the chunks are loaded in from file variables, else will throw an error.
+
+        Notes
+        -----
+        This function acts as a contextmanager, for use in a 'with' block.
+        """
+        try:
+            old_file_mode = self.file_mode
+            old_var_dim_chunksizes = self.var_dim_chunksizes
```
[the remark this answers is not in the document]
ESadek-MO: move assignments out of try

lib/iris/fileformats/netcdf/loader.py
```diff
@@ -673,6 +684,7 @@ def __init__(self, var_dim_chunksizes=None):
 
         """
         self.var_dim_chunksizes = var_dim_chunksizes or {}
+        self.file_mode = False
```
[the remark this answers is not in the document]
ESadek-MO: convert self.file_mode into self.mode, an enum or string.
new mode, as_dask, passed into lazy_data as param

ESadek-MO:
In _lazy_data, there are two checks for as_dask, at `263` and `283`. 
This *might* not be needed, if the check at `280` could be done before `275`.

lib/iris/_lazy_data.py
```diff
@@ -225,7 +225,9 @@ def _optimum_chunksize(
     )
 
 
-def as_lazy_data(data, chunks=None, asarray=False, dims_fixed=None):
+def as_lazy_data(
+    data, chunks=None, asarray=False, dims_fixed=None, as_dask=False
```
trexfeathers, line 229: `as_dask` needs a docstring entry

lib/iris/fileformats/netcdf/loader.py
```diff
@@ -236,7 +237,19 @@ def _get_cf_var_data(cf_var, filename):
             )
             # Get the chunking specified for the variable : this is either a shape, or
             # maybe the string "contiguous".
+            if CHUNK_CONTROL.mode == CHUNK_CONTROL.Modes.AS_DASK:
```
trexfeathers, line 240: Enumeration checks should be `is`, rather than `==` - it's the actual same object rather than just an equivalent one.

lib/iris/fileformats/netcdf/loader.py
```diff
@@ -236,7 +237,19 @@ def _get_cf_var_data(cf_var, filename):
...
+                return as_lazy_data(proxy, chunks=None, as_dask=True)
             chunks = cf_var.cf_data.chunking()
+            #
```
trexfeathers, line 243:
```suggestion
```

lib/iris/fileformats/netcdf/loader.py
```diff
@@ -673,6 +691,8 @@ def __init__(self, var_dim_chunksizes=None):
 
         """
         self.var_dim_chunksizes = var_dim_chunksizes or {}
+        self.Modes = Enum("Modes", ["DEFAULT", "FROM_FILE", "AS_DASK"])
```
trexfeathers, line 694: This object never changes as a result of what's going on in the instance, so it should be a class variable instead - i.e. defined at the top of the class, not in `__init__()` - I think that's common practice for `Enum`s like this.

lib/iris/fileformats/netcdf/loader.py
trexfeathers, line 240:
```suggestion
            if CHUNK_CONTROL.mode == ChunkControl.Modes.AS_DASK:
```

See my comment about a class variable instead of an instance variable (and sorry for getting this wrong when I was demonstrating!).

lib/iris/fileformats/netcdf/loader.py
```diff
@@ -236,7 +237,19 @@ def _get_cf_var_data(cf_var, filename):
             )
             # Get the chunking specified for the variable : this is either a shape, or
             # maybe the string "contiguous".
+            if CHUNK_CONTROL.mode == CHUNK_CONTROL.Modes.AS_DASK:
+                return as_lazy_data(proxy, chunks=None, as_dask=True)
             chunks = cf_var.cf_data.chunking()
+            #
+            if (
+                chunks is None
+                and CHUNK_CONTROL.mode == CHUNK_CONTROL.Modes.FROM_FILE
```
trexfeathers, line 246:
```suggestion
                and CHUNK_CONTROL.mode == ChunkControl.Modes.FROM_FILE
```

See my comment about a class variable instead of an instance variable (and sorry for getting this wrong when I was demonstrating!).

lib/iris/fileformats/netcdf/loader.py
```diff
@@ -247,17 +260,22 @@ def _get_cf_var_data(cf_var, filename):
             dim_chunks = CHUNK_CONTROL.var_dim_chunksizes.get(
                 cf_var.cf_name
             ) or CHUNK_CONTROL.var_dim_chunksizes.get("*")
-            if not dim_chunks:
+            dims = cf_var.cf_data.dimensions
+            if CHUNK_CONTROL.mode == CHUNK_CONTROL.Modes.FROM_FILE:
```
trexfeathers, line 264:
```suggestion
            if CHUNK_CONTROL.mode == ChunkControl.Modes.FROM_FILE:
```

See my comment about a class variable instead of an instance variable (and sorry for getting this wrong when I was demonstrating!).

lib/iris/_lazy_data.py
```diff
@@ -260,26 +262,37 @@ def as_lazy_data(data, chunks=None, asarray=False, dims_fixed=None):
...
     if not is_lazy_data(data):
-        data = da.from_array(
-            data, chunks=chunks, asarray=asarray, meta=np.ndarray
-        )
+        if as_dask:
+            data = da.from_array(data, asarray=asarray, meta=np.ndarray)
+        else:
+            data = da.from_array(
+                data, chunks=chunks, asarray=asarray, meta=np.ndarray
+            )
```
trexfeathers, line 295: >In _lazy_data, there are two checks for as_dask, at `263` and `283`. 
>This *might* not be needed, if the check at `280` could be done before `275`. 

If you're interested in efficiency then the neatest approach IMO would be to contain `asarray` and `meta` in a `dict` - constructed at the top of the function - then optionally add `chunks` to this `dict` in the correct place.

Then you can just have a single call to `da.from_array()`:

```python
da.from_array(data, **my_dict)
```

lib/iris/fileformats/netcdf/loader.py
trexfeathers, line 694: I like this neater "functional syntax". Unfortunately it seems that not all IDEs can handle it, but are fine with the "class syntax". Given code completion etcetera is a big part of the `Enum` utility, I'd prefer the "class syntax" purely to help IDEs.

lib/iris/fileformats/netcdf/loader.py
```diff
@@ -236,7 +237,19 @@ def _get_cf_var_data(cf_var, filename):
             )
             # Get the chunking specified for the variable : this is either a shape, or
             # maybe the string "contiguous".
+            if CHUNK_CONTROL.mode == CHUNK_CONTROL.Modes.AS_DASK:
+                return as_lazy_data(proxy, chunks=None, as_dask=True)
```
trexfeathers, line 241: We typically avoid mutiple `return` statements at different points in a function/method, as it can be harder to follow. I would find it more readable if there is just the one call to `as_lazy_data()` at the end of this function, even though that requires `as_dask` to ALWAYS have a value (`False`/`True`).

trexfeathers (changes requested):
Sorry for so many comments. They're all easy!

lib/iris/fileformats/netcdf/loader.py
```diff
@@ -236,34 +237,51 @@ def _get_cf_var_data(cf_var, filename):
...
-                dims_fixed = None
+            if CHUNK_CONTROL.mode is ChunkControl.Modes.AS_DASK:
+                result = as_lazy_data(proxy, chunks=None, dask_chunking=True)
             else:
-                # Modify the chunks argument, and pass in a list of 'fixed' dims, for
-                # any of our dims which are controlled.
+                chunks = cf_var.cf_data.chunking()
+                if (
+                    chunks is None
+                    and CHUNK_CONTROL.mode == ChunkControl.Modes.FROM_FILE
```
trexfeathers, line 246:
```suggestion
                    and CHUNK_CONTROL.mode is ChunkControl.Modes.FROM_FILE
```

trexfeathers (approved):

PR #5577, opened 2023-11-13
Relicense to from LGPL-3 to BSD-3

lbdreyer:
Closes #5148

It has now passed the deadline we set in #5148 by when objections to relicensing to BSD needed to be raised.

PR #5583, opened 2023-11-15
Chunk Control Tests

### LINKED ISSUE #5399: Finish NetCDF loading chunk control code

ESadek-MO:
## 🚀 Pull Request

Closes #5399.

### Description
Adds tests for changes made on branch [FEATURE_chunk_control](https://github.com/SciTools/iris/tree/FEATURE_chunk_control).

- [x] move tests to netcdf file, and rename to remove leading `netcdf_`
- [x] Convert existing tests to PyTest
- [x] Add tests for `-1` functionality
- [x] Add test for `from_file` context manager
- [x] Add test for `as_dask` context manager
- [x] Add test for `as_lazy_data` to ensure dask bypassing works

Tests are in some cases close enough to add parameterisation, but I'm not convinced it would be readable enough, and would need to pass in parameters and results.


lib/iris/tests/integration/netcdf/test__chunk_control.py
```diff
@@ -0,0 +1,161 @@
...
+
+    loader.as_lazy_data = Mock(side_effect=RuntimeError(message))
+    with CHUNK_CONTROL.as_dask():
+        try:
+            iris.load_cube(tmp_filepath, create_cube[0])
+        except RuntimeError as e:
+            if str(e) == message:
+                pass
+            else:
+                raise e
```
trexfeathers, line 154: What happens if you don't include the `side_effect`? Does that avoid needing the error handling too?

Is this one of the cases where patching is more appropriate?

lib/iris/tests/unit/lazy_data/test_as_lazy_data.py
```diff
@@ -42,6 +44,26 @@ def test_non_default_chunks(self):
...
+        data = np.arange(24)
+        chunks = (12,)
+        optimum = self.patch("iris._lazy_data._optimum_chunksize")
+        optimum.return_value = chunks
+        with self.assertRaises(ValueError) as ar:
+            as_lazy_data(data, chunks=chunks, dask_chunking=True)
+        self.assertEqual(
+            str(ar.exception),
+            f"Dask chunking chosen, but chunks already assigned value {chunks}",
+        )
```
trexfeathers, line 65: Use [`assertRaisesRegex()`](https://docs.python.org/3/library/unittest.html#unittest.TestCase.assertRaisesRegex) to check for the message at the same time.

trexfeathers (changes requested):
Here's some stuff.

Everything that is written is written well. I still need to think about whether anything is missing.

lib/iris/tests/integration/netcdf/test__chunk_control.py
```diff
@@ -0,0 +1,160 @@
...
+@pytest.fixture()
+def create_cube(tmp_filepath):
+    cube = istk.simple_4d_with_hybrid_height()
+    cube_varname = "my_var"
+    sigma_varname = "my_sigma"
+    cube.var_name = cube_varname
+    cube.coord("sigma").var_name = sigma_varname
+    cube.coord("sigma").guess_bounds()
+    iris.save(cube, tmp_filepath)
+    yield cube_varname, sigma_varname
```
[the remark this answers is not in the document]
ESadek-MO: This is a ghost of previous attempts, forgot to revert!

lib/iris/tests/integration/netcdf/test__chunk_control.py
ESadek-MO: Patching is more appropriate, working on that. I'm unfamiliar with Mock and Patch, and the latter doesn't seem to have "assert_called_with".

As to the `side_effect`, i'd have to set an output of the Mock so that it doesn't fail out on a different error. Which is doable, but since all we're checking is how the Mock is called, seems unnecessary, albeit uglier.

ESadek-MO: Hopefully, patching has been fixed, used as a context manager. I still haven't made a decision on side_effect

lib/iris/tests/integration/netcdf/test__chunk_control.py
```diff
@@ -0,0 +1,161 @@
...
+@pytest.fixture(autouse=True)
+def remove_min_bytes():
+    old_min_bytes = loader._LAZYVAR_MIN_BYTES
+    loader._LAZYVAR_MIN_BYTES = 0
+    yield None
+    loader._LAZYVAR_MIN_BYTES = old_min_bytes
+
+
+def test_default(tmp_filepath, create_cube):
+    cube = iris.load_cube(tmp_filepath, create_cube[0])
```
trexfeathers, line 62: Using `iris.load_cube()` within a test makes us vulnerable to a lot of unrelated breakages - it's a big call stack.

These are effectively unit tests, but because they are testing loading we can't get them to be super low level. We can get them a bit lower, though: please could you use `iris.fileformats.netcdf.loader.load_cubes()` instead of `iris.load_cube()`? This bypasses a lot of the call stack.

lib/iris/tests/integration/netcdf/test__chunk_control.py
```diff
@@ -0,0 +1,161 @@
+# Copyright Iris contributors
+#
+# This file is part of Iris and is released under the LGPL license.
+# See COPYING and COPYING.LESSER in the root of the repository for full
+# licensing details.
+"""Integration tests for loading and saving netcdf files."""
```
trexfeathers, line 6: This file is really filling the unit test role. I reckon move these into a section in [`unit/fileformats/netcdf/loader/test_load_cubes.py`](https://github.com/SciTools/iris/blob/80c179251219eebdfd03e72b0df123c2cfca61fd/lib/iris/tests/unit/fileformats/netcdf/loader/test_load_cubes.py)

Or maybe a dedicated module called `test_load_cubes_chunk_control.py`?

I don't think there are any true integration tests that are worth adding for this feature.

lib/iris/tests/integration/netcdf/test__chunk_control.py
trexfeathers, line 1: Testing that @pp-mo felt was worth adding:

>Also, by pinning only certain dimensions, I can get the [automatic operation](https://github.com/SciTools/iris/blob/93b0917bd8152df163e5accaada75c6344e30c78/lib/iris/_lazy_data.py#L50) to optimise the sizes in the other dims.
E.G. if I had automatic chunks of (1, 1, 100, 1000, 2000) ; if I then pin the 3rd dim to 3 I will then get (1, 33, 3, 1000, 2000) instead, which is still "optimally" sized.
(TBD: this code needs properly testing !!)

trexfeathers (changes requested):
Need something in [`test__get_cf_var_data.py`](
https://github.com/SciTools/iris/blob/80c179251219eebdfd03e72b0df123c2cfca61fd/lib/iris/tests/unit/fileformats/netcdf/loader/test__get_cf_var_data.py). Not a full replication of what we have written elsewhere, but just something to demonstrate that `_get_cf_var_data()` responds to `CHUNK_CONTROL`.

lib/iris/tests/integration/netcdf/test__chunk_control.py
```diff
@@ -0,0 +1,157 @@
...
+    with patch("iris.fileformats.netcdf.loader.as_lazy_data") as optimum:
+        optimum.side_effect = RuntimeError(message)
+        with CHUNK_CONTROL.as_dask():
+            try:
+                iris.load_cube(tmp_filepath, create_cube[0])
+            except RuntimeError as e:
+                if str(e) == message:
+                    pass
+                else:
+                    raise e
```
trexfeathers, line 152: This is the alternative, if you are interested:

```suggestion
        class AsDaskError(Exception):
            # Allows calling the Mock without needing defined Mock behaviour.
            pass
        
        optimum.side_effect = AsDaskError(message)
        with CHUNK_CONTROL.as_dask():
            try:
                iris.load_cube(tmp_filepath, create_cube[0])
            except AsDaskErroras:
                pass
```

lib/iris/tests/integration/netcdf/test__chunk_control.py
```diff
@@ -0,0 +1,157 @@
+# Copyright Iris contributors
+#
+# This file is part of Iris and is released under the LGPL license.
+# See COPYING and COPYING.LESSER in the root of the repository for full
+# licensing details.
```
trexfeathers, line 5:
```suggestion
# This file is part of Iris and is released under the BSD license.
# See LICENSE in the root of the repository for full licensing details.
```

#5577

lib/iris/tests/unit/fileformats/netcdf/loader/test__chunk_control.py
```diff
@@ -0,0 +1,173 @@
+# Copyright Iris contributors
+#
+# This file is part of Iris and is released under the LGPL license.
+# See COPYING and COPYING.LESSER in the root of the repository for full
+# licensing details.
+"""Integration tests for loading and saving netcdf files."""
```
trexfeathers, line 6:
```suggestion
"""Unit tests for :class:`iris.fileformats.netcdf.loader.ChunkControl`."""
```

trexfeathers (changes requested):

lib/iris/tests/integration/netcdf/test__chunk_control.py
ESadek-MO: I don't think there's much merit in one over the other, and since I already have one functioning, I'll probably leave as is.

trexfeathers (changes requested):

trexfeathers (approved):

Issue #4988, opened 2022-09-23
split-attrs: equalise-attributes


pp-mo:
Possible need ?

**_At least_** must ensure that the operation makes sense with the new split-form attributes, 
( i.e. using existing equalise code with the new 'cube.attributes' combined access approach : #4983 )

However, we may need to handle newly-possible collisions between matching local+global settings in a more intelligent way
  - possibly provide automatic resolution of some such problems  -- or detect + error/warn
  - possibly add extra keywords to control different resolution modes

pp-mo:
The spec for this, according to the docstring, is 
`This function deletes any attributes which are not the same for all the
given cubes.  The cubes will then have identical attributes, and the
removed attributes are returned.
`

The implementation uses iteration over keys, so that could in some cases break the contract.
E.g. with `cube_a.attributes.globals['x'] = 3` and `cube_b.attributes.locals['x'] = 3`, it will conclude that these are equal, so neither cube will be modified.  But still, afterwards, "cube_a.attributes != cube_b.attributes" .

This could cause breaking changes when cubes are loaded from different sources (at least, if any are netcdf), or constructed manually.
And this is potentially important, because then the cubes would not merge.
( since merge does equality testing of attributes -- see [here ](https://github.com/SciTools/iris/issues/5585#issuecomment-1813015312) )

Probably

Issue #5585, opened 2023-11-15
split-attrs: merge and concatenate


pp-mo:
Following on from https://github.com/SciTools/iris/issues/4988

Since (we think) the number-1 use of `equalise_attributes` is to align cubes for merging, it begs the question whether merge + concatenate may have some cases "broken" by the adoption of split-attributes.

pp-mo:
We are mostly concerned about cases where some new behaviour might break existing code.

From offline discussion with @trexfeathers  ...

Looking into the merge + concatenate code, it seems that 
 * merge [compares attributes](https://github.com/SciTools/iris/blob/v3.7.0/lib/iris/_merge.py#L393), in the course of doing [match checking](https://github.com/SciTools/iris/blob/v3.7.0/lib/iris/_merge.py#L445), but it does _not_ interrogate their contents in any special way (++)
 * concatenate does not deal with cube attributes specifically, but only does [whole metadata comparison](https://github.com/SciTools/iris/blob/v3.7.0/lib/iris/_concatenate.py#L595-L600)

(++)
**_Except_** ... merge does make a key-by-key analysis of mismatched cubes to construct [an account of "why a merge failed"](https://github.com/SciTools/iris/blob/v3.7.0/lib/iris/_merge.py#L394-L415),

This means that, in the new world, we could get a.attributes != b.attributes, but then this message, being based on comparing values per-key, would conclude that there are _no_ keys will different values, leading to a message like
`"cube.attributes values differ for keys: ."`  (an empty comma-separated list, i.e. invisible!)


So, that bit could be improved -- and actually, we could make use of metadata.attributes.difference instead.

PR #5152, opened 2023-02-08
Mergeback of "Feature _split_attrs" branch

pp-mo:
**PURELY DRAFT : do not merge**

The purpose of this is to spot when we get any conflicts with main.

PR #5586, opened 2023-11-15
Splitattrs equalise

pp-mo:
Addresses #4988 
---

### DIFF
--- a/lib/iris/tests/unit/util/test_equalise_attributes.py
+++ b/lib/iris/tests/unit/util/test_equalise_attributes.py
@@ -13,7 +13,6 @@ Unit tests for the :func:`iris.util.equalise_attributes` function.
 import iris.tests as tests  # isort:skip
 
 import numpy as np
-import pytest
 
 from iris.coords import AuxCoord
 from iris.cube import Cube, CubeAttrsDict
@@ -220,7 +219,6 @@ class TestSplitattributes:
         result = self.check_equalised_result(cube1, cube2)
         assert result == ["XX", "XX"]
 
-    @pytest.mark.xfail
     # This case fails without the split-attributes fix.
     def test__oneglobal_onelocal__same(self):
         cube1 = self._sample_splitattrs_cube("AX")
@@ -234,7 +232,6 @@ class TestSplitattributes:
         result = self.check_equalised_result(cube1, cube2)
         assert result == ["XX", "XX"]
 
-    @pytest.mark.xfail
     # This case fails without the split-attributes fix.
     def test__sameglobals_onelocal__same(self):
         cube1 = self._sample_splitattrs_cube("AA")
@@ -242,7 +239,6 @@ class TestSplitattributes:
         result = self.check_equalised_result(cube1, cube2)
         assert result == ["XX", "XX"]
 
-    @pytest.mark.xfail
     # This case fails without the split-attributes fix.
     def test__differentglobals_samelocals(self):
         cube1 = self._sample_splitattrs_cube("AC")
--- a/lib/iris/tests/unit/util/test_equalise_attributes.py
+++ b/lib/iris/tests/unit/util/test_equalise_attributes.py
@@ -13,9 +13,15 @@ Unit tests for the :func:`iris.util.equalise_attributes` function.
 import iris.tests as tests  # isort:skip
 
 import numpy as np
+import pytest
 
-from iris.cube import Cube
+from iris.coords import AuxCoord
+from iris.cube import Cube, CubeAttrsDict
 import iris.tests.stock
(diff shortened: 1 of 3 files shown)


pp-mo:
https://github.com/SciTools/iris/pull/5586/commits/3cb375e7da4a6c993ca696a1f89ae1cd6e49d183 at this point, demonstrated expected-failures
(see [test log](https://github.com/SciTools/iris/actions/runs/6882661416/job/18721610409?pr=5586) )

pp-mo:
@trexfeathers are you interested in this ?

We still have the little thing to fix merge errors in #5585, 
and the merge-back #5152,
but https://github.com/orgs/SciTools/projects/5 is nearly nearly there !

lib/iris/util.py
```diff
@@ -2059,31 +2059,53 @@ def equalise_attributes(cubes):
...
+    from iris.cube import CubeAttrsDict
+
+    cube_attrs = [cube.attributes for cube in cubes]
     removed = []
+
+    # *IF* they are cube attributes, then convert to "paired-keys" form.
+    is_split_dicts = any(
+        isinstance(attrs, CubeAttrsDict) for attrs in cube_attrs
+    )
+    if is_split_dicts:
```
trexfeathers, line 2075: This is already handled by your addition to `_convert_splitattrs_to_pairedkeys_dict()`:

https://github.com/SciTools/iris/blob/35ec095a6293394c599c71a8238db5f10e2f9e13/lib/iris/common/_split_attribute_dicts.py#L39-L41

So can we get away without any checks at either end?

lib/iris/util.py
```diff
@@ -2059,31 +2059,53 @@ def equalise_attributes(cubes):
...
-                and np.all(cube.attributes[key] == cubes[0].attributes[key])
-            )
+            if (key in cube_keys and np.all(attrs[key] == cube_attrs[0][key]))
         ]
     keys_to_remove.difference_update(common_keys)
 
-    # Remove all the other attributes.
+    # rework the keys list if we were handling split-attributes.
+    if is_split_dicts:
+        # Collect all the ('local'/'global, name) keys + retain just the names.
```
trexfeathers, line 2097:
```suggestion
        # Collect all the ('local'/'global', name) keys + retain just the names.
```

lib/iris/util.py
```diff
@@ -2059,31 +2059,53 @@ def equalise_attributes(cubes):
...
+    # deferred imports to avoid circularity problem
+    from iris.common._split_attribute_dicts import (
+        _convert_splitattrs_to_pairedkeys_dict,
+    )
+    from iris.cube import CubeAttrsDict
+
+    cube_attrs = [cube.attributes for cube in cubes]
     removed = []
+
+    # *IF* they are cube attributes, then convert to "paired-keys" form.
```
trexfeathers, line 2071: Can this comment be enhanced to explain why the rest of the code 'magically works' once the keys are paired keys?

lib/iris/tests/unit/util/test_equalise_attributes.py
```diff
@@ -153,5 +158,111 @@ def test_complex_somecommon(self):
...
 
+class TestSplitattributes:
+    """
+    Extra testing for cases where attributes differ specifically by type
+
+    That is, where there is a new possibility of 'mismatch' due to the newer "typing"
+    of attributes as global or local.
+
+    Specifically, it is now possible that although
+    "cube1.attributes.keys == cube2.attributes.keys()",
```
trexfeathers, line 169:
```suggestion
    "cube1.attributes.keys() == cube2.attributes.keys()",
```

trexfeathers (changes requested):

lib/iris/util.py
pp-mo: The original point here was the the conversion to/from paired-keys type is only needed for split-type attributes
(which technically we should _always_ have, if it is only used on cubes, but I still think it's safer to not insist on that)

On reflection it should still _work_, if less efficient, to do this with regular attributes dictionaries too -- i.e. always + regardless of type.
I will try that, as it clearly would make the code simpler.

pp-mo: Yup, it works !
https://github.com/SciTools/iris/pull/5586/commits/c3e2175b7ea7735ed8ab2b49d926d70922a4afde

lib/iris/util.py
pp-mo: I've tried now https://github.com/SciTools/iris/pull/5586/commits/c3e2175b7ea7735ed8ab2b49d926d70922a4afde
Please see what you think !

pp-mo:
Thanks @trexfeathers I think the latest addresses your points so far.
Please take a look + see what you think now.

trexfeathers (approved):

PR #5595, opened 2023-11-21
Splitattrs ncsave deprecation

pp-mo:
Although we handled the split-attrs save control, we forgot to provide a deprecation warning.
So that's what this does.


ESadek-MO (approved):
Thanks @pp-mo, this looks good to me!

PR #5600, opened 2023-11-22
Fix test to ignore extra warnings from later versions of distributed.

pp-mo:
To fix #5547 problem

It seems that, when dask/distributed version 2023.9.3 --> 2023.11.0, some of our testcases trigger an additional warning about transferring large data to workers (only when using distributed).

I've confirmed that this isn't a _new_ inefficiency :  The test takes around 5 seconds via distributed, compared to 1 sec with a threaded or synchronous scheduler, and that performance hit is basically the same both before + after the version change.
So I guess it is just a new _warning_.

The tests that fail are testing the exact "data fill" warnings issued when saving to netcdf.
This PR gets the offending tests to explicitly ignore any warnings which look like that new message.

### DIFF
--- a/lib/iris/tests/integration/netcdf/test_delayed_save.py
+++ b/lib/iris/tests/integration/netcdf/test_delayed_save.py
@@ -5,6 +5,7 @@
 """
 Integration tests for delayed saving.
 """
+import re
 import warnings
 
 from cf_units import Unit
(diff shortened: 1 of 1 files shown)


trexfeathers (approved):
Thanks for your diligence, @pp-mo!

Issue #4990, opened 2022-09-23
Rethink "extra" user info sections


pp-mo:
## 📚 Documentation
Since #4661, what were "Iris Technical Papers" ([were here](https://scitools-iris.readthedocs.io/en/v3.0.0/techpapers/index.html)), are now moved to be part of the "Developer docs" ([now here](https://scitools-iris.readthedocs.io/en/stable/techpapers/index.html))
The existing items here are :
  1. "Iris Handling of PP and Fieldsfiles", and
  2. "Missing Data Handling in Iris"

Meanwhile, with #3890, the User Guide has gained an extra "Further Topics" section, now containing:
  1. "Metadata"
  2. "Lenient Metadata"
  3. "Lenient Cube Maths", and 
  4. "Mesh Support"

I think the result is now a bit confused + inappropriate :  None of the above are really developer-specific information, and the information sections like "Metadata" and "Mesh Support" are if anything of _less_ general interest than the former "Technical Papers".

So I think these could usefully **_all_** go in the "Further Topics" section of the User Guide.  

**What does anyone else think of this ?**
@bjlittle @tkknight @trexfeathers

pp-mo:
Some possible problems, though ?
  * the older "technical papers" might want some rewrite to unify the styles a bit
  * the "metata" sections look like they might usefully be grouped together.  But I don't know if sections + subsections are easily achievable here ?

trexfeathers:
My personal instinct would be a dedicated heading for Further Topics and Technical Papers combined. But I'm not certain how that fits with the '4-quadrant' approach researched by @jamesp and @wjbenfold.

wjbenfold:
I'm not about to completely redesign the docs (probably) but will happily opine on them a bit:

* Making a call about what's important for users and what's important for devs (can we use different words, given that users might consider themselves devs? Anyway...) seems sensible. I'd expect some users to benefit from an understanding of why choices have been made and the mental models behind them, so maybe there will be more in the user section than we'd expect. There could always be stubs in the dev bit that linked to the user bit?
* I think we'd ideally be using a 4-quadrant style of **tutorials**, **how-tos**, **reference** and **explanation** - to me I think further topics tends to fall into explanation (leaving api docs as the reference). Based on that, and based on a quick skim of the contents of both, I'd stick them all in a "explanation" or "background information" or something. There might be an argument for having explanations for users and explanations for devs, though it seems like they're all user-type ones?

rcomer:
[...]

PR #5602, opened 2023-11-24
Relocated the Technical Papers documentation to Further Topics. 

ESadek-MO:
## 🚀 Pull Request

Relocated the Technical Papers documentation to Further Topics. 

Relevant #4990 

4990 suggests removing the Technical Papers heading in general, which I have not done, but could do.


docs/src/developers_guide/contributing_getting_involved.rst
```diff
@@ -61,6 +61,6 @@ If you are new to using GitHub we recommend reading the
 
    ../generated/api/iris
    ../whatsnew/index
-   ../techpapers/index
+   ../../further_topics/techpapers/index
```
ESadek-MO, line 64: Do we still want this? 

It means it'd still be accessible via Developers_Guide(>Sidebar)>References, and would just be a shortcut to it's new home in Further Topics.

I've left this in so it's easier for users to find, but it feels like it undermines the point of it perhaps?

pp-mo: Personally I **_think_** I would have preferred that the old "tech papers" should just be merged into the"further topics" list.
But I note that this would mean there is nothing for that old whatsnew link to point **_at_**.  
Which is a problem I hadn't even thought of before this ... https://github.com/SciTools/iris/issues/5603

I don't think we can fix that problem here+now (!) -- it may want further discussion anyway.

So I think this is OK at least as an interim solution.
Or, if you fancy it, you could merge all topics but also create an overview page for the list, which could act as a target for this link.  @trexfeathers suggests that he'd like to have an overview page for the further topics anyway.

pp-mo:
Hi @ESadek-MO I see this is pretty close.

The way you fixed the failure is interesting ..  
I was rather expecting that the "Further topics" and "Tech papers" would just become one long list.
But if we **_did_** do that, there would no longer be anything for that link to point at. See #5603 

But I think all those objections require more thinking + decisions, a bit of a debate.
So I don't think it's worth trying to resolve any of it here, and we may as well merge this as it currently done.

----
There is also another problem which I think this creates .. separate comment coming (doesn't link to your existing changes)

docs/src/developers_guide/contributing_getting_involved.rst
pp-mo: No, I'd say we **_don't_** want this link any more.

ESadek-MO:
Converted to draft so it doesn't get confused with #5604.

pp-mo (approved):
Thanks @ESadek-MO 

I think this looks great now.  
A lot less fussy, and generally more visible to users.
⭐

PR #5610, opened 2023-12-05
Faster and simpler iris.util.array_equal

bouweandela:
### Description

The function `iris.util.array_equal` realizes the data multiple times, instead of just once. This means that data is pulled from disk multiple times, which is slower than necessary. This pull request simplifies the code so it lets Dask figure out what the best way to handle the computation is.

---

### DIFF
--- a/lib/iris/util.py
+++ b/lib/iris/util.py
@@ -22,7 +22,7 @@ import numpy as np
 import numpy.ma as ma
 
 from iris._deprecation import warn_deprecated
-from iris._lazy_data import as_concrete_data, is_lazy_data, is_lazy_masked_data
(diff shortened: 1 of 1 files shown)


pp-mo:
> The answer is simpler than I expected, have a look at the `dask.array.Array.__bool__` method:
> https://github.com/dask/dask/blob/191d39177009d2cce25b818878118e35329b6db3/dask/array/core.py#L1860-L1867

Thanks,  I see !

I still think it's a shame this is not highlighted in the user docs somewhere, though.
I was surprised when I found out !

bouweandela:
Good point. I opened a pull request to document it: https://github.com/dask/dask/pull/10686

PR #5620, opened 2023-12-11
Add option to specify chunks in `iris.util.broadcast_to_shape`

### LINKED ISSUE #5456: Add the ability to specify chunks to `iris.util.broadcast_to_shape`
## ✨ Feature Request

I would like to be able to specify the target chunks in [`iris.util.broadcast_to_shape`](https://scitools-iris.readthedocs.io/en/latest/generated/api/iris.util.html#iris.util.broadcast_to_shape).

## Motivation

This function is often used to broadcast auxiliary coordinate or cell measure data or weights to the same shape as the data from a cube. If the chunks of the broadcasted data do not match that of the cube, this leads to poor computational performance.

## Additional context
Here is an example implementation (taken from https://github.com/ESMValGroup/ESMValCore/pull/2120):

```python
def broadcast_to_shape(array, shape, chunks, dim_map):
    """Copy of `iris.util.broadcast_to_shape` that allows specifying chunks."""
    if isinstance(array, da.Array):
        chunks = list(chunks)
        for src_idx, tgt_idx in enumerate(dim_map):
            # Only use the specified chunks along new dimensions.
            chunks[tgt_idx] = array.chunks[src_idx]
        broadcast_to = partial(da.broadcast_to, chunks=chunks)
    else:
        broadcast_to = np.broadcast_to

    n_orig_dims = len(array.shape)
    n_new_dims = len(shape) - n_orig_dims
    array = array.reshape(array.shape + (1,) * n_new_dims)

    # Get dims in required order.
    array = np.moveaxis(array, range(n_orig_dims), dim_map)
    new_array = broadcast_to(array, shape)

    if np.ma.isMA(array):
        # broadcast_to strips masks so we need to handle them explicitly.
        mask = np.ma.getmask(array)
        if mask is np.ma.nomask:
            new_mask = np.ma.nomask
        else:
            new_mask = np.broadcast_to(mask, shape)
        new_array = np.ma.array(new_array, mask=new_mask)

    elif is_lazy_masked_data(array):
        # broadcast_to strips masks so we need to handle them explicitly.
        mask = da.ma.getmaskarray(array)
        new_mask = broadcast_to(mask, shape)
        new_array = da.ma.masked_array(new_array, new_mask)

    return new_array

```

bouweandela:
### Description

Make it possible to specify the target chunks in [iris.util.broadcast_to_shape](https://scitools-iris--5620.org.readthedocs.build/en/5620/generated/api/iris.util.html#iris.util.broadcast_to_shape).

This function is often used to broadcast auxiliary coordinate or cell measure data or weights to the same shape as the data from a cube. If the chunks of the broadcasted data do not match that of the cube, this leads to poor computational performance. Rechunking after broadcasting is less efficient, therefore it is really useful to be able to specify the chunks directly as part of the broadcast.

Closes #5456

---

### DIFF
--- a/lib/iris/util.py
+++ b/lib/iris/util.py
@@ -50,9 +50,10 @@ def broadcast_to_shape(array, shape, dim_map, chunks=None):
         If the source array is a :class:`dask.array.Array` and a value is
         provided, then the result will use these chunks instead of the same
         chunks as the source array. Setting chunks explicitly as part of
-        broadcast_to_shape is more efficient than rechunking afterwards. The
-        values provided here will only be used along dimensions that are new on
-        the result or have size 1 on the source array.
+        broadcast_to_shape is more efficient than rechunking afterwards. See
+        also :func:`dask.array.broadcast_to`. Note that the values provided
+        here will only be used along dimensions that are new on the result or
+        have size 1 on the source array.
 
     Examples
     --------
--- a/lib/iris/util.py
+++ b/lib/iris/util.py
@@ -25,7 +25,7 @@ from iris.common.lenient import _lenient_client
 import iris.exceptions
 
 
-def broadcast_to_shape(array, shape, dim_map):
+def broadcast_to_shape(array, shape, dim_map, chunks=None):
     """Broadcast an array to a given shape.
 
     Each dimension of the array must correspond to a dimension in the
@@ -46,6 +46,13 @@ def broadcast_to_shape(array, shape, dim_map):
         the index in *shape* which the dimension of *array* corresponds
         to, so the first element of *dim_map* gives the index of *shape*
         that corresponds to the first dimension of *array* etc.
+    chunks : :class:`tuple`, optional
+        If the source array is a :class:`dask.array.Array` and a value is
+        provided, then the result will use these chunks instead of the same
+        chunks as the source array. Setting chunks explicitly as part of
+        broadcast_to_shape is more efficient than rechunking afterwards. The
+        values provided here will only be used along dimensions that are new on
+        the result or have size 1 on the source array.
 
     Examples
     --------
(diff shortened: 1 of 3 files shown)


bouweandela:
It looks like the link check is failing because of the use of `http://` instead of `https://`. I opened #5621 to see if I can fix this.

trexfeathers:
@pp-mo any update on this?

pp-mo:
> @pp-mo any update on this?

Sorry yes I am reviewing this, I have some queries + suggestions pending .
I has taken me quite a while to get my head around it -- should finally get to posting later today I think !

lib/iris/util.py
```diff
@@ -46,6 +46,13 @@ def broadcast_to_shape(array, shape, dim_map):
         the index in *shape* which the dimension of *array* corresponds
         to, so the first element of *dim_map* gives the index of *shape*
         that corresponds to the first dimension of *array* etc.
+    chunks : :class:`tuple`, optional
```
HGWright, line 49: Sorry for the delay, there was some confusion on my end about if we were asking for Pytest rather than Unit test, but we are grandfathering in everything made before last month so no changes needed on the testing front. The only thing I would ask is that this docstring could do with referencing the Dask docs for broadcast_to where this appears to be inspired from stating that this is the standard behaviour according to dask and not new functionality that we have decided. Thanks

HGWright (changes requested):

lib/iris/util.py
bouweandela: Thank you for reviewing! I added a link in [f12a80c](https://github.com/SciTools/iris/pull/5620/commits/f12a80cafb319e96cef0b47b8b39001c2be1bbd7). Does it look better now?

PR #5623, opened 2023-12-11
Adopt inital noop ruff linter

bjlittle:
### Description

This pull-request adds the configuration for a no-op `ruff` linter i.e., the `ruff` linter runs but all rules are ignored, hence there are no compliance exceptions raised.

Note that, it adds the `pre-commit` hook to ensure that `ruff` (and it's configured rules) are applied for developers using `pre-commit` and on pull-requests via the `pre-commit.ci` service.

Also note that `ALL` linter rules have been selected in the `pyproject.toml`, and each of the currently existing 58 `ruff` linter rules (see [here](https://docs.astral.sh/ruff/rules/)) have been ignored in the `.ruff.toml` configuration file, which extends the `ruff` configuration defined in the `pyproject.toml`.

As a strategy going forwards, I'd recommend that we:

- add **permanent** rule and file exceptions to the `pyproject.toml`
- use the `.ruff.toml` to collate all **temporary** rule and file exceptions, but aim to address these all over time.

Ultimately, the goal would be to have **no temporary exceptions** in the `.ruff.toml` configuration file. Also, we should always aim to minimize our permanent (documented) exceptions.

We can also slowly drop use of all the third-party packages that `ruff` replaces, namely:

- `black`
- `flake8`
- `isort`

Taking the above approach will allow us to adopt and become more compliant with `ruff` over time, at our own pace. 

This strategy is open-ended, in that we have enabled `ALL` linter rules for `ruff` (with explicit `ignores`) so that we automatically adopt new non-`preview` linter rules by default. If new rules are implemented within `ruff` that break us, then either we address them straight away or add temporary/permanent exceptions as we deem fit.

---


tkknight (approved):
This looks great @bjlittle.  It enables ruff but we can adopt the rules over time.

PR #5621, opened 2023-12-11
Update links to https

bouweandela:
### Description

Update `http://` links to `https://` to fix the [failing link check](https://github.com/SciTools/iris/actions/runs/7165540693/job/19507691774?pr=5620).

---
[Consult Iris pull request check list]( https://scitools-iris.readthedocs.io/en/latest/developers_guide/contributing_pull_request_checklist.html)

PR #5628, opened 2023-12-12
Use https for opendap.org link, as http no longer supported.

pp-mo:
Should fix #5619 CI linkcheck failure

### DIFF
--- a/docs/src/conf.py
+++ b/docs/src/conf.py
@@ -396,6 +396,7 @@ linkcheck_ignore = [
     "https://www.metoffice.gov.uk/",
     "https://biggus.readthedocs.io/",
     "https://stickler-ci.com/",
+    "https://twitter.com/scitools_iris",
 ]
 
 # list of sources to exclude from the build.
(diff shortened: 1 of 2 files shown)


pp-mo:
Hang on, now failed elsewhere, on the twitter link.
But I think that may be a one-off
Trying a re-spin ...

bjlittle:
> Hang on, now failed elsewhere, on the twitter link. But I think that may be a one-off Trying a re-spin ...

@pp-mo I've been respinning various `iris` PR with this failure ... I'm tempted to suggest that you add the twitter link to the `conf.py` to ignore as part of the linkcheck

bjlittle (changes requested):
Add the twitter link to the list of exceptions in the `docs/src/conf.py` as it's giving us serious issues atm

bjlittle:
@pp-mo Have you rebased from `main` after #5621 landed?

pp-mo:
> @pp-mo Have you rebased from `main` after #5621 landed?

No, this PR is not independent as it targets auto-update-lockfiles

bjlittle (approved):

pp-mo:
Ok these 2 changes do seems to work around the problem.
@bjlittle is it good to merge this + #5619 
or, given your comments above about "rebased from main after https://github.com/SciTools/iris/pull/5621 landed", is it better to launch a fresh PR, based on a more recent main ?

Issue #4721, opened 2022-04-28
numpydoc adoption tick list


tkknight:
## 📚 Documentation

**GOAL**: Improve the docstring content and ensure numpydocs formatting.

Related to https://github.com/SciTools/iris/issues/5637.

Below is a list of all the python files in ``iris/lib``.  In order to make ground ensuring our code based support numpdoc fully the below checklist can be used to track our progress.

Recommend we slowly work thru this list using multiple pull requests, all may contribute!

For each file we should:
* Apply numpdoc string formatting.  See guidance in our documentation: https://scitools-iris.readthedocs.io/en/latest/developers_guide/documenting/docstrings.html#docstrings
* Ensure it renders ok using the numpdoc formatting when [Building the Documentation](https://scitools-iris.readthedocs.io/en/latest/developers_guide/contributing_documentation_full.html#building)
* Consider improving the docstring content, not just the formatting.
* Clarify laziness status (if appropriate).  See https://github.com/SciTools/iris/issues/3292

## How to Test usinf ruff (preferred)

Install **ruff** into your iris-dev conda environment via:

- ``pip install ruff``

Temporarily enable the ruff checker by commenting out this line: https://github.com/SciTools/iris/blob/main/.ruff.toml#L30  Once all the repo the source is compliant this ignore rule can be removed.

You can then run the checker via (running against `time.py` in this example):

- ``ruff time.py``

## How to Test (not preferred)

Unless we find something better we can use [pydocstyle](http://www.pydocstyle.org/en/stable/usage.html#command-line-interface).  Install into your iris-dev conda environment via:

- ``pip install pydocstyle``

You can then run the docstring checker via (running against `time.py` in this example):

-  ``pydocstyle --convention=numpy --add-ignore=D105 time.py``

We can use the [numpy convention](http://www.pydocstyle.org/en/stable/error_codes.html#default-conventions) otherwise there maybe conflicts in the errors codes (fix one and and the other triggers and vice versa).

Proposed additional [error codes](https://pep257.readthedocs.io/en/latest/error_codes.html#grouping) to ignore:

- `D105` -  Missing docstring in magic method

## Hit List

Commands used to create the list: ``cd lib/iris; find . -name "*.py" -print | grep -v "^./tests"``.  Feel free to remove entries is they are not relevant.

- [x] ./__init__.py (@tkknight #5635)
- [x] ./_concatenate.py (@tkknight #5715)
- [x] ./_constraints.py (@tkknight #5635)
[...]

PR #5630, opened 2023-12-12
ruff checks for numpydocs

tkknight:
### Description
Ruff compliant.

Part of https://github.com/SciTools/iris/issues/4721

---


bjlittle (approved):
@tkknight Great thanks.

So this is an interim no-op change. You've added `Dxxx` rules to the `.ruff.toml` but will only be enabled by removing the `pydocstyle` category rule `"D"` ... plus you're working on other pydocstyle rules to address.

I'm totally okay with this :+1:

Issue #5541, opened 2023-10-11
How to calculate seasonal temporal correlations?


trexfeathers:
### Discussed in https://github.com/SciTools/iris/discussions/5520

<div type='discussions-op-text'>

<sup>Originally posted by **krikru** September 27, 2023</sup>
If I have two iris cubes, A and B, both on the same grid and with the same time points, how can I calculate the intra-seasonal (according the seasons that are added when calling [add_season](https://scitools-iris.readthedocs.io/en/v3.0.0/generated/api/iris/coord_categorisation.html#iris.coord_categorisation.add_season)) Pearson correlations between A and B for each grid point? Can I do this somehow by only using Iris functions (i.e., not extracting the [data](https://scitools-iris.readthedocs.io/en/stable/generated/api/iris.cube.html#iris.cube.Cube.data) property from both cubes and using NumPy), and without using the [definition](https://en.wikipedia.org/wiki/Correlation#Pearson's_product-moment_coefficient) of the Pearson correlation?

Pearson correlation can usually be calculated with [pearsonr](https://scitools-iris.readthedocs.io/en/latest/generated/api/iris.analysis.stats.html#iris.analysis.stats.pearsonr), but this function doesn't seem to enable [aggregation by an auxiliary variable](https://scitools-iris.readthedocs.io/en/stable/generated/api/iris.cube.html#iris.cube.Cube.aggregated_by), and I need to aggregate by season.</div>

We might want to include a version of @krikru's code within `iris.analysis`. Please vote if you'd like this feature!

trexfeathers:
Highly voted, so we should keep this open, even if we are slow to get to it

PR #1714, opened 2015-06-26
Re-write pearsonr to allow broadcasting and other aggregator keyword arguments

rcomer:
Enables specification of `weights` keyword in `iris.analysis.stats.pearsonr`, to calculate a weighted correlation.  I want to use it for spatial correlations.

In the unit tests, I changed the "perfect correlation" tests from `assertArrayEqual` to `assertArrayAlmostEqual` because I only got 1 to seven decimal places.

PR #5638, opened 2023-12-14
Re-write `pearsonr` to use `Resolve`

rcomer:
#### History ####
`pearsonr` was first added to Iris at #555.  To perform the calculation, the cubes' data were pulled out as numpy arrays and reshaped/transposed within the `_get_calc_view` helper so the dimension to be collapsed was always the leading dimension.  This ensured that the collapsed arrays would broadcast to the original, which is required by the calculation.

At #1714 we re-wrote `pearsonr` to use `cube.collapsed` and cube arithmetic.  This simplified the implementation because those cube operations took care of getting the various dimensions lined up correctly.  It also gave us broadcasting for free (previously the input cubes had to be the same shape) as well as the _weights_ and _mdi_ keywords which were handled by the aggregators.  As the aggregators matured we also got lazy evaluation for free in most cases.  However, it did slow `pearsonr` down a bit.

#### Present Change ####

We can now have the best of both worlds: `Resolve` figures out the dimension handling up front and gives us cubes whose arrays are guaranteed to broadcast to each other.  The _keepdims_ keyword for various `numpy` functions means we can collapse any dimension of an array and the result will still broadcast to the original shape.  This PR takes advantage of `Resolve` and _keepdims_ to move back to doing most of the calculations directly on arrays.  Advantages:
* Cleaner code I think
* Lazy data is now preserved in all cases.  Previously it was not if _common_mask_ was `True`, and I couldn't see a neat way to address that with the existing implementation.
* Could be a step towards solving #5541: if we were to make some sort of aggregator to work on two cubes, we would need an underlying functions that works on arrays.
* It's a bit faster.

<details><summary>Speed comparison</summary>

This script adapted from one of the tests
```python

import timeit
import warnings

import dask.array as da
import iris
import iris.analysis.stats as istats
import iris.tests


def setup(lazy=True):
    cube_temp = iris.load_cube(
        iris.tests.get_data_path(
            ("NetCDF", "global", "xyt", "SMALL_total_column_co2.nc")
        )
    )

    if not lazy:
        cube_temp.data

    cube_a = cube_temp[:6]
    cube_b = cube_temp[20:26]
    cube_b.replace_coord(cube_a.coord("time"))
    for name in ["latitude", "longitude"]:
        cube_b.coord(name).guess_bounds()
    weights = iris.analysis.cartography.area_weights(cube_b)

    return cube_a, cube_b, weights


def correlate(a, b, w):
    r = istats.pearsonr(a, b,
            ["latitude", "longitude"],
            weights=w)
    r.data


def timings(lazy):
    timed = timeit.timeit(f"correlate(*args)", number=10,
                          setup=f"from __main__ import setup, correlate;args = setup({lazy})")
    print(f"{lazy=} {timed:.3f}")


if __name__ == '__main__':
    warnings.simplefilter("ignore")
    for lazy in [False, True]:
        timings(lazy)
```
With `main`:
```
lazy=False 0.383
lazy=True 0.830
```
With my branch:
```
lazy=False 0.221
lazy=True 0.682
```

</details>

#### Testing ####

I modified the existing tests so that the input cubes are always lazy, and then subclassed to make non-lazy versions of all of them.  Not sure if this is overkill?

#### Numpy min pin ####

Note that the _keepdims_ keyword was only added to [numpy.ma.average](https://numpy.org/doc/stable/reference/generated/numpy.ma.average.html) at v1.23, so this PR moves the minimum pin.  [NEP29](https://numpy.org/neps/nep-0029-deprecation_policy.html) says we should support v1.22 until 1st January, but I am assuming the chances of this PR making it into a release before New Year are low.

`dask.ma.average` got the _keepdims_ keyword at [v2022-08-0](https://docs.dask.org/en/stable/changelog.html#v2022-8-0).  Our dask min pin is already higher than this.

#### numpydoc ####

While I was in there, I also updated the docstrings to numpydoc style.  I Have verified with `ruff lib/iris/analysis/stats.py`.

---


rcomer:
In draft till I add missing test highlighted by Codecov.

rcomer:
@tkknight I'm happy to pull the numpydoc change into its own PR if that helps those efforts.  Let me know.

rcomer:
A couple more updates because it's generally better to not break your own use cases 👀 😅

docs/src/whatsnew/latest.rst
```diff
@@ -96,6 +96,8 @@ This document explains the changes made to Iris for this release
    lazy data from file. This will also speed up coordinate comparison.
    (:pull:`5610`)
 
+#. `@rcomer`_ modified :func:`~iris.analysis.stats.pearsonr` so it preserves
+   lazy data in all cases and also runs a little faster.  (:pull:`5638`)
```
trexfeathers, line 100: Could we have a benchmark to demonstrate this?

trexfeathers (changes requested):
I tried to review this but I ran out of time.

**I have finished reviewing every module except `stats.py`.** Either myself or someone else can pick this back up in the future.

lib/iris/tests/unit/analysis/stats/test_pearsonr.py
```diff
@@ -126,36 +145,29 @@ def test_non_existent_coord(self):
...
+
         cube_small_2d = self.cube_a[:, 0:2, 0]
-        cube_small.data = ma.array(
-            cube_small.data, mask=np.array([0, 0, 0, 0, 0, 1], dtype=bool)
-        )
-        cube_small_2d.data = ma.array(
-            np.tile(cube_small.data[:, np.newaxis], 2),
-            mask=np.zeros((6, 2), dtype=bool),
+        cube_small_2d.data = iris.util._mask_array(
+            cube_small.core_data().reshape(6, 1), mask_2d
```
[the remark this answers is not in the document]
rcomer: Past Me wrote this a while ago and Current Me agrees with you.  Apologies!

lib/iris/tests/unit/analysis/stats/test_pearsonr.py
```diff
@@ -169,5 +181,27 @@ def test_common_mask_broadcast(self):
...
+
+class TestCoordHandling(Mixin, tests.IrisTest):
+    def test_lenient_handling(self):
+        # Smoke test that mismatched var_name does not prevent operation.
+        self.cube_a.coord("time").var_name = "wibble"
+        stats.pearsonr(self.cube_a, self.cube_b)
+
+    def test_incompatible_cubes(self):
+        with self.assertRaises(ValueError):
+            stats.pearsonr(self.cube_a[:, 0, :], self.cube_b[0, :, :], "longitude")
```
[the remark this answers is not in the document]
rcomer: Oversight.  Now fixed.

docs/src/whatsnew/latest.rst
rcomer: I would probably need some help with that as my last attempt to run the benchmarks locally failed.  See you at next surgery?

lib/iris/tests/unit/analysis/stats/test_pearsonr.py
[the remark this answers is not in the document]
rcomer: Hopefully a bit clearer now.

lib/iris/analysis/stats.py
```diff
@@ -79,90 +82,93 @@ def pearsonr(
...
-        weights_1 = broadcast_to_shape(weights, cube_1.shape, dims_1_common)
-        if cube_2.shape != smaller_shape:
-            dims_2_common = [
-                i for i in range(cube_2.ndim) if dim_coords_2[i] in common_dim_coords
-            ]
-            weights_2 = broadcast_to_shape(weights, cube_2.shape, dims_2_common)
-        else:
-            weights_2 = weights
+        weights_2 = np.broadcast_to(weights, array_2.shape)
+        weights_1 = np.broadcast_to(weights, array_1.shape)
```
trexfeathers, line 147: Is there any merit in:

```python
weights_cube = Cube(weights)
weights_resolver = Resolve(cube_1, weights_cube)
weights_1 = weights_resolver.lhs_cube_resolved
weights_2 = weights_resolver.rhs_cube_resolved
```

In my head this would avoid needing to get `pearsonr()` to track the logic of `Resolve()`, plus make the code simpler.

But I could be missing an important reason against this.

lib/iris/analysis/stats.py
```diff
@@ -4,13 +4,16 @@
...
-import numpy.ma as ma
 
 import iris
-from iris.util import broadcast_to_shape
+from iris.common import SERVICES, Resolve
+from iris.common.lenient import _lenient_client
+from iris.util import _mask_array
 
 
+@_lenient_client(services=SERVICES)
```
trexfeathers, line 16: You'll note that leniency currently only has one option - `"maths"` - which `Resolve` responds to, e.g.:

https://github.com/SciTools/iris/blob/b3b9ddeeaffb7372481815599c6a1469b7694307/lib/iris/common/resolve.py#L1764

Do you consider `pearsonr()` to fall under the `"maths"` category? Or would the API make more sense if it came under a separate category.

Either way, something should be communicated in the docstring so that the user knows they have this level of control. Perhaps some links to [Lenient Cube Maths](https://scitools-iris.readthedocs.io/en/latest/further_topics/lenient_maths.html) or similar.

```
:ref:`lenient maths`
```

trexfeathers (changes requested):
OK, I don't mind admitting that I found that super difficult. Most of `iris.analysis` remains over my head, even after 5 years.

I can follow what the new code is doing, and have come up with a few suggestions. I can't speak to the equivalence with the old code, but I'll rely on the tests to confirm that.

The more stuff we can get relying on `Resolve`, the better, so well done for making the effort. I'm sure @bjlittle will be delighted!

lib/iris/analysis/stats.py
rcomer: It would need to be

```python
weights_cube = smaller_cube.copy(weights)
...etc...
```
so that `Resolve` has all the metadata it needs to figure out the mappings.  The downside is that we'd be going through all `Resolve`'s logic twice, which could be inefficient.  Maybe something to look at once we have the benchmark in place.

lib/iris/analysis/stats.py
```diff
@@ -79,90 +82,93 @@ def pearsonr(
...
+            index = tuple(
+                slice(0, 1)
+                if np.array_equal(mask_1.any(axis=dim), mask_1.all(axis=dim))
+                else slice(None)
+                for dim in range(mask_1.ndim)
+            )
+            mask_1 = mask_1[index]
+
+        array_2 = _mask_array(array_2, mask_1)
+        array_1 = _mask_array(array_1, al.ma.getmaskarray(array_2))
```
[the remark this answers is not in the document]
rcomer: I would value @bjlittle's opinion on that: my understanding of `Resolve` is that it's currently all about metadata, so asking it to also modify data values (in the form of the masks) would be an expansion of scope.  Note this sometimes also involves changing the array type (numpy array vs masked array vs dask array).

lib/iris/analysis/stats.py
rcomer: The only other place I know of that uses `Resolve` is `mask_cube`, where we just stated that it follows [the same broadcasting approach as cube arithmetic](https://scitools-iris.readthedocs.io/en/stable/_modules/iris/util.html#mask_cube).  I would say that `pearsonr` is more mathsy than `mask_cube` is so we should maybe just add a similar note to the part of the docstring that states the two cubes must be broadcastable to each other.

On a practical note, I don't really know what would be involved in creating a separate category, or what the implications would be.  I also suspect that the vast majority of users don't think about it and leave all the leniency settings to default.

lib/iris/analysis/stats.py
rcomer: Or maybe `Resolve`'s logic that does the transpose and broadcasting on the array could be exposed as a method? 🤔

lib/iris/analysis/stats.py
trexfeathers: I've looked around and no other operations mention that they respond to the leniency framework. That's either deliberate, or if not then we would need to run through the API as a whole. Either way it's not a responsibility of this PR.

trexfeathers: >we should maybe just add a similar note to the part of the docstring that states the two cubes must be broadcastable to each other.

Sounds good 👍

lib/iris/analysis/stats.py
rcomer: On closer inspection that would be rather complicated, because `_as_compatible_cubes`, which does the transpose and broadcast, is [sometimes called a second time](https://github.com/SciTools/iris/blob/1e2a85ae14dc4552dae860d6079d157b39d1a9ca/lib/iris/common/resolve.py#L1210-L1225).  I think that also means my existing method is unsafe, so I have gone for a version of your suggestion.  It does not seem to make a difference to my shiny new benchmark:


```
Benchmarks that have stayed the same:

| Change   | Before [045ce020] <pearsonr-resolve~1>   | After [83d2c832] <pearsonr-resolve>   | Ratio   | Benchmark (Parameter)                |
|----------|------------------------------------------|---------------------------------------|---------|--------------------------------------|
|          | 615±300μs                                | 606±100μs                             | 0.99    | import_iris.Iris.time_analysis_stats |
|          | 76.6±10ms                                | 73.1±8ms                              | 0.95    | stats.PearsonR.time_lazy             |
|          | 37.1±6ms                                 | 30.3±5ms                              | ~0.82   | stats.PearsonR.time_real             |
```

lib/iris/analysis/stats.py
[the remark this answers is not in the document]
rcomer: Actually I think this could be a bit gnarly: this procedure can increase the dimensionality of the arrays.  That's pretty straightforward here because they are _just_ arrays.  Since `Resolve` returns cubes, we would have to sort out how that increased dimensionality affects allocation of coordinates.  I'm not sure I want to go that far down the `Resolve` rabbit hole... 🐇

docs/src/whatsnew/latest.rst
rcomer: First attempt now added.

lib/iris/tests/unit/analysis/stats/test_pearsonr.py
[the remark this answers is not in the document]
rcomer: I wanted to do this by verifying that dask's `compute` doesn't get called.  That turned out to be more complicated than I realised because you can call it from different parts of dask.  In particular

`np.asanyarray(lazy_array)` calls `lazy_array.compute()` which calls `dask.base.compute()`

`cube.data` [ultimately calls](https://github.com/SciTools/iris/blob/fa074429ac16456f3a01de72fff1f6744d599347/lib/iris/_lazy_data.py#L310) `dask.array.compute()`

I couldn't figure out a neat way to cover the different versions using `mock.patch` so used pytest's monkeypatch, but that meant converting all the tests over to pytest.

lib/iris/analysis/stats.py
```diff
@@ -79,90 +82,93 @@ def pearsonr(
 
     smaller_shape = cube_2.shape
 
-    dim_coords_1 = [coord.name() for coord in cube_1.dim_coords]
-    dim_coords_2 = [coord.name() for coord in cube_2.dim_coords]
-    common_dim_coords = list(set(dim_coords_1) & set(dim_coords_2))
+    # Get the broadcast, auto-transposed safe versions of the cube operands.
+    resolver = Resolve(cube_1, cube_2)
+    cube_1 = resolver.lhs_cube_resolved
+    cube_2 = resolver.rhs_cube_resolved
```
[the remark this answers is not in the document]
rcomer: Done, and also renamed other variables to lhs/rhs for consistency.

lib/iris/analysis/stats.py
rcomer: Done

lib/iris/tests/unit/analysis/stats/test_pearsonr.py
```diff
@@ -35,18 +40,34 @@ def setUp(self):
...
+        m_compute = mock.Mock(wraps=dask.base.compute)
+
+        # The three dask compute functions are all the same function but monkeypatch
+        # does not automatically know that.
+        # https://stackoverflow.com/questions/77820437
+        monkeypatch.setattr(dask.base, dask.base.compute.__name__, m_compute)
+        monkeypatch.setattr(dask, dask.compute.__name__, m_compute)
+        monkeypatch.setattr(dask.array, dask.array.compute.__name__, m_compute)
+
+        return m_compute
```
rcomer, line 55: Possibly we might want this in a `conftest.py` somewhere so it's available to other tests?

rcomer (commented):
Thanks @trexfeathers for your thorough review.  I think I have now responded to all the comments one way or another.

>  I don't mind admitting that I found that super difficult.

Full disclosure: I originally started trying to do this just to see if I could!  So yeah, it was a challenge for me too.

lib/iris/tests/unit/analysis/stats/test_pearsonr.py
rcomer: Like maybe this one 👀 https://github.com/SciTools/iris/blob/main/lib/iris/tests/unit/conftest.py

trexfeathers: ping @ESadek-MO, who is planning some team discussions on this sort of thing soon.

lib/iris/tests/unit/analysis/stats/test_pearsonr.py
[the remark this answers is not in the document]
trexfeathers: Some detailed work! I personally would have just ended with an assertion that the `Cube`(s) were still lazy, but of course your approach will catch things that I would not have.

lib/iris/analysis/stats.py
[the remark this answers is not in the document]
trexfeathers: Conversation offline:

We confirmed that `Resolve` DOES increase the dimensionality of `lhs_` or `rhs_cube_resolved`.

However it is still true that data management is not currently in the scope of  `Resolve`. `lhs_` and `rhs_cube_resolved` in particular are not expected to be used directly in many cases, so there is no 'contract' for getting the data right in these. What is more: The mask handling of `pearsonr()` is specialised - [see the `common_mask` parameter](https://github.com/SciTools/iris/blob/e16ea932a48a581dd7b383e388efe917341bb09c/lib/iris/analysis/stats.py#L51-L55), making it even less likely that the logic in this function would be reused elsewhere.

No further action

trexfeathers (approved):

rcomer:
Thanks @trexfeathers!

Issue #5611, opened 2023-12-06
Lazy `iris.analysis.cartography.area_weights`


schlunma:
## ✨ Feature Request

Currently, `iris.analysis.cartography.area_weights` always returns a numpy array. Depending on the shape of the input cube, this can use up a lot of memory. It would be really helpful to have a lazy version of it that returns a dask array.

## Motivation

Using a dask distributed scheduler in ESMValTool in combination with a preprocessor that requires the calculation of area weights is currently not possible. A lazy version of `iris.analysis.cartography.area_weights` would solve that.

## Design
Solving this is probably easy since `broadcast_to_shape` that is [used to broadcast the 2D weights to the cube shape](https://github.com/SciTools/iris/blob/db43fde0534af0ee32c77f7747ab506da5552c1d/lib/iris/analysis/cartography.py#L482-L484) now supports dask arrays. I see two options here regarding the API:

1. Add an additional keyword argument for `iris.analysis.cartography.area_weights` (maybe `compute` which is `True` by default?).
2. Let the behavior depend on the input data, i.e., if the cube has lazy data, the function returns lazy data and vice versa.

I guess option 1. is preferable since it is fully backwards-compatible.

If we can agree on an implementation, I can open a PR.

HGWright:
@SciTools/peloton

@schlunma thanks for raising this and offering to put up a PR. We think option 1 is much preferred to option 2.

PR #5658, opened 2023-12-19
Implement lazy area weights

schlunma:
### Description

This PR makes `iris.analysis.cartography.area_weights` lazy by providing the keyword arguments `compute` and `chunks`:

```py
def area_weights(cube, normalize=False, compute=True, chunks=None):
```

These defaults ensure full backwards-compatibility. ~`chunks=None` results in using the cube data's chunks for the weights array, which is a resonable default since these weights will probably be used with the cube's data anyway.~

~This requires https://github.com/SciTools/iris/pull/5620, so still in draft mode. Apart from that, this PR is ready.~

Closes https://github.com/SciTools/iris/issues/5611

---


trexfeathers:
Thanks for the hard work @schlunma. #5620 will be reviewed by @pp-mo in the new year, so #5658 can get moving again once that is sorted.

lib/iris/tests/test_analysis.py
```diff
@@ -1155,6 +1155,55 @@ def test_area_weights(self):
...
+        # Check that chunksizes are as expected
+        if chunks is not None:
+            assert area_weights.chunksize == chunks
+        else:
+            if isinstance(cube_data, da.Array):
+                assert area_weights.chunksize == cube_data.chunksize
+            else:
+                assert area_weights.chunksize == (4, 3, 4)
+
+        # Check that actual weights are as expected
```
trexfeathers, line 1190: Please could you note that these are 'known good output'? I have previously panicked when dependencies have changed test outputs and I had no idea why the original numbers were there!

lib/iris/tests/test_analysis.py
```diff
@@ -1155,6 +1155,55 @@ def test_area_weights(self):
         )
 
 
+@tests.skip_data
+class TestLazyAreaWeights:
```
trexfeathers, line 1159: I'm happy with the coverage from this 👍

trexfeathers (changes requested):
This looks great, thanks @schlunma! Just a tiny request please

lib/iris/tests/test_analysis.py
schlunma: Done in https://github.com/SciTools/iris/pull/5658/commits/8a03706d678de744f616a94fd6a0f87b048c97c1, is this okay like this or do I need to add more?

schlunma:
Just had a second look on this. I think it's actually better to use `cube.lazy_data().chunks` instead of `cube.lazy_data().chunksize` to make sure that the chunk structure is exactly identical in the weights and the original data. It probably won't matter, but it feels better to use `.chunks` here...

schlunma:
After having a third look on it, I think it's better to leave full control of the chunks to the user (i.e., if `chunks=None`, also use `chunks=None` in the `broadcast_to_shape` call instead of the cube's chunks). Otherwise, there is no way a user can specify to use "no special chunking", which might be confusing.

Sorry for all the back and forth, with this I am happy with the PR!

trexfeathers (approved):
OK these changes make equal sense to me

PR #4661, opened 2022-03-24
Implement Pydata theme

tkknight:
### Description

Adoption of the pydata theme (https://pydata-sphinx-theme.readthedocs.io/en/latest/) for Iris.  This has meant some slight restructuring of the rst files to allow for the navigation.

Changes of note:

* Added icons to each section on [homepage ](https://tkknight-iris-test-doc.readthedocs.io/en/latest/index.html)(tired to use all svg but some are png)
* Theme allows for icons in the top right corner
* Added a panel (bottom right one) to goto the Issue Voting.
* [Support](https://tkknight-iris-test-doc.readthedocs.io/en/latest/index.html#support) section on index highlights the legacy channels
* Added a discussions directive so we can use `:discussion:'9999'` and tidied up some discussions references.  Used on the [whatsnew entry](https://tkknight-iris-test-doc.readthedocs.io/en/latest/whatsnew/latest.html#documentation) for this PR.
* Added a [Why Iris](https://tkknight-iris-test-doc.readthedocs.io/en/latest/why_iris.html) page.  
* [Data Model](https://tkknight-iris-test-doc.readthedocs.io/en/latest/further_topics/ugrid/data_model.html#structured-grids-the-old-world) figures are now left aligned.  The new theme used to right align the image **and** the text, it looked weird.
* fixed many bullet point alignment issues
* Used badges to highlight the version of the docs being used (with colour).  See the image for all the possible examples (only one will be displayed of course in the docs)

Screenshot 2022-03-24 142704

Changes that maybe contentious:

* [Gallery ](https://tkknight-iris-test-doc.readthedocs.io/en/latest/generated/gallery/index.html) is not directly accessible.  It is under the [Getting Started](https://tkknight-iris-test-doc.readthedocs.io/en/latest/getting_started.html) top link or panel button.
* [User Guide](https://tkknight-iris-test-doc.readthedocs.io/en/latest/userguide/index.html) menu (left side) has a `FURTHER TOPICS`  and the sections included.  If it was a separate file it would not be expanded in the menu so users would never find it unless they go looking.
* I removed `dev.rst`.  As this was a symlink with [latest.rst](https://tkknight-iris-test-doc.readthedocs.io/en/latest/whatsnew/latest.html). the sphinx build showed a warning.  If I added links to both the sidebar menu would not render as it is not explicitly reference on a toc (only the symlink is).   This may mean the release instructions will need tweaked - happy to have some input on this.

[...]

PR #5684, opened 2024-01-09
Make further topics obviously different

trexfeathers:
### Description

A very simple way to make it obvious in a TOC-tree that `Further Topics` is not just a page, but a subsection of its own.

I chose to edit the title itself, rather than an alias in the existing TOC-tree, so that this will also apply if included in other TOC-trees. I also kinda like the title being this way 😁

---

### DIFF
--- a/docs/src/conf.py
+++ b/docs/src/conf.py
@@ -291,7 +291,6 @@ html_sidebars = {
 html_theme_options = {
     "footer_start": ["copyright", "sphinx-version"],
     "footer_end": ["custom_footer"],
-    "collapse_navigation": True,
     "navigation_depth": 3,
     "show_prev_next": True,
     "navbar_align": "content",
--- a/docs/src/further_topics/index.rst
+++ b/docs/src/further_topics/index.rst
@@ -1,8 +1,8 @@
 .. _further_topics_index:
 
 
-➕ Further Topics
-=================
+Further Topics
+===============
 
 Extra information on specific technical issues.
 
--- a/docs/src/further_topics/index.rst
+++ b/docs/src/further_topics/index.rst
@@ -1,8 +1,8 @@
 .. _further_topics_index:
 
 
-Further Topics
-===============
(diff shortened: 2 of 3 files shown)


ESadek-MO (commented):
I wonder if we should apply this to all nested toc-members? I found the below all expand out when clicked:
- Dask Best Practices
- Mesh Support
- Gallery

I think we should maintain consistency throughout the docs, which'd also mean we'd have to remember the format if creating new nests later down the line (which is a good habit, just thought it was worth noting).

It's a good idea though, and a simple change!

tkknight:
There maybe an option in the pydata theme to do this.  I will check.

trexfeathers:
> There maybe an option in the pydata theme to do this. I will check.

There is! I'm just checking the blame for why we turned it off originally...

trexfeathers:
> > There maybe an option in the pydata theme to do this. I will check.
> 
> There is! I'm just checking the blame for why we turned it off originally...

@tkknight looks like this line was introduced when we introduced PyData Theme (#4661). Would you be comfortable with turning it off?

https://github.com/SciTools/iris/blob/2bfaebda62b22d3203cfa737a5f017d68800a5c8/docs/src/conf.py#L294

[ref](https://pydata-sphinx-theme.readthedocs.io/en/stable/user_guide/navigation.html#remove-reveal-buttons-for-sidebar-items)

tkknight:
Try removing this line and rebuild the docs. 

https://github.com/SciTools/iris/blob/main/docs/src/conf.py#L294

trexfeathers:
I like it

image

ESadek-MO (changes requested):
I really like this new version. Adds a lot of functionality with a single line of code imo, and doesn't seem to break anything. 

Just need to add a whatsnew then I'm happy for this to go!

ESadek-MO (approved):
Super happy with this, thanks @trexfeathers!

PR #5685, opened 2024-01-09
Removed use of catch_warnings

### LINKED ISSUE #5536: Remove problematic use of catch_warnings
## 📰 Custom Issue

Part of the investigation in #5499 found that warning duplication was being caused during loading by these lines of code:
https://github.com/SciTools/iris/blob/b6e39d758fc17e167b9323b25484965453846dbf/lib/iris/fileformats/_nc_load_rules/helpers.py#L417-L431

This code should be refactored to not use `catch_warnings` and it should be demonstrated that this reduces warning duplication.

ESadek-MO:
Closes #5536.

Removed catch_warnings, by moving the functionality to parse_cell_methods()


stephenworsley:
## Before:
image

## After:
image

Compare with https://github.com/SciTools/iris/pull/5506#issuecomment-1729102179

stephenworsley (changes requested):
Looks good, I think this could do with a test to explicitly check that this avoids warning repeats similar to this one (feel free to copy this test) https://github.com/SciTools/iris/pull/5506/files#diff-517f51f31e7b2402608486de8f2f1a79f004d9591eb17869451b9f27a61518b3R495

lib/iris/tests/integration/netcdf/test_general.py
```diff
@@ -484,5 +484,43 @@ def test_path_string_save_same(self):
         self.assertCDL(tempfile_frompath)
 
 
+@tests.skip_data
+class TestWarningRepeats(tests.IrisTest):
+    def test_datum_once(self):
+        fnames = [
+            "false_east_north_merc.nc",
+            "non_unit_scale_factor_merc.nc",
+            "toa_brightness_temperature.nc",
```
stephenworsley, line 493: So I've run some tests and I think this file would actually cause a warning repeat if there were files after it, for the sake of making it clear which files are expected not to cause repeats, I think this should be commented out for the time being with a comment noting that this file still causes warning repeats.

lib/iris/tests/integration/netcdf/test_general.py
```diff
@@ -484,5 +484,43 @@ def test_path_string_save_same(self):
...
+        ]
+
+        with warnings.catch_warnings():
+            warnings.simplefilter("default")
+            with warnings.catch_warnings(record=True) as record:
+                for fpath in fpaths:
+                    iris.load(fpath)
+        assert len(record) == 1
+
+    def test_datum_default(self):
```
stephenworsley, line 506: Upon reflection, I think this test is less relevant to this PR and given that its behaviour might change in future it would be better to be removed.

stephenworsley (changes requested):
Looking good, just a couple thoughts about how the tests might be made more robust to future changes to warning behaviour.

lib/iris/tests/integration/netcdf/test_general.py
```diff
@@ -487,39 +487,35 @@ def test_path_string_save_same(self):
...
-                    iris.load(fpath)
-        assert len(record) == 1
+        Notes
+        -----
+        This test relies on `iris.load` throwing a warning. This warning might
+        be removed in the future, in which case `assert len(record) == 2 should`
+        be change to `assert len(record) == 1`.
 
-    def test_datum_default(self):
+        toa_brightness_temperature.nc has lazy data, and triggers a
```
stephenworsley, line 498: To be precise, it has an AuxCoord with lazy data.

lib/iris/tests/integration/netcdf/test_general.py
```diff
@@ -487,39 +487,35 @@ def test_path_string_save_same(self):
...
+        specific part of dask which contains a `catch_warnings()` call,
+        so has been removed from the `fnames` list.
```
stephenworsley, line 500: I think it's elaborating "... a `catch_warnings()` call _which causes warnings to be repeated,_ so has been removed from the `fnames` list _until a solution is found for such a file._"

lib/iris/tests/integration/netcdf/test_general.py
```diff
@@ -487,39 +487,35 @@ def test_path_string_save_same(self):
...
+
+        """
+        #
         fnames = [
             "false_east_north_merc.nc",
             "non_unit_scale_factor_merc.nc",
-            "toa_brightness_temperature.nc",
+            #
```
stephenworsley, line 507: The file name should still be here, but commented out.

stephenworsley (changes requested):
Just a couple tweaks to the comments here

lib/iris/tests/integration/netcdf/test_general.py
```diff
@@ -495,16 +495,17 @@ def test_datum_once(self):
         be removed in the future, in which case `assert len(record) == 2 should`
         be change to `assert len(record) == 1`.
 
-        toa_brightness_temperature.nc has lazy data, and triggers a
-        specific part of dask which contains a `catch_warnings()` call,
-        so has been removed from the `fnames` list.
+        toa_brightness_temperature.nc has an AucCoord lazy data, and triggers a
```
stephenworsley, line 498: *Au _x_ Coord _with_ lazy data

ESadek-MO: Embarrassing, my apologies

lib/iris/fileformats/_nc_load_rules/helpers.py
```diff
@@ -356,10 +357,15 @@ def parse_cell_methods(nc_cell_methods):
...
-                )
+                msg = msg.format(method_words[0])
+                if cf_name:
+                    name = "{}".format(cf_name)
+                    msg = msg.replace("variable", "variable {!r}".format(name))
+                else:
+                    warnings.warn(
+                        msg,
+                        category=UnknownCellMethodWarning,
+                    )
```
stephenworsley, line 368: I think it might actually be worth adding a `msg = None` here to ensure this warning isn't repeated at the end of the function.

stephenworsley (approved):
Looks good!

trexfeathers:
Should there have been a What's New entry here?

PR #5691, opened 2024-01-11
Faster trivial equality checks for coordinates and arrays

bouweandela:
### Description

This makes the comparison of coordinates, cubes, and arrays to themselves faster.

---

### DIFF
--- a/lib/iris/cube.py
+++ b/lib/iris/cube.py
@@ -3826,6 +3826,9 @@ class Cube(CFVariableMixin):
 
     # START OPERATOR OVERLOADS
     def __eq__(self, other):
+        if other is self:
+            return True
+
         result = NotImplemented
 
         if isinstance(other, Cube):
--- a/lib/iris/cube.py
+++ b/lib/iris/cube.py
@@ -3826,9 +3826,6 @@ class Cube(CFVariableMixin):
 
     # START OPERATOR OVERLOADS
     def __eq__(self, other):
-        if other is self:
-            return True
-
         result = NotImplemented
 
         if isinstance(other, Cube):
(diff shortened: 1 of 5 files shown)


lib/iris/cube.py
```diff
@@ -3826,6 +3826,9 @@ def _deepcopy(self, memo, data=None):
 
     # START OPERATOR OVERLOADS
     def __eq__(self, other):
+        if other is self:
+            return True
```
stephenworsley, line 3830: I believe it is currently the case that when comparing cubes, unlike when comparing coordinates, the presence of NaNs will cause two otherwise identical cubes to be considdered not equal. See: https://github.com/SciTools/iris/blob/9b7b9430856b610cda193f1d8ae9ed4847dc1dca/lib/iris/cube.py#L3868-L3870

This behaviour could be changed by adding an `equal_nan=True` argument to `da.allclose`, but it looks like this behaviour has been maintained, I suppose, for fear of introducing a breaking change (though I could definitely see an argument for changing it).

With things as they are, I think your change would introduce an inconsistency with how cubes with NaN data behaved under equality. I suppose you could maintain the current behaviour by adding a check whether any of the data in the cube is NaN, e.g.
`if other is self and not da.any(da.isnan(self.core_data)).compute():`

It's probably worth thinking about if we still want this behaviour and if it is even intentional (I can't find any tests which explicitly test for it), but I think that should probably be a seperable issue.

bouweandela: Thanks for reviewing! I overlooked that. I created https://github.com/SciTools/iris/issues/5705 and will take the change out of this pull request.

bouweandela: Done in [920120a](https://github.com/SciTools/iris/pull/5691/commits/920120ac16e33c9998b0e75901d781dc376d6b5d)

stephenworsley (approved):
Looks good to me!

PR #5693, opened 2024-01-11
Make the Coord.cell method lazy

### LINKED ISSUE #5692: Lazy `iris.coords.Coord.cell` method
## ✨ Feature Request

At the moment, the `iris.coords.Coord.cell` method has the side effect of loading all the coordinate data from disk into memory.

## Motivation

It would be more efficient if only the requested coordinate point would be loaded. If users want to load all data, e.g. because they plan to access many cells, they can do so themselves.

bouweandela:
### Description

This avoids loading all the coordinate data from disk if only a single cell is requested.

Closes #5692 

---


ESadek-MO (approved):
Thanks @bouweandela, looks good to me!

Issue #5457, opened 2023-09-01
Auxiliary coordinates created by HybridPressureFactory have huge chunks


bouweandela:
When loading a file that contains an auxiliary coordinate that can be computed using a formula term, the auxiliary coordinate ends up having huge chunks. This leads to memory issues when trying to use such coordinates, as the Dask workers will run out of memory and get killed.

## Example

Open the file [clw_Amon_FGOALS-f3-L_historical_r1i1p1f1_gr_196001-196912.nc](https://esgf.ceda.ac.uk/thredds/fileServer/esg_cmip6/CMIP6/CMIP/CAS/FGOALS-f3-L/historical/r1i1p1f1/Amon/clw/gr/v20190927/clw_Amon_FGOALS-f3-L_historical_r1i1p1f1_gr_196001-196912.nc) and list the chunks of the computed coordinate:

```python
In [1]: import iris

In [2]: iris.__version__
Out[2]: '3.7.0'

In [3]: cube = iris.load_cube("clw_Amon_FGOALS-f3-L_historical_r1i1p1f1_gr_196001-196912.nc", "mass_fraction_of_cloud_liquid_water_in_air")
/home/bandela/mambaforge/envs/iris/lib/python3.11/site-packages/iris/fileformats/cf.py:859: UserWarning: Missing CF-netCDF measure variable 'areacella', referenced by netCDF variable 'clw'
  warnings.warn(
/home/bandela/mambaforge/envs/iris/lib/python3.11/site-packages/iris/fileformats/cf.py:1264: UserWarning: Ignoring formula terms variable 'ps' referenced by data variable 'b_bnds' via variable 'lev': Dimensions ('time', 'lat', 'lon') do not span ('lev', 'bnds')
  warnings.warn(msg)
/home/bandela/mambaforge/envs/iris/lib/python3.11/site-packages/iris/fileformats/cf.py:1264: UserWarning: Ignoring formula terms variable 'ps' referenced by data variable 'a_bnds' via variable 'lev': Dimensions ('time', 'lat', 'lon') do not span ('lev', 'bnds')
  warnings.warn(msg)
/home/bandela/mambaforge/envs/iris/lib/python3.11/site-packages/iris/fileformats/netcdf/loader.py:353: UserWarning: Unable to find coordinate for variable 'ps'
  warnings.warn(
[...]

PR #5712, opened 2024-01-24
Use reasonable chunk sizes in derived coordinates

bouweandela:
### Description

Use the iris optimal chunk size algorithm to set reasonable chunk sizes for derived coordinates.

This is one way to solve #5457.

Could I have some feedback on the approach, please? See also the question in [the issue](https://github.com/SciTools/iris/issues/5457#issuecomment-1907866758).

---

### DIFF
--- a/lib/iris/aux_factory.py
+++ b/lib/iris/aux_factory.py
@@ -11,9 +11,31 @@ import cf_units
 import dask.array as da
 import numpy as np
 
+from iris._lazy_data import _optimum_chunksize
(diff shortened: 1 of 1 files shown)


trexfeathers:
@bouweandela please could we try an alternative solution that changes the chunking of the original coordinates (the ones that make up the derived coordinate)? We know that users could override this using `CHUNK_CONTROL` if they wanted.

Thanks 😊

PR #5713, opened 2024-01-24
Consider NaNs equal when comparing cubes

### LINKED ISSUE #5705: Should a cube be considered equal to itself?
## Motivation

At the moment, a cube where the data contains a NaN is not considered equal to itself or to another cube with the same data. Conversely, coordinates/cell measures/ancillary variables with NaNs are equal to themselves. Maybe it would be nicer, and more consistent, to introduce the same behaviour for cubes.

See https://github.com/SciTools/iris/pull/5691#discussion_r1457455386 for some previous discussion.

bouweandela:
### Description

Consider NaNs equal when comparing cubes and speed up comparing cubes to themselves.

Closes  #5705

---

### DIFF
--- a/docs/src/whatsnew/latest.rst
+++ b/docs/src/whatsnew/latest.rst
@@ -91,8 +91,10 @@ This document explains the changes made to Iris for this release
 💣 Incompatible Changes
 =======================
 
-#. `@bouweandela`_ updated cube comparison so cubes with data containing a
-   :obj:`numpy.nan` are now considered equal (:pull:`5713`)
+#. `@bouweandela`_ updated :class:`~iris.cube.Cube` comparison so equality is
+   now possible between cubes with data containing a :obj:`numpy.nan`.
+   e.g. ``Cube([np.nan, 1.0]) != Cube([np.nan, 2.0])``,
+   ``Cube([np.nan, 1.0]) == Cube([np.nan, 1.0])``. (:pull:`5713`)
 
 
 🚀 Performance Enhancements
--- a/docs/src/whatsnew/latest.rst
+++ b/docs/src/whatsnew/latest.rst
@@ -91,7 +91,8 @@ This document explains the changes made to Iris for this release
 💣 Incompatible Changes
 =======================
 
-#. N/A
+#. `@bouweandela`_ updated cube comparison so cubes with data containing a
+   :obj:`numpy.nan` are now considered equal (:pull:`5713`)
 
 
 🚀 Performance Enhancements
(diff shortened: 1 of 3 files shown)


lib/iris/cube.py
```diff
@@ -3862,7 +3865,11 @@ def __eq__(self, other):
             if result:
                 # TODO: why do we use allclose() here, but strict equality in
                 #  _DimensionalMetadata (via util.array_equal())?
-                result = da.allclose(self.core_data(), other.core_data()).compute()
+                result = da.allclose(
+                    self.core_data(),
+                    other.core_data(),
+                    equal_nan=True,
+                ).compute()
```
trexfeathers, line 3872: I know this wasn't you, but I think we would be better with only using Dask when a lazy array is involved, which will save overhead if everything is already NumPy. `np.allclose()` is intelligent enough to do this, and `as_concrete_data()` knows to only use `compute()` when necessary:

```suggestion
                result = np.allclose(
                    self.core_data(),
                    other.core_data(),
                    equal_nan=True,
                )
                result = _lazy.as_concrete_data(result)
```

docs/src/whatsnew/latest.rst
```diff
@@ -91,7 +91,8 @@ This document explains the changes made to Iris for this release
 💣 Incompatible Changes
 =======================
 
-#. N/A
+#. `@bouweandela`_ updated cube comparison so cubes with data containing a
+   :obj:`numpy.nan` are now considered equal (:pull:`5713`)
```
trexfeathers, line 95:
```suggestion
#. `@bouweandela`_ updated :class:`~iris.cube.Cube` comparison so equality is
   now possible between cubes with data containing a :obj:`numpy.nan`.
   e.g. ``Cube([np.nan, 1.0]) != Cube([np.nan, 2.0])``,
   ``Cube([np.nan, 1.0]) == Cube([np.nan, 1.0])``. (:pull:`5713`)
```

I'm worried about misinterpretation that ALL `Cube`s including `NaN` will be considered equal.

trexfeathers (changes requested):
Thanks very much @bouweandela! Would you be happy with these adjustments?

lib/iris/cube.py
bouweandela: That would change the resulting datatype from a `bool` to a `numpy.ndarray` though. Would the solution in https://github.com/SciTools/iris/pull/5713/commits/ec30de862338e6e6b0dbb2e6032731e36b55cdbd also work for you?

trexfeathers (approved):
This is great, thanks @bouweandela. LGTM

PR #5734, opened 2024-02-07
added in a vertical rule for surface fields

### LINKED ISSUE #3820: Incorrect LBLEV in pp save rules for surface fields.
## 🐛 Bug Report

Iris pp save rules result in a LBLEV of 0 for a surface field (i.e. all the non-zero values are gated behind a check for a vertical coordinate: https://github.com/SciTools/iris/blob/master/lib/iris/fileformats/pp_save_rules.py#L653 ).  F03, however, specifies a value of 9999 for surface fields: https://code.metoffice.gov.uk/doc/um/latest/papers/umdp_F03.pdf#page=27  In Ants, we rely on the pp save rules to set this value as we translate a cube into mule fields for saving ancillaries.  We will put in a workaround in Ants to set this ourselves for an immediate fix, but it would be good if this could be fixed upstream.

## How To Reproduce

1. Use iris to load a surface field (e.g. SST or orography)
2. Save the surface field as pp file.
3. Inspect lookup header with e.g. `mule-pumf --headers-only filename.pp | grep 'lblev'`

## Expected behaviour

LBLEV value of 9999 for surface fields.

## Environment 
 - Iris Version: 2.3

ESadek-MO:
## 🚀 Pull Request

Closes #3820.


lib/iris/fileformats/pp_save_rules.py
```diff
@@ -773,6 +773,21 @@ def _vertical_rules(cube, pp):
...
+        height_coord is None
+        and depth_coord is None
+        and pressure_coord is None
+        and soil_mln_coord is None
+        and apt_coord is None
+        and air_pres_coord is None
+        and level_height_coord is None
+        and mln_coord is None
+        and sigma_coord is None
+    ):
```
pp-mo, line 787: FWIW I reckon the logic of this **_ought_** to be captured in [`iris.util.guess_coord_axis`](https://github.com/SciTools/iris/blob/933e29f120fb7d3bc7bd4813971c41475e0d3fe9/lib/iris/util.py#L231),
so that this test could be just `if not cube.coords(axis='Z'):`.

But unfortunately things like 'height' and 'depth' are _not_ recognised by guess_coord_axis, at least at present.

pp-mo:
Note about the currently failing tests ...

These PP save tests are checked by comparing a string representation of all the resulting PP fields against one stored in a text file under [iris/tests/results/cube_to_pp](https://github.com/SciTools/iris/tree/v3.7.0/lib/iris/tests/results/cube_to_pp).  
This is done with the "cube_save_test" method supplied [here](https://github.com/SciTools/iris/blob/v3.7.0/lib/iris/tests/pp.py#L20).

So, you should just _delete_ all those files, and run all tests which will re-make them.
Then commit the resulting changes, which can then be reviewed.

NOTE: for no very good reason AFAICT, this check is implemented separately from the other compare-to-stored-file tests of this nature, and does NOT require IRIS_TEST_CREATE_MISSING=1 to be set  (which was probably an oversight).

lib/iris/fileformats/pp.py
```diff
@@ -2301,12 +2301,12 @@ def save_pairs_from_cube(cube, field_coords=None):
...
         # Run the PP save rules on the slice2D, to fill the PPField,
         # recording the rules that were used
-        pp_field = verify(slice2D, pp_field)
+        pp_field = verify(slice2D, pp_field, label_surface_fields=label_surface_fields)
 
         yield (slice2D, pp_field)
 
 
-def as_fields(cube, field_coords=None):
+def as_fields(cube, field_coords=None, label_surface_fields=False):
```
pp-mo, line 2309: Needs to add a keyword description in the docstring "Parameters"

lib/iris/fileformats/pp.py
```diff
@@ -2156,7 +2156,7 @@ def _load_cubes_variable_loader(
     return result
 
 
-def save(cube, target, append=False, field_coords=None):
+def save(cube, target, append=False, field_coords=None, label_surface_fields=False):
```
pp-mo, line 2159: Needs to add a keyword description in the docstring "Parameters"

docs/src/whatsnew/latest.rst
```diff
@@ -41,6 +41,9 @@ This document explains the changes made to Iris for this release
 #. `@rcomer`_ enabled partial collapse of multi-dimensional string coordinates,
    fixing :issue:`3653`. (:pull:`5955`)
 
+#. `@ESadek-MO`_ has updated :mod:`iris.fileformats.pp_save_rules` to set
+   `pp.lbelv` of surface fields to 9999. (:issue:`3280`, :pull:`5734`)
+
```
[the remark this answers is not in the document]
pp-mo: I'm doubling down on this : 
  * mention the new `pp.save` keyword + link to the docstring
  * mention that you can also pass it to `iris.save`
  * add a subsection on surface fields to the [vertical coords section of the 'further info' page on um file processing](https://scitools-iris.readthedocs.io/en/v3.10.0/further_topics/um_files_loading.html#vertical-coordinates)
    * .. and ref that in the whatsnew, too

pp-mo (changes requested):
Just some docs improvements needed, I think

stephenworsley (changes requested):
Just a couple changes needed. I think it's worth being a little bit more specific in a few places.

pp-mo (approved):

pp-mo:
From @stephenworsley offline : Says he is happy now, once I agree that my comments are addressed. 
So Good To Go :tm: !

PR #5736, opened 2024-02-07
Update line number references in filtering_warnings.rst .

trexfeathers:
### Description

No What's New entry as this page is still new for Iris 3.8 .

---

### DIFF
--- a/docs/src/further_topics/filtering_warnings.rst
+++ b/docs/src/further_topics/filtering_warnings.rst
@@ -125,18 +125,16 @@ Or you can target Warnings raised by specific lines of specific modules, e.g.
 .. doctest:: filtering_warnings
 
     >>> with warnings.catch_warnings():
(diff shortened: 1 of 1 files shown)


pp-mo:
It seems fine, as far as it goes.
But it's clearly a problem that the line number could simply change again before we cut the release ?

Actually, **_no_** .. it should fail on the doctest if that happens.
Let's test that ....

trexfeathers:
> It seems fine, as far as it goes. But it's clearly a problem that the line number could simply change again before we cut the release ?
> 
> Actually, **_no_** .. it should fail on the doctest if that happens. Let's test that ....

CI seems to demonstrate what we were after - 30121c72306f6681d5e22e52dd9fe5509f983537 raises the expected doctest failure.

pp-mo (approved):
Boom, sorted !

Issue #5517, opened 2023-09-26
Iris bridge to GeoVista


pp-mo:
User interest in geovista seems to be blossoming right now.
We have various resources showing how to link to it from Iris, e.g. [the Mesh plotting section + example](https://scitools-iris.readthedocs.io/en/v3.7.0/further_topics/ugrid/operations.html#plotting).
Plus the [region extraction example](https://scitools-iris.readthedocs.io/en/v3.7.0/further_topics/ugrid/operations.html#region-extraction) And in future we might extend this.

Likewise we have produced various demo notebooks containing similar code content.

But there's a lot of duplicated code + boilerplate making the link between cubes and PyVista polydata.
So there is a really useful opportunity to provide an "iris geovista bridge" module.

However it's not clear where this would live, and where it should get tested.
It should **_not_** add to the the core dependencies, or the C.I. testing costs.
It could be a partner package, like iris-grib or iris-esmf-regrid.
The idea of a plugin architecture might also have legs.

trexfeathers:
>The idea of a plugin architecture might also have legs.

I personally think proliferation of plugin packages is ill fitted to our situation.

The two arguments in favour (that I'm aware of):

- Aids innovation in large projects because the plugins can be managed by non-core developers.
  Not really applicable here.
- Makes it easier to introduce dependencies not required by the core package
  This isn't actually true. Xarray (and others I'm sure) have demonstrated that you can have a very cut-down recipe, with errors/warnings if a user tries to invoke an optional dependency without it being installed.

Arguments against:

- Doubling up on repo management - dependabot, lock-files, dev permissions etcetera
- A user has to understand the structure so they can raise the issue in the right place

trexfeathers:
See #4798

trexfeathers:
Discussed during refinement for Iris 3.8.

## Features

The MVP bridge would constitute two operations as previously noted by @pp-mo:

- Conversion from `Cube` to `PolyData`
- Indexing a `Cube` using `PolyData` indices. Could we wrap the whole operation so the user didn't need to learn GeoVista/PyVista?
- _We do NOT expect a need for conversion from `PolyData` to `Cube`_

## Possible dependency problems

To avoid getting bogged down in concerns about dependencies, we recommend giving it a try and seeing what happens:

- Introduce the bridge features in `iris.experimental` - can back out if we run into trouble
- Include GeoVista as an **optional** Iris dependency
- Acknowledge that this means making the testing / development environment larger
[...]

PR #5752, opened 2024-02-16
Introduce ncdata; more info on Xarray CF handling.

pp-mo:
Addresses #5517

### DIFF
--- a/docs/src/community/iris_xarray.rst
+++ b/docs/src/community/iris_xarray.rst
@@ -38,9 +38,28 @@ There are multiple ways to convert between Iris and Xarray objects.
   feasible to save a NetCDF file using one package then load that file using
   the other package. This will be lossy in places, as both Iris and Xarray
   are opinionated on how certain NetCDF concepts relate to their data models.
-* The Iris development team are exploring an improved 'bridge' between the two
-  packages. Follow the conversation on GitHub: `iris#4994`_. This project is
-  expressly intended to be as lossless as possible.
+* `ncdata <https://github.com/pp-mo/ncdata/blob/main/README.md>`_ is a package which
+  the Iris development team have developed to manage netcdf data, which can act as an
+  improved 'bridge' between Iris and Xarray :
+
+Ncdata can convert Iris cubes to an Xarray dataset, or vice versa, with minimal
+overhead and as lossless as possible.
+
+For example :
+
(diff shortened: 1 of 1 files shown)


docs/src/community/iris_xarray.rst
```diff
@@ -107,15 +126,41 @@ However, attribute handling is not so complete: like Iris, it interprets and
...
+  where bounds variables do not usually define their units or standard_names can cause
+  problems.  Certain files containing bounds variables with more than 2 bounds (e.g.
+  unstructured data) may not load at all.
+* missing points are always represented as NaNs, as-per Pandas usage.
+  (See :ref:`xarray_missing_data` ).
+  This means that fill values are not preserved, and that masked integer data is
+  converted to floats.
+  The netCDF default fill-values are not supported, so that variables wit no
+  "_FillValue" attribute will have missing points equal to the fill-value
+  in place of NaNs.  By default, output variables generally have `_FillValue = NaN`.
```
stephenworsley, line 156: I think there should be a double ` here around "_FillValue = NaN".

docs/src/community/iris_xarray.rst
```diff
@@ -107,15 +126,41 @@ However, attribute handling is not so complete: like Iris, it interprets and
...
+which therefore may make it harder to correctly manage this type of data.
+While Xarray CF support is not complete, it may improve, and obviously
+:ref:`cfxarray` may be relevant here.
+There is also relevant documentation
+`at this page <https://docs.xarray.dev/en/stable/user-guide/weather-climate.html#weather-and-climate-data>`_.
+
+In some particular aspects, CF data is not loaded well (or at all), and in many cases
+output is not fully CF compliant (as-per `the cf checker <https://cfchecker.ncas.ac.uk/>`_).
+
+* it has it's own interpretation of coordinates, which is different from the CF-based
```
stephenworsley, line 140: It could be made clearer what is being refered to in this list. When we say "it" are we refering to xarray itself, an xarray object or the output of an xarray object?

stephenworsley (changes requested):
Looks good, a couple minor things, I think there are also other places within this page which use single \`s where they could use double \`\`s which would be good to also change for the sake of cosistency.

stephenworsley (approved):
Looks good!

PR #5756, opened 2024-02-19
Pin ASV - airspeed-velocity/asv#1385

trexfeathers:
### Description

This avoids a bug in Airspeed Velocity's dependencies: airspeed-velocity/asv#1385

---

### DIFF
--- a/.github/workflows/benchmarks_run.yml
+++ b/.github/workflows/benchmarks_run.yml
@@ -42,7 +42,7 @@ jobs:
 
       - name: Install ASV & Nox
         run: |
-          pip install asv nox
+          pip install "asv!=0.6.2" nox
 
       - name: Cache environment directories
         id: cache-env-dir


trexfeathers:
Successful benchmark run demonstrates that this pin is effective

ESadek-MO (approved):
LGTM, thanks @trexfeathers!

HaoZeke:
Note that the upstream bug is a bit spurious, and a better fix is to explicitly put the build project command in the asv configuration.

trexfeathers:
> Note that the upstream bug is a bit spurious, and a better fix is to explicitly put the build project command in the asv configuration.

@HaoZeke thanks, I was looking into that, until I saw that you were investigating including `build` as a core dependency, then it seemed best to wait for that to shake out.

If that doesn't work out then yeah we can add `build_command` back into `asv.conf.json`. I was quite enjoying having a vanilla setup without it, but I appreciate that not every project is a `pyproject.toml` project!

Issue #5753, opened 2024-02-16
Some iris netcdf saves are fetching all the data.


pp-mo:
From a support issue loading a large dataset from a GRIB file and saving to netcdf.
The example data was about 16Gb, and was running out of memory when saving.

It seems that at least **_some_** netcdf saves are not able to be correctly saved in a chunk-by-chunk streamed manner.

### Simple example to reproduce
```
import dask.array as da
import iris
from iris.cube import Cube
import tracemalloc

# Reasonably sized usage example, big enough to show problem
nt, ny, nx = 50, 700, 1400
# The data must be a stack of lazy arrays.
# A single random array with these chunks does *not* show the problem.
chunk_arrays = [
    da.random.uniform(0., 1, size=(ny, nx))
    for _ in range(nt)
]
data = da.stack(chunk_arrays)
cube = Cube(data)
print(f"Cube to save : {cube.summary(shorten=True)}")
data_size_mb = data.size * data.dtype.itemsize * 1.0 / 1024 ** 2
print(f"Cube data shape = {data.shape}, size = {data_size_mb:.1f} Mb")

tracemalloc.start()
iris.save(cube, 'tmp.nc')
_, peak_mem = tracemalloc.get_traced_memory()

memory_mb = peak_mem * 1.0 / 1024 ** 2
print(f"Peak memory used during save : {memory_mb:.1f}Mb")
```
#### from which, a sample output :
```
Cube to save : unknown / (unknown)                 (-- : 50; -- : 700; -- : 1400)
Cube data shape = (50, 700, 1400), size = 373.8 Mb
Peak memory used during save : 377.9Mb
```

### For comparison, it is OK in xarray (!! sorry 😬 !!) :
( using same data array )
```
import xarray as xr
xrda = xr.DataArray(data)

tracemalloc.stop()
tracemalloc.start()
xrda.to_netcdf('tmp.nc')
_, peak_mem = tracemalloc.get_traced_memory()

memory_mb = peak_mem * 1.0 / 1024 ** 2
print(f"Peak memory used during Xarray save : {memory_mb:.1f}Mb")
```
#### from which..
```
Peak memory used during Xarray save : 30.2Mb
```

## Expected result
The total memory required is **_expected_** to be only be around N or maybe N+1 * chunk sizes, 
where N is the number of Dask workers
 -- which here that was =4 (threads or "cpus").
So in this case, approx 8Mb per chunk, expected ~32-40 Mb., as seen in the Xarray case.

[...]

PR #5767, opened 2024-02-21
Fix usage of map_blocks in AreaWeighted and elsewhere

stephenworsley:
## 🚀 Pull Request

Addresses a bug raised by @hdyson  in `AreaWeighted` which caused weights to be retained in memory even after all relevent python objects had been removed. This appears to be due to passing weights to `dask.array.map_blocks` as part of a partial function. It seems that while dask will remove arrays from memory after use it will not do the same for python functions. Passing weights as arguments to `map_blocks` ought to fix this.

### DIFF
--- a/.github/labeler.yml
+++ b/.github/labeler.yml
@@ -1,0 +1,9 @@
+# benchmark_this automatically triggers the benchmark workflow when added by
+#  a user. No triggering happens when GitHub Actions adds the label (this
+#  avoids security vulnerabilities), so alternative triggers for the below
+#  files are therefore included in workflows/benchmarks_run.yml. Automatic
+#  labelling is still included here to make it easier to search pull requests,
+#  and to reinforce the culture of using this label.
+benchmark_this:
+- changed-files:
+  - any-glob-to-any-file: ['requirements/locks/*.lock', "setup.py"]
--- a/.github/pull_request_template.md
+++ b/.github/pull_request_template.md
@@ -7,3 +7,8 @@
 
 ---
 [Consult Iris pull request check list]( https://scitools-iris.readthedocs.io/en/latest/developers_guide/contributing_pull_request_checklist.html)
+
+---
+Add any of the below labels to trigger actions on this PR:
+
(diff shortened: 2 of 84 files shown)


pp-mo:
While I doubt it actually helps here, it's interesting to note that this problem may have similarities to https://github.com/SciTools/iris/issues/5753

What's similar is the way the operations are presented to Dask as a combination of tasks : 
In _this_ case, the operation uses a partial function which fixes a temporary numpy array argument.
In the netcdf-save case, a Dask.delayed is created which references a delayed store and an array calculation on the same data.  But the delayed function has obvious similarities to a partial function operation.

However, in that case, it seems more like dask is unnecessarily assembling a complete array in memory when both operations are entirely chunk-able with the same chunks.   

What makes me think that the save problem is **_not_** quite the same problem as here, is that in that case the memory use is _recovered at the end of the operation_.  
Which I think is not the case here @stephenworsley ?

stephenworsley:
@pp-mo Yes, that sounds like a slightly different issue. In the case of regridding, it doesn't seem like there is any problem with the data that is being *realised by dask*, the problem is with the data that lives in the partial function *already realised* and which isn't being thrown away when we would expect it to.

pp-mo:
> [no problem with] data that is being _realised by dask_, the problem is with the data that lives in the partial function _already realised_

Yes, that's a telling way of looking at it.  Those really are quite different things.
I'd say that broadly Dask is responsible for both, but in this case it actually feels like some kind of _bug_, whereas the other feels like a problem with its scheduling strategies.

trexfeathers:
@bouweandela suggests a possible future implementation that could be more efficient: making the weights array always a Dask array, then `compute()` only when needed, to avoid any memory footprint and make it easy to share information between processes/nodes/etcetera.

pp-mo (approved):
After some offline discussion, I'm now happy with this.
I re-ran a performance test example, and it does seem to be working (process memory usage no longer grows with a repeating large regrid operation).

PR #5773, opened 2024-02-22
Use the SciTools app for labeler permissions

trexfeathers:
### Description

Ran into [this problem](https://github.com/orgs/community/discussions/25565) when testing - benchmarks didn't trigger despite being labelled (#5772).

Logic dictates that we should test this, given the previous problem, but I don't know how that can be achieved given we need the App which is only on SciTools. I might try using `pull_request` instead of `pull_request_target`, in case the App gets round some of the other permissions...

---


trexfeathers:
Looking at other examples, the label needs to be added by `scitools-ci`, not `github-actions`.

trexfeathers:
I don't think this a go-er, based on the fact that [others haven't been able to get it working with pull requests](https://github.com/orgs/community/discussions/25565#discussioncomment-3429468), and that `peter-evans/create-pull-request` has been [designed with this problem in mind](https://github.com/peter-evans/create-pull-request/blob/main/docs/concepts-guidelines.md#triggering-further-workflow-runs), so may be working in a more sophisticated way that perhaps doesn't work for other actions.

I reckon it's SciTools/workflows#28 or bust

PR #5776, opened 2024-02-22
ASV custom build command and file-based benchmark triggers

trexfeathers:
### Description

- Customises `build_command` in `asv.conf.json` to use Iris' `setup.py` file
- Introduces file monitoring to make sure the benchmarks are run against files likely to affect performance or likely to affect the benchmarking itself.

---
[Consult Iris pull request check list]( https://scitools-iris.readthedocs.io/en/latest/developers_guide/contributing_pull_request_checklist.html)

PR #5763, opened 2024-02-20
Add an auto-labeller to always benchmark lock file changes.

trexfeathers:
### Description

Dependency changes are the most common cause of performance regressions, but we often forget to benchmark these changes so things get missed.

#### To do

- [ ] What's New entry, once 3.8 has been cut.

---
[Consult Iris pull request check list]( https://scitools-iris.readthedocs.io/en/latest/developers_guide/contributing_pull_request_checklist.html)

PR #5779, opened 2024-02-23
Revert SciTools/iris#5763

trexfeathers:
### Description

#5776 shows promise but can't currently be demonstrated while the failed attempt at labelling is still present.

---


trexfeathers:
@HGWright if you thought it better for traceability: we could keep the labeler action IN ADDITION to #5776. As we know, adding the label won't always actually trigger the benchmarks, but it would flag up anything that is being benchmarked via the alternative trigger!

trexfeathers:
> @HGWright if you thought it better for traceability: we could keep the labeler action IN ADDITION to #5776? As we know, adding the label won't always actually trigger the benchmarks, but it would flag up anything that is being benchmarked via the alternative trigger!

Closing after offline conversation

PR #5783, opened 2024-02-26
DOC: clarify save_pairs_from_cube docstring

rcomer:
### Description

I'm glad this function exists publicly as it was the solution to my problem today but I found the docstring confusing: because it said it could convert an iterable of cubes, I tried passing a cubelist, which failed because cubelists have no `shape` attribute.

I also note that the `target` parameter does not appear to be used for anything, but I don't think I want to pull on that thread right now...

---

### DIFF
--- a/lib/iris/fileformats/pp.py
+++ b/lib/iris/fileformats/pp.py
@@ -2177,11 +2177,11 @@ def save(cube, target, append=False, field_coords=None):
     of cubes to be saved to a PP file.
 
     """
-    fields = as_fields(cube, field_coords, target)
+    fields = as_fields(cube, field_coords)
     save_fields(fields, target, append=append)
 
 
-def save_pairs_from_cube(cube, field_coords=None, target=None):
+def save_pairs_from_cube(cube, field_coords=None):
     """Use the PP saving rules to generate (2D cube, PP field) pairs from a cube.
 
     Parameters
@@ -2193,8 +2193,6 @@ def save_pairs_from_cube(cube, field_coords=None, target=None):
         reducing the given cube into 2d slices, which will ultimately
         determine the x and y coordinates of the resulting fields.
         If None, the final two  dimensions are chosen for slicing.
-    target : optional
-        A filename or open file handle.
 
     Yields
     ------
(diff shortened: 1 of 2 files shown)


rcomer:
No sign of RTD in the CI 😕

rcomer:
[Renders like this](https://scitools-iris--5783.org.readthedocs.build/en/5783/generated/api/iris.fileformats.pp.html#iris.fileformats.pp.save_pairs_from_cube)

lib/iris/fileformats/pp.py
```diff
@@ -2197,6 +2194,11 @@ def save_pairs_from_cube(cube, field_coords=None, target=None):
     target : optional
         A filename or open file handle.
 
+    Yields
+    ------
+    :class:`iris.cube.Cube`, :class:`iris.fileformats.pp.PPField`.
+        A 2-dimensional slice of the input cube together with its associated pp-field.
```
pp-mo, line 2200: Can this be phrased so it's clear you get a stream of them ?
```suggestion
        2-dimensional slices of the input cube together with their associated pp-fields.
```

Would this read a bit better, to emphasize that there are a sequence of them ?

pp-mo:
> the target parameter does not appear to be used for anything

I chased this down.  It used to have a role in "rules logging", which we got rid of a long time ago.

I'm also not sure how we go about tidying this.
Given how obscure this is, perhaps we could even just remove it (and the one in "as_fields"), and issue a warning in a whatsnew ?
I mean, maybe we don't need to be as seriously backwards-compatible as Iris.
@trexfeathers what do you reckon to this one ?

pp-mo:
> I'm glad this function exists publicly as it was the solution to my problem today 
BTW maybe you can just as well use "as_fields" ?  I'd be interested to know if not.

rcomer:
> BTW maybe you can just as well use "as_fields" ? I'd be interested to know if not.

Yes I could and I think that would be cleaner - thanks!  I got here by looking at the example in [the user guide](https://scitools-iris.readthedocs.io/en/latest/userguide/saving_iris_cubes.html#customising-the-save-process) which I think could also use `as_fields`

trexfeathers:
>Given how obscure this is, perhaps we could even just remove it (and the one in "as_fields"), and issue a warning in a whatsnew ?

Yup, that's my preferred solution. We can always bugfix it back in if someone's calls are still expecting more arguments.

pp-mo (approved):
OK it's backwards-breaking, but we've discussed that enough I think.
Let's run with it + see what happens!

rcomer:
Thanks @pp-mo!

Issue #5574, opened 2023-11-10
Encourage PyTest, discourage old-style, in Iris documentation


trexfeathers:
## 📚 Documentation

We are already encouraging use of PyTest for **running** Iris tests: [Running the Tests — Iris 3.7.0 documentation (scitools-iris.readthedocs.io)](https://scitools-iris.readthedocs.io/en/stable/developers_guide/contributing_running_tests.html)

However we still encourage **writing** old-style tests:

- [Testing tools — Iris 3.7.0 documentation (scitools-iris.readthedocs.io)](https://scitools-iris.readthedocs.io/en/stable/developers_guide/testing_tools.html)
- [Test Categories — Iris 3.7.0 documentation (scitools-iris.readthedocs.io)](https://scitools-iris.readthedocs.io/en/stable/developers_guide/contributing_testing.html)

We should re-write these pages to encourage use of PyTest. Later we could capture any shared conventions on use of PyTest in the same places.

PR #5785, opened 2024-02-26
Added a style guide for iris pytest

ESadek-MO:
## 🚀 Pull Request

Might address #5574?

This is a style guide to offer some consistency on iris' approach to testing.


docs/src/developers_guide/testing_style_guide.rst
```diff
@@ -0,0 +1,67 @@
...
+There should be a conftest.py file in the root/unit and root/integration
+folders. Additional lower level conftests can be added if it is agreed there
+is a need.
+
+Fixtures
+========
+
+As far as is possible, the actual test function should do little else but the
+actual assertion. However, in some cases this will not be applicable, so this
+will have to be decided on a case by case basis.
```
trexfeathers, line 34: I'd prefer less vague phrasing. You want to avoid stuff like:

>This is true except when it is not

You've done a great job of more specific phrasing in the following paragraph. So how to do the same with this one... The conflict here is readability versus maintainability:

- Max readability = the whole sequence written in a single test file - you can see everything in one place
- Max maintainability = all setup abstracted into fixtures - you can alter all relevant tests with a single change, distinction is made between setup and test when things go wrong

Something like the below?

```suggestion
actual assertion. Separating preparation into fixtures can make the code
harder to follow, so compromises are acceptable; e.g. setting up a test ``Cube``
should be a fixture, creating a simple string (``expected = "foo"``) for a single
test should not be a fixture. 
```

docs/src/developers_guide/testing_style_guide.rst
```diff
@@ -0,0 +1,67 @@
...
+
+Where suitable, tests should be located within the relevant directories.
+In most circumstance, that means new tests should not be placed in the
+root test directory, but in the relevant sub-folders.
+
+Conftest.py
+===========
+
+There should be a conftest.py file in the root/unit and root/integration
+folders. Additional lower level conftests can be added if it is agreed there
```
trexfeathers, line 26:
```suggestion
folders. Additional lower level ``conftest``s can be added if it is agreed there
```

Maybe? For consistency and to draw attention?

docs/src/developers_guide/testing_style_guide.rst
```diff
@@ -0,0 +1,67 @@
...
+is a need.
+
+Fixtures
+========
+
+As far as is possible, the actual test function should do little else but the
+actual assertion. However, in some cases this will not be applicable, so this
+will have to be decided on a case by case basis.
+
+New fixtures should always be considered for conftest when added. If it is
```
trexfeathers, line 36:
```suggestion
New fixtures should always be considered for ``conftest`` when added. If it is
```

?

docs/src/developers_guide/testing_style_guide.rst
trexfeathers, line 1: Cross-reference to the pytest documentation whenever a concept is introduced. So perhaps most of the headings could be links.

E.g.

- Parameters: https://docs.pytest.org/en/stable/example/parametrize.html
- Parametrizing fixtures: https://docs.pytest.org/en/stable/how-to/fixtures.html#fixture-parametrize
- Etcetera

docs/src/developers_guide/testing_style_guide.rst
```diff
@@ -0,0 +1,67 @@
...
+
+Though it is a useful tool, we should not be complicating tests to work around
+parameters; they should only be used when it is simple and apparent to implement.
+
+Where you are parameterising multiple tests with the same parameters, it is
+usually prudent to use the parameterisation within fixtures. When doing this,
+ensure within the tests that it is apparent it's being parameterised,
+either within the fixture name or with comments.
+
+All parameterisation benefits from ids, and so should be used where possible.
```
trexfeathers, line 51:
```suggestion
All parameterisation benefits from ``ids``, and so should be used where possible.
```

Definitely deserves a link to provide context.

docs/src/developers_guide/testing_style_guide.rst
```diff
@@ -0,0 +1,67 @@
...
+
+Classes
+=======
+
+How and when to group tests within classes can be based on personal opinion,
+we do not deem consistency on this a vital concern.
```
trexfeathers, line 57: The below will need changing otherwise we had a direct contradiction.

https://github.com/SciTools/iris/blob/9a4a45b874501c2ace720628c7cde9afbd438dc3/docs/src/developers_guide/contributing_testing.rst?plain=1#L48-L52

https://github.com/SciTools/iris/blob/9a4a45b874501c2ace720628c7cde9afbd438dc3/docs/src/developers_guide/contributing_testing.rst?plain=1#L98-L101

docs/src/developers_guide/testing_style_guide.rst
```diff
@@ -0,0 +1,67 @@
...
+
+Mocks
+=====
+
+Any mocking should be done with pytest.mock, and monkeypatching where suitable.
```
trexfeathers, line 62:
```suggestion
Any mocking should be done with ``pytest.mock``, and monkeypatching where suitable.
```

Need some more links. Could maybe consider [intersphinx](https://github.com/SciTools/iris/blob/9a4a45b874501c2ace720628c7cde9afbd438dc3/docs/src/conf.py#L236-L245) so you could use a format like `:func:...`, but that might be overkill.

docs/src/developers_guide/testing_style_guide.rst
trexfeathers, line 1: We're currently expecting people to read several different pages before they have a full picture:

https://github.com/SciTools/iris/blob/9a4a45b874501c2ace720628c7cde9afbd438dc3/docs/src/developers_guide/testing_style_guide.rst?plain=1#L5-L6

https://github.com/SciTools/iris/blob/9a4a45b874501c2ace720628c7cde9afbd438dc3/docs/src/developers_guide/contributing_testing.rst?plain=1#L6-L7

https://github.com/SciTools/iris/blob/9a4a45b874501c2ace720628c7cde9afbd438dc3/docs/src/developers_guide/testing_tools.rst?plain=1#L5-L6

- A confusing read
- Too easy for inconsistencies to creep in (several of which I have mentioned in this review)

For me this would be much better as a single **Writing Tests** page.

trexfeathers (changes requested):
Thanks for this @ESadek-MO!

You've definitely captured everything the team discussed offline, and the spirit of compromise is much more realistic than many style guides I've read in the past 💐

Some changes below

docs/src/developers_guide/contributing_tests.rst
```diff
@@ -0,0 +1,301 @@
...
+
+.. note::
+    If you think we're missing anything important here, please consider creating an
+    issue or discussion and share your ideas with the team!
+
+.. _pytesting_tools:
+
+=============
+Testing tools
+=============
```
trexfeathers, line 230: This heading has not rendered at the correct level

docs/src/developers_guide/contributing_tests.rst
```diff
@@ -0,0 +1,301 @@
...
+After the change from ``unittest`` to ``pytest``, ``IrisTest.patch`` has been
+converted into :meth:`~iris.tests._shared_utils.patch`.
+
+This is currently not implemented, and will raise an error if called.
+
+Graphic tests
+-------------
+
+As a package capable of generating graphical outputs, Iris has utilities for
+creating and updating graphical tests - see :ref:`testing.graphics` for more
```
trexfeathers, line 300: I've read through that dedicated page and I don't believe anything needs changing 👍

docs/src/developers_guide/contributing_tests.rst
trexfeathers, line 1: Can confirm that this is a faithful lift-and-shift from all the 'sub-pages'. The sequence makes sense to me.

Happy to delete those now?

docs/src/developers_guide/contributing_tests.rst
```diff
@@ -0,0 +1,301 @@
...
+
+    Sometimes code changes alter the results expected from a test containing the
+    above methods. These can be updated by removing the existing result files
+    and then running the file containing the test with a ``--create-missing``
+    command line argument, or setting the ``IRIS_TEST_CREATE_MISSING``
+    environment variable to anything non-zero. This will create the files rather
+    than erroring, allowing you to commit the updated results.
+
+Context managers
+----------------
```
trexfeathers, line 278: This heading needs to be a higher level

docs/src/developers_guide/contributing_tests.rst
```diff
@@ -0,0 +1,301 @@
...
+
+This style guide should be approached pragmatically. Most of the guidelines laid out
+below will not be practical in every scenario, and as such should not be considered
+firm rules.
+
+At time of writing, some existing tests have already been written in PyTest,
+so might not be abiding by these guidelines.
+
+Directory
+---------
```
trexfeathers, line 168: This section overlaps significantly with:

https://github.com/SciTools/iris/blob/37a342f3c2f005e945347ed0d58654c72fbc2b3c/docs/src/developers_guide/contributing_tests.rst?plain=1#L11-L12

docs/src/developers_guide/contributing_tests.rst
```diff
@@ -0,0 +1,301 @@
...
+
+Where suitable, tests should be located within the relevant directories.
+In most circumstance, that means new tests should not be placed in the
+root test directory, but in the relevant sub-folders.
+
+`conftest.py <https://docs.pytest.org/en/7.1.x/reference/fixtures.html#conftest-py-sharing-fixtures-across-multiple-files>`_
+----------------------------------------------------------------------------------------------------------------------------
+
+There should be a ``conftest.py`` file in the ``root/unit`` and ``root/integration``
+folders. Additional lower level conftests can be added if it is agreed there
```
trexfeathers, line 178:
```suggestion
folders. Additional lower level ``conftest``s can be added if it is agreed there
```

Maybe? For consistency and to draw attention?

docs/src/developers_guide/contributing_tests.rst
```diff
@@ -0,0 +1,301 @@
...
+------------------------------------------------------------------------------------
+
+As far as is possible, the actual test function should do little else but the
+actual assertion. Separating off preparation into fixtures may make the code
+harder to follow, so compromises are acceptable. For example, setting up a test
+``Cube`` should be a fixture, whereas creating a simple string
+(``expected = "foo"``), or a single use setup, should *not* be a fixture.
+
+
+New fixtures should always be considered for conftest when added. If it is
```
trexfeathers, line 191:
```suggestion
New fixtures should always be considered for ``conftest`` when added. If it is
```

?

docs/src/developers_guide/contributing_tests.rst
```diff
@@ -0,0 +1,301 @@
...
+:meth:`~iris.tests.IrisTest_nometa.assert_no_warnings_regexp` and
+:meth:`~iris.tests.IrisTest_nometa.assert_logs`.
+
+Patching
+--------
+
+After the change from ``unittest`` to ``pytest``, ``IrisTest.patch`` has been
+converted into :meth:`~iris.tests._shared_utils.patch`.
+
+This is currently not implemented, and will raise an error if called.
```
trexfeathers, line 294: I believe we have discovered the `mock.patch` pattern is the perfect replacement for this?

docs/src/developers_guide/contributing_tests.rst
```diff
@@ -0,0 +1,301 @@
...
+Context managers
+----------------
+
+Capturing exceptions and logging
+--------------------------------
+
+:mod:`~iris.tests._shared_utils` includes several context managers that can be used
+to make test code tidier and easier to read. These include
+:meth:`~iris.tests.IrisTest_nometa.assert_no_warnings_regexp` and
+:meth:`~iris.tests.IrisTest_nometa.assert_logs`.
```
trexfeathers, line 286: Needs changing to reference `_shared_utils`

docs/src/developers_guide/contributing_tests.rst
```diff
@@ -0,0 +1,301 @@
...
+    All functions listed on this page are defined within
+    :mod:`iris.tests._shared_utils`. They can be accessed within a test using
+    ``_shared_utils.exampleFunction``.
+
+Custom assertions
+-----------------
+
+:mod:`iris.tests._shared_utils` supports a variety of custom pytest-style
+assertions, such as :meth:`~iris.tests._shared_utils.assert_array_equal`, and
+:meth:`~iris.tests._shared_utils.assert_array_almost_equal`.
```
trexfeathers, line 253: These are now `:func:` rather than `:meth:`

docs/src/developers_guide/contributing_tests.rst
```diff
@@ -0,0 +1,301 @@
...
+
+Saving results
+--------------
+
+Some tests compare the generated output to the expected result contained in a
+file. Custom assertions for this include
+:meth:`~iris.tests._shared_utils.assert_CML_approx_data`
+:meth:`~iris.tests._shared_utils.assert_CDL`
+:meth:`~iris.tests._shared_utils.assert_CML` and
+:meth:`~iris.tests._shared_utils.assert_text_file`. See docstrings for more
```
trexfeathers, line 265: `:func:`

docs/src/developers_guide/contributing_tests.rst
```diff
@@ -0,0 +1,301 @@
...
+support writing tests. Using these makes tests quicker and easier to write, and
+also consistent with the rest of Iris (which makes it easier to work with the
+code). Most of these conveniences are accessed through the
+:mod:`iris.tests._shared_utils` module.
+
+.. tip::
+
+    All functions listed on this page are defined within
+    :mod:`iris.tests._shared_utils`. They can be accessed within a test using
+    ``_shared_utils.exampleFunction``.
```
trexfeathers, line 246: Probably best to use the non-unittest format e.g. `example_function`

trexfeathers (changes requested):

docs/src/developers_guide/contributing_tests.rst
trexfeathers: Revisiting after several months away. I was referring to:

- `testing_style_guide`
- `contributing_testing`
- `testing_tools`

I believe all the content from the above 3 files is now in this file.

docs/src/developers_guide/contributing_tests.rst
ESadek-MO: Yeah I think we can remove the ``Directory`` section

docs/src/developers_guide/contributing_tests.rst
```diff
@@ -0,0 +1,301 @@
...
+
+    :literal:`"""Unit tests for the \`iris.experimental.raster.export_geotiff\` function."""`
+
+All unit tests must be placed and named according to the following
+structure:
+
+
+.. _pytesting.classes:
+
+Classes
```
trexfeathers, line 46: The examples in this section still show classes inheriting from `tests.IrisTest`, which is now deprecated.

docs/src/developers_guide/contributing_tests.rst
```diff
@@ -0,0 +1,301 @@
...
+
+All parameterisation benefits from
+`ids <https://docs.pytest.org/en/stable/example/parametrize.html#different-options-for-test-ids>`_,
+and so should be used where possible.
+
+`Classes <https://docs.pytest.org/en/stable/getting-started.html#group-multiple-tests-in-a-class>`_
+---------------------------------------------------------------------------------------------------
+
+How and when to group tests within classes can be based on personal opinion,
+we do not deem consistency on this a vital concern.
```
trexfeathers, line 208: We need to thin out the `Test Categories` section, which is very prescriptive about class structure at the moment - a direct contradiction with the statement here.

docs/src/developers_guide/testing_style_guide.rst
```diff
@@ -0,0 +1,73 @@
...
+so might not be abiding by these guidelines.
+
+Directory
+=========
+
+Where suitable, tests should be located within the relevant directories.
+In most circumstance, that means new tests should not be placed in the
+root test directory, but in the relevant sub-folders.
+
+`conftest.py <https://docs.pytest.org/en/7.1.x/reference/fixtures.html#conftest-py-sharing-fixtures-across-multiple-files>`_
```
trexfeathers, line 22:
```suggestion
`conftest.py <https://docs.pytest.org/en/stable/reference/fixtures.html#conftest-py-sharing-fixtures-across-multiple-files>`_
```

docs/src/developers_guide/contributing_tests.rst
```diff
@@ -0,0 +1,301 @@
...
+so might not be abiding by these guidelines.
+
+Directory
+---------
+
+Where suitable, tests should be located within the relevant directories.
+In most circumstance, that means new tests should not be placed in the
+root test directory, but in the relevant sub-folders.
+
+`conftest.py <https://docs.pytest.org/en/7.1.x/reference/fixtures.html#conftest-py-sharing-fixtures-across-multiple-files>`_
```
trexfeathers, line 167:
```suggestion
`conftest.py <https://docs.pytest.org/en/stable/reference/fixtures.html#conftest-py-sharing-fixtures-across-multiple-files>`_
```

docs/src/developers_guide/contributing_tests.rst
```diff
@@ -0,0 +1,301 @@
+.. include:: ../common_links.inc
+
+.. _contributing_tests:
+
+***********************
+Contributing Iris Tests
```
trexfeathers, line 6: Looking at this against the other headings (`Adding or Updating Graphics Tests`, `Running the Tests`, etc), I'm wondering if this title would be better as `Writing Tests`, or something similar. What do you think?

docs/src/developers_guide/contributing_testing_index.rst
```diff
@@ -12,3 +12,5 @@ Testing
    contributing_running_tests
    contributing_ci_tests
    contributing_benchmarks
+   testing_style_guide
+   contributing_tests
```
trexfeathers, line 16: `contributing_tests` should the first in the list

docs/src/developers_guide/contributing_tests.rst
```diff
@@ -0,0 +1,301 @@
...
+Iris has various internal convenience functions and utilities available to
+support writing tests. Using these makes tests quicker and easier to write, and
+also consistent with the rest of Iris (which makes it easier to work with the
+code). Most of these conveniences are accessed through the
+:mod:`iris.tests._shared_utils` module.
+
+.. tip::
+
+    All functions listed on this page are defined within
+    :mod:`iris.tests._shared_utils`. They can be accessed within a test using
```
trexfeathers, line 238: I think we decided that we couldn't commit to publicly maintained API at this time, hence having `_shared_utils` private. However it does make this documentation harder to read, since references to API objects (e.g. `assert_logs()`) cannot link to anything in the docs. I would definitely be up for making a public module, analogous to `np.testing` (as opposed to `np.tests`), with documented conveniences that anyone can use.

If you agree, I will raise an issue.

trexfeathers (changes requested):

docs/src/developers_guide/contributing_tests.rst
trexfeathers: I see you've attempted a fix in a33b44c02cab0236ae65e3077faabe4a8417f638. I don't think the problem was the number of `=` characters, it was the _style_ of heading you chose.

If you use a heading style that is not consistent with previous styles, RST (or Sphinx?) will render this at a new, lower level.

docs/src/developers_guide/contributing_tests.rst
trexfeathers: Your fix in a33b44c02cab0236ae65e3077faabe4a8417f638 hasn't worked.

See https://github.com/SciTools/iris/pull/5785#discussion_r1824266194 for an explanation.

docs/src/developers_guide/converting_from_unittest.rst
```diff
@@ -0,0 +1,47 @@
...
+#. Check for references  to ``unittest``. Many of the functions within unittest
+   are also in pytest, so often you can just change where the function is imported
+   from.
+#. Check for references to `self.assert`. Pytest has a lighter-weight syntax for
+   assertions, e.g. ``assert x == 2`` instead of ``assertEqual(x, 2)``. In the
+   case of custom assertions, the majority of these have been replicated in
+   :mod:`iris.tests._shared_utils`, but with snake_case instead of camelCase.
+   Some :class:`iris.tests.IrisTest` assertions have not been converted into
+   :mod:`iris.tests._shared_utils`, as it was deemed these were easy to do
+   without.
```
trexfeathers, line 29:
```suggestion
   :mod:`iris.tests._shared_utils`, as these were deemed easy to achieve via
   simple ``assert ...`` statements.
```

docs/src/developers_guide/converting_from_unittest.rst
```diff
@@ -0,0 +1,47 @@
...
+--------------------
+.. note::
+    Please bear in mind the following checklist is for general use; there may be
+    some cases which require extra context or thought before implementing these changes.
+
+#. Before making any manual changes, run https://github.com/dannysepler/pytestify
+   on the file. This does a lot of the brunt work for you!
+#. Check for references to :class:`iris.tests.IrisTest`. If a class inherits
+   from this, remove it. :class:`iris.tests.IrisTest` has been deprecated, and
+   replaced with the :mod:`iris.tests._shared_utils` module.
```
trexfeathers, line 19:
```suggestion
   from this, remove the inheritance. Inheritance is unnecessary for 
   pytest tests, so :class:`iris.tests.IrisTest` has been deprecated
   and its convenience methods have been moved to the 
   :mod:`iris.tests._shared_utils` module.
```

docs/src/developers_guide/converting_from_unittest.rst
```diff
@@ -0,0 +1,47 @@
...
+#. Check for references  to ``unittest``. Many of the functions within unittest
+   are also in pytest, so often you can just change where the function is imported
+   from.
+#. Check for references to `self.assert`. Pytest has a lighter-weight syntax for
```
trexfeathers, line 23:
```suggestion
#. Check for references to ``self.assert``. Pytest has a lighter-weight syntax for
```

docs/src/developers_guide/converting_from_unittest.rst
```diff
@@ -0,0 +1,47 @@
...
+   assertions, e.g. ``assert x == 2`` instead of ``assertEqual(x, 2)``. In the
+   case of custom assertions, the majority of these have been replicated in
+   :mod:`iris.tests._shared_utils`, but with snake_case instead of camelCase.
+   Some :class:`iris.tests.IrisTest` assertions have not been converted into
```
trexfeathers, line 27:
```suggestion
   case of custom :class:`~iris.tests.IrisTest` assertions, the majority of these
   have been replicated in
   :mod:`iris.tests._shared_utils`, but with snake_case instead of camelCase.
   Some custom assertions have not been converted into
```

docs/src/developers_guide/converting_from_unittest.rst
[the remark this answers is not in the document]
ESadek-MO: That feels like I'd have to force it? contributing_to_pytset_conversion feels clunky

docs/src/developers_guide/converting_from_unittest.rst
```diff
@@ -0,0 +1,47 @@
...
+   :mod:`iris.tests._shared_utils`, as it was deemed these were easy to do
+   without.
+#. Check for references to ``setUp()``. Pytest recognises a specific method called
+   ``_setup()`` instead. Ensure that this is decorated with
+   ``@pytest.fixture(autouse=True)``.
```
trexfeathers, line 32: Not actually true - you can call the method whatever you like. The use of the fixture is what makes it work, and the underscore is for Ruff compliance because this method doesn't (normally) return anything.

If it's possible to do, this would be much clearer to understand as a 2-line code block.

trexfeathers (changes requested):
Great job reconstructing the true meaning of each step from our rough notes! 💐

docs/src/developers_guide/contributing_tests.rst
trexfeathers: Still need to be 1 level lower! I believe `Context managers` are considered a type of `Testing tools`.

docs/src/developers_guide/contributing_tests.rst
```diff
@@ -0,0 +1,263 @@
+.. include:: ../common_links.inc
+
+.. _contributing_tests:
+
+*************
+Writing Tests
+*************
+
+.. note::
+    If you're converting UnitTest tests to PyTest, check out :ref:`converting_tests`.
```
trexfeathers, line 10: `converting_tests` no longer works, has the reference changed name?

docs/src/developers_guide/contributing_pytest_conversions.rst
```diff
@@ -0,0 +1,55 @@
...
+   Some :class:`iris.tests.IrisTest` assertions have not been converted into
+   :mod:`iris.tests._shared_utils`, as these were deemed easy to achieve via
+   simple ``assert ...`` statements.
+#. Check for references to ``setUp()``. Replace this with ``_setup()`` instead.
+   Ensure that this is decorated with ``@pytest.fixture(autouse=True)``.
+   .. codeblock:: python
+
+      @pytest.fixture(autouse=True)
+      def _setup(self):
+         ...
```
trexfeathers, line 39: This has _almost_ rendered correctly, but clearly still needs some tweaking.

image

docs/src/developers_guide/contributing_tests.rst
```diff
@@ -0,0 +1,263 @@
...
+
+.. _developer_pytest_categories:
+
+Test Categories
+===============
+
+There are two main categories of tests within Iris:
+
+- :ref:`testing.unit_test`
+- :ref:`testing.integration`
```
trexfeathers, line 20: These are now broken links.

docs/src/developers_guide/contributing_tests.rst
```diff
@@ -0,0 +1,263 @@
...
+    # Tests focussing on the handling of different projections.
+    def test_no_ellipsoid(self):
+        ...
+
+Unlike unit tests, there is no fixed naming scheme for integration
+tests. But folders and files must be created as required to help
+developers locate relevant tests. It is recommended they are named
+according to the capabilities under test, e.g.
+``metadata/test_pp_preservation.py``, and not named according to the
+module(s) under test.
```
trexfeathers, line 195: More appropriate to be under `Test Categories`, since it is not about how to write a test.

trexfeathers: You're right that they still deserve a mention here, if only to say "write them how you want".

docs/src/developers_guide/contributing_tests.rst
```diff
@@ -0,0 +1,263 @@
...
+    class Test_add_dim_coord:
+        def test_normal_usage(self):
+            ...
+
+        def test_coord_already_present(self):
+            ...
+
+When testing a function all the tests must reside in the module:
+
+    :literal:`lib/iris/tests/unit/<fully/qualified/module>/test_<function_name>.py`
```
trexfeathers, line 156: Why has this moved into the style guide, while guidance on `test_<ClassName>` has not?

docs/src/developers_guide/contributing_tests.rst
```diff
@@ -0,0 +1,263 @@
...
+`Mocks <https://docs.pytest.org/en/stable/how-to/monkeypatch.html>`_
+--------------------------------------------------------------------
+
+Any mocking should be done with ``pytest.mock``, and monkeypatching where suitable.
+
+.. note::
+    If you think we're missing anything important here, please consider creating an
+    issue or discussion and share your ideas with the team!
+
+Naming Test Classes and Functions
```
trexfeathers, line 113: We have now lost the explicit statement that class/function structure is not strictly controlled. With the presence of the examples, this will likely lead to new arguments about strictness.

```
`Classes <https://docs.pytest.org/en/stable/getting-started.html#group-multiple-tests-in-a-class>`_
---------------------------------------------------------------------------------------------------

How and when to group tests within classes can be based on personal opinion,
we do not deem consistency on this a vital concern.
```

trexfeathers (changes requested):

docs/src/developers_guide/contributing_tests.rst
trexfeathers: #6210

docs/src/developers_guide/contributing_tests.rst
```diff
@@ -0,0 +1,262 @@
...
+        def test_int16_big_endian(self):
+            ...
+
+
+    # Tests focussing on the handling of different projections.
+    def test_no_ellipsoid(self):
+        ...
+
+There is no fixed naming scheme for integration tests.
+.. _testing_tools:
```
trexfeathers, line 199:
```suggestion

.. _testing_tools:
```

`_testing_tools` is currently being shown as part of the documentation text.

docs/src/developers_guide/contributing_tests.rst
```diff
@@ -0,0 +1,262 @@
...
+---------------------------------------------------------------------------------------------------
+
+How and when to group tests within classes can be based on personal opinion,
+we do not deem consistency on this a vital concern.
+
+Naming Test Classes and Functions
+---------------------------------
+
+Within a test module each tested method may have corresponding test classes,
+for example:
```
trexfeathers, line 129: We're missing some context since we moved things around. This section is talking about testing of **classes and their methods**; we should make that clear with a subheading or a sentence.

docs/src/developers_guide/contributing_tests.rst
```diff
@@ -0,0 +1,262 @@
...
+
+    # Tests for the Cube.add_dim_coord() method.
+    class Test_add_dim_coord:
+        def test_normal_usage(self):
+            ...
+
+        def test_coord_already_present(self):
+            ...
+
+Within this test module there may be test classes, for example:
```
trexfeathers, line 166: We're missing some context since we moved things around. This section is talking about testing of **functions**; we should make that clear with a subheading or a sentence.

trexfeathers (changes requested):
OK, once the below 3 things are actioned, I will merge 👍

trexfeathers (approved):

PR #5740, opened 2024-02-13
Iris to GeoVista conversion

HGWright:
### Description

From the issue #5517. This PR adds the ability to convert an Iris cube into a polydata mesh and then extract a region of that to be able to be plotted using geovista. This is a one way operation as it does not include the ability to go the other way from geovista to iris. There was plenty of discussion of the scope of this PR in the original issue.

---
[Consult Iris pull request check list]( https://scitools-iris.readthedocs.io/en/latest/developers_guide/contributing_pull_request_checklist.html)

PR #5788, opened 2024-02-27
Avoid computing lazy scalar coordinates when printing a Cube

### LINKED ISSUE #5743: Cube printout should not compute lazy scalar coordinates
To address #5723

We've accepted in the discussion there that we really don't want complex lazy calculations performed each time you print a cube.  
Nor would we like it to "realise" a scalar coordinate, so it only happens once, since you really wouldn't expect printing to _modify_ a cube.

It seems reasonable instead to just print something like "complex_scalarvalue_coord: \<lazy\>" in this case.
The user could of course realize it, if + when the value is actually _wanted_. 

Note: IMHO this is not a major change, and can just be done
  -- we have changed printout format before without considering it a breaking change.
Also relevant here : _most_ scalar coords, ones loaded from files, are not lazy, since #5229 
( so in many/most cases, adopting this _won't_ change the output )

fnattino:
### Description

The point and bound of a scalar coordinate are printed out only when the corresponding data is not lazy. Fix #5743. 

---

### DIFF
--- a/lib/iris/_representation/cube_summary.py
+++ b/lib/iris/_representation/cube_summary.py
@@ -5,6 +5,9 @@
 """Provides objects describing cube summaries."""
 import re
 
+import numpy as np
+
+import iris._lazy_data as _lazy
 from iris.common.metadata import hexdigest
 import iris.util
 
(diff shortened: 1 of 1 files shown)


pp-mo:
Note: we are planning now a minor "interim" release to include #5833  and #5740
Ping @trexfeathers can we fit this in ?

Status : I think ..
  * this basically works
  * a fix for #5743 is very desirable
  * the new behaviour (coords print differently ) should not be considered a breaking change
  * the broken test is a simple fix
  * however, some more specific tests would be a good thing

pp-mo:
@fnattino if it's ok with you, I'm going to take this + add a couple of commits to fix + extend the testing slightly.
Then we can bank it.
Because we're intending to cut Iris 3.9.0.rc0 today

trexfeathers:
>Because we're intending to cut Iris 3.9.0.rc0 today

tomorrow*

ESadek-MO:
Completed in #5896 .

fnattino:
Many apologies for the lack of response, I have been busy with other things recently and away for the last week or so. Thanks a lot @pp-mo for picking this up, happy to hear #5896 has been merged.

pp-mo:
> Many apologies for the lack of response, I have been busy with other things recently and away for the last week or so. Thanks a lot @pp-mo for picking this up, happy to hear #5896 has been merged.

No problem at all.  Thanks for contributing !

PR #5797, opened 2024-03-04
Rethink netcdf delayed saving to avoid memory problem.

pp-mo:
Addresses #5753

This solution does **_not_** follow [the pattern explained in 5753](https://github.com/SciTools/iris/issues/5753#issuecomment-1948994177)
Because, it seems, that solution is specific to cases where the arrays stacked by merge are all _whole_ chunks (e.g. 2d fields from PP or GRIB files).  
Whereas, if those component arrays are _themselves_ divided into chunks, those chunks get ignored when a `map_blocks` call with `drop_axis` is used -- i.e. the result of that cannot be chunked across the 'missing' axes.  And in consequence, the save still ends up fetching large numbers of chunks into memory at once.

So in effect, that earlier solution does not generalise well.

Instead, I have now modified the write-proxy objects to do the fill-value checking when they are written to, and I think that fixes the memory problem.
At least, the test provided in #5753 now shows the lower memory usage, equivalent to xarray.

**WIP** : pushing up to check correctness against full tests

**TODO** :  
  - [ ] check that this does still work when component arrays of a `da.stack` are themselves chunked.
  - [ ] provide a benchmark ?

### DIFF
--- a/lib/iris/fileformats/netcdf/_thread_safe_nc.py
+++ b/lib/iris/fileformats/netcdf/_thread_safe_nc.py
@@ -352,6 +352,34 @@ class NetCDFDataProxy:
             setattr(self, key, value)
 
 
+def data_fillvalue_check(data, check_value):
+    """Check whether a numpy array is masked, and whether it contains a fill-value.
+
+    Parameters
+    ----------
+    data : array-like
+        Array to check (numpy or dask).
+    check_value : number or None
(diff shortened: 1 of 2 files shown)


pp-mo:
OK this solution is not viable as it doesn't work with a distributed (i.e. process-based scheduler).  
Which on reflection is no surprise !

Will post an alternative approach when I have worked one out ...

PR #5588, opened 2023-11-16
Mergeback of `FEATURE_chunk_control` branch

trexfeathers:
### Description

Still a draft at this time. Raised to test merge-ability, have an account of the full diff, and to get a coverage report.

---
[Consult Iris pull request check list]( https://scitools-iris.readthedocs.io/en/latest/developers_guide/contributing_pull_request_checklist.html)

PR #5801, opened 2024-03-04
Use the correct `chunktype` for Dask arrays

### LINKED ISSUE #5800: `iris.cube.Cube.lazy_data` method results in wrong chunk array type for masked arrays
## How To Reproduce

Run the following code
```python
import numpy as np
import iris.cube

In [1]: iris.cube.Cube(np.ma.array([1, 2], mask=[True, False])).lazy_data()
Out[1]: dask.array<array, shape=(2,), dtype=int64, chunksize=(2,), chunktype=numpy.ndarray>
```
Note that chunktype is `numpy.ndarray`, while the data type is actually a masked array. This causes problems when inspecting the content of the Dask array with `dask.array.utils.meta_from_array`, because that will return the wrong chunk type shown above.

## Expected behaviour

```python
import numpy as np
import iris.cube

In [1]: iris.cube.Cube(np.ma.array([1, 2], mask=[True, False])).lazy_data()
Out[1]: dask.array<array, shape=(2,), dtype=int64, chunksize=(2,), chunktype=numpy.MaskedArray>

```

## Environment 
 - OS & Version: 23.10
 - Iris Version: 3.9.0.dev15

bouweandela:
### Description

Use the correct `chunktype` for Dask arrays. Closes #5800.

---

---

### DIFF
--- a/lib/iris/_lazy_data.py
+++ b/lib/iris/_lazy_data.py
@@ -423,7 +423,7 @@ def stack(
 ) -> da.Array | np.ndarray:
     """Stack a sequence of arrays along a new axis.
 
-Improves on the regular :func:`dask.array.stack` by always respecting a masked
+    Improves on the regular :func:`dask.array.stack` by always respecting a masked
     ``.meta``, as described for :func:`_combine`.
 
     Parameters
--- a/lib/iris/_lazy_data.py
+++ b/lib/iris/_lazy_data.py
@@ -9,6 +9,7 @@ To avoid replicating implementation-dependent test and conversion code.
 """
 
 from functools import lru_cache, wraps
+from typing import Sequence
 
 import dask
 import dask.array as da
@@ -221,7 +222,7 @@ def _optimum_chunksize(
 
 
 def as_lazy_data(
-    data, chunks=None, asarray=False, dims_fixed=None, dask_chunking=False
+    data, chunks=None, asarray=False, meta=None, dims_fixed=None, dask_chunking=False
 ):
     """Convert the input array `data` to a :class:`dask.array.Array`.
 
@@ -235,6 +236,9 @@ def as_lazy_data(
     asarray : bool, default=False
         If True, then chunks will be converted to instances of `ndarray`.
         Set to False (default) to pass passed chunks through unchanged.
+    meta : numpy.ndarray, optional
+        Empty ndarray created with same NumPy backend, ndim and dtype as the
+        Dask Array being created.
     dims_fixed : list of bool, optional
         If set, a list of values equal in length to 'chunks' or data.ndim.
         'True' values indicate a dimension which can not be changed, i.e. the
@@ -258,12 +262,14 @@ def as_lazy_data(
     but reduced by a factor if that exceeds the dask default chunksize.
 
     """
+    if isinstance(data, ma.core.MaskedConstant):
+        data = ma.masked_array(data.data, mask=data.mask)
+
     if dask_chunking:
         if chunks is not None:
             raise ValueError(
                 f"Dask chunking chosen, but chunks already assigned value {chunks}"
             )
-        lazy_params = {"asarray": asarray}
     else:
         if chunks is None:
             # No existing chunks : Make a chunk the shape of the entire input array
@@ -281,14 +287,9 @@ def as_lazy_data(
                 dtype=data.dtype,
                 dims_fixed=dims_fixed,
             )
-        lazy_params = {
-            "chunks": chunks,
-            "asarray": asarray,
-        }
-    if isinstance(data, ma.core.MaskedConstant):
-        data = ma.masked_array(data.data, mask=data.mask)
+
     if not is_lazy_data(data):
-        data = da.from_array(data, **lazy_params)
+        data = da.from_array(data, chunks=chunks, asarray=asarray, meta=meta)
     return data
 
 
@@ -350,7 +351,22 @@ def as_concrete_data(data):
     return data
 
 
-def multidim_lazy_stack(stack):
+def stack(seq: Sequence[da.Array | np.ndarray]) -> da.Array:
+    """Stack arrays along a new axis.
+
+    This version of :func:`da.stack` ensures all slices of the resulting array
+    are masked arrays if any of the input arrays is masked.
+    """
+
+    def is_masked(a):
+        return isinstance(da.utils.meta_from_array(a), np.ma.MaskedArray)
+
+    if any(is_masked(a) for a in seq):
+        seq = [a if is_masked(a) else da.ma.masked_array(a) for a in seq]
+    return da.stack(seq)
+
+
+def multidim_lazy_stack(arr):
     """Recursively build a multidimensional stacked dask array.
 
     This is needed because :meth:`dask.array.Array.stack` only accepts a
@@ -358,7 +374,7 @@ def multidim_lazy_stack(stack):
 
     Parameters
     ----------
-    stack :
+    arr :
         An ndarray of :class:`dask.array.Array`.
 
     Returns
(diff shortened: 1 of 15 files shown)


pp-mo:
Hi @bouweandela 
The existing test breakage is a fairly minor matter, but I think it indicates that removing line 287b is causing the `da.from_array` to do trial access on the data -- so I think it is probably better to continue to provide a `meta` key in this case.

However, I am also concerned that ideally we should _also_ be addressing #5800 if possible (!)
In which case, I think this should have `meta=np.ma.MaskedArray` in place of `np.ndarray`, when appropriate.

From my reading of dask docs, I think it may also be better to pass a zero-shaped instance instead of a class, since the [from_array docs](https://docs.dask.org/en/stable/generated/dask.array.from_array.html#dask.array.from_array) don't actually mandate that usage.
And it also doesn't actually work if you pass `np.ma.MaskedArray` instead : you get a `TypeError: MaskedArray.__new__() got an unexpected keyword argument 'shape'`.

So I think something like `da.from_array(real, meta=real.__class__((), dtype=real.dtype))` works here, and is probably a better way.

Does this make sense to you ?

bouweandela:
@pp-mo I also updated the `concatenate` function so it avoids the slicing issue you mentioned in https://github.com/SciTools/iris/issues/5800#issuecomment-2025165305. This is now ready for review.

I suspect that the pre-commit failure is unrelated to this pull request. When I run `pre-commit run --all-files` on the `main` branch I get the same error.

pp-mo:
> > @bouweandela I suspect that the pre-commit failure is unrelated to this pull request. When I run `pre-commit run --all-files` on the `main` branch I get the same error.
> 
> Yes, puzzling. We are looking into it...

Ok I think we know what this is now  -- fix coming soon.
It's only "odd" because the flake8 error code is bugged !

lib/iris/_lazy_data.py
```diff
@@ -232,17 +236,18 @@ def as_lazy_data(
...
         If True, then chunks will be converted to instances of `ndarray`.
         Set to False (default) to pass passed chunks through unchanged.
+    meta : numpy.ndarray, optional
+        Empty ndarray created with same NumPy backend, ndim and dtype as the
+        Dask Array being created.
     dims_fixed : list of bool, optional
         If set, a list of values equal in length to 'chunks' or data.ndim.
         'True' values indicate a dimension which can not be changed, i.e. the
         result for that index must equal the value in 'chunks' or data.shape.
-    dask_chunking : bool, default=False
```
pp-mo, line 242: I think your merge has removed the changes made in #5588
Is this a mistake ?

lib/iris/_lazy_data.py
```diff
@@ -351,31 +353,110 @@ def as_concrete_data(data):
...
+    func = getattr(array_module, operation)
+    return func(arrays, **kwargs)
+
+
+def concatenate(
+    arrays: Sequence[da.Array | np.ndarray],
+    axis: int = 0,
+) -> da.Array | np.ndarray:
+    """Concatenate a sequence of arrays along a new axis.
+
```
pp-mo, line 396:
```suggestion

    Improves on the regular :func:`dask.array.concatenate` by always respecting a masked
    ``.meta``, as described for :func:`_combine`.

```

lib/iris/_lazy_data.py
```diff
@@ -351,31 +353,110 @@ def as_concrete_data(data):
...
+    """
+    return _combine(arrays, operation="concatenate", axis=axis)
+
+
+def stack(
+    arrays: Sequence[da.Array | np.ndarray],
+    axis: int = 0,
+) -> da.Array | np.ndarray:
+    """Stack a sequence of arrays along a new axis.
+
```
pp-mo, line 419:
```suggestion

Improves on the regular :func:`dask.array.stack` by always respecting a masked
    ``.meta``, as described for :func:`_combine`.

```

lib/iris/tests/unit/fileformats/netcdf/loader/test__chunk_control.py
```diff
@@ -203,7 +203,7 @@ def test_as_dask(tmp_filepath, save_cubelist_with_sigma):
             except RuntimeError as e:
                 if str(e) != message:
                     raise e
-        as_lazy_data.assert_called_with(ANY, chunks=None, dask_chunking=True)
+        as_lazy_data.assert_called_with(ANY, meta=ANY, chunks="auto")
```
pp-mo, line 206: Unfortunately here too I think you've clashed with #5588

lib/iris/fileformats/netcdf/_thread_safe_nc.py
```diff
@@ -323,6 +323,10 @@ def ndim(self):
         # noqa: D102
         return len(self.shape)
 
+    @property
+    def meta(self):
```
pp-mo, line 327:
```suggestion
    def meta(self):
```suggestion
    def dask_meta(self):
        """Make a suitable dask .meta for wrapping this as a dask array."""
```

Since these proxy objects aren't actually dask objects, they don't really have a meta themselves,
so I'm hoping that a slight  rename makes the purpose clearer.

N.B. if you accept, applies to usage below as well.

lib/iris/fileformats/pp.py
```diff
@@ -606,6 +606,10 @@ def fill_value(self):
     def ndim(self):
         return len(self.shape)
 
+    @property
+    def meta(self):
```
pp-mo, line 610: I suggest rename to "dask_meta" method, as for the NetCDFDataProxy one.

lib/iris/tests/unit/lazy_data/test_as_lazy_data.py
```diff
@@ -46,26 +46,28 @@ def test_dask_chunking(self):
         chunks = (12,)
         optimum = self.patch("iris._lazy_data._optimum_chunksize")
         optimum.return_value = chunks
-        _ = as_lazy_data(data, chunks=None, dask_chunking=True)
```
pp-mo, line 49: Probably more collisions with #5588

pp-mo (changes requested):
I think everything here is basically good, and it's a really nice touch to provide a fix for concatenate as well as stack.

However,  I think we need to rethink the merge, since some changes from #5588 seem to have got lost.
I guess that means that merging back into this will cause problems 
( I expect at least some tests ought then to fail, otherwise where is the testing of those new keywords ?? )
So I haven't updated with main, as you may want to resolve this in your own way.

From a quick search I also think that, unlike 'stack', there are multiple uses of 'concatenate' elsewhere in the Iris code that would benefit from using the new routine.
However, given how involved this is getting, it would make good sense to get this working + merged in **_first_** : then I can follow up with another to expand the use of ``_lazy.data.concatenate``.

lib/iris/fileformats/pp.py
bouweandela: Done in 00af67e

lib/iris/fileformats/netcdf/_thread_safe_nc.py
bouweandela: Done in 00af67e.

lib/iris/_lazy_data.py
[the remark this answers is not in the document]
bouweandela: I combined the functionality of the `dask_chunking` argument with the `chunks` argument (`dask_chunking = chunks == 'auto'`): https://github.com/SciTools/iris/blob/00af67eed9fce33a901d646d14034e7135d6bec9/lib/iris/_lazy_data.py#L237-L240
I did this because, with the addition of the `meta` argument, the function started to feel a bit bloated because of the many arguments. However, if you prefer the previous function signature, I would be happy to restore it.

bouweandela:
Thank you for reviewing @pp-mo! I just merged the latest `main` into this branch and the pre-commit test is passing fine now :partying_face: 

> However, I think we need to rethink the merge, since some changes from https://github.com/SciTools/iris/pull/5588 seem to have got lost.
I guess that means that merging back into this will cause problems
( I expect at least some tests ought then to fail, otherwise where is the testing of those new keywords ?? )
So I haven't updated with main, as you may want to resolve this in your own way.

I refactored the `as_lazy_data` function a bit, but the functionality is still there (see https://github.com/SciTools/iris/pull/5801#discussion_r1580022494). If you prefer, I can undo that?

lib/iris/_lazy_data.py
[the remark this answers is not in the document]
pp-mo: Ok, my apologies for being slow!
I can see that you have preserved the behaviour, and simplified the code too.
And I've now also realised that none of this is public API.

pp-mo:
> If you prefer, I can undo that?

Thanks no, I just hadn't understood fully what you were doing.  It's good !

pp-mo (approved):

PR #5798, opened 2024-03-04
v3.7.1 updates

trexfeathers:
# Merge using a merge commit

(Do not squash)

- Revert the mistaken merge of `main` into `v3.7.x` (875e890)
- Cherry-pick area weighted memory fix (c6151e819189ca213b370ddf4ff38d8511c7e012)
- Update What's New

Combined, this should update `v3.7.x` to the state at `v3.7.0` plus the change in #5767.

PR #5802, opened 2024-03-04
v3.7.1 merge-back

trexfeathers:
# Merge via a merge commit

(Do not squash)

No doubt problematic due to my reversion in #5798

The only meaningful change needed is the change to the What's New. Would it be acceptable to just create a pull request that manually makes that change? Or is it important that `main` includes the commits from `3.7.x`?


trexfeathers:
Replaced by #5857

PR #5804, opened 2024-03-05
Include ancillary variables and cell_measures during intersect

### LINKED ISSUE #5413: Cube loses ancillary variables and cell_measures after (some) intersections
Hey folks,

Here's one sneaky one for you:

### Main point

An iris cube that has ancillary variables and cell_measures, upon intersecting it with a region via `cube.intersect` may lose those ancillary variables and cell_measures depending on what the coordinates of the intersected region are.

### Minimal fully reproducible code

```python
import iris


cube = iris.load_cube("cube_to_intersect_smol.nc")

def extract_region(cube, start_longitude, end_longitude, start_latitude,
                   end_latitude):
    print("Cube before intersection", cube)
    print("Ancil variables", cube.ancillary_variables())
    print("Cell measures", cube.cell_measures())
    region_subset = cube.intersection(
        longitude=(start_longitude, end_longitude),
        latitude=(start_latitude, end_latitude),
        ignore_bounds=True,
    )
    print("Cube after intersection", region_subset)


# intersection cube loses cell measures and ancillary variables
extract_region(cube,
               start_longitude=-90,
               end_longitude=40,
               start_latitude=20,
               end_latitude=80
    )

# intersection cube doesn't lose cell measures and ancillary variables
extract_region(cube,
               start_longitude=160,
               end_longitude=280,
               start_latitude=-5,
               end_latitude=5
    )
```

`cube_to_intersect_smol.nc` attached here (rename to .nc, size=1.2M, smol like I says) 
[cube_to_intersect_smol.nc.txt](https://github.com/SciTools/iris/files/12261411/cube_to_intersect_smol.nc.txt)

### Have I looked at docs?

Yes https://scitools-iris.readthedocs.io/en/latest/generated/api/iris.cube.html#iris.cube.Cube.intersection no mention of any side effects, coordinates are circular

### Iris version and installation type

```
(esmvaltool) valeriu@valeriu-PORTEGE-Z30-C:~$ conda list iris
# packages in environment at /home/valeriu/miniconda3/envs/esmvaltool:
#
# Name                    Version                   Build  Channel
iris                      3.6.1              pyha770c72_0    conda-forge
iris-esmf-regrid          0.7.0              pyhd8ed1ab_0    conda-forge
```

### AOBs

Related to https://github.com/SciTools/iris/issues/3483

Great many thanks in advance! Many thanks to @rebeccaherman1 for alerting me to this via https://github.com/ESMValGroup/ESMValCore/issues/2162

### Obligatory pint emoji

:beer:

ESadek-MO:
## 🚀 Pull Request

Closes #5413.

To do:
 - [x] Add rough functionality
 - [x] Refactor functions into one consolidated
 - [x] Add tests

### DIFF
--- a/lib/iris/cube.py
+++ b/lib/iris/cube.py
@@ -3182,9 +3182,7 @@ class Cube(CFVariableMixin):
                         dim_within_coord = dims.index(dim)
                         data = np.concatenate(
                             [
-                                get_metadata(
-                                    chunk, src_metadata.name()
-                                ).core_data()
+                                get_metadata(chunk, src_metadata.name()).core_data()
                                 for chunk in chunks
                             ],
                             dim_within_coord,


ESadek-MO:
Two notes on last commit:

- I decided against consolidating `create_coord` into `create_metadata` as it would require so many if checks and the like it would've been harder to read. That said, I'm happy to be vetoed on this. 
- This PR threw up some flake8 errors from previous changes; `cube` was being used as a variable despite being imported as well. All these instances have been changed to `c` instead

pp-mo:
Hi @ESadek-MO
FYI I just assigned this to you..
I was reviewing my Peloton tasks, and this popped up, 
but being in Draft I think it is still on you.

lib/iris/cube.py
```diff
@@ -3175,8 +3176,41 @@ def create_coords(src_coords, add_coord):
                     add_coord(result_coord, dims)
                     coord_mapping[id(src_coord)] = result_coord
 
+            def create_metadata(src_metadatas, add_metadata, metadata_type):
+                if metadata_type == "cell_measure":
+                    metadata_search_object = cube.Cube.cell_measure
+                elif metadata_type == "ancillary_var":
+                    metadata_search_object = cube.Cube.ancillary_variable
+                else:
+                    raise ValueError
```
pp-mo, line 3185: The ValueError probably ought to have some content.  

But then, **I'm thinking we don't really need it** , because ...
I think it would be simpler to just make the "metadata_search_object" another method argument,
instead of using an identifying string.
So ```def create_metadata(src_metadatas, add_metadata, metadata_search_object):```
and then change the calls (see below)

Given that, I think it would _also_ be clearer to rename "metadata_search_object" as "get_metadata"
(or something like), since that matches "add_metadata" : they are both method arguments.

lib/iris/cube.py
```diff
@@ -3175,8 +3176,41 @@ def create_coords(src_coords, add_coord):
...
+                        )
+                        result_coord = src_metadata.copy(values=data)
+                    else:
+                        result_coord = src_metadata.copy()
+                    add_metadata(result_coord, dims)
+
             create_coords(self.dim_coords, result.add_dim_coord)
             create_coords(self.aux_coords, result.add_aux_coord)
+            create_metadata(
+                self.cell_measures(), result.add_cell_measure, "cell_measure"
```
pp-mo, line 3207:
```suggestion
                self.cell_measures(), result.add_cell_measure, Cube.cell_measure
```
as above suggestion

lib/iris/cube.py
```diff
@@ -3175,8 +3176,41 @@ def create_coords(src_coords, add_coord):
...
+            )
+            create_metadata(
+                self.ancillary_variables(),
+                result.add_ancillary_variable,
+                "ancillary_var",
```
pp-mo, line 3212:
```suggestion
                Cube.ancillary_variable,
```

as above suggestion

pp-mo (changes requested):
Some small suggestions which I think make it simpler
(though you may disagree !)

I did also wonder if we should check that the newly constructed cell_measures / ancillaries remain lazy.   But on second thoughts I don't think it's really needed

pp-mo (approved):
Thanks, looks fine now to me

PR #5807, opened 2024-03-06
Lazystream fix 4 unifiedbyblocks

pp-mo:
Addresses #5753 

Another attempt !

### DIFF
--- a/lib/iris/fileformats/netcdf/saver.py
+++ b/lib/iris/fileformats/netcdf/saver.py
@@ -2469,6 +2469,85 @@ class Saver:
                 # Issue a fill-value warning immediately, if appropriate.
                 _fillvalue_report(fill_info, is_masked, contains_fill_value, warn=True)
 
+    @staticmethod
+    def _create_completion_delayed(sources, targets, fill_infos):
+        # Define a combined store-and-fillvalue-check operation, delayed for
+        #  application to lazy data blocks.
+        @dask.delayed
+        def delayed_block_save_and_fv_check(
+            datablock, target_block, block_slices, fillinfo
+        ):
+            # The inner function of a delayed fill-value check
+            # Save the block and return array([is_masked, contains_value]).
+            target_block[block_slices] = datablock
+            is_masked, contains_value = _data_fillvalue_check(
+                np, datablock, fillinfo.check_value
+            )
+            return np.array([is_masked, contains_value])
+
+        # Map the function across the blocks of the input data.
+        # NOTE: the goal is to process the to-be-saved blockwise via a delayed
+        # operation that will return (is_masked, contains_value) for each block.
+        # These block-wise results are then OR-ed for each source.
+        # N.B. `da.map_blocks` apparently will not do this, at least not without
+        #  `drop_axis` which seems to trigger loading of the whole source array.
+        # TODO: It probably *can* be done with `da.blockwise`, but I have so far
+        #  failed to grasp how you would do that.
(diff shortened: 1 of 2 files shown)


lib/iris/fileformats/netcdf/saver.py
```diff
@@ -2490,24 +2490,66 @@ def delayed_completion(self) -> Delayed:
...
+                return np.array([is_masked, contains_value])
+
+            # Map the function across the blocks of the input data.
+            # NOTE: the goal is to process the to-be-saved blockwise via a delayed
+            # operation that will return (is_masked, contains_value) for each block.
+            # These block-wise results are then OR-ed for each source.
+            # N.B. `da.map_blocks` apparently will not do this, at least not without
+            #  `drop_axis` which seems to trigger loading of the whole source array.
+            # TODO: It probably *can* be done with `da.blockwise`, but I have so far
+            #  failed to grasp how you would do that.
```
bouweandela, line 2517: This looks like a [reduction](https://docs.dask.org/en/latest/generated/dask.array.reduction.html), maybe that could be useful?

pp-mo: :heart: :think:

pp-mo:
### Status update 2024-03-08 
The recent change appears to have got all tests passing.

I discovered some top-level Dask docs I hadn't noticed before - https://docs.dask.org/en/stable/best-practices.html#learn-techniques-for-customization
which may somewhat endorse the slightly messy way of doing things here

Here at MetOffice, we do have some good reasons for wanting to get a solution for this into a bugfix release.  There are quite a few outstanding things we'd like to clear up though, at least in an ideal world ...
  * benchmark to see what effect this has on performance more generally
  * add a benchmark to catch "the problem", which would have ben really useful here
  * raise as an issue on Dask, asking if there is "a better way" of getting what we want
  * investigate alternative, possibly preferable, ways as suggested by @bouweandela 
    * [da.reduce](https://github.com/SciTools/iris/pull/5807#discussion_r1515834878)
    * [dask.graph_manipulation.clone](https://github.com/SciTools/iris/issues/5753#issuecomment-1981685054)
  * remove or rework [test_Saver__lazy.TestStreamed](https://github.com/SciTools/iris/blob/2b1ca33e7f9b15ddea17ca903d56529412fad097/lib/iris/tests/unit/fileformats/netcdf/saver/test_Saver__lazy.py#L82-L114), as since we no longer call da.store, this seems to have become a bit pointless

trexfeathers:
The team are spending the next 2 weeks giving our testing infrastructure some TLC.

I we haven't made further progress on this PR by the beginning of April we should look at merging as-is and getting a bugfix release out.

In the meantime we'll have to encourage sufferers to use Iris 3.5.

pp-mo:
> The team are spending the next 2 weeks giving our testing infrastructure some TLC.
> If we haven't made further progress on this PR by the beginning of April we should look at merging as-is and getting a bugfix release out.

I also raised https://github.com/dask/dask/issues/10991
There may be some answers there.
Even if not, I suspect @bouweandela suggestion of using `da.reduce` will probably prove one way to improve on this solution.
In any case, I'm sure we can make this neater some way. I'd love to see a neater solution to [this tiresome bit of code](https://github.com/SciTools/iris/blob/2b1ca33e7f9b15ddea17ca903d56529412fad097/lib/iris/fileformats/netcdf/saver.py#L2500-L2518)

pp-mo:
Replaced by #5833
KISS !

PR #5808, opened 2024-03-11
added pytest iris class

ESadek-MO:


lib/iris/tests/iris_test_pytest.py
```diff
@@ -0,0 +1,1033 @@
...
+    _EXPORT_DATAPATHS_FILE = None
+
+
+if "--create-missing" in sys.argv:
+    sys.argv.remove("--create-missing")
+    print("Allowing creation of missing test results.")
+    os.environ["IRIS_TEST_CREATE_MISSING"] = "true"
+
+
+def main():
```
trexfeathers, line 107: This hasn't been working for a while. So no point in transferring over, but we do need to investigate, so an issue should be raised.

This statement also applies to the arguments referenced e.g. `--data-files-used`

lib/iris/tests/iris_test_pytest.py
trexfeathers, line 1: Please can we bring back the 'skipper' objects, but using the PyTest API rather than the unitest API.

lib/iris/tests/iris_test_pytest.py
```diff
@@ -0,0 +1,1033 @@
+# Copyright Iris contributors
+#
+# This file is part of Iris and is released under the BSD license.
+# See LICENSE in the root of the repository for full licensing details.
+"""Provides testing capabilities and customisations specific to Iris.
+
+.. note:: This module needs to control the matplotlib backend, so it
+          **must** be imported before ``matplotlib.pyplot``.
```
trexfeathers, line 8: Not true anymore

lib/iris/tests/iris_test_pytest.py
```diff
@@ -0,0 +1,1033 @@
...
+
+The primary class for this module is :class:`IrisTest`.
```
trexfeathers, line 10: Not true anymore

lib/iris/tests/iris_test_pytest.py
trexfeathers, line 1: Rename to `_shared_utils.py`.

This should discourage downstream use. People can always ask for things to be made public, and then we can discuss before doing so.

trexfeathers (changes requested):

lib/iris/tests/iris_test_pytest.py
trexfeathers: Turns out it IS working, I had just misunderstood it. Anyway please remove all that stuff from this file.

Here is the issue: #5809

lib/iris/tests/iris_test_pytest.py
```diff
@@ -0,0 +1,1033 @@
...
+    result = os.path.join(
+        self.get_result_path(""),
+        sub_path.replace("test_", ""),
+        self.__class__.__name__.replace("Test_", ""),
+        filename,
+    )
+    return result
+
+
+def assertCMLApproxData(cubes, reference_filename=None, **kwargs):
```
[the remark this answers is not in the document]
trexfeathers: Still need to do `_recordWarningMatches`, `assertNoWarningsRegexp`, `assertLogs`

lib/iris/tests/_shared_utils.py
```diff
@@ -0,0 +1,929 @@
...
+    messages = [str(warning.message) for warning in w]
+    expr = re.compile(expected_regexp)
+    matches.extend(message for message in messages if expr.search(message))
+
+
+@contextlib.contextmanager
+def assertLogs(caplog, logger=None, level=None, msg_regex=None):
+    """Also adds the ``msg_regex`` kwarg:
+    If used, check that the result is a single message of the specified
+    level, and that it matches this regex.
```
trexfeathers, line 516: Docstring reads quite oddly now

lib/iris/tests/_shared_utils.py
```diff
@@ -0,0 +1,929 @@
...
+    cube.regrid, has maintained its behaviour.
+
+    """
+    assert result.shape == shape
+    assert_array_all_close(result.data.mean(), mean, rtol=rtol)
+    assert_array_all_close(result.data.std(), std_dev, rtol=rtol)
+
+
+def assert_dict_equal(lhs, rhs, msg=None):
+    """Dictionary Comparison."""
```
trexfeathers, line 681: The original docstring explained why this was necessary. It would be good to restore that.

lib/iris/tests/_shared_utils.py
```diff
@@ -0,0 +1,929 @@
...
+    #             break
+    # filename = basename + ext
+    #
+    # result = os.path.join(
+    #     get_result_path(""),
+    #     sub_path.replace("test_", ""),
+    #     self.__class__.__name__.replace("Test_", ""),
+    #     filename,
+    # )
+    # return result
```
trexfeathers, line 246: Replace this with a `TODO:`

trexfeathers (approved):

PR #5813, opened 2024-03-11
Convert `tests.unit.concatenate.test_concatenate` to pytest.

pp-mo:


bjlittle:
@pp-mo This pull-request is still in `draft` ... did you want me to hold off reviewing just now?

pp-mo:
> @pp-mo This pull-request is still in `draft` ... did you want me to hold off reviewing just now?

Not now it passes !
Knock yourself out ...

lib/iris/tests/unit/concatenate/test_concatenate.py
```diff
@@ -261,147 +230,156 @@ def test_asc_points(self):
...
-        c1 = self.build_lazy_cube([1, 2], aux_coords=True)
-        c2 = self.build_lazy_cube([3, 4, 5], aux_coords=True)
+    def test_lazy_concatenate(self, sample_lazy_cubes):
+        (cube,) = concatenate(sample_lazy_cubes)
+        assert cube.has_lazy_data()
+        assert not ma.isMaskedArray(cube.data)
+
+    def test_lazy_concatenate_aux_coords(self, sample_lazy_cubes):
+        c1, c2 = sample_lazy_cubes
+        c2 = self.build_lazy_cube([3, 4, 5])
```
bjlittle, line 366: @pp-mo Why the need to call `build_lazy_cube` here?

Isn't this the same as `sample_lazy_cubes[1]` ?

bjlittle (changes requested):
@pp-mo Lookin' good! 💯 

Just a simple query to service then we can bank this 👍 

Nice one!

lib/iris/tests/unit/concatenate/test_concatenate.py
pp-mo: NO, I think I was missing something obvious!
I guess just because I didn't initially equate these with the "standard setup", due to use of "aux_coords=True" in the original.

pp-mo: And in that case, "build_lazy_cubes" doesn't need to be provided separately at all ...

pp-mo: https://github.com/SciTools/iris/pull/5813/commits/2427de8a3b93c22fa95ee4b5c1a67b42e2f8a560

bjlittle (approved):

bjlittle:
@pp-mo Awesome, cheers 🍻

PR #5822, opened 2024-03-12
pytest migration for unit.common.test_Lenient

bjlittle:
### Description

---

---

### DIFF
--- a/lib/iris/tests/unit/common/lenient/test_Lenient.py
+++ b/lib/iris/tests/unit/common/lenient/test_Lenient.py
@@ -28,7 +28,7 @@ class Test___init__:
     def test_kwargs(self, lenient):
         actual = Lenient(maths=False)
         expected = dict(maths=False)
-        assert actual.__dict__, expected
+        assert actual.__dict__ == expected
 
     def test_kwargs_invalid(self, lenient):
         emsg = "Invalid .* option, got 'merge'."
@@ -46,7 +46,7 @@ class Test___contains__:
 
 class Test___getitem__:
     def test_in(self, lenient):
-        assert lenient["maths"] is True
+        assert bool(lenient["maths"]) is True
 
     def test_not_in(self, lenient):
         emsg = "Invalid .* option, got 'MATHS'."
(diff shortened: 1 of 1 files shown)


bjlittle:
Interesting ... `pytest-xdist` is exposing failures elsewhere within CI :thinking:

lib/iris/tests/unit/common/lenient/test_Lenient.py
```diff
@@ -4,179 +4,166 @@
...
 
+@pytest.fixture()
+def lenient():
+    # setup
+    state = {key: _LENIENT.__dict__[key] for key in _LENIENT_PROTECTED}
+    # call
+    yield Lenient()
+    # teardown
+    for key, value in state.items():
+        _LENIENT.__dict__[key] = value
```
bjlittle, line 20: @pp-mo As I suspected ... `unittest` gives us this `setup` and `teardown` behaviour for free.

Very interesting to know and be aware of ... for reference see https://docs.pytest.org/en/latest/how-to/fixtures.html#yield-fixtures-recommended

lib/iris/tests/unit/common/lenient/test_Lenient.py
```diff
@@ -4,179 +4,166 @@
...
 
-class Test___init__(tests.IrisTest):
-    def test_default(self):
-        lenient = Lenient()
+
+class Test___init__:
+    def test_default(self, lenient):
         expected = dict(maths=True)
-        self.assertEqual(expected, lenient.__dict__)
+        assert expected == lenient.__dict__
```
pp-mo: I think ideally this should go the other way around.
Ideally, all the pytest style equality checks should by "assert actual == expected" and not "expected == actual".
I believe this matters only because if it fails, in some cases the message could be a bit incomprehensible if the order is wrong.

N.B. the existing "way around" **_is_** correct for unittest , as in "assertEqual(expected, ...)".  Personally I always found that a bit non-obvious anyway.

Please treat as a general point : I'm not going to stick comments everywhere just for this !

lib/iris/tests/unit/common/lenient/test_Lenient.py
```diff
@@ -4,179 +4,166 @@
...
-        self.assertTrue(_LENIENT.enable)
-        self.lenient["maths"] = True
-        self.assertTrue(self.lenient.__dict__["maths"])
-        self.assertTrue(_LENIENT.enable)
-
-    def test_maths_enable__lenient_enable_false(self):
+        assert _LENIENT.enable is False
+        lenient["maths"] = False
+        assert lenient.__dict__["maths"] is False
+        assert _LENIENT.enable is False
```
pp-mo, line 86: Are you happy that the testing for the exact values False/True here is the best expression of what was intended ?

I.E. I assume you want to ensure that the value is actually a "True", and not a 1 or "x" or anything else for which "if x" or bool(x) counts true. And similar for the "is False" tests.

If that is **_not_** so, then I think the style here is not correct : we should just be using "assert x" / "assert not x" instead.

The problem is, I think in _unittest_ language you would commonly do assertTrue/assertFalse for "assert x"/"assert not x" anyway.
But if it _really_ meant the same as what is here, that should probably be an "assertIs".

Again, that applies to a whole lot of cases here !

pp-mo (changes requested):
One definite bug, and some possible style queries to look into..

pp-mo:
Sorry we just crossed over.
Please let me know when you're ready for me to re-check.

bjlittle:
@pp-mo Done, thanks ... back to you 👍

lib/iris/tests/unit/common/lenient/test_Lenient.py
```diff
@@ -4,179 +4,166 @@
...
-        self.assertEqual(expected, lenient.__dict__)
+        assert expected == lenient.__dict__
 
-    def test_kwargs(self):
-        lenient = Lenient(maths=False)
+    def test_kwargs(self, lenient):
+        actual = Lenient(maths=False)
         expected = dict(maths=False)
-        self.assertEqual(expected, lenient.__dict__)
+        assert expected, actual.__dict__
```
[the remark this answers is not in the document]
bjlittle: Oopsie 😊

pp-mo:
OK, good stuff!

PR #5823, opened 2024-03-12
pytest migration ruff PT compliance for unit.concatenate

bjlittle:
### Description

@pp-mo I've updated your changes so that they're `ruff check --select PT` ([flake8-pytest-style](https://docs.astral.sh/ruff/rules/#flake8-pytest-style-pt)) compliant.

---

---


pp-mo:
A wee bit of a downer, because when I spotted that I'd added the empty parentheses then I explicitly removed them !

But certainly can't argue with the standards + consistency message.

PR #5824, opened 2024-03-12
Feature pytest conversion

pp-mo:
For now, just to get a docs build with the pytest style guide from https://github.com/SciTools/iris/pull/5785


pp-mo:
Now dead

PR #5827, opened 2024-03-13
Better implementation of `_shared_utils.results_path()` - using PyTest `request`

trexfeathers:
The previous implementation was flaky given we could not know where in the call stack the test actually was, and I have no idea how I could have got it working with PyTest parameters - probably something even more flaky.

This version uses publicly available attributes from the PyTest [`FixtureRequest`](https://docs.pytest.org/en/stable/reference/reference.html#pytest.FixtureRequest), which seems much safer.

### DIFF
--- a/lib/iris/tests/_shared_utils.py
+++ b/lib/iris/tests/_shared_utils.py
@@ -11,7 +11,6 @@ import difflib
 import filecmp
 import functools
 import gzip
-import inspect
 import json
 import math
 import os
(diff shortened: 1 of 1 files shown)


lib/iris/tests/_shared_utils.py
```diff
@@ -207,17 +206,36 @@ def get_result_path(relative_path):
...
+        )
+        raise ValueError(message)
+
+
+def result_path(request: pytest.FixtureRequest, basename=None, ext=""):
     """Generate the path to a test result; from the calling file, class, method.
 
     Parameters
     ----------
+    request : pytest.FixtureRequest
```
ESadek-MO, line 228: Perhaps make it clear that all it needs to have is the word request, and pytest does the rest?

ESadek-MO (changes requested):
LGTM, thanks @trexfeathers! 

Just a small suggestion, but otherwise awesome!

lib/iris/tests/_shared_utils.py
trexfeathers: 6829c759d9017499dca3fd2bc185a2aaa83a8e02

ESadek-MO (approved):
Awesome, thanks!

PR #5839, opened 2024-03-14
Convert `tests/test_plot` to PyTest

trexfeathers:
`ruff check --select PT` ([flake8-pytest-style](https://docs.astral.sh/ruff/rules/#flake8-pytest-style-pt)) compliant

### DIFF
--- a/lib/iris/tests/test_plot.py
+++ b/lib/iris/tests/test_plot.py
@@ -3,10 +3,6 @@
 # This file is part of Iris and is released under the BSD license.
 # See LICENSE in the root of the repository for full licensing details.
 
-# import iris tests first so that some things can be initialised before
-# importing anything else
-import iris.tests as tests  # isort:skip
-
 from contextlib import nullcontext
 
 import cf_units
@@ -20,7 +16,7 @@ from iris.tests import _shared_utils
 import iris.tests.stock
 
 # Run tests in no graphics mode if matplotlib is not available.
-if tests.MPL_AVAILABLE:
+if _shared_utils.MPL_AVAILABLE:
     import matplotlib.pyplot as plt
 
     import iris.plot as iplt
@@ -80,7 +76,7 @@ class TestMissingCoord(_shared_utils.GraphicsTest):
 class TestMissingCS(_shared_utils.GraphicsTest):
     @_shared_utils.skip_data
     def test_missing_cs(self):
-        cube = tests.stock.simple_pp()
+        cube = iris.tests.stock.simple_pp()
         cube.coord("latitude").coord_system = None
         cube.coord("longitude").coord_system = None
         qplt.contourf(cube)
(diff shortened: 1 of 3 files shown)


lib/iris/tests/test_plot.py
```diff
@@ -760,46 +695,26 @@ def test_t_dates(self):
...
-    key = (filename, str(constraint))
-    cube = _load_cube_once_cache.get(key, None)
-
-    if cube is None:
-        cube = iris.load_cube(filename, constraint)
-        _load_cube_once_cache[key] = cube
-
-    return cube
-
-
```
trexfeathers, line 802: See:

https://github.com/SciTools/iris/blob/efa00207bb3ba2cd21a748e6748a09cddf4bb981/lib/iris/tests/test_plot.py#L735-L739

lib/iris/tests/test_plot.py
```diff
@@ -131,23 +127,25 @@ def test_orography(self):
...
 
         # TODO: Test bounds once they are supported.
-        with self.assertRaises(NotImplementedError):
-            qplt.pcolor(self.cube)
+        qplt.pcolor(self.cube)
+        with pytest.raises(NotImplementedError):
             iplt.orography_at_bounds(self.cube)
-            iplt.outline(self.cube)
-            self.check_graphic()
+            # iplt.outline(self.cube)
```
ESadek-MO, line 133: commented code

ESadek-MO (changes requested):
Thanks @trexfeathers. Only a couple of queries, otherwise looks great!

lib/iris/tests/test_plot.py
```diff
@@ -339,16 +334,19 @@ def test_circular_changes(self):
...
-@tests.skip_data
-@tests.skip_plot
-class Test1dFillBetween(tests.GraphicsTest):
-    def setUp(self):
-        super().setUp()
+@_shared_utils.skip_data
+@_shared_utils.skip_plot
+class Test1dFillBetween(_shared_utils.GraphicsTest):
+    @pytest.fixture(autouse=True)
+    def _set_draw_method(self):
```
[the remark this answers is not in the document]
trexfeathers: This class is inherited. The only difference in the inheriting class is `self.draw_method`. I was therefore able to separate this into its own fixture so that the rest of `_setup()` didn't need to be re-written in the inheriting class.

https://github.com/SciTools/iris/blob/098466411b26e8c539028569b55686fea1ad607c/lib/iris/tests/test_plot.py#L428-L433

BTW this is what I meant when I said I had seen the benefits of using fixtures rather than `setup_method()`.

lib/iris/tests/test_plot.py
trexfeathers: I believe this commented code is here to signify how the test should look once the `# TODO` has been addressed.

https://github.com/SciTools/iris/blob/098466411b26e8c539028569b55686fea1ad607c/lib/iris/tests/test_plot.py#L129

Is there a better format that you would prefer?

lib/iris/tests/test_plot.py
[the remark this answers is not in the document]
ESadek-MO: Awesome, happy with that.

ESadek-MO (approved):
Consider me rebutted! LGTM, thanks @trexfeathers

PR #5840, opened 2024-03-14
Pytest convert the remainder of the aux-factory test

pp-mo:
That is, all the "Ocean" ones.


bjlittle (approved):
@pp-mo Lovely work! Thanks 💯

PR #5845, opened 2024-03-15
pytest migration for unit.common.lenient.test__qualname

bjlittle:
### Description

---

---

### DIFF
--- a/lib/iris/tests/unit/common/lenient/test__qualname.py
+++ b/lib/iris/tests/unit/common/lenient/test__qualname.py
@@ -6,14 +6,11 @@
 
 from inspect import getmodule
 
-import pytest
-
 from iris.common.lenient import _qualname
 
 
 class Test:
-    @pytest.fixture(autouse=True)
-    def _setup(self):
+    def setup_method(self):
         module_name = getmodule(self).__name__
         self.locals = f"{module_name}" + ".Test.{}.<locals>.{}"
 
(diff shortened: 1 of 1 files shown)


lib/iris/tests/unit/common/lenient/test__qualname.py
```diff
@@ -4,39 +4,34 @@
...
 from inspect import getmodule
-from unittest.mock import sentinel
 
 from iris.common.lenient import _qualname
 
 
-class Test(tests.IrisTest):
-    def setUp(self):
+class Test:
+    def setup_method(self):
```
pp-mo, line 13: I only just realised that this **_does_** work, and why pytestify does it this way.
The docs are rather quiet about it, but it is covered [here](https://docs.pytest.org/en/7.1.x/how-to/xunit_setup.html?highlight=setup_method#method-and-function-level-setup-teardown)

Nevertheless., I think I still prefer `@fixture(autouse=True)`, and maybe we should standardise on that ??
In particular, I think a "setup_method" is **_not_** a fixture, and is not called in quite the same way, and it can't reference another fixture, e.g. mocker for a 'mocker.patch'.
```suggestion
    @pytest.fixture(auto_use=True)
    def _setup(self):
```

pp-mo (approved):
One detail to consider,
OWOK!

lib/iris/tests/unit/common/lenient/test__qualname.py
bjlittle: Yup, all true.

Perhaps worth a sprint discussion?

I've erred on the side of using a `fixture(autouse=True)` when I don't need to use the `mocker` fixture in the `setup`, otherwise default to the `pytest` convenience of `setup_method`, which like you mention isn't a fixture.

I'm happy either way

PR #5853, opened 2024-03-18
Convert tests in unit/util to pytest

### LINKED ISSUE #5855: Convert unit/util test files to pytest

ESadek-MO:
closes #5855


lib/iris/tests/unit/util/test_equalise_attributes.py
```diff
@@ -66,21 +65,21 @@ def _test(self, cubes, expect_attributes, expect_removed):
...
         # Check resulting attributes all match the expected set
         for cube in working_cubes:
-            self.assertEqual(cube.attributes, expect_attributes)
+            assert cube.attributes == expect_attributes
         # Check removed attributes all match as expected
-        self.assertEqual(len(actual_removed), len(expect_removed))
+        assert len(actual_removed) == len(expect_removed)
         for actual, expect in zip(actual_removed, expect_removed):
-            self.assertEqual(actual, expect)
+            assert all(actual) == all(expect)
```
trexfeathers, line 75: I assume you added `all()` to get it to shut up about comparing iterables. But this isn't doing what you think - it's just checking that everything in `actual`/`expect` is truthy:

```python
>>> all(["a", 1, 2])
True
>>> all(["a", 1, 0])
False
```

Basically: the fact that this is always returning `True` for this test is not evidence that the code is working.

The correct change will depend on what `actual` and `expect` actually are.

lib/iris/tests/unit/util/test_file_is_newer_than.py
```diff
@@ -44,7 +42,7 @@ def setUp(self):
             mtime += 5.0 + 10.0 * i_file
             os.utime(file_path, (mtime, mtime))
 
-    def tearDown(self):
+    def teardown_method(self):
```
trexfeathers, line 45: You can use the `setup -> yield -> teardown` style all within one fixture.

But in this case there is a built-in mechanism where you won't need a teardown at all:

[How to use temporary directories and files in tests — pytest documentation](https://docs.pytest.org/en/stable/how-to/tmp_path.html)

lib/iris/tests/unit/util/test_mask_cube.py
```diff
@@ -29,24 +26,27 @@ def full2d_global():
 
 
 class MaskCubeMixin:
-    def assertOriginalMetadata(self, cube, func):
+    def assertOriginalMetadata(self, request, cube, func):
```
trexfeathers, line 29: Should probably change the syntax of this name as it currently looks like unittest

lib/iris/tests/unit/util/test_mask_cube.py
```diff
@@ -29,24 +26,27 @@ def full2d_global():
...
         """Check metadata matches that of input cube.  func is a string indicating
         which function created the original cube.
 
         """
         reference_dir = pathlib.Path("unit/util/mask_cube")
         reference_fname = reference_dir / f"original_cube_{func}.cml"
-        self.assertCML(
+        _shared_utils.assert_CML(
+            request,
```
trexfeathers, line 37:
```suggestion
            self.request,
```

Since this is a method it should always be able to access `self.request` (or error if it has not been set).

That would also simplify all the calls to the method since you don't need `request` as an argument.

trexfeathers (changes requested):
Superb job, especially with so many files. Some useful patterns too.

A few pieces to change

lib/iris/tests/unit/util/test_equalise_attributes.py
ESadek-MO: Yeah that was a brain fart, apologies! They seem to be either a dict or a list, which gives me two different cases to handle is all

lib/iris/tests/unit/util/test_mask_cube.py
trexfeathers: The failing tests are just because you're still passing `self.request` in various calls and you don't need to anymore.

trexfeathers (approved):

PR #5858, opened 2024-03-18
`v3.8.1` merge-back

trexfeathers:
Does not need to be a merge-commit since this is a 'manual' recreation of what the merge-back should be (no features are exclusive to the release branch, only What's New content).


bjlittle (approved):
@trexfeathers Thanks! 💯

PR #5859, opened 2024-03-18
Make test_analysis PyTest

trexfeathers:

### DIFF
--- a/lib/iris/tests/test_analysis_calculus.py
+++ b/lib/iris/tests/test_analysis_calculus.py
@@ -3,12 +3,8 @@
 # This file is part of Iris and is released under the BSD license.
 # See LICENSE in the root of the repository for full licensing details.
 
-# import iris tests first so that some things can be initialised before importing anything else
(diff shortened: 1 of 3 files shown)


lib/iris/tests/test_analysis_calculus.py
```diff
@@ -733,19 +760,22 @@ def test_contrived_spherical_curl2(self):
         np.testing.assert_array_almost_equal(
```
ESadek-MO, line 760: I've been tending to convert these calls to _shared_utils for sake of consistency.

ESadek-MO (changes requested):
Only thing I could see was the use of np.testing functions, otherwise LGTM!

lib/iris/tests/test_analysis_calculus.py
trexfeathers: bcaca28a6a14cf62f30e034db5b6aa02e1029303

ESadek-MO (approved):
thanks @trexfeathers

PR #5862, opened 2024-03-19
Convert ABF test to pytest.

pp-mo:
In this case I went a bit beyond the brief, and have converted the roll-your-own call counting to a reasonably standard approach using Mock.
Which is not strictly a pytest change (!)

**NOTE:**
@bjlittle suggested using "mocker.spy" instead, which does sound like the nicest way.
However, I found that "mocker.spy" does not work on a `__getattr__` call, even though "mock.patch" actually did.
So it looks like the implementation of "mocker.spy" is less obvious than you might expect (probably since it now it supports async methods, etc).

For the sake of simplicity, I just removed the check that `__getattr__` is called :   I think checking that `_read` was called is good enough in itself.

### DIFF
--- a/lib/iris/tests/unit/fileformats/abf/test_ABFField.py
+++ b/lib/iris/tests/unit/fileformats/abf/test_ABFField.py
@@ -11,15 +11,14 @@ class Test_data:
         path = "0000000000000000jan00000"
         field = ABFField(path)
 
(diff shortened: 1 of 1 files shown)


bjlittle (approved):
@pp-mo Pleasing to see this code melt away.

Nice one 👍

PR #5863, opened 2024-03-19
Convert unit/fileformats/__init__

pp-mo:
Move this code into pp_load_rules where it belongs.
While converting just these test routines into assertion routines

Because it's simpler to do that **_first_**, before converting all the pp_load_rules tests themselves.

***NOTE:*** 
I subsequently discovered that one of these 2 routines is specific to a single test, so it [can go there](https://github.com/SciTools/iris/pull/5865/files#diff-cf81a5dc388b8df78c9dacbe64e1d1077f9a7c7542a123594695c716abb9d5e8R19-R21).
But we may as well do it this way this first, anyway.


bjlittle (approved):
@pp-mo I understand the need for reshuffle here, thanks.

PR #5866, opened 2024-03-19
Convert `unit.plot` and `.quickplot` to PyTest

trexfeathers:


ESadek-MO (approved):
Thanks @trexfeathers.

All seems to be mostly boilerplate, and a quick scan can't find anything else I'd chnage, LGTM!

PR #5876, opened 2024-03-20
pytest migration of unit.common.metadata.test_metadata_manager_factory

bjlittle:
### Description

---

---

### DIFF
--- a/lib/iris/tests/unit/common/metadata/test_metadata_manager_factory.py
+++ b/lib/iris/tests/unit/common/metadata/test_metadata_manager_factory.py
@@ -177,7 +177,7 @@ class Test_instance__fields:
             metadata = metadata_manager_factory(base)
             assert metadata.fields == fields
             for field in fields:
-                hasattr(metadata, field)
+                assert hasattr(metadata, field)
 
 
 class Test_instance__values:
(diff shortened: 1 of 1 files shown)


lib/iris/tests/unit/common/metadata/test_metadata_manager_factory.py
```diff
@@ -156,40 +156,38 @@ def setUp(self):
...
         self.bases = BASES
 
     def test(self):
         for base in self.bases:
             fields = base._fields
             metadata = metadata_manager_factory(base)
-            self.assertEqual(metadata.fields, fields)
+            assert metadata.fields == fields
             for field in fields:
                 hasattr(metadata, field)
```
pp-mo, line 180: Not a pytest issue, but I can't help being puzzled by this "hasattr" bit.
Is it a mistake, or does it merely test for "does not crash" ?

pp-mo (approved):
One silly question, otherwise all good :+1:

lib/iris/tests/unit/common/metadata/test_metadata_manager_factory.py
bjlittle: I guess it's a pattern that I used to ensure that the fields are genuinely accessible.

We could add an `assert` here

bjlittle: See lines +62 and +64 ... yup, you've spotted a genuine omission 👍

PR #5883, opened 2024-03-22
More root pytest

trexfeathers:

### DIFF
--- a/lib/iris/tests/test_pp_stash.py
+++ b/lib/iris/tests/test_pp_stash.py
@@ -3,104 +3,93 @@
 # This file is part of Iris and is released under the BSD license.
 # See LICENSE in the root of the repository for full licensing details.
 
-# import iris tests first so that some things can be initialised before importing anything else
-import iris.tests as tests  # isort:skip
+import pytest
 
 import iris
 import iris.fileformats.pp
 import iris.io
+from iris.tests import _shared_utils
 import iris.tests.stock
 import iris.util
 
 
-class TestPPStash(tests.IrisTest):
-    @tests.skip_data
+class TestPPStash:
+    @_shared_utils.skip_data
     def test_cube_attributes(self):
-        cube = tests.stock.simple_pp()
-        self.assertEqual("m01s16i203", cube.attributes["STASH"])
-        self.assertNotEqual("m01s16i999", cube.attributes["STASH"])
-        self.assertEqual(cube.attributes["STASH"], "m01s16i203")
-        self.assertNotEqual(cube.attributes["STASH"], "m01s16i999")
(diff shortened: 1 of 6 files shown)


lib/iris/tests/test_pp_stash.py
```diff
@@ -3,104 +3,93 @@
...
     def test_ppfield(self):
-        data_path = tests.get_data_path(("PP", "simple_pp", "global.pp"))
+        data_path = _shared_utils.get_data_path(("PP", "simple_pp", "global.pp"))
         pps = iris.fileformats.pp.load(data_path)
         for pp in pps:
-            self.assertEqual("m01s16i203", pp.stash)
-            self.assertNotEqual("m01s16i999", pp.stash)
-            self.assertEqual(pp.stash, "m01s16i203")
-            self.assertNotEqual(pp.stash, "m01s16i999")
+            assert "m01s16i203" == pp.stash
```
ESadek-MO, line 30: I'm confused as to why this needs to happen twice?

lib/iris/tests/test_util.py
```diff
@@ -90,43 +89,36 @@ def test_oversize_string(self):
...
         # Test with a clip length that is longer than the string
 
         clip_length = 10999
         result = iris.util.clip_string(self.test_string, clip_length, self.rider)
-        self.assertEqual(
-            len(result),
-            len(self.test_string),
-            "String was clipped when it should not have been.",
-        )
+        assert len(result) == len(
```
ESadek-MO, line 106: Most asserts with a message attached are wrapped in brackets, might be worth keeping it consistent (unless I've missed something)!

lib/iris/tests/test_util.py
```diff
@@ -90,43 +89,36 @@ def test_oversize_string(self):
...
     def test_invalid_clip_lengths(self):
         # Clip values less than or equal to zero are not valid
         for clip_length in [0, -100]:
             result = iris.util.clip_string(self.test_string, clip_length, self.rider)
-            self.assertEqual(
-                len(result),
-                len(self.test_string),
-                "String was clipped when it should not have been.",
-            )
+            assert len(result) == len(
```
ESadek-MO, line 119: bracket wrap here also

ESadek-MO (changes requested):
Thanks @trexfeathers. A couple comments it'd be good to have your thoughts on, but otherwise LGTM!

lib/iris/tests/test_pp_stash.py
ESadek-MO: Oh that hadn't even occurred to me. Cool

ESadek-MO (approved):
Thanks @trexfeathers, assume all tests pass happy to merge this in!

PR #5894, opened 2024-04-02
nep29: drop py39 and support py312

bjlittle:
### Description

---

---


bjlittle:
As this pull-request is being automatically benchmarked, I'll add an associated `whatsnew` entry in a follow-up pull-request :+1:

lib/iris/tests/unit/experimental/ugrid/load/test_load_mesh.py
```diff
@@ -24,7 +24,8 @@ def test_calls_load_meshes(self):
         args = [("file_1", "file_2"), "my_var_name"]
         with PARSE_UGRID_ON_LOAD.context():
             _ = load_mesh(args)
-        self.assertTrue(self.load_meshes_mock.called_with(args))
+        assert self.load_meshes_mock.call_count == 1
+        assert self.load_meshes_mock.call_args == ((args, None),)
```
bjlittle, line 28: The `called_with` isn't playing ball in `py312` for some reason.

Didn't investigate any deeper, instead used an alternative pattern.

rcomer: I don’t get how that ever worked 😕 there is `assert_called_with` or `assert_called_once_with`
https://docs.python.org/3.12/library/unittest.mock.html#unittest.mock.Mock.assert_called_with

bjlittle: Me either 🤷

bjlittle:
As this pull-request bumps our python support to include `py312` the benchmarking task will fail, as there is no equivalent `py312` on `main` to compare with.

After discussing options with @trexfeathers, we've added a temporary workaround to force the benchmarking to complete successfully so that we can review the final performance report, see [here](https://github.com/SciTools/iris/pull/5894#issuecomment-2032411468).

We're happy to see that there are many performance wins using `py312` 🔥🚀🤩 

I'm now going to back out the workaround, which will result in the benchmarking failing. This is expected, and it's acceptable to merge this pull-request with failing benchmarking.

pyproject.toml
```diff
@@ -73,7 +73,6 @@ src = [
     "lib",
     "docs/src",
 ]
-target-version = "py39"
```
trexfeathers, line 76: I'm in favour of keeping a `target-version`. I do occasionally get burned by things that are not compatible with all our supported Python versions.

Any particular reason you would like to remove it?

If we do keep it then it would be another thing that should be included here:

https://github.com/SciTools/iris/blob/44bc4b856a1582c0f6a8bcf15d0ea6fa4df13f42/lib/iris/tests/test_coding_standards.py#L66-L71

trexfeathers (commented):
Thanks @bjlittle, I'm happy that you got 'em all.

One question below

pyproject.toml
bjlittle: It's duplicate metadata that's not necessary due to the `requires-python` clause in the `pyproject.toml` file. `ruff` doesn't actually need it.

In fact, `repo-review` insists on removing it ... so it's something that we're going to remove now or later if we want to be compliant.

trexfeathers: My bad for not knowing that

trexfeathers (approved):

Issue #5947, opened 2024-05-09
Investigate tracemalloc


pp-mo:
@pp-mo has some ideas on this

End goal: replace existing Iris memory benchmarking with a tracemalloc approach, hopefully as a basis for almost all SciTools memory benchmarking in future

Brief team discussion Thu 9th afternoon

pp-mo:
See https://github.com/SciTools/iris/pull/5946 for investigation

PR #5946, opened 2024-05-09
Trial tracemalloc as memory tracker (replacement v2 PR)

pp-mo:
# ***WIP DO NOT MERGE THIS, EVER***

Some modified code for memory benchmarks, evaluating "tracemalloc" against our existing Linux process-RSS measuring technique.

### Additional context : sample output from a desktop run
```
$ asv run -b Memcheck* --quick -e -E existing:/tmp/persistent/newconda-envs/iris3/bin/python 
Couldn't load asv.plugins._mamba_helpers because
No module named 'libmambapy'
· Ignoring ASV setting(s): `python`. Benchmark environment management is delegated to third party script(s).
· Discovering benchmarks
· Running 4 total benchmarks (1 commits * 1 environments * 4 benchmarks)
[ 0.00%] ·· Benchmarking existing-py_tmp_persistent_newconda-envs_iris3_bin_python
[ 0.00%] ··· Importing benchmark suite produced output:
[ 0.00%] ···· /net/home/h05/itpp/git/iris/iris_main/benchmarks/benchmarks/generate_data/__init__.py:52: UserWarning: No BENCHMARK_DATA env var, defaulting to /net/home/h05/itpp/git/iris/iris_main/benchmarks/.data. Note that some benchmark files are GB in size.
              warn(message)

[12.50%] ··· memtrace_evaluation.MemcheckBlocksAndWorkers.time_calc                                                                                                                                 ok
[12.50%] ··· ========= ========== ========= =========
             --                   nworkers           
             --------- ------------------------------
              nblocks      1          4         9    
             ========= ========== ========= =========
                 1      568±0ms    543±0ms   514±0ms 
                 4      144±0ms    181±0ms   151±0ms 
                 9      55.9±0ms   116±0ms   109±0ms 
             ========= ========== ========= =========

[25.00%] ··· memtrace_evaluation.MemcheckBlocksAndWorkers.track_addmem_calc                                                                                                                         ok
[25.00%] ··· ========= ======= ======= =======
             --                nworkers       
             --------- -----------------------
              nblocks     1       4       9   
             ========= ======= ======= =======
                 1      305.3   305.3   305.3 
                 4       19.2    76.4    76.4 
                 9       4.0     15.2    33.9 
             ========= ======= ======= =======

[37.50%] ··· memtrace_evaluation.MemcheckRunstyles.time_calc                                                                                                                                        ok
[37.50%] ··· ============= ================= ================= =================== ===================
             --                                          runtype / ysize                              
[...]

PR #5948, opened 2024-05-09
Use tracemalloc for memory measurements.

pp-mo:
Addresses https://github.com/SciTools/iris/issues/5947

Replaces the current process RSS ("resident set size") based mechanism.
See #5946 for info on how they compare.

N.B. this is hard to test, possibly only takes effect when merged ?
@trexfeathers can you advise? I'm a bit vague on this.


benchmarks/benchmarks/__init__.py
```diff
@@ -63,27 +65,32 @@ class TrackAddedMemoryAllocation:
...
     def __enter__(self):
-        self.mb_before = self.process_resident_memory_mb()
+        tracemalloc.start()
         return self
 
     def __exit__(self, *_):
-        self.mb_after = self.process_resident_memory_mb()
+        _, peak_mem = tracemalloc.get_traced_memory()
+        tracemalloc.stop()
+        self._peak_mb = peak_mem * 1.0 / 1024**2
```
stephenworsley, line 86: Would this be clearer as `peak_mem * 2**-20`?

stephenworsley (changes requested):

stephenworsley (approved):
After running the benchmarks in #5946, I'm satisfied that tracemalloc gives a more accurate view of memory for more cases than the current set up. It seems like there are still some potential difficulties with memory benchmarking for some parallelisation setups, but I don't believe any more problems are being introduced. Otherwise, I think this PR looks good to go.

pp-mo:
Thanks @stephenworsley !

PR #5951, opened 2024-05-13
Provide a Nox `benchmarks` session as the recommended entry point

### LINKED ISSUE #5943: Make Nox the entry point for benchmarks
## 📰 Custom Issue

Nox (or Tox in future) is supposed to be the 1-stop shop.

We could keep all the logic in `bm_runner.py`, but have a Nox entry point that just called `bm_runner.py`.

Need to think about how to keep the useful `--help` content.

trexfeathers:
### Description

Closes #5943 

---

---


noxfile.py
```diff
@@ -291,3 +291,27 @@ def wheel(session: nox.sessions.Session):
         "import iris; print(f'{iris.__version__=}')",
         external=True,
     )
+
+
+@nox.session
+def benchmarks(session: nox.sessions.Session):
+    """Run the Iris benchmark runner. Run session with `-- -h` for help.
```
stephenworsley, line 298: I appreciate that both work, but since everywhere else in the documentation uses `-- --help`, it seems slightly inconsistent to use `-- -h` here.

stephenworsley (changes requested):
Looks good, just one very minor nitpick about documentation.

noxfile.py
trexfeathers:
```suggestion
    """Run the Iris benchmark runner. Run session with `-- --help` for help.
```

PR #5955, opened 2024-05-15
Partial collapse of multi-dim string coords: take 2

### LINKED ISSUE #3653: cube.collapsed fails with multi-dimensional string coordinates

rcomer:
### Description

This is the same bugfix as #4294.  In there, I also tried to add type hints following my interpretation of [the dev guide](https://scitools-iris.readthedocs.io/en/stable/developers_guide/contributing_code_formatting.html#type-hinting).  This turned out to be [complicated](https://github.com/SciTools/iris/discussions/4383#discussioncomment-1913397) and I suspect is the main reason the PR has not made progress.

I think it would be good to fix the bug regardless of typing questions, so here is a version without the type hints which I'm hoping will be an easier review.

Fixes #3653.

If you have a 2D string coordinate on a cube, and collapse one of the dimensions spanned by that coordinate, the coordinate's points (and bounds, if any) are now serialised along the collapsed dimension. This seems to have been missed when partial collapse was implemented for the numeric coordinates.

---

---


lib/iris/coords.py
```diff
@@ -2114,22 +2114,38 @@ def collapsed(self, dims_to_collapse=None):
...
+
+                return np.array(joined).reshape(out_shape)
 
             bounds = None
             if self.has_bounds():
-                shape = self._bounds_dm.shape[1:]
-                bounds = []
-                for index in np.ndindex(shape):
-                    index_slice = (slice(None),) + tuple(index)
-                    bounds.append(serialize(self.bounds[index_slice]))
```
rcomer, line 2126: If I've understood the old slicing method, it and the new method are consistent for the basic case of 2D bounds (n_points, n_bounds). For higher dimensions, the old method doesn't look right to me: if we had a 2D (n, m) coordinate with bounds (n, m, n_bounds), it looks like we would get m * n_bounds elements in the list. But I think we would have wanted just n_bounds elements.

I didn't even know bounded string coords were a thing!

lib/iris/coords.py
```diff
@@ -2114,22 +2114,38 @@ def collapsed(self, dims_to_collapse=None):
...
+                else:
+                    dims_to_collapse = tuple(
+                        dim % self.ndim for dim in dims_to_collapse
+                    )
+                bounds = serialize(self.bounds, dims_to_collapse)
+
+            points = serialize(self.points, dims_to_collapse)
             # Create the new collapsed coordinate.
-            coord = self.copy(points=np.array(points, dtype=dtype), bounds=bounds)
+            coord = self.copy(points=np.array(points), bounds=bounds)
```
rcomer, line 2148: I don't know why it was necessary to set the `dtype`, but the existing tests pass fine without this now. The old code is 8 years old, so I guess numpy has moved on and is likely doing more for us automatically.

rcomer:
For reference, here is what we currently use for string coordinates in `aggregated_by`, which handles 2D coords but not bounded string coords.
https://github.com/SciTools/iris/blob/95b7ffe5f75cd362fa71eb43644a18d6ae04eb06/lib/iris/analysis/__init__.py#L2437-L2460

lib/iris/coords.py
```diff
@@ -2114,22 +2114,38 @@ def collapsed(self, dims_to_collapse=None):
...
             # Collapse the coordinate by serializing the points and
             # bounds as strings.
-            def serialize(x):
-                return "|".join([str(i) for i in x.flatten()])
+            def serialize(x, axis):
+                if axis is None:
+                    return "|".join(str(i) for i in x.flatten())
+
+                # np.apply_along_axis does not work with str.join, so we
+                # need to loop through the array directly. First move (possibly
```
stephenworsley, line 2122: Is this also true for functions which call `str.join`? i.e.
```python
def np_join(x):
    return "|".join(str(i) for i in x)
```

Which seems to work with `np.apply_along_axis(np_join, 0, arr)` for the arrays I've tested with.

lib/iris/coords.py
```diff
@@ -2114,22 +2114,38 @@ def collapsed(self, dims_to_collapse=None):
         if np.issubdtype(self.dtype, np.str_):
             # Collapse the coordinate by serializing the points and
             # bounds as strings.
-            def serialize(x):
-                return "|".join([str(i) for i in x.flatten()])
+            def serialize(x, axis):
+                if axis is None:
+                    return "|".join(str(i) for i in x.flatten())
```
stephenworsley, line 2119: I think it's worth discussing if it's wise to allow collapsing over multiple dimensions (as supported by `x.flatten()` here). I believe the original case in #3653 only collapses one dimension. One thing worth considering is that with the old behaviour it ough to be possible to reconstruct the original coordinate from the collapsed coordinate, but this is broken when collapsing multiple dimensions. I suppose this boils down to if we interpret `"|"` to imply a positional relation or if we consider it to be "commutative".

If we don't want to handle collapsing multiple dimensions then an appropriate error message would be good here. If we're happy handling collapsing multiple dimensions then I think it would be worth adding a test which demonstrated this behaviour.

lib/iris/coords.py
rcomer: Sorry I should have been clearer about what I meant by "doesn't work".  The problem comes if the output strings would have different length, some of them get truncated.  E.g.

```python
a = np.arange(25).reshape(5, 5)

print(np.apply_along_axis(np_join, 1, a))
```
```
['0|1|2|3|4' '5|6|7|8|9' '10|11|12|' '15|16|17|' '20|21|22|']
```
See [my StackOverflow question](https://stackoverflow.com/questions/52677947/join-strings-along-axis) from when I first tripped over this trying to fix `aggregated_by`.  I will update the comment.

lib/iris/coords.py
rcomer: I think it's preferable to support it even if the output coord is less informative than the single dimension case:  if we raise here then a user has to remove the coordinate before they can perform their collapsing operation, so they will have even less information on the output cube.  They may well not even care about this particular coordinate but just see "I have to remove this coord" as an annoying extra step to get where they want to go.

Happy to add an extra test.

rcomer: Actually I think this is already covered by [test_serialize](https://github.com/SciTools/iris/blob/1bfe3113cb6c1460886ce07dfe495c32782f1aa8/lib/iris/tests/unit/coords/test_Coord.py#L437-L476)?

stephenworsley: It looks like the 2D string arrays in that test are for the bounds (for which only the slices are serialized), the points array being serialized is still 1D. I think there's still room for a more explicit test here.

lib/iris/coords.py
rcomer: https://github.com/SciTools/iris/pull/5955/commits/5d5d6aee8915d5fb2f51bd61bd0281e4beb6a7fe

lib/iris/coords.py
rcomer: https://github.com/SciTools/iris/pull/5955/commits/638d4910ae05b4efa32c36ed4f58ffa2491a9517

stephenworsley (approved):
Looks good.

rcomer:
Thanks @stephenworsley! 🎉

PR #5957, opened 2024-05-17
Add bm_runner 'trialrun' subcommand.

### LINKED ISSUE #5944: Benchmark runner convenience for "can I run X benchmark?"
## 📰 Custom Issue

`python bm_runner.py test_run my_benchmark`

Is something we do a lot when writing a new benchmark.

Should automatically select the various things to speed stuff up like `--quick` and `--attribute rounds=1` etcetera.

@pp-mo suggests forcing the use of an existing environment (`-E` argument), and also making that the `DATA_GEN_PYTHON` environment too. Would be easy enough to pass a path in as a argument or an env variable, could maybe limit it to just the name of a Conda environment?

pp-mo:
Closes #5944

Notes: I tried to make this as foolproof and self-contained as possible. 
so it sets up DATA_GEN_PYTHON and ON_DEMAND_BENCHMARKS.

However you must still provide iris-test-data, via **OVERRIDE_TEST_DATA_REPOSITORY** (at least for most benchmarks).
If you don't, the resulting error is not that obvious (!) -- see below
You may also need to set **BENCHMARK_DATA** if you don't want to overfill your homespace or /tmp/persistent (here at MetOffice).
So, this is potentially still tricky to use, and even dangerous if you total your disk quota!

You may also provide a suitable environment, to prevent it creating one.
Omitting this does work, but can be slow as although we can build a suitable test env, we still don't have a reliable way of automatically caching + re-using them.
**If the "Speed up environment prep" task currently on the board comes good, it might improve that case ?**

### Detail of error report due to missing test-data 
```
$ nox -s benchmarks -- trialrun CubeCreation /tmp/persistent/newconda-envs/iris3/bin/python
nox > Running session benchmarks
nox > Re-using existing virtual environment at .nox/benchmarks.
nox > python -m pip install asv nox
nox > cd /tmp/persistent/git/iris/benchmarks
nox > python bm_runner.py trialrun CubeCreation /tmp/persistent/newconda-envs/iris3/bin/python
Couldn't load asv.plugins._mamba_helpers because
No module named 'libmambapy'
BM_RUNNER DEBUG: Using existing data generation environment.
BM_RUNNER DEBUG: Setting up ASV ...
BM_RUNNER DEBUG: asv machine --yes
Couldn't load asv.plugins._mamba_helpers because
No module named 'libmambapy'
I will now ask you some questions about this machine to identify it in the benchmarks.
  . . .
BM_RUNNER DEBUG: Setup complete.
BM_RUNNER DEBUG: asv run --bench CubeCreation --quick -e --environment existing:/tmp/persistent/newconda-envs/iris3/bin/python
Couldn't load asv.plugins._mamba_helpers because
No module named 'libmambapy'
· Ignoring ASV setting(s): `python`. Benchmark environment management is delegated to third party script(s).
· Discovering benchmarks
· Running 1 total benchmarks (1 commits * 1 environments * 1 benchmarks)
[ 0.00%] ·· Benchmarking existing-py_tmp_persistent_newconda-envs_iris3_bin_python
[50.00%] ··· cube.CubeCreation.time_create                                                                                                                                              failed
[50.00%] ··· =============== ============= ===========
             --                Cube creation strategy 
             --------------- -------------------------
              Cube has mesh   instantiate   construct 
             =============== ============= ===========
                  False          failed       failed  
                   True          failed       failed  
             =============== ============= ===========
             For parameters: False, 'instantiate'

<<multiple tracebacks like this ...>>
              For parameters: True, 'construct'
                File "/tmp/persistent/newconda-envs/iris3/lib/python3.11/site-packages/asv_runner/server.py", line 179, in _run_server
                  _run(run_args)
                File "/tmp/persistent/newconda-envs/iris3/lib/python3.11/site-packages/asv_runner/run.py", line 65, in _run
                  skip = benchmark.do_setup()
                         ^^^^^^^^^^^^^^^^^^^^
                File "/tmp/persistent/newconda-envs/iris3/lib/python3.11/site-packages/asv_runner/benchmarks/time.py", line 80, in do_setup
                  result = Benchmark.do_setup(self)
                           ^^^^^^^^^^^^^^^^^^^^^^^^
                File "/tmp/persistent/newconda-envs/iris3/lib/python3.11/site-packages/asv_runner/benchmarks/_base.py", line 632, in do_setup
                  setup(*self._current_params)
                File "/tmp/persistent/git/iris/benchmarks/benchmarks/cube.py", line 21, in setup
                  source_cube = realistic_4d_w_everything(w_mesh=w_mesh)
                                ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
                File "/tmp/persistent/git/iris/benchmarks/benchmarks/generate_data/stock.py", line 182, in realistic_4d_w_everything
                  _ = run_function_elsewhere(_external, w_mesh_=w_mesh, save_path_=str(save_path))
                      ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
                File "/tmp/persistent/git/iris/benchmarks/benchmarks/generate_data/__init__.py", line 95, in run_function_elsewhere
                  result = run(
                           ^^^^
                File "/tmp/persistent/newconda-envs/iris3/lib/python3.11/subprocess.py", line 571, in run
                  raise CalledProcessError(retcode, process.args,
              subprocess.CalledProcessError: Command '['/tmp/persistent/newconda-envs/iris3/bin/python', '-c', "def _external(w_mesh_: str, save_path_: str):\n    import iris\n    from iris.tests.stock import realistic_4d_w_everything\n\n    cube = realistic_4d_w_everything(w_mesh=bool(w_mesh_))\n    iris.save(cube, save_path_)\n\n_external(w_mesh_=True,save_path_='/data/users/itpp/benchmarks_testdata/realistic_4d_w_everything_True.nc')"]' returned non-zero exit status 1.
              asv: benchmark failed (exit status 1)

  File "/tmp/persistent/git/iris/benchmarks/bm_runner.py", line 636, in <module>
    main()
  File "/tmp/persistent/git/iris/benchmarks/bm_runner.py", line 632, in main
    parsed.func(parsed)
  File "/tmp/persistent/git/iris/benchmarks/bm_runner.py", line 596, in func
    _subprocess_runner(asv_command, asv=True)
  File "/tmp/persistent/git/iris/benchmarks/bm_runner.py", line 49, in _subprocess_runner
    return subprocess.run(args, **kwargs)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/tmp/persistent/newconda-envs/tp_asvtst/lib/python3.12/subprocess.py", line 571, in run
    raise CalledProcessError(retcode, process.args,
subprocess.CalledProcessError: Command '['asv', 'run', '--bench', 'CubeCreation', '--quick', '-e', '--environment', 'existing:/tmp/persistent/newconda-envs/iris3/bin/python']' returned non-zero exit status 2.
nox > Command python bm_runner.py trialrun CubeCreation /tmp/persistent/newconda-envs/iris3/bin/python failed with exit code 1
nox > Session benchmarks failed.
```

### DIFF
--- a/benchmarks/bm_runner.py
+++ b/benchmarks/bm_runner.py
@@ -537,6 +537,65 @@ class Custom(_SubParserGenerator):
         _subprocess_runner([args.asv_sub_command, *args.asv_args], asv=True)
 
 
+class TrialRun(_SubParserGenerator):
+    name = "trialrun"
+    description = (
+        "Fast trial-run a given benchmark, to check it works : "
+        "in a provided or latest-lockfile environment, "
+        "with no repeats for accuracy of measurement."
+    )
+    epilog = (
+        "e.g. python bm_runner.py trialrun "
+        "MyBenchmarks.time_calc /tmp/testpython/bin/python"
+        "\n  NOTE: setting $DATA_GEN_PYTHON is equivalent to the "
+        "'runpath' argument."
+    )
+
+    def add_arguments(self) -> None:
+        self.subparser.add_argument(
+            "benchmark",
+            type=str,
+            help=(
+                "A benchmark name, possibly including wildcards, "
+                "as supported by the ASV '--benchmark' argument."
+            )
+        )
+        self.subparser.add_argument(
+            "runpath",
+            type=str,
+            nargs="?",
+            help=(
+                "A path to an existing python environment, "
+                "to completely bypass environment building."
+            )
+        )
+
+    @staticmethod
+    def func(args: argparse.Namespace) -> None:
+        print(args)
+        if args.runpath:
+            # Shortcut creation of a data-gen environment
+            # - which is also the trial-run env.
+            environ["DATA_GEN_PYTHON"] = args.runpath
+        _setup_common()
+        # get path of data-gen environment, setup by previous call
+        python_path = environ["DATA_GEN_PYTHON"]
+        # allow 'on-demand' benchmarks
+        environ["ON_DEMAND_BENCHMARKS"] = "1"
+        asv_command = [
+            "run",
+            "--bench",
+            args.benchmark,
+            # no repeats for timing accuracy
+            "--quick",
+            # show any errors
+            "-e",
+            # do not build a unique env : run test in data-gen environment
+            "--environment", f"existing:{python_path}"
+        ] + args.asv_args
+        _subprocess_runner(asv_command, asv=True)
+
+
 class GhPost(_SubParserGenerator):
     name = "_gh_post"
     description = (
@@ -566,7 +625,7 @@ def main():
     )
     subparsers = parser.add_subparsers(required=True)
 
-    for gen in (Overnight, Branch, CPerf, SPerf, Custom, GhPost):
+    for gen in (Overnight, Branch, CPerf, SPerf, Custom, TrialRun, GhPost):
         _ = gen(subparsers).subparser
 
     parsed = parser.parse_args()
--- a/benchmarks/bm_runner.py
+++ b/benchmarks/bm_runner.py
@@ -547,12 +547,7 @@ class TrialRun(_SubParserGenerator):
     epilog = (
         "e.g. python bm_runner.py trialrun "
         "MyBenchmarks.time_calc /tmp/testpython/bin/python"
-        "\n  NOTE: 'runpath' is also used for any data-generation, i.e. it"
-        "replaces $DATA_GEN_PYTHON during this run."
-        "\n  NOTE#2: setting $OVERRIDE_TEST_DATA_REPOSITORY avoids the runner "
-        "installing iris-test-data."
-        "\n  NOTE#3: setting $BENCHMARK_DATA may be desirable to specify "
-        "where is safe to create potentially large (Gb) test data."
+        "\n\nNOTE: 'runpath' also replaces $DATA_GEN_PYTHON during the run."
     )
 
     def add_arguments(self) -> None:
@@ -624,7 +619,20 @@ class GhPost(_SubParserGenerator):
 def main():
     parser = ArgumentParser(
         description="Run the Iris performance benchmarks (using Airspeed Velocity).",
-        epilog="More help is available within each sub-command.",
+        epilog=(
+            "More help is available within each sub-command."
+            "\n\nNOTE(1): iris-test-data is downloaded and cached within the "
+            "benchmarks code directory.\n   Set $OVERRIDE_TEST_DATA_REPOSITORY "
+            "to avoid the cost of this."
+            "\nNOTE(2): a separate python environment is created to write "
+            "benchmark test data.\n   Set $DATA_GEN_PYTHON to avoid the cost "
+            "of this."
+            "\nNOTE(3): test data is cached within the "
+            "benchmarks code directory, and uses a lot of disk space "
+            "of disk space (Gb).\n   Set $BENCHMARK_DATA to specify where this "
+            "space can be safely allocated."
+        ),
+        formatter_class=argparse.RawTextHelpFormatter,
     )
     subparsers = parser.add_subparsers(required=True)
 
--- a/benchmarks/bm_runner.py
+++ b/benchmarks/bm_runner.py
@@ -546,7 +546,7 @@ class TrialRun(_SubParserGenerator):
     )
     epilog = (
         "e.g. python bm_runner.py trialrun "
-        "MyBenchmarks.time_calc /tmp/testpython/bin/python"
+        "MyBenchmarks.time_calc ${DATA_GEN_PYTHON}"
         "\n\nNOTE: 'runpath' also replaces $DATA_GEN_PYTHON during the run."
     )
 
@@ -621,12 +621,12 @@ def main():
         description="Run the Iris performance benchmarks (using Airspeed Velocity).",
         epilog=(
             "More help is available within each sub-command."
-            "\n\nNOTE(1): iris-test-data is downloaded and cached within the "
-            "benchmarks code directory.\n   Set $OVERRIDE_TEST_DATA_REPOSITORY "
-            "to avoid the cost of this."
-            "\nNOTE(2): a separate python environment is created to write "
-            "benchmark test data.\n   Set $DATA_GEN_PYTHON to avoid the cost "
+            "\n\nNOTE(1): a separate python environment is created to "
+            "construct test files.\n   Set $DATA_GEN_PYTHON to avoid the cost "
             "of this."
+            "\nNOTE(2): iris-test-data is downloaded and cached within the "
+            "data generation environment.\n   Set "
+            "$OVERRIDE_TEST_DATA_REPOSITORY to avoid the cost of this."
             "\nNOTE(3): test data is cached within the "
             "benchmarks code directory, and uses a lot of disk space "
             "of disk space (Gb).\n   Set $BENCHMARK_DATA to specify where this "
--- a/benchmarks/bm_runner.py
+++ b/benchmarks/bm_runner.py
@@ -558,7 +558,7 @@ class TrialRun(_SubParserGenerator):
             help=(
                 "A benchmark name, possibly including wildcards, "
                 "as supported by the ASV '--benchmark' argument."
-            )
+            ),
         )
         self.subparser.add_argument(
             "runpath",
@@ -567,12 +567,11 @@ class TrialRun(_SubParserGenerator):
             help=(
                 "A path to an existing python environment, "
                 "to completely bypass environment building."
-            )
+            ),
         )
 
     @staticmethod
     def func(args: argparse.Namespace) -> None:
-        print(args)
         if args.runpath:
             # Shortcut creation of a data-gen environment
             # - which is also the trial-run env.
@@ -591,7 +590,8 @@ class TrialRun(_SubParserGenerator):
             # show any errors
             "-e",
             # do not build a unique env : run test in data-gen environment
-            "--environment", f"existing:{python_path}"
+            "--environment",
+            f"existing:{python_path}",
         ] + args.asv_args
         _subprocess_runner(asv_command, asv=True)
 
--- a/benchmarks/bm_runner.py
+++ b/benchmarks/bm_runner.py
@@ -547,8 +547,8 @@ class TrialRun(_SubParserGenerator):
     epilog = (
         "e.g. python bm_runner.py trialrun "
         "MyBenchmarks.time_calc /tmp/testpython/bin/python"
-        "\n\n  NOTE#1: setting $DATA_GEN_PYTHON is equivalent to the "
-        "'runpath' argument."
+        "\n  NOTE: 'runpath' is also used for any data-generation, i.e. it"
+        "replaces $DATA_GEN_PYTHON during this run."
         "\n  NOTE#2: setting $OVERRIDE_TEST_DATA_REPOSITORY avoids the runner "
         "installing iris-test-data."
         "\n  NOTE#3: setting $BENCHMARK_DATA may be desirable to specify "
@@ -567,7 +567,6 @@ class TrialRun(_SubParserGenerator):
         self.subparser.add_argument(
             "runpath",
             type=str,
-            nargs="?",
             help=(
                 "A path to an existing python executable, "
                 "to completely bypass environment building."
--- a/benchmarks/bm_runner.py
+++ b/benchmarks/bm_runner.py
@@ -325,7 +325,7 @@ class _SubParserGenerator(ABC):
     def add_asv_arguments(self) -> None:
         self.subparser.add_argument(
             "asv_args",
-            nargs=argparse.REMAINDER,
+            nargs="*",
             help="Any number of arguments to pass down to the ASV benchmark command.",
         )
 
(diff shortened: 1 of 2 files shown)


trexfeathers:
Since `OVERRIDE_TEST_DATA` is mentioned as **required** in the README, perhaps we ought to have something in `bm_runner.py` that insists it is there. It would probably be unwelcome to automate the installation of iris-test-data, and I don't know how else we could automate the detection, but insisting on the environment variable would at least provide a quick failure and meaningful error message?

benchmarks/bm_runner.py
```diff
@@ -537,6 +537,66 @@ def func(args: argparse.Namespace) -> None:
...
+        "'runpath' argument."
+    )
+
+    def add_arguments(self) -> None:
+        self.subparser.add_argument(
+            "benchmark",
+            type=str,
+            help=(
+                "A benchmark name, possibly including wildcards, "
+                "as supported by the ASV '--benchmark' argument."
```
trexfeathers, line 560:
```suggestion
                "as supported by the ASV '--bench' argument."
```

benchmarks/bm_runner.py
```diff
@@ -537,6 +537,66 @@ def func(args: argparse.Namespace) -> None:
...
+            ),
+        )
+        self.subparser.add_argument(
+            "runpath",
+            type=str,
+            nargs="?",
```
trexfeathers, line 566: Why `nargs` instead of [`required`](https://docs.python.org/3/library/argparse.html#required)?

benchmarks/bm_runner.py
```diff
@@ -537,6 +537,66 @@ def func(args: argparse.Namespace) -> None:
...
+            help=(
+                "A path to an existing python environment, "
```
trexfeathers, line 568: Is this more accurate?

```suggestion
                "A path to an existing python executable, "
```

benchmarks/bm_runner.py
```diff
@@ -537,6 +537,66 @@ def func(args: argparse.Namespace) -> None:
...
+        # allow 'on-demand' benchmarks
+        environ["ON_DEMAND_BENCHMARKS"] = "1"
+        asv_command = [
+            "run",
+            "--bench",
+            args.benchmark,
+            # no repeats for timing accuracy
+            "--quick",
+            # show any errors
+            "-e",
```
trexfeathers, line 592:
```suggestion
            "-show-stderr",
```

Self-explaining, and consistent with the other fully named arguments.

benchmarks/bm_runner.py
trexfeathers: Possibly related - if I don't provide any arguments I get a misleading error message:

```
$ python bm_runner.py trialrun       
usage: bm_runner.py trialrun [-h] benchmark [runpath] ...
bm_runner.py trialrun: error: the following arguments are required: benchmark, asv_args
```

Misleading because `asv_args` is not actually required

benchmarks/bm_runner.py
```diff
@@ -537,6 +537,66 @@ def func(args: argparse.Namespace) -> None:
...
+            help=(
+                "A path to an existing python environment, "
+                "to completely bypass environment building."
+            ),
+        )
+
+    @staticmethod
+    def func(args: argparse.Namespace) -> None:
+        print(args)
+        if args.runpath:
```
trexfeathers, line 576: `pathlib` could `resolve()` `args.runpath` so that relative paths can be passed (my testing indicates that only absolute paths work at the moment).

benchmarks/bm_runner.py
trexfeathers: Is `nargs` perhaps to allow commands like this to work fine (`--version` being an ASV argument)?

```
python bm_runner.py trialrun CubeCreation --version
```

trexfeathers (changes requested):
I love this. Delivers exactly what is needed, especially useful as a guide to someone inexperienced with ASV.

benchmarks/bm_runner.py
pp-mo: > Why nargs instead of [required](https://docs.python.org/3/library/argparse.html#required)?
I initially tried "required=False", but it causes an error - it says it's not allowed for positional arguments.

pp-mo:
Sorry I just force-pushed because the pre-commit bot had added corrections.
I think I was using an env without pre-commit installed 🤦

benchmarks/bm_runner.py
pp-mo: OK I read it up + I think that "nargs='?'" **_is_** correct + intended usage.
https://docs.python.org/3/library/argparse.html#nargs : where it says 
> If no command-line argument is present, the value from [default](https://docs.python.org/3/library/argparse.html#default) will be produced. Note that for optional arguments, there is an additional case ...

I think that implies that it does make sense for positionals too, and the following examples seem to confirm it.

pp-mo: > a misleading error message

This appears to be just standard, unfortunately..
```
>>> parser = argparse.ArgumentParser()
>>> _ = parser.add_argument("x")
>>> _ = parser.add_argument("y", nargs="*")
>>> parser.parse_args(["asb", "def"])
Namespace(x='asb', y=['def'])
>>> parser.parse_args(["asb"])
Namespace(x='asb', y=[])
>>> parser.parse_args([])
usage: [-h] x [y ...]
: error: the following arguments are required: x, y

>>>
```

So, the summary shows 'y' as being optional, but the error message doesn't say that :-(

pp-mo: .. and here's another oddity.

The "argparse.REMAINDER" constant appears to be undocumented, but it translates to "...".
The effect is odd though, as seen from the error when you don't give a benchmark :

With "nargs=argparse.REMAINDER" :
`usage: bm_runner.py trialrun [-h] benchmark [runpath] ...`
But with "nargs='*'" :
`usage: bm_runner.py trialrun [-h] benchmark [runpath] [asv_args ...]`
(the rest being unchanged)

So, I think the latter is more correct + I've now adopted `nargs="*"`.

Sadly, the syntax error message is still unchanged
`bm_runner.py trialrun: error: the following arguments are required: benchmark, asv_args`
 -- i.e. still "wrong" about the optional nature of asv_args.

On second thoughts though, perhaps an interpretation of that message is that, since it consumes any number of args, including zero, then it _is_ always present (and, coincidentally, always "required").

benchmarks/bm_runner.py
pp-mo: This was trickier than it sounds, because the noxfile sets cwd when running the bm_runner, and bm_runner does the same when running asv.
So, I **_think_** it works to leave the path relative and run bm_runner in the original cwd.
I've done that, and it seems to work.  But I'd be grateful if you'd check it out any way you can think of @trexfeathers

trexfeathers: I raised this because I tried using a relative path and it couldn't find it, but worked fine when I gave the full path. I don't understand the harm of just fully resolving the path? I had assumed that a fully resolved path would work no matter what the cwd was?

trexfeathers: Also ab09e95760eae1f0e7060316d9480d54a1220d38 has actioned my suggestion so I'm not sure what we're debating any more

pp-mo:
Ok, [latest commit](https://github.com/SciTools/iris/pull/5957/commits/ab09e95760eae1f0e7060316d9480d54a1220d38) I think addresses the review comments.
But I did make some more changes too, and notably this bit may not be quite correct in advance of #5958

benchmarks/bm_runner.py
```diff
@@ -537,6 +537,69 @@ def func(args: argparse.Namespace) -> None:
...
+        "in a provided or latest-lockfile environment, "
+        "with no repeats for accuracy of measurement."
+    )
+    epilog = (
+        "e.g. python bm_runner.py trialrun "
+        "MyBenchmarks.time_calc /tmp/testpython/bin/python"
+        "\n\n  NOTE#1: setting $DATA_GEN_PYTHON is equivalent to the "
+        "'runpath' argument."
+        "\n  NOTE#2: setting $OVERRIDE_TEST_DATA_REPOSITORY avoids the runner "
+        "installing iris-test-data."
```
pp-mo, line 553: NOTE: this one isn't really true -- until #5958 is adopted.

benchmarks/bm_runner.py
pp-mo: > the harm of just fully resolving the path?
You can't resolve the path when running via nox, since nox resets the cwd before calling the bm_runner (and passing the arg).
So if left as it was, a relative path always resolved relative to the ./benchmarks dir instead of the user's cwd.
I did check this + it didn't work !

trexfeathers: OK I understand now.

Well what you have done works when either `nox` or `python bm_runner.py` is called so that's all good.

benchmarks/bm_runner.py
```diff
@@ -325,7 +325,7 @@ def add_arguments(self) -> None:
     def add_asv_arguments(self) -> None:
         self.subparser.add_argument(
             "asv_args",
-            nargs=argparse.REMAINDER,
+            nargs="*",
```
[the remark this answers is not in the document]
pp-mo: That is tiresome.

I have now found a reference on StackOverflow explaining ["what happened to argparse.REMAINDER"](https://stackoverflow.com/questions/76956139/why-did-python-argparse-stop-documenting-nargs-remainder) 
It's a bit of a sorry story! 
I assume that the problems you demonstrate are due the fact that, as it describes (paraphrased) "options are always special to argparse, unless REMAINDER is used".

I really would rather get rid of this usage, which certainly puzzled me as it is (now) completely undocumented -- and at least potentially unsupported I guess.  But the list of alternatives in the StackOverflow answer is really not encouraging.

I couldn't find a good alternative, **_unless_** I make a special implementation for a '--' [as used by nox for this purpose](https://nox.thea.codes/en/stable/config.html#passing-arguments-into-sessions) , so may seem familiar.
E.G. `python bm_runner.py trialrun CubeCreation -- --version`.
So, that seems familiar to me, but I can't find any hint of it elsewhere (but it is very hard to search for!)

This does require special support -- argparse can't do it directly.  But I now have a trial solution running.
Basically I pre-process the arguments before calling argparse.
**Do you think that is an acceptable solution ?**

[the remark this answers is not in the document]
pp-mo: P.S. when used with nox itself though, the usage can get a little weird.
e.g. `$ nox -s benchmarks -- trialrun CubeCreation -- -=version`
It does work, though.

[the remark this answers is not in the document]
trexfeathers: Not a fan, sorry.

I think we should make the second argument always required. If the developer does not want to pass in an path they should enter `NEW` or something similar. That leaves us with just 1 object of arbitary length (`asv_args`), which I hope would work fine.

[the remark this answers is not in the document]
pp-mo: Update : on re-reading this seems a bit obscure.
I'm going to sort out those changes + push them up so you can see what I mean...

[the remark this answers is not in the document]
trexfeathers: Thanks for the commit. I still feel that the syntax is too confusing, as you mentioned:



> P.S. when used with nox itself though, the usage can get a little weird. e.g. `$ nox -s benchmarks -- trialrun CubeCreation -- --version` It does work, though.

pp-mo:
Thanks @trexfeathers 
Latest I hope addresses all the outstanding -- please re-review !

benchmarks/bm_runner.py
```diff
@@ -537,6 +537,69 @@ def func(args: argparse.Namespace) -> None:
...
+    epilog = (
+        "e.g. python bm_runner.py trialrun "
+        "MyBenchmarks.time_calc /tmp/testpython/bin/python"
+        "\n\n  NOTE#1: setting $DATA_GEN_PYTHON is equivalent to the "
+        "'runpath' argument."
+        "\n  NOTE#2: setting $OVERRIDE_TEST_DATA_REPOSITORY avoids the runner "
+        "installing iris-test-data."
+        "\n  NOTE#3: setting $BENCHMARK_DATA may be desirable to specify "
+        "where is safe to create potentially large (Gb) test data."
+    )
```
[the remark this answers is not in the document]
pp-mo: See https://github.com/SciTools/iris/pull/5957/commits/7bae96bc31673cfbaef7a65d694367d5ce6f6567

benchmarks/bm_runner.py
```diff
@@ -562,11 +619,24 @@ def add_asv_arguments(self) -> None:
 def main():
     parser = ArgumentParser(
         description="Run the Iris performance benchmarks (using Airspeed Velocity).",
-        epilog="More help is available within each sub-command.",
+        epilog=(
+            "More help is available within each sub-command."
+            "\n\nNOTE(1): iris-test-data is downloaded and cached within the "
+            "benchmarks code directory.\n   Set $OVERRIDE_TEST_DATA_REPOSITORY "
```
trexfeathers, line 625: It's actually in the `DATA_GEN_PYTHON` environment, so this message needs tweaking and would make more sense to come after the current `NOTE(2)`.

benchmarks/bm_runner.py
```diff
@@ -537,6 +537,63 @@ def func(args: argparse.Namespace) -> None:
...
+    description = (
+        "Fast trial-run a given benchmark, to check it works : "
+        "in a provided or latest-lockfile environment, "
+        "with no repeats for accuracy of measurement."
+    )
+    epilog = (
+        "e.g. python bm_runner.py trialrun "
+        "MyBenchmarks.time_calc /tmp/testpython/bin/python"
+        "\n\nNOTE: 'runpath' also replaces $DATA_GEN_PYTHON during the run."
+    )
```
trexfeathers, line 551: I liked your idea of also including an example that uses $DATA_GEN_PYTHON - for people who aren't in the detailed headspace that we are, and might forget.

```
python bm_runner.py trialrun MyBenchmarks.time_calc $DATA_GEN_PYTHON
```

trexfeathers (changes requested):
All seems to work, but the `--help` is slightly misleading so I have a suggestion

trexfeathers (approved):

pp-mo:
Thanks for your patience, @trexfeathers !

PR #5958, opened 2024-05-17
Automatically install iris-test-data for benchmark data generation

trexfeathers:
### Description

#5957 demonstrates that trouble with iris-test-data is a barrier to understanding at the moment.

---

---

### DIFF
--- a/benchmarks/bm_runner.py
+++ b/benchmarks/bm_runner.py
@@ -111,11 +111,13 @@ def _prep_data_gen_env() -> None:
             ]
         )
 
-        echo("Installing iris-test-data into data generation environment ...")
-        test_data_dir = clone_resource(
-            "iris-test-data", "https://github.com/SciTools/iris-test-data.git"
-        )
-        environ["OVERRIDE_TEST_DATA_REPOSITORY"] = str(test_data_dir / "test_data")
+        test_data_var = "OVERRIDE_TEST_DATA_REPOSITORY"
+        if test_data_var not in environ:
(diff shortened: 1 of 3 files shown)


benchmarks/bm_runner.py
```diff
@@ -112,6 +111,12 @@ def _prep_data_gen_env() -> None:
             ]
         )
 
+        echo("Installing iris-test-data into data generation environment ...")
```
pp-mo, line 114: Could we skip this if there is an existing setting ?
i.e. `if "OVERRIDE_TEST_DATA_REPOSITORY" not in environ`

pp-mo (changes requested):
Look OK, but I think would it be nicer to respect an existing OVERRIDE_TEST_DATA_REPOSITORY and just use that.

If you agree, I think you should also reinstate a mention in the README

benchmarks/bm_runner.py
trexfeathers: 47fe9a8a66089c7a6d8193f0763aca027cc04060

pp-mo: As discussed offline ...
From one angle, it looks like the fix will re-install every time as it can't set OVERRIDE.. permanently in the caller's env.
But actually it is OK because the download is in a fixed location, and [once it exists there it will not be re-downloaded](https://github.com/SciTools/iris/blob/47fe9a8a66089c7a6d8193f0763aca027cc04060/benchmarks/bm_runner.py#L98-L99).

pp-mo (approved):

PR #4621, opened 2022-03-04
Sperf & Cperf Benchmarks

trexfeathers:
### Description

* Migrates the SPerf and CPerf benchmarks for the UK Met Office NG-VAT project.
* Refactors some benchmark modules to account for these additions.
* Adds the `@on_demand_benchmark` decorator, since S/CPerf benchmarks are inappropriate for inclusion by default.

---
[Consult Iris pull request check list]( https://scitools-iris.readthedocs.io/en/latest/developers_guide/contributing_pull_request_checklist.html)

PR #5959, opened 2024-05-17
Remove "on-demand" from some benchmarks

pp-mo:
Since we believe the new [tracemalloc memory measurement method](https://github.com/SciTools/iris/pull/5948)  (TMMM :tm:) ought to deliver more stable results, 
Let's see what happens if we put those back in the standard set.

There are a lot more in cperf/sperf, but I think we [decided to remove all of those from the standard set](https://github.com/SciTools/iris/pull/4621).
I'd be open to reconsidering that -- if anyone thinks appropriate, @trexfeathers ?

### DIFF
--- a/benchmarks/benchmarks/experimental/ugrid/regions_combine.py
+++ b/benchmarks/benchmarks/experimental/ugrid/regions_combine.py
@@ -18,7 +18,7 @@ from iris import load, load_cube, save
 from iris.experimental.ugrid import PARSE_UGRID_ON_LOAD
 from iris.experimental.ugrid.utils import recombine_submeshes
 
-from ... import TrackAddedMemoryAllocation, on_demand_benchmark
+from ... import TrackAddedMemoryAllocation
 from ...generate_data.ugrid import make_cube_like_2d_cubesphere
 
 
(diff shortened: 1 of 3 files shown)


trexfeathers:
>There are a lot more in cperf/sperf, but I think we https://github.com/SciTools/iris/pull/4621.
I'd be open to reconsidering that -- if anyone thinks appropriate, @trexfeathers ?

CPerf and SPerf would be inappropriate for other reasons - CPerf relies on local files, and SPerf can take a long time due to deliberately large data.

benchmarks/benchmarks/experimental/ugrid/regions_combine.py
```diff
@@ -183,7 +183,6 @@ def time_compute_data(self, n_cubesphere):
         _ = self.recombined_cube.data
 
     # Vulnerable to noise, so disabled by default.
```
trexfeathers, line 185: You left the comments behind!

trexfeathers (changes requested):
Please could we get a What's New entry. The `Internal` section is beginning to serve as a good account of where we've been busy recently.

benchmarks/benchmarks/experimental/ugrid/regions_combine.py
pp-mo: Good spot !
https://github.com/SciTools/iris/pull/5959/commits/1b2a9f140a0dfeeda142c1cdad217cece8790c99 fixes, I think

trexfeathers (approved):

pp-mo:
Oh, I still missed the whatsnew.
I'll do something about that ...

pp-mo:
> missed the whatsnew.

See #5961

PR #5960, opened 2024-05-20
Add memory benchmarks

stephenworsley:
### Description

Extends existing benchmarks by adding a version which tracks memory for functions where iris may be responsible for memory handling (e.g. by calling dask in a specific way, as resolved in #5767). Memory benchmarks are repeated so that memory leaks may be better detected.

### DIFF
--- a/benchmarks/benchmarks/__init__.py
+++ b/benchmarks/benchmarks/__init__.py
@@ -114,6 +114,33 @@ class TrackAddedMemoryAllocation:
         decorated_func.unit = "Mb"
         return _wrapper
 
+    @staticmethod
+    def decorator_repeating(repeats=3):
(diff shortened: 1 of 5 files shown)


pp-mo (approved):
Some questions did occur to me about whether we are measuring the right things, 
but I think this is a good start, so let's merge + see if the results prove stable enough to use as regular checks for performance changes.

pp-mo:
Note: I [added a test to the evaluation branch + results are interesting](https://github.com/SciTools/iris/pull/5946#issuecomment-2121406650) 
From that it would seem that we are measuring somehow the _effects_ of memory allocations, not simply the original requests,
because 
   1. results show variation with repeat runs, and 
   2. repeated operations which regularly increase allocated space don't see a regular rise in the measured peak memory.

PR #5989, opened 2024-06-06
Specify meta in dask.array.map_blocks

### LINKED ISSUE #5988: Regridding produces wrong chunktype for lazy masked arrays
## 🐛 Bug Report

Regridding sets the chunktype to `numpy.ndarray` when it should be `numpy.MaskedArray`. This can result in ignoring the mask when subsequently using `iris.cube.Cube.rolling_window`.

## How To Reproduce

```python
import iris
import dask.array as da

global_air_temp = iris.load_cube(iris.sample_data_path('air_temp.pp'))
rotated_psl = iris.load_cube(iris.sample_data_path('rotated_pole.nc'))

global_air_temp.data = da.ma.masked_array(global_air_temp.core_data())
print(global_air_temp.core_data())
# prints dask.array<masked_array, shape=(73, 96), dtype=float32, chunksize=(73, 96), chunktype=numpy.MaskedArray>

rotated_air_temp = global_air_temp.regrid(rotated_psl, iris.analysis.Linear())
print(rotated_air_temp.core_data())
# prints dask.array<_regrid, shape=(22, 36), dtype=float32, chunksize=(22, 36), chunktype=numpy.ndarray>
```
## Expected behaviour

I would expect the second print statement above to produce
```
dask.array<_regrid, shape=(22, 36), dtype=float32, chunksize=(22, 36), chunktype=numpy.MaskedArray>
```

## Environment 
 - OS & Version: Ubuntu 23.10
 - Iris Version: 3.9

bouweandela:
### Description

Specify meta in `dask.array.map_blocks` to avoid having the wrong chunktype in the output array.

Closes #5988 

---

---


pp-mo:
Hmm linkcheck fail is this one...
(    whatsnew/1.4: line  131) broken    https://www.opendap.org/ - 500 Server Error: Internal Server Error for url: https://www.opendap.org/

The website really is down right now :-(

lib/iris/_lazy_data.py
```diff
@@ -596,8 +599,14 @@ def map_complete_blocks(src, func, dims, out_sizes, *args, **kwargs):
...
-        )
+        # Assume operation does not change dtype and meta if not specified.
+        if "meta" not in kwargs:
+            kwargs["meta"] = da.utils.meta_from_array(data)
+        if "dtype" in kwargs:
+            kwargs["meta"] = kwargs["meta"].astype(kwargs["dtype"])
+        else:
+            kwargs["dtype"] = kwargs["meta"].dtype
+
+        result = data.map_blocks(func, *args, chunks=out_chunks, **kwargs)
```
pp-mo, line 610: I'm finding this special handling of "dtype" and/or "meta" rather confusin,.
And I note that it  is currently **_only_** exercised in tests, and nowhere in the actual code.

So I'm wondering if we can do without ??

If not, maybe they need documenting as a special case, or even listing as explicit keys :  in the current code, it is rather confusing that these keywords, which specifically control `map_blocks`, are  are part of kwargs but  **_not_** passed down to `func` like the docstring says they are (!).

lib/iris/_lazy_data.py
```diff
@@ -536,9 +536,12 @@ def lazy_elementwise(lazy_array, elementwise_op):
     # This makes good practical sense for unit conversions, as a Unit.convert
     # call may cast to float, or not, depending on unit equality : Thus, it's
     # much safer to get udunits to decide that for us.
-    dtype = elementwise_op(np.zeros(1, lazy_array.dtype)).dtype
+    meta = da.utils.meta_from_array(lazy_array)
+    new_meta = elementwise_op(meta)
```
pp-mo, line 540: This now requires `elementwise_op` to work on an empty-shaped array, which was not the case before, and may not work for all array operations.
So I think that passing a minimal array of the correct dtype, as before, would be "safer".

pp-mo (changes requested):
This does seem basically sound, but I have a couple of worries to raise.
We can go with this if you don't like the suggestions.

lib/iris/_lazy_data.py
bouweandela: I restored the previous behaviour in c1e5e1a

lib/iris/_lazy_data.py
bouweandela: I was concerned about the assumption that the output `dtype` is always the same as the input `dtype`, that's why I thought it might be good to be able to specify this. I checked the `dtypes` coming out of the functions that use `map_complete_blocks` and tried to specify those in c1e5e1a. Does it look better now?

pp-mo:
Thanks for your patience @bouweandela !
We can get this into v3.11

pp-mo (approved):

PR #5999, opened 2024-06-11
Faster time coordinate categorization

bouweandela:
### Description

Make time coordinate categorization faster by converting the time points to datetimes in a single call.

Example script:
```python
In [1]: import numpy as np

In [2]: import cf_units

In [3]: from iris.cube import Cube

In [4]: from iris.coords import DimCoord

In [5]: from iris.coord_categorisation import add_year

In [6]: coord = DimCoord(np.arange(10000), standard_name='time', units=cf_units.Unit("days since epoch", "standard"))

In [7]: cube = Cube(np.empty(10000), dim_coords_and_dims=[(coord, 0)])

In [8]: %timeit add_year(cube.copy(), 'time')
9.99 ms ± 24.1 µs per loop (mean ± std. dev. of 7 runs, 100 loops each)
```

while with the current `main` branch, the timing result is
```python
345 ms ± 2.07 ms per loop (mean ± std. dev. of 7 runs, 1 loop each)
```

---

---


bouweandela:
Would it make sense to add the example code above as a benchmark?

lib/iris/coord_categorisation.py
```diff
@@ -42,7 +51,11 @@ def add_categorised_coord(cube, name, from_coord, category_function, units="1"):
         coordinate point-value.
     units : str, default="1"
         Units of the category value, typically 'no_unit' or '1'.
-
+    category_function_expects_date : bool, default=False
```
bouweandela, line 54: As an alternative to adding this extra argument, I could also add a new function `add_categorised_time_coord`. This function could then also take a simpler `category_function` argument, which would accept a function with only the point `value` argument, as the `coord` argument is no longer needed in any of the time categorization functions below.

trexfeathers: I have a third alternative: use `inspect.signature()` to check for type hinting on the second parameter. This would avoid needing `category_function_expects_date` or `add_categorised_time_coord`; it would all just work™. Obviously the docstring would need to make this behaviour clear. It seems a good way to open up the functionality for custom user coord categorisation, too.

`lambda`s don't properly support type hinting, so we'd have to abandon those, which I admit is a bit of a shame.

What do you think?

lib/iris/coord_categorisation.py
trexfeathers, line 1: >Would it make sense to add the example code above as a benchmark?

I believe we decided in a @SciTools/peloton meeting that this would be too specific given the limited running time available for benchmarks.

trexfeathers (commented):
Excellent idea @bouweandela, we just need to come to an agreement on how best to identify the cases where we're dealing with dates.

docs/src/whatsnew/latest.rst
```diff
@@ -51,6 +51,9 @@ This document explains the changes made to Iris for this release
 #. N/A
 
 
+#. `@bouweandela`_ made the time coordinate categorisation functions in
+   :mod:`iris.coord_categorisation` faster. (:pull:`5999`)
+
```
trexfeathers, line 56:
```suggestion
#. `@bouweandela`_ made the time coordinate categorisation functions in
   :mod:`iris.coord_categorisation` faster. (:pull:`5999`)

```

lib/iris/coord_categorisation.py
bouweandela: Thanks for reviewing @trexfeathers! I implemented the proposed approach, could you have another look?

docs/src/whatsnew/latest.rst
```diff
@@ -48,8 +48,8 @@ This document explains the changes made to Iris for this release
 🚀 Performance Enhancements
 ===========================
 
-#. N/A
-
+#. `@bouweandela`_ made the time coordinate categorisation functions in
+   :mod:`iris.coord_categorisation` faster. (:pull:`5999`)
```
trexfeathers, line 52: `add_categorised_coord()` is public API because it also allows user custom categorisation. It would be good to make clear that this improvement is available to anyone using `add_categorised_coord()` with `cftime.datetime` .

trexfeathers (changes requested):
OK, final couple of comments and I think we're ready to go 🚀

docs/src/whatsnew/latest.rst
bouweandela: Done in 16b20b3

trexfeathers (approved):

Issue #5908, opened 2024-04-09
Rationalising mesh dimension names


bjlittle:
## 🐛 Bug Report

Given two valid CF-UGRID netCDF mesh files, the **first** of which has the following netCDF `dimensions`:

```
dimensions:
        time = 4 ;
        dim0 = 221184 ;
        num_node = 221186 ;
        num_vertices = 4 ;
```

The **second** of which has the following netCDF `dimensions`:

```
dimensions:
        time = 22 ;
        bnds = 2 ;
        dim0 = 221184 ;
        num_node = 221186 ;
        num_vertices = 4 ;
```

I easily can use `iris` to load these files into as single cube list and then save to a single combined netCDF file as follows:

```python
import iris
from iris.experimental.ugrid import PARSE_UGRID_ON_LOAD

with PARSE_UGRID_ON_LOAD.context():
    cubes = iris.load(fnames)

iris.save(cubes, "combined.nc")
```

Inspecting the `dimensions` of `combined.nc` gives:

```
dimensions:
        num_node = 221186 ;
        dim0 = 221184 ;
        time = 22 ;
        dynamics_face_N_nodes = 4 ;
        bnds = 2 ;
        num_node_0 = 221186 ;
        dim0_0 = 221184 ;
        time_0 = 4 ;
        num_node_1 = 221186 ;
        dim0_1 = 221184 ;
```

😢 

Consequently, attempting to then load `combined.nc` results in the following traceback:

```python
Traceback (most recent call last):
  File "<snip>/script.py", line 36, in <module>
    combined = iris.load(fname) 
               ^^^^^^^^^^^^^^^^
  File "<snip>/iris/lib/iris/__init__.py", line 326, in load
    return _load_collection(uris, constraints, callback).merged().cubes()
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "<snip>/iris/lib/iris/__init__.py", line 294, in _load_collection
    result = _CubeFilterCollection.from_cubes(cubes, constraints)
             ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "<snip>/iris/lib/iris/cube.py", line 97, in from_cubes
    for cube in cubes:
  File "<snip>/iris/lib/iris/__init__.py", line 275, in _generate_cubes
    for cube in iris.io.load_files(part_names, callback, constraints):
[...]

Issue #5165, opened 2023-02-20
Common agreement on loading CF non-compliant NetCDF files


trexfeathers:
Iris needs a public statement on how it handles NetCDF files that deviate from the CF conventions. This will serve multiple benefits:

- More certainty when discussing if/how Iris should load a particular file.
- Clearer direction when developing the codebase.
- Set user expectations.

Writing this statement will involve making some difficult decisions. A working group is tackling this now: @tkknight, @bjlittle, @lbdreyer, @pp-mo, @trexfeathers, @stephenworsley, @ESadek-MO, @scottrobinson02, @HGWright

### Factors at play

- More CF compliance means smoother collaboration between institutions, and Iris can play a part in raising awareness.
- CF evolves over time, so may develop 'opinions' on things that previously didn't matter and invalidate older files.
- The available tooling can make it difficult to address non-compliances in a file.
- UX - being strict/verbose about CF compliance makes the user experience more awkward.
- Iris has a place in the scientific Python community - people choose Iris / Xarray / raw netCDF4 / something else / for different purposes, and CF handling plays a part in that.
- Continuing to work in the face of CF non-compliances could need more defensive code.

### Items affected

(please edit if you know of others)

- #5119
- #5126
- #5068
- #5067
- #5003
- #4495
- #1801
- #5171
- #4453
- #5257 
```[tasklist]
### Tasks
- [ ] https://github.com/SciTools/iris/issues/5068
- [ ] https://github.com/SciTools/iris/issues/5119
```

trexfeathers:
# Summary from working group conversations

`2023-02-02`, `2023-02-14`, `2023-03-22`

Note this issue is not intended as a debate, hence why it is not posted as a discussion. The below conversations took place in real time, with a group deliberately sized to aid decision making.

## Outcome - our ideal implementation

**When loading NetCDF files, Iris will load all CF-compliant elements. A container of non-compliant variables and attributes will be attached to the `Cube`(s).**

Encourage users:

>If this causes you problems, please reach out to us to see if we can collaborate on a solution.

## Implementation considerations

- How to contain things that can't be represented properly?
- Associate things with `Cube`s or isolated in own list?
- Activate behaviour with a `FUTURE` flag?

## Working group summary comments

- @trexfeathers: embrace imperfection, skipping non-compliances sounds good if warnings work.
- @stephenworsley: CF compliance is a good aim, but can't always be expected.
[...]

PR #6004, opened 2024-06-12
Mesh saveload fix

pp-mo:
### Description

---

---


lib/iris/experimental/ugrid/load.py
```diff
@@ -485,9 +485,27 @@ def _build_mesh_coords(mesh, cf_var):
         "edge": mesh.edge_dimension,
         "face": mesh.face_dimension,
     }
-    mesh_dim_name = element_dimensions[cf_var.location]
-    # (Only expecting 1 mesh dimension per cf_var).
-    mesh_dim = cf_var.dimensions.index(mesh_dim_name)
+    location = getattr(cf_var, "location", "<empty>")
```
pp-mo, line 488: N.B. This relates to https://github.com/SciTools/iris/issues/5165

It would be really nice to be more "tolerant" here : I had hoped to issue appropriate warnings (instead of errors), and simply not attach the mesh in any of these problematic cases.

However, when I tried to code that, I hit problems with **_testing_** it, mostly because of the context in which the UGRID code was written : 
  * it is using logging instead of warnings, which was probably is mis-step + wants "changing back" (and is awkward to test for)
  * the UGRID loading code has a _particularly_ "hardline" approach to the unexpected : mostly just "fail if not as expected" -- e.g. in [the motivating issue](https://github.com/SciTools/iris/issues/5908), you just get an obscure indexing error.  So, my attempts to code some "tolerant loading in this respect only" just hit a bunch of snags, adding a lot of additional complexity in unrelated areas

**_So,_** to avoid exploding the cost beyond the scope of this current effort, I instead cut my losses here + settled for just producing more intelligible error messages.
Hopefully we can still address the tolerant loading approach in future, but right now that seems to be a bigger challenge for the UGRID code than the 'regular' netcdf loading

lib/iris/tests/unit/fileformats/netcdf/saver/test_Saver__ugrid.py
```diff
@@ -1172,7 +1172,50 @@ def test_connectivity_names(self):
             )
             self.assertEqual(expected_names, result_names, fail_msg)
 
-    def _check_two_different_meshes(self, vars):
+    def test_multiple_equal_mesh(self):
```
pp-mo, line 1175: NOTE: this bit looks complicated, but only because I removed a subsidiary routine.
Check out the individual commits, which make a lot more sense in this case :
  * https://github.com/SciTools/iris/pull/6004/commits/5365adea44e7b01f4dc55043c62b2cbb1ebe5838
  * https://github.com/SciTools/iris/pull/6004/commits/e57c30d23906525f5b115faec50dbedd31ab37cf

lib/iris/tests/unit/fileformats/netcdf/saver/test_Saver__ugrid.py
```diff
@@ -1215,63 +1258,19 @@ def _check_two_different_meshes(self, vars):
...
         if "edge_coordinates" in vars["Mesh2d_0"]:
-            self.assertEqual(vars["longitude_0"][_VAR_DIMS], ["Mesh2d_edge_0"])
+            self.assertEqual(vars["longitude_0"][_VAR_DIMS], ["Mesh2d_edge"])
             self.assertEqual(
                 vars["mesh2d_edge_0"][_VAR_DIMS],
-                ["Mesh2d_edge_0", "Mesh2d_0_edge_N_nodes"],
+                ["Mesh2d_edge", "Mesh2d_0_edge_N_nodes"],
             )
 
-    def test_multiple_equal_mesh(self):
```
pp-mo, line 1231: NOTE: Actually not much has _really_ changed here  : see https://github.com/SciTools/iris/pull/6004/files#r1638046009

pp-mo:
Hi @ESadek-MO 
I fixed the test problems + I think this is probably good to review now.
Let me know what you think !

ESadek-MO (changes requested):
Thanks @pp-mo! Looks good to me, just need a whatsnew then happy for this to go.

pp-mo:
> Thanks @pp-mo! Looks good to me, just need a whatsnew then happy for this to go.

Thanks @ESadek-MO .
Hope this does it.

PR #6005, opened 2024-06-14
Added more descriptive errors within concatenate

### LINKED ISSUE #5683: Include helpful error if non-identical meshes have prevented concatenation or merging
## ✨ Feature Request

As discovered in #5576, calling `concatenate_cube()` on cubes where the `MeshCoord`s appear identical, but are derived from non-identical `Mesh`es, produces this un-informative error:

```
ConcatenateError: failed to concatenate into a single cube.
  An unexpected problem prevented concatenation.
```

This is unlike other concatenation errors that hint at the potential blockers (e.g. different coordinates). **We should add an informative error about differences in the cubes' `Mesh`es.**

I'm unaware if merging has similar problems, but this should also be looked into.

## Motivation

Concatenation and merging are both a common use of Iris, and rather difficult to get right. Various errors attempt to help the user understand problems; without these errors the problem becomes almost impossible to debug and users are forced to as developers for help. If we can provide more information about impossible concatenation/merging then we should.

## Additional context

I believe this stems from the fact that users are led to expect differences in coordinate values and metadata to be the problem, but `MeshCoord` equality ALSO checks `Mesh` equality. The concatenate code is unaware of this so it becomes an 'invisible' concern.

ESadek-MO:
Closes #5683.


lib/iris/_concatenate.py
```diff
@@ -926,6 +930,12 @@ def register(
                         or candidate_axis not in coord_b.dims
                     ):
                         if not coord_a == coord_b:
+                            mismatch_error_msg = (
+                                "Auxiliary coordinates differ for phenomenon"
```
trexfeathers, line 934: I'm looking at the `match()` method (code below). The word `differ` is used there too, meaning there's no decent distinction between those existing errors (where a coord is absent, or the metadata is different), and these new errors.

**I prefer the word `unequal` here**, because we are actually using equality whereas the existing code is doing more basic comparisons. That should give us the distinction we need.

https://github.com/SciTools/iris/blob/e92e36ac0e845318635a507b7e58914c5633d03c/lib/iris/_concatenate.py#L554-L562

lib/iris/_concatenate.py
```diff
@@ -926,6 +930,12 @@ def register(
...
+                                f" {self._cube_signature.defn.name()}:\n"
```
trexfeathers, line 935:
```suggestion
                                f" `{self._cube.name()}`:\n"
```

We were doing our best to mimic `match()`, but we have access to extra info here that makes for more readable code.

I feel the backticks help highlight the name within the message, too.

Obvs this comment applies to the other cases too.

lib/iris/_concatenate.py
```diff
@@ -991,6 +1019,14 @@ def register(
             if existing_order == _CONSTANT and this_order != _CONSTANT:
                 self._coord_signature.dim_order[dim_ind] = this_order
 
+        if mismatch_error_msg:
+            if not match and error_on_mismatch:
+                raise iris.exceptions.ConcatenateError([mismatch_error_msg])
+            elif not match:
```
trexfeathers, line 1025:
```suggestion
        if mismatch_error_msg and not match:
            if error_on_mismatch:
                raise iris.exceptions.ConcatenateError([mismatch_error_msg])
            else:
```

I _think_ this is the same?

Neat structure, by the way 🎖

lib/iris/tests/unit/concatenate/test_concatenate.py
```diff
@@ -246,6 +246,45 @@ def test_dim_coords_overlap_message(self):
             _ = concatenate([cube_1, cube_2], True)
 
 
+class TestNonMetadataMessages(TestMessages):
```
trexfeathers, line 249: This unfortunately inherits a lot of other tests, too, so those are all running twice. The correct thing to do here would be to:

- create a `_MessagesMixin` class that ONLY has the `setUp()` method
- have `TestMessages` inherit from `_MessagesMixin`
- have your new class also inherit from `_MessagesMixin`, leaving you free to write whatever tests you want with whatever names you want

lib/iris/_concatenate.py
```diff
@@ -901,17 +902,20 @@ def register(
             match = candidate_axis is not None and (
                 candidate_axis == axis or axis is None
             )
+            if not match:
+                mismatch_error_msg = (
+                    f"Cannot find an axis to concatenate over for phenomenon "
+                    f"{self._cube_signature.defn.name()}"
+                )
```
trexfeathers, line 909: We need test coverage for this one

lib/iris/_concatenate.py
```diff
@@ -991,6 +1019,14 @@ def register(
             if existing_order == _CONSTANT and this_order != _CONSTANT:
                 self._coord_signature.dim_order[dim_ind] = this_order
 
+        if mismatch_error_msg:
+            if not match and error_on_mismatch:
+                raise iris.exceptions.ConcatenateError([mismatch_error_msg])
+            elif not match:
+                warnings.warn(
+                    mismatch_error_msg, category=iris.warnings.IrisUserWarning
+                )
```
trexfeathers, line 1028: Need test coverage for this block. I know the tests are still using `unittest`, but it should be possible to sneak `pytest.mark.parametrize` and `pytest.warns()` in there without it crashing.

trexfeathers (changes requested):
Nice work, and glad to see @rcomer likes it too!

Some comments for you

Issue #6002, opened 2024-06-12
Lazy cube interpolation


fnattino:
It would be great if the Iris cube (rectilinear) interpolation could run lazily, since this would enable a few of the ESMValCore preprocessor [to work lazily as well](https://github.com/ESMValGroup/ESMValCore/issues/2446).

Right now the cube data is immediately realised when setting up the interpolator: 
https://github.com/SciTools/iris/blob/57096f29f6f2c5ff278d616df5e2da4d2bab95e2/lib/iris/analysis/_interpolation.py#L208-L209

and the [`_RegularGridInterpolator`](https://github.com/SciTools/iris/blob/57096f29f6f2c5ff278d616df5e2da4d2bab95e2/lib/iris/analysis/_scipy_interpolate.py#L42) that does the actual weight calculation and interpolation does not seem to support lazy data due to the interaction with scipy's sparse matrices.  

Could we maybe contribute in making the interpolation lazy by dropping the immediate data realisation, and by including support for Dask arrays in the  `_RegularGridInterpolator`?

HGWright:
completed by #6084

PR #6006, opened 2024-06-14
Lazy rectilinear interpolator, with `sparse`

fnattino:
### Description

Enable the rectilinear interpolator to run lazily  #6002.

The original implementation makes use of a CSR matrix from `scipy.sparse` to represent the weights. However, this does not seem to work well with Dask arrays, since it does not adhere (completely) to the array interface. Switching to sparse arrays from [`sparse`](https://sparse.pydata.org/en/stable/) seems to work well, allowing the sparse matrix to be wrapped in a Dask array.  

Would this be an acceptable change? On the long run, it looks like `sparse` will replace `scipy.sparse` in the PyData ecosystem..


fnattino:
Marking it as ready for review to hear your thoughts on this. What is in here for now seems to work (see e.g. code snippet below), provided that `sparse` is installed. 

<details>

<summary> Test code snippet </summary>

```python
import iris

from iris.analysis import RectilinearInterpolator


LATITUDE = [16., 16.1]
LONGITUDE = 226.


def get_cube():
    filename = iris.sample_data_path('E1_north_america.nc')
    return iris.load_cube(filename, 'air_temperature')


def interpolate(method):
    assert method in ('linear', 'nearest')
    cube = get_cube()
    coords = ('latitude', 'longitude')
    points = (LATITUDE, LONGITUDE)
    interpolator = RectilinearInterpolator(cube, coords, method, "mask")
    print('Cube is lazy: ', cube.has_lazy_data())
    # Cube is lazy:  True
    result = interpolator(points, collapse_scalar=True)
    print('Result is lazy: ', result.has_lazy_data())
    # Result is lazy:  True


if __name__ == '__main__':
    interpolate('linear')
    interpolate('nearest')

```

</details>

trexfeathers:
Hi @fnattino, thanks for your hard work. Just to warn you I expect several weeks' delay before we can get to this as we are accumulating a backlog of ESMValTool changes while we prepare GeoVista for [SciPy 2024](https://www.scipy2024.scipy.org/) and complete the mesh-focused [Iris 3.10 release](https://github.com/SciTools/iris/discussions/5570).

fnattino:
I have worked on a different approach that makes the intepolator lazy by using a similar approach as used in the regridder, see #6084. This has the disadvantage that it requires merging the chunks along the interpolating dimensions, but it is not as "disruptive" as in it does not add new dependencies, so maybe worth to have a look at #6084  first keeping this for the longer run? In the meantime I can also try to run some benchmarks to compare the performance of `sparse` vs `scipy.sparse`.

HGWright:
From discussion at the @SciTools/peloton. Is this still required since #6084 has been completed and merged in? @fnattino

fnattino:
Hi @HGWright, this was indeed initial work for an alternative to #6084 -- it can now be closed.

Issue #5557, opened 2023-10-30
If a face-coord in an unstructured cube has a non-None coordinate system, this breaks cube.copy()


pp-mo:
## How To Reproduce
Minimal (non!-)working example :
```
>>> from iris.coord_systems import GeogCS
>>> from iris.tests.stock.mesh import sample_mesh_cube
>>> 
>>> cube = sample_mesh_cube()
>>> coord = cube.mesh.coords(include_faces=True)[0]
>>> coord.coord_system = GeogCS(1.0)
>>> cube2 = cube.copy()
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "/tmp/persistent/newconda-envs/own_antsdev/lib/python3.10/site-packages/iris/cube.py", line 3579, in copy
    cube = self._deepcopy(memo, data=data)
  File "/tmp/persistent/newconda-envs/own_antsdev/lib/python3.10/site-packages/iris/cube.py", line 3595, in _deepcopy
    new_aux_coords_and_dims = deepcopy(self._aux_coords_and_dims, memo)
  File "/tmp/persistent/newconda-envs/own_antsdev/lib/python3.10/copy.py", line 146, in deepcopy
    y = copier(x, memo)
  File "/tmp/persistent/newconda-envs/own_antsdev/lib/python3.10/copy.py", line 206, in _deepcopy_list
    append(deepcopy(a, memo))
  File "/tmp/persistent/newconda-envs/own_antsdev/lib/python3.10/copy.py", line 146, in deepcopy
    y = copier(x, memo)
  File "/tmp/persistent/newconda-envs/own_antsdev/lib/python3.10/copy.py", line 211, in _deepcopy_tuple
    y = [deepcopy(a, memo) for a in x]
  File "/tmp/persistent/newconda-envs/own_antsdev/lib/python3.10/copy.py", line 211, in <listcomp>
    y = [deepcopy(a, memo) for a in x]
  File "/tmp/persistent/newconda-envs/own_antsdev/lib/python3.10/copy.py", line 153, in deepcopy
    y = copier(memo)
  File "/tmp/persistent/newconda-envs/own_antsdev/lib/python3.10/site-packages/iris/experimental/ugrid/mesh.py", line 2991, in __deepcopy__
    return self.copy()
  File "/tmp/persistent/newconda-envs/own_antsdev/lib/python3.10/site-packages/iris/experimental/ugrid/mesh.py", line 2977, in copy
    new_coord = MeshCoord(
[...]

Issue #3388, opened 2019-08-30
Support loading/saving netCDF with multiple coordinate systems


lbdreyer:
In CF-1.7 the use of the grid_mapping attribute was expanded to store multiple coordinate systems. See [example 5.10](http://cfconventions.org/Data/cf-conventions/cf-conventions-1.7/cf-conventions.html#british-national-grid)
Ref (for background, if needed): [trac ticket 70](https://cf-trac.llnl.gov/trac/ticket/70)

We should handle reading and writing files of these types of files.

lbdreyer:
Agreed to delay with til Iris 3, as it is no longer such a high priority for the user requesting this

pp-mo:
Note the relevance of #3796.  
It probably makes good sense to do this first.
However, as noted there, the feature described in #3796 can affect the implementation of this in some cases.  So it makes sense to bear that in mind.

pp-mo:
Just found that this is also has unexpected relevance to UGRID encoding.

So...  for regular structured data, with X+Y dimensions and 2 separate X+Y coordinates,
where the data is sampled on a grid which is **_not_** true-latlon based, 
  1. data-variable has a "grid_mapping" attribute, specifying a coordinate system, and 
  2. the X+Y coords will have the appropriate standard_names (i.e. "grid_lat(long)itude" OR "projection_x(y)", instead of plain "lat(long)itude". Also
  3. optionally, there may be true-latlon equivalent coordinate information.  This is strictly redundant, but CF recommend that it is supplied.  It will consist of variables called plain "lat(long)itude", which will be 2D (Y,X).

For example (using a rotated pole grid) ...
```
dimensions
  xmain = <nx>
  ymain = <ny>

variables
  float xmain(xmain)
    xmain:standard_name = 'grid_longitude'
    xmain:units = 'degrees'
 float  ymain(ymain)
    ymain:standard_name = 'grid_latitude'
    ymain:units = 'degrees'
 float xtrue(ymain, xmain):
    xtrue:standard_name = "longitude"
    xtrue:units = "degrees_east"
        # NB the distinction between 'degrees' and 'degrees_east' is canonical, but I don't think the unit is necessarily intended to be definitive :  e.g. we can reasonably use "radians" instead, and I don't think "radians_east" is a thing
 float ytrue(ymain, xmain):
    ytrue:standard_name = "latitude"
    ytrue:units = "degrees_north"

  int gridvar
[...]

PR #6016, opened 2024-06-20
Allow MeshCoord to have a coord-system

pp-mo:
This makes a MeshCoord have the coord-system of the appropriate location coordinate.

Note though that, as loaded from NetCDF-CF,  location coords can't actually have a coord-system at present.
Because the load rules only call `build_auxiliary_coordinate` in one place, and a coord-system (though supported) is never passed  :
   - see [here](https://github.com/SciTools/iris/blob/v3.9.0/lib/iris/fileformats/_nc_load_rules/actions.py#L482) 

In CF terms, this is because mesh coords are (at best) aux-coords of the data variable, and never _dimension_ coords.
And a grid-mapping can currently _only_ apply to dim-coords, because [we don't yet support the extended grid-mapping syntax](https://github.com/SciTools/iris/issues/3388#issuecomment-1263389772)

However ... this **does fix** the [copy error caused by a manually added coord-system](https://github.com/SciTools/iris/issues/5557)


lib/iris/experimental/ugrid/mesh.py
```diff
@@ -2806,8 +2811,19 @@ def axis(self):
 
     @property
     def coord_system(self):
-        """The coordinate-system of a MeshCoord is always 'None'."""
-        return None
+        """The coordinate-system of a MeshCoord comes from its origin coords."""
+        select_kwargs = {
+            f"include_{self.location}s": True,
```
pp-mo, line 2816: On second thoughts, by analogy with #4860 this should probably *always* use "include_nodes".

pp-mo: > by analogy with https://github.com/SciTools/iris/issues/4860 this should probably always use "include_nodes

No, that's the **_reverse_** of the change in #4860 !
This stays as it is.

lib/iris/experimental/ugrid/mesh.py
```diff
@@ -2784,6 +2784,11 @@ def fix_repr(val):
                     )
                     raise ValueError(msg)
 
+        # Don't use 'coord_system' as a constructor arg, since for
+        # MeshCoords it is deduced from the mesh.
+        # (Otherwise a non-None coord_system breaks the 'copy' operation)
+        use_metadict.pop("coord_system")
```
stephenworsley, line 2790: Does this fix a bug which already existed? If so, it might be worth adding a bugfix part to the whatsnew.

stephenworsley: Also, is there a test which shows that this copy operation is working correctly?

stephenworsley (changes requested):
Looks good, I think the main thing I'd like to be sure of is that this point about the copy operation being fixed now is properly tested and documented.

lib/iris/experimental/ugrid/mesh.py
pp-mo: > Does this fix a bug which already existed? If so, it might be worth adding a bugfix part to the whatsnew.

Well, effectively #5557 is that bug.
Since you can't get a mesh coordinate with a coord-system from loading a file, manual assigning one is the only way you would hit this problem, at present.

pp-mo: > is there a test which shows that this copy operation is working correctly

I guess not. I will add one.

pp-mo:
Thanks @stephenworsley 
Hopefully addressed those points now ?
Please re-review!

stephenworsley (approved):
Looks good.

PR #6017, opened 2024-06-20
Update test_Saver__ugrid.py

pp-mo:
Correct comment, since after #6004 the behaviour, and the testcase changed but the comment did not !

@ESadek-MO tiny fix : hope this makes sense to you.


lib/iris/tests/unit/fileformats/netcdf/saver/test_Saver__ugrid.py
```diff
@@ -178,7 +178,7 @@ def mesh_location_size(mesh, location):
 
 
 # A simple "standard" test mesh for multiple uses, which we can use for cubes
-# that *share* a mesh (since we don't support mesh equality).
+# that ** a mesh (since we don't support mesh equality).
```
ESadek-MO, line 181: Not sure what this is meant to mean?

"... we can use for cubes that ** a mesh..."

ESadek-MO (changes requested):

lib/iris/tests/unit/fileformats/netcdf/saver/test_Saver__ugrid.py
pp-mo: Good spot -- definitely a mistake ! 🤭

pp-mo:
N.B. I just realised this PR uses a branch of the main repo, instead of my fork.
That's probably just a mistake, but no big deal.  I will remove the unwanted branch afterwards.

ESadek-MO (approved):
LGTM! Thanks @pp-mo

PR #6035, opened 2024-07-02
Demo Numpy v2

trexfeathers:
### Description

This PR expands on the original demonstration (a82bea0c62a6f4235702451eac38ec8a17cae2ac) to make Iris fully NumPy v2 compliant. Notable changes:

- Adapting to [the new representation of NumPy scalars](https://numpy.org/doc/stable/release/2.0.0-notes.html#representation-of-numpy-scalars-changed).
- Restructuring for [the new dtype promotion rules](https://numpy.org/doc/stable/numpy_2_0_migration_guide.html#changes-to-numpy-data-type-promotion).
- Requirements files
- What's New entry

<details><summary>Expand for the text from the original demonstration:</summary>

We already know that Iris' _syntax_ is NumPy v2 compliant, [thanks to Ruff](https://docs.astral.sh/ruff/rules/#numpy-specific-rules-npy), but that does not prove that the code will all _function_ with NumPy v2.

A vanilla Conda install of Iris does get NumPy v2, but unfortunately our full test environment does not (some of the dependencies must be pinned). So I have forced the issue by manually editing `requirements/locks/py312-linux-64.lock`, which does thankfully produce a working environment since none of NumPy's dependencies have changed.

- [Vanilla install dependencies](https://github.com/conda-forge/iris-feedstock/blob/db2e0ab3c29fb11a05b8970109ce8751f980f9e4/recipe/meta.yaml#L18-L40)
- [Development environment dependencies](https://github.com/SciTools/iris/blob/1a73e8428e8ed5bc3320c64a02c7a3e1a920c549/requirements/py312.yml)

The test failures show us what work we need to do for Iris to be compliant with NumPy v2. It seems to mostly be around changes in:

- Handling of array `dtype`s
- Array string representation
- Array difference tolerances

</details>

---

---


trexfeathers:
Note to self: `tests.test_ff.TestFF2PP2Cube.test_unit_pass_0` keeps flipping between expecting `float32` or `float64`. I suspect a NumPy v2 change now means that something is no longer deterministic.

trexfeathers:
I propose that the performance changes (e.g. https://github.com/SciTools/iris/pull/6035#issuecomment-2416956573) are:

- Acceptable - only a couple of minor regressions
- Too specific to report in a What's New entry

trexfeathers:
> Note to self: `tests.test_ff.TestFF2PP2Cube.test_unit_pass_0` keeps flipping between expecting `float32` or `float64`. I suspect a NumPy v2 change now means that something is no longer deterministic.

99061d83843c568324ced2ca9810c47ba0cd1551

trexfeathers:
Thanks @HGWright!

PR #6036, opened 2024-07-02
Adapt setup.py for pypa/setuptools@2db55275f

trexfeathers:
### Description

We were mistakenly passing a tuple into something that needed a list - needs to be a mutable object.

pypa/setuptools@2db55275f

---

---


HGWright (approved):
Approving and merging this while the benchmarks are still in progress, since a full run is not necessary to test the change that we have made. We just wanted to make sure that the benchmarks could run and iris could be installed. (This is something that changing `setup.py` could affect.) We have confirmed that the benchmarks have completed full runs with this commit checked out and is able to run the benchmarks where as the commit on main is having benchmarks fail, (the change that we are making fixed that).

PR #6050, opened 2024-07-15
Quieter datum warning

### LINKED ISSUE #5749: Improve datum FUTURE warning
## 🐛 Bug Report

When loading any netCDF file, a warning is produced from iris:

> FutureWarning: Ignoring a datum in netCDF load for consistency with existing behaviour. In a future version of Iris, this datum will be applied. To apply the datum when loading, use the iris.FUTURE.datum_support flag.

The warning text implies that there is a datum in the file being loaded and it is being ignored.  Can the behaviour to trigger the warning be aligned with the warning text, so the warning only fires if there is a datum in the source file?

This would make the warning much more useful for end users: users would only see the warning for those cases where there is a behaviour change controlled by the iris.FUTURE.datum_support flag (and hence where the user may need to understand or do something).  At the moment, the warning is emitted for files where there will be no behaviour change too.

## How To Reproduce

1. Load a netCDF file with iris 3.7 (or iris since 3.3ish?)

## Expected behaviour

Only seeing the warning for netCDF files that do have a datum defined.

trexfeathers:
### Description

Closes #5749 

---

---

### DIFF
--- a/docs/src/whatsnew/latest.rst
+++ b/docs/src/whatsnew/latest.rst
@@ -55,7 +55,7 @@ This document explains the changes made to Iris for this release
 
 #. `@trexfeathers`_ fixed the datum :class:`python:FutureWarning` to only be raised if
    the ``datum_support`` :class:`~iris.Future` flag is disabled AND a datum is
-   present the loaded NetCDF grid mapping. (:issue:`5749`, :pull:`6050`)
+   present on the loaded NetCDF grid mapping. (:issue:`5749`, :pull:`6050`)
 
 
 💣 Incompatible Changes
--- a/docs/src/whatsnew/latest.rst
+++ b/docs/src/whatsnew/latest.rst
@@ -53,6 +53,10 @@ This document explains the changes made to Iris for this release
 #. `@pp-mo`_ corrected the use of mesh dimensions when saving with multiple
    meshes.  (:issue:`5908`, :pull:`6004`)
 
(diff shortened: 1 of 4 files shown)


docs/src/whatsnew/latest.rst
```diff
@@ -53,6 +53,10 @@ This document explains the changes made to Iris for this release
 #. `@pp-mo`_ corrected the use of mesh dimensions when saving with multiple
    meshes.  (:issue:`5908`, :pull:`6004`)
 
+#. `@trexfeathers`_ fixed the datum :class:`python:FutureWarning` to only be raised if
+   the ``datum_support`` :class:`~iris.Future` flag is disabled AND a datum is
+   present the loaded NetCDF grid mapping. (:issue:`5749`, :pull:`6050`)
```
HGWright, line 58: "AND a datum is present the loaded NetCDF grid mapping." I think this should be "AND a datum is present (ON) the loaded NetCDF grid mapping."

HGWright (changes requested):
Thanks @trexfeathers, just a small comment about the whatsnew entry grammar.

docs/src/whatsnew/latest.rst
trexfeathers:
```suggestion
   present on the loaded NetCDF grid mapping. (:issue:`5749`, :pull:`6050`)
```

HGWright (approved):
Thanks for that change.

Issue #5990, opened 2024-06-06
Set deadline for making `iris.experimental.ugrid` non-experimental


trexfeathers:

trexfeathers:
See #5725

trexfeathers:
Existing requests that are (or might be) breaking changes:

- #5993
  Ideally this would only add functionality rather than changing anything existing, but it's still being specced out.
- #5940
  We have defended our design decision based on still being experimental when the change is made.
- #6052
  Specifically designed as a quick way to get the breaking change in quickly while still experimental, allowing the larger proportion of the work (#5994) to take place at any time.

Other improvements I have looked over could deliberately be implemented in a way that would avoid breaking most existing user scripts.

trexfeathers:
### Future feedback that might require breaking changes

I can obviously only speculate. Every month Iris' mesh API gets exercised more, but this is by definition early adopters. We will only know if it is truly fit for purpose once it is used operationally, and that won't happen while it is marked as `experimental` (indeed we have already had conversations with someone about this being a problem).

We do know from the volume of issues and offline conversations that many people have already used Iris' mesh API and it has worked for them.

Any date we set on this basis will be arbitrary. **There is no concrete argument for delay here** - there isn't a key person we're waiting to hear from or anything like that.

trexfeathers:
### [3.10](https://github.com/SciTools/iris/discussions/5570) (now) is the best time

We're in a developer sprint with multiple experts available to tackle this, and coordination of potential conflicts.

PR #6055, opened 2024-07-17
add location argument to mesh.coords

### LINKED ISSUE #5993: Better API for accessing coordinates on a `Mesh`
## ✨ Feature Request

It should be easier to access the desired coordinate within a `Mesh`, taking into account that `location` and `axis` are commonly used filters, and that these are often dynamic.

## Motivation

I'm always frustrated when I have to write code like this:

https://github.com/SciTools/iris/blob/6adac1b6f83baa2e1d4a13ced0d6748ff91092ee/lib/iris/experimental/ugrid/mesh.py#L2736-L2738

stephenworsley:
## 🚀 Pull Request
Closes #5993

### DIFF
--- a/benchmarks/benchmarks/unit_style/ugrid.py
+++ b/benchmarks/benchmarks/unit_style/ugrid.py
@@ -93,9 +93,8 @@ class Mesh(UGridCommon):
         )
 
         def get_coords_and_axes(location):
-            search_kwargs = {f"include_{location}s": True}
             return [
-                (source_mesh.coord(axis=axis, **search_kwargs), axis)
+                (source_mesh.coord(axis=axis, location=location), axis)
                 for axis in ("x", "y")
             ]
 
--- a/benchmarks/benchmarks/unit_style/ugrid.py
+++ b/benchmarks/benchmarks/unit_style/ugrid.py
@@ -113,7 +113,7 @@ class Mesh(UGridCommon):
         self.node_x = self.object.node_coords.node_x
         # Kwargs for reuse in search and remove methods.
         self.connectivities_kwarg = dict(cf_role="edge_node_connectivity")
-        self.coords_kwarg = dict(include_faces=True)
+        self.coords_kwarg = dict(location="face")
 
         # TODO: an opportunity for speeding up runtime if needed, since
         #  eq_object is not needed for all benchmarks. Just don't generate it
--- a/lib/iris/experimental/ugrid/mesh.py
+++ b/lib/iris/experimental/ugrid/mesh.py
@@ -1553,7 +1553,7 @@ class Mesh(CFVariableMixin):
             include_edges = location == "edge"
             include_faces = location == "face"
         else:
-            incude_nodes = include_edges = include_faces = None
+            include_nodes = include_edges = include_faces = None
 
         result = self._coord_manager.filter(
             item=item,
@@ -1641,7 +1641,7 @@ class Mesh(CFVariableMixin):
             include_edges = location == "edge"
             include_faces = location == "face"
         else:
-            incude_nodes = include_edges = include_faces = None
+            include_nodes = include_edges = include_faces = None
 
         result = self._coord_manager.filters(
             item=item,
@@ -1791,7 +1791,6 @@ class Mesh(CFVariableMixin):
             ``Y``, ``Z`` and ``T`` (case-insensitive).
         location : str, optional
             The desired location. Accepts the values ``node``, ``edge`` or ``face``.
-            TODO: should this allow a list?
 
         Returns
         -------
@@ -1802,6 +1801,7 @@ class Mesh(CFVariableMixin):
         """
         # Filter out absent arguments - only expecting face coords sometimes,
         # same will be true of volumes in future.
+        # TODO: should location be allowed to be a list?
         if location is not None:
             if location not in ["node", "edge", "face"]:
                 raise ValueError(
@@ -1811,7 +1811,7 @@ class Mesh(CFVariableMixin):
             include_edges = location == "edge"
             include_faces = location == "face"
         else:
-            incude_nodes = include_edges = include_faces = None
+            include_nodes = include_edges = include_faces = None
 
         kwargs = {
             "item": item,
--- a/lib/iris/experimental/ugrid/mesh.py
+++ b/lib/iris/experimental/ugrid/mesh.py
@@ -1555,6 +1555,8 @@ class Mesh(CFVariableMixin):
 
         """
         if location is not None:
+            if location not in ["node", "edge", "face"]:
+                raise ValueError(f"Expected location to be one of `node`, `edge` or `face`, got `{location}`")
             include_nodes_new = location == "node"
             include_edges_new = location == "edge"
             include_faces_new = location == "face"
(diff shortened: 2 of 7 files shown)


lib/iris/experimental/ugrid/mesh.py
```diff
@@ -1631,6 +1662,39 @@ def coords(
             :class:`Mesh` that matched the given criteria.
 
         """
+        if location is not None:
```
trexfeathers, line 1665: I can understand why philosophically you are performing the validation here, but it would be cleaner to pass straight down to the underlying filtering methods. There is already some handling there:

https://github.com/SciTools/iris/blob/96cb02600b52a18ae3c30b8b145dd927d57a84fd/lib/iris/experimental/ugrid/mesh.py#L2279-L2284

Would that also offer an opportunity to avoid the duplication we currently have?

lib/iris/experimental/ugrid/mesh.py
```diff
@@ -1631,6 +1662,39 @@ def coords(
...
+            for loc in _location:
+                if loc not in ["node", "edge", "face"]:
+                    raise ValueError(
+                        f"Expected location to contain only `node`, `edge` or `face`, got `{location}`"
+                    )
+            include_nodes_new = "node" in _location
+            include_edges_new = "edge" in _location
+            include_faces_new = "face" in _location
+            if include_nodes not in [None, include_nodes_new]:
+                msg = f"Value of `include_nodes` ({include_nodes}) is incompatible with `location` ({location})"
```
trexfeathers, line 1683: Let's not bother with full-on support for both 'types' of parameter. @stephenworsley and I discussed a `DeprecationWarning`, but given the impending ambition to drop the `experimental` label (#5990) I guess we would be stuck supporting the old parameters for a long time.

It's all still `experimental` right now - could we just drop `include_nodes`, `include_edges`, and `include_faces` abruptly. It's not expected to be popular API anyway. Thoughts @bjlittle @pp-mo @stephenworsley?

trexfeathers: At the very least I don't think we need to be so careful aligning possible duplicate arguments. `location` is in charge, fall back on the other arguments if present.

trexfeathers (changes requested):
Some fairly wide scope things for you. I will hold off more detailed reviewing until we have some conclusions.

Thanks!

lib/iris/experimental/ugrid/mesh.py
```diff
@@ -2737,8 +2801,7 @@ def __init__(
         use_metadict = node_metadict.copy()
         if location != "node":
             # Location is either "edge" or "face" - get the relevant coord.
-            kwargs = {f"include_{location}s": True, "axis": axis}
-            location_coord = self.mesh.coord(**kwargs)
+            location_coord = self.mesh.coord(axis=axis, location=location)
```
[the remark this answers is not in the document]
trexfeathers: You've already made some changes but it could get a lot simpler ❤

```python
        mesh, location, axis = self.mesh, self.location, self.axis
        node_coord = mesh.coord(location="node", axis=axis)
        
        if location == "node":
            points_coord = node_coord
            bounds_connectivity = None
        else:
            points_coord = mesh.coord(location=location, axis=axis)
            bounds_connectivity = getattr(mesh, f"{location}_node_connectivity")
```

lib/iris/experimental/ugrid/mesh.py
trexfeathers: We discussed offline this morning and I still think it's important to push any handling down to this level, especially with the edits to `remove_coords()` as well. The code block I linked to also needs to be removed since the 'tri-state' is no longer something on offer for the user.

benchmarks/benchmarks/unit_style/ugrid.py
trexfeathers, line 1: Missed a spot:

https://github.com/SciTools/iris/blob/63c2cf25ab4a962a740a3306ffbd0ff562f48f48/benchmarks/benchmarks/unit_style/ugrid.py#L116

lib/iris/tests/unit/experimental/ugrid/mesh/test_Mesh.py
```diff
@@ -238,6 +241,7 @@ def test_coords(self):
             {"long_name": "long_name"},
             {"var_name": "node_lon"},
             {"attributes": {"test": 1}},
+            {"location": "node"},
```
trexfeathers, line 244: Why did this need adding? I has assumed that not passing `location` was equivalent to not passing any of the original 3 arguments?

lib/iris/tests/unit/experimental/ugrid/mesh/test_Mesh.py
```diff
@@ -248,6 +252,7 @@ def test_coords(self):
             {"long_name": "foo"},
             {"var_name": "foo"},
             {"attributes": {"test": 2}},
+            {"location": "face"},
```
trexfeathers, line 255: ?

lib/iris/tests/unit/experimental/ugrid/mesh/test_Mesh.py
```diff
@@ -571,32 +569,8 @@ def test_coords_elements(self):
...
-            (
-                {"include_faces": False, "include_edges": False},
-                ["node_x", "node_y"],
-            ),
-            (
-                {"include_faces": False, "include_edges": True},
-                ["edge_x", "edge_y"],
-            ),
+            ({"location": "node"}, ["node_x", "node_y"]),
+            ({"location": "edge"}, ["edge_x", "edge_y"]),
```
trexfeathers, line 573: This 2D version should be testing `face`, too.

lib/iris/experimental/ugrid/mesh.py
trexfeathers, line 1: Please could we have a What's New entry as well?

trexfeathers (changes requested):
Beautiful work @stephenworsley - neater, less confusing, and true to the original philosophy of identical signatures all eventually passing down to the `filters()` method.

lib/iris/tests/unit/experimental/ugrid/mesh/test_Mesh.py
stephenworsley: I believe this test is set up to loop over this list of kwarg dictionaries and check that each one returns the expected output. Adding an extra dictionary here improves the coverage of this test to include the new argument.

lib/iris/tests/unit/experimental/ugrid/mesh/test_Mesh.py
stephenworsley: As above, the addition of this argument expands the coverage of this test to check that the location arument _excludes_ appropriate coords.

lib/iris/tests/unit/experimental/ugrid/mesh/test_Mesh.py
trexfeathers: Closing after discussion offline

lib/iris/tests/unit/experimental/ugrid/mesh/test_Mesh.py
trexfeathers: Agreed offline:

```suggestion
            {"location": "edge"},
```

pp-mo (approved):
All good I think, just needs an approval.

Issue #6082, opened 2024-07-24
Loading cube fails if `cell_methods` is not valid


ukmo-ccbunney:
## 🐛 Bug Report
Iris fails to load some netCDF files due to a malformed `cell_method` attribute, but is throwing a slightly unhelpful error:
```python
File /opt/scitools/conda/deployments/default-2024_02_20/lib/python3.11/site-packages/iris/fileformats/_nc_load_rules/helpers.py:267, in _split_cell_methods(nc_cell_methods)
    265 for ii in range(len(name_start_inds) - 1):
    266     method_indices.append((name_start_inds[ii], name_start_inds[ii + 1]))
--> 267 method_indices.append((name_start_inds[-1], len(nc_cell_methods)))
    269 # Index the string and match against each substring
    270 nc_cell_methods_matches = []

IndexError: list index out of range
```

Inspection of the stack trace reveals it was failing to correctly match the `cell_method` attribute value because it was missing a space after the colon separator, i.e.
```
cell_methods = "time:point"
```
rather than:
```
cell_methods = "time: point"
```

This should probably raise a `Warning` that the `cell_method` could not be parsed and continue loading the cube.

## How To Reproduce
Steps to reproduce the behaviour:

1. Create a netCDF file where a variable has the following attribute (no space after colon): `cell_methods = "time:point"`
2. Load using `iris.load_cube(nc_filename)`

Throws error `IndexError: list index out of range` with stack trace shown above in description.

## Expected behaviour
Cube should be loaded but with a `Warning` stating the `cell_method` attribute could not be parsed.

## Environment 
 - Linux; RHEL7.9
 - 3.2.0.dev978

PR #6083, opened 2024-07-24
Update to `cell_method` parsing

ukmo-ccbunney:
### Description
Iris fails to load cubes from a netCDF file if the `cell_method` attribute is malformed and throws this slightly unhelpful error:
```python
File /opt/scitools/conda/deployments/default-2024_02_20/lib/python3.11/site-packages/iris/fileformats/_nc_load_rules/helpers.py:267, in _split_cell_methods(nc_cell_methods)
    265 for ii in range(len(name_start_inds) - 1):
    266     method_indices.append((name_start_inds[ii], name_start_inds[ii + 1]))
--> 267 method_indices.append((name_start_inds[-1], len(nc_cell_methods)))
    269 # Index the string and match against each substring
    270 nc_cell_methods_matches = []

IndexError: list index out of range
```

For the netCDF files in my particular case, inspection of the stack trace reveals it was failing to correctly parse the `cell_method` attribute which was missing a space between the colon separator and time method. I.e.
```
cell_methods = "time:point"
```
rather than:
```
cell_methods = "time: point"
```

This PR makes two changes to the cell method parsing in `fileformats/_nc_load_rules/helpers.py`:
1. In `_split_cell_methods`: If the value of `cell_methods` does not match the expected `name: ` pattern, a `Warning` is issued and the function returns an empty list so cube loading can continue. This is consistent with the behaviour further down the function if the `name: value` pattern cannot be matched.

2. Updates `_CM_PARSE_NAME` and `_CM_PARSE` regexes to make the pattern matching a little more lenient by making the space after the colon separator between the `name` and `method` pair optional. The [CF Convention](https://cfconventions.org/Data/cf-conventions/cf-conventions-1.7/build/ch07s03.html#:~:text=To%20describe%20the%20characteristic%20of%20a%20field%20that,words%20of%20the%20form%20%22%20name%3A%20method%20%22.) states that the pairs are "blank-separated list of `name: method` pairs". It _does_ show a space after the colon in the examples it gives, but the [CF Compliance checker ](https://cfchecker.ncas.ac.uk/) is happy with no space after the colon.

Unit tests have been updated to test that:
  1. Malformed `cell_method` raises a warning; e.g. name but no method, such as `cell_method='time'`
  2. Space after colon between name and method is optional; e.g. `cell_method='time:mean'`

Tested on Iris version 3.10.0.dev96.

- fixes: #6082


ukmo-ccbunney:
Leaving this one as draft for the moment whilst I address some unit test failures.

ukmo-ccbunney:
Modified slightly to address some failing unit tests.
All tests passing locally for me now.

trexfeathers:
From @SciTools/peloton:

Related to #5165.

At the very least there needs to be an informative warning, otherwise what's the point in Iris being strict about CF.

Also, if the CF checker is OK with this, maybe we should be OK with it also? @pp-mo would love your opinion on the CF checker's place in these decisions.

HGWright:
From @SciTools/peloton we want to accept this as CF-compliant.

ESadek-MO (changes requested):
Thanks @ukmo-ccbunney, changes all look good to me! I think it's worth just adding a small whatsnew entry for this, then it'll be good to go!

ukmo-ccbunney:
> Thanks @ukmo-ccbunney, changes all look good to me! I think it's worth just adding a small whatsnew entry for this, then it'll be good to go!

Done.

docs/src/whatsnew/latest.rst
```diff
@@ -52,6 +52,9 @@ This document explains the changes made to Iris for this release
 #. `@rcomer`_ enabled partial collapse of multi-dimensional string coordinates,
    fixing :issue:`3653`. (:pull:`5955`)
 
+#. `@ukmo-ccbunney`_ improved error handling for malformed `cell_method`
+   attribute. Also made cell_method string parsing more lenient w.r.t.
+   whitespace. (:pull:`6181`)
```
ESadek-MO, line 57:
```suggestion
   whitespace. (:pull:`6083`)
```

ESadek-MO (changes requested):
Wrong PR number!

docs/src/whatsnew/latest.rst
ukmo-ccbunney: Whoops! Good spot 🙄

ESadek-MO (approved):

PR #6084, opened 2024-07-25
Lazy rectilinear interpolator 

fnattino:
### Description

Different take to enable the rectilinear interpolator to run lazily #6002 . 

Trying to address the same issue as #6006, but there I have made the underlying `_RegularGridInterpolator`  from `_scipy_interpolate` to run on lazy data, which required switching from `scipy.sparse` to `sparse` (not ideal since it would add numba as a dependency). 

Here I have tried instead to implement a similar approach as used for regridding, which works on lazy data as well. The downside is that the chunks in the dimensions we are interpolating over need to be merged, but at least we could run interpolation in parallel over the chunks in the other dimensions (and we do not need to add extra dependencies to iris).


lib/iris/analysis/_interpolation.py
```diff
@@ -296,44 +312,51 @@ def _interpolate(self, data, interp_points):
...
-            # so we set it afterwards instead. Sneaky. ;-)
-            self._interpolator = _RegularGridInterpolator(
-                self._src_points,
-                data,
-                method=self.method,
-                bounds_error=mode.bounds_error,
-                fill_value=None,
-            )
-        else:
-            self._interpolator.values = data
```
fnattino, line 331: `_interpolate` is now mapped to all the chunks of the data we are interpolating, thus we cannot cache the `_RegularGridInterpolator` instance anymore. Is this acceptable? 

One can still cache the  `RectilinearInterpolator` thought, and, as far as I can say, this would be similar to the way in which the caching of the regridder works?

bouweandela:
Nice to see this progressing @fnattino! Did you notice that CI is failing on this pull request?

lib/iris/tests/unit/analysis/interpolation/test_RectilinearInterpolator.py
```diff
@@ -499,24 +499,37 @@ def test_orthogonal_cube_squash(self):
...
+        # Perform interpolation and check the data is real.
+        interpolator = RectilinearInterpolator(
+            self.cube, ["latitude"], LINEAR, EXTRAPOLATE
+        )
+        res = interpolator([[1.5]])
+        self.assertFalse(res.has_lazy_data())
+
+
 class Test___call___lazy_data(ThreeDimCube):
     def test_src_cube_data_loaded(self):
```
fnattino, line 518: This test used to check that setting up the interpolator would trigger data loading in the source cube, in order to avoid that new interpolations would have to load the same data again and again (see #1222). I have modified the test in favour of a couple of tests that should make more sense for a lazy interpolator.

trexfeathers:
@fnattino just want to reassure you that I have been looking at this, but since I have never worked with our interpolators before it is slow progress. Having another go this afternoon with help from some coffee ☕

fnattino:
No worries @trexfeathers, but thanks for the heads-up! :)

lib/iris/analysis/_interpolation.py
trexfeathers: Having read through the code base: I believe the caching is because the same interpolation routine gets used on a series of coordinates - start here and follow the call stack down to `_points()`:

https://github.com/SciTools/iris/blob/4c73724b2bad6469c2b75e7ed54d5be8195d8988/lib/iris/analysis/_interpolation.py#L656-L658

... meaning multiple uses of `_RegularGridInterpolator` but with only `.values` changing. I assume that there is noticeable overhead when creating a new instance of `_RegularGridInterpolator`, but that there is no problem applying the same instance to a series of different arrays, hence the caching.

There is a risk that your refactoring - which makes the routine much faster when parallelising large datasets - will make small-to-medium cases slower, especially for those cases that need to interpolate a larger number of coordinates. So I am very keen that we try out my suggestion using `args=[self]`, which I believe would allow us to retain the existing caching? If that does not work then I'll need to assess the performance impact with some benchmarking.

lib/iris/analysis/_interpolation.py
```diff
@@ -554,9 +571,20 @@ def _points(self, sample_points, data, data_dims=None):
             # Transpose data in preparation for interpolation.
             data = np.transpose(data, interp_order)
 
-        # Interpolate and reshape the data ...
-        result = self._interpolate(data, interp_points)
-        result = result.reshape(interp_shape)
+        # Interpolate the data, merging the chunks in the interpolated
+        # dimensions.
+        dims_merge_chunks = [dmap[d] for d in di]
+        result = map_complete_blocks(
```
trexfeathers, line 577: TODO for @trexfeathers: confirm that this use of `map_complete_blocks()` does not suffer the memory leaks described in #5767.

lib/iris/tests/unit/analysis/interpolation/test_RectilinearInterpolator.py
trexfeathers, line 1: TODO for @trexfeathers: review the tests.

trexfeathers (changes requested):
Hi @fnattino, thank you for your hard work on this.

Here is a partial review. I have left myself a couple of TODO comments. But the suggestions I have already might take some time, and may change the code significantly, so it seemed important to get these suggestions to you as soon as possible.

Also thank you to @HarriKeenan for helping me with this review last week 🤜🤛

lib/iris/analysis/_interpolation.py
```diff
@@ -12,6 +12,8 @@
 from numpy.lib.stride_tricks import as_strided
 import numpy.ma as ma
 
+from iris._lazy_data import map_complete_blocks
+from iris.analysis._scipy_interpolate import _RegularGridInterpolator
```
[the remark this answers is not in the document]
fnattino: I noticed that `_RegularGridInterpolator` was imported at module level in `_regrid.py`, and it seemed to work here as well, so I naively moved this out - but I will put it back to where it was :)

lib/iris/analysis/_interpolation.py
```diff
@@ -296,44 +312,51 @@ def _interpolate(self, data, interp_points):
...
-            )
-        else:
-            self._interpolator.values = data
+        # Determine the shape of the interpolated result.
+        ndims_interp = len(interp_shape)
+        extra_shape = data.shape[ndims_interp:]
+        final_shape = [*interp_shape, *extra_shape]
+
+        mode = EXTRAPOLATION_MODES[extrapolation_mode]
+        _data = np.ma.getdata(data)
```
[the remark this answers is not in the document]
fnattino: Passing a masked array to the `_RegularGridInterpolator` used to work, but the mask is actually ignored: when multiplying by the sparse matrix of weights, the mask is discarded and only the data is considered for the matrix multiplication (I think that scipy.sparse is not mask-aware). So I thought that it would be "cleaner" to pass only the data here (in the same way that only the mask is passed to the interpolator few lines below).

fnattino (commented):
@trexfeathers thanks a lot for the review and apologies for the scattered response. 

As I have tried to explain in reply to your comments, I am a bit hesitant to implement the solution that would copy the current instance of the `RectilinearInterpolation` - but I am very curious to hear your thoughts on this!

lib/iris/analysis/_interpolation.py
trexfeathers: Have confirmed 👍

lib/iris/analysis/_interpolation.py
trexfeathers, line 1: I tried writing a benchmark to demonstrate the benefits of this but everything either stayed the same or got slower.

Could you share with me the results you have seen? Especially if it's something I/we could turn into a benchmark. Thanks

trexfeathers (approved):
OK I got my head back in the space and I understand your logic better now. Nearly there

trexfeathers (changes requested):
Clicked the wrong button with my previous review!

fnattino:
Thanks a lot for the effort of getting back to this @trexfeathers ! I should have time to get back to this next week, but I will keep you posted!

lib/iris/analysis/_interpolation.py
fnattino: I have run few tests one some high resolution CMIP6 data (I have used [this](https://esgf.ceda.ac.uk/thredds/catalog/esg_cmip6/CMIP6/HighResMIP/MOHC/HadGEM3-GC31-HM/highresSST-present/r1i1p1f1/E3hrPt/ta/gn/v20170831/catalog.html?dataset=esg_cmip6/CMIP6/HighResMIP/MOHC/HadGEM3-GC31-HM/highresSST-present/r1i1p1f1/E3hrPt/ta/gn/v20170831/ta_E3hrPt_HadGEM3-GC31-HM_highresSST-present_r1i1p1f1_gn_201412010000-201412302100.nc) NetCDF file of approx 2 GB) and tested the performance of extracting one coordinate point from the cube.

<details>

<summary> Benchmark script used </summary>

```python
import argparse

import iris
import iris.analysis

# https://esgf.ceda.ac.uk/thredds/fileServer/esg_cmip6/CMIP6/HighResMIP/MOHC/HadGEM3-GC31-HM/highresSST-present/r1i1p1f1/E3hrPt/ta/gn/v20170831/ta_E3hrPt_HadGEM3-GC31-HM_highresSST-present_r1i1p1f1_gn_201412010000-201412302100.nc
DEFAULT_FILEPATH = "ta_E3hrPt_HadGEM3-GC31-HM_highresSST-present_r1i1p1f1_gn_201412010000-201412302100.nc"
SCHEME = {
    "linear": iris.analysis.Linear(),
    "nearest": iris.analysis.Nearest(),
}
SAMPLE_POINTS = [("latitude", 52.377956), ("longitude", 4.897070)]  # Amsterdam


def parse_args():
    parser = argparse.ArgumentParser()
    parser.add_argument("--cube-path",
                        help="Cube file path.",
                        type=str,
                        default=DEFAULT_FILEPATH)
    parser.add_argument("--scheme",
                        help="Interpolation scheme.",
                        type=str,
                        choices=("linear", "nearest"),
                        default="linear")
    parser.add_argument("--ntimes",
                        help="Only consider the first `ntimes` time elements.",
                        type=int,
                        default=None)
    args = parser.parse_args()
    return vars(args)


def interpolate(cube_path, scheme="linear", ntimes=None):
    print(f"Loading cube: {cube_path}")
    cube = iris.load_cube(cube_path)
    if ntimes is not None:
        cube = cube[0:ntimes, :, :, :]
    print(f"Cube shape: {cube.shape}")
    print(f"Cube has lazy data: {cube.has_lazy_data()}")
    print(f"Interpolation scheme: {scheme}")
    interpolated = cube.interpolate(SAMPLE_POINTS, SCHEME[scheme])
    print(f"Interpolated cube shape: {interpolated.shape}")
    print(f"Interpolated cube has lazy data: {interpolated.has_lazy_data()}")
    interpolated.data  # if interpolator is lazy, run it here!
    assert not interpolated.has_lazy_data()
    print("Done!")


def main():
    args = parse_args()
    interpolate(**args)


if __name__ == "__main__":
    main()

```
</details>

The data set (~4.9 GB when decompressed) is chunked in the time dimension (40 chunks, ~120 MB each):

```python
import iris
c = iris.load_cube("ta_E3hrPt_HadGEM3-GC31-HM_highresSST-present_r1i1p1f1_gn_201412010000-201412302100.nc")
c.core_data()
# dask.array<array, shape=(240, 7, 768, 1024), dtype=float32, chunksize=(6, 7, 768, 1024), chunktype=numpy.MaskedArray>
c.core_data().nbytes / 2**30
# 4.921875 GB
```

Interpolating this cube with the non-lazy interpolator tries to load and setup the interpolator for the whole cube. The lazy interpolator instead preserves the time chunks, so it is faster (it interpolates chunks in parallel) and more memory efficient.

I have run the benchmark locally ("old" MacBook Pro with Intel 4 core CPU), comparing this version of iris vs the `main` branch (as of today). 

When considering only the first 30 elements in the time dimension of the cube (`--ntimes 30`),
and timing the benchmark script as:
```shell
time python benchmark.py --scheme linear --ntimes 30
```
I get 7.262s (current branch) vs 11.398s (main branch) as wall times (~1.5 speedup). With the interpolator from the current branch, I can also run the benchmark on the full cube:
```shell
time python benchmark.py --scheme linear
``` 
which runs in 38.577s, while I run out of memory when trying to do the same using the interpolator from the main branch.

I also did some memory profiling using [`memory_profiler`](https://github.com/pythonprofilers/memory_profiler), running as follows to consider the time subset:
```shell
mprof run python benchmark.py --scheme linear --ntimes 30
```
The comparison of the memory usage vs time for the current branch (black) and main branch (blue):
linear-30

Also the memory profile when interpolating the cube while keeping the whole time axis:
```
mprof run python benchmark.py --scheme linear
``` 
linear-all

Hope it is clear, please let me know if you think I should run other tests!

lib/iris/analysis/_interpolation.py
```diff
@@ -163,6 +165,15 @@ def snapshot_grid(cube):
     return x.copy(), y.copy()
 
 
+def _interpolated_dtype(dtype, method):
+    """Determine the minimum base dtype required by the underlying interpolator."""
+    if method == "nearest":
+        result = dtype
+    else:
+        result = np.result_type(_DEFAULT_DTYPE, dtype)
+    return result
```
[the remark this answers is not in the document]
fnattino: Oops - I did not realize that I have left `_interpolated_dtype()` as a method of the `RectilinearInterpolator` as well. This should be fixed now!

lib/iris/analysis/_interpolation.py
```diff
@@ -490,9 +504,8 @@ def _points(self, sample_points, data, data_dims=None):
 
         Returns
         -------
-        :class:`~numpy.ndarray` or :class:`~numpy.ma.MaskedArray`
-            An :class:`~numpy.ndarray` or :class:`~numpy.ma.MaskedArray`
-            instance of the interpolated data.
+        ndarray
+            The interpolated data array.
```
fnattino, line 508: `_points` can now return a Dask array if input data is lazy

lib/iris/analysis/_interpolation.py
```diff
@@ -585,14 +609,22 @@ def __call__(self, sample_points, collapse_scalar=True):
...
             the number of scalar coordinates, if collapse_scalar is True.
 
+        Notes
+        -----
+        .. note::
+
+            If the source cube has lazy data,
+            `chunks <https://docs.dask.org/en/latest/array-chunks.html>`__
+            in the interpolated dimensions will be combined before regridding.
+
```
fnattino, line 619: Added note analogous to the one in the [`RectilinearRegridder.__call__` docstring](https://github.com/SciTools/iris/blob/f30980e64d88e25cc2e3efc41eb793619077ed7f/lib/iris/analysis/_regrid.py#L890-L897)

fnattino:
Thanks again for the review @trexfeathers  and sorry with the delay in replying. In addition to replying to your comments, I have updated a couple of docstrings (where I left comments). I have also realized that the user guide should probably also be updated, especially [the section on caching the interpolator](https://scitools-iris.readthedocs.io/en/latest/userguide/interpolation_and_regridding.html#caching-an-interpolator)? I can take care of this tomorrow. Any other docs that should be update?

docs/src/userguide/interpolation_and_regridding.rst
```diff
@@ -417,22 +417,22 @@ In each case ``result`` will be the input cube regridded to the grid defined by
...
     >>> air_temp
     <iris 'Cube' of air_temperature / (K) (time: 240; latitude: 37; longitude: 49)>
     >>> air_temp.has_lazy_data()
     True
 
-and the regridding scheme supports lazy data. All regridding schemes described
-here support lazy data. If you still run out of memory even while using lazy
-data, inspect the
+and the interpolation or regridding scheme supports lazy data. All interpolation and
+regridding schemes described here with exception of the the point-in-cell regridder
```
fnattino, line 434: The `CurvilinearRegridder` [used by the point-in-cell regridder](https://github.com/SciTools/iris/blob/67d0b0771834510b782ca57af9e445f6e2e21427/lib/iris/analysis/__init__.py#L3101) says it does not support lazy data: 
https://github.com/SciTools/iris/blob/67d0b0771834510b782ca57af9e445f6e2e21427/lib/iris/analysis/_regrid.py#L453-L454

fnattino:
I have made some small updates to the userguide to add that the linear and nearest-neighbour interpolators can now work with lazy data as well. 

Just wanted to double check, but I think we should  also remove the subsections on ["caching an interpolator"](https://scitools-iris.readthedocs.io/en/latest/userguide/interpolation_and_regridding.html#caching-an-interpolator) and ["caching a regridder"](https://scitools-iris.readthedocs.io/en/latest/userguide/interpolation_and_regridding.html#caching-a-regridder): as discussed in the thread below [this comment](https://github.com/SciTools/iris/pull/6084#discussion_r1691122976), almost all the work is done when calling the interpolator (within `RectilinearInterpolator._interpolate`) so there seem to be no actual benefit in constructing and reusing the interpolator objects? And I think this is also true for the regridder, for which all the work seems to done within [`RectilinearRegridder._regrid`](https://github.com/SciTools/iris/blob/67d0b0771834510b782ca57af9e445f6e2e21427/lib/iris/analysis/_regrid.py#L608)?  

PS: the linkchecker action seems to fail for something unrelated - the website https://fesom.de (linked [here](https://github.com/SciTools/iris/blob/67d0b0771834510b782ca57af9e445f6e2e21427/docs/src/further_topics/ugrid/other_meshes.rst?plain=1#L363)) seems to be unreachable.

lib/iris/analysis/_interpolation.py
trexfeathers: Thanks very much! I can even see where your laptop had to start using swap memory on the `main` branch 😉

trexfeathers (changes requested):

docs/src/userguide/interpolation_and_regridding.rst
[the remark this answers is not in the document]
fnattino: Thanks a lot for checking the documentation updates and for sharing about the future plans for interpolation and regridding! I have added the two fixes, please let me know if there is anything left to do on this!

trexfeathers (approved):
Thanks for your patience - I have been on leave since December.

Everything looks good 🏆

PR #6088, opened 2024-07-26
Load performance improvement (ignoring UGRID)

stephenworsley:
## 🚀 Pull Request

Attempts to address performance slowdowns highlighted here #6075


pp-mo:
> Initial [benchmark results](https://github.com/SciTools/iris/pull/6088#issuecomment-2252631900) are interesting. not all straightforward. I'm going to run that again ...

WHOOPS sorry, I did that to the wrong PR

lib/iris/fileformats/cf.py
```diff
@@ -1348,6 +1353,25 @@ def __exit__(self, exc_type, exc_value, traceback):
         # When used as a context-manager, **always** close the file on exit.
         self._close()
 
+    def _has_meshes(self):
+        result = False
+        for _, variable in self._dataset.variables.items():
```
bjlittle, line 1358:
```suggestion
        for variable in self._dataset.variables.values():
```

lib/iris/fileformats/netcdf/ugrid_load.py
```diff
@@ -56,6 +56,8 @@ def _meshes_from_cf(cf_reader):
     # Mesh instances are shared between file phenomena.
     # TODO: more sophisticated Mesh sharing between files.
     # TODO: access external Mesh cache?
+    if not cf_reader._with_ugrid:
+        return {}
```
bjlittle, line 60: @stephenworsley A stylistic thing ... but it's better to only have one `return` from the function. 

So I'd suggest this like this instead:

```python
meshes = {}

if cf_reader._with_ugrid:
    mesh_vars = cf_reader.cf_group.meshes
    meshes = {
        name: _build_mesh(cf_reader, var, cf_reader.filename)
        for name, var in mesh_vars.itesm()
    }

return meshes
```

bjlittle (changes requested):
@stephenworsley Looks good! 💯 

Just a couple of comments to service, then we can bank this 👍

bjlittle (approved):

PR #6092, opened 2024-07-26
"Fixed" benchmarks testing for changes in #6077 -- version 2

pp-mo:
A trial for benchmark run on #6077, since it looks like the benchmark code in there won't run _prior_ to the PR
(a mistake!)

So, this replaces tonight's overnight run on this, which we expect to fail.


pp-mo:
Initial [benchmark results](https://github.com/SciTools/iris/pull/6092#issuecomment-2252654582) are interesting. not all straightforward.
I'm going to run that again ...

pp-mo:
OK I think the results make sense.  The surprising `track_addedmem_netcdf_save(600, True)` one did not repeat.
The `time_common_metadata` makes some sense, as the module has grown from 7 to 9 classes, all about the same size.

PR #6119, opened 2024-08-15
Add format agent copy and equality methods

### LINKED ISSUE #6108: Comparisons for FormatAgent()
## ✨ Feature Request

Allow objects of type FormatAgent to be compared to each other to see if they are equal. For example:
```
>>>import iris.fileformats
>>>from copy import deepcopy
>>>agent_copy = deepcopy(iris.fileformats.FORMAT_AGENT)
>>>agent_copy == iris.fileformats.FORMAT_AGENT
True
```
Currently the result would be false.

## Motivation
It would be useful for testing purposes.

## Additional context
<details>
<summary>Use Case</b></summary>
ANTS uses iris.load and modifies iris.fileformats.FORMAT_AGENT to modify some FormatSpecifications and add a few more. We want it to do this within a context manager so that iris behaviour is not modified by importing ants. To test this we need to test that once outside of the context manager, the FORMAT_AGENT is the same as it was before the context manager was called. This requires comparing the FORMAT_AGENT to a copy of itself.
</details>

jrackham-mo:
### Description

Closes #6108.
Adds a copy method to `FormatAgent`, which creates a new `FormatAgent` from the format specs. Also adds an equality method which compares the format specs.

The use case for this is to temporarily modify the `FORMAT_AGENT` in a custom load, and then restore it so that we don't change standard `iris.load` behaviour outside of the custom load.

### DIFF
--- a/docs/src/whatsnew/latest.rst
+++ b/docs/src/whatsnew/latest.rst
@@ -30,7 +30,9 @@ This document explains the changes made to Iris for this release
 ✨ Features
 ===========
 
-#. N/A
+#. `@jrackham-mo`_ added :meth:`~iris.io.format_picker.FormatAgent.copy` and
+   equality methods to :class:`iris.io.format_picker.FormatAgent`, as requested
+   in :issue:`6108`.
 
 
 🐛 Bugs Fixed
@@ -79,7 +81,7 @@ This document explains the changes made to Iris for this release
     Whatsnew author names (@github name) in alphabetical order. Note that,
     core dev names are automatically included by the common_links.inc:
 
-
+.. _@jrackham-mo: https://github.com/jrackham-mo
 
 
 .. comment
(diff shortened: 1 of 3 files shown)


docs/src/whatsnew/latest.rst
```diff
@@ -30,7 +30,9 @@ This document explains the changes made to Iris for this release
 ✨ Features
 ===========
 
-#. N/A
+#. `@jrackham-mo`_ added :meth:`~iris.io.format_picker.FormatAgent.copy` and
+   equality methods to :class:`iris.io.format_picker.FormatAgent`, as requested
+   in :issue:`6108`.
```
ESadek-MO, line 35: Could you also add a link to this pull request here for ease of access?

```suggestion
   in :issue:`6108`, actioned in :pull:`6119`.
```

ESadek-MO (changes requested):
Thanks @jrackham-mo, code all looks good to me, just a minor change in the whastnew and should be good to go!

docs/src/whatsnew/latest.rst
jrackham-mo: Thanks @ESadek-MO that should now be fixed

ESadek-MO (approved):
lgtm, thanks!

Issue #4757, opened 2022-05-19
Ensure MeshCoord points and bounds always agree with the Mesh (while keeping Mesh mutable)


stephenworsley:
## 📰 Custom Issue

See also #4749 for a simpler fix and a description of the problem.

Currently, both the MeshCoord points and bounds and the data on the mesh they refer to are all mutable. In the case where the Mesh has already been realised before the MeshCoord has been created, the points will have the same underlying array as the Mesh data so changing one will change the other, however this is not true in all other cases. Furthermore, when the Mesh is lazy, all MeshCoords that derive from it will instantiate their own points and bounds so that if all derived MeshCoords are realised, the same data will exist in different copies, taking up more memory than may be necessary.

A possible suggestion to improve matters:
- Make points and bounds on MeshCoords read only properties.
- Have the points and bounds properties fetch data from the Mesh object. This may involve storing a private `_mesh_bounds` array on the Mesh when called for.

This approach would improve the memory performance when referencing the same Mesh multiple times and, unlike the solution proposed in #4749, may also allow the safe editing of the Mesh.

PR #6125, opened 2024-08-19
added functional timestamps

ESadek-MO:
### Description

---

---

### DIFF
--- a/lib/iris/mesh/components.py
+++ b/lib/iris/mesh/components.py
@@ -1139,6 +1139,10 @@ class MeshXY(Mesh):
         return self._connectivity_manager.all_members
 
     @property
+    def timestamp(self):
+        return MAX(self._coord_manager.timestamp, self._connectivity_manager.timestamp)
+
+    @property
     def all_coords(self):
         """All the :class:`~iris.coords.AuxCoord` coordinates of the :class:`MeshXY`."""
         return self._coord_manager.all_members
@@ -2001,6 +2005,7 @@ class _Mesh1DCoordinateManager:
         # optional coordinates
         self.edge_x = edge_x
         self.edge_y = edge_y
+        self.timestamp = datetime.now()
 
     def __eq__(self, other):
         # TBD: this is a minimalist implementation and requires to be revisited
@@ -2031,6 +2036,7 @@ class _Mesh1DCoordinateManager:
         return f"{self.__class__.__name__}({', '.join(args)})"
 
     def _remove(self, **kwargs):
+        self.timestamp = datetime.now()
         result = {}
         members = self.filters(**kwargs)
 
@@ -2045,6 +2051,7 @@ class _Mesh1DCoordinateManager:
         return result
 
     def _setter(self, element, axis, coord, shape):
+        self.timestamp = datetime.now()
         axis = axis.lower()
         member = f"{element}_{axis}"
 
@@ -2138,6 +2145,7 @@ class _Mesh1DCoordinateManager:
         self._setter(element="node", axis="y", coord=coord, shape=self._node_shape)
 
     def _add(self, coords):
+        self.timestamp = datetime.now()
         member_x, member_y = coords._fields
 
         # deal with the special case where both members are changing
@@ -2377,6 +2385,7 @@ class _MeshConnectivityManagerBase(ABC):
         self.ALL = self.REQUIRED + self.OPTIONAL
         self._members = {member: None for member in self.ALL}
         self.add(*connectivities)
+        self.timestamp = datetime.now()
 
     def __eq__(self, other):
         # TBD: this is a minimalist implementation and requires to be revisited
@@ -2423,6 +2432,7 @@ class _MeshConnectivityManagerBase(ABC):
         # manager.
         # No warning is raised for duplicate cf_roles - user is trusted to
         # validate their outputs.
+        self.timestamp = datetime.now()
         add_dict = {}
         for connectivity in connectivities:
             if not isinstance(connectivity, Connectivity):
@@ -2564,6 +2574,7 @@ class _MeshConnectivityManagerBase(ABC):
         contains_edge=None,
         contains_face=None,
     ):
+        self.timestamp = datetime.now()
         removal_dict = self.filters(
             item=item,
             standard_name=standard_name,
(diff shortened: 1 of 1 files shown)


lib/iris/mesh/components.py
```diff
@@ -2808,6 +2752,31 @@ def location(self):
...
+    def bounds(self):
+        if self.timestamp < self._mesh.timestamp or self.timestamp is None:
+            self._load_points_and_bounds()
+        return self.bounds
+
+    @bounds.setter
+    def bounds(self, value):
+        if value:
+            msg = "Cannot set 'bounds' on a MeshCoord."
+            raise ValueError(msg)
```
trexfeathers, line 2778: Gonna need some more nuance somewhere. Calling `super.__init__()` runs the below line:

https://github.com/SciTools/iris/blob/7b32abe214330d294b7f540c8774a2fca668781c/lib/iris/coords.py#L1509

This is unrelated to WHEN we call `super().__init__()` (a separate thread), because even if we change that we will still need to call it somewhere during `MeshCoord.__init__()`.

lib/iris/mesh/components.py
trexfeathers, line 1: Need to import `datetime`!

lib/iris/mesh/components.py
```diff
@@ -3017,6 +2986,70 @@ def summary(self, *args, **kwargs):
...
+                    ]
+                    msg = (
+                        f"Node coordinate {node_coord!r} disagrees with the "
+                        f"{location} coordinate {location_coord!r}, "
+                        f'in having a "{key}" value of {nodes_value} '
+                        f"instead of {bounds_value}."
+                    )
+                    raise ValueError(msg)
+            super().__init__(points, bounds=bounds, **use_metadict)
+            self.timestamp = self._mesh.timestamp
```
trexfeathers, line 3051:
```suggestion
            self.timestamp = self.mesh.timestamp
```

If there is a `mesh` property then that's how we're expected to access it unless there is a good reason not to. This is because there might be some important post-processing before the mesh is returned to us.

lib/iris/mesh/components.py
```diff
@@ -3017,6 +2986,70 @@ def summary(self, *args, **kwargs):
             result = "\n".join(lines)
         return result
 
+    def _load_points_and_bounds(self):
```
trexfeathers, line 2989: All references to `location` need to be `self.location` or this won't be able to run.

trexfeathers: Ditto `axis` -> `self.axis`.

lib/iris/mesh/components.py
```diff
@@ -2808,6 +2752,31 @@ def location(self):
...
+    @property
+    def points(self):
+        """The coordinate points values as a NumPy array."""
+        if self.timestamp < self._mesh.timestamp or self.timestamp is None:
+            self._load_points_and_bounds()
+        return self._values
+
+    @points.setter
+    def points(self, value):
+        if value:
```
trexfeathers, line 2764: Can't use the truthy syntax here because it trips over the ambiguous truth value of array problem.

lib/iris/mesh/components.py
```diff
@@ -2808,6 +2752,31 @@ def location(self):
...
+
+    @property
+    def bounds(self):
+        if self.timestamp < self._mesh.timestamp or self.timestamp is None:
+            self._load_points_and_bounds()
+        return self.bounds
+
+    @bounds.setter
+    def bounds(self, value):
+        if value:
```
trexfeathers, line 2776: Can't use the truthy syntax here because it trips over the ambiguous truth value of array problem.

trexfeathers (changes requested):
Thanks @ESadek-MO 

I'm confident that you have captured all the manager methods that need to update the timestamp.

Some requests for you

lib/iris/mesh/components.py
```diff
@@ -3017,6 +2986,70 @@ def summary(self, *args, **kwargs):
...
+                        fix_repr(val) for val in (nodes_value, bounds_value)
+                    ]
+                    msg = (
+                        f"Node coordinate {node_coord!r} disagrees with the "
+                        f"{location} coordinate {location_coord!r}, "
+                        f'in having a "{key}" value of {nodes_value} '
+                        f"instead of {bounds_value}."
+                    )
+                    raise ValueError(msg)
+            super().__init__(points, bounds=bounds, **use_metadict)
```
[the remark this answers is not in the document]
trexfeathers: I've had a look through the parent 'stack' and I reckon if you can work out a way to update the points, the bounds, and the metadata ([`MeshCoordMetadata`](https://scitools-iris.readthedocs.io/en/stable/generated/api/iris.common.metadata.html#iris.common.metadata.MeshCoordMetadata)) then you'll be in the clear.

lib/iris/mesh/components.py
```diff
@@ -2724,73 +2741,7 @@ def __init__(
...
-                    )
-                    raise ValueError(msg)
-
-        # Don't use 'coord_system' as a constructor arg, since for
-        # MeshCoords it is deduced from the mesh.
-        # (Otherwise a non-None coord_system breaks the 'copy' operation)
-        use_metadict.pop("coord_system")
-
-        # Call parent constructor to handle the common constructor args.
+        points, bounds = self._load_points_and_bounds()
```
trexfeathers: I think you forgot to unpack the third value that is returned by `_load_points_and_bounds()`.

lib/iris/mesh/components.py
```diff
@@ -2808,6 +2759,31 @@ def location(self):
...
+    def bounds(self):
+        if self.timestamp < self._mesh.timestamp or self.timestamp is None:
+            _, self.bounds, _ = self._load_points_and_bounds()
+        return super.bounds
+
+    @bounds.setter
+    def bounds(self, value):
+        if len(value) > 0 and self.bounds:
+            msg = "Cannot set 'bounds' on a MeshCoord."
+            raise ValueError(msg)
```
trexfeathers: These two together create some kind of infinite loop, since the setter accesses the getter then the getter accesses the setter...

Also when I tested this, asking for `my_coord.bounds` returns `None`, but perhaps that problem will go away once the above problem is fixed.

lib/iris/mesh/components.py
```diff
@@ -2808,6 +2759,31 @@ def location(self):
     def axis(self):
         return self._metadata_manager.axis
 
+    @property
+    def points(self):
+        """The coordinate points values as a NumPy array."""
+        if self.timestamp < self.mesh.timestamp or self.timestamp is None:
+            self._values, _, _ = self._load_points_and_bounds()
```
trexfeathers: This line updates `self.timestamp`. This means that accessing `points` first will prevent `bounds` from being updated, and the same in reverse.

It would be safer to update both `points` and `bounds` when EITHER is accessed.

lib/iris/mesh/components.py
```diff
@@ -3017,6 +2993,69 @@ def summary(self, *args, **kwargs):
...
 
+    def _load_points_and_bounds(self):
+        points, bounds = self._construct_access_arrays()
+        if points is None:
+            # TODO: we intend to support this in future, but it will require
+            #  extra work to refactor the parent classes.
+            msg = "Cannot yet create a MeshCoord without points."
+            raise ValueError(msg)
+
+        # Get the 'coord identity' metadata from the relevant node-coordinate.
```
trexfeathers, line 3004: Something missed in #4757 and in previous discussions: the metadata such as `units`, `standard_name`, etcetera is also subject to change.

Thankfully it is quick to recalculate, so we don't need to protect it behind a date check. But we do need a way to make sure it is recalculated on the fly.

Metadata classes a tricksy and it's probably best if we don't make any changes there. Instead we can put `self._metadata_manager` behind a `@property' so that we can regenerate it every time before returning it:


#### 1

```diff
- self._metadata_manager = metadata_manager_factory(MeshCoordMetadata)
+ self._some_other_name = metadata_manager_factory(MeshCoordMetadata)
```

#### 2

```python
@property
def _metadata_manager(self):
    # An explanatory comment.
    
    self._some_other_name.standard_name = something
    # Etcetera for all standard coord metadata
    # THIS INCLUDES DETERMINING THE CORRECT VALUE, AS IN
    #  THE CURRENT BLOCK WITHIN _load_points_and_bounds

    return self._some_other_name
```

#### 3

I don't _think_ this will be necessary since I don't _think_ there will be any calls to set `self._metadata_manager`?

```python
@_metadata_manager.setter
def _metadata_manager(self, value):
    self._some_other_name = value
```

trexfeathers (changes requested):

ESadek-MO:
This PR was old, out of date and rather difficult to debug, so I've refreshed it in #6405

PR #6135, opened 2024-09-02
Proj ci fix WIP

HGWright:
### Description

---
[Consult Iris pull request check list]( https://scitools-iris.readthedocs.io/en/latest/developers_guide/contributing_pull_request_checklist.html)

---
Add any of the below labels to trigger actions on this PR:

- https://github.com/SciTools/iris/labels/benchmark_this

PR #6141, opened 2024-09-04
Lockfiles debug

pp-mo:
Actually an attempt to dig a bit deeper into the #6096 #6135 problems


pp-mo:
The env dump in the py3.12 tests went like this, in part ...
```
2024-09-04T16:44:11.9294931Z nox > python -m pip install --no-deps --editable .
2024-09-04T16:44:18.3399127Z nox > echo '"NOX TEST ENV DUMP == START "'
2024-09-04T16:44:18.3409259Z "NOX TEST ENV DUMP == START "
2024-09-04T16:44:18.3413493Z nox > env 
2024-09-04T16:44:18.3423399Z SYSTEMROOT=
2024-09-04T16:44:18.3424153Z SELENIUM_JAR_PATH=/usr/share/java/selenium-server.jar
2024-09-04T16:44:18.3424913Z CONDA=/home/runner/miniconda3
2024-09-04T16:44:18.3425544Z GITHUB_WORKSPACE=/home/runner/work/iris/iris
2024-09-04T16:44:18.3426381Z JAVA_HOME_11_X64=/usr/lib/jvm/temurin-11-jdk-amd64
2024-09-04T16:44:18.3427105Z CONDA_EXE=/home/runner/miniconda3/bin/conda
2024-09-04T16:44:18.3427693Z _CE_M=
2024-09-04T16:44:18.3428105Z ENV_NAME=ci-tests
  . . .
2024-09-04T16:44:18.3490069Z PROJ_DATA=/home/runner/miniconda3/envs/ci-tests/share/proj
 . . .

2024-09-04T16:44:18.3522405Z PROJ_NETWORK=ON
 . . .
2024-09-04T16:44:18.3523626Z NOX_CURRENT_SESSION=tests
2024-09-04T16:44:18.3524201Z TMPDIR=/home/runner/work/iris/iris/.nox/tests/tmp
2024-09-04T16:44:18.3524969Z NPY_DISABLE_CPU_FEATURES=AVX512F,AVX512CD,AVX512_SKX
2024-09-04T16:44:18.3525782Z nox > echo '".. == END NOX TEST ENV DUMP."'
2024-09-04T16:44:18.3526703Z nox > pytest -n auto --cov=lib/iris --cov-report=xml lib/iris/tests
2024-09-04T16:44:18.3527501Z ".. == END NOX TEST ENV DUMP."
```

So, I have a string suspicion that the `PROJ_DATA=/home/runner/miniconda3/envs/ci-tests/share/proj` here 
**is** the key problem.

It does seem from the error messages that this key variable is not re-set to the correct path for the installed test environment, as it should have been.  E.G. the message :
```

We've seen

**Comment by pp-mo** [comment]
Also tested : a "minimal" noxfile like this :
```
import nox

#: Default to reusing any pre-existing nox environments.
nox.options.reuse_existing_virtualenvs = False  # was True for Iris


def show_env(session: nox.sessions.Session):
    session.run("echo", '"NOX TEST LIVE-ENV DUMP == START "', external=True)
    session.run("env", external=True)
    session.run("echo", '".. == END NOX TEST LIVE-ENV DUMP."', external=True)
    session.run("echo", 'NOX TEST session env."', external=True)
    msg = "\n".join(("  " + str(item)) for item in session.env.items())
    print(msg)
    session.run("echo", '.. = END NOX TEST session env."', external=True)


@nox.session(venv_backend="conda")
def basic(session: nox.sessions.Session):
    show_env(session)

@nox.session(venv_backend="conda")
def withproj(session: nox.sessions.Session):
    session.conda_install("proj", "pyproj")
    show_env(session)
```

### Results
Too long+tedious to copy here, but it suffices to say that PROJ_DATA is not set in the environment where proj is installed.

Checking the nox code, I believe that the [`Session.conda_install()`](https://github.com/wntrblm/nox/blob/e0868b6c092c80837f86463a54d34379851d43d7/nox/sessions.py#L603) call eventually calls popen, setting any specific variables from `session.env`, but there is **_no_** effort to capture any env changes that the install makes. 
I have also checked that the user "conda install" **does** set env variables, i.e. effectively it actions activation scripts just like "conda activate" does.

So I conclude that `nox.Session.conda_install()` just **_does not support_** additional activation steps.
This is how an unsuitable PROJ_DATA setting manages to upset the installation of a _later_ version of `proj`.

However, consulting also our [SSS modulefiles](https://github.com/MetOffice/ssstack/tree/2024_08_27/modulefiles/central), which likewise include variables from activation scripts so as to mimic a "conda activate", it seems that PROJ_DATA is quite possibly the _only_ one which is likely to cause trouble in this way.  So in practice it's a marginal problem.

In this specific case, another way to fix it could be to set `PROJ_DATA="invalid"`, since Proj seems to ignore any value which is not a suitable filepath.  I think you can do that by setting `session.env["PROJ_DATA"] = "invalid"` in the nox code

PR #6154, opened 2024-09-30
configure dependabot for github composite actions

bjlittle:
### Description

This pull-request updates the `dependabot` configuration to include coverage of our composite GHAs.

Note that, `dependabot` has not updated the [actions/cache](https://github.com/actions/cache) GHA version used by the composite actions, which has been at `v4.0.2` since 19 Mar 2024.

As a result, the composite actions have been using a deprecated version of `node.js` and forced to run on `node20`.

---

---

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -6,7 +6,9 @@ version: 2
 updates:
 
   - package-ecosystem: "github-actions"
-    directory: "/"
+    directories:
+      - "/"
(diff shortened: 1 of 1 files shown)


bjlittle:
Let's not manually update the `actions/cache` version in the composite actions, and leave it to `dependabot`, which is a good test of this PR change.

pp-mo (approved):
Look good, as you say.
Thanks for chasing this down !

Issue #4039, opened 2021-02-26
Which statistics in `iris.analysis` are lazy?


stefsmeets:
## 📰 Custom Issue

Hi everyone, we are currently working on a feature to make our multimodel calculations lazy in ESMValTool by depending on `iris.analysis` to perform the calculations (https://github.com/ESMValGroup/ESMValCore/pull/968). The [documentation](https://scitools-iris.readthedocs.io/en/latest/generated/api/iris/analysis.html) states that `MEAN`, `STD_DEV` and `VARIANCE` already have lazy implementations via `dask`. 

Looking through the code, I have noticed that `iris.analysis.MIN` and `iris.analysis.MAX` also have lazy functions associated with them, but these are not mentioned in the documentation as being lazy. I'm wondering if I'm missing something or if this information is not yet available in the documentation.

We would be very interested to also make some of the other statistics lazy on our side, i.e. `MEDIAN` and `PERCENTILE`, which also have implementations available via [dask](https://docs.dask.org/en/latest/array-api.html).

rcomer:
Hi @stefsmeets, thanks for this.  I agree we ought to have something in the docstrings that tells us which aggregators are lazy.  We have an open issue about doing this for functions generally (#3292), but nobody has got to it yet.

For percentiles, I have an open PR at #3901.  I’d welcome any feedback on that as I’m pretty new to dask.

rcomer:
It looks like `dask.array.median` is using `numpy.median` under the hood, so doesn't respect masks:

```python
import numpy.ma as ma
import dask.array as da

arr = ma.array(range(4), mask=[0,0,0,1])
print(ma.median(arr))

larr = da.from_array(arr)
print(da.median(larr, axis=0).compute())
```

Output:
```
1.0
1.5
```

So I think we would need something extra to make lazy median consistent with our existing median aggregator.

stefsmeets:
Hi @rcomer , I just noticed that a `nanmedian` function exists in [dask](https://docs.dask.org/en/latest/array-api.html#dask.array.nanmedian). Would this be a way to make the median operation lazy in iris?

rcomer:
Hi @stefsmeets, yes it looks like that should work in principle.  Something like

```python
import numpy as np
import numpy.ma as ma
import dask.array as da

def lazy_median(array, axis):
    array = array.astype(np.float_)
    nan_array = da.ma.filled(array, np.nan)
    median = da.nanmedian(nan_array, axis)
[...]

PR #6167, opened 2024-10-08
Lazy median aggregator

fnattino:
### Description

Adding the lazy implementation of the median aggregator, following the implementation discussed in https://github.com/SciTools/iris/issues/4039 . 

It is true that the issue above suggests that there was not much interest for a lazy median, but it looked like it would need only little effort. And we thought it could still be useful to some ESMValTool users who might not be aware of the workaround of using the 50th percentile.

CC: @bouweandela


bouweandela:
Ping @SciTools/peloton

ESadek-MO (commented):
Thanks @fnattino!

All the code looks great, happy with this. I've rerun the tests, as there was an issue with readthedocs that we've since fixed, so assuming they pass, I'm happy with this. 

The only thing to say is that we're now trying to ensure any new tests are written in pytest, using the _shared_utils file for all conveniences. That said, I believe this was put in place after you pushed this PR up, so the fault lies on me for my late review! If you fancy rewriting them in pytest, ideal, if not I'm happy to do so myself once this is merged in.

fnattino:
Thanks for the review @ESadek-MO ! No problem, I have now updated the test module to pytest. Please let me know whether there is anything else that should still be addressed here!

ESadek-MO (approved):
Thanks @fnattino, looks good to me!

Issue #6165, opened 2024-10-07
Workaround for time-dependent hybrid height


pp-mo:

pp-mo:
## Comments following above
> @stephenworsley It looks like derived coordinates are being handled by new_axis and concatenate
> does the result look correct to you (Matt) ?

> > (Matt) Looks good to me -- interesting to see that the altitude data is also lazy

pp-mo:
## Further comments
@pp-mo I tried adding a third month (2063oct + 2063dec + 2064 jan).
Then you get one cube (oct + dec) with 2 timepoints, and 1 cube with a scalar time (jan).
So, I applied a transform like 
```
if not cube.coord_dims('time'):
        cube = new_axis(cube, scalar_coord="time", expand_extras=("surface_altitude", "forecast_period")) 
```
But the problem is, the 2-timepoint cube already **_has_** a time dimension, but the surface_altitude naturally does not map to it.
So you can't concatenate them after all.
Those cubes cubes like (**before** new_axis processing)...
```
 ...
Cube # 1
x_wind / (m s-1)                    (latitude: 144; longitude: 192)
    Dimension coordinates:
        latitude                             x               -
        longitude                            -               x
    Auxiliary coordinates:
        surface_altitude                     x               x
    Derived coordinates:
        altitude                             x               x
    Scalar coordinates:
        forecast_period             1365480.0 hours, bound=(1365120.0, 1365840.0) hours
        forecast_reference_time     2006-01-01 00:00:00
        level_height                10.000004 m, bound=(0.0, 19.999998) m
        model_level_number          1
        sigma                       0.9988703, bound=(1.0, 0.9977413)
        time                        2164-01-16 00:00:00, bound=(2164-01-01 00:00:00, 2164-02-01 00:00:00)
```
so **After** new-axis processing
```
Cube # 0
x_wind / (m s-1)                    (time: 2; latitude: 144; longitude: 192)
    Dimension coordinates:
        time                             x            -               -
        latitude                         -            x               -
        longitude                        -            -               x
    Auxiliary coordinates:
        forecast_period                  x            -               -
        surface_altitude                 -            x               x
    Derived coordinates:
        altitude                         -            x               x
    Scalar coordinates:
        forecast_reference_time     2006-01-01 00:00:00
        level_height                10.000004 m, bound=(0.0, 19.999998) m
        model_level_number          1
        sigma                       0.9988703, bound=(1.0, 0.9977413)
...
Cube # 1
[...]

Issue #5369, opened 2023-07-06
Time dimension in hybrid height


trexfeathers:
## ✨ Feature Request

[As described in CF conventions](https://cfconventions.org/cf-conventions/cf-conventions.html#atmosphere-hybrid-height-coordinate)

## Motivation

From @matthew-mizielinski. Considering how to represent hybrid height over glaciers, where the orography moves/changes over time.

## Additional context

@pp-mo expects problems with FF or PP loading, as it would require a specific sequence of merge steps.

<details>
<summary>Click to expand <b>this section...</b></summary>

```
Please add additional verbose information in this section e.g., references, screenshots, listings etc
```
</details>

```[tasklist]
### Tasks
- [ ] https://github.com/SciTools/iris/issues/6162
- [ ] https://github.com/SciTools/iris/issues/6163
- [ ] https://github.com/SciTools/iris/issues/6203
- [ ] https://github.com/SciTools/iris/issues/6204
```

trexfeathers:
@matthew-mizielinski has confirmed that data like this currently generates 1 `Cube` for each time point, rather than a single `Cube` with a time dimension.

trexfeathers:
It will be problematic for UK Met Office strategy (climate - IPCC) if this missed the 3.11 release.

stephenworsley:
I believe it is currently possible to construct a hybrid height coordinate that varies over time. What is not possible is to merge multiple 2D cubes with varying orographies together. This would require a substantial change to merge behaviour. I suspect this may be covered by #5375 which has been a particularly stubborn issue to untangle.

trexfeathers:
> I believe it is currently possible to construct a hybrid height coordinate that varies over time. What is not possible is to merge multiple 2D cubes with varying orographies together. This would require a substantial change to merge behaviour. I suspect this may be covered by #5375 which has been a particularly stubborn issue to untangle.

@stephenworsley let us know what you need. If necessary we have a whole team of developers (given the strategic importance of this).

stephenworsley:
@matthew-mizielinski minimal test data would absolutely be appreciated, and yes, I think it would be good to set up a discussion when possible.

stephenworsley:
One possible idea for resolving the merge issue:

[...]

PR #6168, opened 2024-10-09
Enable factory references to create new dimensions on load.

pp-mo:
Relates to #5369 , #6165


pp-mo:
Now updated with prototype loading-control object
NOTE: 
  * no proper tests yet
  * work needed to implement the loading policy in load_cubes/load_cube

lib/iris/fileformats/rules.py
```diff
@@ -345,6 +376,34 @@ def _resolve_factory_references(
...
+
+    return cube
+
+
+class MultipleReferenceFieldDetector(threading.local):
+    def __init__(self):
+        self.found_multiple_refs = False
+
+
+# A single global object (per thread) to record whether multiple reference fields
```
stephenworsley, line 400: If there is an instance per thread, is there any risk that behaviour is dependant on how multiprocessing is done? Or is the relevant loading done entirely within a single thread?

lib/iris/fileformats/rules.py
```diff
@@ -345,6 +376,34 @@ def _resolve_factory_references(
...
+# (e.g. time-dependent orography, or surface pressure fields) have been detected during
+# the latest load operation.
+# This is used purely to implement the iris.LOAD_POLICY.multiref_triggers_concatenate
+# functionality.
+_MULTIREF_DETECTION = MultipleReferenceFieldDetector()
```
stephenworsley, line 405: I don't understand why this would be instantiated at the module level rather than during a load call. Would this not change your loading behaviour depending on what files you have already loaded during your session?

stephenworsley (commented):
Looking good, just a couple comments.

lib/iris/__init__.py
```diff
@@ -303,7 +312,118 @@ def _load_collection(uris, constraints=None, callback=None):
...
+            f"use_concatenate={self.use_concatenate}, "
+            f"use_merge={self.use_merge}, "
+            f"cat_before_merge={self.cat_before_merge}, "
+            f"repeat_until_done={self.repeat_until_done}"
+            ")"
+        )
+        return msg
+
+    def copy(self):
+        return LoadPolicy(**{key: getattr(self, key) for key in self._allkeys})
```
trexfeathers, line 358: I think data classes give you some or all of this, and you can still add your own methods in addition.

lib/iris/__init__.py
```diff
@@ -303,7 +312,118 @@ def _load_collection(uris, constraints=None, callback=None):
...
+class LoadPolicy(threading.local):
+    """Object defining a general loading strategy."""
+
+    _allkeys = (
+        "support_multiple_references",
+        "multiref_triggers_concatenate",
+        "use_concatenate",
+        "use_merge",
+        "cat_before_merge",
+        "repeat_until_done",
```
trexfeathers, line 324: I like the idea of giving users more controls to play with, but having 6 booleans is very confusing, so I'm keen to look for ways it could be simpler to understand.

1. It doesn't look like `support_multiple_references` is necessary
2. `use_concatenate`, `use_merge` and `cat_before_merge` could become some sort of controllable sequence e.g. `{"c", "m"}` / `{"m"}` / `{"m", "c"}`.

lib/iris/__init__.py
```diff
@@ -303,7 +312,118 @@ def _load_collection(uris, constraints=None, callback=None):
...
+
+
+LOAD_POLICY = LoadPolicy()
+LOAD_POLICY_LEGACY = LoadPolicy()
+LOAD_POLICY_RECOMMENDED = LoadPolicy(
+    support_multiple_references=True, multiref_triggers_concatenate=True
+)
+LOAD_POLICY_COMPREHENSIVE = LoadPolicy(
+    support_multiple_references=True, use_concatenate=True, repeat_until_done=True
+)
```
trexfeathers, line 394: Might be simpler to follow if these 'presets' were all part of another object. Could they be class attributes e.g. `LoadPolicy.recommended`? It would be good if the only global constant were `LOAD_POLICY`.

lib/iris/__init__.py
```diff
@@ -303,7 +312,118 @@ def _load_collection(uris, constraints=None, callback=None):
...
+
+    def copy(self):
+        return LoadPolicy(**{key: getattr(self, key) for key in self._allkeys})
+
+    @contextlib.contextmanager
+    def context(self, policy=None, **kwargs):
+        """Return context manager for temporary options.
+
+        Modifies the given parameters within a context, for the active thread.
+        """
```
trexfeathers, line 365: Could do with more detail and examples - I'm struggling to picture use cases.

pp-mo: Agreed. This is not intended to be complete yet -- still just draft ideas

trexfeathers (commented):
I've had my rough look around.

I think the user experience is almost there. Seems like a good compromise that still holds the user's hand as much as possible (they don't write the code themselves), while still offering more customisation than currently.

The developer experience feels very complicated. I think this is worth some attention, since the developer experience around loading is _already_ too complicated and I have noticed can be a big barrier for less experienced people. Afraid I don't have any immediate suggestions, only that it's worth some time.

Thanks!

pp-mo:
> @trexfeathers  I think the user experience is almost there. Seems like a good compromise that still holds the user's hand as much as possible (they don't write the code themselves), while still offering more customisation than currently.

I'm not providing additional docstrings yet, but it's clearly needed (if not a userguide section?) 
Likewise what @trexfeathers already said : the new functionality badly needs usage examples
-- on the TODO list

pp-mo:
> @trexfeathers  The developer experience feels very complicated. I think this is worth some attention, since the developer experience around loading is _already_ too complicated and I have noticed can be a big barrier for less experienced people. Afraid I don't have any immediate suggestions, only that it's worth some time.

You're totally right, and I'm currently planning to remove all the pieces here to a single sub-module "iris.io.loading", which I think will help a bit.  
I have a draft of that, and it gathers the _CubeFilterCollection and _CubeFilter classes, supporting routines "_generate_cubes" and "_load_collection", all the new "policy" code **_and_** all the load_xxx functions.
Along with that, I am considering removing the _CubeFilterCollection and _CubeFilter classes altogether as I think they may obscure rather than help.

However, I want to add some proper testcases before I attempt a lift-and-shift, as otherwise some breakages might go unnoticed.

lib/iris/__init__.py
pp-mo: OK, as discussed offline, I think we can simplify all this with benefits.

lib/iris/__init__.py
pp-mo: Agreed, I don't like this much.
Under review, watch this space!

lib/iris/__init__.py
```diff
@@ -303,7 +312,118 @@ def _load_collection(uris, constraints=None, callback=None):
...
+    if not policy.use_concatenate and policy.multiref_triggers_concatenate:
+        from iris.fileformats.rules import _MULTIREF_DETECTION
+
+        if _MULTIREF_DETECTION.found_multiple_refs:
+            policy = policy.copy()
+            policy.use_concatenate = True
+    return policy
+
+
+def _apply_loading_policy(cubes, policy=None):
```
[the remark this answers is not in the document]
pp-mo: Yes, I agree this could be a useful general tool in its own right.
In that case it would be useful to expose the "POLICY" settings as simple arguments here
(especially if they get simpler, as described in the "new settings arrangement").
Going forward, I'll try + make that workable.

pp-mo:
### Current best ideas for a "new control strategy"
Following offline discussion wuith @trexfeathers 

Just 3 control keywords:
  * `support_multiple_references` = True/False
     *  (as before, but now **_implies_** the old `multiref_triggers_concatenate=True` behaviour)
  * `merge_concatenate_sequence` = "m" / "c" / "mc" / "cm"
  * `repeat_until_no_changes` = True/False

Plus... 
  * combine settings in a simple dictionary or namedtuple
  * use simple constants for common choices
  * make  `support_multiple_references` **True by default**, so it ..
    * _includes_ the triggering of a final concatenate
    * delivers required solution for the key new cases (i.e. time-dependent hybrid height)
    * should deliver existing result for 99% of cases
    * for 1% of broken legacy loads, can be manually forced to "existing behaviour"

lib/iris/fileformats/rules.py
pp-mo: It's explained [here](https://github.com/SciTools/iris/blob/e52e20b5c7cba5ebec58e2af7f3029b71d7a4e41/lib/iris/__init__.py#L298-L304), that all the iris.load_xxx load functions call via `_load_collection`, which resets this.  It is admittedly a weak contract (!).
We need it to behave like the settings controls. i.e. to exist as a common (but per-thread) global reference, so it can affect low-level behaviour without the need to pass the control down through all functions in the call chain.

lib/iris/fileformats/rules.py
pp-mo: I believe the loading code is all single-threaded in operation.  It has to be since, while it has the input file open, it might not be safe to run parallel loading tasks since the file system interface may not be thread-safe (e.g. as for netcdf4-python) .

pp-mo:
### UPDATED Progress Update 2024-10-23 :

I'm now targetting a feature branch, 
so we can review the work in stages before committing to main.

The remaining planned work is :
  * [ ] move all the disparate moving parts into a "iris.io.loading" submodule
  * add tests 
     * [ ] tests for LoadPolicy behaviour (init, set + context methods)
     * [ ] tests for new "combine_cubes" user routine
        * (aside : should this maybe be cast as a Cubelist method, like merge + concatenate ?)
     * [ ]  ?possibly? separate tests for how rules.py now handles multiple references
        * TBH I think this may be adequately proved in the integration tests
  * [ ] reinstate + fixes skipped tests (i.e. the plotting crashes)
  * [ ] change actual default loading policy from "legacy" to "default" -- and fix any tests that breaks
  * [ ] add whatsnew

lib/iris/__init__.py
```diff
@@ -303,7 +315,301 @@ def _load_collection(uris, constraints=None, callback=None):
...
+        # Add a concatenate to implement the "multiref triggers concatenate" mechanism
+        from iris.fileformats.rules import _MULTIREF_DETECTION
+
+        if _MULTIREF_DETECTION.found_multiple_refs:
+            options["merge_concat_sequence"] += "c"
+
+    return combine_cubes(cubes, options, merge_require_unique=merge_require_unique)
+
+
+def load(uris, constraints=None, callback=None, policy=None):
```
pp-mo, line 612: The addition here of "policy=None" is a mistake, as it is not implemented here.
I will remove this when we've reviewed the rest.

So, we might actually add this later,
but if so it should probably have the common style now established elsewhere
 \- i.e. `options=None, **kwargs`
And it should be supported by all of `load`/`load_cube`/`load_cubes`.

lib/iris/__init__.py
```diff
@@ -303,7 +315,301 @@ def _load_collection(uris, constraints=None, callback=None):
...
+def combine_cubes(cubes, options=None, merge_require_unique=False):
+    """Combine cubes as for load, according to "loading policy" options.
+
+    Applies :meth:`~iris.cube.CubeList.merge`/:meth:`~iris.cube.CubeList.concatenate`
+    steps to the given cubes, as determined by the 'settings'.
+
+    Parameters
+    ----------
+    cubes : list of :class:`~iris.cube.Cube`
+        A list of cubes to combine.
```
stephenworsley, line 556: It looks like, as written, `cubes` would be required to be a CubeList. I think it would actually be quite neat if this was allowed to take in a list of cubes as well as a CubeList, though I think there should be some appropriate error checking logic added for this.

lib/iris/cube.py
```diff
@@ -64,6 +64,8 @@ class _CubeFilter:
     """A constraint, paired with a list of cubes matching that constraint."""
 
     def __init__(self, constraint, cubes=None):
+        from iris.cube import CubeList
```
stephenworsley, line 67: Why are we importing from `iris.cube` while within the `iris.cube` module?

lib/iris/__init__.py
```diff
@@ -303,7 +315,301 @@ def _load_collection(uris, constraints=None, callback=None):
...
+        kwargs : dict
+            Option values, as for :meth:`~LoadPolicy.set`.
+
+        Examples
+        --------
+        .. testsetup::
+
+            import iris
+            from iris import LOAD_POLICY, sample_data_path
+            path = sample_data_path("hybrid_height.nc")
```
stephenworsley, line 522: Would it be an idea to point this at the data I added for #6180?

stephenworsley (commented):
Just a couple comments on my first pass looking at this. Looking like this is going in the right direction.

lib/iris/__init__.py
pp-mo: Yes, I have just fixed this while writing tests (subsequent PR).
In future, it will convert an input ordinary list to a CubeList, and always returns a CubeList.

lib/iris/cube.py
pp-mo: Oof 🤦 
Will fix !

lib/iris/__init__.py
pp-mo: OK yes, we can update now that is in.
Will do

pp-mo:
Ok @stephenworsley I think I have covered the outstanding.  
Please check against comments + resolve if appropriate.
Let me know if+when you have more points to raise -- or not !

lib/iris/__init__.py
```diff
@@ -303,7 +315,312 @@ def _load_collection(uris, constraints=None, callback=None):
     return result
 
 
-def load(uris, constraints=None, callback=None):
+class LoadPolicy(threading.local):
+    """A container for loading strategy options.
+
+    Controls merge/concatenate usage, and the handling of cases where multiple reference
+    fields merge to define an additional dimension (e.g. a time-varying orography).
```
stephenworsley, line 322: When you say "reference fields" here, is this talking about a reference field *for a derived coordinate*? If so, it would be worth making that more explicit.

stephenworsley (changes requested):
A couple more minor comments. One thing I would like to flesh out is the difference in behaviour between the case where time varying orography is reference field for a derived coordinate and the case where there is no derived coordinate. From what I gather, the logic involved in loading involves checking derived references, but I think it could be better documented and tested what differences in behaviour there are.

lib/iris/tests/integration/varying_references/test_roundtrip_time_varying_references.py
```diff
@@ -0,0 +1,276 @@
...
+    else:
+        ref_shape = (NT, NY, NX)
+    ref_data = 1000.0 + np.arange(np.prod(ref_shape)).reshape(ref_shape)
+    if zcoord_type == "pressure" and file_extension == "grib2":
+        # values come back in Pa not hPa
+        ref_data *= 100.0
+    assert np.array_equal(ref_data, result_ref_cube.data)
+
+
+_file_formats = ["pp", "nc"]
```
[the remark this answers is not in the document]
pp-mo: I don't think so because usually in netcdf, a dataset is **_not_** split into separate fields, such as 2D slices, but retains all its necessary dimensions.
In principle, if we had different variables containing individual sections of data, with suitable additional scalar coords (e.g. time), then a merge would occur on load.  But this would be highly unusual, even if they were in separate files.
In practice, if we have data in different files it may be a candidate for _concatenate_, but I've never seen a case where a _merge_ can happen.

To be clear, I wouldn't expect any of the changes here to affect netcdf load/save in "normal" cases.
I only included "nc" as a format to confirm that it _doesn't_ impact the normal load/save operations.
The data generate by these tests will all save as netcdf in the "normal" way.

lib/iris/tests/unit/fileformats/test_load_functions.py
```diff
@@ -0,0 +1,231 @@
...
+            iris.Constraint(height=[1, 2]),
+            iris.Constraint(height=[1, 4, 5]),
+        ]
+        c12 = cu(z=[1, 2])
+        c145 = cu(z=[1, 4, 5])
+        expected_results = {
+            "load": [c12, c145],
+            "load_cube": "ValueError.*only a single constraint is allowed",
+            "load_cubes": [c12, c145],  # selected parts merge, as for load
+            "load_raw": [c1, c2, c1, c4, c5],  # THIS VERY STRANGE BEHAVIOUR!!
```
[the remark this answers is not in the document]
pp-mo: Yes it is.  See #6191 - this shows that all these tests pass with the code in current main.

lib/iris/tests/integration/varying_references/test_roundtrip_time_varying_references.py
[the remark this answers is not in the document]
pp-mo: Ok I added  https://github.com/SciTools/iris/pull/6168/commits/cb1ddff49e042705c221476137b81ee88033de9a to cover that idea.

It splits the test-data into 2D fields + saves each to a different file.
The result combines ok as long as concatenation is enabled from the outset (the auto-detect doesn't work because the multiple reference cubes can concat but not merge).

I am increasingly impressed by concatenate : it can do a lot more than I realised.

pp-mo:
### Status 2024-10-23 @1615 :
Essentials from previous comments still to resolve :
  * [explain meaning of reference fields in LoadPolicy docs](https://github.com/SciTools/iris/pull/6168#discussion_r1814616688)
  * [adjust LoadPolicy.context examples](https://github.com/SciTools/iris/pull/6168#discussion_r1814503754) 
  * [remove spurious "policy" arg in load function](https://github.com/SciTools/iris/pull/6168#discussion_r1814503754)

We think that unresolved comments by @trexfeathers are all either now done, or not important to fix

docs/src/further_topics/controlling_merge.rst
```diff
@@ -200,5 +200,9 @@ Once merged, we can now concatenate all these cubes into a single result cube, w
             source                      'Data from Met Office Unified Model'
             um_version                  '12.1'
 
-.. todo::
-    Mention the work done in #6168
+See Also
+--------
+* :func:`iris.combine_cubes` can perform similar operations automatically
```
stephenworsley, line 205: This reference should be changed now the function is private

pp-mo:
Hi @stephenworsley thanks for all your patience !
I think I have done everything I was planning now, and I _think_ it addresses all the outstanding review queries. 
If not drop me a line - I'll check in sometime tomorrow morning, and can put a little time into it if that will get it finished off.

stephenworsley (approved):
Good job! I believe this is in a releasable state now. There's a small bump in some of our loading benchmarks, but I think that is to be expected given our expanded functionality. Looking at the scaling within the benchmarks, it looks to be more overhead than an actual concern. The remaining work looks seperable, I don't believe anything here is functionally incomplete.

Issue #5970, opened 2024-05-22
Configurable colorbar in `iris.quickplot`


trexfeathers:
## ✨ Feature Request

I would like to be able to configure colorbar creation via kwargs in `iris.quickplot` functions.

## Motivation

If I want to customise a plot using `iris.quickplot`, I can modify every element after creating the plot EXCEPT the colorbar (due to how Matplotlib works).

rcomer:
I think the work here is:

1. Add a `colorbar_kwargs` keyword to the [`_label` function](https://github.com/SciTools/iris/blob/3e1e80ecdb3784f25f4f2bc03bc8e0dbb0a1afa0/lib/iris/quickplot.py#L51), which is a dictionary of keywords to use in the [`colorbar` call](https://github.com/SciTools/iris/blob/3e1e80ecdb3784f25f4f2bc03bc8e0dbb0a1afa0/lib/iris/quickplot.py#L60-L62).  If "orientation" or "drawedges" is in that dictionary, they should override the current choices.

2. Do the "plumbing" to expose `colorbar_kwargs` in every public plotting function that ultimately calls `_label`.

3. Add tests.

pp-mo:
I'm not quite sure of the exact scope or intention here, but I would probably oppose any additions to the quickplot api.
IMHO iris.quickplot is a quick + simple one-stop shop.  If you need more control, you should simply be using iris.plot instead.
If there is still a need for additional controls _there_, to do things which can't be done otherwise, then I'd say that is appropriate.

trexfeathers:
From @SciTools/peloton:

We prefer the simpler solution of 1 new parameter: `colorbar`, which defaults to `True`. If the user wants a customised colorbar then they should use `False` then add their own after creating the plot.

rcomer:
I think the work here is:

1. Add the boolean `colorbar` keyword to the [`_label` function](https://github.com/SciTools/iris/blob/3e1e80ecdb3784f25f4f2bc03bc8e0dbb0a1afa0/lib/iris/quickplot.py#L51), which controls whether the colorbar is added.

2. Do the "plumbing" to expose `colorbar` in every public plotting function that ultimately calls `_label`.

3. Add tests.

ukmo-ccbunney:
Just so I understand the problem, I assume it is not enough to have a handle to the `colorbar` instance so you can adjust stuff? E.g.:
```
cm = qplt.pcolormesh(cube)
cb = cm.colorbar

# change ticks:
[...]

PR #6169, opened 2024-10-09
`colorbar` keyword for iris.quickplot routines

ukmo-ccbunney:
### Description
New boolean `colorbar` keywork allows user to control automatic generation of a colorbar in iris.quickplot plotting routines.
Only implemented for `pcolormesh`, `colormesh` and `coutourf` as the other plotting routines do not generate a colorbar.

Fixes: #5970 

---

### DIFF
--- a/lib/iris/quickplot.py
+++ b/lib/iris/quickplot.py
@@ -48,14 +48,15 @@ def _title(cube_or_coord, with_units):
     return title
 
 
-def _label(cube, mode, result=None, ndims=2, coords=None, axes=None):
+def _label(cube, mode, result=None, ndims=2, coords=None, axes=None, colorbar=True):
     """Put labels on the current plot using the given cube."""
     if axes is None:
         axes = plt.gca()
 
     axes.set_title(_title(cube, with_units=False))
 
-    if result is not None:
+    # optional colorbar
+    if colorbar and result is not None:
         draw_edges = mode == iris.coords.POINT_MODE
         bar = plt.colorbar(
             result, ax=axes, orientation="horizontal", drawedges=draw_edges
@@ -89,12 +90,12 @@ def _label(cube, mode, result=None, ndims=2, coords=None, axes=None):
         raise ValueError(msg)
 
 
-def _label_with_bounds(cube, result=None, ndims=2, coords=None, axes=None):
-    _label(cube, iris.coords.BOUND_MODE, result, ndims, coords, axes)
+def _label_with_bounds(cube, result=None, ndims=2, coords=None, axes=None, colorbar=True):
+    _label(cube, iris.coords.BOUND_MODE, result, ndims, coords, axes, colorbar)
 
 
-def _label_with_points(cube, result=None, ndims=2, coords=None, axes=None):
-    _label(cube, iris.coords.POINT_MODE, result, ndims, coords, axes)
+def _label_with_points(cube, result=None, ndims=2, coords=None, axes=None, colorbar=True):
+    _label(cube, iris.coords.POINT_MODE, result, ndims, coords, axes, colorbar)
 
 
 def _get_titles(u_object, v_object):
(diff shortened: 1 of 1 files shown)


ukmo-ccbunney:
@trexfeathers I've added 3 new tests for this change in `test_quickplot.py::TestLabels` and generated the new images and `imagerepo.json` files. If you're happy with the tests I've added, then I'll update raise a PR for the `iris-test-data` repo.

lib/iris/quickplot.py
```diff
@@ -89,12 +90,16 @@ def _label(cube, mode, result=None, ndims=2, coords=None, axes=None):
...
+):
+    _label(cube, iris.coords.BOUND_MODE, result, ndims, coords, axes, colorbar)
 
 
-def _label_with_points(cube, result=None, ndims=2, coords=None, axes=None):
-    _label(cube, iris.coords.POINT_MODE, result, ndims, coords, axes)
+def _label_with_points(
+    cube, result=None, ndims=2, coords=None, axes=None, colorbar=True
+):
+    _label(cube, iris.coords.POINT_MODE, result, ndims, coords, axes, colorbar)
```
trexfeathers, line 102: As well as `contour()` and `contourf()`, this is also called by the `points()` function. Is there a reason why you didn't add the `colorbar` parameter there as well?

lib/iris/quickplot.py
```diff
@@ -89,12 +90,16 @@ def _label(cube, mode, result=None, ndims=2, coords=None, axes=None):
         raise ValueError(msg)
 
 
-def _label_with_bounds(cube, result=None, ndims=2, coords=None, axes=None):
-    _label(cube, iris.coords.BOUND_MODE, result, ndims, coords, axes)
+def _label_with_bounds(
+    cube, result=None, ndims=2, coords=None, axes=None, colorbar=True
+):
+    _label(cube, iris.coords.BOUND_MODE, result, ndims, coords, axes, colorbar)
```
trexfeathers, line 96: As well as `pcolor()` and `pcolormesh()`, this is also called by the `outline()` function. Is there a reason why you didn't add the `colorbar` parameter there as well?

trexfeathers (changes requested):
Thanks @ukmo-ccbunney, I agree with your logic, please do put up that iris-test-data pull request - I assume you are following this guidance: [Adding or Updating Graphics Tests — Iris 3.10.0 documentation (scitools-iris.readthedocs.io)](https://scitools-iris.readthedocs.io/en/stable/developers_guide/contributing_graphics_tests.html)

lib/iris/quickplot.py
ukmo-ccbunney: The `points` method calls `_label_with_ponts` with no "mappable" results in the `results` keyword, i.e.
https://github.com/SciTools/iris/blob/ca6470e126c97c188918d910f9732ca882b144e6/lib/iris/quickplot.py#L301

A colorbar is only generated for a plot if `result` is not `None` in the `_label` method:
https://github.com/SciTools/iris/blob/ca6470e126c97c188918d910f9732ca882b144e6/lib/iris/quickplot.py#L59-L63

I could add the `colorbar` keyword to this method (and the `outline` method mentioned in your other comment) for the sake of consistency, but it would simply be ignored/

lib/iris/quickplot.py
ukmo-ccbunney: See comment above regarding `points` method.

trexfeathers:
Oh and don't forget a What's New entry in `latest.rst`

trexfeathers:
Great! Now that iris-test-data is sorted, we need to reference [2.27](https://github.com/SciTools/iris-test-data/releases/tag/v2.27) in the below places.

While you're there, could you add a comment in `ci-tests.yml` to remind people that iris-test-data is also referenced in `benchmarks_run.yml` (they end up mis-aligned a lot!)?

https://github.com/SciTools/iris/blob/e9a73195af318baa21cd76a8245fe64f64c0926f/.github/workflows/ci-tests.yml#L52-L54

https://github.com/SciTools/iris/blob/e9a73195af318baa21cd76a8245fe64f64c0926f/.github/workflows/benchmarks_run.yml#L60-L66

trexfeathers (approved):

PR #6171, opened 2024-10-10
Update lock files, pin Cartopy!=0.23

### LINKED ISSUE #5977: Remove `GeoAxes` patch at next Cartopy release
## 📰 Custom Issue

SciTools/cartopy#2390

https://github.com/SciTools/iris/blob/57096f29f6f2c5ff278d616df5e2da4d2bab95e2/lib/iris/plot.py#L46-L60

trexfeathers:
### Description

Includes the removal of the `GeoAxes` monkey patch - closes #5977.

Forced to do this now since we haven't yet got Read the Docs locked down (SciTools/workflows#38), so that's already getting Cartopy v0.24 and therefore failing our CI.

---

---


trexfeathers:
https://github.com/SciTools/iris/pull/6171#issuecomment-2405378513

As far as I can tell: ALL collapse benchmarks have become slower. Some have not triggered the threshold. This appears to be a function of how long the benchmark takes - longer benchmarks have changed less, suggesting the regression is an overhead of some sort (the regression itself does not scale).

It's hard to tell but I believe the weighted collapse benchmarks `w_collapsed_by` have also regressed, but not by much.

The aggregated_by benchmarks appear unchanged.

There aren't many changes in the lock files, and I struggle to see how any of the changes could affect collapsing but not aggregating - they both use the same operation under the hood.

trexfeathers:
692d3332bffd4c7f44b28e611849749e4e66140f

SciTools/cartopy#2414 made a minor change that had two effects for Iris:

- Changed how plot edges might be rendered
- Changed the aspect ratio of some plots

Did the due diligence on the image tests that had failed and confirmed that there was no evidence of something malfunctioning, just a minor visual change going forward.

trexfeathers:
>As far as I can tell: ALL collapse benchmarks have become slower. Some have not triggered the threshold. This appears to be a function of how long the benchmark takes - longer benchmarks have changed less, suggesting the regression is an overhead of some sort (the regression itself does not scale).

I was hoping to replicate locally, then I could do some scalability testing, but my local results suffer from too much variance to identify regressions.

Regardless: the only action we could reasonably take would be to warn users of the regression (we can't fix it if it's in a dependency), but I propose that a regression this small, and one that seemingly does not scale, is not worth warning users about.

ESadek-MO (approved):
Happy with this, thanks @trexfeathers

PR #6180, opened 2024-10-17
Document use of new_axis to control merge

### LINKED ISSUE #6162: Document `new_axis` utility
`new_axis` could be used as a potential existing solution to #5369, but this utility isn't currently very visible.

Improved documentation, and a guiding error message, should increase visibility and offer a first step to handling time dimension in hybrid heights.

stephenworsley:
### Description

Closes #6162.
Requires https://github.com/SciTools/iris-sample-data/pull/33.
May need to be adapted to account for #6168.


docs/src/further_topics/controlling_merge.rst
```diff
@@ -0,0 +1,198 @@
...
+    >>> test_cubes.concatenate_cube()
+    Traceback (most recent call last):
+    ...
+    iris.exceptions.ConcatenateError: failed to concatenate into a single cube.
+      Scalar coordinates values or metadata differ: forecast_period != forecast_period
+
+This error alerts to us that the ``forecast_period`` coordinate is also varying over time. To get concatenation to work,
+we will have to expand the dimension of this coordinate also by passing it to the ``expand_extras`` keyword.
+
+    >>> processed_cubes = CubeList(
```
pp-mo, line 86: As this is a "retry" of the previous failed command, I'd suggest that the result should have the same name - i.e. both should be either "test_cubes" or "processed_cubes".

docs/src/further_topics/controlling_merge.rst
```diff
@@ -0,0 +1,198 @@
...
+    ... )
+    >>> print(processed_raw_cubes)
+    0: x_wind / (m s-1)                    (time: 1; latitude: 144; longitude: 192)
+    1: x_wind / (m s-1)                    (time: 1; latitude: 144; longitude: 192)
+    ...
+    73: x_wind / (m s-1)                    (time: 1; latitude: 144; longitude: 192)
+    74: x_wind / (m s-1)                    (time: 1; latitude: 144; longitude: 192)
+
+Then we merge along the ``model_level_number`` dimension. Note that merging these cubes does not affect the ``time``
+dimension since merging only applies along scalar coordinates, not dimension coordinates of length 1.
```
pp-mo, line 161:
```suggestion
Then we merge, which will create a new ``model_level_number`` dimension, since there is one cube for each level, at each time point.  Note, however, that merging these cubes does _not_ affect the ``time``
dimension, since merging only applies to scalar coordinates, not dimension coordinates of length 1.
```

I'm uncomfortable with "merge along the `model_level_number` dimension" since, of course, merge chooses the merge dimensions automatically +  we have no control over it.
I also generally don't like "merge along" because we don't have a dimension giving a direction to be "along", whereas (for me) "concatenate along" **_is_** clear + meaningful. 

Not entirely sure about my alternative proposals, though.  Maybe you can think of something better.

docs/src/further_topics/controlling_merge.rst
```diff
@@ -0,0 +1,198 @@
...
+
+    >>> merged_cubes = processed_raw_cubes.merge()
+    >>> print(merged_cubes)
+    0: x_wind / (m s-1)                    (model_level_number: 5; time: 1; latitude: 144; longitude: 192)
+    1: x_wind / (m s-1)                    (model_level_number: 5; time: 1; latitude: 144; longitude: 192)
+    ...
+    13: x_wind / (m s-1)                    (model_level_number: 5; time: 1; latitude: 144; longitude: 192)
+    14: x_wind / (m s-1)                    (model_level_number: 5; time: 1; latitude: 144; longitude: 192)
+
+Once merged, we can now concatenate these cubes together:
```
pp-mo, line 171:
```suggestion
Once merged, we can now concatenate all these cubes into a single result cube, which is what we wanted:
```

docs/src/further_topics/controlling_merge.rst
```diff
@@ -0,0 +1,198 @@
...
+    >>> from iris.cube import CubeList
+    >>> test_cubes = CubeList([new_axis(cube, scalar_coord="time", expand_extras=["surface_altitude"]) for cube in cubes])
+    >>> test_cubes.concatenate_cube()
+    Traceback (most recent call last):
+    ...
+    iris.exceptions.ConcatenateError: failed to concatenate into a single cube.
+      Scalar coordinates values or metadata differ: forecast_period != forecast_period
+
+This error alerts to us that the ``forecast_period`` coordinate is also varying over time. To get concatenation to work,
+we will have to expand the dimension of this coordinate also by passing it to the ``expand_extras`` keyword.
```
pp-mo, line 84:
```suggestion
we will have to expand the dimensions of this coordinate to include "time", by passing it also to the ``expand_extras`` keyword.
```

pp-mo (changes requested):
The process seems very solid, but I have suggested a few possible improvements for clarity.

Obviously we aren't quite ready to merge this anyway, owing to the link with #6168 as you already noted.

docs/src/further_topics/controlling_merge.rst
```diff
@@ -0,0 +1,204 @@
+.. _controlling_merge:
+
+=================
+Controlling Merge
```
pp-mo, line 4: Is it anyway better to call this "Controlling Merge and Concatenate" ?

pp-mo (changes requested):
Sorry, nearly there but there are a few problems remaining.
   * a couple of further suggestions on the diff
   * you need to include the new page in the 'further topics' TOC, since that doesn't happen automatically, and I couldn't view the built page in the docs even by substituting the page name
     I _think_ you just need to add it [here](https://github.com/SciTools/iris/blob/569a0a3aef88739d8d3bb50f4a9a1dde46ab14bd/docs/src/further_topics/index.rst?plain=1)
   * the docs build is now working, but the doctests are still failing.  I think it's a cartopy compatibility problem, which I hope may be fixed by just updating the branch, but I'll leave you to do that in case something goes wrong

pp-mo (approved):
All good now, I think !

Issue #6027, opened 2024-06-26
Cannot load cubes that contain multiple variables with array-type attributes with different sizes


schlunma:
## How To Reproduce
Steps to reproduce the behaviour:

```py
import iris
from iris.cube import Cube, CubeList
print("iris version", iris.__version__)

iris.FUTURE.save_split_attrs = True  # commenting this does not change the outcome

cubes = CubeList([
    Cube(0.0, var_name='a', attributes={'att': [0, 1]}),
    Cube(0.0, var_name='b', attributes={'att': [0, 1, 2]}),
])

path = 'data/cubes_with_attributes_bug.nc'
iris.save(cubes, path)
print("Saved", path)

loaded_cubes = iris.load(path)
```

gives

```bash
iris version 3.9.0
Saved data/cubes_with_attributes_bug.nc
Traceback (most recent call last):
  File "/home/b/b309141/scripts/iris/attributes.py", line 20, in <module>
    loaded_cubes = iris.load(path)
                   ^^^^^^^^^^^^^^^
  File "/work/bd0854/b309141/miniforge3/envs/esm/lib/python3.11/site-packages/iris/__init__.py", line 326, in load
    return _load_collection(uris, constraints, callback).merged().cubes()
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/work/bd0854/b309141/miniforge3/envs/esm/lib/python3.11/site-packages/iris/cube.py", line 126, in merged
    return _CubeFilterCollection([pair.merged(unique) for pair in self.pairs])
                                 ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/work/bd0854/b309141/miniforge3/envs/esm/lib/python3.11/site-packages/iris/cube.py", line 126, in <listcomp>
    return _CubeFilterCollection([pair.merged(unique) for pair in self.pairs])
                                  ^^^^^^^^^^^^^^^^^^^
  File "/work/bd0854/b309141/miniforge3/envs/esm/lib/python3.11/site-packages/iris/cube.py", line 85, in merged
    return _CubeFilter(self.constraint, self.cubes.merge(unique))
                                        ^^^^^^^^^^^^^^^^^^^^^^^^
  File "/work/bd0854/b309141/miniforge3/envs/esm/lib/python3.11/site-packages/iris/cube.py", line 482, in merge
[...]

PR #6181, opened 2024-10-18
Fix attribute array comparison

ukmo-ccbunney:
### Description
Fixes: #6027 

Logic for comparing arrays in attributes was relying on old numpy behaviour where an array compared to another any object would return a scalar `True` or `False`. Current numpy behaviour is to do a element-wise comparison which throws an error if the arrays are of different size.

Fix is to use the numpy function `array_eqaul`.

---

### DIFF
--- a/lib/iris/tests/unit/common/mixin/test_LimitedAttributeDict.py
+++ b/lib/iris/tests/unit/common/mixin/test_LimitedAttributeDict.py
@@ -42,6 +42,9 @@ class Test(tests.IrisTest):
         right = LimitedAttributeDict(**values)
         self.assertEqual(left, right)
         self.assertEqual(left, values)
+
+        # ChrisB: I am not sure the follow test is really valid. It is comparing
+        # a one-element numpy array with a scalar. Is that what we want?
         values = dict(one=np.arange(1), two=np.arange(1), three=np.arange(1))
         left = LimitedAttributeDict(dict(one=0, two=0, three=0))
         right = LimitedAttributeDict(**values)
--- a/lib/iris/_merge.py
+++ b/lib/iris/_merge.py
@@ -388,7 +388,16 @@ class _CubeSignature(
                 diff_attrs = [
                     repr(key[1])
                     for key in attrs_1
-                    if np.all(attrs_1[key] != attrs_2[key])
+                    if not np.array_equal(attrs_1[key], attrs_2[key])
+                    ## CB: Is the original line below really what was intended?
+                    ##     It only evaluates to true if ALL the values are different
+                    ##     in an array.
+                    ##     Surely it should be either:
+                    ##       np.any( attrs_1[key] != attrs_2[key] )
+                    ##     or
+                    ##       not np.all( attrs_1[key] == attrs_2[key] )
+                    ##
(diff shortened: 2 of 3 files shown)


ukmo-ccbunney:
Note, this change fixes what I believe is a bug here:
https://github.com/SciTools/iris/blob/d3071ff90f9e347ad4f1e1b9b097a74f28beb21d/lib/iris/_merge.py#L388-L391
Line 391 should be checking for **any** differences in the array elements, not **all**.

It also cleans up a now unnecessary `try...except` block that I believe was only needed for older versions of numpy. 
https://github.com/SciTools/iris/blob/d3071ff90f9e347ad4f1e1b9b097a74f28beb21d/lib/iris/common/mixin.py#L108-L111

I have extended the existing unit test to check for correct handling inequality.
However, I question the validity of this test
https://github.com/SciTools/iris/blob/d3071ff90f9e347ad4f1e1b9b097a74f28beb21d/lib/iris/tests/unit/common/mixin/test_LimitedAttributeDict.py#L45-L49
which now fails as it is testing a scalar value against a single element array. With the old logic, this would work, but it is flawed as `np.all( np.array([0]) == 0 )` and `np.all( np.zeros(100) == 0 )` would both return `True`, even though the latter is not what is intended.

I have commented out this test at the moment and would appreciate the opinion @pp-mo.

lib/iris/tests/unit/common/mixin/test_LimitedAttributeDict.py
```diff
@@ -42,11 +42,24 @@ def test___eq___numpy(self):
...
+        self.assertNotEqual(values, right)
+
+        # ChrisB: I am not sure the follow test is really valid. It is comparing
+        # a one-element numpy array with a scalar. Is that what we want?
+
+    #    values = dict(one=np.arange(1), two=np.arange(1), three=np.arange(1))
+    #    left = LimitedAttributeDict(dict(one=0, two=0, three=0))
+    #    right = LimitedAttributeDict(**values)
+    #    self.assertEqual(left, right)
+    #    self.assertEqual(left, values)
```
pp-mo, line 62: I think this usage **is** intentional.
We do want x to be the same as [x], because that's how netcdf attributes work.
So 1 = [1], but (obviously) 1 != [1, 0].

I think at least one way to handle this would be to wrap both sides with np.array(x, ndmin=1).
Interestingly, that leaves open the question of whether attributes can be multidimensional, or indeed non-array objects.  Iris does allow that, but they can't exist in files:   We use this. for example, in allowing attributes to be [STASH objects](https://github.com/SciTools/iris/blob/cb475400c471f79fa129188fae9fd97140156ea2/lib/iris/fileformats/pp.py#L246) -- but these are specifically intercepted + converted (both ways) on netcdf read and write.

pp-mo (changes requested):

pp-mo (changes requested):
See comment where I think this changes too much from as-is behaviour.
Otherwise all good , I think

ESadek-MO (approved):
Happy to build on top of @pp-mo 's review and consider this good to go!
Thanks @ukmo-ccbunney !

Issue #3678, opened 2020-03-05
Bounds of derived variable are not read correctly


schlunma:
Hi,

in ESMValTool we found the following issue concerning files with derived variables (in particular ` atmosphere_hybrid_sigma_pressure_coordinate`, see ESMValGroup/ESMValCore#543):

The bounds of the derived variable of model output that is consistent with the [CF conventions](http://cfconventions.org/Data/cf-conventions/cf-conventions-1.7/cf-conventions.html#cell-boundaries) cannot be read correctly with `iris`. Here is a minimal example using the [nc file](https://github.com/SciTools/iris/files/4292678/a_new_file.zip) that is used on the CF convention webpage (without the optional attributes and the units of `A` replaced by `1`):

```
netcdf a_new_file {
dimensions:
        eta = 1 ;
        lat = 1 ;
        lon = 1 ;
        bnds = 2 ;
variables:
        double eta(eta) ;
                eta:long_name = "eta at full levels" ;
                eta:positive = "down" ;
                eta:standard_name = "atmosphere_hybrid_sigma_pressure_coordinate" ;
                eta:formula_terms = "a: A b: B ps: PS p0: P0" ;
                eta:bounds = "eta_bnds" ;
        double eta_bnds(eta, bnds) ;
                eta_bnds:formula_terms = "a: A_bnds b: B_bnds ps: PS p0: P0" ;
        double A(eta) ;
                A:long_name = "a coefficient for vertical coordinate at full levels" ;
                A:units = "1" ;
        double A_bnds(eta, bnds) ;
        double B(eta) ;
                B:long_name = "b coefficient for vertical coordinate at full levels" ;
                B:units = "1" ;
        double B_bnds(eta, bnds) ;
        double PS(lat, lon) ;
                PS:units = "Pa" ;
        double P0 ;
                P0:units = "Pa" ;
        float temp(eta, lat, lon) ;
                temp:standard_name = "air_temperature" ;
                temp:units = "K" ;
```

 Reading this file with `iris`

```python
import iris
print(iris.__version__)
print("")

path = os.path.expanduser('~/a_new_file.nc')
cubes = iris.load(path)
print(cubes)
print("")

cube = cubes.extract_strict(iris.Constraint('air_temperature'))
air_pressure_coord = cube.coord('air_pressure')
print(air_pressure_coord)
```

gives

```
2.4.0

[...]

PR #6190, opened 2024-10-24
Bounds on derived coords

HGWright:
### Description

This PR does the loading agreed in #3678 ([from this comment](https://github.com/SciTools/iris/issues/3678#issuecomment-1490268672))

---

---


lib/iris/fileformats/cf.py
```diff
@@ -1425,7 +1425,41 @@ def _translate(self):
...
+                                        ):
+                                            if cf_terms in form_terms:
+                                                to_attach_to = example_dict[cf_terms]
+                                                attach_from = cf_vari.cf_name
+                                                if (
+                                                    to_attach_to.lower()
+                                                    != attach_from.lower()
+                                                ):
+                                                    cf_var.bounds = cf_vari
+                                                    print(cf_vari.bounds)
```
trexfeathers, line 1462:
```suggestion
                bounds_name = None
                cf_root_coord = self.cf_group.coordinates.get(cf_root, None)
                with contextlib.suppress(AttributeError):
                    # Copes with cf_root_coord not existing, OR not having
                    #  `bounds` attribute.
                    bounds_name = _getncattr(cf_root_coord, "bounds")
                if bounds_name is not None:
                    # This will error if more or less than 1 variable is found.
                    #  TODO: does this ever need to be more permissive?
                    (bounds_var,) = [
                        f for f in formula_terms.values()
                        if f.cf_terms_by_root.get(bounds_name) == cf_term
                    ]
                    self.cf_group[cf_var.cf_name].bounds = bounds_var.cf_name
                    # TODO: get bounds_var into the appropriate 'cf_group' for
                    #  discovery during helpers.get_cf_bounds_var().
                    self.cf_group[bounds_var.cf_name] = CFBoundaryVariable(
                        bounds_var.cf_name, bounds_var.cf_data
                    )
```

trexfeathers: This might also work:

```diff
-cf_root_coord = self.cf_group.coordinates.get(cf_root, None)
+cf_root_coord = self.cf_group.coordinates.get(cf_root)
```

trexfeathers: Note that I did not exclude variables that are duplicated between the original formula terms and the bounds formula terms. E.g. `ps` and `p0` in #3678.

This needs doing.

trexfeathers:
> After a long and productive discussion with @trexfeathers and @pp-mo (thanks guys). We have decided that for the upcoming 3.12 release we will put this behind a `Future:` flag as a different method of loading, this means that we can keep the existing Iris behaviour. We can also avoid making this save to the best version of CF until a later date to avoid this code going stale. (This will be written up and spec'd out in an issue.)

Something else from that discussion, which affects the work in this PR: we need to think about how to handle a case where both approaches to recording bounds have been used, but they produce conflicting information.

trexfeathers:
### My understanding of our goal

... after my most recent conversation with @HGWright

Ref: [CF Conventions 7.1.4. Boundaries and Formula Terms](https://cfconventions.org/Data/cf-conventions/cf-conventions-1.12/cf-conventions.html#boundaries-and-formula-terms)

|   | `FUTURE` OFF [^1]  | `FUTURE` ON |
| - | - | - |
| Primary method [^2] | ❌ | ✔ |
| Secondary method [^3] | ✔ | ✔ |
| `hybrid_height.nc` [^4] | ✔ | ❌ |

[^1]: the current situation

[^2]: "If a parametric coordinate variable with a formula_terms attribute (section 4.3.2) also has a bounds attribute, its boundary variable must have a formula_terms attribute too. In this case the same terms would appear in both (as specified in Appendix D), since the transformation from the parametric coordinate values to physical space is realized through the same formula. For any term that depends on the vertical dimension, however, the variable names appearing in the formula terms would differ from those found in the formula_terms attribute of the coordinate variable itself because the boundary variables for formula terms are two-dimensional while the formula terms themselves are one-dimensional."

[^3]: "Whenever a formula_terms attribute is attached to a boundary variable, the formula terms may additionally be identified using a second method: variables appearing in the vertical coordinates' formula_terms may be declared to be coordinate, scalar coordinate or auxiliary coordinate variables, and those coordinates may have bounds attributes that identify their boundary variables. In that case, the bounds attribute of a formula terms variable must be consistent with the formula_terms attribute of the boundary variable. Software digesting legacy datasets (constructed prior to version 1.7 of this standard) may have to rely in some cases on the first method of identifying the formula term variables and in other cases, on the second. Starting from version 1.7, however, the first method will be sufficient."

[^4]: [`hybrid_height.nc`](https://github.com/SciTools/iris-sample-data/blob/main/iris_sample_data/sample_data/hybrid_height.nc) in iris-sample-data is [acknowledged to be incorrect CF](https://github.com/SciTools/iris/issues/3678#issuecomment-2441373336), but has been the Iris status quo for years. Iris internals/tests expect this format, and we must assume some users might also depend on this, hence why dropping support for this invalid format must be protected behind a `FUTURE` flag.

HGWright:
I have messed up the PR trying to revive this. Closing and moving changes to a new PR.

PR #6194, opened 2024-10-25
Enable factory references to create new dimensions on load. (#6168)

stephenworsley:
Merge feature branch with changes made by #6168.


ESadek-MO (approved):
Thanks @stephenworsley. I'm content this is essentially #6168, which has already been reviewed and merged, so happy to approve this too.

PR #6195, opened 2024-10-25
SPEC 0: drop py310 and support py313

### LINKED ISSUE #6177: Support Python 3.13
3.13 was [released 2024-10-07](https://peps.python.org/pep-0719/#lifespan)
Aligning with [SPEC0](https://scientific-python.org/specs/spec-0000/), this would mean we support 3.11, 3.12, 3.13 + drop 3.10.
Presumably, that will be ok.

stephenworsley:
### Description
closes #6177


benchmarks/asv.conf.json
```diff
@@ -37,16 +37,16 @@
...
             "PY_VER=3.12 nox --envdir={env_parent} --session=tests --install-only --no-error-on-external-run --verbose"
         ],
-        "d58fca7e": [
+        "44fae030": [
             "ENV_PARENT={conf_dir}/.asv/env/nox311",
             "PY_VER=3.11 nox --envdir={env_parent} --session=tests --install-only --no-error-on-external-run --verbose"
-        ],
-        "44fae030": [
-            "ENV_PARENT={conf_dir}/.asv/env/nox310",
-            "PY_VER=3.10 nox --envdir={env_parent} --session=tests --install-only --no-error-on-external-run --verbose"
```
trexfeathers, line 49: In this case I would prefer that we keep the old one, as we occasionally have need to benchmark further back in history.

benchmarks/asv.conf.json
```diff
@@ -37,16 +37,16 @@
     ],
     "delegated_env_commands": {
         "c8a663a0": [
+            "ENV_PARENT={conf_dir}/.asv/env/nox313",
```
trexfeathers, line 40: This is awkward - we don't know what the new commit hash will be once this is merged. I didn't think of that 🙄

trexfeathers: Discussed with @ESadek-MO: if you just provide one of the commit hashes that is on your branch, and then we'll merge this one with a merge-commit (not a squash).

stephenworsley: Like this https://github.com/SciTools/iris/pull/6195/commits/bd6495c401f81e6a46a939ab1f3e0ef7c752df0b ?

stephenworsley:
Looks like python 3.13 is causing a lot of test failures. Most of these seem to be in cml files where I'm seeing `&#10` show up a lot. I think this may be due to how python is rendering newlines. Seems like this can't be fixed by rerendering all the cml files since that would break the tests for the older versions so this seems like a bit of a blocker for the moment.

benchmarks/asv.conf.json
trexfeathers: Almost, however there are now duplicate `3.13`, which is what is causing the benchmark failure.

trexfeathers:
> hi folks, we are ready to roll out support for ESMValCore and Python 3.13 in [ESMValGroup/ESMValCore#2566](https://github.com/ESMValGroup/ESMValCore/pull/2566) - @bouweandela brought this issue to my attention, do we understand that `iris==3.11` doesn't really-really support Python 3.13 and we have to wait for `iris==3.12` or an eventual bugfix `iris==3.11.1`? Cheers muchly! 🍺

The test failures are all caused by a change in XML character representation, which stops us comparing against known-good-outputs. This has made 5% of our tests 'blind' until we can work on a solution. But we have NO evidence of any Iris functionality breaking, and 95% of our tests still work and they show Iris functioning fine. Your own passing tests (https://github.com/ESMValGroup/ESMValCore/pull/2566#issuecomment-2470346860) back this up even more.

So the Iris 'project' cannot support Python 3.13 until we can get our CI working with 3.13; we have no immediate plans for finishing this work as we need space to finish some other things. But it looks like the Iris **software** supports 3.13 just fine already 👍. This happens with all new Python versions - our users have success with it before we even get a chance to test against it.

valeriupredoi:
> > hi folks, we are ready to roll out support for ESMValCore and Python 3.13 in [ESMValGroup/ESMValCore#2566](https://github.com/ESMValGroup/ESMValCore/pull/2566) - @bouweandela brought this issue to my attention, do we understand that `iris==3.11` doesn't really-really support Python 3.13 and we have to wait for `iris==3.12` or an eventual bugfix `iris==3.11.1`? Cheers muchly! 🍺
> 
> The test failures are all caused by a change in XML character representation, which stops us comparing against known-good-outputs. This has made 5% of our tests 'blind' until we can work on a solution. But we have NO evidence of any Iris functionality breaking, and 95% of our tests still work and they show Iris functioning fine. Your own passing tests ([ESMValGroup/ESMValCore#2566 (comment)](https://github.com/ESMValGroup/ESMValCore/pull/2566#issuecomment-2470346860)) back this up even more.
> 
> So the Iris 'project' cannot support Python 3.13 until we can get our CI working with 3.13; we have no immediate plans for finishing this work as we need space to finish some other things. But it looks like the Iris **software** supports 3.13 just fine already 👍. This happens with all new Python versions - our users have success with it before we even get a chance to test against it.

wonderful, exactly the answer I was looking for, both wrt clarity and content, cheers, Martin! :beer: 
@bouweandela we're good to go with  https://github.com/ESMValGroup/ESMValCore/pull/2566 then :firecracker:

trexfeathers:
We have a remaining problem with how our expected checks are set up - they are specific to Python versions:

image

- I can switch these over to expect the new versions, but that will block ALL other PRs
- I can turn off the expected checks, although that would defeat the object of having them (to allow auto-merge). Maybe could do that temporarily then quickly turn back on?

This has previously affected Stratify, but considering the lower traffic that wasn't such a problem there. @HGWright can you offer any extra insight?

trexfeathers:
>We have a remaining problem with how our [expected checks](https://docs.github.com/en/repositories/configuring-branches-and-merges-in-your-repository/managing-protected-branches/managing-a-branch-protection-rule) are set up - they are specific to Python versions:

Decided as a team to allow repository admins to override the expected checks.

PR #6200, opened 2024-10-28
Update CF standard names table.

ESadek-MO:


HGWright (changes requested):
@ESadek-MO Just missing a `whatsnew` entry and this will be complete

HGWright (approved):

Issue #5690, opened 2024-01-11
Pytest adoption tracker


trexfeathers:
# Iris approximate pytest adoption

## Generated using this script

<details>
<summary>Click to expand <b>this section...</b></summary>

```python
import ast
from pathlib import Path
from sys import argv

# Also requires the `tabulate` package.
import pandas as pd


if not (len(argv) == 2 and Path(argv[1]).exists()):
    message = "Correct script usage = `pytest_adoption.py <iris_root_dir>`"
    raise ValueError(message)

IRIS_ROOT = Path(argv[1])
IRIS_TREE_URL = "https://github.com/SciTools/iris/tree/main/"
markdown_path = Path(__file__).with_suffix(".md")

tests_dir = IRIS_ROOT / "lib" / "iris" / "tests"
dataframe_data = []
total_asserts = 0
total_asserts_pytest = 0

for file_path in tests_dir.rglob("*.py"):
    file_text = file_path.read_text()
    parsed = ast.parse(source=file_text)
    calls = filter(lambda node: hasattr(node, "func"), ast.walk(parsed))
    assert_calls = filter(
        lambda c: getattr(c.func, "attr", "")[:6] == "assert", calls
    )
    assert_calls_methods = filter(lambda c: hasattr(c.func, "value"), calls)
    assert_calls_methods_self = filter(
        lambda m: getattr(m.func.value, "id", "") == "self", assert_calls_methods
    )
    num_asserts_unittest = len(list(assert_calls_methods_self))

    assert_keywords = filter(
        lambda node: isinstance(node, ast.Assert), ast.walk(parsed)
    )
    num_asserts_pytest = len(list(assert_keywords))

    both_asserts = num_asserts_unittest + num_asserts_pytest
    if both_asserts == 0:
        percent_adoption = "N/A"
    else:
        percent_adoption = f"{num_asserts_pytest / both_asserts * 100:.0f}"

    total_asserts += both_asserts
    total_asserts_pytest += num_asserts_pytest

    file_url = IRIS_TREE_URL + str(file_path.relative_to(IRIS_ROOT))
    file_link_text = f"{file_path.relative_to(tests_dir).stem}"

    dataframe_data.append(
        {
            "% adoption": percent_adoption,
            # "File Path": f"[{file_link_text}]({file_url})",
            "File Path": f"{file_link_text}",
            "Number of `self.assertSomething()`": num_asserts_unittest,
            "Number of `assert something`": num_asserts_pytest,
        }
    )

[...]

PR #6213, opened 2024-11-01
Add a pytest-unittest translator to the conversion guide  

ESadek-MO:
Added a translation guide, as suggested in https://github.com/SciTools/iris/issues/5690#issuecomment-1988712822.

Additionally, corrected myself; ``mocker`` has the same api as ``mock``, and can be used as a context manager. In the example I based that judgement off, it was just decided it wasn't needed. 

See https://pytest-mock.readthedocs.io/en/latest/usage.html

### DIFF
--- a/docs/src/developers_guide/contributing_pytest_conversions.rst
+++ b/docs/src/developers_guide/contributing_pytest_conversions.rst
@@ -41,7 +41,7 @@ Conversion Checklist
 
 #. Check for references to ``@tests``. These should be changed to ``@_shared_utils``.
 #. Check for references to ``with mock.patch("...")``. These should be replaced with
-   ``mocker.patch("...")``, which can be passed into functions as a fixture.
+   ``mocker.patch("...")``. ``mocker`` is a fixture, and can be passed into functions.
 #. Check for ``np.testing.assert...``. This can usually be swapped for
    ``_shared_utils.assert...``.
 #. Check for references to ``super()``. Most test classes used to inherit from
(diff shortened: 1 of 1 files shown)


docs/src/developers_guide/contributing_pytest_conversions.rst
```diff
@@ -41,7 +41,7 @@ Conversion Checklist
 
 #. Check for references to ``@tests``. These should be changed to ``@_shared_utils``.
 #. Check for references to ``with mock.patch("...")``. These should be replaced with
-   ``mocker.patch("...")``. Note, ``mocker.patch("...")`` is NOT a context manager.
+   ``mocker.patch("...")``, which and be passed into functions as a fixture.
```
ESadek-MO, line 44:
```suggestion
   ``mocker.patch("...")``, which can be passed into functions as a fixture.
```

docs/src/developers_guide/contributing_pytest_conversions.rst
```diff
@@ -41,7 +41,7 @@ Conversion Checklist
 
 #. Check for references to ``@tests``. These should be changed to ``@_shared_utils``.
 #. Check for references to ``with mock.patch("...")``. These should be replaced with
-   ``mocker.patch("...")``. Note, ``mocker.patch("...")`` is NOT a context manager.
+   ``mocker.patch("...")``, which can be passed into functions as a fixture.
```
trexfeathers, line 44: I don't understand this phrasing. Are you saying tests should look like this:

```python
def test_something(mocker.patch):
    mocker.patch(...)
```

Or like this:

```python
def test_something(mocker):
    mocker.patch(...)
```

ESadek-MO: mocker is the fixture here, so the latter. 

```suggestion
   ``mocker.patch("...")``. ``mocker`` is a fixture, and be passed into functions.
```

trexfeathers (approved):

PR #6219, opened 2024-11-06
removed double reference of typehints and enhancements

ESadek-MO:
The 3.11 whatsnew currently double mentions type hints and enhancements. I've removed the offending line. 

This should go into the 3.11 branch, and then a mergeback.


HGWright (approved):
LGTM

PR #6231, opened 2024-11-15
Use custom tokenization function for NetCDFDataProxy objects

bouweandela:
### Description

Use custom tokenization function for NetCDFDataProxy objects. This speeds up loading small NetCDF files. With the multiple_cubes.nc file from https://github.com/SciTools/iris/issues/6223, the speedup is 10 to 20 %.

---

---


pp-mo:
> the speedup is 10 to 20 %, though in the benchmarks below it is not really significant.

Looks really neat + I don't doubt you're right this is worth it for some cases.
I just wonder though, is it worth _adding_ a benchmark which demonstrates the advantage ?

bouweandela:
The benchmark I added just shows:
```
| Change   | Before [0fdedb46]    | After [23e9462b]    |   Ratio | Benchmark (Parameter)                                                                       |
|----------|----------------------|---------------------|---------|---------------------------------------------------------------------------------------------|
|          | 1.10±0.01s           | 1.06±0.01s          |    0.96 | load.ManyCubes.time_many_cube_load                                                          |
```

On my laptop the timings are a bit more positive, but maybe this isn't really worth the effort. Shall I just close this?

bouweandela:
As discussed at the peloton meeting, this is probably useful after all. The benchmark is likely slower in CI than on my laptop because it has different hardware and therefore spends more time on I/O, reducing the relative speedup of this change on that particular machine.

bouweandela:
The improvement is still ~50 ms, but since the overall time is now shorter the benefit becomes relatively larger. CI results:
```
| Change   | Before [ba9a6e2d]    | After [a09fe715]    |   Ratio | Benchmark (Parameter)                                                                       |
|----------|----------------------|---------------------|---------|---------------------------------------------------------------------------------------------|
|          | 469±10ms             | 417±4ms             |    0.89 | load.ManyCubes.time_many_cube_load                                                          |
```
and on my laptop
```

| Change   | Before [ba9a6e2d] <main>   | After [9a1a14df] <netcdfdataproxy-dask-tokenize>   |   Ratio | Benchmark (Parameter)              |
|----------|----------------------------|----------------------------------------------------|---------|------------------------------------|
|          | 177±0.9ms                  | 154±4ms                                            |    0.87 | load.ManyCubes.time_many_cube_load |
```

bouweandela:
Benchmark result from my laptop with the most recent commit (`python bm_runner.py branch upstream/main --bench=load.ManyCubes.time_many_cube_load`):

```
| Change   | Before [ba9a6e2d] <main>   | After [31abeaf7] <netcdfdataproxy-dask-tokenize>   |   Ratio | Benchmark (Parameter)              |
|----------|----------------------------|----------------------------------------------------|---------|------------------------------------|
| -        | 168±7ms                    | 102±1ms                                            |    0.61 | load.ManyCubes.time_many_cube_load |
````

bouweandela:
It looks like the advantage is rather minimal for the benchmarked case now that we are creating fewer arrays since #6252. Let's just close this.

ukmo-ccbunney:
> It looks like the advantage is rather minimal for the benchmarked case now that we are creating fewer arrays since #6252. Let's just close this.

Sincere apologies for not getting round to looking at this in good time @bouweandela.

PR #6253, opened 2024-12-11
Implement dataless cubes

### LINKED ISSUE #4447: A Dataless Cube
## ✨ Feature Request
I think it's healthy to challenge established norms...

I want the ability to create a dataless cube. By this I mean the ability to create a hyper-space defined only by metadata i.e., no data payload

Once data is added to the cube, then the dimensionality is established and locked down, as we traditionally know and accept.

## Motivation
Such hyper-spaces could be used in various ways e.g.,
- as a factory to manufacture fully formed cubes for test data
- as the target hyper-space in a regridding or interpolation transformation

I'm sure there are more concrete use cases... Please do share them on this issue 🙏

Traditionally, there are many situations where a cube **enforcing** that it has data is simply an inconvenience. Given the natural progression of model resolutions it seems "just wrong" to abuse dask to create lazy data that will never be used. It reeks of something not being quite right to me.

Let's do something about that 😉

Please up vote this issue if you'd like to see this happen 👍

ESadek-MO:
Closes #4447.

I plan to do this in four stages:

- [x] Ensure that cubes can handle `data is None` when handed a shape value, and can essentially round-trip removing and adding data. This shouldn't break existing tests. [71c7ae8](https://github.com/SciTools/iris/pull/6253/commits/71c7ae8eb6c5af40350a673c4f00e9f7cc8960c8)
- [x] Create new tests for the above. [b2fb2f8](https://github.com/SciTools/iris/pull/6253/commits/b2fb2f8de32d48665cfbc4fea45909994293f91f)
- [x] Ensure that all `DataManager` methods all make sense and work with `None` data. `DataManager.copy()` is an example of a method that won't make sense with `None` data.
- [x] Create new tests for the above.


ESadek-MO:
The FUTURE flag  causes an issue in a lot of places, because copying with `None` explicitly is used often within our code. I can either work to change our code, with a lot of feature flags, or just make the behaviour opt-in rather than opt-out. This would perhaps be more confusing to newcomers, but less confusing to veterans, and would require less refactoring.

lib/iris/_data_manager.py
```diff
@@ -219,15 +239,20 @@ def data(self, data):
             managed.
 
         """
+        # If data is None, ensure previous shape is maintained, and that it is
+        # not wrapped in an np.array
+        dataless = data is None
+        if dataless:
```
bjlittle, line 245:
```suggestion
        if (dataless := data is None):
```

lib/iris/_data_manager.py
```diff
@@ -219,15 +239,20 @@ def data(self, data):
...
+            self._shape = self.shape
```
bjlittle, line 246: @ESadek-MO Don't we require to set `self._real_array = None` and `self._lazy_array = None` then check the axiom **only** in the `dataless` case?

i.e., do that and nothing further.

lib/iris/_data_manager.py
```diff
@@ -219,15 +239,20 @@ def data(self, data):
...
 
-        # Determine whether the class instance has been created,
-        # as this method is called from within the __init__.
-        init_done = self._lazy_array is not None or self._real_array is not None
-
-        if init_done and self.shape != data.shape:
+        # Determine whether the class already has a defined shape,
+        # as this method is called from __init__.
+        has_shape = self.shape is not None
+        if has_shape and not dataless and self.shape != data.shape:
```
bjlittle, line 255: @ESadek-MO In the dataless case, we shouldn't get here.

lib/iris/_data_manager.py
```diff
@@ -242,7 +267,8 @@ def data(self, data):
         else:
             if not ma.isMaskedArray(data):
                 # Coerce input data to ndarray (including ndarray subclasses).
-                data = np.asarray(data)
+                if not dataless:
+                    data = np.asarray(data)
```
bjlittle, line 271: @ESadek-MO We shouldn't need this defensive code for the dataless case i.e., we shouldn't get here.

lib/iris/_data_manager.py
```diff
@@ -261,12 +287,26 @@ def dtype(self):
...
     def shape(self):
         """The shape of the data being managed."""
-        return self.core_data().shape
+        if self.core_data() is None:
+            result = self._shape
+        else:
+            result = self.core_data().shape
+        return result
+
+    def is_dataless(self):
```
bjlittle, line 301:
```suggestion
    def is_dataless(self) -> bool:
```

lib/iris/_data_manager.py
```diff
@@ -261,12 +287,26 @@ def dtype(self):
...
+        """Determine whether the cube is dataless.
```
bjlittle, line 302: @ESadek-MO Perhaps it's best to not use `dataless` to describle `is_dataless` e.g., maybe something like `Determine whether the cube has no data.` instead?

lib/iris/_data_manager.py
```diff
@@ -261,12 +287,26 @@ def dtype(self):
...
 
     @property
     def shape(self):
         """The shape of the data being managed."""
-        return self.core_data().shape
+        if self.core_data() is None:
+            result = self._shape
+        else:
+            result = self.core_data().shape
+        return result
```
bjlittle, line 299:
```suggestion
        return self._shape if self._shape else self.core_data().shape
```

lib/iris/_data_manager.py
```diff
@@ -261,12 +287,26 @@ def dtype(self):
...
+
+    def is_dataless(self):
+        """Determine whether the cube is dataless.
+
+        Returns
+        -------
+        bool
+
+        """
+        return (self.core_data() is None) and (self.shape is not None)
```
bjlittle, line 309: @ESadek-MO Given our axiom (and if this isn't true then something is wrong) it must always be the case that:

- `self._shape = None` and (`self._lazy_array is not None` or `self._real_array is not None`)
- `self._shape is not None` and (`self._lazy_array is None` and `self._real_array is None`)

Therefore, `is_dataless` should be defined simply as `return self._shape is not None`, right?

bjlittle (changes requested):
@ESadek-MO I'm only partially through the review, but I thought I'd push these review comments early so that you can see them and address them ASAP.

Note that you also need to refactor the following `DataManager` methods:

- `__equal__` to account for dataless cubes, particularly when dataless cubes are involved in the operation i.e., for the new use cases between data and dataless, and  dataless and dataless
- `lazy_data` to deal with the dataless case
- akin to the `lazy_data` method, you also have to deal with `core_data` for the dataless case
-`  __repr__` requires to cope with the data less case i.e., provide the shape

lib/iris/_data_manager.py
```diff
@@ -10,32 +10,42 @@
...
         # Initialise the instance.
         self._lazy_array = None
         self._real_array = None
 
         # Assign the data payload to be managed.
+        self._shape = shape
         self.data = data
 
-        # Enforce the manager contract.
-        self._assert_axioms()
```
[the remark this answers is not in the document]
ESadek-MO: It didn't seem necessary, since `self.data = data` already has assert_axioms within it

lib/iris/_data_manager.py
```diff
@@ -126,11 +136,16 @@ def __repr__(self):
...
+        empty = self._lazy_array is None and self._real_array is None
+        overfilled = self._lazy_array is not None and self._real_array is not None
+        if overfilled:
+            msg = "Unexpected data state, got both lazy and real data."
+            raise ValueError(msg)
+        elif (
+            empty and self._shape is None
+        ):  # if I remove the second check, allows empty arrays, like old behaviour
+            msg = "Unexpected data state, got no lazy or real data, and no shape."
+            raise ValueError(msg)
```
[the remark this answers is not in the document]
ESadek-MO: I'm so glad of this correction, I couldn't get my head around the correct way to do this and was pulling out what precious hair I have left.

lib/iris/_data_manager.py
```diff
@@ -148,25 +163,30 @@ def _deepcopy(self, memo, data=None):
...
         try:
             if data is None:
                 # Copy the managed data.
                 if self.has_lazy_data():
                     data = copy.deepcopy(self._lazy_array, memo)
-                else:
+                elif self._real_array is not None:
                     data = self._real_array.copy()
+                else:
+                    shape = self.shape
```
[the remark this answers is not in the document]
ESadek-MO: I'll come back to this one

ESadek-MO:
Something is wrong with the _assert_axiom code now. I need to catch a bus, so will have to figure out why tomorrow.

ESadek-MO:
Although the tests are passing, next time I look at this, I need to investigate the data setter. Currently, line 256 causes issues, as self._shape should be changable if self._shape = (), and it won't be.

lib/iris/_data_manager.py
```diff
@@ -216,57 +241,76 @@ def data(self, data):
...
     def ndim(self):
         """The number of dimensions covered by the data being managed."""
-        return self.core_data().ndim
+        return len(self.shape)
 
     @property
     def shape(self):
         """The shape of the data being managed."""
-        return self.core_data().shape
+        return self._shape if self._shape else self.core_data().shape
```
ESadek-MO, line 303: This is defensive; it wouldn't cause any issues **that I can think of** to just have `return self._shape`, but this should protect us in case I've missed something.

ESadek-MO: I've actually removed this now. 

() is falsy. I could have easily done `...if self._shape is not None`, but I decided there's no point in adding redundant code. Easy undo if my reviewer disagrees!

lib/iris/_data_manager.py
```diff
@@ -128,9 +143,14 @@ def _assert_axioms(self):
         # Ensure there is a valid data state.
         is_lazy = self._lazy_array is not None
         is_real = self._real_array is not None
-        emsg = "Unexpected data state, got {}lazy and {}real data."
-        state = is_lazy ^ is_real
-        assert state, emsg.format("" if is_lazy else "no ", "" if is_real else "no ")
+
+        if not (is_lazy ^ is_real):
```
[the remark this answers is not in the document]
ESadek-MO: I did see this, but forgot to comment. 
*Personally*, I think I prefer it as it is, but if there's significant desire for your suggestion, I don't mind.

lib/iris/_data_manager.py
```diff
@@ -10,32 +10,43 @@
...
         """
+        if (shape is not None) and (data is not None):
+            msg = '"shape" should only be provided if "data" is None'
+            raise ValueError(msg)
+
+        self._shape = shape
+
         # Initialise the instance.
         self._lazy_array = None
         self._real_array = None
```
ESadek-MO, line 45:
```suggestion
        # Initialise the instance.
        self._shape = shape
        self._lazy_array = None
        self._real_array = None
```

lib/iris/_data_manager.py
```diff
@@ -10,32 +10,42 @@
...
+        shape : tuple, optional
+            A tuple, representing the shape of the data manager. This can only
+            be used in the case of ``data=None``, and will render the data manager
+            dataless.
 
         """
+        if (shape is not None) and (data is not None):
+            msg = '"shape" should only be provided if "data" is None'
+            raise ValueError(msg)
+
```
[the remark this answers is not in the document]
bjlittle: There is a question here that if the user provides `data` and `shape` and the `shape` actually matches the `data.shape`, then true they have overspecified things, but would it not be more forgiving to not raise an exception?

bjlittle (changes requested):
@ESadek-MO Awesome thanks :+1: 

A couple of minor comments to service.

The big gotcha is that we need to be able to handle equality that involves a dataless cube, which is broken ATM e.g.,

```python
>>> from iris.cube import Cube
>>> a = Cube(1)
>>> b = Cube(shape=(1,))
>>> a == b
  File "/home/users/bill.little/projects/iris/lib/iris/cube.py", line 4163, in __eq__
    np.allclose(
  File "/home/users/bill.little/.conda/envs/iris-dev/lib/python3.12/site-packages/numpy/_core/numeric.py", line 2329, in allclose
    res = all(isclose(a, b, rtol=rtol, atol=atol, equal_nan=equal_nan))
              ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/users/bill.little/.conda/envs/iris-dev/lib/python3.12/site-packages/numpy/_core/numeric.py", line 2447, in isclose
    result = (less_equal(abs(x-y), atol + rtol * abs(y))
                             ~^~
TypeError: unsupported operand type(s) for -: 'int' and 'NoneType'
>>> b == b
True
>>> c = Cube(shape=(1,))
>>> b == c
  File "/home/users/bill.little/projects/iris/lib/iris/cube.py", line 4163, in __eq__
    np.allclose(
  File "/home/users/bill.little/.conda/envs/iris-dev/lib/python3.12/site-packages/numpy/_core/numeric.py", line 2329, in allclose
    res = all(isclose(a, b, rtol=rtol, atol=atol, equal_nan=equal_nan))
              ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/users/bill.little/.conda/envs/iris-dev/lib/python3.12/site-packages/numpy/_core/numeric.py", line 2447, in isclose
    result = (less_equal(abs(x-y), atol + rtol * abs(y))
                             ~^~
TypeError: unsupported operand type(s) for -: 'NoneType' and 'NoneType'
```

lib/iris/_data_manager.py
```diff
@@ -10,32 +10,42 @@
...
         """Create a data manager for the specified data.
 
         Parameters
         ----------
-        data :
+        data : np.typing.ArrayLike, optional
             The :class:`~numpy.ndarray` or :class:`~numpy.ma.core.MaskedArray`
             real data, or :class:`~dask.array.core.Array` lazy data to be
-            managed.
+            managed. If a value of None is given, the data manager will be
```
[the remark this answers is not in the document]
ESadek-MO: https://github.com/SciTools/iris/pull/6253/commits/afd63b3df94e0da3c00405f2107a89468a4f81e2

lib/iris/_data_manager.py
```diff
@@ -10,32 +10,42 @@
...
     """Provides a well defined API for management of real or lazy data."""
 
-    def __init__(self, data):
+    def __init__(self, data, shape=None):
         """Create a data manager for the specified data.
 
         Parameters
         ----------
-        data :
+        data : np.typing.ArrayLike, optional
```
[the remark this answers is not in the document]
ESadek-MO: https://github.com/SciTools/iris/pull/6253/commits/afd63b3df94e0da3c00405f2107a89468a4f81e2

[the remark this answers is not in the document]
ESadek-MO: I have a feeling this needs importing actually, I'll go back and do that in a sec.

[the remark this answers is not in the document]
bjlittle: You don't need to import it for the doc-string, but you will if you add a type annotation.

Put that at the top of the module using the following pattern:

```python
from typing import TYPE_CHECKING

if TYPE_CHECKING:
    from numpy.typing import ArrayLike
```

lib/iris/_data_manager.py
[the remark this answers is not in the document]
ESadek-MO: https://github.com/SciTools/iris/pull/6253/commits/1b4917e5728264f3a42cb85c0d0627ccea9c98af, with a minor correction for warnings usage in https://github.com/SciTools/iris/pull/6253/commits/4a9befbbc39da039759625dcc6fd9fadf9847c51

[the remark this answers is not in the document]
bjlittle: @ESadek-MO Are you adding some test coverage for this warning behaviour?

lib/iris/_data_manager.py
```diff
@@ -5,37 +5,56 @@
...
+            A tuple, representing the shape of the data manager. This can only
+            be used in the case of ``data=None``, and will render the data manager
+            dataless.
 
         """
+        if (shape is None) and (data is None):
+            msg = 'one of "shape" or "data" should be provided; both are None'
+            raise ValueError(msg)
+        elif (shape is not None) and (data is not None):
+            if shape == data.shape:
```
bjlittle, line 42: @ESadek-MO This breaks with `iris.cube.Cube(1, shape=(1,))` i.e., `data` may not be an instance with a shape method.

TBH thinking about this, doing a` data.shape` here is kinda dangerous. To avoid complications, it probably better to dial this back to ensuring that simply both `data` and `shape` are not provided.

lib/iris/tests/unit/data_manager/test_DataManager.py
```diff
@@ -18,6 +20,26 @@
 from iris._lazy_data import as_lazy_data
 
 
+class Test__init__:
+    @pytest.fixture(autouse=True)
+    def _setup(self):
+        self.data = np.array([1])
+
+    def test_data_same_shape(self):
+        with pytest.warns(match=""):
```
bjlittle, line 29: @ESadek-MO Can we match against the expected warning message.

lib/iris/tests/unit/data_manager/test_DataManager.py
```diff
@@ -158,6 +190,13 @@ def test_lazy(self):
         expected = "{}({!r})".format(self.name, self.lazy_array)
         self.assertEqual(result, expected)
 
+    def test_dataless(self):
+        print(self.real_array.shape)
```
bjlittle, line 194: @ESadek-MO Remove the `print`

lib/iris/tests/unit/data_manager/test_DataManager.py
```diff
@@ -158,6 +190,13 @@ def test_lazy(self):
         expected = "{}({!r})".format(self.name, self.lazy_array)
         self.assertEqual(result, expected)
 
+    def test_dataless(self):
```
bjlittle, line 193: @ESadek-MO Rename to describe better what it's testing ... `test_dataless_repr` ?

lib/iris/tests/unit/data_manager/test_DataManager.py
```diff
@@ -452,6 +492,46 @@ def test_nd_lazy_to_nd_lazy(self):
...
 
+    def test_nd_lazy_to_dataless(self):
+        shape = (2, 3, 4)
+        size = np.prod(shape)
+        real_array = np.arange(size).reshape(shape)
+        lazy_array = as_lazy_data(real_array)
+        dm = DataManager(lazy_array * 10)
+        self.assertTrue(dm.has_lazy_data())
+        dm.data = None
+        self.assertTrue(dm.core_data() is None)
```
bjlittle, line 503: @ESadek-MO `self.assertIsNone(dm.core_data())` instead?

Repeat this pattern through-out ?

bjlittle (changes requested):
A few more comments to service ...

We're getting there 👍

bjlittle (approved):

PR #6256, opened 2024-12-16
Adapt Iris to cf-units precision change.

trexfeathers:
### Description

See https://github.com/SciTools/iris/pull/6215#issuecomment-2473331889

Going to work on the `FUTURE` flag next

---

---


trexfeathers:
This will never pass all tests, due to the bug described in #6258.

It would be sensible to complete the code in this pull request, to prove that all other tests pass, but then raise a new PR based on `main` - with the current approved lock file - to get a passing PR that we can merge.

Issue #6249, opened 2024-12-10
list of additional "equalise" operations to assist merge/concatenate


pp-mo:
**Note:**  
the suggested operations listed in #5375 contain multiple fussy "mismatch" problems like dtype mismatches and minor-floating-point-differences
Inevitably, these probably miss quite a few possiblilties + should be generalised (e.g. across different component types)

pp-mo:
another similar idea : https://github.com/SciTools/iris/issues/5050

This _might_ point to a need to supply a generic user-defined equalisation operation ??

pp-mo:
Initial draft list, drawn (inclusively!) mostly from #5375 :
  * coords differing in dtype
  * coords with very small floating-point differences (rounding?)
  * coords (dim-coords / aux-coords) : 
    * equalise or remove where points differ
    * equalise bounds : with/without
  * equalise/remove differing cell-measures/ancils
  * equalise/remove differing derived-coords  ((**think** this is a misconception, probably always based on _actual_ coords ??)
  * equalise attributes
  * equalise units (?some cases?)
  * remove/promote selected scalar coords (new-axis, include selected components)
  * "merge" attributes (as-per #5050 : selected ones)
  * "unify units" (would also need to be selective)

pp-mo:
## Another possible required processing operation
One thing that seems probably wanted is to enable merge/concat of data with a coord (e.g. time) which in some parts is multi-valued (vector) and in other parts is single-valued (i.e. scalar).
You can imagine applying a coord-name-specific 'promote' function to this to ensure it all has a 'time' dimension.
It looks like it should then concatenate.
But having tried this, you probably will **_also_** need a processing step that transposes the data into a standard dimension order, otherwise (I think) it will tend to come out with different structures, 
e.g. `[(level:10, y:100, x:100,), (level:10, time:5, y:100, x:100), ...]`
  --> `[(time:1, level:10, y:100, x:100,), (level:10, time:5, y:100, x:100), ...]` 
and this **_won't_** concatenate, until they are all arranged (t, z, y, x) or (z, t, y, x)

pp-mo:
### See also
some ideas further elaborated under "Other possible, future options" in [this comment](https://github.com/SciTools/iris/pull/6257#issuecomment-2549841347),
but nothing actually _beyond_ the above list

pp-mo:
Follow-on updated clean list of todos for implementation:
#6383

PR #6257, opened 2024-12-16
Equalise cubes

### LINKED ISSUE #6248: inital draft "equalise_cubes" operation to assist merge/concatenate

pp-mo:
Closes #6248


pp-mo:
### Status update
I think this is now good enough to consider as it is, though I'm still anticipating other useful features could be added.
It's worth considering what's included here, and why, and also what is anticipated to be added in future

Currently included options :
  * ```unify_time_units``` because 
     * it already exists + has proved useful
  * ```equalise_attributes``` because ...
     * it already exists + has proved useful
     * it has a particular relation to cube metadata, and interaction with input 'grouping' 
       \--both to affect it + to be affected by it-- so this is an opportunity to sort out how that needs to function
  * ```unify_names``` because ...
     * this operation may be needed to enable netcdf data to be concatenated, 
       which is more likely to be wanted now we've made that [available on loading](https://scitools-iris.readthedocs.io/en/v3.11.0/generated/api/iris.html#iris.LOAD_POLICY)
     * it has a particular relation to cube metadata, and interaction with input 'grouping',  
       so it's worth resolving + documenting how that works in the initial version of the function

lib/iris/util.py
```diff
@@ -2189,3 +2190,127 @@ def mask_cube_from_shapefile(cube, shape, minimum_weight=0.0, in_place=False):
...
+        When True, apply the :func:`unify_time_units` operation to each input group.
+        Note : while this may convert units of time reference coordinates, it does
+        not affect the units of the cubes themselves.
+
+    Returns
+    -------
+    :class:`~iris.cube.CubeList`
+        A CubeList containing the original input cubes, ready for merge or concatenate
+        operations.  The cubes are possibly modified (in-place), and possibly in a
+        different order.
```
stephenworsley, line 2236: I think this can be edited a little so that it doesn't suggest that the cubes might be modified in place in a different order. Something like "..., and are possibly returned in a different order".

lib/iris/util.py
```diff
@@ -2189,3 +2190,127 @@ def mask_cube_from_shapefile(cube, shape, minimum_weight=0.0, in_place=False):
...
+
+    The input cube groups also depend on the equalisation operation(s) selected :
+    Operations which equalise a specific cube metadata element (names, units,
+    attributes or cell-methods) exclude that element from the input grouping criteria.
+
+    """
+    from iris.common.metadata import CubeMetadata
+    from iris.cube import CubeList
+
+    if unify_names or apply_all:
```
stephenworsley, line 2256: Is `unify_names` the correct name for this operation? Compared to `unify_time_units`, this operation works quite differently. Only removing information rather than editing it. `unify_names` might suggest that equivalent `standard_name`s are made the same for example. This may also cause us problems if we ever wanted to add that sort of functionality in the future. A better name might be `standardise_names` or `trim_names`.

stephenworsley (changes requested):
Looking good, just a couple things, mostly about documentation. This will also want a whatsnew describing the new function.

lib/iris/tests/unit/util/test_equalise_cubes.py
```diff
@@ -0,0 +1,266 @@
...
+    n_inputs = len(inputs)
+    # NOTE: make object array of explicit shape + fill it,
+    # since np.array(inputs) *fails* specifically with a list of metadata objects
+    inputs_array = np.empty((n_inputs,), dtype=object)
+    inputs_array[:] = inputs
+    n_inputs = inputs_array.shape[0]
+    scramble_inds = rng.permutation(n_inputs)
+    inputs_array = inputs_array[scramble_inds]
+    # Modify input list **BUT N.B. IN PLACE**
+    inputs[0:] = inputs_array
```
stephenworsley, line 30: Could `[0:]` not be replaced by `[:]`?

lib/iris/util.py
```diff
@@ -2189,3 +2190,127 @@ def mask_cube_from_shapefile(cube, shape, minimum_weight=0.0, in_place=False):
...
+                cube_group_keys.append(cube_group_key)
+                cube_group_cubes.append([cube])
+            else:
+                i_at = cube_group_keys.index(cube_group_key)
+                cube_group_cubes[i_at].append(cube)
+
+        # Apply operations to the groups : in-place modifications on the cubes
+        for group_cubes in cube_group_cubes:
+            for op in equalisation_ops:
+                op(group_cubes)
```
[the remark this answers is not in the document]
pp-mo: OK, see https://github.com/SciTools/iris/pull/6257/commits/8a0080844894775fec91de75fbc34c32ca4d99b7

lib/iris/util.py
```diff
@@ -2189,3 +2190,127 @@ def mask_cube_from_shapefile(cube, shape, minimum_weight=0.0, in_place=False):
...
+            if cube.standard_name:
+                cube.long_name = None
+                cube.var_name = None
+            elif cube.long_name:
+                cube.var_name = None
+
+    # Snapshot the cube metadata elements which we use to identify input groups
+    # TODO: we might want to sanitise practically comparable types here ?
+    #  (e.g. large object arrays ??)
+    cube_grouping_keys = [
```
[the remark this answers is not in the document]
pp-mo: I've done an extensive rename, without trying to be explicit about the exact type content of the data variables.
Please see what you think : https://github.com/SciTools/iris/pull/6257/commits/f6150deac6ab04a7c7f737dd1bfc6d4fb8c4f8f4

But don't be shy if you still don't like it !

lib/iris/util.py
```diff
@@ -2189,3 +2190,127 @@ def mask_cube_from_shapefile(cube, shape, minimum_weight=0.0, in_place=False):
...
+        # However, if 'unify-names' was done, we *already* modified cubes in-place.
+
+        # Compute the cube groups
+        # N.B. *can't* use sets, or dictionary key checking, as our 'keys' are not
+        #  always hashable -- e.g. especially, array attributes.
+        # I fear this can be inefficient (repeated array compare), but maybe unavoidable
+        # TODO: might something nasty happen here if attributes contain weird stuff ??
+        cube_group_keys = []
+        cube_group_cubes = []
+        for cube, cube_group_key in zip(cubes, cube_grouping_keys):
```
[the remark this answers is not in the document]
pp-mo: See as https://github.com/SciTools/iris/pull/6257#discussion_r1892020363

lib/iris/util.py
pp-mo: I've since changed the code, so the order is now guaranteed [**unchanged** from the input](https://github.com/SciTools/iris/blob/ae992ffd333afc9493627916ba5564d73b24d953/lib/iris/util.py#L2327).

I have found it a bit hard to believe, but I **_think_** everything we may want to do (as [listed so far](https://github.com/SciTools/iris/issues/6249), anyway) can be achieved by modifying cubes in place.
So I fixed the docstring : https://github.com/SciTools/iris/pull/6257/commits/8b56344b77090ebb31f7f9a86a536ce723bed66f

lib/iris/util.py
pp-mo: now renamed to 'normalise_names' : https://github.com/SciTools/iris/pull/6257/commits/844b18d99bda75e54031c341002c94944ba0c256

I didn't like "standardise_names" because of the crossover with "standard_name"

lib/iris/tests/unit/util/test_equalise_cubes.py
pp-mo: Yes it can (!) : https://github.com/SciTools/iris/pull/6257/commits/ae992ffd333afc9493627916ba5564d73b24d953

I had simply misunderstood how this works

pp-mo:
### Update: 
oops, forgot a whatsnew.  I'll get onto that, too.

stephenworsley (changes requested):
Looking good, I think the logic of the way you've implemented things makes sense, I think starting with a simpler implementation makes sense and this looks like it should be a pretty nice convenience in a lot of common use cases.

Just a couple more minor quibles here and I'm happy merging this.

lib/iris/tests/unit/util/test_equalise_cubes.py
```diff
@@ -0,0 +1,291 @@
...
+from numpy.random import Generator
+import pytest
+
+from iris.coords import DimCoord
+from iris.cube import Cube
+from iris.util import equalise_cubes
+from iris.warnings import IrisUserWarning
+
+
+def _scramble(inputs, rng=95297):
```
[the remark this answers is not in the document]
pp-mo: Changing a different seed should *not* invalidate the checks in fact, since mostly the only difference it makes is that where the tests construct the expected result metadata, they sometimes need to be re-sorted in the same way as the input cubes (like [this example](https://github.com/SciTools/iris/blob/62285b58e1986145215a731402eedcef90268575/lib/iris/tests/unit/util/test_equalise_cubes.py#L149-L150)).
The exception is [the unify-units test](https://github.com/SciTools/iris/blob/62285b58e1986145215a731402eedcef90268575/lib/iris/tests/unit/util/test_equalise_cubes.py#L245-L246), because it resolves to the first-encountered reference time, which obviously depends on the ordering. 

So, the "magic number" was in fact chosen so that the re-sorted order is non-trivial for all small numbers of entries (approx 2-10), in particular that neither the first or last entries should stay the same.  Though naturally that isn't terribly obvious !
I will add  a comment about it, while trying to keep it shorter than all this ...

lib/iris/tests/unit/util/test_equalise_cubes.py
```diff
@@ -0,0 +1,291 @@
...
+        if usage != "off":
+            stdnames = [sn1, None, None, None, sn2, None]
+            longnames = [None, "long2", None, None, None, None]
+            varnames = [None, None, "var3", "var4", None, None]
+            expected_metadatas = [
+                _cube(stdname=stdname, longname=longname, varname=varname).metadata
+                for stdname, longname, varname in zip(stdnames, longnames, varnames)
+            ]
+            if usage == "scrambled":
+                expected_metadatas = _scramble(expected_metadatas)
```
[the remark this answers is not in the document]
pp-mo: Oh yes, good spot !

pp-mo:
## Update
Just realized a possible problem, hence put into draft temporarily..

Somewhat as @bjlittle [suggested](https://github.com/SciTools/iris/issues/6248#issuecomment-2551059998), it might have been better to use the metadata APIs for logic.
I just now realized that I have not tested this with real metadata content. The problem with the [metadata "snapshot" dictionaries](https://github.com/SciTools/iris/blob/e841ec2ba695f7fcd345fbcfa945d5b937196b17/lib/iris/util.py#L2272) I am using is that they will not support correct equality testing if they contain array attributes. Whereas actual metadata objects would do.

My purpose was (a) to make an independent stable copy and (b) that it should be easily modifiable, which actual metadata objects aren't. But I think now I will look into doing this with actual metadata instead

pp-mo:
## Update
OK panic over, sorry about that.  Please re-consider now @stephenworsley !

I added a test for array-attribute handling, and to my surprise it all **_just worked anyway_** (!)

Debugging, I quickly realised that the reason is that in the "cube_grouping_values" dictionaries, the 'attributes' values are deep-copies of the original ```CubeAttrsDict```s.   
So those **do** of course support the required comparison operation, and all works as it should.  
As I originally said, the 'grouping' is based only on equality, it has no use for more sophisticated metadata behaviours.

stephenworsley (approved):
Looks good to me.

PR #6260, opened 2024-12-17
`date_microseconds` FUTURE flag

trexfeathers:
### Description

#6256 proves that this protects our tests, and our users, against the new precision introduced in cf-units version 3.3. Unfortunately that PR is blocked by other dependency issues that will be addressed elsewhere, so I want to introduce these code changes onto `main` independently, which I have agreed with @stephenworsley.

I will run the benchmarks to check whether my change was also responsible for the performance shifts in #6256.

---

---

### DIFF
--- a/docs/src/whatsnew/latest.rst
+++ b/docs/src/whatsnew/latest.rst
@@ -30,7 +30,12 @@ This document explains the changes made to Iris for this release
 ✨ Features
 ===========
 
-#. N/A
+#. `@trexfeathers`_ added a new :class:`~iris.Future` flag -
+   ``date_microseconds`` - which sets whether Iris should use the new
+   microsecond-precision units (see :class:`cf_units.Unit`) when the unit
+   is a time unit. The previous maximum precision was seconds. You should check
+   your code for new floating point problems if activating this (e.g. when
+   using the :class:`~iris.Constraint` API). (:pull:`6260`)
 
 
 🐛 Bugs Fixed
@@ -50,7 +55,10 @@ This document explains the changes made to Iris for this release
 🚀 Performance Enhancements
 ===========================
 
-#. N/A
+#. Note that due to the new ``date_microseconds`` :class:`~iris.Future` flag,
+   the time coordinate categorisation speedup introduced in
+   :doc:`/whatsnew/3.11` will only be available when
+   ``iris.FUTURE.date_microseconds == True``.
 
 
 🔥 Deprecations
--- a/docs/src/whatsnew/latest.rst
+++ b/docs/src/whatsnew/latest.rst
@@ -32,7 +32,8 @@ This document explains the changes made to Iris for this release
 
 #. `@trexfeathers`_ added a new :class:`~iris.Future` flag -
    ``date_microseconds`` - which sets whether Iris should use the new
-   microsecond-precision units (see :class:`cf_units.Unit`) when the unit
+   microsecond-precision units (see :class:`cf_units.Unit`, microseconds
+   introduced in version 3.3) when the unit
    is a time unit. The previous maximum precision was seconds. You should check
    your code for new floating point problems if activating this (e.g. when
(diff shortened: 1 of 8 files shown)


docs/src/whatsnew/latest.rst
```diff
@@ -50,7 +55,10 @@ This document explains the changes made to Iris for this release
 🚀 Performance Enhancements
 ===========================
 
-#. N/A
+#. Note that due to the new ``date_microseconds`` :class:`~iris.Future` flag,
+   the time coordinate categorisation speedup introduced in
+   :doc:`/whatsnew/3.11` will only be available when
+   ``iris.FUTURE.date_microseconds == True``.
```
trexfeathers, line 61: @bouweandela please take note.

I spent about an hour trying to work out a way to use cftime's vectorisation when rounding a large number of time units, without success.

So to access your performance improvements you will need to activate this future flag as described.

trexfeathers:
Linkcheck should be fixed by #6261

trexfeathers:
The lack of coverage is only because we're testing against cf-units version 3.2 in this PR.

lib/iris/common/metadata.py
```diff
@@ -54,6 +57,55 @@
...
+            )
+
+    result = _num2date_original(
+        self, time_value, only_use_cftime_datetimes, only_use_python_datetimes
+    )
+    if FUTURE.date_microseconds is False:
+        message = (
+            "You are using legacy date precision for Iris units - max "
+            "precision is seconds. In future, Iris will use microsecond "
+            "precision - from cf-units version 3.3 - which may affect core "
```
trexfeathers, line 90: I could instead say "available since cf-units version 3.3" if it were felt to be important. I'm uncertain how much it matters.

lib/iris/common/metadata.py
```diff
@@ -54,6 +57,55 @@
...
+            vfunc = np.vectorize(_round)
+            result = vfunc(result)
+        else:
+            result = _round(result)
+
+    return result
+
+
+# See the note in _num2date_to_nearest_second.
+cf_units.Unit.num2date = _num2date_to_nearest_second
```
bouweandela, line 106: This will change the function even when used outside of Iris, which may be unexpected to users. Maybe it would be cleaner to only change it in those places where it is used in iris internally.

stephenworsley: I think for this to work, units in iris would have to be a subclass of cf_units.Unit . This does seem like a cleaner way of doing things.

stephenworsley: If everything works out nicely, it might just be a case of modifying the class and instantiating it here: https://github.com/SciTools/iris/blob/01bbdf68524a9c6375b3f7b1ce52ec7385d404a3/lib/iris/common/mixin.py#L209-L210

lib/iris/common/metadata.py
stephenworsley: that does sound clearer to me

stephenworsley (changes requested):
Looks good in theory, though I think there's a way to achieve this without changing the behaviour of cf_units for code outside of iris by limiting this to a change in the `Unit` class which iris objects have as their `units` attribute.

lib/iris/common/metadata.py
trexfeathers: d673480cf8eb6da7e1c2fac983417e8658263a82

lib/iris/common/metadata.py
trexfeathers: 5c8fb3f9947a02999799efec19f287bb69c7e8d1

My tests still work, and the docstrings and hints still come from the original `Unit` class thankfully.

stephenworsley (approved):
Nice one. This looks a lot more robust now.

trexfeathers:
Thanks @stephenworsley1

PR #6261, opened 2024-12-17
Remove freepik.com link due to linkcheck breakage

trexfeathers:
### Description

---

---

### DIFF
--- a/docs/src/conf.py
+++ b/docs/src/conf.py
@@ -404,7 +404,6 @@ linkcheck_ignore = [
     "https://twitter.com/scitools_iris",
     "https://stackoverflow.com/questions/tagged/python-iris",
     "https://www.flaticon.com/",
-    "https://www.freepik.com/",
 ]
 
 # list of sources to exclude from the build.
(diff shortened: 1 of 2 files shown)


docs/src/conf.py
```diff
@@ -404,6 +404,7 @@ def _dotv(version):
     "https://twitter.com/scitools_iris",
     "https://stackoverflow.com/questions/tagged/python-iris",
     "https://www.flaticon.com/",
+    "https://www.freepik.com/",
```
trexfeathers, line 407: This is coming up as "forbidden", which suggests that linkcheck doesn't want to touch it, rather than it being broken, so that feels permanent hence me adding it to the list.

trexfeathers:
@ukmo-ccbunney thanks for volunteering your time, this should be a quick one! I decided the link didn't deliver any value so it wasn't worth pursuing and just removed it instead.

ukmo-ccbunney (approved):
Odd how the link ignore didn't work, that was going to be my suggestion until I saw you already tried it.

I agree that we can just remove it, especially as it is blocking other work.

trexfeathers:
Thanks @ukmo-ccbunney!

PR #6289, opened 2025-01-17
Add support for saving string stash codes to PP

### LINKED ISSUE #6239: String stash code silently ignored on pp-save
## 🐛 Bug Report

Before saving to pp-file, a user may have reason to modify or add a stash code.  If the code is added in the string form, it is silently ignored.

## How To Reproduce

1. Run code
```python
import iris

fname = iris.sample_data_path('air_temp.pp')
cube = iris.load_cube(fname)

cube.attributes['STASH'] = 'm01s34i001'

iris.save(cube, 'test.pp')
```

2. Inspect the file `test.pp`.

3. The stash code in the file is zero.

## Expected behaviour

Either the stash code in the file is 34001 or I get a type error because I should have used a `STASH` object instead of a string.

## Environment 
 - OS & Version: RHEL9 :tada: 
 - Iris Version: 3.11

## Additional context

<details>

```
Please add additional verbose information in this section e.g., code, output, tracebacks, screenshots etc
```
</details>

rcomer:
### Description

Closes #6239.  String stash codes now "just work".

---

---

### DIFF
--- a/docs/src/whatsnew/latest.rst
+++ b/docs/src/whatsnew/latest.rst
@@ -41,7 +41,8 @@ This document explains the changes made to Iris for this release
 🐛 Bugs Fixed
 =============
 
-#. N/A
+#. `@rcomer`_ added handling for string stash codes when saving pp files.
+   (:issue:`6239`, :pull:`6289`)
 
 
 💣 Incompatible Changes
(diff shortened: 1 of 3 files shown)


rcomer:
Ah linkcheck, I have not missed you.

pp-mo:
> Ah linkcheck, I have not missed you.

😞 I think we should remove this from PRs into a periodic check like benchmarks
https://github.com/SciTools/iris/issues/6292

pp-mo (approved):
I think this is totally good enough.

I did wonder about replacing arbitrary odd object with string reprs, but I think that is a step too far :  STASH is a special recognised attribute name, and we do expect to convert them back on loading, so we can save UM data and re-save correctly as PP fields with the original stashes.
FWIW we should probably so the same for GRIB too, I currently have an [issue open about that](https://github.com/SciTools/iris-grib/issues/596).

Issue #5909, opened 2024-04-10
iris.coord_categorisation.add_season_membership creates a boolean coordinate that cannot be saved to netcdf


leosaffin:
## 🐛 Bug Report
Trying to save a cube that has had iris.coord_categorisation.add_season_membership applied to it fails because netcdf does not allow boolean types

## How To Reproduce
Running
```python
from iris import coord_categorisation
from iris.tests import stock

cube = stock.realistic_4d()
iris.save(cube, "cube_with_season_membership.nc")
```
gives the following error from the netcdf4 library (full traceback below)
```
TypeError: Illegal primitive data type, must be one of dict_keys(['S1', 'i1', 'u1', 'i2', 'u2', 'i4', 'u4', 'i8', 'u8', 'f4', 'f8']), got bool (variable 'season_membership', group '/')
```

## Expected behaviour
It's fairly simple to work around this by converting the coordinate data to integers, but I would expect it to "just work". I can think of two solutions but both have downsides, so I wasn't sure what was best.

1. Change add_season_membership to return an integer 1s/0s array. This means the cube never becomes incompatible with netCDF, but might be an issue for people with code expecting a boolean array.
2. Change the saving behaviour in iris to convert boolean arrays to integers. This means add_season_membership can work the same but I don't know how you stop the cube becoming inconsistent between saving and reloading.

## Additional context

<details>
<summary>Click to expand <b>this section...</b></summary>

```
---------------------------------------------------------------------------
TypeError                                 Traceback (most recent call last)
Cell In[14], line 1
----> 1 iris.save(cube, "cube_with_season_membership.nc")

File ~/miniforge3/envs/core/lib/python3.11/site-packages/iris/io/__init__.py:497, in save(source, target, saver, **kwargs)
    495 # Single cube?
    496 if isinstance(source, Cube):
--> 497     result = saver(source, target, **kwargs)
    499 # CubeList or sequence of cubes?
    500 elif isinstance(source, CubeList) or (
    501     isinstance(source, (list, tuple))
    502     and all([isinstance(i, Cube) for i in source])
    503 ):
    504     # Only allow cubelist saving for those fileformats that are capable.

[...]

Issue #3680, opened 2020-03-06
Allow saving boolean data to NetCDF


PAGWatson:
Hello - I wanted to save a data mask from one dataset, so that I could load it and apply it to other datasets I wanted to compare with, without necessarily needing to load the first dataset in its entirety. I made a cube and set cube.data equal to the data mask (an array of Booleans). However, when I tried saving the cube with iris.save(), I got error
`TypeError: illegal primitive data type, must be one of ['i8', 'f4', 'f8', 'S1', 'i2', 'i4', 'u8', 'u4', 'u1', 'u2', 'i1'], got bool`

In NetCDF4, it is possible to save Boolean data with type 'b' (https://unidata.github.io/netcdf4-python/netCDF4/index.html#netCDF4.Dataset.createVariable) - could this not be added in Iris fairly simply to allow this kind of thing to be done?

PR #6305, opened 2025-02-05
Adds a warning for `iris.coord_categorisation.add_season_membership` that this is not saveable as a boolean coordinate.

HGWright:
### Description

This addresses #5909, by raising a warning that since this adds a boolean, the result will not be saveable to a Netcdf file since it contains a boolean coordinate.

**Note for reviewer:**
I was not 100% sure on the type of error to raise either `iris.warnings.IrisSaveWarning` or `iris.warnings.IrisCfSaveWarning` since this a problem with Netcdf files.


lib/iris/coord_categorisation.py
```diff
@@ -433,6 +435,12 @@ def add_season_membership(cube, coord, season, name="season_membership"):
     months = _months_in_season(season)
 
     def _season_membership(_, value: cftime.datetime) -> bool:
+        warnings.warn(
```
trexfeathers, line 438: This is an opportunity to inform Iris users in ALL cases of boolean data, since ALL cases will experience this problem.

Examples where the user will remain in the dark even after this PR is merged:

- Saving a boolean `Cube` (#3680)
- If we ever add another boolean coordinate category
- Some sort of flag on an `AncillaryVariable`. I have personally created examples like this to demonstrate Iris; e.g. a 2-dimensional `AncillaryVariable` varying over X and Y, indicating whether a cell is within a certain region (`True`) or not (`False`).

So could you move the warning and have it catch all cases of a boolean payload?

lib/iris/coord_categorisation.py
trexfeathers, line 1: Please can you add a What's New entry?

lib/iris/coord_categorisation.py
trexfeathers: Sorry I hadn't realised this was in `coord_categorisation` - a very immediate place for warning the user.

My suggestion was assuming this was part of the saving code. I retract my comment!

trexfeathers (approved):

Issue #6302, opened 2025-02-03
Provide "combine_cubes" function to integrate new merge/concatenate/load facilities


pp-mo:
See as proposed https://github.com/pp-mo/iris/pull/82, which targets the preliminary refactor in #6199 

The intention is to merge in several factors in one place :
  * make the old `LoadPolicy` an inheritor of a `CombinePolicy`, for more consistent naming ...
     * and incorporate a 'merge_unique' keyword ...
     * as mentioned [here](https://github.com/pp-mo/iris/pull/82#discussion_r1822451976)
  * include the new `equalise_cubes` operation as a programmable pre-merge/concatenate step within the combine operation, 
     * as mentioned [here](https://github.com/SciTools/iris/pull/6257)

I'm making a separate issue of this from https://github.com/pp-mo/iris/pull/82, just to be clear + label that I think we should aim to tidy all this up and get it into Iris v3.12

pp-mo:
Finished with v3.12 #6334

PR #6321, opened 2025-02-19
Move loading and combine code into their own submodules.

pp-mo:
Replaces #6199
This is supposedly a "pure refactor"
Except for some rewrites of docstrings, and changes to tests where they imported some (non-public) routines which have moved

Effectively the following things happened ..
  * `iris.LoadPolicy` has been renamed `CombineOptions` -- but `LoadPolicy` still exists as an alternative name
    * I'm intending to go on to extend the features here, as explained in #6302
    * I'm still not totally sure whether these two will remain identical.  For now they do.
  * the following definitions have moved from `iris.__init__` and are now defined in `iris.loading` 
     (but still publicly declared in the `iris` module) :
     `LoadPolicy`, `LOAD_POLICY`, `load`, `load_cube`, `load_cubes` and `load_raw`
     \-- though `LoadPolicy` is now "really" `CombinePolicy` and the definition is actually in `iris._combine`
  * some private routines also moved from `iris/__init__` to `iris.loading` : 
     `_generate_cubes`, `_load_collection`
  * the following definitions have likewise moved from `iris.cube` to `iris.loading` (though these ones aren't public) :
    `_CubeFilter`, `_CubeFilterCollection`
  * the following definitions moved from `iris.__init__` to `iris._combine` :
    `_combine_cubes`, `_combine_load_cubes`, and `CombineCubes` (which before was `LoadPolicy`)


lib/iris/__init__.py
```diff
@@ -283,515 +282,23 @@ def context(self, **kwargs):
...
-    with _raw_structured_loading():
-        return _load_collection(uris, constraints, callback).cubes()
+# NOTE: we define an independent local 'LOAD_POLICY' object here, just so that we can
+# ensure an entry in the API documentation output page.
 
+#: A control object containing the current file loading strategy options.
+LOAD_POLICY = iris.loading.LOAD_POLICY
 
-save = iris.io.save
+from iris.io import save
```
stephenworsley, line 301: Is there a reason for the import order for these save/load functions being so delayed? Would this not make more sense going next to `import iris.io`?

pp-mo: No real reason, and I'm surprised that the isort and ruff-type checks didn't complain about it.
To me it does make sense that, if I were defining these things, I'd put them here in the file, because that's where they "used to live".  But now I'm importing them from somewhere else.

If you think it now makes more sense., I'll treat these like imports, "LOAD_POLICY" included, and move it all to the top.

stephenworsley (approved):
Looks good, much neater this way.

pp-mo:
Thanks @stephenworsley !

PR #6324, opened 2025-02-19
Replace references of `conda install build` to `conda install python-build`

ESadek-MO:
### Description
`build` is now called `python-build` on conda (it is still just `build` in pip). It seems that the [feedstock was archived August '24](https://github.com/conda-forge/build-feedstock), and nothing else has broken, so I don't think this is a problem for Iris' automated processes, outside of references to it. 

This PR changes instructions in the docs for doing a manual pypi release.


bjlittle (approved):

trexfeathers:
Went looking for this so I could review. @bjlittle you're just so fast! 🔥

PR #6333, opened 2025-02-23
Add a text feature to quickplot

### LINKED ISSUE #6247: Add the option to have security classifications on plots
We've had a request for adding the option to have [security classifications](https://www.gov.uk/government/publications/government-security-classifications) on Iris plots (OFFICAL, OFFICIAL-SENSITIVE, etc) in a way that is somewhat controlled and consistent. 
 
We believe this could be possible as an optional parameter when creating a plot, but it wouldn't be realistic to automatically have it added due to external users. (In other words, you'd have to specify if you want a classification on your plot). 
 
We think this should only be a small or medium effort issue to add, (days rather than weeks).

ESadek-MO:
### Description
See #6247


docs/src/whatsnew/latest.rst
```diff
@@ -51,6 +48,11 @@ This document explains the changes made to Iris for this release
    However, :meth:`~iris.cube.Cube.transpose` will work, as will
    :meth:`~iris.cube.Cube.copy`. Note that, ``cube.copy(data=iris.DATALESS)``
    will provide a dataless copy of a cube. (:issue:`4447`, :pull:`6253`)
+   
+#. `@ESadek-MO`_ added a convenience kwarg to :mod:`iris.quickplot` to
+   add a consisently laid-out footer to plots.
```
bjlittle, line 53:
```suggestion
#. `@ESadek-MO`_ added the :mod:`iris.quickplot` ``footer`` kwarg to
   render text in the bottom right of the plot figure.
```

lib/iris/quickplot.py
```diff
@@ -7,11 +7,15 @@
 These routines work much like their :mod:`iris.plot` counterparts, but they
 automatically add a plot title, axis titles, and a colour bar when appropriate.
 
+These also have the optional kwarg `footer`, which adds text to the bottom right of
+the plot.
```
bjlittle, line 11:
```suggestion
These also have the optional kwarg ``footer``, which adds text to the bottom right of
the plot figure.
```

lib/iris/quickplot.py
```diff
@@ -139,6 +143,15 @@ def _label_1d_plot(*args, **kwargs):
     axes.set_ylabel(ylabel)
 
 
+def _footer(footer_text):
```
bjlittle, line 146:
```suggestion
def _footer(footer_text: str) -> None:
```

lib/iris/quickplot.py
```diff
@@ -139,6 +143,15 @@ def _label_1d_plot(*args, **kwargs):
...
+    if footer_text:
+        text = plt.figtext(x=0.99, y=0.01, s=footer_text, fontsize=12, ha="right")
```
bjlittle, line 148:
```suggestion
        text = plt.figtext(x=0.99, y=0.01, s=footer_text, fontsize=12, ha="right", va="bottom")
```

bjlittle (changes requested):
@ESadek-MO LGTM

Just a couple of minor suggestions to consider 👍

lib/iris/quickplot.py
bjlittle: @ESadek-MO See https://matplotlib.org/stable/gallery/text_labels_and_annotations/text_alignment.html

lib/iris/quickplot.py
```diff
@@ -139,6 +143,15 @@ def _label_1d_plot(*args, **kwargs):
...
+        text.set_path_effects(
+            [patheffects.Stroke(linewidth=3, foreground="white"), patheffects.Normal()]
+        )
+        plt.subplots_adjust(bottom=0.15, left=0.1)
```
bjlittle, line 152: Do you have an example of why this is needed?

Just curious ...

ESadek-MO: ...I can't replicate the original thing that made this necessary. I *do* think it looks better, do you foresee it causing problems?

bjlittle (approved):

Issue #6203, opened 2024-10-28
Make `_combine_cubes` public, expand API


stephenworsley:
## ✨ Feature Request

As captured in this to-do list here: https://github.com/SciTools/iris/pull/6168#issuecomment-2431699767, work done on #6168 can be expanded beyond the context of loading.

PR #6334, opened 2025-02-24
Combine cubes 2

pp-mo:
Rework replaces #6331, addresses #6203


pp-mo:
## Status update 2025-02-25
* functional code and APIs are basically complete and correct
* basic API documentation is correct
* doctests and examples need review
* all existing tests are passing
* most additional tests are now present
* some extra testing may be needed

The currently-failing doctests in the docsring of the `combine_cubes()` utility routine are just a placeholder.
(We do need a working example there...)

### My current details to-do list 
(maybe not exhaustive) ...
  * add unit tests for
     * [ ] `LoadPolicy.context()`
     * [ ]  `equalise_cubes_kwargs`control
  * add a proper example/doctests for
     * [ ] `CombineOptions`
     * [ ] `LOAD_POLICY` 
     * [ ] `comibine_cubes`
  * cross-reference user  documentation between (?all-of ?) ...
    * load functions
    * iris main module
    * LoadPolicy
    * user-guide loading chapter

pp-mo:
I'm going to mark this "ready for review", since it is (I think) functionally complete.
But I must admit the list of to-dos is a bit daunting.

For a functional query, I'm _still_ a bit unsure of the relationship complexity between CombineOptions and LoadPolicy,
not least because the default for the `options` of a `combine_cubes` call is defined by `LOAD_POLICY`.
And you can see that in the subsidiary [_combine_options_asdict](https://github.com/SciTools/iris/blob/88be3be20dd6f83d1eca3be2f4c9ac134e897e21/lib/iris/util.py#L2351) routine which I needed to pull out for sharing, which 'ought' to live in iris._combine, but actually relies on iris.loading .  Hmmm

Hopefully I can get a big chunk of that TODOs today.

pp-mo:
## Note: 
Following considerable offline  discussion with @trexfeathers @stephenworsley , 
We decided it was better to re-merge the definitions of `CombineOptions` and `LoadPolicy`
So I've now done that : https://github.com/SciTools/iris/pull/6334/commits/f7ecf8433e45f8159dab03a0bbe8e15647a2c472

In short, LOAD_POLICY and LoadPolicy are now just aliases for COMBINE_POLICY (object) and CombineOptions (class).
The notable reasons were :
  * since all APIs (internal included) took settings dictionaries  and not settings objects, there was no need ever to create a CombineOptions object
     * ... and clearly it was over-complex to make a **_separate_** global control object for `combine_cubes`, so I was going to be using `LOAD_POLICY` to control that too
  * in order to make either or both of `CombineOptions` and `LoadPolicy`docs legible, I was finding I had to replicate information between the two
  * the "settings" classes of a CombineObject wanted to be the same as those of a LoadPolicy, but these were determined by the needs of the loading modes -- and notably tje distinction between "default" and "legacy" which is meaningless for `combine_cubes`)
  
So in the end ... it made more sense to have a single definition, parts of which where irrelevant, than to separate the two for the sake of a small difference

pp-mo:
## Update 2025-02-26
I finally completed my checklist for adding tests :champagne: 

I've also added a whatsnew, and some minimal redirection links from the iris main module API and the userguide loading chapter.
HOWEVER I have also realised that it seems we are missing a mention of combine in the UG chapter on merge+concatenate.  I don't know quite how to contemplate that at this point.

For now, I'm just awaiting comments

pp-mo:
Rebased to update with main -- needed to fix whatsnew conflict

stephenworsley (changes requested):
Looking good. I'm pretty happy with the decision to simplify the API by making LOAD_POLICY a copy of COMBINE_POLICY, though I think it's worth adding another test to ensure that modifying one does indeed modify the other (and vice versa). I just have a couple suggestions for the documentation for the moment, I plan to go look through the testing next. So far, the code itself looks like a definite improvement.

lib/iris/cube.py
```diff
@@ -637,6 +637,20 @@ def concatenate(
             check_derived_coords=check_derived_coords,
         )
 
+    def combine(self, options: str | dict | None = None, **kwargs) -> CubeList:
```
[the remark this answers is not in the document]
pp-mo: 😮 how did I miss these ?!?

pp-mo:
> making LOAD_POLICY a copy of COMBINE_POLICY, though I think it's worth adding another test to ensure that modifying one does indeed modify the other

I don't think there's a real point to answer for this one, as they aren't copies but different names for the _actual same object_.
Both references are set to the same thing [here](https://github.com/SciTools/iris/blob/cc74cdaa3608144f2e9aecd4bbd8dd965b7e4c09/lib/iris/__init__.py#L115-L121)

pp-mo:
Update : thanks @stephenworsley 
I _think_ I addressed all comments to-date
please re-review !

stephenworsley:
> I don't think there's a real point to answer for this one, as they aren't copies but different names for the _actual same object_.
> Both references are set to the same thing [here](https://github.com/SciTools/iris/blob/cc74cdaa3608144f2e9aecd4bbd8dd965b7e4c09/lib/iris/__init__.py#L115-L121)

ah yes, that's what I meant, I was just imprecise (or incorrect) in my language. I ask because I have seen some confusion in the past over how to expect configuration objects to behave when being modified or set (e.g. code of the form `iris.COMBINE_POLICY = ...` rather than `iris.COMBINE_POLICY[key] = ...`, which I imagine is not what we would expect to support). I think I recall that for some non-standard use cases, it mattered how these objects were imported so I wondered if there was any call for tests for what happens to the same object imported with different names. If there was any non-trivial interaction we wanted to ensure was supported, that might be worth considering a test for, but it may be the case as you say that everything is either trivial or ought to be unsupported.

lib/iris/tests/unit/combine/test_CombineOptions.py
```diff
@@ -0,0 +1,215 @@
...
+        set_arg = {"merge_concat_sequence": "c", "repeat_until_unchanged": True}
+        expect_after = {"merge_concat_sequence": "c", "repeat_until_unchanged": True}
+        self.do_check(
+            op_arg=set_arg,
+            before_checks=expect_before,
+            after_checks=expect_after,
+            op_is_set=op_is_set,
+        )
+
+    def test_arg_kwargs(self, op_is_set):
```
stephenworsley, line 144: I think it's worth adding a similar test for when the arg is a string.

lib/iris/tests/unit/util/test_combine_cubes.py
```diff
@@ -0,0 +1,139 @@
+# Copyright Iris contributors
+#
+# This file is part of Iris and is released under the BSD license.
+# See LICENSE in the root of the repository for full licensing details.
+"""Unit tests for the :func:`iris.io.loading.combine_cubes` function.
+
+Note: These tests are fairly extensive to cover functional uses within the loading
+operations.
+TODO: when function is public API, extend testing to the extended API options,
```
stephenworsley, line 9: I believe this TODO is completed and can be removed now.

pp-mo: OK thanks!
[811168b](https://github.com/SciTools/iris/pull/6334/commits/811168bf9674023e330edf1f2e153b282a91cae0)

lib/iris/tests/unit/combine/test_CombineOptions.py
pp-mo: Yes I see, nice spot !
[0c11048](https://github.com/SciTools/iris/pull/6334/commits/0c1104890cc80bd3efc7b446f1492f1bda6e3103)

NOTE: I found I'm not actually using the "initial_options" arg of the `do_check` method any more, as all the current tests are in fact compatible with starting from a "default" state.  But I have left it in for now.
If you think this should be removed, I will do so.

pp-mo:
## Update 2024-03-6 
Thanks @stephenworsley responded to comments so-far + updated.
Please re-check !
Other things are merging at present, unfortunately every merge tends to create conflict in the whastnew/latest.rst.

lib/iris/util.py
```diff
@@ -2398,3 +2401,142 @@ def _print_xml(doc):
...
+    ...     print(combine_cubes(cubes))
+    ...
+    0: unknown / (unknown)                 (time: 5)
+
+    >>> # .. BUT NOTE : this modifies the original input cubes
+    >>> print(cubes[0].attributes.get("x"))
+    None
+
+    """
+    # TODO: somehow, provide a real + useful working code example
```
stephenworsley, line 2534: I believe this can be removed now.

lib/iris/_combine.py
```diff
@@ -278,36 +371,48 @@ def _combine_cubes(cubes, options, merge_require_unique):
...
 
         if sequence[0] == "c":
             # concat if it comes first
-            cubes = cubes.concatenate()
+            cubelist = cubelist.concatenate()
         if "m" in sequence:
             # merge if requested
-            cubes = cubes.merge(unique=merge_require_unique)
+            # NOTE: this needs "unique=False" to make "iris.load()" work correctly.
+            # TODO: make configurable via options.
```
stephenworsley, line 397: This can be removed, as with line 367 above.

stephenworsley (changes requested):
Just a couple TODOs to remove.

pp-mo:
OK I had to rebase yet again, but hopefully that addresses the stale TODOs you pointed out.

stephenworsley (approved):
Looks good, I think this is mergable now.

pp-mo:
Many thanks @stephenworsley !

PR #6337, opened 2025-02-26
Faster dimension lookup for derived coordinates

bouweandela:
### Description

This makes looking up the dimensions of derived coordinates faster by first looking if the metadata of the requested coordinate matches that of an AuxFactory instead of creating the derived coordinate and then comparing the metadata. Creating a derived coordinate is a relatively slow operation.

---

---


pp-mo (approved):
LGTM

pp-mo:
I see from [your comment](https://github.com/SciTools/iris/discussions/6134#discussioncomment-12350990) that this was sort-of in pursuit of something else,  but it's an easy win so let's dispose of this one first !

Issue #6316, opened 2025-02-18
Simplify netcdf loading rules/actions


pp-mo:
In the course of working on loading tolerance work for https://github.com/SciTools/iris/issues/5165#issuecomment-2349326640 within the [3.12 sprint work](https://github.com/orgs/SciTools/projects/45/views/1), we have been reminded that the original work to [replace Pyke rules with our own "custom rules engine" approach](https://github.com/SciTools/iris/pull/4198) has effectively left **a lot of inefficient and complex (non-functional) scaffolding in place**

The netcdf loading "actions" code was written this way, basically, to maximise similarities with what went before, but it leaves considerable opportunities to simplify + improve.

**(1)** We (I!) wrote a lot of code designed to replicate the firing of the equivalent original logic rules ([here](https://github.com/SciTools/iris/blob/v3.11.1/lib/iris/fileformats/_nc_load_rules/actions.py#L344-L374) for example), allowing us to debug the operation of the code and compare it to the original implementation, in case problems arose with loading some files (compared to older Iris versions).
In practice no such problems really happened, so this complex debugging/logging is pretty redundant (as it was in the original rules code, really).

**(2)** Likewise, the way the "facts" of the current loading case (per cube) are stored in a database is pretty cumbersome + obscure (notably [here](https://github.com/SciTools/iris/blob/v3.11.1/lib/iris/fileformats/_nc_load_rules/engine.py#L48-L61)), and could be usefully streamlined.

So in both those areas, there's a big opportunity to simplify the code and improve maintainability, and probably efficiency too.

trexfeathers:
From @SciTools/peloton: this continues to be important - NetCDF loading is a frequently visited area of the codebase, and each change we want to make is slowed down by the legacy structure.

PR #6338, opened 2025-02-26
Tolerant handling of `standard_name` and dimension coordinate loading

### LINKED ISSUE #6318: Capture un-loadable NetCDF dimension coords and `Cube` names
See #6317 for more specifics

- [ ] Iris `__init__.py`
- [ ] `helpers.py`
- [ ] `actions.py`
- [ ] Tests
- [ ] Confirm this can fix known cases (#5165)
- [ ] Docstrings
- [ ] What's New

trexfeathers:
### Description

Closes #6318

#### To-do

- [x] Raise a warning at the end of loading if anything is present in `iris.loading.LOAD_PROBLEMS`
- [x] Docstring and UX improvements to `iris.loading.LOAD_PROBLEMS`.
- [x] Get feedback from reviewer on my approach of adding new routines called `build_and_add_*`
  - Is the philosophy sound?
  - Is the name too clunky?
- [x] Fix existing unit tests now that I have changed how things work. These failures don't look scary
- [x] Add new tests for `iris.loading.LOAD_PROBLEMS` and the new warning

#### Deviation from the original spec

To avoid making too many enhancements to `actions.py` - which we consider legacy (#6316) - I decided it was better to create new routines for adding elements to the `Cube`, and house them in `helpers.py`, with the rest of the non-Pyke code. If the reviewer is happy with this approach, I will rewrite the spec in the issues.

---

---


lib/iris/fileformats/_nc_load_rules/actions.py
```diff
@@ -367,9 +375,34 @@ def action_build_dimension_coordinate(engine, providescoord_fact):
...
+            message += rule_name
+            error = ValueError(message)
+
+        try:
+            # TODO: is there a better way to create a stack trace?
+            #  Is an ST even valuable? `error` can be passed straight into
+            #   from_exception() without raising it first.
+            raise error
+        except error.__class__ as error:
+            tb_exception = TracebackException.from_exception(error)
```
trexfeathers, line 397: 📢

lib/iris/fileformats/_nc_load_rules/helpers.py
```diff
@@ -1198,33 +1376,70 @@ def build_dimension_coordinate(
...
-            )
-            coord_skipped = True
 
-    if not coord_skipped:
+    return coord
+
+
+# TODO: propagate the the build-and-add pattern to all other objects (iris#6319).
+# TODO: this naming convention is clunky. Suggestions welcome.
+def build_and_add_dimension_coordinate(
```
trexfeathers, line 1385: 📢

lib/iris/fileformats/netcdf/loader.py
```diff
@@ -655,6 +659,17 @@ def load_cubes(file_sources, callback=None, constraints=None):
...
+    # TODO: consider propagating to other loaders in future.
+    #  Could not apply at a higher level as user support and ncdata both access
+    #  this lower-level loader.
+    if _profile_load_problems() != load_problems_original:
+        # TODO: should this print iris.loading.LOAD_PROBLEMS?
+        message = (
+            "Not all file objects were parsed correctly. See "
+            "iris.loading.LOAD_PROBLEMS for details."
+        )
+        warnings.warn(message, category=IrisLoadWarning)
```
trexfeathers, line 671: 📢

lib/iris/loading.py
```diff
@@ -322,3 +328,207 @@ class LoadPolicy(CombineOptions):
...
+
+    - ``"".join(stack_trace.format())``: the full stack trace as a string - the
+      same way this would be seen at the command line.
+    - ``stack_trace.exc_type``: the exception type e.g. :class:`ValueError`.
+    """
+
+
+# TODO: could this be a context manager in future?
+# TODO: include in an __all__ somewhere
+LOAD_PROBLEMS: dict[Path, list[LoadProblemsEntry]] = {}
```
trexfeathers, line 370: 📢

lib/iris/loading.py
```diff
@@ -322,3 +328,207 @@ class LoadPolicy(CombineOptions):
...
+        basic_value: str | None = None
+        if hasattr(entry.loaded, "keys"):
+            assert isinstance(entry.loaded, dict)
+            (basic_value,) = entry.loaded.keys()
+        elif hasattr(entry.loaded, "var_name"):
+            assert isinstance(entry.loaded, Cube)
+            attributes = entry.loaded.attributes[LimitedAttributeDict.IRIS_RAW]
+            # TODO: should "var_name" also be encoded as a constant?
+            #  Is there a better way to store this info in the first place?
+            basic_value = attributes["var_name"]
```
trexfeathers, line 525: 📢

trexfeathers (commented):
Here are some to-do's I would appreciate input on.

lib/iris/fileformats/_nc_load_rules/actions.py
pp-mo: I've tried to follow through the docs on this, and they do say 
[on Traceback objects](https://docs.python.org/3/reference/datamodel.html#traceback-objects) : "A traceback object is implicitly created when an exception occurs, and may also be explicitly created by calling `types.TracebackType`".
But TBH it seems that it is _normally_ the `raise` statement itself tdoes the work of walking the stack and creating a list of `Traceback` objects.  So I think it's still **_far_** easier to just use one to do that, even in this case where you are using a 'raise' yourself for the sake of it, just to produce a stacktrace consistent with other failure mechanisms.

pp-mo: After some reading, I'm also in agreement on the use of `TracebackException.from_exception(error)`.
This way, I think we get the equivalent of a stacktrace **_printout_**, but it loses references to actual _objects_ in the frames (like pdb would use).  Which is the right level of information (and will even serialise if needed).

It's possible this still stores too much, if I misunderstood, but in that case I think you could replace it with a printout of the same thing, i.e. a list of strings (like the `.format() you mention).

lib/iris/loading.py
pp-mo: I think for safety you should define this to be a special class, and provide defined operations to access it.

I'm thinking that this will future-proof it, in case we needed to guard operations for thread-safety to prevent some loading operations happening in overlapping actions.  
I rather doubt that is possible now, but it would force a breaking change to the interface if it became necessary, which we can avoid by providing a defined interface (with its own type) now.

I would also like to solve the problem of resetting the record for a particular file when you start to load it, which I don't think we currently do , so re-loading just adds to the existing record.  
(see my other comment) 
Unless I misunderstood ?

Suggest:
```
class LoadProblems():
    def reset(filename: str):
        ...
    def record(filename: str , entry: LoadProblemsEntry):
         ...
    def report(filename:str | None) -> list[LoadProblemsEntry]:
         ...

LOAD_PROBLEMS: LoadProblems = LoadProblems()
```
If needed, we can add a single, or a per-file mutex (later)

pp-mo (changes requested):
Posting what I got so far.
Possibly ongoing, though

lib/iris/fileformats/netcdf/loader.py
trexfeathers: Offline, we agreed:

- I am going to move the warning up to `_load_collection`. Anyone directly calling lower level operations is a super-user and can cope without the warning. Or they are using ncdata and @pp-mo is hoping to change how ncdata interacts with the Iris loader
- There is no need to do any automatic clearing, as this makes sweeping assumptions about user needs. However it should be well advertised that `LOAD_PROBLEMS` _can_ be cleared, perhaps when making a specialised class object - see https://github.com/SciTools/iris/pull/6338#discussion_r1985128826

lib/iris/loading.py
trexfeathers: Extensive refactor to this effect: bc98b2f72362998320567d1f4d86a823c55b3009

Once I had gone through the effort of making a dedicated `LoadProblems` class, this also provided several opportunities to neaten the code.

lib/iris/fileformats/netcdf/loader.py
trexfeathers: >Offline, we agreed:
>...

The refactor in bc98b2f72362998320567d1f4d86a823c55b3009 addresses both these bullets:

- Having a dedicated `LoadProblems` class allowed me to address warnings directly in there, so I no longer need to worry about where to raise the warning. Note that this meant several more tests in `tests/unit/fileformats/nc_load_rules/actions/` raised warnings where they weren't raised before.
  https://github.com/SciTools/iris/blob/5a7f68c10fd6f468457bb768e1064ad586aca91e/lib/iris/loading.py#L579-L587
- The new `LoadProblems` class implements the suggested `reset()` method, which includes a docstring, and a parameter to select a specific file to 'clear'.

lib/iris/tests/unit/loading/test_load_problems.py
```diff
@@ -0,0 +1,163 @@
...
+    assert str(load_problems_instance) == expected
+
+
+def test_problems_property(load_problems_instance):
+    assert load_problems_instance.problems == load_problems_instance._problems
+
+
+def test_problems_by_file_property(load_problems_instance):
+    expected = {p.filename: [p] for p in load_problems_instance._problems}
+    assert load_problems_instance.problems_by_file == expected
```
pp-mo, line 98: This bit looks neat, but it doesn't work if there are >1 entries for a given filepath (which you don't have in your test example).  You get a result with a list of only one result (the last entry encountered) for each file.
The implementing code with its loop + setdefault **_does_** work.

Sort-of-equivalent demonstration:
```
>>> # example seems to work ...
>>> ll = [('a', 1, 'this'), ('b', 2, 'that')]
>>> {x[0]: x for x in ll}
{'a': ('a', 1, 'this'), 'b': ('b', 2, 'that')}
>>> 
>>> # expanded, shows it doesn't ...
>>> ll = [('a', 1, 'this'), ('a', 2, 'other'), ('b', 2, 'that')]
>>> {x[0]: x for x in ll}
{'a': ('a', 2, 'other'), 'b': ('b', 2, 'that')}
```

pp-mo (changes requested):
Pretty solid !
Just a few small changes. mostly to tests

lib/iris/tests/unit/fileformats/nc_load_rules/helpers/test_build_and_add_dimension_coordinate.py
```diff
@@ -233,17 +232,16 @@ def test_aux_coord_construction(self):
...
+        with self.deferred_load_patch, self.get_cf_bounds_var_patch:
+            build_and_add_dimension_coordinate(self.engine, self.cf_coord_var)
 
             # Test that expected coord is built and added to cube.
             self.engine.cube.add_aux_coord.assert_called_with(expected_coord, [0])
+            load_problem = LOAD_PROBLEMS.problems[-1]
             self.assertIn(
-                "creating 'wibble' auxiliary coordinate instead",
-                warnings.warn.call_args[0][0],
+                "creating wibble auxiliary coordinate instead",
```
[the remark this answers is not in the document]
trexfeathers: 7e6c958f336034eff3cd33428a3f17637c450206

lib/iris/fileformats/_nc_load_rules/helpers.py
```diff
@@ -1177,16 +1341,20 @@ def build_dimension_coordinate(
...
-    except ValueError as e_msg:
+    except ValueError as dim_error:
         # Attempt graceful loading.
-        msg = (
-            "Failed to create {name!r} dimension coordinate: {error}\n"
-            "Gracefully creating {name!r} auxiliary coordinate instead."
+        coord_var_name = str(cf_coord_var.cf_name)
+        dim_error.add_note(
+            f"Failed to create {coord_var_name} dimension coordinate:\n"
+            f"Gracefully creating {coord_var_name} auxiliary coordinate instead."
```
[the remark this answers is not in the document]
trexfeathers: 7e6c958f336034eff3cd33428a3f17637c450206

lib/iris/tests/unit/loading/test_load_problems.py
trexfeathers: 93045b35cba8f78ea738262717f9eada9ac8b18d

pp-mo (approved):
All good now !

trexfeathers:
Thanks @pp-mo!

PR #6339, opened 2025-02-26
Use realized data in benchmarks when requested

bouweandela:
### Description

Benchmarking with realized data did not actually happen. This pull request updates the patch method so realized data is used in benchmarks that request it. Note that this does not apply to derived coordinates, which are always lazy.

---

---


pp-mo (approved):
Sorry, I should really have merged this one some time ago,
but I sort-of forgot it was not already done !

Issue #6149, opened 2024-09-20
Iris won't load netCDF files from the new CDS-beta


leosaffin:
## 🐛 Bug Report

Iris won't load netCDF files from the new CDS-beta. Someone has also reported this on a post about the updates to the CDS API here (https://forum.ecmwf.int/t/changes-to-grib-to-netcdf-converter-on-cds-beta-ads-beta/4322/19). I suspect the issue comes from the changes to their converter described there. However, I think iris not loading the files is an iris bug. If I change the offending line (traceback below) in iris from

`total_bytes = cf_var.size * cf_var.dtype.itemsize`

to

`total_bytes = cf_var.size * np.dtype(cf_var.dtype).itemsize`

Then the file loads fine. Simple fix but I thought it's worth reporting in case this might lead to other issues elsewhere.

Running ncdump on the file shows up the difference too. The variables look like this. The `string` in front of the string variables wouldn't have been there before. I don't know if that is an issue or not.

```
double latitude(latitude) ;
    latitude:_FillValue = NaN ;
    string latitude:units = "degrees_north" ;
    string latitude:standard_name = "latitude" ;
    string latitude:long_name = "latitude" ;
    string latitude:stored_direction = "decreasing" ;
```

## How To Reproduce
Steps to reproduce the behaviour:

1. Download a netCDF file from CDS-beta. I just ticked the first box for each value here (https://cds-beta.climate.copernicus.eu/datasets/reanalysis-era5-pressure-levels?tab=download), i.e. Divergence at 1hPa, 1st January 1940
2. Try to open it with iris

## Expected behaviour

The netCDF file should load. It works with xarray and CF python

## Additional context

<details>
<summary>Click to expand <b>this section...</b></summary>

---------------------------------------------------------------------------
AttributeError                            Traceback (most recent call last)
Cell In[2], line 1
----> 1 cubes = iris.load("test_era5.nc")

File ~/miniforge3/envs/core/lib/python3.12/site-packages/iris/__init__.py:326, in load(uris, constraints, callback)
    302 def load(uris, constraints=None, callback=None):
    303     """Load any number of Cubes for each constraint.
    304 
    305     For a full description of the arguments, please see the module
   (...)
    324 
    325     """
--> 326     return _load_collection(uris, constraints, callback).merged().cubes()

File ~/miniforge3/envs/core/lib/python3.12/site-packages/iris/__init__.py:294, in _load_collection(uris, constraints, callback)
    292 try:
[...]

PR #6340, opened 2025-02-26
Handle NetCDF variable length strings (and other VLen types)

ukmo-ccbunney:
### Description
NetCDF provides a "variable length" data type (`VLen`) that can be used to store arrays where one dimension size is unknown/variable.

This variable length datatype is used when creating a new variable with the `str` type (crucially this is different from a `NC_CHAR` type).

Iris fails to load `VLen` `str` types as it cannot determine the itemsize of a `str` type when trying to calculate whether the array is big enough to lazy load.

Other `VLen` types load fine (such as floats, chars, ints, etc), but the calculation for the total array size will be incorrect as the size of the variable length array dimension is not known until the data has been read from disk.

This PR makes the following changes
1.  If the datatype is a NetCDF VLen type a Warning is raised to uniform the user that the total datasize cannot be known until the data has been loaded and this will affect the lazy loading.
2. If the datatype is a `str` type, a default itemsize of 4 bytes is used (sufficient for storing a single Unicode character)
3. The default_fill_value for `str` types is not a think in netCDF4, so I have made it the same as the 'S1' type (a null byte).
4. If the user has some apriori knowledge of the mean length of the variable dimension, this can be passed in via the `CHUNK_CONTROL` context manager using a special `_vl_hint` dimension name.

Fixes: #6149


pp-mo:
The problems with `test_coding_standards` are because it _specifically_ checks that no code is using the `netCDF4` module directly, but only via `iris.fileformats.netcdf.+thread_safe_nc`.
My suggested solution : https://github.com/ukmo-ccbunney/iris/pull/1

lib/iris/fileformats/netcdf/loader.py
```diff
@@ -215,10 +215,41 @@ def _get_cf_var_data(cf_var, filename):
...
+
+            # A stab in the dark at the mean size of the variable length arrays (this could be anything):
+            mean_vl_array_len = 10
+
+            # Give user the chance to pass a hint of the average variable length array size via
+            # the chunk control context manager. This allows for better decisions to be made on
+            # whether the data should be lazy-loaded or not.
+            if CHUNK_CONTROL.mode is not CHUNK_CONTROL.Modes.AS_DASK:
+                if chunks := CHUNK_CONTROL.var_dim_chunksizes.get(cf_var.cf_name):
+                    if vl_chunk_hint := chunks.get("_vl_hint"):
```
pp-mo, line 241: This is quite neat, but I guess it needs documenting **_somewhere_**.
I guess it has to be in the CHUNK_CONTROL docs around [here](https://scitools-iris.readthedocs.io/en/stable/generated/api/iris.fileformats.netcdf.loader.html#iris.fileformats.netcdf.loader.ChunkControl.set) somewhere 
(among the 'dimensions' param, and/or the subsequent Notes section)

pp-mo (changes requested):
Good stuff, solves the problem nicely I think :+1:

Apart from the outstanding test fail 
 -- which I have proposed a fix for (see [comment](https://github.com/SciTools/iris/pull/6340#issuecomment-2688698975))
 
Also a couple of small comments + docs to adjust, as noted.

pp-mo:
BTW I think it's **_really_** important that people can get what they want out of these cases by taking control of the variable chunking.  Which currently doesn't work since it hits the 'itemsize' line and crashes.
So, I think you've now nailed that problem and bullet-proofed us against that failure mode.  No need for a special integration test.
I think it's rather cool that we have a drop on xarray here : we can specify chunking on a per-variable basis, but they don't

lib/iris/fileformats/netcdf/loader.py
```diff
@@ -215,10 +215,41 @@ def _get_cf_var_data(cf_var, filename):
         # See https://github.com/SciTools/iris/issues/4994 "Xarray bridge".
         result = cf_var._data_array
     else:
-        total_bytes = cf_var.size * cf_var.dtype.itemsize
+        # Determine size of data; however can't do this for variable length (VLEN)
+        # arrays as the size of the array can only be known by reading the data.
+        # "Variable length" netCDF types have a datatype of `nc.VLType`.
+        if isinstance(
+            getattr(cf_var, "datatype", None), _thread_safe_nc.netCDF4.VLType
+        ):  # TODO(ChrisB): I am accessing netCDF4 directly here - is this ok? Just for type comparison.
```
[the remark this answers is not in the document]
ukmo-ccbunney: Great. Removed comment here (along with some other comments changes); fc871ba12
This commit also adds `_MEAN_VL_ARRAY_LEN` as a module level variable.

lib/iris/fileformats/netcdf/loader.py
ukmo-ccbunney: Yes - since you're happy with this approach I will now put the some docs together.

ukmo-ccbunney:
> BTW I think it's **_really_** important that people can get what they want out of these cases by taking control of the variable chunking. Which currently doesn't work since it hits the 'itemsize' line and crashes. So, I think you've now nailed that problem and bullet-proofed us against that failure mode - **yes??** In whcih case, OK, and no need for a special integration test (which I might otherwise ask for).
> 
> I think it's rather cool that we have a drop on xarray here : we can specify chunking on a per-variable basis, but they don't

Yes - to a degree.

Users can now pass down custom chunk parameters for variable length arrays, but as the "ragged array size" is not defined by an actual dimension we still have not way of passing this down to Dask. We can hint on the mean ragged array length to aid the decision for lazy loading, but that's about the extent of it.

Your chunks will always be bigger in memory than you expect for **variable length arrays** as the chunking is only working over the explicit known dimensions.

This does at least allow is to load variable length string data now in a way that is consistent with how other variable length data was being handled.

pp-mo:
> Users can now pass down custom chunk parameters for variable length arrays, but as the "ragged array size" is not defined by an actual dimension we still have not way of passing this down to Dask. We can hint on the mean ragged array length to aid the decision for lazy loading, but that's about the extent of it.
> 
> Your chunks will always be bigger in memory than you expect for **variable length arrays** as the chunking is only working over the explicit known dimensions.

I totally agree with that -- the chunks of course don't have known sizes.  
But you do get to choose how it is chunked.  So, if the user knows something about the worst typical cases, they can make sensible decisions.
Which is just as much as anyone can ask for.

lib/iris/fileformats/netcdf/loader.py
```diff
@@ -215,23 +220,53 @@ def _get_cf_var_data(cf_var, filename):
...
+            itemsize = 4 if cf_var.dtype is str else cf_var.dtype.itemsize
+
+            # For `VLType` cf_var.size will just return the known dimension size.
+            total_bytes = cf_var.size * mean_vl_array_len * itemsize
+        else:
+            # Normal NCVariable type:
+            total_bytes = cf_var.size * cf_var.dtype.itemsize
+
         if total_bytes < _LAZYVAR_MIN_BYTES:
-            # Don't make a lazy array, as it will cost more memory AND more time to access.
```
pp-mo, line 220: Why did you nuke this comment? 
it seems useful to me

ukmo-ccbunney: Oops - that was a mistake. Thanks for spotting.
Reinstated; 0db937126

lib/iris/fileformats/netcdf/loader.py
ukmo-ccbunney: Docstring added to the CHUNK_CONTROL class.
New "Variable-length Datatypes" section added to "Further Topics - NetCDF I/O".

ukmo-ccbunney:
New sample data is now being picked up from [iris-sample-data](https://github.com/SciTools/iris-sample-data/pull/34) and doctests are now succeeding. This one is ready for a final review now.

pp-mo:
Some earlier comments still to address, otherwise updated sample-data now functioning and "good to go" :tm: I think !

ukmo-ccbunney:
> Some earlier comments still to address (above, all in the docs)
> 
> Otherwise it has updated + the sample-data now functioning then "good to go" ™️ I think !

The docs should already be up-to-date:
- The `CHUNK_CONTROL.set` docstring has been updated here: [iris.fileformats.netcdf.loader.ChunkControl.set](https://scitools-iris--6340.org.readthedocs.build/en/6340/generated/api/iris.fileformats.netcdf.loader.html#iris.fileformats.netcdf.loader.ChunkControl.set)
- I've also added some details on variable-length types here: [Further Topics - NetCDF I/O Handling - Variable-length datatypes](https://scitools-iris--6340.org.readthedocs.build/en/6340/further_topics/netcdf_io.html#variable-length-datatypes)

However, the doctest GHA is failing due to the RTD environment having an older version of iris-sample-data (2.5.1).
I assume it is lockfiles need cycling?

pp-mo:
> However, the doctest GHA is failing due to the RTD environment having an older version of iris-sample-data (2.5.1). I assume it is lockfiles need cycling?

No I don't think it uses lockfiles.
AFAIK it builds an env from the `./requirements/readthedocs.yml` --> `./requirements/iris.yml` --> `./requirements/py312.yml`
(they are softlinks), so ultimately it is just an env spec containing `iris-sample-data >=2.4.0`.
Perhaps it is cached by RTD and somehow that needs a 'poke' to tell it that something newer is available.

I'll see if I can find something about it ..

pp-mo:
> However, the doctest GHA is failing due to the RTD environment having an older version of iris-sample-data (2.5.1). I assume it is lockfiles need cycling?

Oh perhaps it could be..
the docstests are run with sphinx, via nox.
So I guess they do use lockfiles.
I think you can just cheat and update that tiny bit of a lockfile, since iris-sample-data has no deps

ukmo-ccbunney:
> > However, the doctest GHA is failing due to the RTD environment having an older version of iris-sample-data (2.5.1). I assume it is lockfiles need cycling?
> 
> Oh perhaps it could be.. the docstests are run with sphinx, via nox. So I guess they do use lockfiles. I think you can just cheat and update that tiny bit of a lockfile, since iris-sample-data has no deps

OK - have updated these and doctests are succeeding now.

pp-mo:
@ukmo-ccbunney testing all good now, 
but pending comments still to fix.
Feel free to push back.

I'm really keen to get this in !

pp-mo (changes requested):

docs/src/further_topics/netcdf_io.rst
```diff
@@ -122,6 +122,47 @@ Iris' optimisation all together, and will take its chunksizes from Dask's behavi
     (70, 37, 49)
 
 
+Variable-length datatypes
+-------------------------
+
+The NetCDF4 provides support for variable-length (or "ragged") data types (``VLType``);
+see https://unidata.github.io/netcdf4-python/#netCDF4.Dataset.vltypes
```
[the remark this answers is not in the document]
ukmo-ccbunney: Yes - that was my intended link target. Not sure how I ended up with that other one - thanks for spotting that.
Fixed in da1855c55

docs/src/further_topics/netcdf_io.rst
```diff
@@ -122,6 +122,47 @@ Iris' optimisation all together, and will take its chunksizes from Dask's behavi
...
+whether the load the data lazily or not. If the user has some apriori knowledge of
+the average size of the variable length dimension, this can be provided as a hint
+to  Iris via the ``CHUNK_CONTROL`` context manager and the special ``_vl_hint``
+keyword targeting the variable, e.g. ``CHUNK_CONTROL.set("varname", _vl_hint=5)``.
+
+For example, consider a netCDF file with an auxiliary coordinate
+``experiment_version`` that is stored as a variable-length string type.
+By default, Iris will make a guess of the total array size based on the known
+dimension sizes (``time=150`` in this example) and load the data lazily. However,
+it is known prior to loading that the experiment versions are all no longer than
```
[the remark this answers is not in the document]
ukmo-ccbunney: This has been addressed in https://github.com/SciTools/iris/pull/6340/commits/252fa637d8438ad4953c998c0be10027d2cf93bd

docs/src/further_topics/netcdf_io.rst
```diff
@@ -122,6 +122,47 @@ Iris' optimisation all together, and will take its chunksizes from Dask's behavi
...
+particular array element can vary the values are stored as an array of ``VLType``.
+
+Unfortunately, the size of a ``VLType`` variable cannot be determined without loading
+the data first, which can make it difficult for Iris to make an informed decision on
+whether the load the data lazily or not. If the user has some apriori knowledge of
+the average size of the variable length dimension, this can be provided as a hint
+to  Iris via the ``CHUNK_CONTROL`` context manager and the special ``_vl_hint``
+keyword targeting the variable, e.g. ``CHUNK_CONTROL.set("varname", _vl_hint=5)``.
+
+For example, consider a netCDF file with an auxiliary coordinate
```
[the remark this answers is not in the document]
ukmo-ccbunney: Done; see f6cb8fe98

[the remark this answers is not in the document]
ukmo-ccbunney: Actually, this adds a "For example" in the side bar structure, which looks a bit odd.
I'll change it slightly...

docs/src/further_topics/netcdf_io.rst
```diff
@@ -122,6 +122,61 @@ Iris' optimisation all together, and will take its chunksizes from Dask's behavi
...
+
+VLType size hinting
+^^^^^^^^^^^^^^^^^^^
+
+If the user has some *a priori* knowledge of the average length of the data in
+variable-length ``VLType``, this can be provided as a hint to Iris via the
+``CHUNK_CONTROL`` context manager and the special ``_vl_hint`` keyword
+targeting the variable, e.g. ``CHUNK_CONTROL.set("varname", _vl_hint=5)``.
+This allows Iris to make a more informed decision on whether to load the
+data lazily.
```
pp-mo, line 155: This section is great now, but I still have a slight doubt about this last bit.

In my mind, the primary purpose is to decide chunk sizes, not whether to make the content lazy at all -- that's just a bonus in special cases.

How do you feel about tweaking this ?
If you disagree, or you don't have time now, we can be pragmatic + merge as-is !

pp-mo, line 155: This section is great now, but I still have a slight doubt about this last bit.

In my mind, the primary purpose is to decide chunk sizes, not whether to make the content lazy at all -- that's just a bonus in special cases.

How do you feel about tweaking this ?
If you disagree, or you don't have time now, we can be pragmatic + merge as-is !

pp-mo (changes requested):
Just one tiny further suggestion ...

docs/src/further_topics/netcdf_io.rst
ukmo-ccbunney: The `_vl_hint` only affects the lazy loading calculation so I think this comment is still pertinent.

pp-mo: OK I get it now.

I had been working under the mistaken assumption that we somehow passed an estimated 'itemsize' to the Iris automatic chunking algorithm in "as_lazy_data".
But now I understand that this only uses "var.shape" and "dtype.itemsize", so it basically makes a nonsense of the variable-length cases.
Potentially we could do something better by passing the "_vl_hint" into that calculation too, but perhaps that is a step too far ?
What the user **_can_** (already) do is to take control of the chunking for the variable, but it must be complete control since otherwise limiting chunking in one dimension will cause Iris to expand the chunksize in other dims.

PR #6312, opened 2025-02-11
Unpin `dask` and pin it to `>=2025.1.0` and unSKIP doctests

valeriupredoi:
### Description

Dask solved their severe issue reported in https://github.com/SciTools/iris/issues/6251 in 2025.1.0, and unpinning - and pinning it to the earliest version that solves that issue was raised in https://github.com/SciTools/iris/issues/6264

I am also unSKIPping the doctests that @trexfeathers  mentioned in https://github.com/SciTools/iris/issues/6264#issuecomment-2551193762

Closes https://github.com/SciTools/iris/issues/6264

---
[Consult Iris pull request check list]( https://scitools-iris.readthedocs.io/en/latest/developers_guide/contributing_pull_request_checklist.html)

---
Add any of the below labels to trigger actions on this PR:

- https://github.com/SciTools/iris/labels/benchmark_this

PR #6342, opened 2025-02-27
Unpin dask 2

stephenworsley:
Completes #6312.


ESadek-MO (changes requested):
Happy with all the changes, tests seem to work as they used to (as a bonus, is more readable in places too).

Only thing left is I think it might be worth mentioning this is the dependencies whatsnew, as the pin was mentioned there originally.

ESadek-MO (approved):

PR #6348, opened 2025-02-28
Correct unit to units in AuxCoord docstring

### LINKED ISSUE #6347: Docstring for iris.coords.AuxCoord incorrectly has a `unit` argument, rather than `units`
## 📚 Documentation

In the docstring for `iris.coords.AuxCoord` there is incorrectly listed a `unit` argument, rather than `units`. All other Coords document a `units`, and the implementation requires it be called `units`, so I reason it is the documentation that is wrong.

You can see it it the docs at the following link, but this also affects type hints in editors. (Which is what got me.)
https://scitools-iris.readthedocs.io/en/stable/generated/api/iris.coords.html#iris.coords.AuxCoord

jfrost-mo:
### Description

Fixes #6347 by correcting the AuxCoord docstring.

---

---


ESadek-MO (approved):
Thanks @jfrost-mo, looks good to me, thanks for catching that!

PR #6349, opened 2025-02-28
Trial combination of bouwe array-equal-fix and non-lazy-benchmarks

pp-mo:
My manual merge of #6325 and #6339 just to see how performance is affected.

NB since I already merged #6337 the benefit from that has already been had.
So, if this benchmarks slower in the expected cases, we should reference to those improvements too, since they were intended to mitigate possible costs of #6325


pp-mo:
## Update after re-benchmark run
the #6337 speedups were about -20%..-30%, so roughly cancel out the slowdowns.
but they only apply to the ones named `cube.CubeEquality.time_equality(... 'coord_inequality')`
so that doesn't cover everything that gets hit here

All of this is pretty much the same as what we saw before.
We do have some slowdowns which are still to be taken on board.  But the existing behaviour does seem to be plain wrong.

pp-mo:
Rebased from updated #6325 with added commit "@bouweandela
[Use dask.array.blockwise for array comparison](https://github.com/SciTools/iris/pull/6325/commits/1abf173a9fafe448ad40bf6650483db5b94647d6)" 

And re-booted benchmarks ...

Issue #6317, opened 2025-02-18
Capture un-loadable NetCDF objects instead of discarding/crashing


trexfeathers:
Closes #5165

## User story

- **As an**: Iris user loading NetCDF files written according to the CF Conventions.
- **I want**: Iris to capture malformed CF information during loading, instead of crashing or disposing of it.
- **So that**: I can fix CF problems within my Iris script - avoiding the complexity of multiple scripts/tools.

## Why this is hard

There are many ways Iris crashes when it encounters bad CF. It is tempting to think of these crashes as deliberate - with easy-to-modify code blocks for each rule - and there are a few of these, but Iris is not a CF-checker. Instead we have used CF to make assumptions so that the code can be simpler/smaller; barely any of the crashes are raised from dedicated lines, and they are often hard to predict.

Any 'fix' must therefore be a form of generic error handling, which cannot have knowledge of what precisely might go wrong.

## The architecture

- `iris.LOAD_PROBLEMS` - a global object where Iris can capture objects that could not be loaded, and the stack trace error that was raised.
- `build_raw_cube()` - a routine that will represent any `CFVariable` as a very basic `Cube`, with as little interpretation as possible.
- Separate building objects versus adding them to the `Cube` being loaded.
- Ensure all objects - including names, units, etcetera - are contained within their own building routine.
- `_add_or_capture()`:
  - Can't build the object? Use `build_raw_cube()` and store in `iris.LOAD_PROBLEMS`.
  - Object built but can't add to the `Cube`? Store in `iris.LOAD_PROBLEMS`.
- Issue a warning at the end of loading if anything is found within `iris.LOAD_PROBLEMS`.
- Make it easier to convert `Cube`s - output by `raw_cube_from_cf_var()` - into other objects e.g. `DimCoord`s? Documentation at the very least.

Note that I have checked `cf.py` and believe it can remain unchanged. This has a defensive philosophy already, which involves checking if variables can be interpreted as different types, and the remainder are all represented as `CfDataVariables`, so we already have an existing fallback in place. Anything here that is not formatted correctly just shows up as extra `Cube`(s) in the loaded `CubeList`.

## More specifics on implementation

For reference when writing #6318 and #6319

- ✔ ~~Iris `__init__.py`~~:
  - Create the `LOAD_PROBLEMS` object
  - Example structure: `{"file/path/1": [(problem_object_1, error_or_stacktrace), (problem_object_1, error_or_stacktrace)]}`
- `helpers.py`:
[...]

PR #6350, opened 2025-02-28
reuploading changes for bounds of derived variables

HGWright:
### Description

I seem to have royally screwed up the old PR for this. So here are all of the relevant current changes for that.

Will copy over all of the discussions from the PR soon.

---

---

### DIFF
--- a/lib/iris/fileformats/cf.py
+++ b/lib/iris/fileformats/cf.py
@@ -16,6 +16,7 @@ References
 
 from abc import ABCMeta, abstractmethod
 from collections.abc import Iterable, MutableMapping
+import contextlib
 import os
 import re
 from typing import ClassVar, Optional
@@ -93,6 +94,8 @@ class CFVariable(metaclass=ABCMeta):
 
         #: CF-netCDF formula terms that his variable participates in.
         self.cf_terms_by_root = {}
+
+        self._to_be_promoted = False
 
         self.cf_attrs_reset()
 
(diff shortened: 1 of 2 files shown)


pp-mo:
Note (for my benefit!)
This basically replaces #6190
[This comment](https://github.com/SciTools/iris/pull/6190#issuecomment-2669304020) is a useful summary of the agreed behaviour we should be delivering

lib/iris/fileformats/cf.py
```diff
@@ -1539,19 +1600,42 @@ def _span_check(
...
+        if iris.FUTURE.derived_bounds:
+            for cf_var in self.cf_group.formula_terms.values():
+                if self.cf_group[cf_var.cf_name] is CFBoundaryVariable:
+                    continue
+                else:
+                    for cf_root, cf_term in cf_var.cf_terms_by_root.items():
+                        cf_root_var = self.cf_group[cf_root]
+                        if not hasattr(cf_root_var, "standard_name"):
+                            continue
+                        name = cf_root_var.standard_name or cf_root_var.long_name
```
stephenworsley, line 1612: if you reach this point you will have a standard name, so `or cf_root_var.long_name` is unnecessary.

lib/iris/fileformats/cf.py
```diff
@@ -1416,16 +1419,66 @@ def _translate(self, variables):
...
+            for cf_var in formula_terms.values():
+                # eg. eta:'a' | cf_root = eta and cf_term = a. cf_var.cf_terms_by_root = {'eta': 'a'} (looking at all appearances in formula terms)
+                for cf_root, cf_term in cf_var.cf_terms_by_root.items():
+                    # gets set to the bounds of the coord from cf_root_coord
+                    bounds_name = None
+                    # cf_root_coord = CFCoordinateVariable of the coordinate relating to the root
+                    cf_root_coord = self.cf_group.coordinates.get(cf_root)
+                    if cf_root_coord is None:
+                        cf_root_coord = self.cf_group.auxiliary_coordinates.get(cf_root)
+                    with contextlib.suppress(AttributeError):
```
stephenworsley, line 1433: I feel like this is probably not necessary. Would you not want to do something like `bounds_name = getattr(cf_root_coord, "bounds", default=None)`?

lib/iris/fileformats/cf.py
```diff
@@ -1416,16 +1419,66 @@ def _translate(self, variables):
...
+                        # Copes with cf_root_coord not existing, OR not having
+                        #  `bounds` attribute.
+                        bounds_name = cf_root_coord.bounds
+
+                    if bounds_name is not None:
+                        try:
```
stephenworsley, line 1439: Ultimately, I think you'll probably want to replace this try-except call. I expect you ought to be able to replace this with something like:

```python
bounds_vars = [...]
if len(bounds_vars) == 1:
  bounds_var = bounds_vars[0]
  ...
else:
  ...
```

trexfeathers:
A reminder, the work of #6317 is relying on certain behaviour of `cf.py`:

>Note that I have checked `cf.py` and believe it can remain unchanged. This has a defensive philosophy already, which involves checking if variables can be interpreted as different types, and the remainder are all represented as `CfDataVariables`, so we already have an existing fallback in place. Anything here that is not formatted correctly just shows up as extra `Cube`(s) in the loaded `CubeList`.

If you need to break the above 'contract' please make me aware as I'll need to re-scope my work. Last I checked @HGWright was hopeful we could keep this behaviour though 🤞

HGWright:
This is superseded by #6481. That makes all of the changes and more from this PR.

PR #6386, opened 2025-03-27
Run linkchecks regularly, instead of on PRs

pp-mo:
Replaces the nox-based, per-PR link check with an occasional scheduled (or manually triggered) linkcheck.

This is based on ["lychee"](https://github.com/lycheeverse/lychee/commit/e29a3c5b7845e6e3c69ad7dd62122b18537e4b82), which can check the RSTs directly

### Pros:
  * doesn't require the docs build
  * independent of Sphinx
  * really simple to configure
  * the resulting workflow should port unchanged to other repos (so ripe for templating)
  * already supports a manually triggered action, as well as time-triggered
  * produces a nicely formatting check summary, which also appears on the action
  * produces a repo Issue, including the summary, _only_ if there were any link failures

### Cons:
  * doesn't check our links written as Sphinx roles, i.e. `:pep:`, `:PR:`, and `:Issue:`, as seen (mostly) in whatsnews.
    * N.B. we were hitherto excluding most of these as skipiing all of `github.com`, but I have been able to re-include them by [limiting the ignore to `github.com/[^/]*$`](https://github.com/SciTools/iris/blob/1074909873d2b5ea4d4bc12cb8bcc0010c5b05e1/.lycheeignore#L17), as the damage seemed to be mostly done by all the user ids
 * doesn't provide source-line references for failing urls


.github/workflows/ci-linkchecks.yml
```diff
@@ -0,0 +1,29 @@
...
+      - name: Link Checker
+        id: lychee
+        uses: lycheeverse/lychee-action@v2
+        with:
+          fail: false
+          args: "--verbose './docs/**/*.rst' './docs/**/*.inc'"
+
+      - name: Create Issue From File
+        if: steps.lychee.outputs.exit_code != 0
+        uses: peter-evans/create-issue-from-file@v5
```
trexfeathers, line 25:
```suggestion
        uses: peter-evans/create-issue-from-file@e8ef132d6df98ed982188e460ebb3b5d4ef3a9cd
```

[Security hardening for GitHub Actions - GitHub Docs](https://docs.github.com/en/actions/security-for-github-actions/security-guides/security-hardening-for-github-actions#using-third-party-actions)

.github/workflows/ci-linkchecks.yml
```diff
@@ -0,0 +1,29 @@
+name: Linkcheck
+
+on:
+  workflow_dispatch:
+  schedule:
+    - cron: "00 06 * * 1-5"
```
trexfeathers, line 6: Do you think running at the weekend too would give valuable extra information about the persistence of some link problems?

>It's been failing for 3 mornings in a row, must be a real problem

.github/workflows/ci-linkchecks.yml
```diff
@@ -0,0 +1,29 @@
...
+  linkChecker:
+    runs-on: ubuntu-latest
+    permissions:
+      issues: write # required for peter-evans/create-issue-from-file
+    steps:
+      - uses: actions/checkout@v4
+
+      - name: Link Checker
+        id: lychee
+        uses: lycheeverse/lychee-action@v2
```
trexfeathers, line 18:
```suggestion
        uses: lycheeverse/lychee-action@f613c4a64e50d792e0b31ec34bbcbba12263c6a6
```

[Security hardening for GitHub Actions - GitHub Docs](https://docs.github.com/en/actions/security-for-github-actions/security-guides/security-hardening-for-github-actions#using-third-party-actions)

.lycheeignore
trexfeathers, line 1: Given the limited number of to-do comments, I'd be keen to remove/replace those links as part of this PR.

Either that or raise an issue for doing it in future.

.github/workflows/ci-linkchecks.yml
trexfeathers, line 1: Can you link me to a demonstration of this in action?

.lycheeignore
trexfeathers, line 1: It's a shame to introduce another config file into root (I know you're also not a fan of these), but I'm assuming you have investigated and there is no way to get this into `pyproject.toml` or similar?

.github/workflows/ci-linkchecks.yml
trexfeathers, line 1: Should we include a badge?

https://github.com/lycheeverse/lychee-action?tab=readme-ov-file#fancy-badge

.github/workflows/ci-linkchecks.yml
```diff
@@ -0,0 +1,29 @@
...
+          fail: false
+          args: "--verbose './docs/**/*.rst' './docs/**/*.inc'"
+
+      - name: Create Issue From File
+        if: steps.lychee.outputs.exit_code != 0
+        uses: peter-evans/create-issue-from-file@v5
+        with:
+          title: Link Checker Report
+          content-filepath: ./lychee/out.md
+          labels: report, automated issue
```
trexfeathers, line 29: I appreciate this is a faithful copy of [the example in their README](https://github.com/lycheeverse/lychee-action?tab=readme-ov-file#usage), but we at least ought to be using our own labels!

.github/workflows/ci-linkchecks.yml
```diff
@@ -0,0 +1,29 @@
...
+      issues: write # required for peter-evans/create-issue-from-file
+    steps:
+      - uses: actions/checkout@v4
+
+      - name: Link Checker
+        id: lychee
+        uses: lycheeverse/lychee-action@v2
+        with:
+          fail: false
+          args: "--verbose './docs/**/*.rst' './docs/**/*.inc'"
```
trexfeathers, line 21: ### Possible show-stopper

We are actually losing a decent amount of coverage here, because we can't check our docstrings anymore, and there are definitely some links in there.

trexfeathers (changes requested):
A few changes for you, and one possible show-stopper.

.lycheeignore
pp-mo: https://github.com/SciTools/iris/issues/6387

pp-mo: I thought they all needed a bit of individual thought

.github/workflows/ci-linkchecks.yml
pp-mo: Hmmm.  You have a point. I captured the output of the existing linkcheck, and from grepping I reckon there are 120 such links from within the API docs

.lycheeignore
pp-mo: OK I think it can be done, but it means putting **_all_** your config into a toml file.

.github/workflows/ci-linkchecks.yml
pp-mo: This issues shows a failure report : https://github.com/pp-mo/iris/issues/102

Also link to the action at the very end of that

pp-mo: In the README

pp-mo: But only if we can get it to reflect to OK-ness of the result

.github/workflows/ci-linkchecks.yml
pp-mo: OK we got this to work also by making it check the `./lib/**/*.py` files as well !
Needs to add some more exclusions (no surprise), but entirely feasible. :rocket:

.github/workflows/ci-linkchecks.yml
pp-mo: Maybe does no harm, but we might end up with multiple issues stacking up?

@trexfeathers @pp-mo Decided: let's just try it

.github/workflows/ci-linkchecks.yml
pp-mo: See https://github.com/SciTools/iris/pull/6386/commits/62cfabc0d7a3198737319651785dd07b501e876e

You can't see this running here until merged.
But examples are visible on my fork, e.g. https://github.com/pp-mo/iris/issues/110

.lycheeignore
pp-mo: As discussed, didn't do this. OK?

.github/workflows/ci-linkchecks.yml
pp-mo: See https://github.com/SciTools/iris/pull/6386/commits/1da5538cf126b4f3f386fdeb3c59fa97df6c89b5

.github/workflows/ci-linkchecks.yml
pp-mo: See https://github.com/SciTools/iris/pull/6386/commits/94f54f3efc3fe442412df31f8405291689ca80e7

Took a while to sort out the syntax, but I think this choice makes reasonable sense 
( though NB we don't usually apply more than one "Type: xxx" label ).  
You can see a test result [here](https://github.com/pp-mo/iris/issues/115).  
Of course, they don't look like Iris issues, but I presume they will just connect by text-name equality .. it's a bit weak.

pp-mo: **Update** oh cool, you can just hover over an issue link, and the report summary appears as a tooltip ! :sunglasses:

pp-mo: err ... not so cool ... it only shows so much, so you can't see that  it errored :facepalm:

pp-mo:
?hopefully https://github.com/trexfeathers/SciTools.github/pull/19 fixes the outstanding test fail ?

.lycheeignore
trexfeathers: Agreed offline that we prefer this implementation as the most 'vanilla' possible

.github/workflows/ci-linkchecks.yml
trexfeathers: Neat! Well done for going the extra mile.

trexfeathers (approved):

trexfeathers:
Super, thanks for this exciting improvement @pp-mo!

PR #6395, opened 2025-03-31
TST: make image comparison fail nicely

### LINKED ISSUE #6393: TST: "coordinates given" image comparison fails badly
## 📰 Custom Issue

`lib/iris/tests/test_plot.py` has two instances where it catches the image comparison failure and tries to raise a different error:

https://github.com/SciTools/iris/blob/69f775ed7f7fc2af74424d7786d2197c7dcdbb70/lib/iris/tests/test_plot.py#L807-L810

https://github.com/SciTools/iris/blob/69f775ed7f7fc2af74424d7786d2197c7dcdbb70/lib/iris/tests/test_plot.py#L819-L822

These rely on its `fail` method, which apparently does not exist.

```
    def run_tests_1d(self, cube, results):
        # there is a different calling convention for 1d plots
        for draw_method, rcoords in results:
            draw_method(cube.coord(rcoords[0]), cube)
            try:
                self.check_graphic()
            except AssertionError as err:
                msg = (
                    "Draw method {!r} failed with coords: {!r}. Assertion message: {!s}"
                )
>               self.fail(msg.format(draw_method, rcoords, err))
E               AttributeError: 'TestPlotCoordinatesGiven' object has no attribute 'fail'

lib/iris/tests/test_plot.py:822: AttributeError

```
I tripped over this because I tested against [Matplotlib's nightly build](https://matplotlib.org/stable/install/index.html#install-a-nightly-build).

rcomer:
### Description

Fixes #6393.  I assume this `fail` method existed to expand the error message but got removed somewhere along the way.  I just learned about the [add_note](https://docs.python.org/3/library/exceptions.html#BaseException.add_note) method, which seems a nice simple way to achieve this.

Failure messages now look like

`run_tests_1D`
```
E                           AssertionError: Image comparison failed: Bad phash a2ffb6127f0dc9992085d960c6748d369b121ca69d6a9b048df34ce789ff7205 with hamming distance 4 for test iris.tests.test_quickplot.TestQuickplotCoordinatesGiven.test_y.0.
E                           Draw method iris.quickplot.plot failed with coords: ['grid_latitude'].
```

`run_tests`
```
E                           AssertionError: Image comparison failed: Bad phash ead77a0cf751c5acb40195acd56999e2952899f2d5ec0bf2982c6a536a57b700 with hamming distance 6 for test iris.tests.test_quickplot.TestQuickplotCoordinatesGiven.test_tx.5.
E                           Draw method iris.quickplot.points failed with coords: ['grid_longitude', 'time'].
```

---

---

### DIFF
--- a/lib/iris/tests/test_plot.py
+++ b/lib/iris/tests/test_plot.py
@@ -804,10 +804,10 @@ class TestPlotCoordinatesGiven(_shared_utils.GraphicsTest):
             try:
                 self.check_graphic()
             except AssertionError as err:
-                self.fail(
-                    "Draw method %r failed with coords: %r. "
-                    "Assertion message: %s" % (draw_method, rcoords, err)
+                err.add_note(
+                    f"Draw method {draw_method!r} failed with coords: {rcoords!r}."
                 )
+                raise
 
     def run_tests_1d(self, cube, results):
         # there is a different calling convention for 1d plots
@@ -816,10 +816,10 @@ class TestPlotCoordinatesGiven(_shared_utils.GraphicsTest):
             try:
                 self.check_graphic()
             except AssertionError as err:
-                msg = (
-                    "Draw method {!r} failed with coords: {!r}. Assertion message: {!s}"
+                err.add_note(
+                    f"Draw method {draw_method!r} failed with coords: {rcoords!r}."
                 )
-                self.fail(msg.format(draw_method, rcoords, err))
+                raise
 
     def test_yx(self):
         test_cube = self.cube[0, 0, :, :]


rcomer:
Oh, it was a unittest thing https://docs.python.org/3/library/unittest.html#unittest.TestCase.fail

ESadek-MO:
Yeah, it seems like something that was missed when we converted to pytest. I think there's 3 other instances where `self.fail()` is used, I'll create a sepearte PR to add a TODO to them.

ESadek-MO (changes requested):
As it is, looks fine, nice simple change. 

Only question I have; do we think `raise ThisException("msg) from err` would achieve better? They seem quite similar, but offers a slightly different message. As I understand it, `add_note` just adds information, whereas `raise ... from ...` creates a new exception from the previous one.  

I've only looked into this today, so my opinion isn't strong, nor fully fleshed out. Happy to be overridden on this.

rcomer:
I have no strong opinion either.  I picked `add_note` because I only just learned about it so wanted an excuse to use it 😆 

I think if we do `raise from` we will still want it to be an `AssertionError`, and we will want the string about it being an image comparison failure to be included in the final message so you know up front it's an image test that went wrong.  So, turning it around, is there an advantage to `raise from`?

ESadek-MO (approved):
You've convinced me, this seems good to me!

PR #6401, opened 2025-04-03
Remove unnecessary `global` declarations

ukmo-ccbunney:
### Description

Remove several `global` variable declarations that were unnecessary due to the variables never being **assigned** to in their declared scope (accessing them is not ambiguous in the namespace scope).

This was highlighted by the flake8 linter in a recent pre-commit hook update.

Unblocks: #6396


ESadek-MO (approved):
Thanks @ukmo-ccbunney, LGTM!

PR #6405, opened 2025-04-09
Make MeshCoords immutable and sync updates with the attached mesh

ESadek-MO:
### Description
Closes https://github.com/SciTools/iris/issues/4757, continues #6125.

This pull request aims to make MeshCoords immutable, and when it's corresponding Mesh updates, for the MeshCoord to refresh its Points, Bounds, and Metadata from the Mesh.

<details>
<summary>Notes to self</summary>
<br/>

- First problem, where `node` had no bounds, has been solved.
- Next problem is that bounds aren't being rendered in the cml (and presumably, don't exist). There wasa thread to pull, in that `bounds` is set to an error message about `bounds_dm`, but it seems like that's a red herring, and is solved within `__init__`.
- There's a chance this only solves another 10 or so errors, but progress is progress
</details>


lib/iris/tests/unit/cube/test_Cube.py
```diff
@@ -2336,20 +2336,25 @@ def test_multi_meshcoords(self):
...
-        # constructor args, but we can adjust common metadata afterwards.
-        meshco_2.rename("junk_name")
-
-        n_faces = meshco_1.shape[0]
-        cube = Cube(
-            np.zeros(n_faces),
-            aux_coords_and_dims=[(meshco_1, 0), (meshco_2, 0)],
-        )
-        assert cube.mesh == meshco_1.mesh
+    # def test_multi_meshcoords_same_axis(self):
```
pp-mo: @ESadek-MO agreed we can lose this test, as this awkward possibility no longer exists.
Since meshcoord metadata is now taken from the mesh itself, I think we can't now get 2 different meshcoords on the same axis

lib/iris/tests/unit/cube/test_Cube.py
```diff
@@ -2422,17 +2427,17 @@ def test_add_compatible(self):
...
-        cube = self.cube
-        meshco_y = self.meshco_y
-        # Add the y-meshco back into the cube.
-        cube.add_aux_coord(meshco_y, 1)
-        # Make a duplicate y-meshco, renamed so it can add into the cube.
-        new_meshco_y = meshco_y.copy()
-        new_meshco_y.rename("alternative")
-        cube.add_aux_coord(new_meshco_y, 1)
-        assert len(cube.coords(mesh_coords=True)) == 3
+    # def test_add_multiple(self):
```
pp-mo: As [above](https://github.com/SciTools/iris/pull/6405#discussion_r2161785129), think this possibility is now obsolete + we can ditch this test.

lib/iris/tests/unit/mesh/components/test_MeshCoord.py
```diff
@@ -445,23 +429,35 @@ def test_cube_dims(self):
...
+    #     # "guessed axis" approach.  So the standard_name does not control it.
+    #     coord_on_mesh = meshcoord.mesh.coord(
+    #         location=meshcoord.location,
+    #         axis=meshcoord.axis
+    #     )
+    #     #doesn't like renaming to latitude.......
+    #     #any other one works
+    #     coord_on_mesh.standard_name = "grid_longitude"
+    #     self.assertIs(cube.coord(axis="x"), meshcoord)
+    #     self.assertEqual(cube.coords(axis="y"), [])
```
pp-mo: I think it's still worth testing the "find by axis" approach, 
but the second part showing "rename to something inappropriate + check it doesn't affect the stated axis" obviously won't work any more.  
Actually you probably _can_ still usefully demonstrate something like this, but not by renaming the MeshCoord -- you will have to adjust the coord in the underlying sample mesh instead.

lib/iris/coords.py
```diff
@@ -569,6 +572,17 @@ def reindent_data_string(text, n_indent):
 
         return "\n".join(output_lines)
 
+    def __setattr__(self, key, value):
+        if getattr(self, "_mesh_parents", None) is not None:
```
ESadek-MO: Is this line needed?

lib/iris/coords.py
```diff
@@ -569,6 +572,17 @@ def reindent_data_string(text, n_indent):
...
+            for parent in self._mesh_parents:
+                parent.timestamp = datetime.now()
+
+        prop = getattr(type(self), key, None)
+        if isinstance(prop, property):
+            prop.__set__(self, value)
+        else:
+            super.__setattr__(self, key, value)
+
```
ESadek-MO: __setattr__ should just work here, no?

lib/iris/cube.py
```diff
@@ -2785,7 +2785,7 @@ def shape(self) -> tuple[int, ...]:
 
     @property
     def dtype(self):
-        """The data type of the values in the data array of this :class:`~iris.cube.Cube`."""
+        """The dtype of the values in the data array of this :class:`~iris.cube.Cube`."""
```
ESadek-MO: Why?

lib/iris/mesh/components.py
```diff
@@ -840,6 +841,7 @@ def from_coords(cls, *coords):
         # Validate points and bounds shape match.
         def check_shape(array_name):
             attr_name = f"core_{array_name}"
+            coords
```
ESadek-MO: Why?

lib/iris/mesh/components.py
```diff
@@ -2928,6 +3055,10 @@ def copy(self, points=None, bounds=None):
         new_coord = MeshCoord(mesh=self.mesh, location=self.location, axis=self.axis)
         return new_coord
 
+    def rename(self, name: str | None) -> None:
+        """Ensure you that renaming MeshCoords is done via the Coord attached to the Mesh."""
+        self.mesh.coord(location=self.location, axis=self.axis).rename(name)
+
```
ESadek-MO: Delete this

ESadek-MO: I've made this error, instead of just deleting this.

lib/iris/coords.py
ESadek-MO: It is. things such as DimCoords would fail otherwise

lib/iris/coords.py
```diff
@@ -569,6 +572,17 @@ def reindent_data_string(text, n_indent):
...
 
+    def __setattr__(self, key, value):
+        if getattr(self, "_mesh_parents", None) is not None:
+            for parent in self._mesh_parents:
+                parent.timestamp = datetime.now()
+
+        # prop = getattr(type(self), key, None)
+        # if isinstance(prop, property):
+        #     prop.__set__(self, value)
+        # else:
```
stephenworsley: I take it this has been tested to be unnecessary, this commented out code should probably be removed

lib/iris/coords.py
stephenworsley: could this be replaced by `if hasattr(self, "_mesh_parents"):`?

lib/iris/mesh/components.py
```diff
@@ -1141,7 +1154,9 @@ def all_connectivities(self):
     @property
     def timestamp(self):
         """The time and date that the mesh coordinates and or connecitivities were last edited."""
-        return max(self._coord_manager.timestamp, self._connectivity_manager.timestamp)
+        return max(
+            self._coord_manager.timestamp._dt, self._connectivity_manager.timestamp._dt
+        )
```
stephenworsley, line 1161: This timestamp property behaves quite differently to those in the managers. I think it would clarify things to give them different names. Maybe this could be called something like `_last_updated`.

lib/iris/mesh/components.py
```diff
@@ -88,6 +89,19 @@
 )
 
 
+class _Timestamp:
+    """Makes timestamps mutable.
+
+    This behaviour is used to ensure garbage collection works as expected.
```
stephenworsley: While this is the purpose we're creating a class rather than storing a raw datetime, I think that's the secondary rather than the primary purpose of this class. The primary purpose is to store the last time the associated object has been updated. You could go for something more like

"""Stores the last time the timestamped object has been updated.

The timestamp for a parent object may be held and updated by child objects without the need for the child objects to contain a reference to the parent object. This ought to prevent large parent objects from being held unnecessarily in memory. 
"""

lib/iris/mesh/components.py
```diff
@@ -1137,6 +1151,13 @@ def all_connectivities(self):
         """All the :class:`~iris.mesh.Connectivity` instances of the :class:`MeshXY`."""
         return self._connectivity_manager.all_members
 
+    @property
+    def timestamp(self):
```
stephenworsley: This probably makes more sense as a private object

lib/iris/mesh/components.py
```diff
@@ -88,6 +89,19 @@
...
 
 
+class _Timestamp:
+    """Makes timestamps mutable.
+
+    This behaviour is used to ensure garbage collection works as expected.
+    """
+
+    def __init__(self):
+        self._dt = None
```
stephenworsley: shouldn't this be set to `datetime.now()` ?

lib/iris/mesh/components.py
```diff
@@ -2687,12 +2767,15 @@ class MeshCoord(AuxCoord):
...
     def __init__(
         self,
-        mesh: MeshXY,
-        location: Location,
-        axis: Literal["x", "y"],
+        mesh,
+        location,
+        axis,
     ):
+        self.timestamp = None
```
stephenworsley: This should probably be private. It might also benefit froma different name to identify that this has a different role to the other variables named `timestamp`. Maybe something like `_last_recorded_mesh_update`.

lib/iris/mesh/components.py
```diff
@@ -2928,6 +3071,10 @@ def copy(self, points=None, bounds=None):
         new_coord = MeshCoord(mesh=self.mesh, location=self.location, axis=self.axis)
         return new_coord
 
+    def rename(self, name: str | None) -> None:
+        """Ensure you that renaming MeshCoords is done via the Coord attached to the Mesh."""
```
stephenworsley: A more accurate description would be:

"""Raises an error.

Standard renaming is overriden to ensure that...

lib/iris/mesh/components.py
```diff
@@ -2928,6 +3071,10 @@ def copy(self, points=None, bounds=None):
...
+        raise ValueError("To rename a MeshCoord, you have to go via the attached mesh.")
```
stephenworsley: A more descriptive error message would be helpful, rather than "... go via the attached mesh." maybe "... rename the coordinate on the mesh from which this derives."

stephenworsley (changes requested):
Looking good! Here's some comments from the code itself, I'll go through the tests next. The timestamp stuff looks like it should do the job now, I think mostly I just have virable name and documentation quibbles.

lib/iris/mesh/components.py
ESadek-MO: This isn't to hold when the mesh was last updated, but rather when the MeshCoord was last updated. This is then compared against the mesh one. 

As such, I think this *should* have the same name as on the mesh, (I'm changing both to `_last_modified`), because then the line:

`if self._last_modified != ...mesh._last_modified:` is more clear. Is this okay?

stephenworsley: From what I can tell, this timestamp is set here: https://github.com/SciTools/iris/blob/a69e9678e83d23a533608c12b3c85585ce384b21/lib/iris/mesh/components.py#L3172
This suggests that this doesn't record the timestamp from when the MeshCoord updates from the Mesh, but from when the Mesh last updated to cause the MeshCoord to update.

stephenworsley: With that said, changing both to `_last_modified` is definitely an improvement, I think it's probably fine these having the same name.

lib/iris/mesh/components.py
ESadek-MO: I was indecisive on this. Should the mesh be considered "updated" when the timestamp is first initialised at the start of __init__? 

However, I think it's overwritten promptly anyway, so happy to follow your suggestion.

lib/iris/tests/unit/mesh/components/test_MeshCoord.py
```diff
@@ -751,20 +775,22 @@ def fetch_sources_from_mesh():
...
-        meshcoord.points
-        meshcoord.bounds
-        self.assertFalse(meshcoord.has_lazy_points())
-        self.assertFalse(meshcoord.has_lazy_bounds())
+        assert meshcoord.has_lazy_points() is True
+        assert meshcoord.has_lazy_bounds() is True
+        mc_points = meshcoord.points
+        mc_bounds = meshcoord.bounds
+        assert meshcoord.has_lazy_points() is True
+        assert meshcoord.has_lazy_bounds() is True
```
stephenworsley, line 786: Is it expected that these remain lazy? If so this is definitely worth mentioning in the whatsnew (which is worth adding)

ESadek-MO: Yeah, that's a good point. 

In discussion with @pp-mo, we thought the best contract was that it returns real data, but the points remain lazy. 

Otherwise, you're realizing the coord_on_mesh points as well, which you shouldn't be able to do accidentally. 

The whatsnew is gonna be a behemoth I fear, on it now. Ill try and pare it down

docs/src/whatsnew/latest.rst
```diff
@@ -30,7 +30,11 @@ This document explains the changes made to Iris for this release
 ✨ Features
 ===========
 
-#. N/A
+#. `@ESadek-MO` made MeshCoords immutable. :class:`iris.MeshCoord`s are now updated automatically when
+   changing the attached mesh. All changes from to the :class:`iris.MeshCoord` should instead be done to
```
stephenworsley: "... changes from to ..." remove the from

stephenworsley (changes requested):

lib/iris/mesh/components.py
stephenworsley: Yeah, I think it just makes it neater if _Timestamp always contains the same type of object.

lib/iris/tests/unit/mesh/components/test_MeshCoord.py
```diff
@@ -1083,3 +1083,17 @@ def test_standard_name(self):
...
+            np.zeros(
+                3,
+            ),
+            np.zeros((3, 3)),
+        )
+        self.coord_on_mesh.points = np.zeros(3)
+        # Only the first update should actually update
+        _ = self.meshcoord.update_from_mesh()
+        _ = self.meshcoord.update_from_mesh()
+        mocked.assert_called_once()
```
stephenworsley, line 1099: It's probably worth writing a similar test to check that it's called twice when expected.

lib/iris/mesh/components.py
```diff
@@ -2999,11 +3002,13 @@ def update_from_mesh(self):
...
+                self._last_modified < self.mesh._last_modified
             ):
                 points, bounds = self._load_points_and_bounds()
                 super(MeshCoord, self.__class__).points.fset(self, points)
                 super(MeshCoord, self.__class__).bounds.fset(self, bounds)
+                object.__setattr__(self, "_last_modified", self.mesh._last_modified)
+        # Ensure errors aren't bypassed
         except Exception as e:
             raise e
         finally:
```
stephenworsley, line 3014: might be worth adding a note to the finally call saying that it ensures _updating isn't left on after a failure

stephenworsley (changes requested):
Looking good, I believe these are the last things I can find to request changes on, after this it should be good to merge!

lib/iris/tests/unit/mesh/components/test_MeshCoord.py
ESadek-MO: It should never be called twice. If it's updated once, that means it works as expected. i.e. it's called the first time when timestamps are expected to be different, it's not called the second time when timestamps are expected to be aligned. 

I'd argue that it being called once asserts that it works when a change happens, and it being called *only* once asserts it doesn't call when a change hasn't happened.

lib/iris/mesh/components.py
ESadek-MO: I think that's redundant? That's generally what finally does, I think that would be over explaining perhaps?

lib/iris/tests/unit/mesh/components/test_MeshCoord.py
stephenworsley: what I mean is that it is actually called a second time after timestamps change. I guess the other tests do sort of cover this by showing the effect of this function, but this would explicitly test that the logic for timestamps works both ways. I suppose this introduces some redundancy to the tests, but I don't think that's too bad a thing for testing and it would introduce a nice symmetry to this part of the tests.

lib/iris/mesh/components.py
stephenworsley: I suppose there's still a case for explaining _why_ it's so important that `_updating` must never be left `True` outside of this context, i.e. that this would break the link between a MeshCoord and it's Mesh and mean that it could no longer be updated.

lib/iris/tests/unit/mesh/components/test_MeshCoord.py
```diff
@@ -1097,3 +1097,9 @@ def test_updates_only_once(self, mocker):
         _ = self.meshcoord.update_from_mesh()
         _ = self.meshcoord.update_from_mesh()
         mocked.assert_called_once()
+
+        # Ensure it updates more than once if the mesh has been updated
+        # a second time
+        self.coord_on_mesh.points = np.ones(3)
+        _ = self.meshcoord.update_from_mesh()
+        mocked.assert_has_calls([mock.call(), mock.call()])
```
stephenworsley: It might be neater to write as `assert mocked.call_count == 2` (see https://docs.python.org/3/library/unittest.mock.html#unittest.mock.Mock.call_count)

stephenworsley (changes requested):

pp-mo:
Thanks @ESadek-MO @stephenworsley 
Just to confirm, I'm looking at this now, looks very  close but I need a bit longer to think ....
Hopefully I will just merge later today.

lib/iris/tests/unit/mesh/components/test_MeshCoord.py
pp-mo: Now fixed :+1:

docs/src/whatsnew/latest.rst
pp-mo: Now done

lib/iris/tests/unit/mesh/components/test_MeshCoord.py
pp-mo: Now done :+1:

lib/iris/tests/unit/mesh/components/test_MeshCoord.py
pp-mo: Ok I think this has now been addressed 👍

pp-mo (approved):
Now looking all ok

PR #6421, opened 2025-04-25
Update Iris benchmarking to align with templating

trexfeathers:
### Description

Now we've create templates for benchmarking files (SciTools/.github#119), there are slight deviations between Iris and the templates. This PR removes those deviations.

---

---


benchmarks/_asv_delegated_abc.py
trexfeathers, line 1: Recommend using a manual diff for the two `delegated` Python modules - renaming has confused Git.

- `_asv_delegated_abc.py` used to be `asv_delegated.py`
- `asv_delegated.py` used to be `asv_delegated_iris.py`

benchmarks/bm_runner.py
```diff
@@ -85,9 +86,8 @@ def _prep_data_gen_env() -> None:
         )
         # Find the environment built above, set it to be the data generation
         #  environment.
-        data_gen_python = next(
-            (ROOT_DIR / ".nox").rglob(f"tests*/bin/python{python_version}")
-        ).resolve()
```
pp-mo, line 90: There's a bit here I don't understand...
Previously the end part of this path construction had the equivalent of .. 
`data_gen_python = (env_directory / "bin" / "python{python_version}").resolve()`
But now we have lost the `{python_version}` part of this.
Does that still work + if so why ??

pp-mo (changes requested):
One small query, otherwise OK

benchmarks/bm_runner.py
trexfeathers: It was unnecessary, since all environments include various symlinks in their `bin/` directory (using Python 3.13 as the example):

- `python` -> `python3.13`
- `python3` -> `python3.13`
- `python3.1` -> `python3.13`
- `python3.13`

Removing this unnecessary element hopefully makes the file less complex in future templating.

pp-mo (approved):
:+1:

PR #6422, opened 2025-04-25
Realign with re-styled template.

pp-mo:
MANIFEST.in should now compare nicely with [the latest template](https://github.com/SciTools/.github/blob/7363ee13d94e662d97d5a4e9f8166a09e5ccff88/templates/MANIFEST.in)


HGWright (approved):
Thanks @pp-mo. I was too quick on the trigger last time but this is good thanks.

Issue #6437, opened 2025-05-01
Consider switching from `array.compute()` to `dask.compute(array)`


trexfeathers:
The conversation on dask/dask#11921, including how relaxed Dask are about it, suggests that they expect 'correct' Dask usages to use the `dask.compute()` function, rather than the `.compute()` method of Dask objects. It makes me wonder whether the method approach makes us vulnerable to more breaking changes in future (similar to #6417).

I tried making the switch this afternoon, but it produced a lot of test failures.

- Small value changes, with @ukmo-ccbunney suggests might mean a different order of operations
- Integer results contained in length-1 tuples
- Other failures I didn't investigate

This is not affecting us right now, but I wanted to 'plant a flag' in case we get repeats of #6417 happen in future.

PR #6438, opened 2025-05-01
Lockfile updates plus adapt to Dask changes to delayed and internals

### LINKED ISSUE #6417: issue with iris saver with delayed and dask==2025.4.0
Hi folks, apologies I had to renounce the issue template, this is more of a heads up that may need further investigation (it most probably is a Dask issue, as per usual): see https://github.com/ESMValGroup/ESMValCore/issues/2716 and our [failed test](https://github.com/ESMValGroup/ESMValCore/actions/runs/14607786539/job/40980114812) as of last night with the brand new dask==2025.4.0

```
/opt/conda/envs/esmvaltool/lib/python3.13/site-packages/iris/io/__init__.py:476: in save
    result = saver(source, target, **kwargs)
/opt/conda/envs/esmvaltool/lib/python3.13/site-packages/iris/fileformats/netcdf/saver.py:2746: in save
    with Saver(filename, netcdf_format, compute=compute) as sman:
/opt/conda/envs/esmvaltool/lib/python3.13/site-packages/iris/fileformats/netcdf/saver.py:437: in __exit__
    self.complete()

self = <iris.fileformats.netcdf.saver.Saver object at 0x7f7a5b668470>

    def complete(self) -> None:
        """Complete file by computing any delayed variable saves.
    
        This requires that the Saver has closed the dataset (exited its context).
    
        """
        if self._dataset.isopen():
            msg = (
                "Cannot call Saver.complete() until its dataset is closed, "
                "i.e. the saver's context has exited."
            )
            raise ValueError(msg)
    
        # Complete the saves now
>       self.delayed_completion().compute()
E       AttributeError: 'tuple' object has no attribute 'compute'
```

Cheers 🍺

trexfeathers:
### Description

- Closes #6417 
- Closes #6412 
- Xref: #6437

---

---


lib/iris/tests/unit/fileformats/pp/test__interpret_field.py
```diff
@@ -91,43 +90,6 @@ def test_deferred_fix_lbrow_lbnpt(self):
         self.assertEqual(f1.lbrow, 3)
         self.assertEqual(f1.lbnpt, 4)
 
-    @tests.skip_data
-    def test_landsea_unpacking_uses_dask(self):
```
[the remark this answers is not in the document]
trexfeathers: Update: the change is from dask/dask#11736

I believe the information we needed to prove that `lazy_mask_array` and `lazy_soildata_array` are appropriately linked is no longer available due to the more sophisticated HLG wrapping.

So we will probably need to find a different way of proving the absence of nested computes, assuming it is even possible to do so.

lib/iris/tests/unit/fileformats/pp/test__interpret_field.py
```diff
@@ -112,21 +112,32 @@ def test_landsea_unpacking_uses_dask(self):
...
+        layer.mapping[layer_key] = mock_proxy
+
+        self.assertFalse(mock_proxy.__getitem__.called)
+        with contextlib.suppress(TypeError):
+            # Expect this to crash - data realisation is complex, and it is not
+            #  worth mocking this fully.
+            _ = lazy_soildata_array.compute()
+        # Proves that the soil data graph includes the mask graph - computing
+        #  the soil data has accessed the mask graph object which we mocked.
+        self.assertTrue(mock_proxy.__getitem__.called)
```
trexfeathers, line 140: ([see previous thread](https://github.com/SciTools/iris/pull/6438#discussion_r2071572251))

This change in approach is more agnostic to Dask's internals (although still somewhat vulnerable). It successfully runs, only failing in a classic test failure rather than falling over.

But all I have managed to do is prove that Dask `>=2025.4.0` breaks this test on a philosophical level - not just because the structures have changed.

### Results of updated test

| `fileformats/pp.py` | Dask version | Pass/Fail | what this proves |
| --- | --- | --- | --- |
| `v2.2.0` | `2025.3.0` | Fail ❌ |  |
| #3255 | `2025.3.0` | Pass ✔ | The updated test still detects the case it was designed for |
| `main` | `2025.3.0` | Pass ✔ |  |
| `main` | `2025.4.0` | Fail ❌ | The Dask graph for the data payload no longer incorporates the mask graph |

Understanding if this is cause for concern, and if so how to fix things, will be complex. The development team are fully booked on other projects for the foreseeable, so I would suggest a patch release with a `dask < 2025.4` pin. @SciTools/iris-devs what are your thoughts?

trexfeathers: Discussed at length with @pp-mo, who explained that this test was not written to demonstrate the other changes in #3255, but because there was a concern about something else similar happening again.

Our increased experience with Dask in the 6 years since has given us the confidence that:

1. We now write code that is much less vulnerable to things like #3237
2. Dask itself has matured, and is less vulnerable to unexpected corner cases

It looks like the refactoring that made this test fail (dask/dask#11736) was technical debt work to make Dask safer and more robust, and is NOT evidence that we have become vulnerable to nested `compute()` calls.

Based on all of this, we have decided to remove the test.

trexfeathers:
The performance shifts are most pronounced with smaller data sizes, becoming smaller as the data gets larger. This indicates increased overhead in setting up objects - perhaps dask/dask#11736 - but not something any part of the code that scales with data size. The impact will therefore remain in the milliseconds, even for very large data payloads, and is not a concern (agreed with @pp-mo).

trexfeathers:
Note I am now targetting the changes for a patch release (#6451), which we can then merge back that should unblock future lock-file PRs

trexfeathers:
Closed by #6412

PR #6441, opened 2025-05-02
Temporary Nox negation pin - see wntrblm/nox#961

trexfeathers:
### Description

The new PRs raised on the Nox GitHub suggest that this will be fixed in the next release of Nox.

The correct thing to do would be to update our dependencies and our lock files. Unfortunately we are half way through fixing a separate lock file problem (#6438) so I felt the pragmatic thing to do was only unblock the failing CI and nothing else.

---

---

### DIFF
--- a/.github/workflows/benchmarks_run.yml
+++ b/.github/workflows/benchmarks_run.yml
@@ -78,7 +78,7 @@ jobs:
 
       - name: Install Nox
         run: |
-          pip install nox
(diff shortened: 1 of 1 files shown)


trexfeathers:
Closes #6439

ESadek-MO (approved):
LGTM! I think this is the right approach, we can sort lockfiles later if needs be.

PR #6451, opened 2025-05-09
Compatibility with Dask `2025.4`

trexfeathers:
This has already been discussed on #6438, but I wanted to have a PR targetting `v3.12.x` as this makes the Git history less confusing.


bjlittle (approved):
@trexfeathers Awesome, thanks!

LGTM 👍

PR #6452, opened 2025-05-09
Whats new updates for v3.12.2

trexfeathers:


docs/src/whatsnew/3.12.rst
```diff
@@ -11,7 +11,6 @@ This document explains the changes made to Iris for this release
    :color: primary
    :icon: info
    :animate: fade-in
-   :open:
```
trexfeathers, line 14: It was getting a bit crowded so I closed the other two drop-downs. If we think that's a bad idea then I'm happy to change it back.

bjlittle (approved):

PR #6455, opened 2025-05-09
Fix linkcheck failures

trexfeathers:
### Description

Not sure how to test this without merging. Maybe we just give it a go?

- #6442 
- #6443 
- #6444 
- #6446 
- #6447 
- #6449 
- #6450
- #6454

---
[Consult Iris pull request check list]( https://scitools-iris.readthedocs.io/en/latest/developers_guide/contributing_pull_request_checklist.html)

---
Add any of the below labels to trigger actions on this PR:

- https://github.com/SciTools/iris/labels/benchmark_this

PR #6460, opened 2025-05-12
Various linkcheck fixes and improvements

trexfeathers:
### Description

#6455 did not fix the GitHub rate limiting failures, and in the process made us more vulnerable by widening access to the `scitools-ci` token. I've investigated in more detail now; see comments below for explanations.

---

---

### DIFF
--- a/.github/workflows/ci-linkchecks.yml
+++ b/.github/workflows/ci-linkchecks.yml
@@ -19,7 +19,7 @@ jobs:
         with:
           token: ${{secrets.GITHUB_TOKEN}}
           fail: false
-          args: "--verbose './docs/**/*.rst' './docs/**/*.inc' './lib/**/*.py'"
+          args: "--verbose --max-concurrency 1 './docs/**/*.rst' './docs/**/*.inc' './lib/**/*.py'"
 
       - name: Create Issue From File
         if: steps.lychee.outputs.exit_code != 0
--- a/.github/workflows/ci-linkchecks.yml
+++ b/.github/workflows/ci-linkchecks.yml
@@ -13,18 +13,11 @@ jobs:
     steps:
       - uses: actions/checkout@v4
 
-      - name: "Generate token"
-        uses: actions/create-github-app-token@v2
-        id: generate-token
-        with:
-          app-id: ${{ secrets.AUTH_APP_ID }}
-          private-key: ${{ secrets.AUTH_APP_PRIVATE_KEY }}
-
       - name: Link Checker
         id: lychee
         uses: lycheeverse/lychee-action@1d97d84f0bc547f7b25f4c2170d87d810dc2fb2c
         with:
-          token: ${{ steps.generate-token.outputs.token }}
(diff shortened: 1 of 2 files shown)


.github/workflows/ci-linkchecks.yml
```diff
@@ -13,20 +13,13 @@ jobs:
...
-
       - name: Link Checker
         id: lychee
         uses: lycheeverse/lychee-action@1d97d84f0bc547f7b25f4c2170d87d810dc2fb2c
         with:
-          token: ${{ steps.generate-token.outputs.token }}
+          token: ${{secrets.GITHUB_TOKEN}}
           fail: false
-          args: "--verbose './docs/**/*.rst' './docs/**/*.inc' './lib/**/*.py'"
+          args: "--verbose --max-concurrency 1 './docs/**/*.rst' './docs/**/*.inc' './lib/**/*.py'"
```
trexfeathers, line 22: We're in no hurry since this action is running overnight.

Preventing concurrency (which defaulted to 128) should help reduce the chance of rate limiting, either from GitHub or other providers.

.github/workflows/ci-linkchecks.yml
```diff
@@ -13,20 +13,13 @@ jobs:
...
-        id: generate-token
-        with:
-          app-id: ${{ secrets.AUTH_APP_ID }}
-          private-key: ${{ secrets.AUTH_APP_PRIVATE_KEY }}
-
       - name: Link Checker
         id: lychee
         uses: lycheeverse/lychee-action@1d97d84f0bc547f7b25f4c2170d87d810dc2fb2c
         with:
-          token: ${{ steps.generate-token.outputs.token }}
```
trexfeathers, line 27: This was a slightly desperate attempt to sledgehammer the problems out of the way - the `scitools-ci` app has highly elevated permissions!

I have since read the documentation and it seems that ANY token ought to be enough to give us access to the higher tier of allowed GitHub requests.

https://lychee.cli.rs/troubleshooting/rate-limits/#github-rate-limiting

.lycheeignore
```diff
@@ -26,6 +26,10 @@ https://geoport.whoi.edu:80/thredds/
 # catch (at least) github userids, of which many in whatsnews, too many --> "too many requests" failures
 https://github.com/[^/]*$
 
+# GitHub seems especially keen to rate-limit "blob" URLs.
+#  (maybe because these are especially associated with abuse?)
+https://github.com/[^/]*/[^/]*/blob/.*
+
```
trexfeathers, line 32: Afraid I can't see any other way around this.

You can see that all the problems in #6459 are rate limits from GitHub, and specifically associated with "blob" URLs - the many other GitHub URLs are all fine.

Slowing lychee right down - no concurrency, 1 thread, slow retries - reduced the number of network errors but never fully eliminated them when I tested locally, even when using the super-powered `scitools-ci` token.

bjlittle (approved):
@trexfeathers Seems reasonable to me.

Let's give it a whirl and see what happens 👌

Issue #3584, opened 2019-12-06
TypeError when merging masked scalar coord


trexfeathers:
The handling of scalar coordinates in _merge.py includes the [build_indexes](https://github.com/SciTools/iris/blob/242a62f719213eb315074621ccb8d00e51c466cf/lib/iris/_merge.py#L529) step. This cannot handle the scalar point being a masked value since it attempts to look the value up in the dictionary `name_index_by_scalar`, resulting in `TypeError: unhashable type: 'MaskedConstant'`.

This was discovered when attempting to load a NetCDF file that had previously been converted from NIMROD to NetCDF format. NetCDF had recognized a fill value for a scalar coordinate and therefore saved it as masked. [load_cube](https://github.com/SciTools/iris/blob/242a62f719213eb315074621ccb8d00e51c466cf/lib/iris/__init__.py#L307) includes a merge step, resulting in the error. This is demonstrated in the code below.

The presence of a masked point for a scalar coordinate is a possible scenario and should therefore be handled rather than producing an error.

```Python
import numpy as np
import iris
from iris import coords

my_cube = iris.cube.Cube(data=[1, 2])

# Generate a single-point array with a value NetCDF will recognise as a fill
# value and convert to a mask.
# Use the array as a scalar coordinate on my_cube.
my_array = np.array([-32767], dtype=np.int16)
my_scalar_coord = coords.DimCoord(points=my_array, standard_name="realization")
my_cube.add_aux_coord(my_scalar_coord)

# Save my_cube in NetCDF format.
iris.save(my_cube, "masked_scalar.nc")
# Attempt to load back the saved file. Results in the TypeError.
iris.load_cube("masked_scalar.nc")
```

schlunma:
I just discovered a very similar problem: if the input cubes for merging have a scalar coordinate that uses a masked array, the scalar coordinate of the merged cube is not a masked array anymore.

MWE:

```py
import numpy as np
import iris
import iris.coords
import iris.cube

# Setup cubes
scalar_coord = iris.coords.AuxCoord(np.ma.masked_array([0]), var_name='x')
merge_coord_1 = iris.coords.AuxCoord([1], var_name='merge_coord')
merge_coord_2 = iris.coords.AuxCoord([2], var_name='merge_coord')
cube_1 = iris.cube.Cube(1, var_name='cube',
                        aux_coords_and_dims=[(scalar_coord, ()), (merge_coord_1, ())])
[...]

PR #6468, opened 2025-05-15
Allow masked data in coords

ukmo-ccbunney:
## 🚀 Pull Request
Closes: #3584 

### Description
Allows masked data to be passed in to a `Cell`, which by extension allows for masked arrays in an `AuxCoord`.

---
---

### DIFF
--- a/lib/iris/_merge.py
+++ b/lib/iris/_merge.py
@@ -1406,10 +1406,20 @@ class ProtoCube:
                     # TODO: Consider appropriate sort order (ascending,
                     # descending) i.e. use CF positive attribute.
                     cells = sorted(indexes[name])
-                    points = np.array(
-                        [cell.point for cell in cells],
-                        dtype=metadata[name].points_dtype,
-                    )
+                    points = [cell.point for cell in cells]
+
+                    # If any points are masked then create a masked array type,
+                    # otherwise create a standard ndarray.
+                    if np.ma.masked in points:
+                        # Need to explicitly specify fill_type for list with `masked` elements to
+                        # avoid numpy raising warning about "converting masked element to NaN":
+                        dtype = metadata[name].points_dtype
+                        fill_value = np.ma.default_fill_value(dtype)
+                        points = np.ma.masked_array(
+                            points, dtype=dtype, fill_value=fill_value
+                        )
+                    else:
+                        points = np.array(points, dtype=metadata[name].points_dtype)
                     if cells[0].bound is not None:
                         bounds = np.array(
                             [cell.bound for cell in cells],
@@ -1613,22 +1623,34 @@ class ProtoCube:
             # Attempt to build a DimCoord and add it to the cube. If this
             # fails e.g it's non-monontic or multi-dimensional or non-numeric,
             # then build an AuxCoord.
-            try:
-                coord = iris.coords.DimCoord(
-                    template.points, bounds=template.bounds, **template.kwargs
-                )
-                if len(template.dims) == 1 and template.dims[0] not in covered_dims:
-                    dim_coords_and_dims.append(_CoordAndDims(coord, template.dims))
-                    covered_dims.append(template.dims[0])
-                else:
-                    aux_coords_and_dims.append(_CoordAndDims(coord, template.dims))
-            except ValueError:
+
+            # Check here whether points are masked? If so then it has to be an AuxCoord
+            if np.ma.is_masked(template.points):
+                print("[CB] Masked data - forcing AuxCoord")
+                # Masked data can only ever be an AuxDim
                 # kwarg not applicable to AuxCoord.
                 template.kwargs.pop("circular", None)
                 coord = iris.coords.AuxCoord(
                     template.points, bounds=template.bounds, **template.kwargs
                 )
                 aux_coords_and_dims.append(_CoordAndDims(coord, template.dims))
+            else:
+                try:
+                    coord = iris.coords.DimCoord(
+                        template.points, bounds=template.bounds, **template.kwargs
+                    )
+                    if len(template.dims) == 1 and template.dims[0] not in covered_dims:
+                        dim_coords_and_dims.append(_CoordAndDims(coord, template.dims))
+                        covered_dims.append(template.dims[0])
+                    else:
+                        aux_coords_and_dims.append(_CoordAndDims(coord, template.dims))
+                except (ValueError, TypeError):
+                    # kwarg not applicable to AuxCoord.
+                    template.kwargs.pop("circular", None)
+                    coord = iris.coords.AuxCoord(
+                        template.points, bounds=template.bounds, **template.kwargs
+                    )
+                    aux_coords_and_dims.append(_CoordAndDims(coord, template.dims))
 
         # Mix in the vector coordinates.
         for item, dims in zip(
--- a/lib/iris/coords.py
+++ b/lib/iris/coords.py
@@ -1234,6 +1234,9 @@ class Cell(namedtuple("Cell", ["point", "bound"])):
     # Make this class's comparison operators override those of numpy
     __array_priority__ = 100
 
+    # pre-computed hash for un-hashable `np.ma.masked` value
+    _MASKED_VALUE_HASH = hash("<<##MASKED_VALUE##>>")
+
     def __new__(cls, point=None, bound=None):
         """Construct a Cell from point or point-and-bound information."""
         if point is None:
@@ -1274,13 +1277,17 @@ class Cell(namedtuple("Cell", ["point", "bound"])):
 
     def __hash__(self):
         # See __eq__ for the definition of when two cells are equal.
+        point = self.point
+        if np.ma.is_masked(point):
+            # `np.ma.masked` is unhashable
+            point = Cell._MASKED_VALUE_HASH
         if self.bound is None:
-            return hash(self.point)
+            return hash(point)
         bound = self.bound
         rbound = bound[::-1]
         if rbound < bound:
             bound = rbound
-        return hash((self.point, bound))
+        return hash((point, bound))
 
     def __eq__(self, other):
         """Compare Cell equality depending on the type of the object to be compared."""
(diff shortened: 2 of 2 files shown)


ukmo-ccbunney:
There is only one small snag now w.r.t. the `dtype` returned when working with `Cells`:

```python
# Scalar coord points now have correct dtype
coord = cube.coord('realization')
coord.points
>> masked_array(data=[--],
>>           mask=[ True],
>>      fill_value=np.int64(999999),
>>            dtype=int16)

# ... but accessing them via the `cell` method still looses it:
coord.cell(0).point
>> masked

coord.cell(0).point.dtype
>> dtype('float64')   # <-- This is the dtype of np.ma.masked, not the coord

File: lib/iris/coords.py [hunk 1]
```diff
@@ -2083,7 +2090,8 @@ def cell(self, index):
         """
         index = iris.util._build_full_slice_given_keys(index, self.ndim)
 
-        point = tuple(np.array(self.core_points()[index], ndmin=1).flatten())
+        # Use `np.asanyaray` to preserve any masked values:
+        point = tuple(np.asanyarray(self.core_points()[index]).flatten())
```
trexfeathers, line 2094: Can confirm that:

- `asanyarray()` preserves masked arrays
- While `asanyarray()` does not have the `ndmin` parameter, the final `flatten()` call prevents a scalar array anyway

👍

lib/iris/_merge.py
```diff
@@ -1406,10 +1406,35 @@ def axis_and_name(name):
...
+
+                            # TODO: Perhaps use this method for all arrays, not just integer typea?
+                            # This will avoid numpy warnings about converting masked values to NaN
+                            # and also remove the need for this try..except block.
+                            arr_points = np.ma.masked_all(len(points), dtype=dtype)
+
+                            # slow! :(
+                            for i, p in enumerate(points):
+                                if p is not np.ma.masked:
+                                    arr_points[i] = p
```
trexfeathers: Surely this can be vectorised?

lib/iris/_merge.py
```diff
@@ -1406,10 +1406,35 @@ def axis_and_name(name):
...
+
+                    # If any points are masked then create a masked array type,
+                    # otherwise create a standard ndarray.
+                    if np.ma.masked in points:
+                        dtype = metadata[name].points_dtype
+
+                        try:
+                            # This will always raise a numpy warning about "converting masked elements to Nan"
+                            # TODO: Can we pre-fill masked data in `points`? Would require
+                            #       generation of separate mask array...
```
trexfeathers: Still in the works?

lib/iris/_merge.py
```diff
@@ -1406,10 +1406,35 @@ def axis_and_name(name):
...
+                            points = np.ma.masked_array(points, dtype=dtype)
+                        except np.ma.MaskError:
+                            # Fails for integer types as cannot convert a np.ma.masked to int type.
+                            # Need to loop over list and add points manually to a pre-masked array:
+
+                            # TODO: Perhaps use this method for all arrays, not just integer typea?
+                            # This will avoid numpy warnings about converting masked values to NaN
+                            # and also remove the need for this try..except block.
```
trexfeathers: Still in the works?

lib/iris/_merge.py
```diff
@@ -1613,22 +1644,32 @@ def _build_coordinates(self):
...
+                try:
+                    coord = iris.coords.DimCoord(
+                        template.points, bounds=template.bounds, **template.kwargs
+                    )
+                    if len(template.dims) == 1 and template.dims[0] not in covered_dims:
+                        dim_coords_and_dims.append(_CoordAndDims(coord, template.dims))
+                        covered_dims.append(template.dims[0])
+                    else:
+                        aux_coords_and_dims.append(_CoordAndDims(coord, template.dims))
+                except (ValueError, TypeError):
```
trexfeathers, line 1678: Can you remind me why we're now checking for `TypeError` as well?

lib/iris/_merge.py
```diff
@@ -1613,22 +1644,32 @@ def _build_coordinates(self):
...
+                    # kwarg not applicable to AuxCoord.
+                    template.kwargs.pop("circular", None)
+                    coord = iris.coords.AuxCoord(
+                        template.points, bounds=template.bounds, **template.kwargs
+                    )
+                    aux_coords_and_dims.append(_CoordAndDims(coord, template.dims))
```
trexfeathers, line 1684: I would prefer to see this in a little inner function, rather than copied twice.

trexfeathers (changes requested):
Thanks for looking into this @ukmo-ccbunney!

lib/iris/_merge.py
```diff
@@ -1406,10 +1406,47 @@ def axis_and_name(name):
...
+                        # This avoids the following problems when trying to do `np.ma.masked_array(points, dtype...)`:
+                        #   - Can't convert a np.ma.masked to an integer type
+                        #   - For floating point arrays, numpy raises a warning about "converting masked elements to NaN"
+                        dtype = metadata[name].points_dtype
+                        arr_points = np.ma.masked_all(len(points), dtype=dtype)
+                        # Is the below potentially slow? Perhaps no slower than the list comprehension used to
+                        # build the `points`` list above though...?
+                        for i, p in enumerate(points):
+                            if p is not np.ma.masked:
+                                arr_points[i] = p
```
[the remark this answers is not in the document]
trexfeathers: (I tried to make `points` itself into an object-type array, but that played havoc with various other operations due to it containing the masked constant).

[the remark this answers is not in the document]
trexfeathers: Update: `select` doesn't need to be an array. It's fine as a list

lib/iris/coords.py
ukmo-ccbunney: Indeed. I did leave a comment to state this originally, but have since cleaned it up. Good that you've confirmed though!

lib/iris/_merge.py
ukmo-ccbunney: Recent commit 5c9e553 uses it for both cases rather than the try/except block.

lib/iris/_merge.py
ukmo-ccbunney: `TypeError` was originally catching when a DimCoord failed to build due to masked data in the coord points.
Now I am checking for the existence of masked data explicitly and building an AuxCoord instead (and not trying to build a DimCoord). I have removed now (assuming you are happy with this approach).
See 2df6d050a

lib/iris/_merge.py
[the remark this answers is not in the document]
ukmo-ccbunney: Here's my take on this: 9d599db00
Note I'm using a default fill value now (truncated accordingly), rather than 999 as we discussed.
Also, to avoid numpy warnings about "converting masked value to nan", I need to filter the `points` array of masked values before adding to `arr_points`.

lib/iris/_merge.py
ukmo-ccbunney: See 9d599db00

lib/iris/tests/unit/coords/test_Cell.py
```diff
@@ -267,8 +264,34 @@ def test_cell_rhs(self):
...
+        cell = Cell(point=point, bound=None)
+        hash(cell)
+
+        if isinstance(point, np.number):
+            cell = Cell(point=input, bound=((point - 1, point + 1)))
 
 
 if __name__ == "__main__":
-    tests.main()
+    pytest.main()
```
trexfeathers, line 297: [Converting From unittest to pytest — Iris 3.12.2 documentation](https://scitools-iris.readthedocs.io/en/stable/developers_guide/contributing_pytest_conversions.html)

>11. Check for ``if __name__ == 'main'``. This is no longer needed with pytest.

lib/iris/tests/unit/coords/test_Cell.py
```diff
@@ -267,8 +264,34 @@ def test_cell_rhs(self):
...
+            pytest.param(datetime.datetime(2001, 1, 1), id="datetime"),
+        ),
+    )
+    def test_cell_is_hashable(self, point):
+        """Test a Cell object is hashable with various point/bound types."""
+        cell = Cell(point=point, bound=None)
+        hash(cell)
+
+        if isinstance(point, np.number):
+            cell = Cell(point=input, bound=((point - 1, point + 1)))
```
trexfeathers, line 293: This is either placed in the wrong order, or you intended to have a second `hash(cell)` call afterwards?

trexfeathers (changes requested):

lib/iris/tests/test_merge.py
```diff
@@ -1098,5 +1099,48 @@ def test_splitattrs_keys_local_match_masks_global_mismatch(self):
...
+        cubes = iris.cube.CubeList(
+            [
+                self._build_cube(np.ma.masked_array(i, dtype=dtype, mask=i % 2))
+                for i in range(n)
+            ]
+        )
+        merged = cubes.merge_cube()
+        c = merged.coord("realization")
+        assert np.ma.isMaskedArray(c.points)
+        assert np.any(c.points.mask) and not np.all(c.points.mask)
```
trexfeathers, line 1141: There is enough information to assert that the original mask has been preserved.

```suggestion
        assert all([c.points.mask[i] == i % 2 for i in range(len(c.points))])
```

... or some less-ugly equivalent

lib/iris/tests/unit/coords/test_Cell.py
ukmo-ccbunney: Fixed in ec157e2c5

lib/iris/tests/unit/coords/test_Cell.py
ukmo-ccbunney: Oops, yes - missing a second `hash` test.
Fixed (and commented) in cd1a7a83f

lib/iris/tests/test_merge.py
ukmo-ccbunney: That's a good point. Updated with more rigorous test: e9551164e

lib/iris/tests/unit/merge/test_merge.py
trexfeathers, line 1: Thanks for converting/migrating this! Unfortunately I think the content ought to be split into two destinations:

- `tests/integration/merge/test_merge.py`: The stuff that relies on `iris.load()` and/or uses CML testing
- `tests/unit/merge/test_merge.py`: the other stuff that is making its own cubes on the fly

trexfeathers (changes requested):
Awesome work @ukmo-ccbunney. Just one change left I think

lib/iris/tests/unit/merge/test_merge.py
ukmo-ccbunney: Good point. I'm on it!

ukmo-ccbunney: Done; 802eeca61

PR #6481, opened 2025-05-28
Pp derived bounds

pp-mo:
Separate PR direct to main
cf https://github.com/HGWright/iris/pull/4


HGWright (approved):
Thanks Patrick, it was a big collaborative effort.

PR #6483, opened 2025-05-29
Pp derived bounds  mergedpaths

pp-mo:
Proposed logic rework to #6481 
For discussion
Only here, as a separate PR, to check if it broke any other tests...

(Update : added some further ideas (in followon commits) :
  * replace contextlib.suppress
  * replace try/except
  * improved(?) varnames + comments
 
)


pp-mo:
OK tests look good, 
so I hope that validates the logic of changes in https://github.com/SciTools/iris/pull/6483/commits/b7dd7bce017d2c2f579df5a5f60f17dc95ac23e6

pp-mo:
Benchmark is also clean !
This covers the whole change, including all commits since @HGWright started the work on this

PR #6504, opened 2025-06-10
Lockfile update plus testing fix

trexfeathers:
### Description

Closes #6497 - this the same commits plus a minor fix to a PyTest fixture. I couldn't find anything suspect in the release notes for PyTest 8.4, especially since the error was specific to Python 3.13, but regardless this is better coding practice AND fixes the error so no worries 🙂

---

---


ESadek-MO (approved):
Thanks @trexfeathers, seems to work, I shan't look a gifthorse in the mouth! 

I might investigate this afterwards, but I think this is good to go as is.

Issue #6505, opened 2025-06-11
Loading files with coordinates that have non-string units fails


schlunma:
## How To Reproduce
Steps to reproduce the behaviour:

Download the file [`n2oFlux-Yang2020.nc`](https://darchive.mblwhoilibrary.org/bitstreams/c407fd4e-0fb9-52ba-93b8-5188994d02dd/download) with `time:units = 1. ;`.

<details>
<summary>ncdump -h n2oFlux-Yang2020.nc</summary>

```
netcdf n2oFlux-Yang2020 {
dimensions:
        lat = 720 ;
        lon = 1440 ;
        time = 12 ;
variables:
        double lat(lat) ;
                lat:standard_name = "latitude" ;
                lat:long_name = "latitude" ;
                lat:units = "degrees north" ;
                lat:_CoordinateAxisType = "Lat" ;
        double lon(lon) ;
                lon:standard_name = "longitude" ;
                lon:long_name = "longitude" ;
                lon:units = "degrees north" ;
                lon:_CoordinateAxisType = "Lon" ;
        double time(time) ;
                time:long_name = "Time variable" ;
                time:units = 1. ;
                time:_CoordinateAxisType = "Time" ;
        double latitude_2d(lat, lon) ;
                latitude_2d:_FillValue = NaN ;
                latitude_2d:standard_name = "latitude_2d" ;
                latitude_2d:long_name = "center-cell latitude" ;
                latitude_2d:units = "degrees north" ;
        double longitude_2d(lat, lon) ;
                longitude_2d:_FillValue = NaN ;
                longitude_2d:standard_name = "longitude_2d" ;
                longitude_2d:long_name = "center-cell longitude" ;
                longitude_2d:units = "degrees east" ;
        double cellArea_m2(lat, lon) ;
                cellArea_m2:_FillValue = NaN ;
                cellArea_m2:standard_name = "cellArea_m2" ;
                cellArea_m2:long_name = "cell area in m^2" ;
                cellArea_m2:units = "m^2" ;
        double n2oFlux_EnsMean_g-pm2-pyr(time, lat, lon) ;
                n2oFlux_EnsMean_g-pm2-pyr:_FillValue = NaN ;
                n2oFlux_EnsMean_g-pm2-pyr:standard_name = "n2oFlux_EnsMean_g-pm2-pyr" ;
                n2oFlux_EnsMean_g-pm2-pyr:long_name = "n2o flux -- ensemble mean prediction" ;
[...]

PR #6506, opened 2025-06-11
Fix loading files with coordinates that have non-string units

schlunma:
### Description

The function [`_is_lat_lon`](https://github.com/schlunma/iris/blob/588e48c0851f4193c4f113bdd8257e832d8e6283/lib/iris/fileformats/_nc_load_rules/helpers.py#L1793) fails with an `AttributeError` if the coordinate variable has non-string units.

This PR adapts this function so it can handle this case.

Fixes https://github.com/SciTools/iris/issues/6505

---

---


bjlittle (approved):
@schlunma LGTM 👍

PR #6508, opened 2025-06-12
Make the `Cube` class available from Iris root

### LINKED ISSUE #891: Proposal: move iris.cube.Cube to iris.Cube
I find myself constructing a lot of cubes and cubelists manually when I implement new functions to manipulate cubes.

Since these are core API objects, I suggest they should be moved from `iris.cube.Cube` and `iris.cube.CubeList` to `iris.Cube` and `iris.CubeList` (via an import statement in `/lib/iris/__init__.py`). This would be consistent with many other Python libraries (e.g., [Pandas](http://pandas.pydata.org/pandas-docs/stable/api.html)), for which the core API  is also usually exposed at the top level.

As a library user, it's just easier not to need to know about the structure of Iris's code to find a class or function. I can see a case for consolidating other functions into a broader namespace as well, but Cube and CubeList are the worse offenders (in my view). In my view, exposing submodules in a library's API should be done with caution, and only when it is clearly easier for library users.

trexfeathers:
### Description

Closes #891

- [x] Circular imports - didn't happen
- [x] Documentation - importing, but not including in `__all__`, adds the convenience without polluting the documentation. Precedent comes from `geovista.GeoPlotter` ([undocumented](https://geovista.readthedocs.io/en/latest/reference/generated/api/geovista/index.html)), which actually comes from `geovista.geoplotter.GeoPlotter` ([documented](https://geovista.readthedocs.io/en/latest/reference/generated/api/geovista/geoplotter/index.html#geovista.geoplotter.GeoPlotter))
- [ ] Performance regressions - tested locally, but just confirming via CI...

---

---

### DIFF
--- a/lib/iris/__init__.py
+++ b/lib/iris/__init__.py
@@ -104,6 +104,7 @@ from iris._combine import COMBINE_POLICY as _COMBINE_POLICY
 from iris._combine import CombineOptions
 import iris._constraints
 import iris.config
+from iris.cube import Cube
 import iris.io
 from iris.io import save
 from iris.loading import (


trexfeathers:
Other nominations from the team:

- `CubeList`
- `DimCoord`
- `AuxCoord`

